PR #1289, opened 2021-01-26
Quick tutorial on time series

herve-gross:
This is very brief, but it may be useful. Not sure this should be in the tutorials though. A "how to" section?


joshua-white:
@herve-gross, I agree, I suggest we add this in the User Guide somewhere, rather than as a tutorial.

rrsettgast:
@wrtobin Can you have a look at this and make any corrections that are required?

joshua-white:
I'd suggest we also restructure this as just a regular User Guide entry, rather than a Tutorial.  It can still be written "mostly" like a Tutorial, but maybe with a bit more comprehensive info.

rrsettgast:
@joshua-white @wrtobin Should this be merged?

PR #1291, opened 2021-01-26
Bugfix/sytuan/camel case minor

sytuannguyen:
The build was fail if GEOSX_PTP is set to OFF. This is because of some remaining function names that were not changed with respect to the camelCase rule. This PR is to solve this issue. Should check also other precompilation filters.

### DIFF
--- a/src/coreComponents/physicsSolvers/surfaceGeneration/SurfaceGenerator.cpp
+++ b/src/coreComponents/physicsSolvers/surfaceGeneration/SurfaceGenerator.cpp
@@ -633,9 +633,9 @@ int SurfaceGenerator::separationDriver( DomainPartition & domain,
 #else
 
     GEOSX_UNUSED_VAR( neighbors );
-    AssignNewGlobalIndicesSerial( nodeManager, modifiedObjects.newNodes );
-    AssignNewGlobalIndicesSerial( edgeManager, modifiedObjects.newEdges );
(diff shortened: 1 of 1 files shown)


src/coreComponents/physicsSolvers/surfaceGeneration/SurfaceGenerator.cpp
```diff
@@ -633,9 +633,9 @@ int SurfaceGenerator::separationDriver( DomainPartition & domain,
 #else
 
     GEOSX_UNUSED_VAR( neighbors );
-    AssignNewGlobalIndicesSerial( nodeManager, modifiedObjects.newNodes );
-    AssignNewGlobalIndicesSerial( edgeManager, modifiedObjects.newEdges );
-    AssignNewGlobalIndicesSerial( faceManager, modifiedObjects.newFaces );
+    assignNewGlobalIndicesSerial( nodeManager, modifiedObjects.newNodes );
+    assignNewGlobalIndicesSerial( edgeManager, modifiedObjects.newEdges );
+    assignNewGlobalIndicesSerial( faceManager, modifiedObjects.newFaces );
```
TotoGaz, line 638: Thx @sytuannguyen for the fix. Obviously some issue with the PTP submodule. Otherwise it's OK for me.

TotoGaz:
Hi @sytuannguyen I've done another PR with the modifications, so you can forget about it ;)

PR #1299, opened 2021-01-27
Correct the variable definitions with respect to the constant rule

sytuannguyen:
This PR homogenizes the variable definitions with respect to the constant rule. 
For example all the definitions "const string", "string const", "std::string const" and "const std::string" are homogenized to
"std::string const". Similar correction can be easily done for other major variable types such as real64, bool, int and Integer (that should be also homogenized).

The succesful build was verified and executable was tested!


TotoGaz (commented):
You are changing the code here, it's not only naming conventions.

This `string` is not defined as a standard string, it's a `geosx::string` which happens to be defined as a `std::string`. leaving us the possibility to use `std::wstring` in a single switch.
Personally I would be ok to use `std::string` everywhere, but we could also decide to use `geosx::string` everywhere instead.
If we do not use `geosx::string`, did you suppress its definition in `DataTypes.hpp`?
While being tempting, you must refrain your self from mixing purposes in you PR, specially if it's so huge that it will be difficult to control everything.

And last, we are 2 days after the huge naming conventions patch that everyone did not merge yet. Do we want to merge such an other huge PR?

rrsettgast:
@sytuannguyen This is not how we want things to look. We should not be using `std::string` in GEOSX proper, but instead use the alias.

PR #1301, opened 2021-01-28
Replace std::string with string

rrsettgast:
This PR replaces instances of `std::string` with the `string` alias inside of GEOSX.

### DIFF
--- a/src/coreComponents/codingUtilities/StringUtilities.cpp
+++ b/src/coreComponents/codingUtilities/StringUtilities.cpp
@@ -29,7 +29,7 @@ namespace stringutilities
 /**
  * String tokenizing function
  **/
-string_array Tokenize( const std::string & str, const std::string & delimiters )
+string_array Tokenize( const string & str, const string & delimiters )
 {
   string_array tokens;
 
@@ -41,7 +41,7 @@ string_array Tokenize( const std::string & str, const std::string & delimiters )
   {
 
     bool usesNonWhitespaceDelimiters = false;
-    std::string::size_type i =0;
+    string::size_type i =0;
     while( delimiters[i] && !usesNonWhitespaceDelimiters )
     {
       usesNonWhitespaceDelimiters |= !isspace( int(delimiters[i]) );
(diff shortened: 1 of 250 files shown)


TotoGaz:
You may be willing to remove the `#include <string>` too? There are still a bunch of them.

rrsettgast:
> You may be willing to remove the `#include <string>` too? There are still a bunch of them.

Oh. Yes.

TotoGaz:
There are a couple of occurrences of `std::to_string`; don't know what to do with it though.

src/coreComponents/codingUtilities/StringUtilities.cpp
```diff
@@ -29,7 +29,7 @@ namespace stringutilities
 /**
  * String tokenizing function
  **/
-string_array Tokenize( const std::string & str, const std::string & delimiters )
+string_array Tokenize( const string & str, const string & delimiters )
```
sytuannguyen, line 32: You may want to change also "const string" to "string const" as the rule said:

image

sytuannguyen:
Do you want to take also a look at the types: int and integer? Are they both related to the alias localIndex?

src/coreComponents/codingUtilities/StringUtilities.cpp
TotoGaz: Position of `const` is not restrained to `string`. It will probably be more efficient to deal with `const` position all at once, for all the argument types.

rrsettgast: Yes....we should do the const enforcement in a stand-alone PR.

rrsettgast:
> Do you want to take also a look at the types: int and integer? Are they both related to the aliases localIndex/globalIndex?

If they are actually a local or global index, then that they should be `localIndex` or `globalIndex`. Otherwise if they are an integer data type in the repository, they should be `integer`. If they are in a kernel, they should be `int` unless it is an index into an `ArrayView`

TotoGaz (approved):

rrsettgast:
> There are a couple of occurrences of `std::to_string`; don't know what to do with it though.

I think we keep them. If we ever change the `string` alias to something other than `std::string` we will have to make sure there is a UDC to `std::string`.

PR #1304, opened 2021-01-29
Fix to boundedPlane to avoid missing some embedded elements.

CusiniM:
This PR fixes an issue with the check of whether a point belong to the bounded plane or not. The tolerance was too tight causing issues in some cases.


rrsettgast (approved):

ishovkun (approved):
This is exactly what I needed!

PR #1305, opened 2021-01-29
Improve logs when product is not found in catalog.

TotoGaz:
The user will receive a more comprehensive message in case of typo or error in the input file.

Another example of issue https://github.com/GEOSX/GEOSX/issues/705

### DIFF
--- a/src/coreComponents/codingUtilities/StringUtilities.cpp
+++ b/src/coreComponents/codingUtilities/StringUtilities.cpp
@@ -26,6 +26,15 @@ namespace geosx
 namespace stringutilities
 {
 
+string toLower( string const & input )
+{
+  string output;
+  output.resize( input.size() );
+  auto const toLowerCase = []( unsigned char c ) { return std::tolower( c ); };
+  std::transform( input.cbegin(), input.cend(), output.begin(), toLowerCase );
+  return output;
+}
+
 /**
  * String tokenizing function
  **/
--- a/src/coreComponents/codingUtilities/StringUtilities.hpp
+++ b/src/coreComponents/codingUtilities/StringUtilities.hpp
@@ -19,13 +19,13 @@
 #ifndef GEOSX_CODINGUTILITIES_STRINGUTILITIES_HPP_
 #define GEOSX_CODINGUTILITIES_STRINGUTILITIES_HPP_
 
-#include <cxxabi.h>
+#include <algorithm>
 #include <cstring>
+#include <cxxabi.h>
+#include <iomanip>
+#include <map>
 #include <memory>
 #include <sstream>
- #include <iomanip>
-#include <algorithm>
-#include <map>
 
 
 #include "common/DataTypes.hpp"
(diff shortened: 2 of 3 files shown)


joshua-white (approved):
I think this will be super useful.

herve-gross (approved):
Nice! A good way to catch XML errors, with good guidance on how to correct these issues.

TotoGaz:
@rrsettgast do you want to merge your `string` PR first?

TotoGaz:
> A good way to catch XML errors

Actually I thought we already had something like this...

> I think this will be super useful.

It may save hours and hours.

rrsettgast (approved):
@TotoGaz Can you show the error message as it is printed out in the case of an incorrect key lookup?

TotoGaz:
> @TotoGaz Can you show the error message as it is printed out in the case of an incorrect key lookup?

It displays this
```
***** ERROR
***** LOCATION: ../coreComponents/dataRepository/ObjectCatalog.hpp:190
***** Rank 0: Could not find "Banana" in this catalog. Please be sure that all your keywords are properly spelled or that input file parameters have not changed.
All available keys are: [
Dirichlet,
FieldSpecification,
SourceFlux
]
```
or 
```
***** ERROR
***** LOCATION: ../coreComponents/dataRepository/ObjectCatalog.hpp:190
***** Rank 0: Could not find "Banana" in this catalog. Please be sure that all your keywords are properly spelled or that input file parameters have not changed.
All available keys are: [
HaltEvent,
PeriodicEvent,
SoloEvent
]
```

rrsettgast:
@TotoGaz This looks great to me...but my only concerns are:
1.  whether a user should know what a "Catalog" is.
2. the dependency between `ObjectCatalog.hpp` and `StringUtilites.hpp`. More to the point the circular dependency that is being introduced between `dataRepository` and `codingUtilites`?

TotoGaz:
> 1. whether a user should know what a "Catalog" is. Perhaps add a message indicating that the incorrect keyword is specified in their input file?

I wanted to emphasize that an invalid keyword could be OK in other context.
As per using the catalog word, it's already in the error message just above.
I could say `Could not find keyword "Banana" for this context.`?

> 2. the dependency between `ObjectCatalog.hpp` and `StringUtilites.hpp`. More to the point the circular dependency that is being introduced between `dataRepository` and `codingUtilites`?

That could be troublesome indeed.
But code like `StringUtilites.hpp` in `codingUtilites` are more low-level utility c++ coding helpers.
Why would should such code depend on `dataRepository`?
How could a circular dependency _legitimately_ occur?

rrsettgast:
> > 1. whether a user should know what a "Catalog" is. Perhaps add a message indicating that the incorrect keyword is specified in their input file?
> 
> I wanted to emphasize that an invalid keyword could be OK in other context.
> As per using the catalog word, it's already in the error message just above.
> I could say `Could not find keyword "Banana" for this context.`?
> 
I am just emphasizing that user doesn't know what a "Catalog" is. We could just indicate that there could be an incorrect keyword in their input file in addition to the message.


> > 1. the dependency between `ObjectCatalog.hpp` and `StringUtilites.hpp`. More to the point the circular dependency that is being introduced between `dataRepository` and `codingUtilites`?
> 
> That could be troublesome indeed.
> But code like `StringUtilites.hpp` in `codingUtilites` are more low-level utility c++ coding helpers.
> Why would should such code depend on `dataRepository`?
> How could a circular dependency _legitimately_ occur?

I am mistaken. I was interrupted while reviewing and got crossed up.

TotoGaz:
@rrsettgast is the error message OK ?
`Could not find keyword "Banana" in this context.` and `"Banana" could be found. But the builder is invalid.`

TotoGaz:
yes of course ;)

PR #1307, opened 2021-01-30
Feature: the implementation of the improved version of Projection EDFM

ishovkun:
In this PR we implement a first version of the Projection EDFM to account for impermeable embedded fractures/faults.                                                                                                                                                                                                                        
 
Steps:
- Slightly refactored the EDFM-part of TwoPointFluxApproximation to split M-M and F-M connection computation;
- Added a ProjectionEDFMHelper that does all the pEDFM stuff;
- Added a pEDFM sinplePhase integratedTest;
- Added a pEDFM compositional multiphase integratedTest;
                                                                                                                                                                                                                                                                              
What works:
- The code has all the face selection, geometric proccessing, and transmissibility calculation

What's missing:
- No full projection, but assume that it occupies all the selected faces.
- No fracture-fracture intersections


src/coreComponents/finiteVolume/ProjectionEDFMHelper.hpp
```diff
@@ -0,0 +1,107 @@
...
+  /* This is a convenience struct to avoid passing too many argyuments (one object instead of three indices) */
+  struct CellID
+  {
+    CellID(localIndex r, localIndex sr, localIndex i)
+        : region(r), subRegion(sr), index(i)
+    {}
+
+    localIndex region;
+    localIndex subRegion;
+    localIndex index;
```
TotoGaz, line 32: We need more of there ;)

rrsettgast: Copy the triple index data into a struct for passing between functions called inside a kernel? Am I reading this correctly?

ishovkun: @rrsettgast Is passing three contiguous values of elementary types much more expensive than passing three separate entities of these elementary types/

ishovkun: Although I am passing it by reference, I still wonder.

ishovkun:
The code does not fail any more. 
Now what's left is to get rid of the hard-coded fracture permeability.

src/coreComponents/finiteVolume/ProjectionEDFMHelper.hpp
```diff
@@ -0,0 +1,94 @@
...
+
+class ProjectionEDFMHelper {
+ public:
+  ProjectionEDFMHelper( MeshLevel const & mesh,
+                        GeometricObjectManager const * geometricObjManager,
+                        std::string const & coeffName,
+                        CellElementStencilTPFA & stencil );
+
+  // add Fracture-matrix connections to the cell stencil
+  void addNonNeighboringConnections(EmbeddedSurfaceSubRegion const & fractureSubRegion);
```
CusiniM, line 17: I wonder if instead of creating a `ProjectionEDFMHelper` each time, it would be better to make it inherit from `Group` and register it as a `subGroup` of the `TwoPointFluxApproximation`. We could then also add a method to update the stencils whenever the aperture changes by grabbing it from the `TwoPointFluxApproximation`. 
To do so you would have to change the constructor though and populate the object in a separate method because I think that when the `TwoPointFluxApproximation` is generated a lot of the info needed is not available yet. 

@rrsettgast, @francoishamon, @klevzoff, what do you think?

francoishamon: Yes I agree with @cusiniM. If these helpers can be used in multiple places in the code and can be called every time the aperture changes, then it is worth trying to make the object persistent by registering a `ProjectionEDFMHelper` on the `TwoPointFluxApproximation` object. 
- Then, once all the views that you need in the helpers have been created, you would populate the registered object and copy all the views into the data members of `ProjectionEDFMHelper`. Then you would grab the `ProjectionEDFMHelper` object, and just call `addNonNeighboringConnections` and or the additional function that @cusiniM is mentioning.
- Another possibility is to remove all the data members from `ProjectionEDFMHelper` and just extend the list of input parameters in `addNonNeighboringConnections`. You would have to pass `mesh`, `geometricObjManager`, and `stencil` in addition to `fractureSubRegion` in the call to `addNonNeighboringConnections`. With this option you would not have to worry about populating the `ProjectionEDFMHelper` object.

If the object `ProjectionEDFMHelper` is not registered on the `TwoPointFluxApproximation` but still needs to be used in multiple places, there is always the option to make the functions in the class `ProjectionEDFMHelper` `static` and pass them all the parameters that they need everytime they are called. 

That being said, I am not sure about the best option here. If the `mesh`, `geometricObjManager`, and `stencil` objects are still available when/where you want to update the stencils after the aperture change, maybe there is no need to store them in the class?

klevzoff: I wouldn't go as far as making `ProjectionEDFMHelper` inherit from `Group` because it doesn't actually allocate and permanently store any data, just grabs some views to use. On the other hand, to grab those views it does a few key lookups, so we do want to cache those lookups within the object (with the caveat that views may get invalidated if the underlying arrays are resized - make sure this is not the case here).

I think the way @ishovkun has gone is to create a subobject of `FluxApproximation` (managed through a smart pointer, so it can be created optionally) but not register it as a subgroup. Seems like a reasonable compromise?

src/coreComponents/physicsSolvers/surfaceGeneration/EmbeddedSurfaceGenerator.cpp
```diff
@@ -213,6 +212,8 @@ real64 EmbeddedSurfaceGenerator::solverStep( real64 const & GEOSX_UNUSED_PARAM(
 
 void EmbeddedSurfaceGenerator::addToFractureStencil( DomainPartition & domain )
 {
+  GeometricObjectManager * geometricObjManager = domain.getParent()->getGroup< GeometricObjectManager >( "Geometry" );
```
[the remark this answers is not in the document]
ishovkun: done in 00f50f4

src/coreComponents/finiteVolume/ProjectionEDFMHelper.hpp
ishovkun: The initialization of ProjectionEDFMHelper is moved to TwoPointFluxApproximation::initializePreSubgroups (775d1d0).

@klevzoff, @francoishamon, @CusiniM Please let me know what y'all think.

PR #1311, opened 2021-02-02
Compositonal multiphase flow in fractured media.

CusiniM:


CusiniM:
@joshua-white @francoishamon @klevzoff @rrsettgast let me know what you think of this. For it to work with the poroelastic and teh hydrofracture solvers I need a few more modifications and I will wait for @castelletto1 's work on the poroelastic kernels.

src/coreComponents/finiteVolume/CellElementStencilTPFA.cpp
```diff
@@ -54,4 +57,23 @@ void CellElementStencilTPFA::add( localIndex const numPts,
...
+  m_cellToFaceVec.resize( newSize );
+  m_transMultiplier.resize( newSize );
+
+  m_transMultiplier[oldSize] = transMultiplier;
+
+  LvArray::tensorOps::copy< 3 >( m_faceNormal[oldSize], faceNormal );
+  for( localIndex a=0; a<2; a++ )
+  {
+    LvArray::tensorOps::copy< 3 >( m_cellToFaceVec[oldSize][a], cellToFaceVec[a] );
+  }
```
francoishamon, line 76: Do you have a mechanism somewhere to `reserve` before filling all these vectors?

src/coreComponents/finiteVolume/CellElementStencilTPFA.hpp
```diff
@@ -97,8 +164,108 @@ class CellElementStencilTPFA : public StencilBase< CellElementStencilTPFA_Traits
...
+  real64 const arithmeticWeight = ( halfTrans[0]+halfTrans[1] )/ 2;
+
+  real64 const meanPermCoeff = 1.0; //TODO make it a member if it is really necessary
+
+  real64 const value = meanPermCoeff * harmonicWeight + (1 - meanPermCoeff) * arithmeticWeight;
+  for( localIndex ke = 0; ke < 2; ++ke )
+  {
+    transmissibility[ke] = m_transMultiplier[iconn] * value * (ke == 0 ? 1 : -1);
+  }
+}
```
francoishamon, line 245: Would it be feasible/desirable/easier to compute the derivatives with respect to pressure in this function, instead of in `dTrans_dPressure`, since you will need the half-transmissibilities to compute them?

Also, do you plan to keep the averaging between `harmonicWeight` and `arithmeticWeight` for the `CellElementSubRegion`?

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBaseKernels.hpp
```diff
@@ -312,10 +310,9 @@ struct AccumulationKernel
           arrayView1d< globalIndex const > const & dofNumber,
           arrayView1d< integer const > const & elemGhostRank,
           arrayView1d< real64 const > const & volume,
-          arrayView1d< real64 const > const & porosityOld,
-          arrayView1d< real64 const > const & porosityRef,
-          arrayView2d< real64 const > const & pvMult,
-          arrayView2d< real64 const > const & dPvMult_dPres,
+          arrayView2d< real64 const > const & porosityOld,
+          arrayView2d< real64 const > const & porosityNew,
+          arrayView2d< real64 const > const & dPoro_dPres,
```
francoishamon, line 315: It is great that you were able to get rid of `pvMult` with your new developments!

src/coreComponents/physicsSolvers/fluidFlow/FlowSolverBaseKernels.hpp
```diff
@@ -0,0 +1,134 @@
...
+          arrayView2d< real64 const > const & porosity )
+  {
+    forAll< POLICY >( size, [=] GEOSX_HOST_DEVICE ( localIndex const k )
+    {
+      for( localIndex q = 0; q < permWrapper.numGauss(); ++q )
+      {
+        permWrapper.update( k, q, porosity[k][q] );
+      }
+    } );
+  }
```
francoishamon, line 76: Do you anticipate future permeability models that will not be written as a function of porosity only, and do you think it will be easy to modify this if we encounter a model that does not fit? Recently, we were came across two use cases for which the highlighted lines above may need some changes:
- Permeability (and porosity) tabulated directly as a function of pressure. For permeability, this is probably not something that we want inside GEOSX (?)
- Permeability tabulated as a function of strain

src/coreComponents/physicsSolvers/fluidFlow/integratedTests/singlePhaseFlow/compressible_1d.xml
```diff
@@ -92,48 +99,18 @@
...
-      name="rock"
+    <PressureDependentPorosity
+      name="porosityModel"
       referencePressure="0.0"
+      defaultReferencePorosity ="0.2"
       compressibility="1e-9"/>
+      
+    <ConstantPermeability
+      name="permeabilityModel"
+      permeabilityComponents="1e-12, 1e-12, 1e-12"/>
```
francoishamon, line 110: These new XML tags look good to me. It would be nice to keep the possibility to import permeabilities and porosities embedded inside external meshes (but I know there is already a lot of work to this PR :) )

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseHybridFVM.cpp
```diff
@@ -302,6 +303,23 @@ void SinglePhaseHybridFVM::assembleFluxTerms( real64 const GEOSX_UNUSED_PARAM( t
...
+  GEOSX_UNUSED_VAR( dt );
+  GEOSX_UNUSED_VAR( domain );
+  GEOSX_UNUSED_VAR( dofManager );
+  GEOSX_UNUSED_VAR( localMatrix );
+  GEOSX_UNUSED_VAR( localRhs );
+
+  GEOSX_ERROR("Not yet implemented");
+
+}
+
```
francoishamon, line 322: If you want to leave these functions not implemented in this PR, feel free, I will be glad to implement them in a follow-up PR. From what I have seen in your PR, nothing will prevent the new constitutive model for permeability from being used in the hybrid flow solvers.

francoishamon (commented):
@CusiniM This PR looks very good to me so far. You have done a lot of work already! If you feel that the amount of code needed to implement everything for both single-phase and compositional flow is going to be large, maybe we should find ways to split this into multiple PRs? To handle separately poroelasticity? Or hybrid solvers? Let me know if I can help.

src/coreComponents/constitutive/permeability/ParallelPlatesPermeability.hpp
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+
+  struct viewKeyStruct : public PermeabilityBase::viewKeyStruct
+  {
+    static constexpr char const * dPerm_dApertureString() { return "dPerm_dAperture"; }
+  } viewKeys;
+
+private:
+  array3d< real64 > m_dPerm_dAperture;
```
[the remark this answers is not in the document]
CusiniM: Yes, any suggestion about how to do this would be very useful. It would be great if the solver could throw an error if someone tried to use a constitutive model that does not make sense. 

The reason why I am not registering all variables in the base class and I am not enforcing a generic `update ` function is to avoid carrying around useless data. For example, I think that if we run a flow-only problem there should not be the need for storing derivatives of the permeability w.r.t. to mechanics unknowns. As a matter of fact I was trying to make the flow solver completely agnostic of all mechanics related quantities.

src/coreComponents/constitutive/porosity/PorosityBase.cpp
```diff
@@ -0,0 +1,73 @@
...
+namespace geosx
+{
+
+using namespace dataRepository;
+
+namespace constitutive
+{
+
+
+PorosityBase::PorosityBase( string const & name, Group * const parent ):
```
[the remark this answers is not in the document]
CusiniM: I don't think the update is necessary for porosity since I don't see much common interfaces. 

Yes, I was planning to have a poroelastic model that computes porosity as a function of the strain and of the pressure.  This will have its own update class and will be used directly in the SinglePhasePoroelasticKernel to substitute https://github.com/GEOSX/GEOSX/blob/2de5fd5961cbe3d1e4c155ecb04b3c5e47dea06d/src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoroelasticKernel.hpp#L242-254 .

src/coreComponents/finiteVolume/CellElementStencilTPFA.cpp
CusiniM: Unless the stencils were already doing it for all their fields I have not implemented anything specific myself.

src/coreComponents/finiteVolume/CellElementStencilTPFA.hpp
CusiniM: 1. Yes, it would be certainly feasible. I kind of like the idea to keep functions to do only one task and I also thought that eventually we will need a functions to compute `dTrans_dDisp` and so I thought it would be cleaner to have 2 separate functions. Putting everything in a single function would not be hard though.

2. I am no sure. I saw that the arithmetic av. was there for the fractures and so I kept it but I m not sure why someone would want to use it. If we think it s useless I don't mind removing it.

src/coreComponents/physicsSolvers/fluidFlow/FlowSolverBaseKernels.hpp
CusiniM: So, my plan was to add at least an interface for the poroelastic case in which the permeability is not a function of the porosity but of the strain. I would have a separate interface for this case though. I think that it makes sense to assume that for flow-only problems the strain is not registered and does not exist. Basically I was planning on having some `PoroelasticFlowKernels` to assemble fluxes and update `permeability` and `porosity` for the `poroelastic` case. I know that it means having some more code rewriting but I think that it keeps our assembly kernels much cleaner and allows us to avoid polluting simple tpfa flow kernels with things that are only relevant for the coupled problems.

src/coreComponents/physicsSolvers/fluidFlow/integratedTests/singlePhaseFlow/compressible_1d.xml
CusiniM: So, it is certainly possible to use the `FieldSpecification` to set the permeability but I am not sure how to import the field from and external mesh. It is certainly something we need to be able to do though. If not in this PR we should address it right after.

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseHybridFVM.cpp
CusiniM: Yes, I won't implement this. I just want to make create all the interfaces to make sure that the design works well. You can have fun with this later on.

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoroelasticKernel.hpp
```diff
@@ -293,9 +274,9 @@ class SinglePhase :
 
     for( integer a = 0; a < numNodesPerElem; ++a )
     {
-      stack.localDispFlowJacobian[a*3+0][0] += dNdX[a][0] * m_biotCoefficient * detJxW;
-      stack.localDispFlowJacobian[a*3+1][0] += dNdX[a][1] * m_biotCoefficient * detJxW;
-      stack.localDispFlowJacobian[a*3+2][0] += dNdX[a][2] * m_biotCoefficient * detJxW;
+      stack.localDispFlowJacobian[a*3+0][0] += dNdX[a][0] * dPorosity_dVolStrainIncrement * detJxW;
+      stack.localDispFlowJacobian[a*3+1][0] += dNdX[a][1] * dPorosity_dVolStrainIncrement * detJxW;
+      stack.localDispFlowJacobian[a*3+2][0] += dNdX[a][2] * dPorosity_dVolStrainIncrement * detJxW;
```
castelletto1, line 279: Here we need the following derivative: `dTotalStress_dPressure`.

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoroelasticKernel.hpp
```diff
@@ -318,9 +304,9 @@ class Multiphase :
 
     for( integer a = 0; a < numNodesPerElem; ++a )
     {
-      stack.localDispFlowJacobian[a*3+0][0] += dNdX[a][0] * m_biotCoefficient * detJxW;
-      stack.localDispFlowJacobian[a*3+1][0] += dNdX[a][1] * m_biotCoefficient * detJxW;
-      stack.localDispFlowJacobian[a*3+2][0] += dNdX[a][2] * m_biotCoefficient * detJxW;
+      stack.localDispFlowJacobian[a*3+0][0] += dNdX[a][0] * dPorosity_dVolStrainIncrement * detJxW;
+      stack.localDispFlowJacobian[a*3+1][0] += dNdX[a][1] * dPorosity_dVolStrainIncrement * detJxW;
+      stack.localDispFlowJacobian[a*3+2][0] += dNdX[a][2] * dPorosity_dVolStrainIncrement * detJxW;
```
castelletto1, line 309: Here we need the following derivative: `dTotalStress_dPressure`.

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoroelasticKernel.hpp
CusiniM: true, but aren't the two derivatives the same? I thought they were and didn't want to create and extra variable but correct me if I am wrong.

CusiniM:
This PR is now ready for review. The only thing that needs to be decided is what to do about the solid model refactoring. Based on what is decided in https://github.com/GEOSX/GEOSX/discussions/1422 the solid models should be changed either in this PR or in a separate one.

src/coreComponents/physicsSolvers/multiphysics/SinglePhaseReservoir.cpp
```diff
@@ -64,12 +64,12 @@ void SinglePhaseReservoir::setupSystem( DomainPartition & domain,
...
-  {
-    SinglePhaseFVM< SinglePhaseBase > * fvmSolver = dynamicCast< SinglePhaseFVM< SinglePhaseBase > * >( m_flowSolver );
-    fvmSolver->setUpDflux_dApertureMatrix( domain, dofManager, localMatrix );
-  }
+//  // we need to set the dR_dAper CRS matrix in SinglePhaseFVM to handle the presence of fractures
+//  if( dynamicCast< SinglePhaseFVM< SinglePhaseBase > * >( m_flowSolver ) )
+//  {
+//    SinglePhaseFVM< SinglePhaseBase > * fvmSolver = dynamicCast< SinglePhaseFVM< SinglePhaseBase > * >( m_flowSolver );
+//    fvmSolver->setUpDflux_dApertureMatrix( domain, dofManager, localMatrix );
+//  }
```
klevzoff, line 72:
```suggestion
```

klevzoff (commented):
Solid work! Haven't gone through the whole thing yet, but here are a few comments

src/coreComponents/constitutive/solid/CoupledSolid.hpp
```diff
@@ -157,57 +153,28 @@ class CoupledSolid : public ConstitutiveBase
...
           typename PERM_TYPE >
 void CoupledSolid< SOLID_TYPE, PORO_TYPE, PERM_TYPE >::initializePreSubGroups()
 {
-  if( SOLID_TYPE::catalogName() != getSolidModel().getCatalogName() )
-  {
-    GEOSX_ERROR( " The coupled solid "<<this->getName()<<
-                 " expects a solid model of type "<<SOLID_TYPE::catalogName()<<
-                 " but the specified solid model \""<<m_solidModelName<<
-                 "\" is of type" << getSolidModel().getCatalogName() );
-  }
```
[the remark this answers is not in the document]
CusiniM: I actually don't remember why  I did this. I will move it back.

src/coreComponents/constitutive/solid/porosity/PorosityBase.hpp
```diff
@@ -121,10 +132,10 @@ class PorosityBase : public ConstitutiveBase
 
   struct viewKeyStruct : public ConstitutiveBase::viewKeyStruct
   {
-    static constexpr char const * newPorosityString() { return "newPorosity"; }
+    static constexpr char const * newPorosityString() { return "porosity"; }
```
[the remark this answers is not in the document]
CusiniM: I wanted to keep the same naming convention as for the `stress`. So, I am ok with changing it but I would then change it for both.

src/coreComponents/finiteVolume/CellElementStencilTPFA.hpp
```diff
@@ -49,14 +49,102 @@ struct CellElementStencilTPFA_Traits
...
+class CellElementStencilTPFAWrapper : public StencilWrapperBase< CellElementStencilTPFA_Traits >,
+  public CellElementStencilTPFA_Traits
+{
+public:
+
+  template< typename VIEWTYPE >
+  using ElementViewConst = ElementRegionManager::ElementViewConst< VIEWTYPE >;
+
+  template< typename VIEWTYPE >
+  using PermeabilityViewAccessor = ElementRegionManager::MaterialViewAccessor< VIEWTYPE >;
```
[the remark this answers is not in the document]
CusiniM: this sounds good!

src/coreComponents/constitutive/solid/porosity/PorosityBase.hpp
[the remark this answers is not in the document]
klevzoff: I seem to vaguely remember a similar discussion happening around stress, indeed. I'm sure there were good reasons to keep the name `newStress`. For me personally, the `new`/`old` monikers only make sense inside the update function; once the update is done, and for the majority of the code, "new" stress/porosity is just the current value - I would maybe actually prefer `curr`/`prev` over `new`/`old` (though I've introduced a lost of `old` variables in the past myself). But consistency across models is also important.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVMKernels.cpp
```diff
@@ -231,18 +234,21 @@ INST_PhaseMobilityKernel( 5, 3 );
...
 FluxKernel::
   compute( localIndex const numPhases,
            localIndex const stencilSize,
+           localIndex const numFluxElems,
            arraySlice1d< localIndex const > const seri,
            arraySlice1d< localIndex const > const sesri,
            arraySlice1d< localIndex const > const sei,
-           arraySlice1d< real64 const > const stencilWeights,
+           real64 const (&transmissibility)[2],
+           real64 const (&dTrans_dPres)[2],
```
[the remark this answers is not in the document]
CusiniM: I agree, although I don't love how the `SurfaceElementStencil` computes the transmissibility between multiple elements (more than 2). We should use Mohammad's method.

src/coreComponents/physicsSolvers/fluidFlow/FlowSolverBase.hpp
```diff
@@ -186,6 +163,44 @@ class FlowSolverBase : public SolverBase
 
 };
 
+template< typename POROUSWRAPPER_TYPE >
+void execute1( POROUSWRAPPER_TYPE porousWrapper,
```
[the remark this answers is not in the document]
CusiniM: suggestion for the name? I can for sure move them to the `cpp` file.  They are only here coz NVCC didn't like to have nested lambdas.

src/coreComponents/physicsSolvers/fluidFlow/ProppantTransportKernels.cpp
```diff
@@ -824,53 +824,144 @@ void FluxKernel::
...
-                                ElementViewConst< arrayView1d< real64 const > > const & dSettlingFactor_dProppantConc,
-                                ElementViewConst< arrayView2d< real64 const > > const & dSettlingFactor_dComponentConc,
-                                ElementViewConst< arrayView1d< real64 const > > const & collisionFactor,
-                                ElementViewConst< arrayView1d< real64 const > > const & dCollisionFactor_dProppantConc,
-                                ElementViewConst< arrayView1d< integer const > > const & isProppantMobile,
-                                ElementViewConst< arrayView1d< real64 const > > const & proppantPackVf,
-                                ElementViewConst< arrayView1d< real64 const > > const & aperture,
-                                CRSMatrixView< real64, globalIndex const > const & localMatrix,
-                                arrayView1d< real64 > const & localRhs )
+  launch< EmbeddedSurfaceToCellStencil >( EmbeddedSurfaceToCellStencil const & GEOSX_UNUSED_PARAM( stencil ),
```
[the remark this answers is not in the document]
CusiniM: I think I had to add them because the compiler was not happy otherwise but I can try to remove them and check again.

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseFVM.cpp
```diff
@@ -18,9 +18,12 @@
 
 #include "SinglePhaseFVM.hpp"
 
+#include "../multiphysics/SinglePhasePoromechanicsFluxKernels.hpp"
```
klevzoff, line 21: Should be
```suggestion
#include "physicsSolvers/multiphysics/SinglePhasePoromechanicsFluxKernels.hpp"
```
and moved down in the list of includes

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseFVM.cpp
```diff
@@ -29,6 +32,7 @@
 #include "fieldSpecification/FieldSpecificationManager.hpp"
 #include "physicsSolvers/fluidFlow/SinglePhaseBaseKernels.hpp"
 #include "physicsSolvers/fluidFlow/SinglePhaseFVMKernels.hpp"
+#include "physicsSolvers/solidMechanics/SolidMechanicsEmbeddedFractures.hpp"
```
klevzoff, line 35: This looks unused
```suggestion
```

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.hpp
```diff
@@ -175,6 +175,44 @@ class SinglePhaseBase : public FlowSolverBase
...
+   * @param time_n previous time value
+   * @param dt time step
+   * @param domain the physical domain object
+   * @param dofManager degree-of-freedom manager associated with the linear system
+   * @param localMatrix the system matrix
+   * @param localRhs the system right-hand side vector
+   * @param jumpDofKey dofKey of the displacement jump
+   */
+  virtual void
+  assemblePoroelasticFluxTerms( real64 const time_n,
```
klevzoff, line 189: These things look a bit hacky, but I don't have a concrete refactoring suggestion at the moment. Unless somebody has one, let's just merge as is and refactor later.

klevzoff (approved):
This looks great and ready to go IMO (as soon as CI is passed)

PR #1319, opened 2021-02-12
Adding asyncronous kernel launching and restructuring the CommunicationTools routines to expose more async to allow for latency hiding

wrtobin:


src/coreComponents/managers/initialization.cpp
```diff
@@ -110,10 +110,14 @@ void addUmpireHighWaterMarks()
...
-    GEOSX_LOG_RANK_0( "Umpire " << std::setw( 15 ) << allocatorName << " high water mark: " <<
+    std::size_t const maxMark = MpiWrapper::max( mark );
+    GEOSX_LOG_RANK_0( "Umpire " << std::setw( 15 ) << allocatorName << " total high water mark: " <<
                       std::setw( 9 ) << LvArray::system::calculateSize( totalMark ) );
+    GEOSX_LOG_RANK_0( "Umpire " << std::setw( 15 ) << allocatorName << " max high water mark: " <<
+                      std::setw( 9 ) << LvArray::system::calculateSize( maxMark ) );
 
-    pushStatsIntoAdiak( allocatorName + " high water mark", mark );
+    pushStatsIntoAdiak( allocatorName + " total high water mark", mark );
+    pushStatsIntoAdiak( allocatorName + " max high water mark", mark );
```
wrtobin, line 120: I can remove this or restructure it to make the output a bit cleaner, it is useful in the packing scaling since the max pinned memory directly correlates to our packing data.

wrtobin:
Need to return/collect the RAJA events coming from the async forall launches and poll them for completion instead of just device synchronizing to be fully correct IIRC, that will take some restructuring to collect the active events and poll them at a higher call level than the device packing functions. 

Also need to either (a) retain a synchronous packing for all the other locations we pack in code since those are now using the async packing without fully accounting for that, or (b) restructure all of our packing calls around the async packing with the events changes.

wrtobin:
Also need to add host/omp versions to compile on quartz, etc.

wrtobin:
requires merging in https://github.com/GEOSX/GEOSX_PTP/pull/23

wrtobin:
Ah okay need to add support for machines without openmp or raja, shouldn't be too bad.

src/coreComponents/mesh/ElementRegionManager.cpp
```diff
@@ -272,6 +272,7 @@ ElementRegionManager::PackPrivate( buffer_unit_type * & buffer,
   packedSize += bufferOps::Pack< DOPACK >( buffer, this->getName() );
   packedSize += bufferOps::Pack< DOPACK >( buffer, numRegions() );
 
+  parallelDeviceEvents noEvents;
```
rrsettgast, line 275: make sure this is empty?

src/coreComponents/mesh/ElementRegionManager.cpp
```diff
@@ -332,6 +328,7 @@ int ElementRegionManager::unpackPrivate( buffer_unit_type const * & buffer,
   localIndex numRegionsRead;
   unpackedSize += bufferOps::Unpack( buffer, numRegionsRead );
 
+  parallelDeviceEvents noEvents;
```
rrsettgast, line 331: error check empty?

src/coreComponents/mpiCommunications/CommunicationTools.cpp
```diff
@@ -748,21 +750,45 @@ void CommunicationTools::synchronizePackSendRecvSizes( const std::map< string, s
...
-  MPI_iCommData sizeComm;
+  if( onDevice )
+  {
+    localIndex eventCount = 0;
+    for( const auto & sa : fieldNames )
+    {
+      eventCount += sa.second.size( );
+    }
+    eventCount *= neighbors.size( );
+    events.reserve( eventCount );
```
rrsettgast, line 769: we are still packing each field independently?

src/coreComponents/mpiCommunications/CommunicationTools.hpp
```diff
@@ -110,24 +111,48 @@ class CommunicationTools
   void synchronizeFields( const std::map< string, string_array > & fieldNames,
                           MeshLevel & mesh,
                           std::vector< NeighborCommunicator > & allNeighbors,
-                          bool on_device = false );
+                          bool onDevice );
```
rrsettgast, line 114: Yay....removal of default parameters is always a good thing!

src/coreComponents/mpiCommunications/MpiWrapper.cpp
```diff
@@ -174,6 +174,90 @@ int MpiWrapper::test( MPI_Request * request, int * flag, MPI_Status * status )
   return 0;
 }
 
+int MpiWrapper::testany( int count, MPI_Request array_of_requests[], int * idx, int * flag, MPI_Status array_of_statuses[] )
```
rrsettgast, line 177:
```suggestion
int MpiWrapper::testAny( int count, MPI_Request array_of_requests[], int * idx, int * flag, MPI_Status array_of_statuses[] )
```

src/coreComponents/mpiCommunications/MpiWrapper.cpp
```diff
@@ -174,6 +174,90 @@ int MpiWrapper::test( MPI_Request * request, int * flag, MPI_Status * status )
...
+{
+#ifdef GEOSX_USE_MPI
+  return MPI_Testany( count, array_of_requests, idx, flag, array_of_statuses );
+#endif
+  *flag = 0;
+  return 0;
+}
+
+int MpiWrapper::testall( int count, MPI_Request array_of_requests[], int * flag, MPI_Status array_of_statuses[] )
```
rrsettgast, line 186:
```suggestion
int MpiWrapper::testAll( int count, MPI_Request array_of_requests[], int * flag, MPI_Status array_of_statuses[] )
```

src/coreComponents/mpiCommunications/MpiWrapper.cpp
```diff
@@ -174,6 +174,90 @@ int MpiWrapper::test( MPI_Request * request, int * flag, MPI_Status * status )
...
+int MpiWrapper::check( MPI_Request * request, int * flag, MPI_Status * status )
+{
+#ifdef GEOSX_USE_MPI
+  return MPI_Request_get_status( *request, flag, status );
+#endif
+  *flag = 0;
+  return 0;
+}
+
+int MpiWrapper::checkany( int count, MPI_Request array_of_requests[], int * idx, int * flag, MPI_Status array_of_statuses[] )
```
rrsettgast, line 204:
```suggestion
int MpiWrapper::checkAny( int count, MPI_Request array_of_requests[], int * idx, int * flag, MPI_Status array_of_statuses[] )
```

src/coreComponents/mpiCommunications/MpiWrapper.cpp
```diff
@@ -174,6 +174,90 @@ int MpiWrapper::test( MPI_Request * request, int * flag, MPI_Status * status )
...
+  {
+    *flag = flagCache;
+  }
+  return rval;
+#endif
+  *flag = 0;
+  return 0;
+}
+
+int MpiWrapper::checkall( int count, MPI_Request array_of_requests[], int * flag, MPI_Status array_of_statuses[] )
```
rrsettgast, line 235:
```suggestion
int MpiWrapper::checkAll( int count, MPI_Request array_of_requests[], int * flag, MPI_Status array_of_statuses[] )
```

src/coreComponents/mpiCommunications/MpiWrapper.cpp
```diff
@@ -182,10 +266,10 @@ int MpiWrapper::wait( MPI_Request * request, MPI_Status * status )
   return 0;
 }
 
-int MpiWrapper::waitany( int count, MPI_Request array_of_requests[], int * indx, MPI_Status * status )
+int MpiWrapper::waitany( int count, MPI_Request array_of_requests[], int * indx, MPI_Status array_of_statuses[] )
```
rrsettgast, line 269: of course...camelcase isn't really enforced?

src/coreComponents/mpiCommunications/NeighborCommunicator.cpp
```diff
@@ -289,9 +289,10 @@ inline int GhostSize( NodeManager & nodeManager, arrayView1d< localIndex const >
   bufferSize += edgeManager.packUpDownMapsSize( edgeAdjacencyList );
   bufferSize += faceManager.packUpDownMapsSize( faceAdjacencyList );
   bufferSize += elemManager.PackUpDownMapsSize( elementAdjacencyList );
-  bufferSize += nodeManager.packSize( {}, nodeAdjacencyList, 0 );
-  bufferSize += edgeManager.packSize( {}, edgeAdjacencyList, 0 );
-  bufferSize += faceManager.packSize( {}, faceAdjacencyList, 0 );
+  parallelDeviceEvents noEvents;
```
rrsettgast, line 292: error check empty

rrsettgast (approved):
looks pretty good. only minor comments

src/coreComponents/mpiCommunications/MpiWrapper.cpp
wrtobin: Think this is a holdover from when we were implementing the mpiwrapper in a way that mimics the MPI interface closely minus the MPI_ prefix. When we did the large camelCase restructuring across the codebase that changed in this file e.g. Comm_rank to commRank, so we're not mimicking the mpi interfaces as closely anymore (there is actually a comment in the wrapper.cpp referring to the old methodology which I'm removing in my followup PR).

src/coreComponents/mpiCommunications/CommunicationTools.cpp
wrtobin: Yeah we're still launching packing kernels for each field for each type ('node','edge','face','elem') for each neighbor.

The sizing on this is pulled inside the neighbor communicator and more precise (this doesn't include the type count) in the follow-up I'm currently prepping.

src/coreComponents/managers/initialization.cpp
corbett5: Maybe rename it to like "sum across ranks" and "rank max" or something.

src/coreComponents/mpiCommunications/CommunicationTools.cpp
corbett5: If this loop is only about pre-allocating enough space, I wouldn't bother with it. Calling `emplace_back` a handful (100s) of times on an empty vector or array1d should pale in comparison with either kernel launches or MPI calls, not to mention both.

src/coreComponents/mpiCommunications/CommunicationTools.cpp
```diff
@@ -777,28 +803,97 @@ void CommunicationTools::synchronizePackSendRecv( const std::map< string, string
...
+void CommunicationTools::finalizeUnpack( MeshLevel & mesh,
+                                         std::vector< NeighborCommunicator > & neighbors,
+                                         MPI_iCommData & icomm,
+                                         bool onDevice,
+                                         parallelDeviceEvents & events )
+{
+  GEOSX_MARK_FUNCTION;
+
+  // wait to unpack any remaining buffers
+  while( !asyncUnpack( mesh, neighbors, icomm, onDevice, events ) )
```
corbett5, line 892: I'd add some comment here or some mention that we have a spinning loop here. If we were to say launch this function asynchronously it would be terrible because it would be wasting tons of cycles.

src/coreComponents/mpiCommunications/NeighborCommunicator.cpp
corbett5: I see a bunch of these comments, error check that `noEvents` contains no events? That seems like something that should be in a unit tests and not here. Like each time you allocate an `array1d` you don't check that it's `size() == 0`.

corbett5 (approved):

src/coreComponents/mpiCommunications/NeighborCommunicator.cpp
corbett5: Ahhh I see, and so Randy wanted to check check that they were empty after passing to the functions.

src/coreComponents/mpiCommunications/MpiWrapper.cpp
rrsettgast: ok

src/coreComponents/mpiCommunications/NeighborCommunicator.cpp
rrsettgast: Wait...did I miss something? We expect `noEvents` to be empty, and if it isn't empty, then the events will be async'ed. However we were expecting nothing out of `noEvents`. I don't know that this is actually a problem, but it seems like we should enforce the intent.

What is the difference between "check" if their empty from "check check"?

src/coreComponents/mpiCommunications/CommunicationTools.cpp
rrsettgast: Do we have other options besides a while loop? It seems like there is a while loop here, and inside `asyncUnpack` there is one as well. I always prefer for loops with some high bound and a `break` to `while` loops if there is any danger of spinning.

corbett5: I'd agree that a for-loop is probably better. Ideally we'd check and then sleep for some time before checking again, but that would involve some thought to pick a good amount of time to sleep for and a max timeout. It doesn't matter now, but it might in the future.

src/coreComponents/mpiCommunications/NeighborCommunicator.cpp
wrtobin: Don't think you missed anything, no. It's just a preference between a (currently) erroneous call to `waitAllDeviceEvents` and an error check against the size of noEvents. Either will guard against a future error.

wrtobin:
Comments should be all resolved by this commit with the exception of changing the async wait loop from a while to a for.

I implemented an `GEOSX_ASYNC_WAIT` macro as both a `for` and `while` version and even at a 60s wait ( with polling every ~10 nanosec ) in the `for` version we still got some fails in the integrated tests. The async wait is used underneath all the `CommunicationTools` packing functionality (even those that are externally synchronous) and the timeout on an async wait is very related to the underlying latencies and amount of data moving around for the op, haven't looked into the particular fails because there were several and they were nondeterministic ( depending on machine/network load while running the tests presumably ).

I'd like to have a timeout, but determining a realistic expectation for the different use scenarios and finding the max expected wait time would take some doing.

src/coreComponents/managers/ObjectManagerBase.hpp
```diff
@@ -362,7 +362,7 @@ class ObjectManagerBase : public dataRepository::Group
 
   /**
    * @brief Extract map from object and assign global indices.
-   * @param nodeManager The instance.
+   * @param nodeManager The node manager.
```
wrtobin: Thanks for catch this ___conflict___ and making me do a merge, git.

rrsettgast:
> Comments should be all resolved by the last commit with the exception of changing the async wait loop from a `while` to a `for`.
> 
> I implemented a `GEOSX_ASYNC_WAIT` macro as both a `for` and `while` version and even at a 60s wait ( with polling every ~10 nanosec ) in the `for` version we still got some fails in the integrated tests. The async wait is used underneath all the `CommunicationTools` packing functionality (even those that are externally synchronous) and the timeout on an async wait is very related to the underlying latencies and amount of data moving around for the op, haven't looked into the particular fails because there were several and they were nondeterministic ( depending on machine/network load while running the tests presumably ).
> 
> I'd like to have a timeout on the async wait, but determining a realistic expectation for the different use scenarios and finding the max expected wait time would take some doing.

OK. So the while loop passes all the tests? Perhaps just merge it and tackle the for loop in a follow on PR. Tag it as "ready to merge" when you are good with merging.

src/coreComponents/rajaInterface/GEOS_RAJA_Interface.hpp
```diff
@@ -98,6 +150,26 @@ template< typename POLICY >
...
+    }
+  }
+  return allDone;
+}
+
+RAJA_INLINE void waitAllDeviceEvents( parallelDeviceEvents & events )
+{
+  // poll device events for completion then wait 10 nanoseconds 6,000,000,000 times (60 sec timeout)
+  // 10 nsecs ~= 30 clock cycles @ 3Ghz
+  GEOSX_ASYNC_WAIT( 6000000000, 10, testAllDeviceEvents( events ) );
```
corbett5, line 171: Why not just wait on each event in a loop? With CUDA events this will call `cudaEventSynchronize(m_event)`, which should be much more efficient (possibly even waiting for an OS signal). The only downside is that we don't have a timeout period, but I think that's more than worth the tradeoff.

wrtobin: Ah yeah using builtin blocking waits in the waitall would probably be better. I probably just defaulted to building the blocking on top of the nonblocking implementation.

I'll probably put that in the follow-up which has a couple other items for cleanup as well.

wrtobin:
@rrsettgast This is fine to merge, the only failures are the submodule (expected due to GEOSX/GEOSX_PTP#23) and the return check, though all the commands in there seem to be passing so I'm not sure whats going on with that check.

PR #1321, opened 2021-02-15
Cleanup transverse isotropic constitutive class

sytuannguyen:
The class LinearElasticTransverseIsotropic is not properly written as you can see in the screenshot below ( the copypaste "Elastic Shear Modulus Parameter" is not modified yet !!! )

image

This PR aims to clean this up as well as other stuff that I found. For example Poisson becomes PoissonsRatio to be consistent with YoungsModulus where the "s" is present here I think because of the term "Young's Modulus", etc.

I profit this occation to add also a possibility to define five elastic stiffnesses cij in the input xml (this is an additional option, the old option of defining the five engineering parameters i.e. Et, Ea, ... remains valid). It is sometime convenient to define these stiffnesses because they can be directly obtained from sonic logs.

image


src/coreComponents/constitutive/solid/LinearElasticTransverseIsotropic.hpp
```diff
@@ -375,17 +389,16 @@ class LinearElasticTransverseIsotropic : public SolidBase
...
-    m_defaultPoissonTransverse = input;
+    m_defaultPoissonsRatioTransverse = input;
   }
 
-
   /**
    * @brief Getter for default axial Poisson's ratio
    * @return The value of the default axial/transverse Poisson's modulus.
    */
   real64 getDefaultPoissonsRatioAxialTransverse()
```
corbett5, line 399: This should be `const`.

src/coreComponents/constitutive/solid/LinearElasticTransverseIsotropic.cpp
```diff
@@ -42,91 +40,128 @@ LinearElasticTransverseIsotropic::LinearElasticTransverseIsotropic( string const
...
-    setDescription( "Elastic Shear Modulus Parameter" );
+    setDescription( "Elastic Axial-Transverse Poisson's ratio Parameter" );
 
   registerWrapper( viewKeyStruct::defaultShearModulusAxialTransverse, &m_defaultShearModulusAxialTransverse )->
     setApplyDefaultValue( -1 )->
     setInputFlag( InputFlags::OPTIONAL )->
-    setDescription( "Elastic Shear Modulus Parameter" );
+    setDescription( "Elastic Axial-Transverse (out-plane) Shear Modulus Parameter" );
+
+  registerWrapper< real64 >( viewKeyStruct::defaultC11 )->
```
corbett5, line 65: I know we're inconsistent about this, but I would prefer to have almost all sub-views and sub-groups be members, and then register them with the repository. This way we do less lookup and it makes debugging substantially easier. There are cases (such as below) where you operate on the `Wrapper` instead of the wrapped object.

```c++
  this->getWrapper< array1d< real64 > >( viewKeyStruct::c66 )->
    setApplyDefaultValue( c66Default );
    setApplyDefaultValue( c66 );
```

So maybe we should instead of having the wrapped object (`array1d< real64 >`) as a member, instead have the `Wrapper< array1d< real64 > >`. This would require adding a `Group::registerWrapper( KEY_TYPE const &, WrapperBase * )` overload but it should be pretty straightforward. This would allow for the maximum debuggability.

 @rrsettgast thoughts?

joshua-white:
@sytuannguyen, can you get this caught up with develop?  I also propose we leave PoissonRatio as is, but instead use YoungModulus.  That way we are consistent with BiotCoefficient, etc.

sytuannguyen:
> @sytuannguyen, can you get this caught up with develop? I also propose we leave PoissonRatio as is, but instead use YoungModulus. That way we are consistent with BiotCoefficient, etc.

Agree

src/coreComponents/constitutive/solid/LinearElasticTransverseIsotropic.hpp
sytuannguyen: const is now added to all the get... functions in this class

sytuannguyen:
> @sytuannguyen, can you get this caught up with develop? I also propose we leave PoissonRatio as is, but instead use YoungModulus. That way we are consistent with BiotCoefficient, etc.

@joshua-white I skip the change YoungsModulus and YoungsMod to YoungModulus and YoungMod for another PR because it touches a list of files including unitTests and documents. I found also youngs, etc. the 's' appears systematically.

sytuannguyen:
@joshua-white  this one passed also, please have a look

sytuannguyen:
@joshua-white I will also merge this PR to #1323 to facilitate the update with develop, also to change systematically YoungsM... to YoungM... as you suggested.

PR #1320, opened 2021-02-15
Correct the isotropic Poisson's ratio range

sytuannguyen:
This PR corrects the range of the isotropic Poisson's ratio in LinearElasticIsotropic constitutive model. 

The isotropic Poisson's ratio is actually limited to a wrong range: nu >= 0. 
It is now corrected to the valid range: -0.5 < nu < 0.5

The range of the elastic bulk, shear and Young moduli are also corrected from >=0 to >0 to avoid the division to zero. For example, the calculation of the Poisson's ratio will not accept the value K = 0:

nu = 0.5 * (1 - E / 3 / K ) 

You can find also some minor cleanup

PR #1323, opened 2021-02-16
Elastic Orthotropic constitutive model

sytuannguyen:
The Elastic Orthotropic constitutive model is important for many applications. For example sandstone or fractured rock mass ( that are sensitive to stress state in the earth crust which is orthotropic) can exhibit significant orthotropic anisotropy. However, this constitutive solver is not implemented yet. Using the existing fully anisotropic model for this problem is not optimal because we need to deal with 36 parameters in place of 9 parameters.

This PR aims to fill this gap. We can define either nine engineering elastic constants obtained from triaxial tests (i.e. E1, E2, E3, nu12, etc.) or nine independent elastic stiffnesses c11, c22, ... that can be computed directly from anisotropic sonic velocities (and density).

This PR include also some corrections of the ElasticIsotropic and ElasticTransverseIsotropic classes that are initially addressed in the PRs #1320 and #1321

Rebaseline PR: https://github.com/GEOSX/integratedTests/pull/148


joshua-white:
@sytuannguyen, this seems very useful.  can you get it caught up to develop, passing checks, and in a state to merge?

sytuannguyen:
@joshua-white After uncrustify as you suggested, now it passes. Please take a look.

joshua-white (approved):
Hey @sytuannguyen, this looks good.  Do you pass all the integrated tests?  What platforms have you tested on?  Will this require rebaselining the integrated tests?

sytuannguyen:
@joshua-white I verified that the xml that is modified was not included in the integratedTests. So rebaseline the integratedTests is not required.

rrsettgast:
@joshua-white Can you merge this if it fits in your plan?

PR #1225, opened 2020-11-07
Add energy balance equation to Compositional solver.

CusiniM:
This original content of the PR has been split into multiple pieces that have been implemented separately:
- PR 1: Add a model for solid internal energy  (https://github.com/GEOSX/GEOSX/pull/1660)
- PR 2: Add the enthalpy and internal energy functions to the CO2-brine fluid model (https://github.com/GEOSX/GEOSX/pull/1689)
- PRs 3, 4, 5: Refactor accumulation, flux, and property kernels (https://github.com/GEOSX/GEOSX/pull/1696, https://github.com/GEOSX/GEOSX/pull/1672, and https://github.com/GEOSX/GEOSX/pull/1698)
- PR 6: Add thermal conductivity constitutive model (https://github.com/GEOSX/GEOSX/pull/1732)
- PR 7: Reduce the size of the fluid kernel wrapper (https://github.com/GEOSX/GEOSX/pull/1779)
- PR 8: Put everything together in a `CO2BrineThermal` fluid model (https://github.com/GEOSX/GEOSX/pull/1755)
- This PR: Add the energy balance equation to the compositional solver

This PR implements the kernel interfaces that add the thermal derivatives and assemble the energy balance (internal energy term, enthalpy flux, conduction flux). 

- [X] Unit test (finite difference derivatives)
- [X] Integrated test (CO2-brine simulation)

Rebaseline in https://github.com/GEOSX/integratedTests/pull/213

PR #1325, opened 2021-02-17
Implement CO2Brine model on GPU

francoishamon:
This PR modifies the existing CO2-brine model to make it run on the GPU. The main modifications are:
- The use of `TableFunction` to replace the custom tables implemented in `src/coreComponents/constitutive/fluid/PVTFunctions/UtilityFunctions.hpp`.
- The use of hand-written derivatives to replace the automatic differentiation.
- The use of wrappers (update classes) for each "PVT function" to replace the shared pointers and avoid virtual calls in the compute functions.
- [X] Produces the same results as with the "old" implementation on Quartz.
- [X] Works on GPU
- [X] Remove the old implementation.
- [x] Templatize the class on the PVTFunction classes.
- [x] Coordinate with Matteo: this PR is going to break the thermal PR.

Related to https://github.com/GEOSX/integratedTests/pull/125


src/coreComponents/constitutive/fluid/NewMultiPhaseMultiComponentFluid.hpp
```diff
@@ -0,0 +1,870 @@
...
+  PVTProps::FenghourCO2Viscosity::KernelWrapper const m_co2ViscosityWrapper;
+
+  /// Kernel wrapper for brine density updates
+  PVTProps::BrineCO2Density::KernelWrapper const m_brineDensityWrapper;
+
+  /// Kernel wrapper for CO2 density updates
+  PVTProps::SpanWagnerCO2Density::KernelWrapper const m_co2DensityWrapper;
+
+  /// Kernel wrapper for phase fraction and phase component fraction updates
+  PVTProps::CO2Solubility::KernelWrapper const m_co2SolubilityWrapper;
```
francoishamon, line 225: Hello @klevzoff,

On Lassen, I struggled with the data members in the update class of the new CO2-brine model (currently called `NewMultiPhaseMultiComponentFluid`):

1) In the current version of this PR, the data members are the wrappers of the PVTFunctions, such as, for instance, `PVTProps::BrineViscosity::KernelWrapper const m_brineViscosityWrapper;`. But, when I try to compile on Lassen, I get a compilation error saying:
```
Error: Formal parameter space overflowed
```
located where we launch the [fluid update](https://github.com/GEOSX/GEOSX/blob/11dffd4c2b3ad2e9716d86b53e5e8b929c36aaf5/src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFlowKernels.hpp#L197) (unfortunately I forgot to save the exact error message, I do this in another branch when I remove the non-GPU callable fluid models from the `CMakeLists.txt`...). I found a [thread online](https://forums.developer.nvidia.com/t/formal-parameter-space-overflowed-4096-bytes-max/134513/3) saying that this error might be due to the way I pass the data to the GPU, so I tried 
something else (see point 2 below).

2) If I modify the update class to store views on the wrappers of the PVTFunctions:
```
 /// Kernel wrapper for brine viscosity updates                                                                                              
  arrayView1d< PVTProps::BrineViscosity::KernelWrapper const > const m_brineViscosityWrapper;                                                 
                                                                                                                                              
  /// Kernel wrapper for CO2 viscosity updates                                                                                                
  arrayView1d< PVTProps::FenghourCO2Viscosity::KernelWrapper const > const m_co2ViscosityWrapper;                                             
                                                                                                                                              
  /// Kernel wrapper for brine density updates                                                                                                
  arrayView1d< PVTProps::BrineCO2Density::KernelWrapper const > const m_brineDensityWrapper;                                                  
                                                                                                                                              
  /// Kernel wrapper for CO2 density updates                                                                                                  
  arrayView1d< PVTProps::SpanWagnerCO2Density::KernelWrapper const > const m_co2DensityWrapper;                                               
                                                                                                                                              
  /// Kernel wrapper for phase fraction and phase component fraction updates                                                                  
  arrayView1d< PVTProps::CO2Solubility::KernelWrapper const > const m_co2SolubilityWrapper;
```
then I can compile and run the CO2-brine simulations on Lassen, and everything seems to work fine.
However, this forces me to have these `arrayView1d`s of size 1 in the update class, which is a little odd. What do you think would be a good way to solve this issue?

klevzoff: According to https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#function-parameters the total size of kernel arguments must not exceed 4 KB. We haven't hit this limit before, but it seems this object may indeed be quite large (if you add up the sizes of all `TableKernelWrapper`s and other members in its subobjects). Out of curiosity, can you print out `sizeof(NewMultiPhaseMultiComponentFluidUpdate)` somewhere?

An `ArrayView` of size 1 may be (much) smaller than its contained value, so I'm not surprised that it helps. This may be a reasonable workaround if it's not possible to reduce the size of the object. The only problem is that you also need to place the underlying `Array` object somewhere (inside `NewMultiPhaseMultiComponentFluid`?). I suppose we could also develop a host-device version of `unique_ptr` to provide a slightly lighter interface to manage a single object across memory spaces. 

@corbett5 @rrsettgast thoughts?

corbett5: Interesting. Yeah I'm not a huge fan of the `arrayView1D` containing a single object. A `unique_ptr` would be cleaner, but it wouldn't much easier to use than an array containing a single value. It would still need an owning class somewhere and a view class.

I wonder if captured variables count towards the parameter limit.

klevzoff: > I wonder if captured variables count towards the parameter limit.

I think this is exactly what happens here. The captured-by-value object is a member of the capture object and contributes to its size. I wonder if RAJA could work around this issue on their side by dynamically allocating device memory for the device-side copy of the lambda, rather than passing it directly by-value through the CUDA launch API (conditionally, only when `sizeof(LABMDA)` above limit). Might be worth opening an issue with them.

francoishamon: 1) When I do `sizeof(NewMultiPhaseMultiComponentFluidUpdate)` before launching the kernel on the branch that has the size-1 `arrayView1d`s (see [here](https://github.com/GEOSX/GEOSX/blob/36311100493ee0a1f2a45acfb84e13ef2c0eeac4/src/coreComponents/constitutive/fluid/NewMultiPhaseMultiComponentFluid.hpp#L212-L225)), I get: 2.336 KB and it seems to work well.

2) When I try to compile on Lassen with the current version of this PR (no `arrayView1d` to store the `KernelWrapper`s), I get the following error message that, I think, contains the size of the kernel (5.920 KB):
```
/usr/WS1/GEOS/GEOSX/TPLs_2021-01-19/install-lassen-clang@upstream-release/raja/include/RAJA/policy/cuda/forall.hpp(137): 
Error: Formal parameter space overflowed (5920 bytes required, max 4096 bytes allowed) in function 
_ZN4RAJA6policy4cuda4impl18forall_cuda_kernelILm256ENS_9Iterators16numeric_iteratorIllPlEEZN5geosx34CompositionalMultiphaseFlowKernels17FluidUpdateKernel6launchINS1_9cuda_execILm256ELb0EEENS8_12constitutive38NewMultiPhaseMultiComponentFluidUpdateEEEvlRKT0_RKN7LvArray9ArrayViewIKdLi1ELi0ElNSJ_10ChaiBufferEEESP_dRKNSK_ISL_Li2ELi1ElSM_EEEUllE_lEEvT1_SG_T2_
```
And yes I forgot to say that for the version of 1) with the size-1 `arrayView1d`s I keep an `array1d` in `NewMultiPhaseMultiComponentFluid` [here](https://github.com/GEOSX/GEOSX/blob/36311100493ee0a1f2a45acfb84e13ef2c0eeac4/src/coreComponents/constitutive/fluid/NewMultiPhaseMultiComponentFluid.hpp#L335-L348).

src/coreComponents/constitutive/fluid/PVTFunctions/NewBrineCO2DensityFunction.cpp
```diff
@@ -0,0 +1,173 @@
...
+    P += dP;
+  }
+
+  real64 T = TStart;
+  while( T <= TEnd )
+  {
+    coordinates[1].emplace_back( T );
+    T += dT;
+  }
+
```
TotoGaz, line 111: See below

src/coreComponents/constitutive/fluid/PVTFunctions/NewBrineCO2DensityFunction.cpp
```diff
@@ -0,0 +1,173 @@
...
+  bool notFound = 1;
+  for( localIndex i = 0; i < componentNames.size(); ++i )
+  {
+    if( componentNames[i] == "CO2" || componentNames[i] == "co2" )
+    {
+      m_CO2Index = i;
+      notFound = 0;
+      break;
+    }
+  }
```
TotoGaz, line 48: see below

TotoGaz (commented):
Some stupid shallow comments

src/coreComponents/constitutive/fluid/NewMultiPhaseMultiComponentFluid.cpp
```diff
@@ -0,0 +1,182 @@
...
+  }
+  GEOSX_ERROR_IF( notFound, "Phase water/liquid is not found!" );
+
+  createPVTModels();
+}
+
+void NewMultiPhaseMultiComponentFluid::createPVTModels()
+{
+  // 1) Create the viscosity and density models
+  for( string & filename : m_phasePVTParaFiles )
```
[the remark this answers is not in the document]
francoishamon: Done

src/coreComponents/constitutive/fluid/PVTFunctions/NewSpanWagnerCO2DensityFunction.cpp
```diff
@@ -0,0 +1,316 @@
...
+
+SpanWagnerCO2Density::SpanWagnerCO2Density( array1d< string > const & inputPara,
+                                            array1d< string > const & componentNames,
+                                            array1d< real64 > const & componentMolarWeight ):
+  PVTFunctionBase( inputPara[1],
+                   componentNames,
+                   componentMolarWeight )
+{
+
+  bool notFound = 1;
```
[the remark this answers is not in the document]
francoishamon: Good catch, done

src/coreComponents/constitutive/fluid/PVTFunctions/NewSpanWagnerCO2DensityFunction.cpp
```diff
@@ -0,0 +1,316 @@
...
+
+  constexpr real64 D[] = {0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 275, 275, 275};
+
+  constexpr real64 T_c = 304.1282;
+  constexpr real64 rho_c = 467.6;
+
+  constexpr real64 R = 188.9241;
+
+  real64 tau = T_c / T;
+  real64 delta = rho / rho_c;
```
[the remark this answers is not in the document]
francoishamon: Yes you are right, I added the `const`s

src/coreComponents/constitutive/fluid/PVTFunctions/NewPVTFunctionBase.hpp
```diff
@@ -0,0 +1,124 @@
...
+  PVTFunctionBase( string const & name,
+                   array1d< string > const & componentNames,
+                   array1d< real64 > const & componentMolarWeight )
+    :
+    m_functionName( name ),
+    m_componentNames( componentNames ),
+    m_componentMolarWeight( componentMolarWeight )
+  {}
+
+  virtual ~PVTFunctionBase() {}
```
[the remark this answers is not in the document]
francoishamon: I added the `= default` and removed the dummy implementation of the destructor.

src/coreComponents/constitutive/fluid/PVTFunctions/NewFlashModelBase.hpp
```diff
@@ -0,0 +1,124 @@
...
+
+  FlashModelBase( string const & name,
+                  array1d< string > const & componentNames,
+                  array1d< real64 > const & componentMolarWeight ):
+    m_modelName( name ),
+    m_componentNames( componentNames ),
+    m_componentMolarWeight( componentMolarWeight )
+  {}
+
+  virtual ~FlashModelBase() {}
```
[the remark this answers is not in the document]
francoishamon: same here

src/coreComponents/constitutive/fluid/PVTFunctions/NewSpanWagnerCO2DensityFunction.cpp
```diff
@@ -0,0 +1,316 @@
...
+
+  real64 theta, Delta, Phi, dPhi_delta, dDelta_delta, dDelta_delta_b;
```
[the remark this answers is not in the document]
francoishamon: Ahah, anywhere else in the code, I would agree with you. Here, I think we should keep these names (and all the other names in this very complex function) because they match exactly the names used in the paper that proposed to solve this equation (Span and Wagner, 1996). So for instance here, the `dDelta_delta` variable corresponds to a `\frac{\partial \Delta}{\partial \delta}` in the paper, and so on. If we change these names, it is going to be very difficult to map these variables to those of the paper. In the latest commit, I added to all these PVT functions the equation number or the table number from the journal papers that defined them.

src/coreComponents/constitutive/fluid/MultiFluidPVTPackageWrapper.hpp
```diff
@@ -155,13 +157,15 @@ class MultiFluidPVTPackageWrapperUpdate final : public MultiFluidBaseUpdate
...
 
+  GEOSX_HOST_DEVICE
   GEOSX_FORCE_INLINE
   virtual void update( localIndex const k,
                        localIndex const q,
                        real64 const pressure,
                        real64 const temperature,
                        arraySlice1d< real64 const > const & composition ) const override
   {
+#ifndef __CUDACC__
```
TotoGaz, line 168: I'm a bit confused by all these function becoming `void` when not using `CUDA`.
Do we actually intend them to be void? If the `GPU` codes goes through them, will the results be OK?

Could it be possible that we simply remove them fore `CUDA` by _e.g._ inheriting a specialised class that would be compiled iff we do `CUDA` directly in the `CMakeLists.txt`. Or by putting the `#ifndef __CUDACC__` outside of the functions so they do not exist when not using `CUDA`?

src/coreComponents/constitutive/fluid/PVTFunctions/BrineCO2DensityFunction.cpp
```diff
@@ -26,69 +29,57 @@ using namespace stringutilities;
...
-  real64 P, T, m;
-
-  dT = -1.0;
-  dP = -1.0;
-  TStart = -1.0;
-  TEnd = -1.0;
-  PStart = -1.0;
-  PEnd = -1.0;
+  array1d< array1d< real64 > > coordinates;
+  coordinates.resize( 2 );
```
TotoGaz, line 69: Would it make sense to build some `TableCoordinates class` that would whether inherit from `array1d< array1d< real64 > >` (🙄) or has a `array1d< array1d< real64 > >` attribute to bring more readability to your code.

E.g. the `coordinates.resize( 2 )` could be imposed in the constructor (instead of being done 4 different times in your code).
And we could replace `coordinates[0].size()` by `coordinates.nPressures()`, or `coordinates[0].emplace_back( P )` by `coordinates.appendPressure( P )` and `coordinates[1][j]` by `coordinates.getPressure( j )`. 
The indices `0` and `1` would be hidden when unnecessary (maybe always?) and we could still get no call cost by imposing `inline` (IIUC).

src/coreComponents/constitutive/fluid/PVTFunctions/BrineCO2DensityFunction.cpp
```diff
@@ -97,93 +88,46 @@ void BrineCO2DensityFunction::makeTable( string_array const & inputPara )
...
-
+  real64 P = PStart;
   while( P <= PEnd )
   {
-
-    pressures.emplace_back( P );
+    coordinates[0].emplace_back( P );
     P += dP;
-
   }
```
TotoGaz, line 103: Still a `for` loop. Maybe next time 😉 ?

src/coreComponents/constitutive/fluid/PVTFunctions/FenghourCO2ViscosityFunction.cpp
```diff
@@ -13,158 +13,152 @@
...
+
+  // equation (1) of Fenghour and Wakeham (1998)
+  viscosity = 1e-6 * (vlimit + vxcess + vcrit);
 }
 
-void FenghourCO2ViscosityFunction::calculateCO2Viscosity( real64_array const & pressure, real64_array const & temperature, real64_array2d const & density,
-                                                          real64_array2d const & viscosity )
+void FenghourCO2Viscosity::calculateCO2Viscosity( array1d< array1d< real64 > > const & coordinates,
+                                                  array1d< real64 > const & valuesDensity,
+                                                  array1d< real64 > const & valuesViscosity )
```
TotoGaz, line 137: How is `valuesDensity` and `valuesViscosity` differ from `densities` and `viscosities`.
I like the simple names most, but it's a matter of taste.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.cpp
```diff
@@ -110,30 +110,32 @@ void CompositionalMultiphaseFVM::assembleFluxTerms( real64 const dt,
...
+     MultiFluidBase const & fluid = getConstitutiveModel< MultiFluidBase >( subRegion, fluidModelNames()[targetIndex] );
+     arrayView4d< real64 const > const & phaseCompFrac = fluid.phaseCompFraction();
+     arrayView4d< real64 const > const & dPhaseCompFrac_dPres = fluid.dPhaseCompFraction_dPressure();
+     arrayView5d< real64 const > const & dPhaseCompFrac_dComp = fluid.dPhaseCompFraction_dGlobalCompFraction();
+
+     arrayView3d< real64 const > const & phaseMassDens = fluid.phaseMassDensity();
+     arrayView3d< real64 const > const & dPhaseMassDens_dPres = fluid.dPhaseMassDensity_dPressure();
+     arrayView4d< real64 const > const & dPhaseMassDens_dComp = fluid.dPhaseMassDensity_dGlobalCompFraction();
+
+     forAll< parallelDevicePolicy<> >( subRegion.size(),
```
TotoGaz, line 125: remove (we have `git`)?

TotoGaz (commented):
Comments no to forget them. (we'll forget them, but at least they are somewhere).

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.cpp
francoishamon: @TotoGaz  Thanks for the comments, I am going to address all of them next week. 

The modification above is temporary for my verification tests and the tests on Pangea3: since the fluid updates are done on device for CO2-brine, we do not need this piece of code that is manually moving to the GPU some arrays used in the flux computations (`phaseCompFrac` et `phaseMassDens`). As soon as we have the Pangea3 timings and we are ready to merge this PR, this will need to be uncommented back (integrated tests failures will remind us to uncomment this).

src/coreComponents/constitutive/fluid/MultiFluidPVTPackageWrapper.hpp
klevzoff: I feel like I should chime in since I proposed this. Let me briefly recap the discussion from https://github.com/GEOSX/GEOSX/pull/1313 (and sorry if I repeat points that are already clear).

  * The goal was, for a CUDA build (non-CUDA works as before), to allow different models to "opt-in" into different execution policies, such that models that aren't GPU-ready yet could still be run on the host.
  * To do this, each model declares (through a typedef) the policy it's ready to be run with, and the solver extracts the policy from the static type of the model (after constitutive dispatch) and uses it.
  * That by itself would work just fine. Problems only arise because we have used `virtual`/`override` keywords to enforce/document a common interface between these models. Even though we never use virtual dispatch, CUDA requires that all virtual/overriding functions share the same `host`/`device` designation.
  * Because **some** models want device code, **all** models now have to be marked `device`, which means `nvcc` will try to generate device code even for those functions that aren't device-ready, and will expectedly fail.
  * I saw two solutions to this: a) remove `virtual`/`override` that isn't mandatory; or b) make sure non-device-ready code is simply excluded from device compilation - but host version should still be generated - this is most easily achieved by just `#ifdef`-ing it out.
  * This function will never be called on GPU - the execution policy "opt-in" mechanism, if done correctly, makes sure of that. It becomes part of the class contract, if you will - the user **must** only use the policy the class presribes. So the function being a dummy is perfectly fine. 
  * Now, we *might* be able to `#ifdef` out the entire function definition as you suggest - I'm not sure if this will result in compiler or linker errors. Linker should be ok with this in theory, as the function is never called and thus doesn't need to be linked. Compiler should be ok in theory too - it does see the `override` declaration, so it knows the class claims implements all methods and can this be instantiated. But one can never be too sure with `nvcc`.
  * As for your other suggestion, I'm not sure I fully understood it - perhaps you can elaborate. I'll reiterate that we **want** this class to exist even in a CUDA build, and be usable on host, just not have compilation errors. And in the compilation model we use, any file that is part of the project is always compiled by `nvcc` in both host and device modes. The only way (that I'm aware of) to have those passes "sees" different code is through conditional compilation using `__CUDACC__` macro.

klevzoff: > Linker should be ok with this in theory, as the function is never called and thus doesn't need to be linked.

I take it back. Since the function is virtual, its address needs to be put in the vtable regardless of whether it's going to be called or not. Not providing a definition (at least a dummy one) leads to an `unresolved reference to vtable` linker error.

TotoGaz: Hi @klevzoff so IIUC the root cause is that we use an abstract base class to enforce a class signature for a template pattern. Where we would like to use c++20's future `concept` but it's not ready yet. (Will it work with `__host__`/`__device__` stuff is another question).

My concern is that dummy function are still generated and therefore could be used by somebody not mastering completely the pattern (or simply someone tired 🤷 , that happens).
As such there is a risk that these functions are being indeed called, providing wrong results that may (I don't know) be complicated to detect, specially if we are unlucky enough to give this to an end-user.

I could propose two options to try to keep the abstract base class pattern while improving safety.

First immediate solution would be not to accept such that pattern:
```c++
__something__ void foo(...){
#ifndef __CUDACC__
   bar();
#endif
}
```
but just force the program to die in the catastrophic case where this function would still be called (I say catastrophic because if we end up calling this function with `CUDA` activated there is something really really wrong and the whole simulation may be corrected).
```c++
__something__ void foo(...){
#ifdef __CUDACC__
   die("NotImplementedError, you should not be there!");
#else
  bar();
#endif
}
```

An other, (but maybe less pattern protective I don't know) would be to have this `die` call directly in the base class (which may not be pure abstract anymore, depending on the need). Therefore a non overriden member function which would still be called would kill the simulation, and the derived class would require no dummy function.
I do not know the whole hierarchy implications, so that may that's not possbile.

src/coreComponents/constitutive/fluid/PVTFunctions/NewFlashModelBase.hpp
```diff
@@ -0,0 +1,124 @@
...
+#include "codingUtilities/StringUtilities.hpp"
+#include "dataRepository/ObjectCatalog.hpp"
+
+namespace geosx
+{
+
+namespace PVTProps
+{
+
+class FlashModelBaseUpdate
```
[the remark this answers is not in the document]
francoishamon: Done

src/coreComponents/constitutive/fluid/PVTFunctions/FenghourCO2ViscosityFunction.cpp
francoishamon: I changed the names to `densities` and `viscosities`

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.cpp
francoishamon: I uncommented this piece of code (we still need it for non-GPU fluid models)

src/coreComponents/constitutive/unitTests/testMultiFluid.cpp
```diff
@@ -109,6 +111,16 @@ static const char * pvdo_str = "#P[Pa]\tBo[m3/sm3]\tVisc(Pa.s)\n"
 static const char * pvdw_str = "#\tPref[bar]\tBw[m3/sm3]\tCp[1/bar]\t    Visc[cP]\n"
                                "\t30600000.1\t1.03\t\t0.00000000041\t0.0003";
 
+// CO2-brine model
+
+static const char * pvtliquid_str = "DensityFun BrineCO2Density 1e6 1.5e7 5e4 94 96 1 0.2\n"
```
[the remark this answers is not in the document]
francoishamon: I changed the names to `PVTLiquidString`, `PVTGasString` in the other test file, but I just realized that I did not do it here. Next commit

src/coreComponents/constitutive/unitTests/testMultiFluid.cpp
```diff
@@ -561,11 +679,202 @@ TEST_F( DeadOilFluidTest, numericalDerivativesMass )
...
+    parent.initialize( &parent );
+    parent.initializePostInitialConditions( &parent );
+  }
+
+  ~MultiPhaseMultiComponentFluidTest()
+  {
+    removeFile( "pvtliquid.txt" );
+    removeFile( "pvtgas.txt" );
+    removeFile( "co2flash.txt" );
+  }
```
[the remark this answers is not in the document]
francoishamon: The lines that you referenced are in the "new" test format for fluid models, so I sticked to it in this PR. But, in the latest commit I changed the format in `testCO2BrinePVTModels.cpp` to remove `SetUp` and `TearDown` and use the new format instead. So everything is consistent now.

src/coreComponents/constitutive/fluid/MultiFluidPVTPackageWrapper.hpp
klevzoff: > First immediate solution would be not to accept that such pattern:
> ...
> but at least to force the program to die in the catastrophic case where this function would still be called. (I say catastrophic because if we end up calling this function with `CUDA` activated there is something really really wrong and the whole simulation may be flawed).

I understand now and completely agree - the function should throw/die when called in the unsupported execution space. Nice catch, and indeed extra protection like this is useful to have.

> An other, (but maybe less pattern protective I don't know) would be to have this `die` call directly in the base class (which may not be pure abstract anymore, depending on the need). Therefore a non overriden member function which would still be called would kill the simulation, and the derived class would require no dummy function.
> I do not know the whole hierarchy implications, so that may not be possbile.

So here the problem is we actually want to override `compute()` function in **all** derived classes - but some of those overrides want to be `__host__`-only and others `__host__ __device__`. The possibility to override in one execution space and have the other automatically call non-abstract base implementation would be great, but this scenario unfortunately isn't supported by `nvcc`: https://godbolt.org/z/cGP35E

src/coreComponents/constitutive/fluid/PVTFunctions/NewSpanWagnerCO2DensityFunction.cpp
```diff
@@ -0,0 +1,316 @@
...
+    coordinates[0].emplace_back( P );
+    P += dP;
+  }
+
+  real64 T = TStart;
+  while( T <= TEnd )
+  {
+    coordinates[1].emplace_back( T );
+    T += dT;
+  }
```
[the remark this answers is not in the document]
francoishamon: I changed this piece of code in this file and in all the other files to address your comment, now this looks like:
```
for( real64 P = PStart; P <= PEnd; P += dP )
{
  coordinates[0].emplace_back( P );
}
for( real64 T = TStart; T <= TEnd; T += dT )
{
  coordinates[1].emplace_back( T );
}
```

src/coreComponents/constitutive/fluid/PVTFunctions/NewCO2SolubilityFunction.cpp
```diff
@@ -0,0 +1,321 @@
...
+  bool notFound = 1;
+  for( localIndex i = 0; i < componentNames.size(); ++i )
+  {
+    if( componentNames[i] == "CO2" || componentNames[i] == "co2" )
+    {
+      m_CO2Index = i;
+      notFound = 0;
+      break;
+    }
+  }
```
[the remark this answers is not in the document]
francoishamon: I added a helper function to do find the component and phase names. It is now used in all the `PVTFunction`s, and in `MultiPhaseMultiComponentFluid`.

src/coreComponents/constitutive/fluid/PVTFunctions/PVTFunctionHelpers.hpp
```diff
@@ -0,0 +1,71 @@
...
+      {
+        if( input == expectedNames[j] )
+        {
+          id = i;
+          return true;
+        }
+      }
+    }
+    return false;
+  }
```
klevzoff, line 60: I recognize a standard algorithm in this:
```suggestion
  template<typename InputRange, typename ExpectedRange>
  static bool
  findName( InputRange const & inputNames,
            ExpectedRange const & expectedNames,
            localIndex & id )
  {
    using std::begin;
    using std::end;
    auto it = std::find_first_of( begin(inputNames), end(inputNames), begin(expectedNames), end(expectedNames) );
    id = std::distance( begin(inputNames), it );
    return it != end(inputNames);
  }
```
The template parameters are just so that you can mix and match `array1d`s, `std::vector`s and c-arrays (the last two have convenient list initialization, see the other suggestion above).
Moreover, I'm not a fan of passing an output parameter by reference, would rather have it be a return value. So either returning a `std::pair<bool, localIndex>`; or even just returning a `localIndex`, assigning it to the member variable and then asserting `0 <= m_compIndex && m_compIndex < componentNames.size()`.
(feel free to accept what you like of these suggestions, or nothing)

src/coreComponents/constitutive/fluid/PVTFunctions/BrineCO2DensityFunction.cpp
```diff
@@ -18,77 +18,62 @@
...
+  PVTFunctionBase( inputPara[1],
+                   componentNames,
+                   componentMolarWeight )
+{
+  array1d< string > expectedCO2ComponentNames;
+  expectedCO2ComponentNames.resize( 2 );
+  expectedCO2ComponentNames[0] = "CO2";
+  expectedCO2ComponentNames[1] = "co2";
+  bool found = PVTFunctionHelpers::findName( componentNames, expectedCO2ComponentNames, m_CO2Index );
+  GEOSX_ERROR_IF( !found, "Component CO2 is not found!" );
```
klevzoff, line 48: If you accept my other suggestion, this can become
```suggestion
  char const * expectedCO2ComponentNames[] = { "CO2", "co2" };
  m_CO2Index = PVTFunctionHelpers::findName( componentNames, expectedCO2ComponentNames );
  GEOSX_ERROR_IF( 0 <= m_CO2Index && m_CO2Index < componentNames.size(), "Component CO2 is not found!" );
```
(might need to deal with signed/unsigned comparison here)

TotoGaz (commented):
Still a concern about the indices that are hard coded (e.g. around the `inputPara` or in https://github.com/GEOSX/GEOSX/pull/1325#discussion_r584009994)

src/coreComponents/constitutive/fluid/PVTFunctions/BrineCO2DensityFunction.cpp
francoishamon: I added the class that you suggested here: https://github.com/GEOSX/GEOSX/blob/0c489b262db7da805c43439da199eaa7ef7aca63/src/coreComponents/constitutive/fluid/PVTFunctions/PVTFunctionHelpers.hpp#L54
so there are no `0` and `1` indices any more in the code you referenced and it is nicer.

src/coreComponents/constitutive/fluid/PVTFunctions/PVTFunctionHelpers.hpp
francoishamon: Here, I implemented what you suggested. I now return the `localIndex` that I assign to the member variable, and I use the check that you suggested as well.

src/coreComponents/constitutive/fluid/MultiFluidPVTPackageWrapper.hpp
francoishamon: Here I changed the code as suggested by @TotoGaz  to force the problem to crash if this function is called when it `__CUDA_ACC_` is defined.

src/coreComponents/constitutive/fluid/PVTFunctions/BrineCO2DensityFunction.cpp
francoishamon: Yes, here I implemented your two suggestions, thanks

src/coreComponents/constitutive/fluid/PVTFunctions/FenghourCO2ViscosityFunction.cpp
```diff
@@ -16,155 +16,149 @@
...
 {
-  for( localIndex i = 0; i < pressure.size(); ++i )
+  localIndex const numPressures = coordinates[0].size();
+  localIndex const numTemperatures = coordinates[1].size();
```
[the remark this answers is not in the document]
francoishamon: Oh, I forgot to bring the class that stores the coordinates here, I will change that to: `tableCoords.nPressures()`

src/coreComponents/constitutive/unitTests/testMultiFluid.cpp
```diff
@@ -561,11 +679,202 @@ TEST_F( DeadOilFluidTest, numericalDerivativesMass )
   testNumericalDerivatives( *fluid, parent, P, T, comp, eps, relTol, absTol );
 }
 
+MultiFluidBase * makeMultiPhaseMultiComponentFluid( string const & name, Group * parent )
+{
+  auto fluid = parent->registerGroup< NewMultiPhaseMultiComponentFluid >( name );
+
+  auto & compNames = fluid->getReference< string_array >( MultiFluidBase::viewKeyStruct::componentNamesString );
```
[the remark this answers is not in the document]
francoishamon: In the latest commit, I changed all the `array1d< string >` that I found into `string_array` for consistency (I may have forgotten a few)

src/coreComponents/constitutive/fluid/PVTFunctions/FenghourCO2ViscosityFunction.cpp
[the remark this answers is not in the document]
francoishamon: This is addressed now, I pass the `tableCoords` to this function and I use its getters `nPressures()` and `nTemperatures`

src/coreComponents/constitutive/fluid/PVTFunctions/FenghourCO2ViscosityFunction.cpp
```diff
@@ -16,155 +16,149 @@
...
-  TStart = -1.0;
-  TEnd = -1.0;
-  PStart = -1.0;
-  PEnd = -1.0;
+  real64 TStart = -1.0;
+  real64 TEnd = -1.0;
+  real64 dT = -1.0;
+  real64 PStart = -1.0;
+  real64 PEnd = -1.0;
+  real64 dP = -1.0;
```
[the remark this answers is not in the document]
francoishamon: Here I left the code as it was, because when I do the modifications you suggest I get `error: 'PStart' may be used u\
ninitialized in this function [-Werror=maybe-uninitialized]`

src/coreComponents/constitutive/unitTests/testMultiFluid.cpp
```diff
@@ -31,84 +31,95 @@ using namespace geosx;
...
-                               " 30\t            30000000\t1.07\t    0.000594\n"
-                               " 40\t            40000000\t1.08\t    0.00051\n"
-                               "                50000000\t1.07\t    0.000549  -- another one\n"
-                               "                90000000\t1.06\t    0.00074\n"
-                               " 50\t            50000000.7\t1.09\t    0.000449\n"
-                               "                90000000.7\t1.08\t    0.000605";
-
-static const char * pvtw_str = "#\tPref[bar]\tBw[m3/sm3]\tCp[1/bar]\t    Visc[cP]\n"
-                               "\t30600000.1\t1.03\t\t0.00000000041\t0.0003";
+static const char * PVTGString = "#\tPg(Pa)\t\tRv(sm3/sm3)\tBg(m3/sm3)\tVisc(Pa.s)\n"
```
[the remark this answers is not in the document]
francoishamon: Done

src/coreComponents/constitutive/fluid/PVTFunctions/SpanWagnerCO2DensityFunction.cpp
```diff
@@ -120,217 +134,138 @@ real64 f( const real64 & T, const real64 & P, const real64 & rho )
...
+  expectedCO2ComponentNames[1] = "co2";
+  bool found = PVTFunctionHelpers::findName( componentNames, expectedCO2ComponentNames, m_CO2Index );
+  GEOSX_ERROR_IF( !found, "Component CO2 is not found!" );
 
   makeTable( inputPara );
-
 }
 
-void SpanWagnerCO2DensityFunction::makeTable( string_array const & inputPara )
+void SpanWagnerCO2Density::makeTable( array1d< string > const & inputPara )
```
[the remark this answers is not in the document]
francoishamon: Yes good idea, now it is factorized to avoid the code duplication of the original code. The new function is in `PVTFunctionHelpers`.

src/coreComponents/constitutive/fluid/PVTFunctions/CO2SolubilityFunction.cpp
```diff
@@ -274,121 +199,128 @@ void CO2SolubilityFunction::makeTable( string_array const & inputPara )
...
-  }
-  else
-  {
-    Y = compFraction[m_CO2Index] / (1.0 - compFraction[m_CO2Index]);
-  }
+  V_r = 0.75*Rgas*(T_K_f+T)/(P*P_Pa_f)*(1/V_c);
 
-  if( Y < X )
+  // iterate until the solution of the CO2 equation of state is found
+  for(;; )
```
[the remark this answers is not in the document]
francoishamon: Yes I was reluctant to changing that because I was afraid of breaking things (the equation of state solved in this loop seems quite nasty). I may try to remove it at the very end

francoishamon:
In this PR, one last thing that I have to resolve is the type selection for the `PVTFunction`s. In the current implementation, these types are hard-coded in the class `MultiPhaseMultiComponentFluid`. We could template this class on the types of the `PVTFunction`s, and select these types when the `catalogName` function is called. 

To do that, an example that I was planning to follow is the selection of the `BASE` template in the `SinglePhaseFVM` class:
https://github.com/GEOSX/GEOSX/blob/80ab48fb27b53db20977903ea4743d31dcd0d501/src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseFVM.hpp#L132-L138
This seems to be doing what I need. To extend this to the selection of 5 templates, I am a little bit lost :) Is there a construct that can somehow extend this example to my case?

Besides that, I think the PR is in good shape--provided that we are ok with the `arrayView` of size 0.

src/coreComponents/constitutive/fluid/PVTFunctions/FenghourCO2ViscosityFunction.cpp
```diff
@@ -16,155 +16,113 @@
...
-    for( localIndex j = 0; j < temperature.size(); ++j )
+    for( localIndex j = 0; j < tableCoords.nTemperatures(); ++j )
     {
-      fenghourCO2Viscosity( temperature[j], density[i][j], viscosity[i][j] );
+      fenghourCO2ViscosityFunction( tableCoords.getTemperature( j ),
+                                    densities[j*tableCoords.nPressures()+i],
+                                    viscosities[j*tableCoords.nPressures()+i] );
     }
   }
 }
```
TotoGaz, line 113: The `tableCoords.nPressures()` and `tableCoords.nTemperatures()` should be done outside of the loops.

src/coreComponents/constitutive/fluid/PVTFunctions/PVTFunctionHelpers.hpp
```diff
@@ -0,0 +1,162 @@
...
+  void appendTemperature( const real64 & temp ) { coords[1].emplace_back( temp ); }
+
+  real64 const & getPressure( localIndex i ) const { return coords[0][i]; }
+  real64 const & getTemperature( localIndex i ) const { return coords[1][i]; }
+
+  array1d< array1d< real64 > > const & getCoords() const { return coords; }
+
+private:
+  array1d< array1d< real64 > > coords;
+};
```
TotoGaz, line 50: So do you think it is worth the shot or not?
I do not code the algorithms so I cannot be sure it's useful. You tell me.

Another solution could be to inherit
```c++
class PTTableCoordinates: public array1d< array1d< real64 > >
```
so you could use `PTTableCoordinates` for what it is (an array of array) too.
That would give more possibilities so this may (not?) be what you want.
You shouldn't need the `getCoords` anymore.
What would you prefer?
Also, we should do it `geosx` style, which I don't know.

Last, you should possibly define some `0` and `1` private constants.

src/coreComponents/constitutive/fluid/PVTFunctions/PVTFunctionHelpers.hpp
```diff
@@ -0,0 +1,162 @@
...
+    using std::begin;
+    using std::end;
+    auto it = std::find_first_of( begin( inputNames ), end( inputNames ), begin( expectedNames ), end( expectedNames ) );
+    localIndex const id = std::distance( begin( inputNames ), it );
+    return id;
+  }
+
+  static real64
+  initializePropertyTableWithSalinity( string_array const & inputParameters,
+                                       PTTableCoordinates & tableCoords )
```
TotoGaz, line 75: You could actually just call the function without salinity and just parse the eighth param in this function.

src/coreComponents/constitutive/fluid/PVTFunctions/SpanWagnerCO2DensityFunction.cpp
```diff
@@ -120,217 +133,97 @@ real64 f( const real64 & T, const real64 & P, const real64 & rho )
...
+                                                array1d< real64 > const & densities )
 {
 
-  constexpr real64 T_K_f = 273.15;
-
-  real64 PPa, TK;
+  constexpr real64 TK_f = 273.15;
 
-  for( localIndex i = 0; i < pressure.size(); ++i )
+  for( localIndex i = 0; i < tableCoords.nPressures(); ++i )
```
TotoGaz, line 179: Same here, please take the `nTemperatures` and `nPressures` out of the loops.

src/coreComponents/constitutive/fluid/PVTFunctions/CO2SolubilityFunction.cpp
```diff
@@ -16,379 +16,258 @@
...
-      break;
-    }
-
-  }
-
-  GEOSX_ERROR_IF( notFound, "Component CO2 is not found!" );
+  constexpr real64 eps = 1e-9;
+  constexpr real64 dx = 1e-10;
+  int count = 0;
+  real64 Vr_int = 0.05;
```
TotoGaz, line 226: Is there any way to let the user define those parameters (you can keep these defaults if you want obviously).

klevzoff:
@francoishamon Have we reached a decision on whether we'll try to make this a proper *multiphase* model?

* If yes, there's a lot more work involved in extending this from 2 phases to an arbitrary number. In particular, we would need heterogeneous containers to store PVT function wrappers of different static types for an arbitrary number of phases (think `std::tuple` but device-friendly). Also some template metaprogramming gymnastics to handle multiple variadic packs and such. It's not an easy extensions and I would definitely not recommend attempting it all in one PR.

* If not, I suggest renaming the class to `TwoPhaseFluid` or something like this, and rename the member variables to remove references to "co2" and "brine", just name them something generic (maybe phase1 and phase 2). Then you can template it on the 5 function types.

klevzoff:
> To extend this to the selection of 5 templates, I am a little bit lost :) Is there a construct that can somehow extend this example to my case?

Assuming we stick to the two-phase model (for now at least), I wouldn't use the SFINAE-based pattern for this, it's too verbose. Instead we can delegate to a helper template that will be specialized for the supported combinations of functions.

So in `MultiPhaseMultiComponentFluid.hpp` you just leave the declaration:
```c++
template< typename P1DENS, typename P1VISC, typename P2DENS, typename P2VISC, typename FLASH >
class MultiPhaseMultiComponentFluid
{
  // ...
  static string catalogName(); // leave implementation to cpp file
  // ...
};
```

Then in `MultiPhaseMultiComponentFluid.cpp` we do explicit template instantiations, catalog registration and the helper template that provides catalog names:
```c++
namespace
{
  template< typename P1DENS, typename P1VISC, typename P2DENS, typename P2VISC, typename FLASH >
  TwoPhaseCatalogNames;

  template<>
  TwoPhaseCatalogNames< PVTProps::BrineCO2Density, 
                        PVTProps::BrineViscosity, 
                        PVTProps::SpanWagnerCO2Density,
                        PVTProps::FenghourCO2Viscosity,
                        PVTProps::CO2Solubility >
  {
    static string name() { return "CO2BrineFluid"; }
  }
} // end namespace

// provide a definition for catalogName()
template< typename P1DENS, typename P1VISC, typename P2DENS, typename P2VISC, typename FLASH >
string MultiPhaseMultiComponentFluid< P1DENS, P1VISC, P2DENS, P2VISC, FLASH >::catalogName()
{
  return TwoPhaseCatalogNames< P1DENS, P1VISC, P2DENS, P2VISC, FLASH >::name();
}

// explicit instantiation of the mode template
template class MultiPhaseMultiComponentFluid< PVTProps::BrineCO2Density, 
                                              PVTProps::BrineViscosity, 
                                              PVTProps::SpanWagnerCO2Density,
                                              PVTProps::FenghourCO2Viscosity,
                                              PVTProps::CO2Solubility >;

// finally, register in the factory
using CO2BrineFluid = MultiPhaseMultiComponentFluid< PVTProps::BrineCO2Density, 
                                                     PVTProps::BrineViscosity, 
                                                     PVTProps::SpanWagnerCO2Density,
                                                     PVTProps::FenghourCO2Viscosity,
                                                     PVTProps::CO2Solubility >;

REGISTER_CATALOG_ENTRY( ConstitutiveBase, CO2BrineFluid, string const &, Group * const )
```
(as always, my code is "commentware", meaning it's not guaranteed to compile or link; also, I changed the order of template arguments)

src/coreComponents/constitutive/fluid/PVTFunctions/PVTFunctionHelpers.hpp
```diff
@@ -0,0 +1,162 @@
...
+
+    for( real64 P = PStart; P <= PEnd; P += dP )
+    {
+      tableCoords.appendPressure( P );
+    }
+    for( real64 T = TStart; T <= TEnd; T += dT )
+    {
+      tableCoords.appendTemperature( T );
+    }
+    return salinity;
```
TotoGaz, line 112: Just put this code inside the `try`/`catch` so you can remove the initial declarations and declare your variables `const`.

PS: Obviously you do not require that you have the same number of `P` and `T`.
Why do you store this in a 2d array instead in 2 separate 1d arrays?

src/coreComponents/managers/Functions/TableFunction.hpp
```diff
@@ -177,16 +297,170 @@ class TableFunction : public FunctionBase
...
+TableFunction::KernelWrapper::compute( IN_ARRAY const & input, real64 & value, OUT_ARRAY && derivatives ) const
+{
+  value = 0.0;
+  for( localIndex i = 0; i < m_dimensions; ++i )
+  {
+    derivatives[i] = 0.0;
+  }
 
+  // Linear interpolation
+  if( m_interpolationMethod == TableFunction::InterpolationType::Linear )
```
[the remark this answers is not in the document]
francoishamon: @TotoGaz in the current version of the PR, I don't touch this file anymore, so I will leave this comment for a subsequent PR. There is a lot of room for improvement in the `compute` function of `TableFunction` for use in the CO2-brine model. In particular, since we construct internally the table coordinates with uniform spacing, it does not make much sense to use binary search. Will be addressed in another PR.

src/coreComponents/constitutive/fluid/PVTFunctions/FenghourCO2ViscosityFunction.cpp
francoishamon: Yes good point, done

src/coreComponents/constitutive/fluid/PVTFunctions/PVTFunctionHelpers.hpp
francoishamon: I implemented what you suggested, with `initializePropertyTableWithSalinity` calling `initializePropertyTable` and parsing the eighth parameter

src/coreComponents/constitutive/fluid/PVTFunctions/SpanWagnerCO2DensityFunction.cpp
francoishamon: Done

src/coreComponents/constitutive/fluid/PVTFunctions/CO2SolubilityFunction.cpp
francoishamon: Here I would say that it is not a good idea to let the user choose this type of parameters. I would prefer to let the (expert) users take GEOSX and recompile with a better parameter if needed.

francoishamon:
> Assuming we stick to the two-phase model (for now at least), I wouldn't use the SFINAE-based pattern for this, it's too verbose. Instead we can delegate to a helper template that will be specialized for the supported combinations of functions.

@klevzoff thanks very much for the suggestion, it is very useful.

For the first question about two-phase vs true multiphase, I think it is more reasonable to stick two-phase for now since this is just a refactoring PR, and this is what the original class was doing. I think the plan is to have a true compositional CO2 model this year, but I don't know enough details about this future work to know whether it will fit in this class or not.

Regarding the code that you sent, that is very impressive that you are able to write such complex code just on top of your head :) And it works (at least, on Lassen it works, we will see on Travis)! I copy pasted what you sent and I removed the brine and CO2 words from the variable names in my original code. Now the last point that makes this class not fully generic is the presence of the keywords `SpanWagnerDensityFunction` etc when we read the user-defined tables, but I think I may leave that for another PR (I will think about it more). I may have a use case with a different flash function in the next weeks/months, so we will see then if the design works well. This PR contains all the things that I wanted to do now. Thanks again!

src/coreComponents/constitutive/fluid/MultiPhaseMultiComponentFluid.hpp
```diff
@@ -267,31 +298,618 @@ class MultiPhaseMultiComponentFluid : public MultiFluidBase
...
 
+// this alias will be useful in constitutive dispatch
+using CO2BrineFluid = MultiPhaseMultiComponentFluid< PVTProps::BrineCO2Density,
+                                                     PVTProps::BrineViscosity,
+                                                     PVTProps::SpanWagnerCO2Density,
+                                                     PVTProps::FenghourCO2Viscosity,
+                                                     PVTProps::CO2Solubility >;
+
+// The constructor is in the hpp for one reason: the testMultiFluid won't link if the constructor is in the cpp
+// I could not find a workaround
```
klevzoff, line 363: I wonder if this could be because the explicit instantiation is at the top of the file, before the compiler can see the function definitions? Try moving the instantiation to the bottom of the file (just before the catalog registration macro). If it doesn't help, let's just leave the constructor here and I'll try to figure it out later.

src/coreComponents/constitutive/fluid/PVTFunctions/PVTFunctionHelpers.hpp
francoishamon: Yes I agree, I put the code mentioned above inside the `try/catch`, and I removed the initial declarations as you suggested. 

Regarding the PS, you are right that we don't require the same number of `P` and `T`. But this is exactly why this is stored as an `array1d< array1d< real64 > >`, which supports the case of different numbers of pressures and temperatures. Then `TableFunction` converts this into an `ArrayOfArrays` and everything is good.

src/coreComponents/constitutive/fluid/PVTFunctions/PVTFunctionHelpers.hpp
francoishamon: Here I think the two ways that you proposed (either inheriting from `array1d< array1d< real64 > >`  or have `array1d < array1d< real64 > >` as datamember) are quite similar for what we want to do: avoiding hard-coded 0 and 1 everywhere in the `PVTFunction`s. I just implemented the simplest option for me, which is to have the `array1d< array1< real64 > >` as datamember. To be honest, I think it is fine like that for now.

Also, I added the private constants that you suggested in the more recent commit.

src/coreComponents/constitutive/fluid/MultiPhaseMultiComponentFluid.hpp
francoishamon: Yes great! In the code you looked at, I had applied my usual recipe when things are working on Quartz but not linking on Lassen: moving the implementation to the hpp file... But now with what you suggested, I compiled fine on Lassen with the constructor implementation in the cpp. Thanks

src/coreComponents/constitutive/unitTests/testMultiFluid.cpp
```diff
@@ -698,6 +801,240 @@ TEST_F( DeadOilFluidFromTableTest, numericalDerivativesMolar )
   }
 }
 
+template class MultiPhaseMultiComponentFluid< PVTProps::BrineCO2Density,
+                                              PVTProps::BrineViscosity,
+                                              PVTProps::SpanWagnerCO2Density,
+                                              PVTProps::FenghourCO2Viscosity,
+                                              PVTProps::CO2Solubility >;
```
klevzoff, line 808: I hoped this would not be needed in the test files... can you confirm?

src/coreComponents/constitutive/fluid/PVTFunctions/SpanWagnerCO2DensityFunction.hpp
```diff
@@ -13,61 +13,159 @@
...
+
+  /// Table with brine density tabulated as a function (P,T,sal)
+  TableFunction::KernelWrapper const m_CO2DensityTable;
+
+  /// Index of the CO2 component
+  localIndex const m_CO2Index;
+
+};
+
+class SpanWagnerCO2Density : public PVTFunctionBase
```
klevzoff, line 83: The file name still has "Function" in it and doesn't match the class name, could you rename the file?
(I agree with removing "Function" from class names, just to be clear)

src/coreComponents/constitutive/fluid/PVTFunctions/FenghourCO2ViscosityFunction.hpp
```diff
@@ -16,56 +16,139 @@
...
+                        bool useMass = 0 ) const override;
+
+protected:
+
+  /// Table with viscosity tabulated as a function (P,T)
+  TableFunction::KernelWrapper const m_CO2ViscosityTable;
+
+};
+
+class FenghourCO2Viscosity : public PVTFunctionBase
```
klevzoff, line 78: Same comment re: file name

src/coreComponents/constitutive/fluid/PVTFunctions/CO2SolubilityFunction.hpp
```diff
@@ -13,55 +13,292 @@
...
+
+  /// Index of the gas phase
+  localIndex const m_phaseGasIndex;
+
+  /// Index of the liquid phase
+  localIndex const m_phaseLiquidIndex;
+
+};
+
+class CO2Solubility : public FlashModelBase
```
klevzoff, line 103: Same comment re: file name

src/coreComponents/constitutive/fluid/PVTFunctions/BrineViscosityFunction.hpp
```diff
@@ -16,53 +16,132 @@
...
+protected:
+
+  real64 const m_coef0;
+
+  real64 const m_coef1;
+
+};
+
+
+class BrineViscosity : public PVTFunctionBase
```
klevzoff, line 79: Same comment re: file name

src/coreComponents/constitutive/fluid/PVTFunctions/BrineCO2DensityFunction.hpp
```diff
@@ -16,59 +16,220 @@
...
+
+  /// Index of the CO2 component
+  localIndex const m_CO2Index;
+
+  /// Index of the water component
+  localIndex const m_waterIndex;
+
+};
+
+class BrineCO2Density : public PVTFunctionBase
```
klevzoff, line 88: Same comment re: file name

src/coreComponents/constitutive/fluid/MultiPhaseMultiComponentFluid.cpp
```diff
@@ -19,418 +19,214 @@
...
+        if( strs[1] == "BrineCO2Density" )
+        {
+          m_p1Density = std::make_unique< P1DENS >( strs, m_componentNames, m_componentMolarWeight );
+          m_p1DensityWrapper.emplace_back( m_p1Density->createKernelWrapper() );
+        }
+        else if( strs[1] == "SpanWagnerCO2Density" )
+        {
+          m_p2Density = std::make_unique< P2DENS >( strs, m_componentNames, m_componentMolarWeight );
+          m_p2DensityWrapper.emplace_back( m_p2Density->createKernelWrapper() );
+        }
```
klevzoff, line 158: Can we use something like `P1DENS::catalogName()` and `P2DENS::catalogName()` instead of hard-coded strings here? Same for viscosity below. I know you've addressed this in a comment, but I feel like this would bring some kind of closure to this huge refactoring effort. And it should be simple enough to add.

klevzoff (approved):
Great work!
Just some final comments after I looked at all of it (sorry!)

src/coreComponents/constitutive/fluid/PVTFunctions/PVTFunctionHelpers.hpp
```diff
@@ -0,0 +1,157 @@
...
+
+    try
+    {
+      PStart = stod( inputParameters[2] );
+      PEnd = stod( inputParameters[3] );
+      dP = stod( inputParameters[4] );
+
+      TStart = stod( inputParameters[5] );
+      TEnd = stod( inputParameters[6] );
+      dT = stod( inputParameters[7] );
```
TotoGaz, line 130:
```suggestion
      real64 const PStart = stod( inputParameters[2] );
      real64 const PEnd = stod( inputParameters[3] );
      real64 const dP = stod( inputParameters[4] );

      real64 const TStart = stod( inputParameters[5] );
      real64 const TEnd = stod( inputParameters[6] );
      real64 const dT = stod( inputParameters[7] );
```

src/coreComponents/constitutive/fluid/PVTFunctions/PVTFunctionHelpers.hpp
```diff
@@ -0,0 +1,157 @@
...
+  array1d< array1d< real64 > > coords;
+};
+
+struct PVTFunctionHelpers
+{
+
+  /**
+   * @brief Look for the expectedNames in the names provided by the user
+   * @param[in] inputNames phase or component names provided by the user
+   * @param[in] expectedNames expected names that can be accepted by GEOSX
```
TotoGaz, line 65: Please `@tparam` too.

src/coreComponents/constitutive/fluid/PVTFunctions/CO2SolubilityFunction.cpp
```diff
@@ -16,379 +16,273 @@
...
-      solubiltiy[i][j] = y_CO2 * P / exp( logK );
+      // compute equation (6) of Duan and Sun (2003)
+      real64 const logK = detail::Par( T+T_K_f, P, mu )
+                          - detail::logF( T, P, V_r )
+                          + 2*detail::Par( T+T_K_f, P, lambda ) * salinity
+                          + detail::Par( T+T_K_f, P, zeta ) * salinity * salinity;
 
+      // mole fraction of CO2 in vapor phase, equation (4) of Duan and Sun (2003)
+      real64 const y_CO2 = (P - detail::PWater( T ))/P;
+      values[j*tableCoords.nPressures()+i] = y_CO2 * P / exp( logK );
```
TotoGaz, line 225:
```suggestion
      values[j*nPressures+i] = y_CO2 * P / exp( logK );
```

src/coreComponents/constitutive/fluid/MultiPhaseMultiComponentFluid.hpp
```diff
@@ -267,31 +298,600 @@ class MultiPhaseMultiComponentFluid : public MultiFluidBase
...
+
+  /// Pointer to the CO2 density model
+  array1d< typename P2DENS::KernelWrapper > m_p2DensityWrapper;
 
-  std::shared_ptr< PVTProps::FlashModel const > m_flashModel;
+  /// Pointer to the CO2 viscosity model
+  array1d< typename P2VISC::KernelWrapper > m_p2ViscosityWrapper;
+
+  /// Pointer to the flash model
+  array1d< typename FLASH::KernelWrapper > m_flashWrapper;
```
TotoGaz, line 351: You should really document that these `array1d` are tricks (and explain the issue once somewhere.)

src/coreComponents/constitutive/fluid/MultiPhaseMultiComponentFluid.hpp
```diff
@@ -267,31 +298,600 @@ class MultiPhaseMultiComponentFluid : public MultiFluidBase
...
+         arraySlice1d< real64 > const & dPhaseViscosity_dTemperature,
+         arraySlice2d< real64 > const & dPhaseViscosity_dGlobalCompFraction,
+         arraySlice2d< real64 > const & phaseCompFraction,
+         arraySlice2d< real64 > const & dPhaseCompFraction_dPressure,
+         arraySlice2d< real64 > const & dPhaseCompFraction_dTemperature,
+         arraySlice3d< real64 > const & dPhaseCompFraction_dGlobalCompFraction,
+         real64 & totalDensity,
+         real64 & dTotalDensity_dPressure,
+         real64 & dTotalDensity_dTemperature,
+         arraySlice1d< real64 > const & dTotalDensity_dGlobalCompFraction ) const
```
TotoGaz, line 582: 5 template parameters and 27 arguments, that's something 😉

src/coreComponents/constitutive/fluid/PVTFunctions/CO2SolubilityFunction.cpp
TotoGaz: I think you're going wrong way here.

If I need to test another parameter set I need to spend hours connecting to the remote HPC, set the build env, build and deploy, just to test?
And then I have to re-deploy with all the Quality Assurance process, the tickets with the outsourced team...
And after that, what is the reference version ? Not the same code base, not the same hash...
Imagine that two users have incompatible parameters, do you have to deal with two different versions?

francoishamon: Done, now the tolerance can be chosen from the input file.

src/coreComponents/constitutive/unitTests/testMultiFluid.cpp
francoishamon: Thanks for the catch, I forgot to remove this, and it broke the compilation for the non-GPU builds on Travis... I removed it in the latest commit.

src/coreComponents/constitutive/fluid/PVTFunctions/SpanWagnerCO2DensityFunction.hpp
francoishamon: Done, I renamed this file, and all the others that you mentioned in your review.

src/coreComponents/constitutive/fluid/MultiPhaseMultiComponentFluid.cpp
francoishamon: Yes, you are right it very easy in fact. Now it looks like: ` if( strs[1] == P1DENS::catalogName() )`

src/coreComponents/constitutive/fluid/PVTFunctions/PVTFunctionHelpers.hpp
```diff
@@ -0,0 +1,157 @@
...
+  static real64
+  initializePropertyTableWithSalinity( string_array const & inputParameters,
+                                       PTTableCoordinates & tableCoords )
+  {
+    localIndex const expectedNumParameters = 9;
+    real64 salinity = 0.0;
+    initializePropertyTable( inputParameters, expectedNumParameters, tableCoords );
+    try
+    {
+      salinity = stod( inputParameters[8] );
```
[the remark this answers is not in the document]
francoishamon: Done

src/coreComponents/constitutive/fluid/PVTFunctions/PVTFunctionHelpers.hpp
francoishamon: Done

src/coreComponents/constitutive/fluid/PVTFunctions/CO2SolubilityFunction.cpp
francoishamon: Done

francoishamon:
Hello @CusiniM, 
This is the PR that will break the thermal PR #1225. If I merge this PR (which is ready) before yours, it will make your life difficult because you will have to modify quite significantly the classes computing internal energy and enthalpy that @yue-2018 added to your PR. As we discussed on Slack, I see three possibilities to solve this issue:

1. I merge this PR first, and then, in a separate (small) PR, I modify and add the internal energy and enthalpy classes to GEOSX. They will just be unused until you merge your PR, but I think that's fine.
2. I merge this PR first, and then, in your PR, I modify the internal energy and enthalpy classes to make them fit into the new style, and then everything (energy equation + internal energy and enthalpy in the new style) will be merged at the same time in GEOSX.
3. You merge first the thermal PR, and then I merge this CO2-brine PR. I don't mind waiting.

My preference goes to either 1 or 3. Let me know what you think.

CusiniM:
> Hello @CusiniM,
> This is the PR that will break the thermal PR #1225. If I merge this PR (which is ready) before yours, it will make your life difficult because you will have to modify quite significantly the classes computing internal energy and enthalpy that @yue-2018 added to your PR. As we discussed on Slack, I see three possibilities to solve this issue:
> 
>     1. I merge this PR first, and then, in a separate (small) PR, I modify and add the internal energy and enthalpy classes to GEOSX. They will just be unused until you merge your PR, but I think that's fine.
> 
>     2. I merge this PR first, and then, in your PR, I modify the internal energy and enthalpy classes to make them fit into the new style, and then everything (energy equation + internal energy and enthalpy in the new style) will be merged at the same time in GEOSX.
> 
>     3. You merge first the thermal PR, and then I merge this CO2-brine PR. I don't mind waiting.
> 
> 
> My preference goes to either 1 or 3. Let me know what you think.

I think you should merge this PR since it's ready. We can then go for either option 1 or 2. I do not have any real preference.

src/coreComponents/constitutive/unitTests/testMultiFluid.cpp
```diff
@@ -251,12 +261,24 @@ void testNumericalDerivatives( MultiFluidBase & fluid,
...
+      //       cancellations that happen after when we differentiate wrt component densities. Hence the flag below.
+      if( usePVTPackage )
+      {
+        // renormalize
+        real64 sum = 0.0;
+        for( localIndex ic = 0; ic < NC; ++ic )
+          sum += compNew[ic];
+        for( localIndex ic = 0; ic < NC; ++ic )
+          compNew[ic] /= sum;
+      }
```
klevzoff, line 281: Interesting issue. Though I don't fully understand the comment - you say the new (analytical) models have problems with this test, but you apply the re-normalization fix to the old (FD-based) models that used to work? Does this pass the test for both?

In any case, the problem seems to be applying a perturbation to a set of variables that aren't independent (constrained to unit sum). Perhaps the correct way to do this test (for both kinds of models) is to 1) apply the perturbation to one variable, 2) re-normalize, 3) compute "effective" deltas to each of the variables resulting from re-normalization, and 4) checking the numerical derivative w.r.t. all variables while taking into account the "effective" deltas (instead of `dC`, the original prescribed perturbation)?

francoishamon: Yes sorry for the very unclear comment, and part of it is outdated. The issue comes from the normalization that is used in this test and in `PVTPackage` in the FD derivatives calculations (the GitHub diff is misleading here, I only added the flag `if( usePVTPackage )` to skip the normalization for the native models).

In `PVTPackage` the derivatives are obtained by 1) perturbing a given component fraction, and 2) re-normalizing so that the sum of the component fractions is equal to zero. With the normalization, the FD difference derivative **wrt component fraction** that is obtained is quite different from the analytical one. I tried to explain this for the phase fraction (equal to the component fraction in DO) in a two-phase model in the image below (not sure what GitHub will do with it):

deriv_1
So as long as the normalization is applied in `PVTPackage`, it makes sense to have it for the models implemented there. My impression (not checked yet, will let you know) is that if we use an effective "delta", we are not going to match the `PVTPackage` derivative (Option A), but instead, we are going to get the analytical one (Option B).

In `develop`, the reason why the native DO model passes the test without the flag `if( usePVTPackage)` is that I enforced the "FD difference derivative with normalization" in my implementation here: https://github.com/GEOSX/GEOSX/blob/d4ade8e7be45494ea619e18980f39d1b75309b7d/src/coreComponents/constitutive/fluid/DeadOilFluid.hpp#L705
But I find this way of computing the derivatives not very intuitive, so now I would like to change it in this PR here: https://github.com/GEOSX/GEOSX/blob/2a038ad3548fc77b5aeec138e27738bfe4cc09a6/src/coreComponents/constitutive/fluid/DeadOilFluid.hpp#L707
and if the normalization is used to check derivatives, `testMultiFluid` says that my derivative is wrong (hence the flag)

Eventually, what we care about are the derivatives **wrt component density** that we put in the Jacobian, and both (normalization or no normalization) approaches are equivalent, even if the intermediate derivatives wrt component fractions are quite different. There is indeed a cancellation of terms taking place in the chain rule:
deriv_2

deriv_3

klevzoff: Ok, I understand now, thanks for the detailed explanation! I agree that the new way of assigning `dPhaseFraction_dGlobalCompFraction` in DO is more intuitive (and mathematically correct, given that DO model itself does not apply input composition re-normalization, and leaves it to the user to adjust derivatives if they do so).

francoishamon: For this normalization issue, in the latest commit I tried to improve my explanation in the inline comment just above the `if( usePVTPackage )` flag, and in this comment I reference this PR so people can view this discussion. For now, I left the flag that skips the normalization, but you are definitely right, this will have to be homogenized.

PR #1333, opened 2021-02-23
WIP: Experimental/corbett/jit

corbett5:
This is a proof of concept for a dynamic library based JIT. It takes the constitutive and element dispatch out of `regionBasedKernelApplication` and instead compiles and links in the kernel at runtime. I confirmed that it is passing the `SSLE_sedov_01` problem on Quartz with Clang debug and GCC release and on Lassen with Clang debug.

It should pass all the one rank problems if I were to `#include` the appropriate headers for all the various kernels at the bottom of `KernelBase.hpp` like I did for `SolidMechanicsSmallStrainExplicitNewmarkKernel.hpp`. As is there are some compilation errors (with the main code not the JIT) if I do this, but structuring things like this is a terrible idea so I just proceeded with my hack.

Each run re-compiles the kernels it needs, this is wasteful but would be easily fixed.

Multiple rank runs fail because each rank tries to compile the kernel, easily fixed with some synchronization.

Related to
https://github.com/GEOSX/LvArray/pull/225


corbett5:
@rrsettgast @klevzoff @francoishamon take a look.

TotoGaz:
Hi @corbett5 ! Would this `jit` approach means that instead of deploying `GEOSX` we need to provide the end-user with a complete `GEOSX` build environment? Will I still be able to work with models that I do not want to share (i.e. without accessing the sources)?

corbett5:
@TotoGaz yes you will be able to pre-compile things.

src/coreComponents/finiteElement/kernelInterface/KernelBase.hpp
```diff
@@ -317,16 +292,158 @@ class KernelFactory
...
+//  decide on whether to JIT or not.. but current language restrictions on templating and our in-code restrictions on
+//   the kernels due to the 'using' statements throughout the code to define specific kernel dispatchers make that
+//   devwork intractable.
+// We can't use 'using' here due to differences in the above templates as well... nor a nice preprocessor macro... 
+//   so we're basically left with only one option
+#if JITTI == 1
+  #define KernelDispatch KernelDispatchJIT
+#else
+  #define KernelDispatch KernelDispatchTemplate
+#endif
```
wrtobin, line 446: Per the comments, this is the part of the current implementation I want to change the most. It should be possible, but wound up eating hours while I was trying to get it implemented the way I would prefer.

wrtobin:
I'll need to merge from dev and revert the submodule change prior to merge, but this is functional and in a good place to review. I still need to check some things on lassen, but was able to get 100% of the regression tests passing on quartz. 

Though it did require multiple passes as it is possible for multiple tests to try to compile the same kernel simultaneously resulting in file system errors that kill the runs. This had to be avoiding internally via some mpi checks, but when multiple executables are running simultaneously it might require adding sentinel files or something similar, I'm open to ideas to mitigate that issue.

src/coreComponents/finiteElement/kernelInterface/KernelBase.hpp
klevzoff: It should be possible to make the two dispatcher classes similar by replacing `NAME` and `HEADER` parameters in `KernelDispatchJIT` with with just `KERNEL_TYPE` parameter (i.e. template template parameter, like in the non-JIT dispatcher).

`NAME` could then be extracted as (I'm trying to decide if this is 100% robust):
```c++
string const fullName = LvArray::system::demangleType< KERNEL_TYPE< SUBREGION_TYPE, FE_TYPE, CONSTITUTIVE_TYPE > >();
string const name = fullName.substr( 0, fullName.find( '<' ) );
```

`HEADER` could be replaced by adding inside each "leaf" kernel class something like
```c++
template< typename SUBREGION_TYPE, typename FE_TYPE, typename CONSTITUTIVE_TYPE >
class MyKernel
{
  static constexpr char const source_location[] = __FILE__;
  ...
};

// Sad part: this is required in C++14 to avoid linker errors
template< typename SUBREGION_TYPE, typename FE_TYPE, typename CONSTITUTIVE_TYPE >
constexpr char const MyKernel< SUBREGION_TYPE, FE_TYPE, CONSTITUTIVE_TYPE >source_location[];
```
and accessing as
```c++
string const header = KERNEL_TYPE< SUBREGION_TYPE, FE_TYPE, CONSTITUTIVE_TYPE >::source_location;
```
The downside is having this extra stuff to remember to put in kernels. Maybe there's a better way to associate kernel class with source file name?

These ideas aren't too pretty, just thought I'd mention them. They allow us to gets rid of `JITTI_DECL`/`JITTI_TPARAM` macros entirely and unify the two dispatchers, possibly even making them specializations as the comment above suggests, and allowing for a per-kernel JITting decision.

klevzoff (approved):
Looks great! Just have one suggestion (not necessarily a good one) re: unifying the two dispatchers.

src/coreComponents/finiteElement/CMakeLists.txt
```diff
@@ -37,12 +39,22 @@ endif()
...
 target_include_directories( finiteElement PUBLIC ${CMAKE_SOURCE_DIR}/coreComponents)
 
+add_custom_command( OUTPUT ${CMAKE_BINARY_DIR}/include/kernelJITCompileCommands.hpp
+                    WORKING_DIRECTORY ${CMAKE_BINARY_DIR} 
+                    COMMAND python ${CMAKE_CURRENT_LIST_DIR}/../LvArray/src/jitti/generateCompileCommandsHeader.py 
+                                   ${CMAKE_BINARY_DIR}/compile_commands.json
+                                   --cpp ${CMAKE_CURRENT_LIST_DIR}/kernelInterface/kernelJIT.cpp 
+                                   --hpp ${CMAKE_BINARY_DIR}/include/kernelJITCompileCommands.hpp
+                                   --include ${CMAKE_BINARY_DIR}/include 
+                                   --linker ${CMAKE_CXX_COMPILER} )
```
TotoGaz, line 56: Instead of relying on headers that you generate on the file system, could it have been more robust to store the C++ pre-processing output as a string alongside the compile/link commands directly into the lib?
And that would be more "modern" JIT style imho.

corbett5: Yeah I think you could do that, although it would require the user to list out the functions that they want to JIT at configuration time in CMake. Then for each of those functions we could pre-process `jitti/templateSource.cpp`. But only up to a point, since we don't know what template params the user will want. Then at run time when we know the params we could compile the pre-processed source with the additional command line definitions of `JITTI_TEMPLATE_PARAMS` and `JITTI_TEMPLATE_PARAMS_STRING`.

However, how we pass a string embedded in our library to the compiler without using the file system is beyond me. Not to mention that at the end of the day the compiler is going to spit out a library on the file system that we then have to open. So this frees us from having to have access to the same headers used to build, but unless we do something really fancy I think file system access is a requirement.

Also at least with CUDA the cost of the compilation time itself will greatly outweigh the cost of opening and pre-processing the source.

src/cmake/GeosxOptions.cmake
```diff
@@ -93,6 +93,16 @@ endif(NOT BLT_CXX_STD STREQUAL c++14)
 
 message("CMAKE_CXX_COMPILER_ID = ${CMAKE_CXX_COMPILER_ID}")
 
+option( ENABLE_JITTI "Build all compute kernels just-in-time at runtime." OFF )
+
+if ( ENABLE_JITTI )
+  message( "JITTI is ENABLED")
```
corbett5, line 99: I think the way we've been doing this is `GEOSX_ENABLE_JITTI`, or maybe it's `GEOSX_USE_JITTI`, I forget. But either way I like defined/not defined instead of 1/0.

src/coreComponents/finiteElement/kernelInterface/KernelBase.hpp
corbett5: I think the end goal is to always use `JITTI`, although that doesn't mean that we need to always jit things at run-time. We need to have the capability to pre-jit everything at build-time, and I think that would suffice.

src/coreComponents/finiteElement/CMakeLists.txt
TotoGaz: > Also at least with CUDA the cost of the compilation time itself will greatly outweigh the cost of opening and pre-processing the source.

It's more a problem of having a consistent self contained GEOSX installation than a performance problem imho.
Embedding the pre-processed source files directly into GEOSX would help at non relying on sources that may be modified behind the scenes.
I do not know all the JIT details so maybe it's not a problem anyway: I do not want to solve problems that do not exist.

PR #1337, opened 2021-02-25
Fix Debug-build compile fail on Lassen

wrtobin:
This only effects the Debug build on Lassen. The Release and RelWithDebInfo builds work fine, which is why the nightlies didn't fail. Only one travis might have caught this: Ubuntu18.04_clang8.0.0_cuda10.1.243_debug, but presumably compiler/other version differences prevented it?

There was a host-only call ( `serialReduce::get()` ) inside a host-device lambda.

### DIFF
--- a/src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseHybridFVMKernels.hpp
+++ b/src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseHybridFVMKernels.hpp
@@ -860,7 +860,7 @@ struct PrecomputeKernel
       }
     } );
 
-    forAll< serialPolicy >( faceManagerSize, [=] GEOSX_HOST_DEVICE ( localIndex const iface )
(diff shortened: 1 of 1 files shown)


corbett5 (approved):
Funny, I thought I compiled just a day or two ago. Nice catch!

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseHybridFVMKernels.hpp
```diff
@@ -860,7 +860,7 @@ struct PrecomputeKernel
       }
     } );
 
-    forAll< serialPolicy >( faceManagerSize, [=] GEOSX_HOST_DEVICE ( localIndex const iface )
+    forAll< serialPolicy >( faceManagerSize, [=] GEOSX_HOST ( localIndex const iface )
```
corbett5, line 863:
```suggestion
    forAll< serialPolicy >( faceManagerSize, [=] ( localIndex const iface )
```

corbett5 (approved):

PR #1338, opened 2021-02-26
Enable Internal Mesh and Boundary Conditions for Wellbore Problems

sytuannguyen:
- Enable Internal Mesh and Normal Traction Boundary Condition for Wellbore Problems
- Wellbore integrated tests: solid mechanic, poromechanic, etc.
- Benchmark of wellbore problems

image

**Solid mechanic benchmark**

Good agreement between numerical results and analytical results (e.g. Detournay and Cheng, 1988, loading mode 1)

image

Using a fine mesh

image

**Poroelastic benchmark**

Good agreement with analytical results of Detournay and Cheng (1988), loading mode 2 - pore pressure applied on the borehole wall

image

Displacement field

image


rrsettgast:
@sytuannguyen Is `InternalWellGenerator.cpp` a modified copy of `InternalMeshGenerator.cpp`?

TotoGaz:
> @sytuannguyen Is `InternalWellGenerator.cpp` a modified copy of `InternalMeshGenerator.cpp`?

Indisputably, but it's still a draft, additional work is still to be done. 😉

rrsettgast:
> > @sytuannguyen Is `InternalWellGenerator.cpp` a modified copy of `InternalMeshGenerator.cpp`?
> 
> Indisputably, but it's still a draft, additional work is still to be done. 😉

I don't understand the motivation to copy a bunch of code in this use case. Can someone explain how a radial mesh is anything more than a cartesian mesh with a coordinate transformation and a modification to tie the nodes together at the theta +/- boundary?

sytuannguyen:
> > > @sytuannguyen Is `InternalWellGenerator.cpp` a modified copy of `InternalMeshGenerator.cpp`?
> > 
> > 
> > Indisputably, but it's still a draft, additional work is still to be done. 😉
> 
> I don't understand the motivation to copy a bunch of code in this use case. Can someone explain how a radial mesh is anything more than a cartesian mesh with a coordinate transformation and a modification to tie the nodes together at the theta +/- boundary?

You are totally right!!! 
But don't look at the code duplicate yet because this is just a draft and I just trying to get the functionality works. Feel free to jump in if you have time for this.

rrsettgast:
> > > > @sytuannguyen Is `InternalWellGenerator.cpp` a modified copy of `InternalMeshGenerator.cpp`?
> > > 
> > > 
> > > Indisputably, but it's still a draft, additional work is still to be done. 😉
> > 
> > 
> > I don't understand the motivation to copy a bunch of code in this use case. Can someone explain how a radial mesh is anything more than a cartesian mesh with a coordinate transformation and a modification to tie the nodes together at the theta +/- boundary?
> 
> @rrsettgast: You are totally right!!!
> But don't look at the code duplication yet because this is just a draft and I just trying to get the functionality works. Feel free to jump in if you have time for this.

Yes...but the code was essentially working in the original file. I can't see a substantive change in the new file. Is there a motivation for making a new class?

TotoGaz:
> Yes...but the code was essentially working in the original file. I can't see a substantive change in the new file. Is there a motivation for making a new class?

The way we do it, maybe registering a dedicated product in the factory requires a dedicated constructor?
Maybe could we just provide a `builder function` but I do not know if our `REGISTER_CATALOG_ENTRY` supports it.

sytuannguyen:
> Yes...but the code was essentially working in the original file. I can't see a substantive change in the new file. Is there a motivation for making a new class?

I agree that the existing mesh generator can be employed for cylindrical mesh (even spherical mesh) with some minor modifications. This will provide all the flexibility as a generic mesh generator.
However, we may want some optimizations specified for the wellbore geometry: derived wellbore, open wellbore, cased wellbore, mixing locally refined wellbore mesh and large scale regular resevoir mesh, etc. If we look closely to the wellbore problem we may find some additional features.
Give me sometime to play with this, if at the end we can confirm that a generic cylindrical mesh generator can cover all what we need for the wellbore problem, we will end up with modifying the existing class. Let's see.

rrsettgast:
@sytuannguyen @TotoGaz 
Let me just state that they are both cartesian generators, and both required mesh biasing. In some respect, anything we do the block cartesian mesh, we will want to do do the cylindrical mesh. It would be my strong preference not to fork the capability and thus increasing the amount of code we need to modify when new features are included, or bugs are fixed.

sytuannguyen:
> @sytuannguyen @TotoGaz
> Let me just state that they are both cartesian generators, and both required mesh biasing. In some respect, anything we do the block cartesian mesh, we will want to do do the cylindrical mesh. It would be my strong preference not to fork the capability and thus increasing the amount of code we need to modify when new features are included, or bugs are fixed.

Agree

sytuannguyen:
@rrsettgast @TotoGaz the InternalWellboreGenerator is now herited directly from InternelMeshGenerator. I will add some features specified for wellbore problems to it. The parent class need some additional cleanup related to 2D mesh (accutally it seems that it work only for 3D mesh?).

src/coreComponents/meshUtilities/InternalWellboreGenerator.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+namespace geosx
+{
+using namespace dataRepository;
+
+InternalWellboreGenerator::InternalWellboreGenerator( string const & name, Group * const parent ):
+  InternalMeshGenerator( name, parent )
+{}
+
+REGISTER_CATALOG_ENTRY( MeshGeneratorBase, InternalWellboreGenerator, string const &, Group * const )
```
TotoGaz, line 31: I assume that you have a new class only for registering the product in the factory, am I right?
I cannot see this "name" in your xml entry.
Another possibility is to inherit the specialized features in classes and keep the generic in a base class.

src/coreComponents/meshUtilities/InternalMeshGenerator.cpp
```diff
@@ -363,8 +285,13 @@ void InternalMeshGenerator::generateMesh( DomainPartition & domain )
   SortedArray< localIndex > & zposNodes = nodeSets.registerWrapper< SortedArray< localIndex > >( string( "zpos" ) ).reference();
   SortedArray< localIndex > & allNodes  = nodeSets.registerWrapper< SortedArray< localIndex > >( string( "all" ) ).reference();
 
+  // Nodesets for applying boundary conditions to a borehole problem
+  SortedArray< localIndex > & rnegNodes = nodeSets.registerWrapper< SortedArray< localIndex > >( string( "rneg" ) ).reference();
+  SortedArray< localIndex > & tnegNodes = nodeSets.registerWrapper< SortedArray< localIndex > >( string( "tneg" ) ).reference();
+  SortedArray< localIndex > & tposNodes = nodeSets.registerWrapper< SortedArray< localIndex > >( string( "tpos" ) ).reference();
+  SortedArray< localIndex > & rposNodes = nodeSets.registerWrapper< SortedArray< localIndex > >( string( "rpos" ) ).reference();
```
TotoGaz, line 292: Do these variables `rnegNodes` ... `rposNodes` make sense in strict cartesian grids?
Same, do `xnegNodes` ... `zposNodes`, `allNodes` all make sense for radial meshes?

If not, it could be possible to have a clean little class hierarchy to keep tighter scope on all of this.
We can also say that we don't care.

src/coreComponents/meshUtilities/InternalMeshGenerator.hpp
```diff
@@ -85,17 +42,44 @@ class InternalMeshGenerator : public MeshGeneratorBase
...
+   * @enum MeshType
+   *
+   * The options for mesh type
+   */
+  enum class MeshType : integer
+  {
+    Cartesian,
+    Cylindrical,
+    CylindricalSquareBoundary
+  };
```
TotoGaz, line 82: I like this since it brings clear view of possibilities and adds more readability.

src/coreComponents/meshUtilities/InternalMeshGenerator.hpp
```diff
@@ -335,11 +300,14 @@ class InternalMeshGenerator : public MeshGeneratorBase
    */
   inline bool isRadial()
   {
-    bool rval = (m_mapToRadial > 0);
+    bool rval = ( m_meshType == MeshType::Cylindrical || m_meshType == MeshType::CylindricalSquareBoundary );
     return rval;
```
TotoGaz, line 304: just `return m_meshType == MeshType::Cylindrical || m_meshType == MeshType::CylindricalSquareBoundary;` does the job.

src/coreComponents/meshUtilities/InternalMeshGenerator.cpp
```diff
@@ -873,7 +720,7 @@ void InternalMeshGenerator::generateMesh( DomainPartition & domain )
     }
   }
 
-  if( m_mapToRadial > 0 )
+  if( m_meshType == MeshType::Cylindrical || m_meshType == MeshType::CylindricalSquareBoundary )
```
TotoGaz, line 723: Like `isRadial()` ?

TotoGaz (commented):
Better PR to review: really easier to understand the patch 😉

src/coreComponents/meshUtilities/InternalWellboreGenerator.cpp
sytuannguyen: Yes I will add features specialized for wellbore class then I will update the xml file using the registered name.

src/coreComponents/meshUtilities/InternalMeshGenerator.cpp
sytuannguyen: rnegNodes, rposNodes, etc are now moved to the wellbore mesh class.

src/coreComponents/meshUtilities/InternalMeshGenerator.hpp
sytuannguyen: done

src/coreComponents/meshUtilities/InternalMeshGenerator.cpp
sytuannguyen: done

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.cpp
```diff
@@ -729,8 +730,12 @@ void SolidMechanicsLagrangianFEM::crsApplyTractionBC( real64 const time,
 
     globalIndex_array nodeDOF;
     real64_array nodeRHS;
-    integer const component = bc.getComponent();
 
+    // The flag component = 0, 1 and 2 are for the traction in the three orthogonal directions
+    // in the global cartesian coordinate system
+    // The flag component = -1 is for the normal traction
+
+    integer const component = bc.getComponent();
```
TotoGaz, line 738: So as I already mentioned, this documentation should be moved to the `doxygen` of `bc`.

src/coreComponents/meshUtilities/InternalWellboreGenerator.cpp
```diff
@@ -0,0 +1,89 @@
...
+  SortedArray< localIndex > & rnegNodes = nodeSets.registerWrapper< SortedArray< localIndex > >( string( "rneg" ) ).reference();
+  SortedArray< localIndex > & rposNodes = nodeSets.registerWrapper< SortedArray< localIndex > >( string( "rpos" ) ).reference();
+  SortedArray< localIndex > & tnegNodes = nodeSets.registerWrapper< SortedArray< localIndex > >( string( "tneg" ) ).reference();
+  SortedArray< localIndex > & tposNodes = nodeSets.registerWrapper< SortedArray< localIndex > >( string( "tpos" ) ).reference();
+
+  arrayView2d< real64, nodes::REFERENCE_POSITION_USD > const & X = nodeManager.referencePosition();
+  for( int localNodeIndex=0; localNodeIndex<nodeManager.size(); ++localNodeIndex )
+  {
+    real64 xCoord = X( localNodeIndex, 0 );
+    real64 yCoord = X( localNodeIndex, 1 );
```
TotoGaz, line 51: I don't know but is there an enum somewhere for this `0` and `1`?

Plus you should define your variables `const`.

TotoGaz (commented):
moving on ;)

src/coreComponents/meshUtilities/InternalMeshGenerator.hpp
```diff
@@ -82,20 +39,46 @@ class InternalMeshGenerator : public MeshGeneratorBase
...
+   * @enum MeshType
+   *
+   * The options for mesh type
+   */
+  enum class MeshType : integer
+  {
+    Cartesian,
+    Cylindrical,
+    CylindricalSquareBoundary
+  };
```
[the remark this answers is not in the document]
sytuannguyen: The viewKeyStruct can be defined as protected, I moved it now to protected . But  enum class seems to be linked to EnumString class (not sure). I see that both of them are defined as public in all the existing class. You may want to create a PR for systematically optimizing this over the whole code.

src/coreComponents/meshUtilities/InternalMeshGenerator.cpp
```diff
@@ -569,8 +480,8 @@ void InternalMeshGenerator::generateMesh( DomainPartition & domain )
 
           nodeLocalToGlobal[localNodeIndex] = nodeGlobalIndex( index );
 
-          // cartesian-specific nodesets
-          if( m_mapToRadial == 0 )
+          // Cartesian-specific nodesets
+          if( m_meshType == MeshType::Cartesian )
```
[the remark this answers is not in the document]
sytuannguyen: Agree and changed

src/coreComponents/meshUtilities/InternalMeshGenerator.hpp
```diff
@@ -82,20 +39,46 @@ class InternalMeshGenerator : public MeshGeneratorBase
    * @param[in] name of the InternalMeshGenerator
    * @param[in] parent point to the parent Group of the InternalMeshGenerator
    */
-  InternalMeshGenerator( const string & name,
-                         Group * const parent );
+  InternalMeshGenerator( const string & name, Group * const parent );
 
-  virtual ~InternalMeshGenerator() override;
+  virtual ~InternalMeshGenerator() override = default;
```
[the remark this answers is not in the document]
sytuannguyen: You mean virtual can be removed everywhere in derived classes? I tried this and it works. Is this optimal to do so?

src/coreComponents/meshUtilities/InternalWellboreGenerator.cpp
```diff
@@ -0,0 +1,89 @@
...
+    real64 rCoord = sqrt( xCoord * xCoord + yCoord * yCoord );
+
+    if( isEqual( rCoord, m_min[0], 1e-10 ) )
```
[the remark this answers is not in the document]
sytuannguyen: Added a protected variable m_positionTolerance with default 1e-10. But this is just a matematical precision so user may not need to play with this value.

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.cpp
sytuannguyen: Please jump in because I don't know how doxygen works yet,

src/coreComponents/meshUtilities/InternalWellboreGenerator.cpp
sytuannguyen: No, 0 and 1 are just the first and second components of a node's position.
`const` is added

src/coreComponents/meshUtilities/InternalMeshGenerator.hpp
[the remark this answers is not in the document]
TotoGaz: Yes, `override` implies `virtual` (obviously). So **when overriding** you can remove `virtual`, but you **should put** `override` (which is more powerful in that context).
`clang-tidy` may warn you about that, and we may clean it once for all with some script.

src/coreComponents/meshUtilities/InternalMeshGenerator.hpp
```diff
@@ -82,98 +39,108 @@ class InternalMeshGenerator : public MeshGeneratorBase
...
   int m_dim;
+
+  /// Minimum extent of mesh dimensions
+  real64 m_min[3];
+
+  /// Maximum extent of mesh dimensions
+  real64 m_max[3];
+
+  /// Position tolerance for adding nodes to nodesets
+  real64 m_positionTolerance = 1e-10;
```
TotoGaz, line 117: Just put an entry like
```c++
  registerWrapper( viewKeyStruct::trianglePatternString(), &m_trianglePattern ).
    setApplyDefaultValue( 0 ).
    setDescription( "Pattern by which to decompose the hex mesh into prisms (more explanation required)" );
```
with default value and we are done.

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.cpp
TotoGaz: The current doc is 
```c++
  /**
   * Accessor
   * @return const m_component
   */
  virtual int getComponent() const
  {
    return m_component;
  }
```
Just enter
```c++
  /**
   * Accessing the considered component.
   * The 0, 1 and 2 values are for the traction in the three orthogonal directions in the global cartesian coordinate system.
   * The -1 value is for the normal traction.
   * @return The component axis or a special value.
   */
  virtual int getComponent() const
  {
    return m_component;
  }
```

src/coreComponents/meshUtilities/InternalWellboreGenerator.cpp
TotoGaz: ok then

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.cpp
sytuannguyen: Tks, done

src/coreComponents/meshUtilities/InternalMeshGenerator.hpp
sytuannguyen: done

src/coreComponents/meshUtilities/InternalWellboreGenerator.hpp
```diff
@@ -0,0 +1,98 @@
...
+private:
+
+  /// Wellbore radius
+  real64 m_radius;
+
+  /// Farfield distance
+  real64 m_rOut;
+
+  /// Maximum tangent angle
+  real64 m_theta;
```
TotoGaz, line 80: Please document if it's degrees or radians.

src/coreComponents/meshUtilities/InternalWellboreGenerator.cpp
```diff
@@ -0,0 +1,216 @@
...
+  MeshBody & meshBody = meshBodies.registerGroup< MeshBody >( this->getName() );
+  MeshLevel & meshLevel0 = meshBody.registerGroup< MeshLevel >( string( "Level0" ));
+  NodeManager & nodeManager = meshLevel0.getNodeManager();
+  Group & nodeSets = nodeManager.sets();
+
+  // Wellbore nodesets
+  SortedArray< localIndex > & rnegNodes = nodeSets.registerWrapper< SortedArray< localIndex > >( string( "rneg" ) ).reference();
+  SortedArray< localIndex > & rposNodes = nodeSets.registerWrapper< SortedArray< localIndex > >( string( "rpos" ) ).reference();
+  SortedArray< localIndex > & tnegNodes = nodeSets.registerWrapper< SortedArray< localIndex > >( string( "tneg" ) ).reference();
+  SortedArray< localIndex > & tposNodes = nodeSets.registerWrapper< SortedArray< localIndex > >( string( "tpos" ) ).reference();
```
TotoGaz, line 104: I realize that those variable are indeed part of the user interface of your feature (people will ask for these specific names in their input files).
Could you document above that
```c++
// rneg, rpos, tneg and tpos are the named used by the end-used in the input files. Consider modifying them with care.
```

src/coreComponents/meshUtilities/InternalWellboreGenerator.cpp
```diff
@@ -0,0 +1,216 @@
...
+    // Get Cartesian coordinates of a reference centered vertical wellbore
+    real64 & xCoord = X( localNodeIndex, 0 );
+    real64 & yCoord = X( localNodeIndex, 1 );
+    real64 const & zCoord = X( localNodeIndex, 2 );
+
+    // Compute cylindrical coordinates of a reference centered vertical wellbore
+    real64 rCoord = sqrt( xCoord * xCoord + yCoord * yCoord );
+    real64 tCoord;
+
+    if( rCoord < 1e-10 )
```
TotoGaz, line 118: Please use the precision variable (make it `protected` if needed).

src/coreComponents/meshUtilities/InternalWellboreGenerator.cpp
```diff
@@ -0,0 +1,216 @@
...
+    real64 dx = xBottomCenter - xTopCenter;
+    real64 dy = yBottomCenter - yTopCenter;
+    real64 dz = zBottom - zTop;
+    real64 dr = sqrt( dx*dx + dy*dy );
+    real64 dl = sqrt( dr*dr + dz*dz );
+
+    // Azimuth from x-axis
+    real64 theta0;
+
+    if( dr < 1e-10 )
```
TotoGaz, line 177: Please use the precision variable (make it `protected` if needed).

src/coreComponents/meshUtilities/InternalWellboreGenerator.hpp
sytuannguyen: done

src/coreComponents/meshUtilities/InternalWellboreGenerator.cpp
sytuannguyen: done

src/coreComponents/meshUtilities/InternalWellboreGenerator.cpp
sytuannguyen: done

src/coreComponents/meshUtilities/InternalWellboreGenerator.cpp
sytuannguyen: done

sytuannguyen:
> @sytuannguyen Looks god to me. Are you ready to merge?

I think this is enough for this PR. Just today I got a build issue on Pangea after merging with develop as I mentioned on slack.
 
Need more wellbore PRs to address some features such as benchmark inclined wellbore, wellbore trajectory defined by a polyline, normal displacement boundary condition, Thermo-Hydro-Mechanical coupling for enable thermal loading on the wellbore etc.

rrsettgast:
Wishlist:

- [ ] radial inner mesh block transitioning to square outer boundary block. 
- [ ] auto discretization in radial direction given a number of theta elements to enforce "square" elements in the r-theta plane.

rrsettgast:
@sytuannguyen Just in case it wasn't conveyed properly, I am working on this branch. Please refrain from code changes until I hand it back to you.

cssherman:
@rrsettgast - This is about as good of a mesh you can create assuming a constant bias in the radial direction.  I expect that we could find the appropriate variable bias scheme to do better with a bit of math

```python
# Estimate the optimal bias, Ntheta/Nradial requiring that the first
# and last elements have a 1:1 aspect ratio.  Near the center of the
# region, the aspect ratio can reach ~2:1 
r_in = 0.2
r_out = 2.0

bias = (1.0 - (r_out / r_in)) / (1.0 + (r_out / r_in))
ratio = (2.0 * 3.14159 * r_in) / ((1.0 + bias) * (r_out - r_in))

print('Optimal constant bias = %1.4f' % (bias))
print('Optimal Ntheta / Nradial = %1.4f' % (ratio))
```

mesh_a

mesh_b

cssherman:
@rrsettgast  - Although, it looks like you may have already fixed it!

rrsettgast:
@cssherman This is the result of the approximation I put in:

Screen Shot 2021-03-12 at 11 48 02 AM

rrsettgast:
Ok. Multi-block with a combination of fixed spacing in the steel casing, auto spacing in the cement and rock, and a cartesian boundary applied to the rock.

Screen Shot 2021-03-12 at 1 18 52 PM

src/coreComponents/meshUtilities/InternalMeshGenerator.cpp
```diff
@@ -719,21 +606,26 @@ void InternalMeshGenerator::generateMesh( DomainPartition & domain )
...
+//                    index[1] = -1;
+//                    const localIndex firstNodeIndexR = numNodesInDir[1] * numNodesInDir[2] * ( index[0] - firstElemIndexInPartition[0] ) +
+//                                                       numNodesInDir[2] * ( index[1] - firstElemIndexInPartition[1] ) +
+//                                                       ( index[2] - firstElemIndexInPartition[2] );
+//                    nodeOfBox[2] = numNodesInDir[1] * numNodesInDir[2] + numNodesInDir[2] + firstNodeIndexR;
+//                    nodeOfBox[3] = numNodesInDir[2] + firstNodeIndexR;
+//                    nodeOfBox[6] = numNodesInDir[1] * numNodesInDir[2] + numNodesInDir[2] + firstNodeIndexR + 1;
+//                    nodeOfBox[7] = numNodesInDir[2] + firstNodeIndexR + 1;
+//                  }
+//                }
```
rrsettgast, line 628:
```suggestion
```

rrsettgast (commented):
@sytuannguyen I commented out the trajectory stuff and some of the nodeset stuff.

Can you place anything that moves nodes into the `coordinateTransformation` function? This is probably the best place for your nodeset specification as well.

src/coreComponents/dataRepository/Group.hpp
```diff
@@ -114,10 +114,10 @@ class Group
...
   Group( Group const & ) = delete;
 
-  /**
-   * @brief Deleted move constructor.
-   */
-  Group( Group const && ) = delete;
+//  /**
+//   * @brief Deleted move constructor.
+//   */
+//  Group( Group const && ) = delete;
```
rrsettgast, line 120:
```suggestion
```

src/coreComponents/dataRepository/xmlWrapper.cpp
```diff
@@ -22,27 +22,32 @@ namespace geosx
...
     }
   }
-  GEOSX_ERROR_IF( count!=3, "incorrect number of components specified for R1Tensor" );
+  GEOSX_ERROR_IF( count!=SIZE, "incorrect number of components specified for Tensor" );
 }
 
+
+template void xmlWrapper::stringToInputVariable( Tensor< real64, 3 > & target, string const & inputValue );
+template void xmlWrapper::stringToInputVariable( Tensor< real64, 6 > & target, string const & inputValue );
+
```
rrsettgast, line 50:
```suggestion
template void xmlWrapper::stringToInputVariable( R1Tensor & target, string const & inputValue );
template void xmlWrapper::stringToInputVariable( R2SymTensor & target, string const & inputValue );

```

src/coreComponents/managers/FieldSpecification/FieldSpecificationBase.cpp
```diff
@@ -96,7 +96,9 @@ FieldSpecificationBase::getCatalog()
 
 
 void FieldSpecificationBase::postProcessInput()
-{}
+{
+//  LvArray::tensorOps::l2Norm<3>( m_direction );
```
rrsettgast, line 100: We use `m_direction` as a full vector (direction + magnitude). Perhaps we should change the name. Objections? @klevzoff @francoishamon 

image

src/coreComponents/managers/FieldSpecification/TractionBoundaryCondition.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+using namespace dataRepository;
+
+TractionBoundaryCondition::TractionBoundaryCondition( string const & name, Group * parent ):
+  FieldSpecificationBase( name, parent ),
+  m_tractionType( 0 ),
+  m_inputStress{}//,
+//  m_stressFunctionNames(),
+//  m_useStressFunctions(false),
+//  m_stressFunctions{nullptr}
```
rrsettgast, line 34: These are for future capabilities

src/coreComponents/mesh/FaceManager.cpp
```diff
@@ -604,7 +604,7 @@ void FaceManager::computeGeometry( NodeManager const & nodeManager )
 
     // This needs to be done somewhere else, also we probably shouldn't be orienting the normals like this.
     // Set normal orientation according to a global criterion
-    computationalGeometry::FixNormalOrientation_3D( m_faceNormal[ faceID ] );
+//    computationalGeometry::FixNormalOrientation_3D( m_faceNormal[ faceID ] );
```
rrsettgast, line 607: @andrea-franceschini This breaks the assumed convention of external faces having outward pointing normals. Is this still required?

andrea-franceschini: You are right. This was used to have a consistent orientation for a fracture, that, being an internal surface, does not have an outward direction. But (I assume) we can safely comment it out. @castelletto1 do you think there are specific constraints to keep it?

rrsettgast:
@sytuannguyen How important is your skewed wellbore capability?

sytuannguyen:
> @sytuannguyen How important is your skewed wellbore capability?

It is very important to have a capability of modeling inclined wellbore because of following reasons:

- Most wells are inclined, i.e. deviated from the vertical direction
- Defining inclined wellbore in a mesh body allows applying directly the main in-situ stress as boundary/initial conditions without the need of rotating these stresses to the local wellbore orientation
- Easy to work with inclined wellbore in anisotropic medium (transversely isotropic or orthotropic media) of which the anisotropic directions do not coincide with the well orientation
- Study for optimization of well trajectory, etc.

However, I realized that defining inclined wellbore require one of following two features
- A capability of defining fixed normal displacement to an inclined face. This is to fix the faces at theta=0 and theta=360°. This feature seems complicated to implement.
- Or a capability for connecting the nodesets on the faces at theta=0 and theta=360° as one such that we don't need to define the boundary conditions there.

Anyway, I think it will be OK if we want to skip the development this capability because we can always consider external mesh.

sytuannguyen:
@rrsettgast I will continue with following tasks
- [] Pass Pangea build and then Travis build
- [] Setup and benchmark the case of elastic open hole versus analytical results
- [] Setup and benchmark the case of poroelastic open hole versus analytical results
- [] Merge to develop and resolve conflicts

sytuannguyen:
The benchmark of the elastic openhole is perfect

image

sytuannguyen:
@rrsettgast I testing the poroelastic problem, pore pressure cannot be applied to rneg

image

TotoGaz:
> The benchmark of the elastic openhole is perfect

Great!

Could you plot the difference between the two values also (absolute and relative)?
I suspect the two solutions to be not so close indeed; and I would be interested in knowing the discrepancy.

rrsettgast:
> @rrsettgast I'm testing the poroelastic problem but pore pressure (not normal traction) cannot be applied to rneg by using following input.
> 
> image

There is no pore pressure on the face. Well... @klevzoff do the face pressure boundary conditions still work?

klevzoff:
> > @rrsettgast I'm testing the poroelastic problem but pore pressure (not normal traction) cannot be applied to rneg by using following input.
> > image
> 
> There is no pore pressure on the face. Well... @klevzoff do the face pressure boundary conditions still work?

Yes, we have integrated tests using them, including poroelastic ones (`PoroElastic_Terzaghi_FIM.xml`, `PoroElastic_Mandel_FIM.xml`). The specification seems valid, so I'm not sure what's going on here. Have you verified that face sets are created from node sets correctly for the wellbore generator? I can hop on the branch and debug if you tell me which input file to run.

sytuannguyen:
> I can hop on the branch and debug if you tell me which input file to run.

nvm, I got it.

sytuannguyen:
> Great!
> 
> Could you plot the difference between the two values also (absolute and relative)?
> I suspect the two solutions to be not so close indeed; and I would be interested in knowing the discrepancy.

@TotoGaz Good idea!
image

TotoGaz:
@sytuannguyen Thx! Please do not put the errors and the result on the same plot (like upper), it's more complicated to read.

Also, I think that ~3% is not so little for an analytical comparison.
If you try to increase the mesh or maybe use upper order discretization, are we able to make this error vanish?

rrsettgast:
> * A capability of defining fixed normal displacement to an inclined face. This is to fix the faces at theta=0 and theta=360°. This feature seems complicated to implement because it is not similar to any of existing boundary conditions in GEOSX including the normal traction boundary condition.
> * Or a capability for connecting the nodesets on the faces at theta=0 and theta=360° as one such that we don't need to define the boundary conditions there. This option sounds simpler than the first one, at least we can easily do this using external mesh.
> 

I have tied collapsed the nodes at theta=0,360 if it is a full mesh. So I think this shouldn't be a problem. So it is a contiguous mesh across the theta = 0/360 boundary. The nodeset at that location exists, but is pretty irrelevant.

sytuannguyen:
> @sytuannguyen Thx! Please do not put the errors and the result on the same plot (like upper), it's more complicated to read.
> 
> Also, I think that ~3% is not so little for an analytical comparison.
> If you try to increase the mesh or maybe use upper order discretization, are we able to make this error vanish?

sytuannguyen:
@rrsettgast do you know why all the `cuda` builds are fail?

src/coreComponents/meshUtilities/InternalWellboreGenerator.hpp
```diff
@@ -0,0 +1,116 @@
...
+   * @return string that contains the key name to InternalWellboreGenerator in the Catalog
+   */
+  static string catalogName() { return "InternalWellbore"; }
+
+  virtual void reduceNumNodesForPeriodicBoundary( integer ( &numNodes )[3] ) override final;
+
+  virtual void setNodeGlobalIndicesOnPeriodicBoundary( int (& index)[3],
+                                                       real64 ( &minExtent )[3],
+                                                       real64 ( &maxExtent )[3],
+                                                       arraySlice1d< real64 const > const & X,
```
klevzoff, line 57: @sytuannguyen I think something like this will fix cuda build (obviously, same applied to the parent class):
```suggestion
                                                       arraySlice1d< real64 const, nodes::REFERENCE_POSITION_USD-1 > const & X,
```

sytuannguyen: @klevzoff super! It passes now, tks.

sytuannguyen:
@rrsettgast @TotoGaz @corbett5  It is not possible to run the integrated tests in Pangea (all are fail and I dont know why) so I cannot work on the rebaseline and adding new tests. Could someone help me on this?

sytuannguyen:
> Please do not put the errors and the result on the same plot (like upper), it's more complicated to read.
> Also, I think that ~3% is not so little for an analytical comparison. If you try to increase the mesh or maybe use upper order discretization, are we able to make this error vanish?

@TotoGaz sorry for this delayed answer, I tried to pass the Travis build before addressing your comment. Below are the results obtained by refining the mesh. Obviously we can always reduce error by refining the mesh. We can conclude that the benchmark is perfect.

image

rrsettgast:
replaced with a rebased branch.... new PR is https://github.com/GEOSX/GEOSX/pull/1393

PR #1340, opened 2021-03-02
Changed how the finite element kernel objects are constructed.

corbett5:
This removes the need for the template of templates in `regionBasedKernelApplication`, as well as the variadic parameter pack. The benefit is that it simplifies both the use and the implementation of `regionBasedKernelApplication`. The downside is that now there's an extra class that needs to be created for each of the kernel types.

### DIFF
--- a/src/coreComponents/finiteElement/kernelInterface/KernelBase.hpp
+++ b/src/coreComponents/finiteElement/kernelInterface/KernelBase.hpp
@@ -27,61 +27,6 @@
 #include "mesh/ElementRegionManager.hpp"
 #include "common/GEOS_RAJA_Interface.hpp"
 
-
-
-#if defined(__APPLE__)
-/// Use camp::tuple to hold constructor params.
-#define CONSTRUCTOR_PARAM_OPTION 2
-#else
-/// Use std::tuple to hold constructor params.
-#define CONSTRUCTOR_PARAM_OPTION 1
-#endif
-
-#if CONSTRUCTOR_PARAM_OPTION==1
-namespace std
-{
-namespace detail
-{
-/**
- * @brief Implementation of std::make_from_tuple()
- * @tparam T
- * @tparam Tuple
- * @tparam I
- * @tparam T
- * @tparam Tuple
- * @tparam I
- * @param t
- * @param
- * @return
- */
-template< class T, class Tuple, std::size_t... I >
-constexpr T make_from_tuple_impl( Tuple && t, std::index_sequence< I... > )
-{
-  return T( std::get< I >( std::forward< Tuple >( t ))... );
-}
-} // namespace detail
-
-/**
- * @brief Implementation of std::make_from_tuple()
- * @tparam T
- * @tparam Tuple
- * @tparam T
- * @tparam Tuple
- * @param t
- * @return
- */
-template< class T, class Tuple >
-constexpr T make_from_tuple( Tuple && t )
-{
-  return detail::make_from_tuple_impl< T >( std::forward< Tuple >( t ),
-                                            std::make_index_sequence< std::tuple_size< std::remove_reference_t< Tuple > >::value >{} );
-}
-
-}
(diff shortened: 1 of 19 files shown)


klevzoff (approved):
LGTM. I'm assuming this makes things easier for JITting kernels?

My only suggestion: if we really wanted to avoid writing the `...Wrapper` classes for each kernel that all look the same, we could just put the old tuple-forwarding logic into a reusable "kernel factory" class, sort of like this:
```c++
template<template<typename SUBREGION_TYPE, 
                  typename CONSTITUTIVE_TYPE,
                  typename FE_TYPE> class KERNEL_TYPE,
         typename ... ARGS>
class KernelFactory
{
public:

  KernelFactory(ARGS ... args) : m_args(std::forward_as_tuple(args...)) {}

  template< typename SUBREGION_TYPE, typename CONSTITUTIVE_TYPE, typename FE_TYPE >
  auto
  createKernel( Manager & nodeManager,
                Manager const & edgeManager,
                Manager const & faceManager,
                SUBREGION_TYPE const & elementSubRegion,
                FE_TYPE const & finiteElementSpace,
                CONSTITUTIVE_TYPE & inputConstitutiveType )
  {
    auto std_args = std::forward_as_tuple(nodeManager, 
                                          edgeManager,
                                          faceManager,
                                          elementSubRegion,
                                          finiteElementSpace,
                                          inputConstitutiveType);
    auto all_args = std::tuple_cat(std_args, m_args);
    return std::make_from_tuple<KERNEL_TYPE<SUBREGION_TYPE, CONSTITUTIVE_TYPE, FE_TYPE>>(all_args);
  }

private:
  std::tuple<ARGS & ...> m_args;
};
```
and then use it like
```c++
using LaplaceFEMKernelFactory = KernelFactory<LaplaceFEMKernel,
                                              arrayView1d< globalIndex const > const &,
                                              globalIndex,
                                              CRSMatrixView< real64, globalIndex const > const &,
                                              arrayView1d< real64 > const &,
                                              string const & >
```
The rest would be exactly as you have in this PR ([example](https://godbolt.org/z/1Tj71f))
(btw, I prefer the term "Factory" to "Wrapper" here as a more specific one, although it has its own connotations).

I don't know if this improves readability

TotoGaz:
For what I understood about this code, I would not call these `KernelWrappers` but `KernelBuilders`/`KernelSuppliers`/`KernelProviders` instead.

It could also be interesting not to use a `createKernel` member function but some `operator()`, such that a kernel could be build with a single lambda function (Since I expect some (most?) of these classes are used only once, that would shrink their visibility to the absolute minimum).

corbett5:
@klevzoff yeah it does make the JIT easier, and your suggestion is great.

@TotoGaz I did consider a lambda, but as Sergey mentioned I'm working on a JIT and JIT'ing a function with a lambda as a template parameter might not be possible with my implementation.

rrsettgast:
@corbett5 In combination with @klevzoff's suggestion, I think you may be able to provide a second interface to that maintains the original call signature of `regionBasedKernelApplication` and invokes the factory, then calls the new version of `regionBasedKernelApplication`. Does that break the jitty approach you are using?

corbett5:
@rrsettgast that would work.

corbett5:
I made the changes @klevzoff suggested, including changing the name from `FooWrapper` to `FooFactory`. The only factory I wasn't able to change was the sparsity, but it's much cleaner now.

@rrsettgast I chose not to make the changes you suggested, although if it really was an entirely separate function it would only require a few lines. I'm going to be on vacation for a few days, so if you'd like to make those changes feel free to do so. While going through the existing interface does make things less ugly for the JIT, any approach other than using a lambda should work. Once I have a something fully featured there will probably be a lot of iteration on how to clean it up and make it as user-friendly as possible.

rrsettgast (approved):
@corbett5 Can you update this with develop and merge it?

PR #901, opened 2020-04-22
Add hybrid FVM compositional multiphase solver

francoishamon:
This PR implements a mixed hybrid solver for compositional multiphase flow with buoyancy. I organized the compositional multiphase flow solver as follows:
- A base class `CompositionalMultiphaseBase` that contains the property update functions and the computation of the accumulation term
- A derived class `CompositionalMultiphaseFVM` that implements the cell-centered TPFA-MPFA flux assembly and boundary conditions.
- A derived class `CompositionalMultiphaseHybridFVM` that implements the mixed hybrid flux assembly.

Note that the large number of lines changed is due to an integrated test using an imported mesh file (i.e, not actual code changes).

To-do list:
- [x] Mixed-hybrid compositional multiphase solver working
- [x] Integrated tests
- [x] Unit tests pass on Lassen
- [x] Implement MGR recipe
- [x] Add buoyancy

https://github.com/GEOSX/integratedTests/pull/119

PR #1341, opened 2021-03-02
Fixed bug in the computation of the Peaceman index

francoishamon:
This PR fixes a bug in the computation of the well Peaceman index that occurs when the perforation is placed exactly at the center of the well element (my bad). It also fixes some well function names that were not following the camelCase convention. 

This PR does not require a rebaseline (the bug does not appear in the integrated tests).

### DIFF
--- a/src/coreComponents/meshUtilities/PerforationData.cpp
+++ b/src/coreComponents/meshUtilities/PerforationData.cpp
@@ -21,9 +21,7 @@
 #include "mpiCommunications/MpiWrapper.hpp"
 #include "managers/DomainPartition.hpp"
 #include "mesh/MeshForLoopInterface.hpp"
-#include "meshUtilities/Perforation.hpp"
 #include "meshUtilities/ComputationalGeometry.hpp"
-
 
 namespace geosx
 {
@@ -52,9 +50,9 @@ namespace
 
 /**
  * @brief Check if the well is along the x-, y-, or z- directions.
- * @tparam VEC_TYPE type of @p vecWellElemCenterToPerf
+ * @tparam VEC_TYPE type of @p vecTopNodeToBottom
  * @tparam PERM_TYPE type of @p perm
- * @param[in] vecWellElemCenterToPerf vector connecting the well element center to the perforation
+ * @param[in] vecTopToBottom vector connecting the well element center to the perforation
  * @param[in] dx dimension of the element in the x-direction
  * @param[in] dy dimension of the element in the y-direction
  * @param[in] dz dimension of the element in the z-direction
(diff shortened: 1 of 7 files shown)


src/coreComponents/meshUtilities/PerforationData.cpp
```diff
@@ -78,8 +76,8 @@ void DecideWellDirection( VEC_TYPE const & vecWellElemCenterToPerf,
                           real64 & k2 )
 {
   // vertical well (approximately along the z-direction)
-  if( fabs( vecWellElemCenterToPerf[2] ) > fabs( vecWellElemCenterToPerf[0] )
-      && fabs( vecWellElemCenterToPerf[2] ) > fabs( vecWellElemCenterToPerf[1] ) )
+  if( fabs( vecTopToBottom[2] ) > fabs( vecTopToBottom[0] )
```
TotoGaz, line 79: I would rename `vecTopToBottom` to `topToBottom`. Or at least `topToBottomVec`?

src/coreComponents/meshUtilities/PerforationData.cpp
```diff
@@ -151,15 +150,21 @@ void PerforationData::computeWellTransmissibility( MeshLevel const & mesh,
...
 
-    // compute the vector perforation - well elem center
-    // this vector will be used to decide whether this is a vectical well or not
-    localIndex const wellElemIndex   = m_wellElementIndex[iperf];
-    real64 vecWellElemCenterToPerf[3] = LVARRAY_TENSOROPS_INIT_LOCAL_3( wellElemCenter[wellElemIndex] );
-    LvArray::tensorOps::subtract< 3 >( vecWellElemCenterToPerf, m_location[iperf] );
+    // get the index of the well element, and of the neighboring well nodes
+    localIndex const wellElemIndex = m_wellElementIndex[iperf];
+    localIndex const topNode = elemToNodeMap[wellElemIndex][InternalWellGenerator::NodeLocation::TOP];
+    localIndex const bottomNode = elemToNodeMap[wellElemIndex][InternalWellGenerator::NodeLocation::BOTTOM];
```
TotoGaz, line 156: One funny test could be to swap the `0/1` values of `TOP` and `BOTTOM` to see if it still works. 😉

TotoGaz (approved):
LGTM

src/coreComponents/meshUtilities/PerforationData.cpp
francoishamon: Done, I renamed it to `topToBottomVec`

src/coreComponents/meshUtilities/PerforationData.cpp
francoishamon: I swapped the 0 and 1 (locally, not in the version I committed) and it worked. The integrated tests pass, except for the node list of the `WellElementSubRegion` that is obsviously different if 0 and 1 are swapped.

As a side note, I noticed that I broke an integrated test `deadoil_3ph_staircase_3d` with #901, I will fix it just after this PR using the Dead-Oil branch.

PR #1343, opened 2021-03-03
Fix VTK output to put plot files under output path

klevzoff:
Resolves https://github.com/GEOSX/GEOSX/issues/1342

Collateral "damage" is mostly due to some code cleanup related to path handling.

Rebaseline PR: https://github.com/GEOSX/integratedTests/pull/121
(only for tests using VTK output)

### DIFF
--- a/src/coreComponents/common/Path.cpp
+++ b/src/coreComponents/common/Path.cpp
@@ -24,13 +24,12 @@
 namespace geosx
 {
 
-///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
-void getAbsolutePath( std::string const & path, std::string & absolutePath )
+std::string getAbsolutePath( std::string const & path )
 {
   char absFilePath[ PATH_MAX + 1 ];
   if( realpath( path.data(), absFilePath ) )
   {
-    absolutePath = absFilePath;
+    return absFilePath;
   }
   else
   {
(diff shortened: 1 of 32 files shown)


src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp
```diff
@@ -510,28 +478,42 @@ void VTKPolyDataWriterInterface::writeUnstructuredGrid( vtkSmartPointer< vtkUnst
...
+                                        DomainPartition const & domain )
 {
-  createTimeStepSubFolder( time );
+  string const stepSubFolder = getTimeStepSubFolder( time );
+  if( MpiWrapper::commRank( MPI_COMM_GEOSX ) == 0 )
+  {
+    if( m_previousCycle == -1 )
+    {
+      makeDirsForPath( joinPath( m_outputDir, m_outputName ) );
+    }
```
klevzoff, line 496: The VTK output folder is now created on first write as opposed to writer object creation time, because the name of the directory is set *after* reading the XML input.

src/coreComponents/fileIO/vtk/VTKVTMWriter.hpp
```diff
@@ -33,7 +33,7 @@ class VTKVTMWriter
    * @brief Build the VTM Writer
    * @param[in] filePath path to the file
    */
-  VTKVTMWriter( string const & filePath );
+  explicit VTKVTMWriter( string const & filePath );
```
TotoGaz, line 36: Why didn't you change `string const & filePath` to `string filePath` like you did for others?

src/coreComponents/common/Path.cpp
```diff
@@ -25,12 +25,12 @@ namespace geosx
 {
 
 ///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
-void getAbsolutePath( std::string const & path, std::string & absolutePath )
+std::string getAbsolutePath( std::string const & path )
```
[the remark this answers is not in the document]
klevzoff: Yes... but there's a circular inclusion problem that I didn't want to touch in this PR. Namely, `DataTypes.hpp` needs to include `Path.hpp` in order to provide some related typedefs and lambda dispatch on it. On the other hand, `Path.hpp` would need to include `DataTypes.hpp` in order to gain access to `geosx::string`.

One solution is to split a small portion of `DataTypes.hpp` into a separate header, `BasicTypes.hpp`, that will provide aliases like `localIndex`, 'globalIndex', `real64`, `string`. It can then be independently included in `Path.hpp` and `DataTypes.hpp`. I didn't do it here as it would create merge conflicts for the BoomerAMG PR, but we can address this subsequently.

src/coreComponents/common/Path.cpp
```diff
@@ -39,78 +39,83 @@ void getAbsolutePath( std::string const & path, std::string & absolutePath )
...
+void makeDirectory( std::string const & path )
 {
-  constexpr mode_t mode = 0770;     // user and group rwx permissions
+  constexpr mode_t mode = 0770; // user and group rwx permissions
+  int const err = mkdir( path.c_str(), mode );
+  GEOSX_THROW_IF( err && ( errno != EEXIST ), "Failed to create directory: " << path, std::runtime_error );
+}
 
+///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+void makeDirsForPath( std::string const & path )
```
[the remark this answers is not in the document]
klevzoff: You're right, and https://github.com/GEOSX/GEOSX/issues/1344 is the reason we've been missing docs here and there :) Will add.

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp
```diff
@@ -510,28 +478,42 @@ void VTKPolyDataWriterInterface::writeUnstructuredGrid( vtkSmartPointer< vtkUnst
...
+    makeDirectory( joinPath( m_outputDir, stepSubFolder ) );
+  }
```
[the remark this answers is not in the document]
klevzoff: No, it's an oversight. Good catch!

src/coreComponents/fileIO/vtk/VTKVTMWriter.hpp
klevzoff: Indeed, I should change. I tend to pass-by-value into functions that make a direct copy of the argument, which is also the case for this constructor. I didn't notice because I didn't modify anything in `VTKVTMWriter.cpp` file, which would cause me to spot this pattern.

src/coreComponents/common/Path.cpp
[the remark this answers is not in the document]
TotoGaz: ok then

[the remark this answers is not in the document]
rrsettgast: It is intentional that we kept `std::string` here. `Path` is its own type derived exclusively from `std::string`.

src/coreComponents/common/Path.hpp
```diff
@@ -77,17 +77,21 @@ class Path : public std::string
 /*!
  * @brief Gets the absolute path of a file
  * @param[in] path the relative path to the file
- * @param[out] absolutePath the absolute path to the file
+ * @return the absolute path to the file resolved from current working directory
+ * @exception InputError if absolute path could not be resolved
  */
-void getAbsolutePath( std::string const & path, std::string & absolutePath );
+std::string getAbsolutePath( std::string const & path );
```
rrsettgast, line 83: Re: `string` vs `std::string`. Here we would be passing `string` as the argument, so it should have a UDC to `std::string`, or be an alias for `std::string`.

src/coreComponents/fileIO/vtk/VTKPVDWriter.cpp
```diff
@@ -20,8 +20,8 @@ namespace geosx
 {
 namespace vtk
 {
-VTKPVDWriter::VTKPVDWriter( string const & fileName ):
-  m_fileName( fileName )
+VTKPVDWriter::VTKPVDWriter( string fileName ):
+  m_fileName( std::move( fileName ) )
```
rrsettgast, line 24: why the move?

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.hpp
```diff
@@ -35,7 +35,7 @@
 
 namespace geosx
 {
-using namespace dataRepository;
+//using namespace dataRepository;
```
rrsettgast, line 38:
```suggestion
```

nice catch.

rrsettgast (approved):

PR #1346, opened 2021-03-04
Update Caliper.rst

corbett5:


corbett5:
We can't point to the docs for our version of caliper but this this is better than the previous link.

klevzoff (approved):

PR #1352, opened 2021-03-11
WIP - circular dependencies / object libraries build

bmhan12:
Get all unit tests to compile/link. 
- [x] Move tests that depend on components above their level to `coreComponents/unitTests` 
- [x] Verify pass unit tests, and non-PVT integratedTests.
- [x] Clean up dependencies for unit test. Unit test should only depend on TPL's and the component itself. 

Clean up component dependencies (start with top level components, and config/build after each component).
- [ ] Collapse all the mpi, cuda, openmp, etc dependency inclusions at the lowest level possible, and remove from all upper level components.
- [x] Remove all the branched versions of dependencies for each component. 

Test non-object libraries option.


rrsettgast:
@bmhan12 Are all those circular dependencies in the description still present? Can you checkbox them?

bmhan12:
> @bmhan12 Are all those circular dependencies in the description still present? Can you checkbox them?

Yes, these dependencies are all still present.

TotoGaz:
One potential way to flush those dependencies out could be to find all the **forward declarations** in our code (with regex like `^class [a-zA-Z]+;`).
While some may be legitimate I believe that quite some of them should be replaced by include, and maybe if it's not already the case it's because of circular dependencies.

rrsettgast:
> One potential way to flush those dependencies out could be to find all the **forward declarations** in our code (with regex like `^class [a-zA-Z]+;`).
> While some may be legitimate I believe that quite some of them should be replaced by include, and maybe if it's not already the case it's because of circular dependencies.

I think in general, when we forward declare `FOO` in a header, we `#include "FOO.h"` in the source.

bmhan12:
> @bmhan12 This is looking pretty good. We can link and run `geosx`. Still some work to compile unit tests. I think we can start simplifying....also no PVT or PAMELA enabled until we are ready to merge because it breaks the dependency chain.

Great

klevzoff:
Tests from `linearAlgebra` (except for `testDofManager`) don't actually need to depend on anything except `common`. They only use `geosx::basicSetup` for convenience, but can be changed to manually perform initialization of individual components they need (MPI, lvarray, logger, LAI). It's also a bit strange that in order to call these individual initialization routines they have to depend on `mainInterface` which is very high up the dependency tree. I think all of these routines should live in `common`:
```c++
  setupMPI( argc, argv );
  setupLogger();
  setupLvArray();
  setupOpenMP();
  setupMKL();
```

host-configs/darwin-clang.cmake
```diff
@@ -32,5 +32,5 @@ set(ENABLE_DOXYGEN OFF CACHE BOOL "" FORCE)
 
 set(ENABLE_GEOSX_PTP ON CACHE BOOL "" FORCE)
 
-set(GEOSX_TPL_DIR "/usr/local/GEOSX/GEOSX_TPL" CACHE PATH "" FORCE)
+set(GEOSX_TPL_DIR "/Users/settgast1/Codes/geosx/thirdPartyLibs/install-darwin-clang-release" CACHE PATH "" FORCE)
```
rrsettgast, line 35:
```suggestion
set(GEOSX_TPL_DIR "/usr/local/GEOSX/GEOSX_TPL" CACHE PATH "" FORCE)
```

src/coreComponents/physicsSolvers/CMakeLists.txt
```diff
@@ -133,6 +124,34 @@ if ( ENABLE_CUDA )
...
+         fluidFlow/CompositionalMultiphaseFVM.cpp
+         fluidFlow/CompositionalMultiphaseHybridFVM.cpp
+         fluidFlow/CompositionalMultiphaseBaseKernels.cpp
+         fluidFlow/CompositionalMultiphaseFVMKernels.cpp
+         fluidFlow/CompositionalMultiphaseHybridFVMKernels.cpp
+         multiphysics/CompositionalMultiphaseReservoir.cpp
+         multiphysics/MultiphasePoroelasticSolver.cpp
+         fluidFlow/wells/CompositionalMultiphaseWell.cpp
+         )
+endif()
```
klevzoff, line 153: This is not correct anymore, we have multiphase constitutive models that run natively in GEOSX, without `PVTPackage`. Perhaps constitutive dispatch needs to be fixed to avoid `PVTPackage`-dependent models when it's not enabled. We can do it in a follow up.

src/coreComponents/unitTests/dataRepositoryTests/testRestartExtended.cpp
```diff
@@ -479,7 +478,9 @@ int main( int argc, char * argv[] )
 {
   testing::InitGoogleTest( &argc, argv );
 
-  geosx::GeosxState state( geosx::basicSetup( argc, argv ) );
+//  geosx::GeosxState state( geosx::basicSetup( argc, argv ) );
+
```
klevzoff, line 482:
```suggestion
```

src/coreComponents/unitTests/dataRepositoryTests/testRestartBasic.cpp
```diff
@@ -44,7 +44,7 @@ class SingleWrapperTest : public ::testing::Test
 public:
 
   SingleWrapperTest():
-    m_state( std::make_unique< CommandLineOptions >( g_commandLineOptions ) ),
+//    m_state( std::make_unique< CommandLineOptions >( g_commandLineOptions ) ),
```
klevzoff, line 47:
```suggestion
```

src/coreComponents/unitTests/dataRepositoryTests/testRestartBasic.cpp
```diff
@@ -91,7 +91,7 @@ class SingleWrapperTest : public ::testing::Test
   string const m_wrapperName = "wrapper";
   string const m_fileName = "testRestartBasic_SingleWrapperTest";
 
-  GeosxState m_state;
+//  GeosxState m_state;
```
klevzoff, line 94:
```suggestion
```

src/coreComponents/virtualElement/VirtualElementBase.hpp
```diff
@@ -35,7 +35,7 @@ class VirtualElementBase
 public:
 
   /// Default constructor.
-  VirtualElementBase() = default;
+  VirtualElementBase();
```
klevzoff, line 38: This seems to have been removed by accident?

klevzoff (approved):
Looks good! Mostly cleanup suggestions.

Some additional thoughts/comments:

1. Is the object library build fully functional? If so, would be great to have at least one CI build use it, or it will quickly become dysfunctional again.

1. I don't know how I feel about `FunctionManager`, `FieldSpecificationManager` and `GeometricObjectManager` going back to being "singleton"-like. This prevents two or more simulations existing at the same time (e.g. run in lockstep via python interface). The alternative is to access them like we do occasionally with `DomainPartition`, i.e. `getGroupByPath( "/Problem/functions" )`. Neither is particularly pretty, but singletons are a stronger code smell IMO.

1. There are more header guards that need updating, but I can't comment on specific files since they weren't modified... pretty much any `.hpp` file that was moved with no line changes should be checked. One example is `fieldSpecification/SourceFluxBoundaryCondition.hpp` but there are more. Perhaps this can be done in a dedicated cleanup PR because there are still some inconsistencies in header guard styles.

scripts/config-build.py
```diff
@@ -98,6 +98,12 @@ def setup_ats(scriptsdir, buildpath):
                     action='store_true',
                     help="build third party libraries")
 
+#parser.add_argument("-gvz",
```
corbett5, line 101: Why not keep this?

src/cmake/GeosxOptions.cmake
```diff
@@ -116,8 +116,8 @@ if( ${CMAKE_MAKE_PROGRAM} STREQUAL "ninja" OR ${CMAKE_MAKE_PROGRAM} MATCHES ".*/
 endif()
 
 if( CMAKE_HOST_APPLE )
-    set(GEOSX_LINK_PREPEND_FLAG "-Wl,-force_load" CACHE STRING "")
-    set(GEOSX_LINK_POSTPEND_FLAG "" CACHE STRING "")
+#    set(GEOSX_LINK_PREPEND_FLAG "-Wl,-force_load" CACHE STRING "")
+#    set(GEOSX_LINK_POSTPEND_FLAG "" CACHE STRING "")
```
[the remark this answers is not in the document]
corbett5: Below as well.

src/coreComponents/codingUtilities/CMakeLists.txt
```diff
@@ -18,7 +19,7 @@ set( codingUtilities_sources
      StringUtilities.cpp
    )
 
-set( dependencyList common )
+set( dependencyList common lvarray )
```
corbett5, line 22: Common doesn't depend on LvArray?

src/coreComponents/constitutive/CMakeLists.txt
```diff
@@ -109,9 +109,14 @@ set( constitutive_sources
...
-  set( dependencyList dataRepository)
+  set( dependencyList dataRepository functions)
 else()
-  set( dependencyList common )
+  set( dependencyList 
+       common
+       managers
+       codingUtilities
+       dataRepository
+     )
```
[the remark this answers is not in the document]
corbett5: Same.

src/coreComponents/discretizationMethods/CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
...
+  set( dependencyList ${dependencyList} cuda )
+endif( )
+
+if( ENABLE_MPI )
+  set( dependencyList ${dependencyList} mpi )
+endif()
+
+if( ENABLE_CALIPER )
+  set( dependencyList ${dependencyList} caliper )
+endif()
```
corbett5, line 24: I thought we wouldn't have to specify the common dependencies anymore.

src/coreComponents/events/CMakeLists.txt
```diff
@@ -0,0 +1,64 @@
...
+
+if ( ENABLE_CUDA )
+  set( dependencyList ${dependencyList} cuda )
+endif()
+
+if ( ENABLE_CALIPER )
+  set( dependencyList ${dependencyList} caliper adiak )
+endif()
+
+if ( ENABLE_MKL )
```
corbett5, line 51: Like why does events depend on MKL?

src/coreComponents/linearAlgebra/CMakeLists.txt
```diff
@@ -50,9 +50,20 @@ set( linearAlgebra_sources
      DofManager.cpp )
 
 if( BUILD_OBJ_LIBS )
-  set( dependencyList common blas lapack RAJA managers)
+  set( dependencyList mesh blas lapack RAJA)
```
corbett5, line 53: Yeah so why does functions depend on `MKL` but `linearAlgebra` only depends on `blas` and `lapack`?

src/coreComponents/unitTests/dataRepositoryTests/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
+#
```
corbett5, line 1: Is the goal to eventually move the unit tests back into the components? I think pretty much all of these shouldn't depend on anything outside of `dataRepository`.

corbett5 (approved):
I only looked at the CMake files, but I mean if it builds it's good right? @klevzoff had a lot of good comments. I thought that since we were using actual CMake targets now we wouldn't have to specify the dependency on `OpenMP` and `MPI` and `CUDA` everywhere. Did this not pan out?

src/cmake/GeosxOptions.cmake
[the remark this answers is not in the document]
rrsettgast: no

src/coreComponents/constitutive/CMakeLists.txt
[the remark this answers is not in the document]
rrsettgast: It shouldn't be....The goal is to just to have one. I wanted to get rid of this variance.

src/coreComponents/fileIO/Outputs/RestartOutput.cpp
```diff
@@ -18,11 +18,9 @@
...
 #include "RestartOutput.hpp"
 #include "fileIO/silo/SiloFile.hpp"
-#include "managers/DomainPartition.hpp"
-#include "managers/Functions/FunctionManager.hpp"
-#include "managers/ProblemManager.hpp"
-#include "managers/GeosxState.hpp"
-#include "managers/initialization.hpp"
+#include "mesh/DomainPartition.hpp"
+#include "functions/FunctionManager.hpp"
+#include "mainInterface/ProblemManager.hpp"
```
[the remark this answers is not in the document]
rrsettgast: nice catch

src/coreComponents/physicsSolvers/CMakeLists.txt
rrsettgast: made an issue for this.

src/coreComponents/mesh/NodeManager.hpp
```diff
@@ -19,7 +19,7 @@
 #ifndef GEOSX_MESH_NODEMANAGER_HPP_
 #define GEOSX_MESH_NODEMANAGER_HPP_
 
-#include "managers/ObjectManagerBase.hpp"
+#include "mesh/ObjectManagerBase.hpp"
 #include <string.h>
```
rrsettgast, line 23:
```suggestion
```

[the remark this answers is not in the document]
rrsettgast: Good question.

src/coreComponents/mainInterface/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+# Specify all sources
+#
+
+set(mainInterface_sources
+    ProblemManager.cpp
+    initialization.cpp
+    GeosxState.cpp
+   )
+
+set( dependencyList events fileIO optionparser RAJA conduit)
```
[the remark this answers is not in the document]
rrsettgast: I think the include list in `ProblemManager` is a bit overkill. I know that `fileIO` is dependent on `mesh`, so that dependency should be inherited. And `ProblemManager` should depend on `physicsSolvers`...but it seems to build without the dependency...probably because there is nothing linking with the `mainInterface` that isn't linking with `geosx_core`.

src/coreComponents/finiteVolume/CMakeLists.txt
```diff
@@ -35,9 +35,15 @@ set( finiteVolume_sources
    )
 
 if ( BUILD_OBJ_LIBS )
-  set( dependencyList dataRepository codingUtilities managers)
+  set( dependencyList mesh codingUtilities functions fieldSpecification )
```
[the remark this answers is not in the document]
rrsettgast: I don't think so

src/cmake/GeosxOptions.cmake
```diff
@@ -75,9 +75,9 @@ endif()
 
 ### BUILD & BLT SETUP ###
 
-option( BUILD_OBJ_LIBS "Builds coreComponent modules as object libraries" OFF)
+option( BUILD_OBJ_LIBS "Builds coreComponent modules as object libraries" ON )
```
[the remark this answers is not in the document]
rrsettgast: done

src/coreComponents/CMakeLists.txt
```diff
@@ -76,12 +80,16 @@ else()
 endif()
 
 if( BUILD_OBJ_LIBS )
+  message( "Using cmake  Object Libraries" )
   blt_add_library ( NAME geosx_core
                     SOURCES dummy.cpp
-                    DEPENDS_ON  ${coreLibs}
+                    DEPENDS_ON  mainInterface physicsSolvers
+#                    DEPENDS_ON  ${coreLibs}
```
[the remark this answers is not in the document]
rrsettgast: done

src/coreComponents/unitTests/dataRepositoryTests/CMakeLists.txt
rrsettgast: That is the goal....every unit test should depend only on the component that it tests.

src/coreComponents/events/CMakeLists.txt
rrsettgast: It shouldn't

src/coreComponents/linearAlgebra/CMakeLists.txt
rrsettgast: errr...excellent query Mary.

scripts/config-build.py
rrsettgast: Because it is easier to just pass the `--graphviz` command in at the command line when running `config-build`.

src/coreComponents/virtualElement/VirtualElementBase.hpp
rrsettgast: nope....linking issue.

klevzoff: But now that this constructor is not defined anywhere at all... there is no linking issue? I'm surprised. Anyway, it looks like `VirtualElement` objects aren't instantiated ever, so I would prefer to see a `= delete` default constructor rather than an undefined one.

src/coreComponents/CMakeLists.txt
```diff
@@ -26,6 +29,7 @@ set( geosx_core_defines "" )
 set( buildAsObj FALSE)
 if ( BUILD_OBJ_LIBS)
   set(buildAsObj TRUE)
+  set( LvArrayBuildAsObj TRUE CACHE BOOL "" FORCE)
```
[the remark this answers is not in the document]
rrsettgast: done

src/cmake/GeosxOptions.cmake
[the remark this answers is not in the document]
rrsettgast: well...maybe when object libraries aren't on...but ultimately we are always going to use object libraries.

src/coreComponents/virtualElement/VirtualElementBase.hpp
rrsettgast: Well...It would have helped if I added the source file. But the deletion is a better idea.

src/coreComponents/constitutive/CMakeLists.txt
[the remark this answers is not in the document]
rrsettgast: I got rid of the branching.

src/coreComponents/mesh/NodeManager.hpp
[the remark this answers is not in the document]
rrsettgast: removed.

src/coreComponents/virtualElement/VirtualElementBase.hpp
rrsettgast: deleted

scripts/config-build.py
rrsettgast: I actually added it back in with a unchangeable output name of `dependency.dot`, and then did the conversion to png automatically.

src/coreComponents/codingUtilities/CMakeLists.txt
rrsettgast: yes. removed.

rrsettgast:
> Looks good! Mostly cleanup suggestions.
> 
> Some additional thoughts/comments:
> 
> 1. Is the object library build fully functional? If so, would be great to have at least one CI build use it, or it will quickly become dysfunctional again.

`GEOSX_BUILD_OBJ_LIBS` is on by default. So all CI jobs should be using use object libraries.

> 2. I don't know how I feel about `FunctionManager`, `FieldSpecificationManager` and `GeometricObjectManager` going back to being "singleton"-like. This prevents two or more simulations existing at the same time (e.g. run in lockstep via python interface). The alternative is to access them like we do occasionally with `DomainPartition`, i.e. `getGroupByPath( "/Problem/functions" )`. Neither is particularly pretty, but singletons are a stronger code smell IMO.

Yes. This is an issue....but I thought this wasn't a problem for these objects? @corbett5 ?


> 3. There are more header guards that need updating, but I can't comment on specific files since they weren't modified... pretty much any `.hpp` file that was moved with no line changes should be checked. One example is `fieldSpecification/SourceFluxBoundaryCondition.hpp` but there are more. Perhaps this can be done in a dedicated cleanup PR because there are still some inconsistencies in header guard styles.
I went through and updated header guards. There were a lot that were just wrong regardless of the move. We should make a script to check the guards as part of the CI.

corbett5:
> 
> Yes. This is an issue....but I thought this wasn't a problem for these objects? @corbett5 ?

It's not a problem as long as you only have a single simulation running at a time. What @klevzoff is suggesting is that we could run multiple simulations at once on the same rank, just swapping between them as we see fit. This currently isn't supported, but the singleton nature of those classes would have to go if we did want to support it.

rrsettgast:
> > Yes. This is an issue....but I thought this wasn't a problem for these objects? @corbett5 ?
> 
> It's not a problem as long as you only have a single simulation running at a time. What @klevzoff is suggesting is that we could run multiple simulations at once on the same rank, just swapping between them as we see fit. This currently isn't supported, but the singleton nature of those classes would have to go if we did want to support it.

How about a static function that encapsulates the `getGroup` call? That would be similar to a singleton in usage, with the hidden smelliness hidden a bit. Something like:

```
FunctionManager & FunctionManager::getInstance()
 {
   return this->getGroupByPath< FunctionManager>("/Problem/FunctionManager");
 }
```

src/coreComponents/events/CMakeLists.txt
rrsettgast: removed

PR #1353, opened 2021-03-11
Kernels for poroelastic solvers

castelletto1:
This PR adds assembly kernels for fully-implicit poroelastic simulation with both single-phase (`PoroelasticSolver`) and multiphase (`MultiphasePoroelasticSolver`) flow. Capillary effects are neglected. Flow solvers can be based on either finite volumes (`FVM`) or hybrid schemes (`HybridFVM`). For the muliphase flow case the [maximum number of components](https://github.com/GEOSX/GEOSX/blob/3d3af6704162772e1f3f54131094fabcbd7bebb6/src/coreComponents/physicsSolvers/multiphysics/MultiphasePoroelasticKernel.hpp#L75) is equal to three). 

The kernels assemble:

- all contributions relative to the discrete linear momentum balance equation (note that an inexact linearization is used at the moment for the body force term)
- contributions to residual and Jacobian matrix relative to the the accumulation term in mass balance equations 

Currently the porosity equation considers a linear poroelastic material and assumes incompressible solid grains (`1/N = 0`, with `N` Biot's skeleton modulus):

with the subscript `n` denoting the discrete time level. The porosity treatment will be generalized introducing a dedicated constitutive model (see #1311).

A preliminary 2d example has been added (`PoroElastic_deadoil_3ph_baker_2d.xml`). 

Currently all integrated tests pass, except for the usual ones on Lassen.

Some issues to be addressed in future PR / PRs:

1. Exact linerization for body forces
1. Add capillary effects
1. Add MGR recipes for (i) multiphase poromechanics with `FVM`, and (ii) single- and multiphase poromechanics with `HybridFVM`
1. Rename `PoroelasticSolver` to `SinglePhasePoroelasticSolver` (not done here to avoid new baseline in the integrated tests)
1. Add benchmark based on SPE10 model 1 [see [Camargo et al., IJNAMG (2021)](https://onlinelibrary.wiley.com/doi/10.1002/nag.3192)]
1. Remove the split operator strategy (`SIM_FixedStress` coupling option) from the single-phase poroelastic solver.

Requires a rebaseline:
https://github.com/GEOSX/integratedTests/pull/124


src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoroelasticKernel.hpp
```diff
@@ -0,0 +1,447 @@
...
+
+  arrayView1d< real64 const > const m_bulkModulus;
+
+  real64 const m_biotCoefficient;
+
+
+};
+
+
+} // namespace SolidMechanicsLagrangianFEMKernels
```
sytuannguyen, line 441: /* namespace PoroelasticKernels */

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoroelasticKernel.hpp
```diff
@@ -0,0 +1,447 @@
...
+
+} // namespace geosx
+
+#include "finiteElement/kernelInterface/SparsityKernelBase.hpp"
```
sytuannguyen, line 445: Can this include be moved to the top of the file?

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoroelasticKernel.hpp
```diff
@@ -0,0 +1,447 @@
...
+                               + biotSkeletonModulusInverse * m_deltaFluidPressure[k];
+
+
+    // Evaluate body force vector
+    real64 bodyForce[3] = { m_gravityVector[0],
+                            m_gravityVector[1],
+                            m_gravityVector[2]};
+    if( m_gravityAcceleration > 0.0 )
+    {
+      real64 mixtureDensity = ( 1.0 - porosityNew ) * m_solidDensity( k, q ) + porosityNew * m_fluidDensity( k, q );
```
sytuannguyen, line 282: m_fluidDensity is not a position independant parameter?

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoroelasticKernel.hpp
```diff
@@ -0,0 +1,447 @@
...
+    m_solidDensity( inputConstitutiveType.getDensity() ),
+    m_fluidDensity( elementSubRegion.template getConstitutiveModel< constitutive::SingleFluidBase >( fluidModelNames[targetRegionIndex] ).density() ),
+    m_fluidDensityOld( elementSubRegion.template getReference< array1d< real64 > >( SinglePhaseBase::viewKeyStruct::densityOldString() ) ),
+    m_dFluidDensity_dPressure( elementSubRegion.template getConstitutiveModel< constitutive::SingleFluidBase >( fluidModelNames[targetRegionIndex] ).dDensity_dPressure() ),
+    m_flowDofNumber( elementSubRegion.template getReference< array1d< globalIndex > >( inputFlowDofKey )),
+    m_fluidPressure( elementSubRegion.template getReference< array1d< real64 > >( SinglePhaseBase::viewKeyStruct::pressureString() ) ),
+    m_deltaFluidPressure( elementSubRegion.template getReference< array1d< real64 > >( SinglePhaseBase::viewKeyStruct::deltaPressureString() ) ),
+    m_poroRef( elementSubRegion.template getReference< array1d< real64 > >( SinglePhaseBase::viewKeyStruct::referencePorosityString() ) ),
+    m_biotCoefficient( m_constitutiveUpdate.getBiotCoefficient() )
```
sytuannguyen, line 132: Isotropic Biot's tensor is considered here (single Biot's coefficient), so this class should be called IsotropicSinglePhase? Do we pretend to implement anisotropic Biot's tensor?

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoroelasticKernel.hpp
```diff
@@ -0,0 +1,447 @@
...
+    {
+      stack.localFlowDispJacobian[ 0][a * 3 + 0] += dPorosity_dVolStrainIncrement * m_fluidDensity( k, q ) * dNdX[a][0] * detJxW;
+      stack.localFlowDispJacobian[ 0][a * 3 + 1] += dPorosity_dVolStrainIncrement * m_fluidDensity( k, q ) * dNdX[a][1] * detJxW;
+      stack.localFlowDispJacobian[ 0][a * 3 + 2] += dPorosity_dVolStrainIncrement * m_fluidDensity( k, q ) * dNdX[a][2] * detJxW;
+    }
+
+    stack.localFlowResidual[0] += ( porosityNew * m_fluidDensity( k, q ) - porosityOld * m_fluidDensityOld( k ) ) * detJxW;
+    stack.localFlowFlowJacobian[0][0] += ( dPorosity_dPressure * m_fluidDensity( k, q ) + porosityNew * m_dFluidDensity_dPressure( k, q ) ) * detJxW;
+
+  }
```
sytuannguyen, line 338: Can we extend these residual and jacobian calculations to THM coupling? Are there any difficulty to do so?

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoroelasticKernel.hpp
```diff
@@ -0,0 +1,447 @@
...
+        continue;
+      m_matrix.template addToRowBinarySearchUnsorted< parallelDeviceAtomic >( dof,
+                                                                              stack.localRowDofIndex,
+                                                                              stack.localFlowDispJacobian[i],
+                                                                              nUDof );
+      m_matrix.template addToRowBinarySearchUnsorted< parallelDeviceAtomic >( dof,
+                                                                              stack.localFlowDofIndex,
+                                                                              stack.localFlowFlowJacobian[i],
+                                                                              nPDof );
+      RAJA::atomicAdd< parallelDeviceAtomic >( &m_rhs[ dof ], stack.localFlowResidual[i] );
```
sytuannguyen, line 391: parallelDeviceAtomic is always employed for this task or this is just an option?

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoroelasticKernel.hpp
```diff
@@ -0,0 +1,447 @@
...
+
+  /// The rank-global delta-fluid pressure array.
+  arrayView1d< real64 const > const m_deltaFluidPressure;
+
+  /// The rank-global reference porosity array
+  arrayView1d< real64 const > const m_poroRef;
+
+  arrayView1d< real64 const > const m_bulkModulus;
+
+  real64 const m_biotCoefficient;
```
sytuannguyen, line 435: Comments

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoroelasticKernel.hpp
```diff
@@ -0,0 +1,447 @@
...
+          rankOffset,
+          inputMatrix,
+          inputRhs ),
+    m_X( nodeManager.referencePosition()),
+    m_disp( nodeManager.totalDisplacement()),
+    m_uhat( nodeManager.incrementalDisplacement()),
+    m_gravityVector{ inputGravityVector[0], inputGravityVector[1], inputGravityVector[2] },
+    m_gravityAcceleration( sqrt( inputGravityVector[0] * inputGravityVector[0] +
+                                 inputGravityVector[1] * inputGravityVector[1] +
+                                 inputGravityVector[2] * inputGravityVector[2] ) ),
```
sytuannguyen, line 123: Lvarray norm?

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoroelasticKernel.hpp
```diff
@@ -0,0 +1,447 @@
...
+      for( int i=0; i<3; ++i )
+      {
+#if defined(CALC_FEM_SHAPE_IN_KERNEL)
+        stack.xLocal[ a ][ i ] = m_X[ localNodeIndex ][ i ];
+#endif
+        stack.u_local[ a ][i] = m_disp[ localNodeIndex ][i];
+        stack.uhat_local[ a ][i] = m_uhat[ localNodeIndex ][i];
+        stack.localRowDofIndex[a*3+i] = m_dofNumber[localNodeIndex]+i;
+        stack.localColDofIndex[a*3+i] = m_dofNumber[localNodeIndex]+i;
+      }
```
sytuannguyen, line 214: format [ a ][i] etc.

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoroelasticKernel.hpp
```diff
@@ -0,0 +1,447 @@
...
+    {
+      localIndex const localNodeIndex = m_elemsToNodes( k, a );
+
+      for( int i=0; i<3; ++i )
+      {
+#if defined(CALC_FEM_SHAPE_IN_KERNEL)
+        stack.xLocal[ a ][ i ] = m_X[ localNodeIndex ][ i ];
+#endif
+        stack.u_local[ a ][i] = m_disp[ localNodeIndex ][i];
+        stack.uhat_local[ a ][i] = m_uhat[ localNodeIndex ][i];
```
sytuannguyen, line 211: dispIncrement or displacementIncrement is more readable than uhat

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoroelasticKernel.hpp
```diff
@@ -0,0 +1,447 @@
...
+                              localIndex const q,
+                              StackVariables & stack ) const
+  {
+    // Get displacement: (i) basis functions (N), (ii) basis function
+    // derivatives (dNdX), and (iii) determinant of the Jacobian transformation
+    // matrix times the quadrature weight (detJxW)
+    real64 N[numNodesPerElem];
+    real64 dNdX[numNodesPerElem][3];
+    FE_TYPE::calcN( q, N );
+    real64 const detJxW = m_finiteElementSpace.template getGradN< FE_TYPE >( k, q, stack.xLocal, dNdX );
```
sytuannguyen, line 236: This function is called getGradN but it return also detJxW, ce we optimize this?

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoroelasticKernel.hpp
```diff
@@ -0,0 +1,447 @@
...
+    real64 const dPorosity_dVolStrainIncrement =  m_biotCoefficient;                                                   //
+                                                                                                                       //
+    GEOSX_ERROR_IF_GT_MSG( fabs( m_biotCoefficient - 1.0 ),                                                            //
+                           1e-10,                                                                                      //
+                           "Correct only for Biot's coefficient equal to 1" );                                         //
+    // --------------------------------------------------------------------------------------------------------------- //
+    real64 const porosityNew = porosityOld
+                               + m_biotCoefficient * (strainIncrement[0] + strainIncrement[1] + strainIncrement[2] )
+                               + biotSkeletonModulusInverse * m_deltaFluidPressure[k];
+
```
sytuannguyen, line 274: Do we plan to update Biot's coefficient (and Biot modulus) with porosity also, especially when we will working with plasticity

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoroelasticKernel.hpp
```diff
@@ -0,0 +1,447 @@
...
+      stack.localDispFlowJacobian[ a * 3 + 0][0] += dNdX[a][0] * m_biotCoefficient * detJxW;
+      stack.localDispFlowJacobian[ a * 3 + 1][0] += dNdX[a][1] * m_biotCoefficient * detJxW;
+      stack.localDispFlowJacobian[ a * 3 + 2][0] += dNdX[a][2] * m_biotCoefficient * detJxW;
+    }
+
+    if( m_gravityAcceleration > 0.0 )
+    {
+      // Assumptions: ( i) dMixtureDens_dVolStrain contribution is neglected
+      //              (ii) grains are assumed incompressible
+
```
sytuannguyen, line 316: These assumptions are just temporal?

sytuannguyen:
@castelletto1 @rrsettgast I am interested in poroelastic coupling, so read through your PR and add some comments, Hope this help.

Do you want to add a class for computing the poroelastic constant, i.e. the transformation between the poroelastic constants?

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -640,10 +640,13 @@ void CompositionalMultiphaseBase::assembleSystem( real64 const GEOSX_UNUSED_PARA
 {
   GEOSX_MARK_FUNCTION;
 
-  assembleAccumulationTerms( domain,
-                             dofManager,
-                             localMatrix,
-                             localRhs );
+  if( m_poroElasticFlag != PoroElasticCouplingTypeOption::FIM )
+  {
```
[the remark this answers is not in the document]
castelletto1: Yes, done

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoroelasticSolver.cpp
```diff
@@ -0,0 +1,302 @@
...
+                                                                      //
+                                                                      localMatrix,
+                                                                      localRhs );
+
+  // Face-based contributions (including pressure-dependent terms in the accumulation term of the mass balance equation)
+  m_flowSolver->assembleSystem( time_n, dt,
+                                domain,
+                                dofManager,
+                                localMatrix,
+                                localRhs );
```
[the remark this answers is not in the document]
castelletto1: We have to keep `m_poroElasticFlag` until we decide to get rid of the `SIM_FixedStress` coupling option: 

https://github.com/GEOSX/GEOSX/blob/80ab48fb27b53db20977903ea4743d31dcd0d501/src/coreComponents/physicsSolvers/multiphysics/PoroelasticSolver.hpp#L126-L130

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoroelasticKernel.hpp
castelletto1: The include location mimics what done in SolidMechanicsSmallStrainQuasiStaticKernel.hpp (see [here](https://github.com/GEOSX/GEOSX/blob/80ab48fb27b53db20977903ea4743d31dcd0d501/src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsSmallStrainQuasiStaticKernel.hpp#L334)).

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoroelasticSolver.cpp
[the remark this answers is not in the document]
CusiniM: I have basically killed all the "poroelastic" kernels in the `SinglePhaseFlowKernels` in #1311 to make the flow solver basically only know about porosity and permeability and not about mechanics variables. I think I can remove the `m_poroElasticFlag` flag in that PR or you can go ahead and do it and just turn off the sequentially coupled integrated tests for now.

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoroelasticKernel.hpp
castelletto1: Appropriate Biot's coefficient, porosity, etc. constitutive models will be added in future PRs.

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoroelasticKernel.hpp
```diff
@@ -0,0 +1,554 @@
...
+    {
+      localIndex const localNodeIndex = m_elemsToNodes( k, a );
+
+      for( int i=0; i<3; ++i )
+      {
+#if defined(CALC_FEM_SHAPE_IN_KERNEL)
+        stack.xLocal[ a ][ i ] = m_X[ localNodeIndex ][ i ];
+#endif
+        stack.u_local[ a ][i] = m_disp[ localNodeIndex ][i];
+        stack.uhat_local[ a ][i] = m_uhat[ localNodeIndex ][i];
```
CusiniM, line 238: I wonder if we should use `disp` and `deltaDisp` like we do for the pressure instead of `u` and `u_hat`.

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoroelasticKernel.hpp
castelletto1: Done

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoroelasticSolver.cpp
```diff
@@ -0,0 +1,308 @@
...
+}
+
+void MultiphasePoroelasticSolver::postProcessInput()
+{
+  SolverBase::postProcessInput();
+
+  m_flowSolver = &this->getParent().getGroup< CompositionalMultiphaseBase >( m_flowSolverName );
+  m_solidSolver = &this->getParent().getGroup< SolidMechanicsLagrangianFEM >( m_solidSolverName );
+
+  m_solidSolver->setEffectiveStress( 1 );
```
CusiniM, line 139: if we keep a flag in both the mechanics and the flow solvers ( it seems to me that we can move some stuff to the coupled solver and remove those flags) I would make sure that the flags are the same (`m_poroElastic` or whatever we prefer). Obviously, if we decide to remove the flags and call the coupled kernels directly in the poroelastic solver then it makes no sense changing them.

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoroelasticSolver.cpp
```diff
@@ -0,0 +1,308 @@
...
+
+void MultiphasePoroelasticSolver::applySystemSolution( DofManager const & dofManager,
+                                                       arrayView1d< real64 const > const & localSolution,
+                                                       real64 const scalingFactor,
+                                                       DomainPartition & domain )
+{
+  // update displacement field
+  m_solidSolver->applySystemSolution( dofManager, localSolution, scalingFactor, domain );
+  // update pressure field
+  m_flowSolver->applySystemSolution( dofManager, localSolution, -scalingFactor, domain );
```
CusiniM, line 303: at some point we should fix the signs.

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoroelasticKernel.hpp
castelletto1: Done.

CusiniM: you could use `LvArray::tensorOps::l2Norm<3>(inputGravityVector);`

CusiniM (approved):
This is great! Should also integrate very well with what I have been working on.

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoroelasticKernel.hpp
castelletto1: Done.

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoroelasticKernel.hpp
castelletto1: I agree. The naming convention should homogenized across all solid mechanics solvers / kernels.

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoroelasticKernel.hpp
castelletto1: What do you mean by optimizing?

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoroelasticKernel.hpp
castelletto1: Yes, these are temporary assumptions. However, we should preserve the flexibility of lagging in time some derivatives in the linearization. Note that as soon as `dMixtureDens_dVolStrain` effects are accounted for the "displacement-displacement" block in the Jacobian matrix will result in a nonsymmetric block. As to grain compressibility effects, we should add a bulk modulus member -- or better a bulk compressibility for easily accounting for the incompressible case -- in the base class for all solid constitutive models.

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoroelasticKernel.hpp
castelletto1: The solver uses values cell values or quadrature-point values depending on where the quantities of interest are defined without making any choice. Knowing that the flow solvers is based on a cell-centered method (possibly hybridized) allows for less storage in the constitutive models.

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoroelasticKernel.hpp
castelletto1: I don't anticipate any difficulty. The design question I have is whether we want to have a unique single phase poroelastic solver (and its multiphase counterpart) that has all coupling we're interested in or independent solvers. Based on the last discussion we had, I understand we prefer the second alternative.

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoroelasticKernel.hpp
castelletto1: I agree that the naming convention should be improved. However, it should reflected across all solid mechanics kernels.

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoroelasticSolver.cpp
castelletto1: The only reason for keeping the flag `m_poroelasticFlag` is the split operator approach based on the fixed stress algorithm for the single phase coupled solver. If we agree on removing it (and disabling the corresponding integrated test), we could remove even now.

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoroelasticKernel.hpp
sytuannguyen: I mean: 
- Option 1: the function name should reflect the fact that it is a multi task function that modifies gradN and return detJxW
- Option 2: this multi task function should be splitted to single purpose functions

Persionally I prefer single purpose functions

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoroelasticKernel.hpp
sytuannguyen: The one in SolidMechanics... seem to be the only one include that is placed at the bottom of a file. Any special reason for this?

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoroelasticKernel.hpp
```diff
@@ -0,0 +1,444 @@
...
+ *                            @p SUBREGION_TYPE.
+ * @tparam UNUSED An unused parameter since we are assuming that the test and
+ *                trial space have the same number of support points.
+ *
+ * ### QuasiStatic Description
+ * Implements the KernelBase interface functions required for solving the
+ * quasi-static equilibrium equations using one of the
+ * "finite element kernel application" functions such as
+ * geosx::finiteElement::RegionBasedKernelApplication.
+ *
```
sytuannguyen, line 44: idem to solid elastic solvers?

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoroelasticKernel.hpp
```diff
@@ -0,0 +1,552 @@
...
+ * Copyright (c) 2019-     GEOSX Contributors
+ * All rights reserved
+ *
+ * See top level LICENSE, COPYRIGHT, CONTRIBUTORS, NOTICE, and ACKNOWLEDGEMENTS files for details.
+ * ------------------------------------------------------------------------------------------------------------
+ */
+
+/**
+ * @file SinglePhasePoroelasticKernel.hpp
+ */
```
sytuannguyen, line 17: to verify through the copy paste ^^

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoroelasticKernel.hpp
```diff
@@ -0,0 +1,552 @@
...
+ * In this implementation, the template parameter @p NUM_NODES_PER_ELEM is used
+ * in place of both @p NUM_TEST_SUPPORT_POINTS_PER_ELEM and
+ * @p NUM_TRIAL_SUPPORT_POINTS_PER_ELEM, which are assumed to be equal. This
+ * results in the @p UNUSED template parameter as only the NUM_NODES_PER_ELEM
+ * is passed to the ImplicitKernelBase template to form the base class.
+ *
+ * Additionally, the number of degrees of freedom per support point for both
+ * the test and trial spaces are specified as `3` when specifying the base
+ * class.
+ */
```
sytuannguyen, line 53: idem solid elastic solver?

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoroelasticKernel.hpp
```diff
@@ -0,0 +1,552 @@
...
+template< typename SUBREGION_TYPE,
+          typename CONSTITUTIVE_TYPE,
+          typename FE_TYPE >
+class Multiphase :
```
sytuannguyen, line 57: Should change to MultiPhase to make it consistent with SinglePhase

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoroelasticKernel.hpp
```diff
@@ -0,0 +1,552 @@
...
+  public finiteElement::ImplicitKernelBase< SUBREGION_TYPE,
+                                            CONSTITUTIVE_TYPE,
```
sytuannguyen, line 59: This class share a lot of code with SinglePhase class, can it bases directly on that class?

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoroelasticKernel.hpp
```diff
@@ -0,0 +1,552 @@
...
+    m_dFluidPhaseSaturation_dPressure( elementSubRegion.template getReference< array2d< real64 > >( CompositionalMultiphaseBase::viewKeyStruct::dPhaseVolumeFraction_dPressureString() )),
+    m_dFluidPhaseSaturation_dGlobalCompDensity( elementSubRegion.template getReference< array3d< real64 > >(
+                                                  CompositionalMultiphaseBase::viewKeyStruct:: dPhaseVolumeFraction_dGlobalCompDensityString() )),
+    //
+    m_dGlobalCompFraction_dGlobalCompDensity( elementSubRegion.template getReference< array3d< real64 > >( CompositionalMultiphaseBase::viewKeyStruct::dGlobalCompFraction_dGlobalCompDensityString() )),
+//    //
+    m_dFluidPhaseCompFraction_dGlobalCompFraction( elementSubRegion.template getConstitutiveModel< constitutive::MultiFluidBase >(
+                                                     fluidModelNames[targetRegionIndex] ).dPhaseCompFraction_dGlobalCompFraction() ),
+    //
+    m_flowDofNumber( elementSubRegion.template getReference< array1d< globalIndex > >( inputFlowDofKey )),
```
sytuannguyen, line 146: clean these // // ...

sytuannguyen:
@castelletto1 I added some additional comments to the Multiphase class. Also I have a naive question: if multi phase solver can do the job of single phase solver just by setting 100% saturation to a phase, do we really need single phase solver?

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoroelasticKernel.hpp
castelletto1: yes

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoroelasticKernel.hpp
CusiniM: all these computations are being moved to the proper constitutive models. I would not modify anything else at this stage.

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoroelasticKernel.hpp
castelletto1: I believe the standard nomenclature is:

- single-phase flow (see [here](https://www.glossary.oilfield.slb.com/en/terms/s/single-phase_flow))
- two-phase flow (see [here](https://www.glossary.oilfield.slb.com/en/terms/t/two-phase_flow))
- ...
- multiphase flow (see [here](https://www.glossary.oilfield.slb.com/en/terms/m/multiphase))

Everything is consistent with the logic adopted for fluid flow solver.

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoroelasticKernel.hpp
castelletto1: See my comment above about constitutive models for porosity, Biot's coefficient, ...

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoroelasticKernel.hpp
castelletto1: I briefly modified the description, but more documentation for these kernels will be added. Note that the pressure interpolation (along with the global component densities for the multiphase case) is implicitly assumed piecewise constant since our flow solvers are either based on `FMV` or `HybridFVM`. A formal description of the functional space used to approximate all primary variable is needed.

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoroelasticKernel.hpp
castelletto1: Done

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoroelasticKernel.hpp
castelletto1: See comment above relative to `SinglePhasePoroelasticKernel`.

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoroelasticSolver.hpp
```diff
@@ -0,0 +1,149 @@
...
+
+  // pointer to the flow sub-solver
+  CompositionalMultiphaseBase * m_flowSolver;
+
+  // pointer to the solid mechanics sub-solver
+  SolidMechanicsLagrangianFEM * m_solidSolver;
+
+};
+
+ENUM_STRINGS( MultiphasePoroelasticSolver::CouplingTypeOption, "FIM", "SIM_FixedStress" )
```
klevzoff, line 145: This has an extra string that does not correspond to anything in enum....
Unfortunately there's no mechanism to enforce this list being consistent with enum definition.

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoroelasticKernel.hpp
castelletto1: Done

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoroelasticSolver.cpp
```diff
@@ -0,0 +1,303 @@
...
+ * Copyright (c) 2018-2020 Total, S.A
+ * Copyright (c) 2019-     GEOSX Contributors
+ * All rights reserved
+ *
+ * See top level LICENSE, COPYRIGHT, CONTRIBUTORS, NOTICE, and ACKNOWLEDGEMENTS files for details.
+ * ------------------------------------------------------------------------------------------------------------
+ */
+
+/**
+ * @file PoroelasticSolver.cpp
```
klevzoff, line 16:
```suggestion
 * @file MultiphasePoroelasticSolver.cpp
```

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoroelasticKernel.hpp
```diff
@@ -0,0 +1,537 @@
...
+  using Base = finiteElement::ImplicitKernelBase< SUBREGION_TYPE,
+                                                  CONSTITUTIVE_TYPE,
+                                                  FE_TYPE,
+                                                  3,
+                                                  3 >;
+
+  /// Number of nodes per element...which is equal to the
+  /// numTestSupportPointPerElem and numTrialSupportPointPerElem by definition.
+  static constexpr int numNodesPerElem = Base::numTestSupportPointsPerElem;
+  static constexpr int numMaxComponentsMultiphasePoroelastic = 3;
```
klevzoff, line 66: We need to address this in a follow-up PR

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoroelasticKernel.hpp
castelletto1: Computing `dNdX` requires the Jacobian matrix of the isoparametric mapping. It is convenient to return its determinant, which is needed in the numerical integration. Single purpose functions are nice but we would end up recomputing over and over again the same quantities. We could change the function to something like:
```
getGradN< FE_TYPE >( k, q, stack.xlocal, dNdX, detJxW );
```
Personally I think the return value (`detJxW`) makes sense. But I'm open to suggestions.

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoroelasticKernel.hpp
```diff
@@ -0,0 +1,537 @@
...
+      if( dof < 0 || dof >= m_matrix.numRows() )
+        continue;
+      m_matrix.template addToRowBinarySearchUnsorted< parallelDeviceAtomic >( dof,
+                                                                              stack.localRowDofIndex,
+                                                                              stack.localFlowDispJacobian[i],
+                                                                              nUDof );
+      m_matrix.template addToRowBinarySearchUnsorted< parallelDeviceAtomic >( dof,
+                                                                              stack.localFlowDofIndex,
+                                                                              stack.localFlowFlowJacobian[i],
+                                                                              m_numComponents + 1 );
```
klevzoff, line 458: Use `addToRow< serialAtomic >` instead:
* contributions never collide (1 thread/cell)
* contributions are sorted by column index already
(correct me if any of this is wrong though)

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoroelasticKernel.hpp
```diff
@@ -0,0 +1,435 @@
...
+
+        m_matrix.template addToRowBinarySearchUnsorted< parallelDeviceAtomic >( dof,
+                                                                                stack.localFlowDofIndex,
+                                                                                stack.localDispFlowJacobian[numDofPerTestSupportPoint * localNode + dim],
+                                                                                nPDof );
+
+      }
+    }
+
+    for( localIndex i = 0; i < nPDof; ++i )
```
klevzoff, line 367: I expect the optimizer to remove the loop, but do we want it in the source if `nPDof` is always 1?

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoroelasticKernel.hpp
```diff
@@ -0,0 +1,435 @@
...
+      if( dof < 0 || dof >= m_matrix.numRows() )
+        continue;
+      m_matrix.template addToRowBinarySearchUnsorted< parallelDeviceAtomic >( dof,
+                                                                              stack.localRowDofIndex,
+                                                                              stack.localFlowDispJacobian[i],
+                                                                              nUDof );
+      m_matrix.template addToRowBinarySearchUnsorted< parallelDeviceAtomic >( dof,
+                                                                              stack.localFlowDofIndex,
+                                                                              stack.localFlowFlowJacobian[i],
+                                                                              nPDof );
```
klevzoff, line 379: Same comment re: atomics

klevzoff (approved):
Looks good!

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoroelasticSolver.hpp
castelletto1: Done

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoroelasticSolver.cpp
castelletto1: Done

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoroelasticKernel.hpp
klevzoff: To be clear, this only applies to cell-centered (flow) equation rows of course

castelletto1: Yes, that was clear. Your suggestion is definitely correct. The original motivation for the current implementation is that I wanted to have a coupled solver that could also handle a pure finite-element formulation, i.e. using nodal interpolation also for the pressure field.

castelletto1: The previous comment applies exclusively to single-phase poromechanics. I assume that for multiphase flow we will only support cell-centered (possibly hybridized) discretizations. @rrsettgast, @joshua-white would you agree?

klevzoff: Actually, now that I've read it again, `localRowDofIndex` and `localFlowDispJacobian` are not necessarily sorted? So the first one would be `addToRowBinarySearchUnsorted< serialAtomic >`. Sorry for the confusion.

CusiniM: Yeah, IMO I am not even sure we will have a purely FEM flow solver any time soon. If we wanted to do it for multiphase we would also have to use a mix formulation or another way to enforce mass conservation so I don't know if it's worth worrying about it for now.

castelletto1: Done

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoroelasticKernel.hpp
castelletto1: Done

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoroelasticKernel.hpp
castelletto1: Removed the loop.

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoroelasticSolver.cpp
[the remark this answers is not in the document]
castelletto1: Let's tackle this in #1311.

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoroelasticKernel.hpp
CusiniM: I think this would hide the fact that `detJxW ` gets modified inside the function so I kind of prefere the way it currently is.

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.cpp
```diff
@@ -426,6 +426,7 @@ void SinglePhaseBase::assembleSystem( real64 const time_n,
 
   if( m_poroElasticFlag )
```
rrsettgast, line 427: still can't get rid of these?

rrsettgast (approved):

sytuannguyen (approved):

src/coreComponents/linearAlgebra/interfaces/hypre/HyprePreconditioner.cpp
```diff
@@ -402,7 +402,8 @@ void HyprePreconditioner::createMGR( DofManager const * const dofManager )
   std::vector< HYPRE_Int > mgr_level_interp_type;
   std::vector< HYPRE_Int > mgr_level_frelax_method;
 
-  if( ( m_parameters.mgr.strategy == "Poroelastic" ) | ( m_parameters.mgr.strategy == "Hydrofracture" ) )
+  if( ( m_parameters.mgr.strategy == LinearSolverParameters::MGR::StrategyType::singlePhasePoroelastic) |
+      ( m_parameters.mgr.strategy == LinearSolverParameters::MGR::StrategyType::hydrofracture ) )
```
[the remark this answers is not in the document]
CusiniM: I think that this change is making one test fail (fail check ) on Lassen. It seems to take more iterations on the gpus.

PR #1354, opened 2021-03-12
Parallel synchronization of new embedded surfaces. 

CusiniM:
- PTP Submodule update: https://github.com/GEOSX/GEOSX_PTP/pull/25
- IntegratedTests rebaseline: https://github.com/GEOSX/integratedTests/pull/131


src/coreComponents/dataRepository/BufferOps_inline.hpp
```diff
@@ -784,6 +784,60 @@ localIndex Unpack( buffer_unit_type const * & buffer,
...
+    sizeOfPackedChars += Pack< DO_PACKING >( buffer, localToGlobal[temp[a]] );
+  }
+
+  return sizeOfPackedChars;
+}
+
+//template< typename SORTED >
+//localIndex Unpack( buffer_unit_type * & buffer,
+//                   SortedArrayView< localIndex const > const & var,
+//                   arrayView1d< localIndex const > const & packList,
```
CusiniM, line 820: I wondered if I needed to filter based on the packList but I don't see why it would be necessary when unpacking.

src/coreComponents/dataRepository/BufferOps_inline.hpp
```diff
@@ -784,6 +784,60 @@ localIndex Unpack( buffer_unit_type const * & buffer,
...
 
+template< bool DO_PACKING >
+localIndex Pack( buffer_unit_type * & buffer,
+                 SortedArrayView< localIndex const > const & var,
+                 arrayView1d< localIndex const > const & packList,
+                 arraySlice1d< globalIndex const > const & localToGlobal )
+{
+
+  localIndex length = 0;
+  array1d< localIndex > temp( var.size() );
```
corbett5, line 795: I'd get rid of `temp` and just do
```c++
for( auto a : packList )
{ length += var.count( a ); }

for( localIndex a=0; a<length; ++a )
  {
    if( var.count( packList[ a ] ) )
    {
      sizeOfPackedChars += Pack< DO_PACKING >( buffer, localToGlobal[packList[a]] );
    }
  }
```

corbett5: Just curious though, why are we only packing things that are both in `var` and `packList`?

src/coreComponents/dataRepository/BufferOps_inline.hpp
corbett5: That sounds right to me, you're just unpacking global indices and you don't care where they came from.

corbett5: But if that's the case remove the commented out code.

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp
```diff
@@ -201,12 +201,13 @@ VTKPolyDataWriterInterface::getSurface( FaceElementSubRegion const & esr,
 }
 std::pair< vtkSmartPointer< vtkPoints >, vtkSmartPointer< vtkCellArray > >
 VTKPolyDataWriterInterface::getEmbeddedSurface( EmbeddedSurfaceSubRegion const & esr,
-                                                NodeManager const & nodeManager ) const
+                                                EmbeddedSurfaceNodeManager const & nodeManager ) const
```
corbett5, line 204: So did this add a new `EmbeddedSurfaceNodeManager`? I thought you were trying to unify this with the `NodeManager`?

src/coreComponents/mesh/ElementRegionManager.cpp
```diff
@@ -542,5 +541,105 @@ ElementRegionManager::UnpackUpDownMaps( buffer_unit_type const * & buffer,
...
+
+    localIndex numSubRegionsRead;
+    unpackedSize += bufferOps::Unpack( buffer, numSubRegionsRead );
+    elemRegion.forElementSubRegionsIndex< CellElementSubRegion >(
+      [&]( localIndex const kSubReg, CellElementSubRegion & subRegion )
+    {
+      string subRegionName;
+      unpackedSize += bufferOps::Unpack( buffer, subRegionName );
+
+      /// THIS IS WRONG
```
corbett5, line 634: What should it be / why is it in here if it's WRONG?

src/coreComponents/mesh/EmbeddedSurfaceNodeManager.cpp
```diff
@@ -0,0 +1,416 @@
...
+{
+  m_toEdgesRelation.resize( newSize, 2 * getEdgeMapOverallocation() );
+  m_toElements.m_toElementRegion.resize( newSize, 2 * getElemMapOverAllocation() );
+  m_toElements.m_toElementSubRegion.resize( newSize, 2 * getElemMapOverAllocation() );
+  m_toElements.m_toElementIndex.resize( newSize, 2 * getElemMapOverAllocation() );
+  ObjectManagerBase::resize( newSize );
+}
+
+
+void EmbeddedSurfaceNodeManager::setEdgeMaps( EdgeManager const & embSurfEdgeManager )
```
corbett5, line 61: This is probably the exact same stuff from `NodeManager` right?

src/coreComponents/mesh/SurfaceElementSubRegion.hpp
```diff
@@ -215,9 +215,12 @@ class SurfaceElementSubRegion : public ElementSubRegionBase
     constexpr static char const * creationMassString() { return "creationMass"; }
   };
 
-  /// Map between the face elements and the cells
+  /// Map between the surface elements and the cells
   FixedToManyElementRelation m_surfaceElementsToCells;
 
+  /// Unmapped surface elements to nodes map
+  map< localIndex, array1d< globalIndex > > m_unmappedGlobalIndicesInToNodes;
```
corbett5, line 222: Unless you need to ordering this should be an `unordered_map`.

corbett5 (approved):
Looks good! @wrtobin if you have a second give the packing stuff a look over.

src/coreComponents/mesh/ElementRegionManager.cpp
CusiniM: That comment `THIS IS WRONG??` is there in all the unpacking routines of the `ElementRegionManager` and I copy pasted that line. I am not sure why it would be wrong. I would actually say that they are all correct but I thought of leaving the comment in case someone decided to modify it later on.

src/coreComponents/dataRepository/BufferOps_inline.hpp
wrtobin: Yeah I'm curious about this too, it seems like we also have packing/unpacking versions currently in main that use this logic as well.

CusiniM: so, in this specific case, I think I am using this to pack a set. This set (`var`) contains the indices of the fractured elements. I think that the `packList` contains all ghost indices and I only want to pack the ghosts that are fracture and those contained in the set (`var`).

CusiniM: It is used here https://github.com/GEOSX/GEOSX/blob/404cf125360352eab85b22f68823166e3f298800/src/coreComponents/mesh/CellElementSubRegion.cpp#L212

corbett5: Cool, thanks! Just seemed a little strange but that makes perfect sense!

src/coreComponents/mesh/EmbeddedSurfaceNodeManager.cpp
CusiniM: @TotoGaz will refactor most of the mesh and move all these functions that create the maps so I will leave it for now.

src/coreComponents/mesh/SurfaceElementSubRegion.hpp
CusiniM: I have only moved location of this map but I it was originally created by @rrsettgast who knows better than me how it's used. Does it need to be ordered? 
I should have another one in there though which I am pretty sure that can be made unordered.

src/coreComponents/mesh/CellElementSubRegion.cpp
```diff
@@ -171,6 +173,87 @@ localIndex CellElementSubRegion::unpackUpDownMaps( buffer_unit_type const * & bu
   return unPackedSize;
 }
 
+localIndex CellElementSubRegion::packFracturedElementsSize( arrayView1d< localIndex const > const & packList,
+                                                            arrayView1d< globalIndex const > embeddedSurfacesLocalToGlobal ) const
```
rrsettgast, line 177: why is one of these a `const &` and one a non-const view? I think views should always be const?

src/coreComponents/mesh/CellElementSubRegion.cpp
```diff
@@ -171,6 +173,87 @@ localIndex CellElementSubRegion::unpackUpDownMaps( buffer_unit_type const * & bu
...
+{
+  buffer_unit_type * junk = nullptr;
+  return packFracturedElementsPrivate< false >( junk, packList, embeddedSurfacesLocalToGlobal );
+}
+
+
+localIndex CellElementSubRegion::packFracturedElements( buffer_unit_type * & buffer,
+                                                        arrayView1d< localIndex const > const & packList,
+                                                        arrayView1d< globalIndex const > embeddedSurfacesLocalToGlobal ) const
```
rrsettgast, line 186: here too

src/coreComponents/mesh/CellElementSubRegion.cpp
```diff
@@ -171,6 +173,87 @@ localIndex CellElementSubRegion::unpackUpDownMaps( buffer_unit_type const * & bu
...
+{
+  return packFracturedElementsPrivate< true >( buffer, packList, embeddedSurfacesLocalToGlobal );
+}
+
+template< bool DOPACK >
+localIndex CellElementSubRegion::packFracturedElementsPrivate( buffer_unit_type * & buffer,
+                                                               arrayView1d< localIndex const > const & packList,
+                                                               arrayView1d< globalIndex const > embeddedSurfacesLocalToGlobal ) const
```
rrsettgast, line 194: here too

src/coreComponents/mesh/CellElementSubRegion.cpp
```diff
@@ -171,6 +173,87 @@ localIndex CellElementSubRegion::unpackUpDownMaps( buffer_unit_type const * & bu
...
+                                           packList,
+                                           localToGlobal );
+
+  return packedSize;
+}
+
+
+localIndex CellElementSubRegion::unpackFracturedElements( buffer_unit_type const * & buffer,
+                                                          localIndex_array & packList,
+                                                          unordered_map< globalIndex, localIndex > embeddedSurfacesGlobalToLocal )
```
rrsettgast, line 223:
```suggestion
                                                          unordered_map< globalIndex, localIndex > & embeddedSurfacesGlobalToLocal )
```
maybe const?

src/coreComponents/mesh/CellElementSubRegion.hpp
```diff
@@ -91,6 +91,17 @@ class CellElementSubRegion : public CellBlock
...
+  localIndex packFracturedElementsSize( arrayView1d< localIndex const > const & packList,
+                                        arrayView1d< globalIndex const > embeddedSurfacesLocalToGlobal ) const;
+
+  localIndex packFracturedElements( buffer_unit_type * & buffer,
+                                    arrayView1d< localIndex const > const & packList,
+                                    arrayView1d< globalIndex const > embeddedSurfacesLocalToGlobal ) const;
+
+  localIndex unpackFracturedElements( buffer_unit_type const * & buffer,
+                                      localIndex_array & packList,
+                                      unordered_map< globalIndex, localIndex > embeddedSurfacesGlobalToLocal );
```
rrsettgast, line 103: `&` or `const &`

src/coreComponents/mesh/CellElementSubRegion.hpp
```diff
@@ -253,6 +267,11 @@ class CellElementSubRegion : public CellBlock
   localIndex packUpDownMapsPrivate( buffer_unit_type * & buffer,
                                     arrayView1d< localIndex const > const & packList ) const;
 
+  template< bool DOPACK >
+  localIndex packFracturedElementsPrivate( buffer_unit_type * & buffer,
+                                           arrayView1d< localIndex const > const & packList,
+                                           arrayView1d< globalIndex const > embeddedSurfacesLocalToGlobal ) const;
```
rrsettgast, line 273: &

src/coreComponents/mesh/EmbeddedSurfaceNodeManager.hpp
```diff
@@ -0,0 +1,363 @@
...
+  /**
+   * @brief deleted copy constructor
+   */
+  EmbeddedSurfaceNodeManager( const EmbeddedSurfaceNodeManager & init ) = delete;
+
+  /**
+   * @brief deleted assignement operator
+   */
+  EmbeddedSurfaceNodeManager & operator=( const EmbeddedSurfaceNodeManager & ) = delete;
+  /// @endcond
```
rrsettgast, line 103: Add defaulted move constructor, and deleted move assignment operator.

src/coreComponents/mesh/SurfaceElementSubRegion.hpp
rrsettgast: Does not to be sorted as far as I know.

src/coreComponents/mesh/utilities/ComputationalGeometry.hpp
```diff
@@ -79,8 +79,8 @@ void LinePlaneIntersection( LINEDIR_TYPE const & lineDir,
  * @param[in] normal vector normal to the plane
  */
 template< typename NORMAL_TYPE >
-void orderPointsCCW( arrayView2d< real64 > const & points,
-                     NORMAL_TYPE const & normal )
+std::vector< int >  orderPointsCCW( arrayView2d< real64 > const & points,
```
rrsettgast, line 82: does this have to be `std::vector`?

src/coreComponents/physicsSolvers/fluidFlow/integratedTests/singlePhaseFlow/fractureMatrixFlow_Embedded2d.xml
```diff
@@ -24,7 +24,8 @@
       logLevel="1"
       solidMaterialNames="{ rock }"
       targetRegions="{ Region2, Fracture }"
-      fractureRegion="Fracture"/>
+      fractureRegion="Fracture"
+      mpiCommOrder="1"/>
```
rrsettgast, line 28: what is mpiCommOrder?

rrsettgast (approved):
just some minor suggestions.

src/coreComponents/physicsSolvers/fluidFlow/integratedTests/singlePhaseFlow/fractureMatrixFlow_Embedded2d.xml
CusiniM: This is exactly the same as in the `SurfaceGenerator`.  When syncing the embedded surfaces between ranks it forces the code to process the ranks in order. If I don't do that I randomly fail the integratedTests. I believe that's because the local numbering of the ghosts on each rank will vary depending on which ghosts it receives first.

src/coreComponents/mesh/utilities/ComputationalGeometry.hpp
CusiniM: TBH, it 's an `std::vector` because that's what was used inside the function and I just made it return that object since I needed it. I suspect that the function uses it to be able to use :
```
 // sort the indices
  std::sort( indices.begin(), indices.end(), [&]( int i, int j ) { return angle[ i ] < angle[ j ]; } );
```
If you want I can copy those values in `array1d` and return that. This should only be executed on host though (if that's your concern).

corbett5: That same call works with an `array1d`. But for host only temporary stuff std::vector is fine by me.

CusiniM: I made it an `array1d`.

PR #1362, opened 2021-03-18
Add pygeosx_tools python package

cssherman:
Note: the source branch was very old, and was proving difficult to rebase.  I've manually moved the package/examples to this branch.


src/coreComponents/python/modules/pygeosx_tools_package/pygeosx_tools/mesh_interpolation.py
```diff
@@ -0,0 +1,202 @@
+
```
cssherman, line 1: Some of the functions in this file could be trimmed, as they are not very general

src/coreComponents/python/modules/pygeosx_tools_package/pygeosx_tools/wrapper.py
```diff
@@ -0,0 +1,407 @@
+
```
cssherman, line 1: This is the part of the module where most of the helper functions exist

examples/pygeosxExamples/modifyBoundaryCondition/run_fracture_problem.py
```diff
@@ -0,0 +1,57 @@
...
+                        'scale': 1.0 / 60.0,
+                        'history': []}}
+
+    # Run the code
+    while pygeosx.run() != pygeosx.COMPLETED:
+        time = problem.get_wrapper(time_key).value(False)
+        current_flow_rate = -5.0 - np.sin(np.pi * time / 60.0)
+        wrapper.set_wrapper_to_value(problem, bc_key, current_flow_rate)
+        if (wrapper.rank == 0):
+            print('t = %1.4f, q = %1.4f' % (time, current_flow_rate))
```
TotoGaz, line 47: Probably
```python
print('t = %1.4f, q = %1.4f' % (time, current_flow_rate), flush=True)
```
will do the job si you do not need `sys.stdout.flush()`.

examples/pygeosxExamples/hydaulicFractureWithMonitor/hydraulicFractureWithMonitor.py
```diff
@@ -0,0 +1,57 @@
...
+rank = comm.Get_rank()
+
+
+def run_problem():
+    """
+    Run the GEOSX problem
+    """
+    # Initialize the code
+    xml.apply_xml_preprocessor()
+    problem = pygeosx.initialize(rank, sys.argv)
```
TotoGaz, line 20: Please provide `sys.argv` as input argument of the function to reduce side effects and increase testability.

examples/pygeosxExamples/parametricSurfaceMapping/run_parametric_mesh_problem.py
```diff
@@ -0,0 +1,61 @@
...
+    Function to deform the mesh to match some surface topography
+
+    Args:
+        x (np.ndarray) the object locations
+
+    Returns:
+        np.ndarray: new coordinates
+    """
+    z_transition = 0.5   # Depth of the zone to be deformed
+    z_surf = 1.0         # Average location of the model surface
```
TotoGaz, line 20: Please provide as argument inputs (with default values).

examples/pygeosxExamples/parametricSurfaceMapping/run_parametric_mesh_problem.py
```diff
@@ -0,0 +1,61 @@
...
+import pygeosx
+from pygeosx_tools import wrapper
+
+
+def surface_mapping(x):
+    """
+    Function to deform the mesh to match some surface topography
+
+    Args:
+        x (np.ndarray) the object locations
```
TotoGaz, line 14: Please document the shape.

src/coreComponents/python/modules/pygeosx_tools_package/pygeosx_tools/file_io.py
```diff
@@ -0,0 +1,53 @@
...
+           The output directory will be created if it does not exist yet
+
+    Args:
+        grid (list): A list of numpy ndarrays defining the table axes
+        properties (dict): A dict of numpy ndarrays defning the table values
+        table_root (str): The root path for the output directory
+        axes (list): A list of names for each potential axis (default = ['x', 'y', 'z', 't'])
+    """
+    os.makedirs(table_root, exist_ok=True)
+    for ii, d in zip(range(len(grid)), axes):
```
TotoGaz, line 19: Are you OK if `grid` and `axes` do not share the same length?

src/coreComponents/python/modules/pygeosx_tools_package/pygeosx_tools/file_io.py
```diff
@@ -0,0 +1,53 @@
...
+        np.savetxt('%s/%s.csv' % (table_root, d),
+                   grid[ii],
+                   fmt='%1.5f',
+                   delimiter=',')
```
TotoGaz, line 23: Why not simply
```suggestion
    for g, a in zip(grid, axes):
        np.savetxt('%s/%s.csv' % (table_root, a),
                   g,
                   fmt='%1.5f',
                   delimiter=',')
```
? Is it equivalent?

src/coreComponents/python/modules/pygeosx_tools_package/pygeosx_tools/file_io.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def save_tables(grid, properties, table_root='./tables', axes=['x', 'y', 'z', 't']):
+    """
+    Saves a set of tables in GEOSX format
+    Notes: The shape of these arrays should match the length of each axis in the specified order
+           The output directory will be created if it does not exist yet
+
+    Args:
+        grid (list): A list of numpy ndarrays defining the table axes
```
TotoGaz, line 13: `grid` being an iterable, don't you think you could give a plural name? (maybe `grids`, but I am not sure).

src/coreComponents/python/modules/pygeosx_tools_package/pygeosx_tools/file_io.py
```diff
@@ -0,0 +1,53 @@
...
+        axes (list): A list of names for each potential axis (default = ['x', 'y', 'z', 't'])
+    """
+    os.makedirs(table_root, exist_ok=True)
+    for ii, d in zip(range(len(grid)), axes):
+        np.savetxt('%s/%s.csv' % (table_root, d),
+                   grid[ii],
+                   fmt='%1.5f',
+                   delimiter=',')
+
+    for ka in properties:
```
TotoGaz, line 25: Please iterate over `properties.items()` instead.

src/coreComponents/python/modules/pygeosx_tools_package/pygeosx_tools/file_io.py
```diff
@@ -0,0 +1,53 @@
...
+        axes (list): Axis file names in the target directory (with no extension)
+        properties (list): Property file names in the target directory (with not extension)
+        table_root (str): Root path for the table directory
+        extension (str): Table file extension (default = 'csv')
+
+    Returns:
+        tuple: List of axes values, and dictionary of table values
+    """
+    # Load axes
+    X = [np.loadtxt('%s/%s.%s' % (table_root, f, extension), unpack=True) for f in axes]
```
TotoGaz, line 46: Why not `for axis in axes`?

src/coreComponents/python/modules/pygeosx_tools_package/pygeosx_tools/file_io.py
```diff
@@ -0,0 +1,53 @@
...
+    N = tuple([len(tmp) for tmp in X])
+
+    # Load properties
+    p = {f: np.reshape(np.loadtxt('%s/%s.%s' % (table_root, f, extension)), N, order='F') for f in properties}
```
TotoGaz, line 50: Why not `for prop in properties`?

src/coreComponents/python/modules/pygeosx_tools_package/pygeosx_tools/mesh_interpolation.py
```diff
@@ -0,0 +1,202 @@
...
+        x (np.ndarray): 1D list/array of positions
+        bins (np.ndarray): 1D list/array of bin edges
+        targets (dict): Dict of geosx target keys, inputs to get_random_realization
+
+    Returns:
+        dict: Dictionary of random realizations
+
+    """
+    results = {}
+    for ka in targets.keys():
```
TotoGaz, line 104: Iterate over `targets.items()` instead.

src/coreComponents/python/modules/pygeosx_tools_package/pygeosx_tools/mesh_interpolation.py
```diff
@@ -0,0 +1,202 @@
...
+    Returns:
+        tuple: Surface interpolators (linear-depth, nearest-depth, linear-x, nearest-x)
+    """
+    nodes = []
+    with open(fname, 'r') as f:
+        for line in f:
+            if ('#' not in line):
+                tmp = line.split(',')
+                if ('.' in line):
+                    nodes.append([float(x) for x in tmp])
```
TotoGaz, line 126:
```suggestion
    with open(fname, 'r') as f:
        for line in filter(lambda l: ('#' not in line) and ('.' in line), f):
            nodes.append([float(x) for x in line.split(',')])
```
is more pythonic (if I did nothing wrong).

src/coreComponents/python/modules/pygeosx_tools_package/pygeosx_tools/mesh_interpolation.py
```diff
@@ -0,0 +1,202 @@
...
+
+    Returns:
+        tuple: Surface interpolators (linear-depth, nearest-depth, linear-x, nearest-x)
+    """
+    nodes = []
+    with open(fname, 'r') as f:
+        for line in f:
+            if ('#' not in line):
+                tmp = line.split(',')
+                if ('.' in line):
```
TotoGaz, line 125: What is the role of this '.'?
Plus I am surprised that `numpy.loadtxt` does not provide the proper options not to do this by hand.

src/coreComponents/python/modules/pygeosx_tools_package/pygeosx_tools/mesh_interpolation.py
```diff
@@ -0,0 +1,202 @@
...
+                    nodes.append([float(x) for x in tmp])
+
+    # Convert values
+    nodes = np.array(nodes)
+    x = np.ascontiguousarray(np.squeeze(nodes[:, 2])) / 3.281
```
TotoGaz, line 130: What is this `3.281`?

src/coreComponents/python/modules/pygeosx_tools_package/pygeosx_tools/mesh_interpolation.py
```diff
@@ -0,0 +1,202 @@
...
+    surfaces = np.array(surfaces)[Ia]
+
+    # Interpolate values onto a new grid
+    x_grid, y_grid, z_grid = axes
+    zo = np.concatenate([[-1e6], z_pilot, [0.0]], axis=0)
+    X, Y, Z = np.meshgrid(x_grid, y_grid, z_grid, indexing='ij')
+    conforming_tables = {k: np.zeros(np.shape(X)) for k in property_interps.keys()}
+    surface_elevations = np.zeros((len(x_grid), len(y_grid), len(surfaces)))
+
+    for ii in range(len(x_grid)):
```
TotoGaz, line 170:
```suggestion
    for ii, xx in enumerate(x_grid):
```

src/coreComponents/python/modules/pygeosx_tools_package/pygeosx_tools/mesh_interpolation.py
```diff
@@ -0,0 +1,202 @@
...
+                tmp_z = surface_interps[s][0](x_grid[ii], y_grid[jj])
+                if np.isnan(tmp_z):
+                    # The point is out of the surface xy range
+                    tmp_z = surface_interps[s][1](x_grid[ii], y_grid[jj])
+                    if extrapolate_surfaces:
+                        # Find nearest point, add in the average delta_z from other layers
+                        nearest_x = surface_interps[s][2](x_grid[ii], y_grid[jj])
+                        nearest_y = surface_interps[s][3](x_grid[ii], y_grid[jj])
+                        other_delta_z = []
+                        for kk in range(0, len(surfaces)):
```
TotoGaz, line 183: Do you really want to shadowe previous `kk`?

src/coreComponents/python/modules/pygeosx_tools_package/pygeosx_tools/well_log.py
```diff
@@ -0,0 +1,85 @@
...
+        file_location = 0
+        for line in f:
+            # Preamble
+            if (file_location == 0):
+                if variable_start in line:
+                    file_location += 1
+
+            # Variable definitions
+            elif (file_location == 1):
+                if ('#' not in line):
```
TotoGaz, line 31: Do you really want `'#' not in line` or do you want to check that the line is not a comment?
Splitting on `#` would be more robust then.

src/coreComponents/python/modules/pygeosx_tools_package/pygeosx_tools/well_log.py
```diff
@@ -0,0 +1,85 @@
...
+
+            # Body
+            else:
+                variable_values.append([float(x) for x in line.split()])
+
+    # Format results
+    results = {}
+    variable_values = np.array(variable_values)
+
+    for ii in range(0, len(variable_names)):
```
TotoGaz, line 47: Use enumerate instead.

Or could it be possible to use `zip` (I do not know with `numpy.squeeze`)
```python
for n, u, v in zip(variable_names, variable_units, variable_values):
        results[n] = {'units': u, 'values': np.squeeze(?v?)}
```

TotoGaz (commented):
Every time you use an index in python a kitten dies.

examples/pygeosxExamples/hydaulicFractureWithMonitor/hydraulicFractureWithMonitor.py
```diff
@@ -0,0 +1,57 @@
...
+    # Note: we will be plotting our results.
+    # This will modify the fonts so that they a bit easier to read
+    plot_font = {'weight': 'normal', 'size': 8}
+    plt.rc('font', **plot_font)
+
+    # Setup values to record
+    records = {fracture_location_key: {'label': 'Fracture Extents (m)',
+                                       'scale': 1.0,
+                                       'history': [],
+                                       'fhandle': plt.figure()},
```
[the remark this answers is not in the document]
cssherman: No, we shouldn't need a graphical back-end.  Before you import pyplot you can use a non-graphical one:

```python
import matplotlib
matplotlib.use('TkAgg')
import matplotlib.pyplot as plt
```


Also, if we do keep this functionality it would probably be worth turning this into a class to help end-users

src/coreComponents/python/modules/pygeosx_tools_package/pygeosx_tools/file_io.py
cssherman: I've added some checks to verify that the grid, property sizes are consistent.  I've also added some additional code to handle the default axes names for various lengths:

- N=1, axes=t
- N=3, axes=x, y, z
- N=4, axes=x, y, z, t
- Otherwise, x0, x1, x2, x3, x4, ....

src/coreComponents/python/modules/pygeosx_tools_package/pygeosx_tools/file_io.py
cssherman: Agreed, I've cleaned up the iterators for this and the following block

src/coreComponents/python/modules/pygeosx_tools_package/pygeosx_tools/file_io.py
cssherman: Agreed, I've changed the following to be more clear:

-  "grid" to "axes"
- "axes" to "axes_names"

src/coreComponents/python/modules/pygeosx_tools_package/pygeosx_tools/file_io.py
cssherman: Sure, I've also changed these to be consistent with the write function

src/coreComponents/python/modules/pygeosx_tools_package/pygeosx_tools/mesh_interpolation.py
cssherman: I've deleted this function, because it is too specific to the problem I was working on.

src/coreComponents/python/modules/pygeosx_tools_package/pygeosx_tools/mesh_interpolation.py
cssherman: This was a quirk of the file I was loading, but it is too specific so I've deleted it.

src/coreComponents/python/modules/pygeosx_tools_package/pygeosx_tools/mesh_interpolation.py
cssherman: I've also deleted this function.  The 3.281 is ft/m

src/coreComponents/python/modules/pygeosx_tools_package/pygeosx_tools/mesh_interpolation.py
cssherman: This function was specific to a problem I was working on and probably shouldn't belong in the repository

src/coreComponents/python/modules/pygeosx_tools_package/pygeosx_tools/mesh_interpolation.py
cssherman: This function was specific to a problem I was working on and probably shouldn't belong in the repository

src/coreComponents/python/modules/pygeosx_tools_package/pygeosx_tools/well_log.py
cssherman: It's probably more appropriate to check whether the first character is a comment.  I also need to improve the parsing of the curve information block since my original method is quite fragile.  Looking at some examples I have handy, these are the patterns that I can attempt to match:

```
~Curve Information Block
#MNEM.UNIT                 API CODE     Curve Description
#---------               -----------    ---------------------------
 DEPT.F                 00 000 000 000:  Depth
 RT90.ohmm              00 000 000 000:  90in Resistivity 2ft Res
```

```
~Curve Information Block
#MNEM.UNIT                 API CODE     Curve Description
#---------               -----------    ---------------------------
 DEPT.ft                              :DEPTH 
 GRTL1.API                             :TOTAL GAMMA RAY PASS 1
```

examples/pygeosxExamples/parametricSurfaceMapping/run_parametric_mesh_problem.py
```diff
@@ -0,0 +1,61 @@
...
+
+    # Rather than specifying the wrapper paths explicitly,
+    # search for them using a set of filters
+    location_keys = []
+    location_keys.append(wrapper.get_matching_wrapper_path(problem, ['ReferencePosition']))
+    location_keys.append(wrapper.get_matching_wrapper_path(problem, ['Region2', 'elementCenter']))
+
+    # Map the node, element locations
+    print('Warning: this mapping will change the mesh, but for things to work correctly')
+    print('         we would need to re-calculate the geometric parameters, shape functions')
```
corbett5, line 45: What would be necessary to get this to work correctly? Obviously we could add in a call to re-calculate things, but could we add in another step between `initialize` and `apply_initial_conditions` so that you only calculate these things once?

src/coreComponents/python/modules/pygeosx_tools_package/pygeosx_tools/file_io.py
```diff
@@ -0,0 +1,53 @@
...
+
+    Args:
+        grid (list): A list of numpy ndarrays defining the table axes
+        properties (dict): A dict of numpy ndarrays defning the table values
+        table_root (str): The root path for the output directory
+        axes (list): A list of names for each potential axis (default = ['x', 'y', 'z', 't'])
+    """
+    os.makedirs(table_root, exist_ok=True)
+    for ii, d in zip(range(len(grid)), axes):
+        np.savetxt('%s/%s.csv' % (table_root, d),
```
corbett5, line 20: Just curious if we would eventually want to roll this functionality into the time history stuff, or make time history IO callable as a free function from Python without needing an associated GEOSX instance. Certainly we want to make it easy to read in the time history files in Python. I just would prefer to avoid a proliferation of file types.

src/coreComponents/python/modules/pygeosx_tools_package/pygeosx_tools/wrapper.py
```diff
@@ -0,0 +1,407 @@
...
+                                      pylvarray.CPU)
+    else:
+        local_values.set_access_level(pylvarray.CPU)
+
+    if not isinstance(local_values, np.ndarray):
+        local_values = local_values.to_numpy()
+    return local_values
+
+
+def get_wrapper_par(problem, target_key, allgather=False):
```
corbett5, line 40: This doesn't get rid of duplicate ghosts right? Probably worth mentioning.

corbett5 (approved):
Sorry this took me so long to review, looks great! I'm not that good with all the intricacies of Python, but it looks like @TotoGaz  has that covered.

examples/pygeosxExamples/parametricSurfaceMapping/run_parametric_mesh_problem.py
cssherman: Sure, to get this working we'd one of the following:

- A callback to python somewhere after the mesh is created, but before the shape functions are calculated
- A method to trigger shape function updates

corbett5: Cool, is there any reason we'd want option 2 over 1? Are you ever going to want to recalculate stuff mid run?

src/coreComponents/python/modules/pygeosx_tools_package/pygeosx_tools/file_io.py
cssherman: Don't worry about it, I haven't had the time to work on this myself!

This is the code I typically use to read/write the structured table functions in python, and I could see it being called in either place.  The general use case I have in mind here is as follows:

1) Run the simulation
2) Grab the desired GEOSX variables using pygeosx
3) Interpolate the values onto a regular grid
4) Write these values into a table function, so that they can be used elsewhere

src/coreComponents/python/modules/pygeosx_tools_package/pygeosx_tools/wrapper.py
cssherman: Great point.  Do you think that we should have an option to filter ghosts here, or should we add a note and leave it to the users?

examples/pygeosxExamples/sedovWithStressFunction/run_sedov_problem.py
```diff
@@ -0,0 +1,72 @@
...
+        wrapper.print_global_value_range(problem, stress_key, 'stress')
+
+        # Gather/allgather tests
+        tmp = wrapper.gather_wrapper(problem, stress_key)
+        print(wrapper.rank, 'gather', np.shape(tmp), flush=True)
+
+        tmp = wrapper.allgather_wrapper(problem, stress_key)
+        print(wrapper.rank, 'allgather', np.shape(tmp), flush=True)
+
+        tmp = wrapper.allgather_wrapper(problem, stress_key, ghost_key=ghost_key)
```
cssherman, line 65: @corbett5 - I've added a method to filter ghosts out if the user supplies the appropriate key.  Note that the ghostRank wrapper isn't guaranteed to live at the same level as the wrapper, so we have to manually search for it.

src/coreComponents/python/modules/pygeosx_tools_package/pygeosx_tools/wrapper.py
corbett5: I think it would be a good idea to add that capability at some point, but it's a lot more complicated obviously.

examples/pygeosxExamples/sedovWithStressFunction/run_sedov_problem.py
corbett5: Awesome!

cssherman:
@TotoGaz - any other thoughts on this PR?

TotoGaz:
Oops @cssherman I was somewhere else.

Last comments for this PR. About
```
~Curve Information Block
#MNEM.UNIT                 API CODE     Curve Description
#---------               -----------    ---------------------------
 DEPT.F                 00 000 000 000:  Depth
 RT90.ohmm              00 000 000 000:  90in Resistivity 2ft Res
```
and the `#` question. I thing that some day one of your users will end up with something like
```
 RT90.ohmm              00 000 000 000:  90in Resistivity 2ft Res # Taken from project toto
```
And it will generate bugs. So maybe you should consider this too.

General remarks about the code (not to be considered right now, but for future devs maybe)

- Please consider logging with standard [logging](https://docs.python.org/3/library/logging.html) and not `print`. This will be useful for your users (and you).
- Maybe we shall use standard https://docs.python.org/3/library/typing.html some day?

cssherman:
@TotoGaz - I was busy on other things as well, so it's no problem at all!

I'll update the regex to handle an optional terminal comment in the variable description block.  And the suggestion to use logging is great.  I'll do a quick run-through and see what is worth updating.

TotoGaz:
Actually when you do your
```python
for line in f:
    if (line[0] != '#'):
```
Maybe something like
```python
line = line.split('#')[0]
```
is enough (You need to be dealing with empty strings, or lines with only spaces. But that's already the case, I think.)

cssherman:
> 
> 
> Actually when you do your
> 
> ```python
> for line in f:
>     if (line[0] != '#'):
> ```
> 
> Maybe something like
> 
> ```python
> line = line.split('#')[0]
> ```
> 
> is enough (You need to be dealing with empty strings, or lines with only spaces. But that's already the case, I think.)

Sounds good!

PR #1364, opened 2021-03-19
Triaxial Driver

joshua-white:
This PR adds a simple triaxial driver to GEOSX, so users can test out constitutive models without launching a full finite element simulation.  Current features include:

- The triaxial driver derives from Task and can be added to a standard XML input deck as a SoloEvent.
- Any solid model active in the ConstitutivePassThru can be called by the driver
- Users can run triaxial, volumetric, and oedometer tests using functions in the Function block to control the loading history.  This makes it easy to describe complicated loading tests.
- Results are output to a simple ASCII file.  It includes the stress / strain history, as well as convergence performance metrics like Newton iterations to convergence and residual error norms.
- We added the ability to call the driver within our unit testing framework.  The user can then set up a new material model and validate it with one or more loading tests (against analytical and experimental benchmarks).  The resulting output file(s) can then be saved as baselines.  The unit test will confirm the model results do not subsequently change with new developments.  This makes it easier to add solid material unit tests without a lot of code duplication.

Remaining to do:

- [ ] Extend this driver to allow for True-Triaxial loading conditions (different radial stresses in different directions, as opposed to a uniform confining pressure).
- [x] Sphinx documentation and examples


src/coreComponents/constitutive/unitTests/testTriaxial_druckerPragerExtended.xml
```diff
@@ -0,0 +1,67 @@
...
+      coordinates="{ 0.0, 3.0, 4.0, 7.0, 8.0 }"
+      values="{ 0, -0.003, -0.002, -0.005, -0.004 }"/>
+    <TableFunction
+      name="stressFunction"
+      inputVarNames="{ time }"
+      coordinates="{ 0.0, 8.0  }"
+      values="{ -1.0, -1.0 }"/>
+  </Functions>
+
+  <!-- Mesh is not used, but GEOSX throws an error without one.  Will resolve soon-->
```
joshua-white, line 53: Currently, GEOSX throws an error if we have an empty mesh block, as the initialization routines assume at least one mesh body exists.  Is it worth disabling this check, and how hard do you think it would it be?

src/coreComponents/constitutive/solid/Damage.hpp
```diff
@@ -86,6 +86,7 @@ class DamageUpdates : public UPDATE_BASE
   using UPDATE_BASE::hypoUpdate_StressOnly;
   using UPDATE_BASE::hyperUpdate;
   using UPDATE_BASE::hyperUpdate_StressOnly;
+  using UPDATE_BASE::saveConvergedState;
```
joshua-white, line 89: These changes to saveConvergedState() allow us to call these routines within a single RAJA loop, rather than having to do separate launches for strain updates and saving states.   Currently, there are then two saveConvergedState() methods available: one which is array-wise in SolidBase, and one which is point-wise (called with a specific element / quadrature point index) within SolidBaseUpdates.  I need to find a way to make sure one calls the other, but there is a minor hiccup in doing so.  For the moment, there is a risk the two methods could diverge from one another if the developer is not careful.  I'll think a bit how to fix this.  If we can, it will allow several kernel launches to be merged into one.

joshua-white:
I'm adding the sphinx documentation for the driver on the `docs-solidmodels` PR

src/coreComponents/constitutive/solid/TriaxialDriver.cpp
```diff
@@ -0,0 +1,309 @@
...
+  {
+    getline( file, line );
+  }
+
+  // read data block.  we assume the file size is consistent with m_table,
+  // but check for a premature end-of-file. we then compare results value by value.
+  // we ignore the newton iteration and residual columns, as those may be platform
+  // specific.
+
+  real64 value;
```
sytuannguyen, line 269: const?

src/coreComponents/constitutive/solid/TriaxialDriver.hpp
```diff
@@ -0,0 +1,205 @@
...
+
+  int m_numSteps;              ///< Number of load steps
+  string m_solidMaterialName;  ///< Material identifier
+  string m_mode;               ///< Test mode: triaxial, volumetric, oedometer
+  string m_strainFunctionName; ///< Time-dependent function controlling strain (role depends on test mode)
+  string m_stressFunctionName; ///< Time-dependent function controlling stress (role depends on test mode)
+  string m_outputFile;         ///< Output file (optional, no output if not specified)
+  Path m_baselineFile;         ///< Baseline file (optional, for unit testing of solid models)
+  array2d< real64 > m_table;   ///< Table storing time-history of axial/radial stresses and strains
+
```
sytuannguyen, line 113: ///< is a new way for documenting private parameters?

src/coreComponents/constitutive/solid/TriaxialDriver.hpp
```diff
@@ -0,0 +1,205 @@
...
+  static localIndex const m_numColumns = 9; ///< Number of columns in data table
+  enum columnKeys { TIME, EPS0, EPS1, EPS2, SIG0, SIG1, SIG2, ITER, NORM }; ///< Enumeration of column keys
+
+  static constexpr localIndex m_maxIter = 25;
+  static constexpr real64 m_newtonTol = 1e-6;
+  static constexpr real64 m_baselineTol = 1e-3;
+};
```
sytuannguyen, line 120: These hard coded parameters are randomly chosen or they are optimized?

src/coreComponents/constitutive/unitTests/testTriaxial_elasticIsotropic.xml
```diff
@@ -0,0 +1,62 @@
...
+
+<Problem>
+  <Tasks>
+    <TriaxialDriver
+      name="triaxialDriver"
+      material="elastic"
+      mode="triaxial" 
+      strainFunction="strainFunction"
+      stressFunction="stressFunction"
+      steps="50"
```
sytuannguyen, line 11: Loadings are defined by stress/strain functions? Can we define also pore pressure (for poroelastic coupling test), changing stress path, etc?

sytuannguyen (approved):
Excellent PR! 
Most of time when trying a new solver I come back to a single cell problem and this will simplify such a test. Also, it will be easy to match the simulations with lab test with simple input.

- Can we do this for fluid flow also? 
- I think It may be useful to extend this "driver" concept for some operational problem with pre-defined geometry and loadings

src/coreComponents/constitutive/solid/TriaxialDriver.cpp
```diff
@@ -0,0 +1,309 @@
...
+  fprintf( fp, "# column 1 = time\n" );
+  fprintf( fp, "# column 2 = axial_strain\n" );
+  fprintf( fp, "# column 3 = radial_strain_1\n" );
+  fprintf( fp, "# column 4 = radial_strain_2\n" );
+  fprintf( fp, "# column 5 = axial_stress\n" );
+  fprintf( fp, "# column 6 = radial_stress_1\n" );
+  fprintf( fp, "# column 7 = radial_stress_2\n" );
+  fprintf( fp, "# column 8 = newton_iter\n" );
+  fprintf( fp, "# column 9 = residual_norm\n" );
+
```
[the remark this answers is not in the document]
joshua-white: GEOSX (mostly) let's the user use whatever units they want, so we can't specify them here.  They will be consistent with whatever units the user has for material properties in their input deck.

src/coreComponents/constitutive/solid/TriaxialDriver.cpp
joshua-white: non-const.  this is the buffer the file values are read into.

src/coreComponents/constitutive/solid/TriaxialDriver.hpp
joshua-white: This is a standard doxygen syntax for adding @brief documentation after a member is defined. https://www.doxygen.nl/manual/docblocks.html

src/coreComponents/constitutive/solid/TriaxialDriver.hpp
joshua-white: I would say they are "expertly chosen."  They should work well in most situations, and I don't think we want the user to have to set these parameters.  If they don't work, there is probably something else wrong.

src/coreComponents/constitutive/unitTests/testTriaxial_elasticIsotropic.xml
joshua-white: Yes, maybe in a follow up PR?  I'm keeping a list of additional features, and this would be a good one.

jhuang2601 (approved):
Excellent PR. I've conducted testes with one set of the experimental data and compared to the previous GEOSX results without using TriaxialDriver. I checked the extended Drucker-Prager case with either strain hardening or weakening. The cases with TriaxialDriver were running much fast than full GEOSX simulation and nearly identical results were obtained, which was encouraging. With this new feature, calibrating the existing constitutive models with the corresponding experiments could be done in a cost-effective way. And this feature can be potentially extended for any rock or fluid mechanical model.

jhuang2601:
> Excellent PR!
> Most of time when trying a new solver I come back to a single cell problem and this will simplify such a test. Also, it will be easy to match the simulations with lab test with simple input.
> 
> * Can we do this for fluid flow also?
> * I think It may be useful to extend this "driver" concept for some operational problems with pre-defined geometry and loadings

src/coreComponents/constitutive/unitTests/testTriaxial_elasticIsotropic.xml
jhuang2601: Along the same line, a "wellbore driver" could be developed to solve the operational problems, which could be calibrated with both analytical solutions and experimental/field data.

joshua-white: @jhuang2601, I'll add this to the feature list.  We will need to talk about exactly what this looks like though.

src/coreComponents/unitTests/constitutiveTests/CMakeLists.txt
```diff
@@ -46,3 +51,21 @@ foreach(test ${gtest_geosx_tests})
...
+blt_add_executable( NAME testTriaxial
+                    SOURCES testTriaxial.cpp
+                    OUTPUT_DIR ${TEST_OUTPUT_DIRECTORY}
+                    DEPENDS_ON ${dependencyList} "-lz"
+                  )
+
+foreach(test ${gtest_triaxial_xmls})
+    get_filename_component( test_name ${test} NAME_WE )
+    blt_add_test( NAME ${test_name} 
+                  COMMAND testTriaxial -i ${CMAKE_CURRENT_LIST_DIR}/${test}
```
rrsettgast, line 69: This is more of an integrated test than a unit test? But I guess you can't perform an integratedTest without a mesh?

src/coreComponents/constitutive/solid/TriaxialDriver.hpp
```diff
@@ -0,0 +1,205 @@
...
+
+
+template< typename SOLID_TYPE >
+void TriaxialDriver::runStrainControlTest( SOLID_TYPE & solid, arrayView2d< real64 > & table )
```
rrsettgast, line 124: if this is only called from the source file, it can be defined in the source file instead of the header.

src/coreComponents/constitutive/solid/TriaxialDriver.hpp
```diff
@@ -0,0 +1,205 @@
...
+
+      table( n, ITER ) = 1;
+    }
+  } );
+}
+
+
+template< typename SOLID_TYPE >
+void TriaxialDriver::runMixedControlTest( SOLID_TYPE & solid, arrayView2d< real64 > & table )
+{
```
rrsettgast, line 155: This too

rrsettgast (approved):
Looks good. A couple of things for future consideration:
1) You can unit test this without reading in xml files. 
2) The unit tests are more of an integrated test.
3) It seems that the driver can be run on a GPU, but there are not tests on a GPU. 

It may be a good idea to ask Eric H, or Oleg about some feature of their constitutive driver that have proven useful for them.

src/coreComponents/constitutive/solid/TriaxialDriver.cpp
```diff
@@ -0,0 +1,306 @@
...
+    compareWithBaseline();
+  }
+
+  return false;
+}
+
+
+void TriaxialDriver::outputResults()
+{
+  FILE * fp = fopen( m_outputFile.c_str(), "w" );
```
TotoGaz, line 222: Does that respect the `-o` flag in case of relative path provided?
I think we had some issues with @jhuang2601 with this.

rrsettgast: @TotoGaz Don't we have facilities to convert paths to absolute?

TotoGaz: Do you mean on the user side or on the developer side?
I don't know for both event if I imagine we have something on the dev side.

But here we directly dump the results directly in the string provided by the user; we should "_sanitize_" this on the developer side.

PR #1371, opened 2021-03-23
Eighth-order finite-difference tests with RAJA (for discussion only)

francoishamon:
This PR is here for discussion but should not be reviewed/merged.

It implements an 8th finite-difference scheme for acoustic wave propagation. The goal is to test the performance of the RAJA kernels. So far the implementation is CPU only, but we also have another version based on `LvArray` that could be used on GPU.

We are trying to understand why the CPU version is slower (about two times slower) when the standard `for` loops are replaced (see question below).

### DIFF
--- a/src/fdwave_raja/target_3d.cpp
+++ b/src/fdwave_raja/target_3d.cpp
@@ -1,0 +1,300 @@
+#include "target_3d.hpp"
+#include "grid.hpp"
+#include "constants.hpp"
+
+#include "memoryManager.hpp"
+#include "RAJA/RAJA.hpp"
+
+#define LAP (coef0*u[IDX3_l(i,j,k)] \
+                +coefx[1]*(u[IDX3_l(i+1,j,k)]+u[IDX3_l(i-1,j,k)]) \
+                +coefy[1]*(u[IDX3_l(i,j+1,k)]+u[IDX3_l(i,j-1,k)]) \
+                +coefz[1]*(u[IDX3_l(i,j,k+1)]+u[IDX3_l(i,j,k-1)]) \
+                +coefx[2]*(u[IDX3_l(i+2,j,k)]+u[IDX3_l(i-2,j,k)]) \
+                +coefy[2]*(u[IDX3_l(i,j+2,k)]+u[IDX3_l(i,j-2,k)]) \
+                +coefz[2]*(u[IDX3_l(i,j,k+2)]+u[IDX3_l(i,j,k-2)]) \
+                +coefx[3]*(u[IDX3_l(i+3,j,k)]+u[IDX3_l(i-3,j,k)]) \
+                +coefy[3]*(u[IDX3_l(i,j+3,k)]+u[IDX3_l(i,j-3,k)]) \
+                +coefz[3]*(u[IDX3_l(i,j,k+3)]+u[IDX3_l(i,j,k-3)]) \
+                +coefx[4]*(u[IDX3_l(i+4,j,k)]+u[IDX3_l(i-4,j,k)]) \
+                +coefy[4]*(u[IDX3_l(i,j+4,k)]+u[IDX3_l(i,j-4,k)]) \
+                +coefz[4]*(u[IDX3_l(i,j,k+4)]+u[IDX3_l(i,j,k-4)]))
+
+void compute_inner_3d(RAJA::Index_type const x3, const RAJA::Index_type x4, const RAJA::Index_type y3,
+                      RAJA::Index_type const y4, const RAJA::Index_type z3, const RAJA::Index_type z4,
+                      RAJA::Index_type const lx, const RAJA::Index_type ly, const RAJA::Index_type lz,
+                      RAJA::View< const float, RAJA::Layout<1, RAJA::Index_type, 0> > &__restrict__ coefxViewConst,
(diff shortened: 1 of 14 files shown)


src/fdwave_raja/target_3d.cpp
```diff
@@ -0,0 +1,300 @@
...
+  RAJA::View< float, RAJA::Layout<3, RAJA::Index_type, 2> > phiView( phi, RAJA::Layout<3>(nx, ny, nz) );
+  RAJA::View< const float, RAJA::Layout<3, RAJA::Index_type, 2> > etaViewConst( eta, RAJA::Layout<3>(nx+2, ny+2, nz+2) );    
+
+  compute_pml_3d( RAJA::Index_type(x3), RAJA::Index_type(x4),
+                  RAJA::Index_type(y3), RAJA::Index_type(y4),
+                  RAJA::Index_type(z3), RAJA::Index_type(z4),
+                  RAJA::Index_type(lx), RAJA::Index_type(ly), RAJA::Index_type(lz),
+                  hdx_2, hdy_2, hdz_2,
+                  coefxViewConst, coefyViewConst, coefzViewConst,
+                  uViewConst, vpViewConst, etaViewConst, vView, phiView );
```
francoishamon, line 202: @corbett5 This is the main kernel of the code. To implement it using RAJA, we followed the examples with FD stencils in the RAJA performance suite. 

Here is the problem that we have: if you uncomment the code above (`compute_pml_3d`), and instead use the standard `for` loops below, the code still produces the same results, but is twice faster on the CPU, and we don't understand why. We must be doing something wrong, but I could not find what. If you spot something that we are doing wrong in `compute_pml_3d`, please let us know.

src/fdwave_raja/target_3d.cpp
```diff
@@ -167,6 +167,43 @@ void compute_pml_3d(RAJA::Index_type const x3, const RAJA::Index_type x4, const
                                      *(uViewConst(i+lx,j+ly,k+1+lz)-uViewConst(i+lx,j+ly,k-1+lz))*hdz_2 ) )
                                  / (1.f+etaViewConst(i+1,j+1,k+1));
                              });
+#else
+  for (llint i = x3; i < x4; ++i) {
```
corbett5, line 171: Same loop body as above but with for-loops and it gets the same poor performance.

src/fdwave_raja/target_3d.cpp
```diff
@@ -200,33 +251,82 @@ void target_pml_3d(llint nx, llint ny,llint nz,
...
+    >
+  >;
 
-  //              v[IDX3_l(i,j,k)] =
-  //                  ((2.f-eta[IDX3_eta1(i,j,k)]*eta[IDX3_eta1(i,j,k)]
-  //                  +2.f*eta[IDX3_eta1(i,j,k)])*u[IDX3_l(i,j,k)]
-  //                  -v[IDX3_l(i,j,k)]
-  //                  +vp[IDX3(i,j,k)]*
-  //                  (lap+phi[IDX3(i,j,k)]))/(1.f+2.f*eta[IDX3_eta1(i,j,k)]);
+  RAJA::kernel<POLICY>( RAJA::make_tuple(XRange, YRange, ZRange),
```
corbett5, line 276: Same loop body as the native one below, but with RAJA::kernel and it gets the same (within less than 5%) good performance.

corbett5 (commented):
My guess is that it's something to do with the RAJA::View. In the LvArray benchmarks there are multiple places where the RAJA::View and LvArray classes perform (usually about 2x) worse than a `restrict` pointer. If you remove the `restrict` from the pointer then they achieve parity. These cases (not very common, and more common on Lassen) are almost always the native for-loops, and as soon as you capture them in a lambda they go away.

When I removed `restrict` from the pointers the original for-loop shot up to match `target_pml_3d`. But you are capturing the views in a lambda so I'm not sure what to do about it.

corbett5:
I got the same results with clang 10.0.0. I'd try and write something (both for-loops and RAJA::kernel) using 1D RAJA Views. You should be able to just copy and paste the existing loop body and change the variable names. If you're still getting the same performance I'd create an issue with RAJA.

francoishamon:
> I got the same results with clang 10.0.0. I'd try and write something (both for-loops and RAJA::kernel) using 1D RAJA Views. You should be able to just copy and paste the existing loop body and change the variable names. If you're still getting the same performance I'd create an issue with RAJA.

Thanks a lot for your help! I just pushed a new version using 1D `RAJA::View`. 

In fact, it seems that the only cases in which the code is slow is when I create a `RAJA::View` in one function, and then I pass it to another function (for instance `compute_inner_3d`) in which I do the computation inside a loop (either a RAJA::kernel or a standard for loop. Apparently, passing the `RAJA::View`s makes me "lose" the benefit of the `__restrict__` keyword that had been applied to the underlying pointer (even if I pass the `RAJA::View` by reference with `__restrict__`), which slows down the code considerably. This happens for `KERNEL_OPTION`s 2, 3, and 4.

However, whenever I do all the computations in the function in which the `RAJA::View`s are created, everything works fast, with the for-loops as well as with the `RAJA::kernel`. See for example `KERNEL_OPTION`s 1 and 2.

corbett5:
Very interesting. Yeah I'd create an issue with RAJA tomorrow.

francoishamon:
> Very interesting. Yeah I'd create an issue with RAJA tomorrow.

Ok thanks will do. I am closing this PR for now, I will find a better place for it.

PR #1375, opened 2021-03-26
Reducing visibility of neighborRank in NeighborCommunicator

TotoGaz:
Reducing visibility of `m_neighborRank`.
Inverting `NeighborCommunicator`'s constructor dependency.

`m_neighborRank` could be modified by a public setter while it's unnecessary.
Generic constructor of `NeighborCommunicator` was depending on specialized constructor: dependency inverted.
Remove useless forward declarations.
Deleted commented code.

### DIFF
--- a/src/coreComponents/mpiCommunications/NeighborCommunicator.hpp
+++ b/src/coreComponents/mpiCommunications/NeighborCommunicator.hpp
@@ -12,10 +12,6 @@
  * ------------------------------------------------------------------------------------------------------------
  */
 
-/**
- * @file NeighborCommunicator.hpp
- */
-
 #ifndef GEOSX_MPICOMMUNICATIONS_NEIGHBORCOMMUNICATOR_HPP_
 #define GEOSX_MPICOMMUNICATIONS_NEIGHBORCOMMUNICATOR_HPP_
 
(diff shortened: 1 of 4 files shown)


src/coreComponents/mpiCommunications/NeighborCommunicator.hpp
```diff
@@ -39,16 +35,14 @@ inline int CommTag( int const GEOSX_UNUSED_PARAM( senderRank ),
...
 {
 public:
 
-  NeighborCommunicator();
+  explicit NeighborCommunicator( int rank );
 
-  NeighborCommunicator( int rank ):
-    NeighborCommunicator()
-  { setNeighborRank( rank ); }
+  NeighborCommunicator():
```
corbett5, line 44: Does it even need to be default construct-able? I would say it probably shouldn't be.

corbett5 (approved):

src/coreComponents/mpiCommunications/NeighborCommunicator.hpp
TotoGaz: Actually yes it does since we have a `std::vector< NeighborCommunicator >` which is managed through `Group` and `Wrappers`. This induces some default constructor constraint for some STL stuff (`resize` if I can read correctly).

But this is indeed utility constraint and not business and we were close to be able to remove it. 🤷

PR #1376, opened 2021-03-29
Implement dedicated corner-point mesh reader

francoishamon:
This PR implements a dedicated corner-point mesh reader in `meshUtilities`. The code currently only supports conforming corner-point meshes, although the final goal is to handle non-conforming meshes. The main steps are:
1. Partition the mesh into boxes "by hand" using a simple Cartesian approach.
2. Read keywords `ZCORN`, `COORD`, and `ACTNUM` from the file.
3. On each rank, compute the position of the vertices and filter out duplicates.
4. On each rank, assign global indices for cells and vertices using (i,j,k) indices.

Steps 1, 2, 3, and 4 seems to be working well in my initial tests. Currently for conforming meshes, I can reproduce PAMELA's behavior in serial and in parallel.

Todo-list of things to implement:
- [x] Reverse Z
- [ ] Filter out cells with small pore volume
- [ ] Decide on a temporary solution to treat ill-shaped hexas (remove?)
- [ ] Smarter partitioning
- [ ] Build faces


src/coreComponents/meshUtilities/CPMesh/CPMeshBuilder.cpp
```diff
@@ -0,0 +1,409 @@
...
+  }
+
+  GEOSX_THROW_IF( counter != nOwnedActiveCells,
+                  "Partitioning error for the cells",
+                  InputError );
+
+  // for now, we don't care about global indices for the overlaps...
+}
+
+void CPMeshBuilder::assignGlobalVertexIndices()
```
francoishamon, line 367: My initial plan was to use the position of the corner-point vertices to compute a global index that uniquely identifies each vertex in the mesh (something similar to what `nodeGlobalIndex` does in `InternalMeshGenerator`). However, corner-point meshes contain a lot of duplicate vertices. So, when I filter out duplicates, I remove a lot of vertices from the meshes (with their corresponding global indexes). 

As a result, in the final mesh, global indices are no longer contiguous. The largest global index is larger than the total number of vertices in the mesh, and some numbers in the interval [0, maxGlobalIndex] don't have any vertex associated with them. Hello @rrsettgast  @corbett5 @klevzoff do you know if having non-contiguous global indices can create problems somewhere in GEOSX? Maybe in `communicationTools` or in `dofManager`?

I also have another way to assign global vertex indices (MPI communications, I drafted something below), but it involves synchronization at the MPI partition boundaries and is much more complicated.

klevzoff: AFAIK non-contiguous global indices aren't a problem. About the only thing I know that might not work is `LAIHelperFunctions::createPermutationMatrix`, which is just a debugging tool that has limitations.

src/coreComponents/meshUtilities/CPMesh/CPMeshBuilder.cpp
```diff
@@ -0,0 +1,401 @@
...
+  array2d< real64 > const & cPVertices = m_meshVertices.m_cPVertices;
+  array1d< bool > const & cPVertexIsInsidePartition = m_meshVertices.m_cPVertexIsInsidePartition;
+  array1d< globalIndex > const & cPVertexToGlobalCPVertex = m_meshVertices.m_cPVertexToGlobalCPVertex;
+  array1d< localIndex > & cPVertexToVertex = m_meshVertices.m_cPVertexToVertex;
+  cPVertexToVertex.resize( cPVertices.size() );
+
+  // First step: filter the unique vertices using a set
+
+  std::set< Vertex, CompareVertices > uniqueVerticesHelper;
+  std::set< Vertex, CompareVertices >::iterator it;
```
TotoGaz, line 333: This deserves an `auto` (just checking against `.end()` so...)

src/coreComponents/meshUtilities/CPMesh/CPMeshBuilder.cpp
```diff
@@ -0,0 +1,401 @@
...
+{
+  array2d< real64 > const & cPVertices = m_meshVertices.m_cPVertices;
+  array1d< bool > const & cPVertexIsInsidePartition = m_meshVertices.m_cPVertexIsInsidePartition;
+  array1d< globalIndex > const & cPVertexToGlobalCPVertex = m_meshVertices.m_cPVertexToGlobalCPVertex;
+  array1d< localIndex > & cPVertexToVertex = m_meshVertices.m_cPVertexToVertex;
+  cPVertexToVertex.resize( cPVertices.size() );
+
+  // First step: filter the unique vertices using a set
+
+  std::set< Vertex, CompareVertices > uniqueVerticesHelper;
```
TotoGaz, line 332: `CompareVertices` deserves a threshold! What do you think?

src/coreComponents/meshUtilities/CPMesh/CPMeshComm.hpp
```diff
@@ -0,0 +1,141 @@
...
+   * @brief Const getter for catalog name of this class
+   * @return the catalog name
+   */
+  virtual string getCatalogName() const { return catalogName(); }
+
+  /**
+   * @brief Const getter for the list of my neighbor ranks
+   * @return a set containing the indices of neighbors
+   */
+  std::set< int > const & neighborsList() const { return m_neighborsList; }
```
TotoGaz, line 92: Is it a `std::set` of `int` or `localIndex`  or `globalIndex`?

src/coreComponents/meshUtilities/CornerPointMeshGenerator.cpp
```diff
@@ -0,0 +1,191 @@
...
+      {
+        xMax[dim] = X( iVertex, dim );
+      }
+      if( X( iVertex, dim ) < xMin[dim] )
+      {
+        xMin[dim] = X( iVertex, dim );
+      }
+    }
+  }
+  LvArray::tensorOps::subtract< 3 >( xMax, xMin );
```
TotoGaz, line 107: I think that this whole loop to find some min and max could ce in its own function.

src/coreComponents/meshUtilities/CornerPointMeshGenerator.cpp
```diff
@@ -0,0 +1,191 @@
...
+  // temporary accessors while we only support the conforming case
+  // this is not what we ultimately want, which is instead:
+  //  -> map from elem to faces
+  //  -> map from face to nodes
+  // what is below is a temporary mess (that maybe, should be hidden in CPMeshData)
+  arrayView1d< localIndex const > activeCellInsidePartitionToActiveCell = m_cPMeshBuilder->activeCellInsidePartitionToActiveCell();
+  arrayView1d< globalIndex const > activeCellToGlobalCell = m_cPMeshBuilder->activeCellToGlobalCell();
+  arrayView1d< localIndex const > activeCellToCell = m_cPMeshBuilder->activeCellToCell();
+  arrayView1d< localIndex const > cellToCPVertices = m_cPMeshBuilder->cellToCPVertices();
+  arrayView1d< localIndex const > cPVertexToVertex = m_cPMeshBuilder->cPVertexToVertex();
```
TotoGaz, line 123: Instead of exposing the vectors like that, maybe just having a member function `m_cPMeshBuilder::activeCellToCell` would make the notations lighter?

src/coreComponents/meshUtilities/CornerPointMeshGenerator.hpp
```diff
@@ -0,0 +1,93 @@
...
+   * @brief This function provides capability to post process input values prior to
+   * any other initialization operations.
+   */
+  void postProcessInput() override final;
+
+
+private:
+
+  /// Driver class for the construction of the conformal corner-point mesh
+  std::unique_ptr< CPMesh::CPMeshBuilder > m_cPMeshBuilder;
```
TotoGaz, line 84: I am confused by this `CPMeshBuilder` and `m_cPMeshBuilder` names.
Usually the builder do have to produce something. Here the builder *is* the product of its own build.
For `PAMELAMeshGenerator` the field is `m_pamelaMesh` and it's nice.
Here we could have some `m_cpMesh` e.g..

src/coreComponents/meshUtilities/CPMesh/CPMeshBuilder.hpp
```diff
@@ -0,0 +1,262 @@
...
+  CPMeshCells m_meshCells;
+
+  /// Object in charge of parsing the GRDECL file
+  CPMeshParser m_meshParser;
+
+  /// Object holding the communication info
+  CPMeshComm m_meshComm;
+
+  /// Name of the mesh
+  string m_meshName;
```
TotoGaz, line 254: Maybe remove all the `mesh` from the names?

src/coreComponents/meshUtilities/CPMesh/utilities/GeometryUtilities.hpp
```diff
@@ -0,0 +1,101 @@
...
+{
+
+namespace GeometryUtilities
+{
+
+/**
+ * @struct Vertex
+ * @brief a helper struct used to filter out duplicate nodes
+ */
+struct Vertex
```
TotoGaz, line 35: I'd be surprised we do not have this already. Good for starting nevertheless.

src/coreComponents/mpiCommunications/MpiWrapper.hpp
```diff
@@ -389,6 +389,21 @@ struct MpiWrapper
...
+   * @tparam T the type to scatter
+   * @param[in] sendValues the values to scatter across ranks
+   * @param[out] myValues the arrayView in which I receive my values
+   * @param[in] srcRank The rank of the source process within \p comm.
+   * @param[in] comm The handle to the MPI_Comm
+   * @return
+   */
+  template< typename T >
+  static int scatter( arrayView1d< T const > sendValues,
+                      arrayView1d< T > myValues,
```
TotoGaz, line 403: Here and below. Probably `destValues` or `recvValues`?

TotoGaz (commented):
I do not quite know how you will deal with merged/duplicated/removed/inserted nodes in the close future (you'll have to add node for faces so...) Maybe you'll have to build a dedicated class?

src/coreComponents/meshUtilities/CPMesh/CPMeshBuilder.hpp
```diff
@@ -0,0 +1,262 @@
...
+ * Copyright (c) 2019-     GEOSX Contributors
+ * All rights reserved
+ *
+ * See top level LICENSE, COPYRIGHT, CONTRIBUTORS, NOTICE, and ACKNOWLEDGEMENTS files for details.
+ * ------------------------------------------------------------------------------------------------------------
+ */
+
+/**
+ * @file CPMeshBuilder.hpp
+ */
```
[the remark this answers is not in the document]
corbett5: When I was working on the LvArray documentation comments such as these were necessary for Doxygen to add the header to some of its lists. I can't say exactly what it gets you, but it was worth it.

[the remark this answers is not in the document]
klevzoff: Yeah, it's what enables Doxygen to generate file synopsis pages like [this one](https://geosx-geosx.readthedocs-hosted.com/en/latest/doxygen_output/html/_data_types_8hpp.html). It's a bit strange that it can't do that by default, just using the file name itself, but that's how it is. We have these `@file` tags in every file of the project.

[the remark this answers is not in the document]
TotoGaz: Ok then. And [doc](https://www.doxygen.nl/manual/commands.html#cmdfile) says that maybe just `@file` is enough?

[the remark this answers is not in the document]
klevzoff: > Ok then. And [doc](https://www.doxygen.nl/manual/commands.html#cmdfile) says that maybe just `@file` is enough?

Yes, by the looks of it. What's also interesting is that we include this tag even in `.cpp` files, although nothing in them needs to be documented (for doxygen, anyway). I'm not really sure why we're doing that.

src/coreComponents/meshUtilities/CPMesh/utilities/StringUtilities.cpp
```diff
@@ -0,0 +1,192 @@
...
+    // Read a chunk from the buffer
+    splitBuffer >> chunk;
+    if( chunk.size() > 0 )
+    {
+      // does this word contain a star "*"
+      positionOfStar = int(chunk.find( star ));
+      if( positionOfStar>0 )
+      {
+        std::string const nTimesValueIsRepeated( chunk.substr( 0, positionOfStar ));
+        std::string const valueAsString( chunk.substr( positionOfStar+1, chunk.size()));
```
sytuannguyen, line 53:
```suggestion
        std::string const valueAsString( chunk.substr( positionOfStar+1, chunk.size() ) );
```

src/coreComponents/meshUtilities/CPMesh/utilities/StringUtilities.cpp
```diff
@@ -0,0 +1,192 @@
...
+
+  // Traverse through all chunks
+  do
+  {
+    // Read a chunk from the buffer
+    splitBuffer >> chunk;
+    if( chunk.size() > 0 )
+    {
+      // does this word contain a star "*"
+      positionOfStar = int(chunk.find( star ));
```
sytuannguyen, line 49:
```suggestion
      positionOfStar = int( chunk.find( star ) );
```

src/coreComponents/meshUtilities/CPMesh/utilities/StringUtilities.cpp
```diff
@@ -0,0 +1,192 @@
...
+      if( positionOfStar>0 )
```
sytuannguyen, line 50:
```suggestion
      if( positionOfStar > 0 )
```

src/coreComponents/meshUtilities/CPMesh/utilities/StringUtilities.cpp
```diff
@@ -0,0 +1,192 @@
...
+        repeatedValue = std::stod( chunk ); // conversion to double
+        v.push_back( repeatedValue ); // append the repeatedValue once
+      }
+    }
+    // While there is more to read
+  }
+  while (splitBuffer);
+}
+
+void eclipseDataBufferToVector( std::string & inputBuffer, std::vector< int > & v )
```
sytuannguyen, line 70:
```suggestion
void eclipseDataBufferToVector( string & inputBuffer, std::vector< int > & v )
```

src/coreComponents/meshUtilities/CPMesh/utilities/OutputUtilities.hpp
```diff
@@ -0,0 +1,92 @@
...
+  }
+  myfile.close();
+}
+
+
+} // end namespace OutputUtilities
+
+} // end namespace CPMesh
+
+} // end namespace geosx
```
sytuannguyen, line 90:
```suggestion
} // namespace geosx
```

src/coreComponents/meshUtilities/CPMesh/CPMeshParser.cpp
```diff
@@ -0,0 +1,372 @@
...
+  }
+
+  GEOSX_THROW_IF( !foundCOORD || !foundZCORN,
+                  "At least one of the following keywords was not found: COORD, ZCORN",
+                  InputError );
+
+
+}
+
+void CPMeshParser::readLocalCOORD( std::istringstream & meshFile,
```
sytuannguyen, line 166: meshFile is a const?

src/coreComponents/meshUtilities/CPMesh/CPMeshParser.cpp
```diff
@@ -0,0 +1,372 @@
...
+  string buffer = extractDataBelowKeyword( meshFile ); // TODO: change to save only what is necessary
+  CPMeshStringUtilities::eclipseDataBufferToVector( buffer, work ); // TODO: change to save only what is necessary
+
+  localIndex const nXLocal = cPMeshDims.nXLocal();
+  localIndex const nYLocal = cPMeshDims.nYLocal();
+
+  localIndex const iMinLocal = cPMeshDims.iMinLocal();
+  localIndex const jMinLocal = cPMeshDims.jMinLocal();
+
+  m_coord.resize( 6*(nXLocal+1)*(nYLocal+1) );
```
sytuannguyen, line 185: why 6? `m_coord` contains only x and y coordinates right?

src/coreComponents/meshUtilities/CPMesh/CPMeshParser.cpp
```diff
@@ -0,0 +1,372 @@
...
+  {
+    for( localIndex j = 0; j < nYLocal; ++j )
+    {
+      for( localIndex i = 0; i < nXLocal; ++i )
+      {
+        localIndex const i_xmLocal = k*8*nXLocal*nYLocal + j*4*nXLocal + 2*i;
+        localIndex const i_xpLocal = i_xmLocal + 2*nXLocal;
+        localIndex const i_xm = (k+kMinLocal)*8*nX*nY + (j+jMinLocal)*4*nX + 2*(i+iMinLocal);
+        localIndex const i_xp = i_xm + 2*nX;
+
```
sytuannguyen, line 239: `iXm`,` iXp`, etc.

src/coreComponents/meshUtilities/CPMesh/CPMeshParser.cpp
```diff
@@ -0,0 +1,372 @@
...
+  for( localIndex k = 0; k < nZLocal; ++k )
+  {
+    for( localIndex j = 0; j < nYLocal; ++j )
+    {
+      for( localIndex i = 0; i < nXLocal; ++i )
+      {
+        localIndex const eiLocal = k*nXLocal*nYLocal + j*nXLocal + i;
+        localIndex const ei = (k+kMinLocal)*nX*nY + (j+jMinLocal)*nX + i+iMinLocal;
+        prop( eiLocal ) = work[ ei ];
```
sytuannguyen, line 331: These readers share a lot of codes, is it profitable to mix them up? For example, the reading of `actnum` and `prop` are almost identical.

sytuannguyen (approved):
@francoishamon I am interested in your PR because I worked on a python script to read GRDECL files for an operational problem. Therefore I read through your code and added some minor comments. I'd like to ask some basic questions:
- Will this replace PAMELA?
- Is it possible to mixup/simplify the coordinate and property readers because they share a lot of codes? For example, the readers of `actnum` and `prop` are almost identical

src/coreComponents/meshUtilities/CPMesh/CPMeshBuilder.cpp
```diff
@@ -0,0 +1,401 @@
...
+                                             array1d< real64 > const & coord,
+                                             array1d< real64 > const & zcorn,
+                                             array1d< real64 > & xPos,
+                                             array1d< real64 > & yPos,
+                                             array1d< real64 > & zPos,
+                                             array1d< bool > & cPVertexIsInside ) const
+{
+  localIndex const i_xmLocal = k*8*nXLocal*nYLocal + j*4*nXLocal + 2*i;
+  localIndex const i_xpLocal = i_xmLocal + 2*nXLocal;
+
```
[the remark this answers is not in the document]
sytuannguyen: The factor 2 here for all the indexes is because of node douplication in all the direction?

src/coreComponents/meshUtilities/CMakeLists.txt
```diff
@@ -16,6 +17,13 @@ set(meshUtilities_headers
     SimpleGeometricObjects/Box.hpp
     SimpleGeometricObjects/ThickPlane.hpp
     SimpleGeometricObjects/BoundedPlane.hpp
+    CPMesh/CPMeshBuilder.hpp
+    CPMesh/CPMeshParser.hpp
+    CPMesh/CPMeshData.hpp
+    CPMesh/CPMeshComm.hpp
+    CPMesh/utilities/StringUtilities.hpp
+    CPMesh/utilities/GeometryUtilities.hpp
+    CPMesh/utilities/OutputUtilities.hpp
```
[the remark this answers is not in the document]
francoishamon: Done, I renamed the folder

src/coreComponents/meshUtilities/CPMesh/CPMeshBuilder.hpp
```diff
@@ -0,0 +1,262 @@
...
+   * @return the 2D array of vertex positions
+   */
+  arrayView2d< real64 const > vertices() const
+  { return m_meshVertices.m_vertices.toViewConst(); }
+
+  /**
+   * @brief Const getter for the map from corner-point vertex to (unique) vertex
+   * @return the map from corner-point vertex to (unique) vertex
+   */
+  arrayView1d< localIndex const > cPVertexToVertex() const
```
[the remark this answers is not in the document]
francoishamon: Done, I use the `cp` prefix now.

src/coreComponents/meshUtilities/CPMesh/CPMeshBuilder.hpp
francoishamon: Yes, I removed `mesh` from all these names

src/coreComponents/mpiCommunications/MpiWrapper.hpp
francoishamon: Done, I renamed `myValues` to `recvValues`, and in the serial implementation I now copy `sendValues` into `recvValues`.

src/coreComponents/meshUtilities/CPMesh/CPMeshBuilder.hpp
```diff
@@ -0,0 +1,262 @@
...
+
+#include "dataRepository/ObjectCatalog.hpp"
+#include "meshUtilities/CPMesh/CPMeshParser.hpp"
+#include "meshUtilities/CPMesh/CPMeshData.hpp"
+#include "meshUtilities/CPMesh/CPMeshComm.hpp"
+
+namespace geosx
+{
+
+namespace CPMesh
```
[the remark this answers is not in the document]
francoishamon: Done, the namespace is now called `cornerPointMesh`.

src/coreComponents/meshUtilities/CPMesh/CPMeshParser.cpp
```diff
@@ -0,0 +1,371 @@
...
+
+CPMeshParser::CPMeshParser( string const & name )
+  : m_meshName( name )
+{}
+
+
+void CPMeshParser::readNumberOfCells( Path const & filePath,
+                                      localIndex & nX,
+                                      localIndex & nY,
+                                      localIndex & nZ ) const
```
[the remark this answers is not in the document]
francoishamon: This function is now static.

src/coreComponents/meshUtilities/CPMesh/CPMeshParser.cpp
```diff
@@ -0,0 +1,371 @@
...
+{
+  localIndex const nXLocal = cPMeshDims.nXLocal();
+  localIndex const nYLocal = cPMeshDims.nYLocal();
+  localIndex const nZLocal = cPMeshDims.nZLocal();
+
+  m_actnum.resize( nXLocal*nYLocal*nZLocal );
+  m_actnum.setValues< serialPolicy >( 1 );
+}
+
+std::string CPMeshParser::extractDataBelowKeyword( std::istringstream & stringBlock ) const
```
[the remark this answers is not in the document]
francoishamon: This function is now static.

src/coreComponents/meshUtilities/CPMesh/CPMeshParser.cpp
```diff
@@ -0,0 +1,371 @@
...
+
+
+void CPMeshParser::readNumberOfCells( Path const & filePath,
+                                      localIndex & nX,
+                                      localIndex & nY,
+                                      localIndex & nZ ) const
+{
+  // Note: This is definitely not what we want to do (reading the full file into a string)
+  //       For now, to get a quick start, I ported this from PAMELA. I will improve that in a second step
+  string fileContent = CPMeshStringUtilities::fileToString( filePath );
```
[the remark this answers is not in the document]
francoishamon: I added the `const` keyword.

src/coreComponents/meshUtilities/CPMesh/CPMeshComm.hpp
```diff
@@ -0,0 +1,141 @@
...
+
+  /**
+   * @brief Rank 0 partitions the domain, and the communication pattern is set up
```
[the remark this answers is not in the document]
francoishamon: I updated the inline comment as you suggest here.

src/coreComponents/meshUtilities/CPMesh/CPMeshComm.hpp
```diff
@@ -0,0 +1,141 @@
...
+{
+
+namespace CPMesh
+{
+
+/**
+ * @class CPMeshComm
+ * @brief This class is in charge of MPI communications during the construction of the mesh
+ */
+class CPMeshComm
```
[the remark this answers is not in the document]
francoishamon: I renamed this class `CornerPointMeshPartition`.

src/coreComponents/meshUtilities/CPMesh/CPMeshBuilder.cpp
```diff
@@ -0,0 +1,401 @@
...
+CPMeshBuilder::CPMeshBuilder( string const & name )
+  :
+  m_meshParser( name ),
+  m_meshComm( name ),
+  m_meshName( name )
+{}
+
+void CPMeshBuilder::buildMesh( Path const & filePath )
+{
+  localIndex const iRank = MpiWrapper::commRank( MPI_COMM_GEOSX );
```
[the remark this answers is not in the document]
francoishamon: I use `rank` here now.

src/coreComponents/meshUtilities/CPMesh/CPMeshComm.cpp
```diff
@@ -0,0 +1,142 @@
...
+  array1d< localIndex > neighbors( sizePerRank*m_size );
+
+  if( m_myRank == m_mainRank )
+  {
+    // Temporary code to facilitate debugging: 1D partitioning along the x-direction.
+    // Ultimately here we will partition in both X and Y, taking into account active cells in each partition.
+    // Once I am sure that everything works on large test case, I will rewrite this entirely
+    localIndex const spacing = floor( static_cast< real64 >( meshDims.nX() )
+                                      / static_cast< real64 >( m_size ) );
+    for( localIndex iRank = 0; iRank < m_size; ++iRank )
```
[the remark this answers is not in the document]
francoishamon: I use `rank` here now.

src/coreComponents/meshUtilities/CPMesh/CPMeshComm.cpp
```diff
@@ -0,0 +1,142 @@
...
+}
+
+void CPMeshComm::scatterPartitionBoundaries( CPMeshDimensions & meshDims )
+{
+  localIndex const sizePerRank = 4;
+  array1d< localIndex > boundaries( sizePerRank*m_size );
+  array1d< localIndex > overlaps( sizePerRank*m_size );
+  array1d< localIndex > neighbors( sizePerRank*m_size );
+
+  if( m_myRank == m_mainRank )
```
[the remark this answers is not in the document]
francoishamon: I added a comment for `Step 0` as you suggest.

src/coreComponents/meshUtilities/CPMesh/CPMeshParser.cpp
```diff
@@ -0,0 +1,371 @@
...
+  m_coord.resize( 6*(nXLocal+1)*(nYLocal+1) );
+
+  // Note: Below I mimic what I will ultimately do: not storing the full file in a string,
+  //       but instead going through the file and saving only what the MPI rank needs
+  for( localIndex j = 0; j < nYLocal+1; ++j )
+  {
+    for( localIndex i = 0; i < nXLocal+1; ++i )
+    {
+      localIndex const indPillarLocal = j*(nXLocal+1)+i;
+      localIndex const indPillar = (j+jMinLocal)*(nX+1)+(i+iMinLocal);
```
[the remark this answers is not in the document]
francoishamon: Yes this is the pillar index. I renamed it `iPillar` to be consistent with the convention I used in these classes.

src/coreComponents/meshUtilities/CPMesh/CPMeshParser.cpp
```diff
@@ -0,0 +1,371 @@
...
+                                  CPMeshDimensions const & cPMeshDims,
+                                  array1d< real64 > & prop )
+{
+  localIndex const nX = cPMeshDims.nX();
+  localIndex const nY = cPMeshDims.nY();
+  localIndex const nZ = cPMeshDims.nZ();
+
+  localIndex const maxSize = nX*nY*nZ; // TODO: change
+  std::vector< real64 > work;
+  work.reserve( maxSize );
```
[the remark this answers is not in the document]
francoishamon: I renamed this variable `tmp`.

src/coreComponents/meshUtilities/CPMesh/CPMeshBuilder.cpp
```diff
@@ -0,0 +1,401 @@
...
+            cPVertexToGlobalCPVertex( iFirstVertexLocal + pos ) = iFirstVertexGlobal + order[pos];
+            cPVertexIsInsidePartition( iFirstVertexLocal + pos ) = cPVertexIsInside( order[pos] );
+          }
+        }
+      }
+    }
+  }
+}
+
+bool CPMeshBuilder::processActiveHexahedron( localIndex const i, localIndex const j, localIndex const k,
```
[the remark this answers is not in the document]
francoishamon: This function is now `static`.

src/coreComponents/meshUtilities/CPMesh/CPMeshBuilder.cpp
```diff
@@ -0,0 +1,401 @@
...
+  array1d< bool > cPVertexIsInside( nCPVerticesPerCell );
+
+  // loop over all cells in the MPI domain, including overlaps
+  for( localIndex k = 0; k < nZLocal; ++k )
+  {
+    for( localIndex j = 0; j < nYLocal; ++j )
+    {
+      for( localIndex i = 0; i < nXLocal; ++i )
+      {
+        bool activeAndValid = false;
```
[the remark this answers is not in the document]
francoishamon: I renamed this variable as you suggest.

src/coreComponents/meshUtilities/CPMesh/CPMeshBuilder.hpp
```diff
@@ -0,0 +1,262 @@
...
+   * @param k index in the z-direction
+   * @param nXLocal number of cells in the x-direction on my rank
+   * @param nYLocal number of cells in the y-direction on my rank
+   * @param coord local content of the COORD keyword
+   * @param zcorn local content of the ZCORN keyword
+   * @param xPos x-position of the eight corner-point vertices
+   * @param yPos y-position of the eight corner-point vertices
+   * @param zPos z-position of the eight corner-point vertices
+   */
+  bool processActiveHexahedron( localIndex const i,
```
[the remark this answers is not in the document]
francoishamon: I documented the return value.

src/coreComponents/meshUtilities/CPMesh/utilities/GeometryUtilities.hpp
francoishamon: There is definitely the same thing in Pamela, but I don't think there is something like that inside GEOSX.

src/coreComponents/meshUtilities/CPMesh/CPMeshData.hpp
```diff
@@ -0,0 +1,295 @@
...
+{
+
+namespace CPMesh
+{
+
+/**
+ * @struct CPMeshDimensions
+ * @brief Struct storing global and local mesh dimensions
+ */
+struct CPMeshDimensions
```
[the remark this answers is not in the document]
francoishamon: This class just defines the dimensions of the full mesh and the dimensions of the local MPI partition. It is just a convenience struct storing these dimensions because I need to pass them around all the time to compute local and global indices. But, these dimensions do not correspond to the sub-regions that are later constructed in GEOSX.

src/coreComponents/meshUtilities/CPMesh/CPMeshComm.cpp
```diff
@@ -0,0 +1,142 @@
...
+    // Ultimately here we will partition in both X and Y, taking into account active cells in each partition.
+    // Once I am sure that everything works on large test case, I will rewrite this entirely
+    localIndex const spacing = floor( static_cast< real64 >( meshDims.nX() )
+                                      / static_cast< real64 >( m_size ) );
+    for( localIndex iRank = 0; iRank < m_size; ++iRank )
+    {
+      boundaries( iRank*sizePerRank )   = iRank * spacing;
+      boundaries( iRank*sizePerRank+1 ) = 0;
+      boundaries( iRank*sizePerRank+2 ) = ( iRank < m_size-1 ) ? ( iRank+1 ) * spacing - 1 : meshDims.nZ()-1;
+      boundaries( iRank*sizePerRank+3 ) = meshDims.nY()-1;
```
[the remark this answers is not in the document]
francoishamon: For this comment and most of the others in this class, I would prefer not doing anything for the moment. Currently the partitioning is Cartesian. Each "interior" rank has four neighbors, and could in fact deduce the dimensions of its domain as well as the indices of the neighbors without doing any communication (no scatter), as you remarked. 

For domains with a lot of inactive cells like Norne or SAIGUP, this is not going to work very well because many ranks will not have any active cell. For these cases, we will have to abandon the strictly Cartesian approach, and we will have to introduce a little bit more flexibility in the partitioning. At that point, the dimensions of each partition will have to be communicated and many things in this function will have to change (`sizePerRank`, the neighbor indices, etc).

src/coreComponents/meshUtilities/CPMesh/CPMeshComm.cpp
```diff
@@ -0,0 +1,142 @@
...
+  {
+    meshDims.defineDomainDimensions( nCellsInEachDirection[0],
+                                     nCellsInEachDirection[1],
+                                     nCellsInEachDirection[2] );
+  }
+}
+
+void CPMeshComm::scatterPartitionBoundaries( CPMeshDimensions & meshDims )
+{
+  localIndex const sizePerRank = 4;
```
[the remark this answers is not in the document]
francoishamon: Currently each "interior" partition has exactly four boundaries (i+1, i-1, j+1, j-1) and four neighbors but this is temporary.

src/coreComponents/meshUtilities/CPMesh/CPMeshBuilder.cpp
```diff
@@ -0,0 +1,401 @@
...
+
+/**
+ * @file CPMeshBuilder.cpp
+ */
+
+#include "CPMeshBuilder.hpp"
+
+#include "codingUtilities/Utilities.hpp"
+#include "mpiCommunications/MpiWrapper.hpp"
+#include "rajaInterface/GEOS_RAJA_Interface.hpp"
```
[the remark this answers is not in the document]
francoishamon: I agree that your suggestion is more consistent but this is not a file that I am going to touch in this PR.

src/coreComponents/meshUtilities/CPMesh/utilities/StringUtilities.hpp
```diff
@@ -0,0 +1,61 @@
...
+std::string fileToString( const std::string filePath );
+void trim( std::string & str );
+bool removeStringAndFollowingContentFromLine( std::string toBeRemoved, std::string & line );
+void removeTab( std::string & v );
+void removeEndOfLine( std::string & v );
+void removeExtraSpaces( std::string & v );
+
+template< typename T >
+void fromStringTo( std::string & data, std::vector< T > & v )
+{
```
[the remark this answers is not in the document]
francoishamon: Done

src/coreComponents/meshUtilities/CPMesh/utilities/StringUtilities.cpp
francoishamon: Done

src/coreComponents/meshUtilities/CPMesh/utilities/StringUtilities.cpp
francoishamon: Done

src/coreComponents/meshUtilities/CPMesh/utilities/StringUtilities.cpp
francoishamon: Done

src/coreComponents/meshUtilities/CPMesh/utilities/OutputUtilities.hpp
francoishamon: Done

src/coreComponents/meshUtilities/CPMesh/CPMeshParser.cpp
francoishamon: Done

src/coreComponents/meshUtilities/CornerPointMeshGenerator.cpp
```diff
@@ -0,0 +1,191 @@
...
+  MeshBody & meshBody = meshBodies.registerGroup< MeshBody >( this->getName() );
+
+  MeshLevel & meshLevel0 = meshBody.registerGroup< MeshLevel >( string( "Level0" ));
+  NodeManager & nodeManager = meshLevel0.getNodeManager();
+  CellBlockManager & cellBlockManager = domain.getGroup< CellBlockManager >( keys::cellManager );
+
+  // Step 0: transfer the neighbor list
+
+  // TODO: change the name, we don't use metis
+  domain.getMetisNeighborList() = m_cPMeshBuilder->neighborsList();
```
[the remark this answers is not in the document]
francoishamon: Yes, I agree. In this PR, I will try to keep the changes limited to the `meshUtilities` folder, so I will probably leave that for another PR.

src/coreComponents/fileIO/schema/docs/Mesh.rst
```diff
@@ -1,11 +1,12 @@
...
-Name                Type Default Description                    
-=================== ==== ======= ============================== 
-InternalMesh        node         :ref:`XML_InternalMesh`        
-InternalWell        node         :ref:`XML_InternalWell`        
-PAMELAMeshGenerator node         :ref:`XML_PAMELAMeshGenerator` 
-=================== ==== ======= ============================== 
+======================== ==== ======= =================================== 
+Name                     Type Default Description                         
+======================== ==== ======= =================================== 
+CornerPointMeshGenerator node         :ref:`XML_CornerPointMeshGenerator`
```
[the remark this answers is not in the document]
francoishamon: Done, the catalog name for this class is now `ExternalCornerPointMesh`.

src/coreComponents/meshUtilities/CPMesh/CPMeshBuilder.cpp
```diff
@@ -0,0 +1,401 @@
...
+#include "meshUtilities/CPMesh/utilities/GeometryUtilities.hpp"
+#include "meshUtilities/CPMesh/utilities/OutputUtilities.hpp"
+
+namespace geosx
+{
+
+namespace CPMesh
+{
+
+using namespace GeometryUtilities;
```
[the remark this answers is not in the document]
francoishamon: Done

src/coreComponents/meshUtilities/CPMesh/CPMeshBuilder.cpp
```diff
@@ -0,0 +1,401 @@
...
+
+  // post-process the mesh to make it conforming
+  postProcessMesh();
+}
+
+void CPMeshBuilder::postProcessMesh()
+{
+  // for each active cell, compute the position of the eight CPG vertices
+  // at this point, CPG vertices have not been filtered yet
+  buildCornerPointCells();
```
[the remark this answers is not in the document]
francoishamon: CPG means Corner Point Grid. In the Corner Point format, each cell has 8 corner point vertices (the `cpVertices`) of the code, so that is what these `CPG vertices` in the comment refer to.

src/coreComponents/meshUtilities/CPMesh/utilities/StringUtilities.cpp
francoishamon: Done

src/coreComponents/meshUtilities/CPMesh/CPMeshParser.cpp
francoishamon: I tried to make it `const`, but I cannot because the `getline` function used in this function takes a non-const `std::istringstream`

src/coreComponents/meshUtilities/CPMesh/CPMeshParser.cpp
francoishamon: Here is an explanation from the [OPM documentation](https://opm-project.org/wp-content/uploads/2019/01/OPM-Flow-Documentation-2018-10-Rev-2.pdf): 
```
COORD defines a set of coordinate lines or pillars for a reservoir grid via an array. 
A total of (NX+1) x (NY+1) lines must be specified for each coordinate data set. 
For Cartesian geometry, each line is defined by the (x, y, z) coordinates of two distinct points on the line. 
```
For each of two points, you need the (x,y,z) coordinates, hence 6 entries per pillar.

francoishamon:
> * Will this replace PAMELA?

In PAMELA, there is currently a limitation in the treatment of non-conforming corner-point meshes. This limitation comes from the fact that non-matching faces (along faults for example) are assumed, by default, to be impermeable. 

One of the goals of this PR and future ones on the topic is to overcome this limitation. Non-matching faces will be subdivided into smaller, matching faces to obtain a conforming mesh made of arbitrary shaped polyhedra on which we can simulate flow through these faces. The resulting conforming mesh should also be amenable for virtual element simulations. 

The processing of the faces is currently being implemented (but not pushed yet in the PR) in the `buildFaces` function of `CornerPointMeshBuilder`.

francoishamon:
Closing in favor of @IsaacJu-debug 's PR coming soon

PR #1377, opened 2021-03-29
Hide attributes and member functions in partitions.

TotoGaz:
Functionalities are unmodified but visibility is reduced to the maximum without changing the current architecture.
Commented and dead code is removed too.

I am working on Mesh "_generation_" so I wander in this area; when I think I may need to propose some modification I try to redefine clear contracts for the classes (specially the base classes). Removing dead code and reducing visibility somehow helps.


rrsettgast:
@TotoGaz I am actively moving locations for many files. Please don't merge this until the circular dependencies branch is merged.

PR #1380, opened 2021-04-02
Fixed compilation warning with clang+hypreGPU on Lassen

francoishamon:
This PR fixes the following compilation warning on Lassen when `ENABLE_HYPRE_CUDA` is set to `ON`:
```
warning: format specifies type                                                                                                                                             
      'long' but the argument has type 'geosx::localIndex' (aka 'int') [-Wformat]                                                             
sprintf(filename, "sismoTraceReceiver%0ld.txt", ircv);                                                                                        
                                     ~~~~       ^~~~                                                                                          
                                     %0d 
```


corbett5 (approved):
Thanks!

PR #1382, opened 2021-04-06
Device Packing Optimization

wrtobin:
Improve CUDA buffer packing performance by using assignment-based packing on properly aligned memory.

This reduces our packing time approximately one order of magnitude in observed cases. Further reduction should be possible through kernel register reduction and loop fusion (not in this PR).


wrtobin:
I've already run the integratedTests on both lassen and quartz, so as long as this passes the CI checks it should be good to go.

src/coreComponents/dataRepository/BufferOpsDevice.cpp
```diff
@@ -145,25 +145,25 @@ PackDataByIndexDevice ( buffer_unit_type * & buffer,
...
   localIndex const numIndices = indices.size();
-  localIndex const unitSize = var.size() / var.size( 0 ) * sizeof( T );
-
-  buffer_unit_type * const devBuffer = buffer;
+  localIndex const sliceSize = var.size() / var.size( 0 );
+  localIndex packedSize = numIndices * sliceSize * sizeof( T );
   if( DO_PACKING )
   {
+    T * devBuffer = reinterpret_cast< T * >( buffer );
     parallelDeviceStream stream;
```
corbett5, line 153: We should clean up this stream and events business later. We should not be creating a new stream for every field that we pack and we should probably do stream based synchronization instead of event based.

src/coreComponents/dataRepository/BufferOpsDevice.cpp
```diff
@@ -187,22 +199,22 @@ UnpackDataByIndexDevice ( buffer_unit_type const * & buffer,
...
+  localIndex unpackSize = numIndices * sliceSize * sizeof( T );
+  T const * devBuffer = reinterpret_cast< T const * >( buffer );
   parallelDeviceStream stream;
-  events.emplace_back( forAll< parallelDeviceAsyncPolicy<> >( stream, numIndices, [=] GEOSX_DEVICE ( localIndex const i )
+  events.emplace_back( forAll< parallelDeviceAsyncPolicy<> >( stream, numIndices, [=] GEOSX_DEVICE ( localIndex const ii )
   {
-    buffer_unit_type const * threadBuffer = devBuffer + i * unitSize;
-    LvArray::forValuesInSlice( var[ indices[ i ] ], [&threadBuffer] GEOSX_DEVICE ( T & value )
+    T const * threadBuffer = &devBuffer[ ii * sliceSize ];
+    LvArray::forValuesInSlice( var[ indices[ ii ] ], [&threadBuffer] GEOSX_DEVICE ( T & value )
```
corbett5, line 209: I was thinking about it later and you can only get rid of `forValuesInSlice` if the slice is contiguous. If we don't care the order that it's packed in, as long as it's unpacked in a consistent manner then we can probably do much better but it would take some thinking as to how best to pack things and it would likely involve a single thread packing parts of multiple objects in some cases.

corbett5 (approved):

src/coreComponents/dataRepository/BufferOpsDevice.cpp
```diff
@@ -145,25 +145,25 @@ PackDataByIndexDevice ( buffer_unit_type * & buffer,
...
     {
-      buffer_unit_type * threadBuffer = devBuffer + ii * unitSize;
-      LvArray::forValuesInSlice( var[ indices[ ii ] ], [&threadBuffer] GEOSX_DEVICE ( T const & value )
+      T * threadBuffer = &devBuffer[ ii * sliceSize ];
+      LvArray::forValuesInSlice( var[ indices[ ii ] ], [&] GEOSX_DEVICE ( T const & value )
       {
-        memcpy( threadBuffer, &value, sizeof( T ) );
-        threadBuffer += sizeof( T );
+        *threadBuffer = value;
+        threadBuffer++;
```
rrsettgast, line 160:
```suggestion
        *threadBuffer = value;
        ++threadBuffer;
```

src/coreComponents/dataRepository/BufferOpsDevice.cpp
```diff
@@ -187,22 +199,22 @@ UnpackDataByIndexDevice ( buffer_unit_type const * & buffer,
...
   {
-    buffer_unit_type const * threadBuffer = devBuffer + i * unitSize;
-    LvArray::forValuesInSlice( var[ indices[ i ] ], [&threadBuffer] GEOSX_DEVICE ( T & value )
+    T const * threadBuffer = &devBuffer[ ii * sliceSize ];
+    LvArray::forValuesInSlice( var[ indices[ ii ] ], [&threadBuffer] GEOSX_DEVICE ( T & value )
     {
-      memcpy( &value, threadBuffer, sizeof( T ) );
-      threadBuffer += sizeof( T );
+      value = *threadBuffer;
+      threadBuffer++;
```
rrsettgast, line 212:
```suggestion
      value = *threadBuffer;
      ++threadBuffer;
```

src/coreComponents/dataRepository/BufferOpsDevice.hpp
```diff
@@ -166,10 +166,10 @@ PackDataDevice( buffer_unit_type * & GEOSX_UNUSED_PARAM( buffer ),
 //------------------------------------------------------------------------------
 template< bool DO_PACKING, typename T, int NDIM, int USD, typename T_INDICES >
 typename std::enable_if< can_memcpy< T >, localIndex >::type
-PackDataByIndexDevice( buffer_unit_type * & buffer,
-                       ArrayView< T const, NDIM, USD > const & var,
-                       T_INDICES const & indices,
-                       parallelDeviceEvents & events );
+PackDataByIndexDevice ( buffer_unit_type * & buffer,
+                        ArrayView< T const, NDIM, USD > const & var,
+                        const T_INDICES & indices,
```
rrsettgast, line 171:
```suggestion
                        T_INDICES const & indices,
```

rrsettgast (approved):
Looks good. I don't know if this is "outdated" but I still always pre-increment rather than post-increment. 

@corbett5 can you merge this if the suggested changes are still worthwhile?

PR #1383, opened 2021-04-07
Remove header includes in CMakeLists.txt files

sytuannguyen:
The header dependencies are already ensured by including .hpp and .h files in the corresponding .cpp files. 
My question is: do we really need to include the header files in CMakeLists.txt files?

This PR aims to illustrate this idea. A successful build was observed after removing all the header files in CMakeLists.txt files.


corbett5:
Can you try an install build? I think we're going to have to list the headers somewhere, and at the same place we list the sources makes sense to me. I don't know this for sure but I would be quite surprised if CMake parsed the cpp files to get header dependencies.

https://stackoverflow.com/questions/36174499/why-add-header-files-into-add-library-add-executable-command-in-cmake
https://cmake.org/cmake/help/v3.14/command/add_library.html

sytuannguyen:
> Can you try an install build?

Do you mean release build? You can see that this PR passes all the Travis builds.

sytuannguyen:
> I think we're going to have to list the headers somewhere

If we want to list them somewhere and they are not required for the build, we could list them in the docs.

sytuannguyen:
> I don't know this for sure but I would be quite surprised if CMake parsed the cpp files to get header dependencies.

Can this be confirmed by the fact that all the Travis builds are passed

corbett5:
I meant run `make install`.

TotoGaz:
The people of `cmake` are providing both `cpp` and `.h` files : https://github.com/Kitware/CMake/blob/master/Source/CMakeLists.txt
In a slightly different way though (they do not split headers and sources).

sytuannguyen:
> From what I've read it seems as if you don't need to specify the headers to get a correct build, or a correct incremental build.

Yes it is. Indeed, I tried `make install` and I obsersved a successful build and the executable works well.

sytuannguyen:
> All the headers should be installed in the installation directory. I don't think anyone is actually using GEOSX like this, but we should support it.

I run `make install` on both `develop` branch and this PR branch then I checked the install directory. Both run produce the same thing. More precisely, the run on `develop` branch does not provide the header files in install directory as you expected (please check from your side to confirm this point).

So, until now I don't see any different between including and not including the header files in CMakeLists.txt.

rrsettgast:
I am closing this as I don't see a benefit to this, and I don't really understand the motivation.

PR #1391, opened 2021-04-13
Update Caliper.rst

corbett5:


francoishamon (approved):
Thank you @corbett5 I was not aware of the `max_column_width` option so I struggled with truncated names, that is going to be very useful.

corbett5:
> 
> 
> Thank you @corbett5 I was not aware of the `max_column_width` option so I struggled with truncated names, that is going to be very useful.

Yeah it was a huge improvement when they added it. Definitely should be in the docs.

PR #1392, opened 2021-04-13
Consider the effect of confining stress on SIF calculation

WuHuiLLNL:
In the presence of confining stress, a negative aperture might be possible in the SIF calculation of the SurfaceGenerator.cpp. Since the SIF equation (tipNodeSIF = pow( (fabs...) always give a positive value, the SIF will be extremely large under high confining stress, and the fracture will propagate, which is unphysical.

We manually multiply the SIF by -1 in case of negative aperture to avoid fracture propagation under very high confining stress.

### DIFF
--- a/src/coreComponents/physicsSolvers/surfaceGeneration/SurfaceGenerator.cpp
+++ b/src/coreComponents/physicsSolvers/surfaceGeneration/SurfaceGenerator.cpp
@@ -3044,6 +3044,11 @@ void SurfaceGenerator::calculateNodeAndFaceSif( DomainPartition & domain,
           tipNodeSIF = pow( (fabs( tipNodeForce[0] * trailingNodeDisp[0] / 2.0 / tipArea ) + fabs( tipNodeForce[1] * trailingNodeDisp[1] / 2.0 / tipArea )
                              + fabs( tipNodeForce[2] * trailingNodeDisp[2] / 2.0 / tipArea )), 0.5 );
 
+          if ( LvArray::tensorOps::AiBi< 3 >( trailingNodeDisp, faceNormalVector ) < 0.0)  //In case the aperture is negative with the presence of confining stress.
+          {
+            tipNodeSIF *= -1;
+          }
+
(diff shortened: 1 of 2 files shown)


WuHuiLLNL:
@rrsettgast @jhuang2601 
Hey Randy, I made some minor changes to consider the effect of confining stress on SIF calculation. The change should be straightforward. With the change, the experiment results from Wu et al. (2008) can be appropriately reproduced. Jian wants to test on Pangea2. Can you review the change?

I have also passed the integrated test. The only difference occurs in the test results of "KGD_ZeroToughness.xml", where the SIF value of a tip node changes from positive to negative (new value is something like -77770 and the previous value in baseline is 77770), which is reasonable as the aperture at the fracture tip might be negative. I have rebased the results of "KGD_ZeroToughness", and all the integrated tests are passed. In the IntegratedTest repo, you can find a new branch named "SIFCalculationwithConfiningStress". Please let me know if I also need to create a pull request in the IntegratedTest repo.

src/coreComponents/physicsSolvers/surfaceGeneration/SurfaceGenerator.cpp
```diff
@@ -3044,6 +3044,11 @@ void SurfaceGenerator::calculateNodeAndFaceSif( DomainPartition & domain,
           tipNodeSIF = pow( (fabs( tipNodeForce[0] * trailingNodeDisp[0] / 2.0 / tipArea ) + fabs( tipNodeForce[1] * trailingNodeDisp[1] / 2.0 / tipArea )
                              + fabs( tipNodeForce[2] * trailingNodeDisp[2] / 2.0 / tipArea )), 0.5 );
 
+          if ( LvArray::tensorOps::AiBi< 3 >( trailingNodeDisp, faceNormalVector ) < 0.0)  //In case the aperture is negative with the presence of confining stress.
+          {
+            tipNodeSIF *= -1;
+          }
```
rrsettgast, line 3050: If aperture is negative, we should be be splitting at all??

rrsettgast (approved):

PR #1399, opened 2021-04-16
your-platform.cmake

Scanty20:


TotoGaz:
hello @Scanty20 thanks for your proposal :)

I think that you could follow the documentation and you could not find a `your-platform.cmake` file.
A "dummy" name is provided in the doc so you can create your own.
I suggest you rename `your-platform` with a personal name and parameters that match your environment.

You've also deleted our `darwin` config file which is for our continuous integration mac support, which explains why the checks fail on this PR.

Thus, we won't be able to accept your PR.

PR #1407, opened 2021-04-23
Adding timestamp information required by visit to vtk files

cssherman:


cssherman:
Testing the new timestamp using the sedov problem in visit.  Visit is able to parse the psuedo-cycle from the file names.  In case we wanted the true cycle information included, adding it to the vtk file is similar.

sedov_time

rrsettgast (approved):

PR #1418, opened 2021-05-05
Improvements towards arbitrary polyhedral grids

TotoGaz:
Hello, here is the proposition to move forward to arbitrary polyhedral grids.
It’s not 100% over, but it’s ready to be reviewed.
It relates to https://github.com/GEOSX/GEOSX/issues/1386

GEOSX relies on _element to nodes_, _face to edges_, _node to faces_ mappings...
Most of these mappings are built using the element type information (hexs, tets…). 

The main issue is that for such arbitrary polyhedral grids, there may be strange cells with arbitrary kinds and shapes.
Thus, there may be no rules to build our associative mappings.
For now, the execution flow of GEOSX makes mandatory to go through those assembly functions.
Those assembly codes are held by `NodeManager`, `EdgeManager`, etc.

In order to deal with the issue, I’ve moved all these functions around the `CellBlockManager`/`CellBlock`/`MeshGenerator` area.
The responsibility to provide the consistent mappings relies now entirely on them.
This area is purely geometrical (elements, faces, edges, nodes).

The `Node|Edge|Face|Elem` managers are now just to retrieve the tables from the `CellBlock*` stuff.

In order to leave some room for future arbitrary polyhedral grids, on top of `CellBlockManager` and `CellBlock` are their (reduced) abstract versions `CellBlockManagerABC`/`CellBlockABC`.
Suffix `ABC` stands for `Abstract Base Class`. It could be renamed into `API`, `Base` (while `Base` has implementation flavor, here our `*ABC` classes are contracts), or _nothing_ and have different names for the _implementations_. We could use a dedicated namespace too… Whatever. Discussion is here #1555 
Any grid provider has to implement its own version of `CellBlockManagerABC`/`CellBlockABC` to set the mappings inside of the core of GEOSX. (This can’t be enough since GEOSX strongly assumes that we have sub-regions with only one type of cell).

One benefit from this architecture is that the contract we are waiting from mesh providers is clear.
It’s exactly `CellBlockManagerABC`/`CellBlockABC` (`Group` is still here on top but should not be used as part of the `API`.).
They have respectively 9 and 16 members. Which is quite large but there is some combination of node, edge, face, element, so it’s more concise than it looks like.
```
CellBlockManagerABC
├── getCellBlocks
├── numNodes
├── numEdges
├── numFaces
├── getNodesPositions
├── getNodeLocalToGlobal
├── getNodeSets
├── getNodeToEdges
├── getNodeToFaces
├── getNodeToElements
├── getEdgeToNodes
├── getEdgeToFaces
├── getFaceToNodes
├── getFaceToEdges
└── getFaceToElements
```
```
CellBlockABC
├── getElementTypeString
├── numNodesPerElement
├── numFacesPerElement
├── numElements
├── getElemToNode
├── getElemToEdges
├── getElemToFaces
├── localToGlobalMap
└── forExternalProperties< LAMBDA >( LAMBDA && lambda )
```
* Also, the possibility of misusage is reduced: there is no _mutable_ accessor/setter (except `Group` which grants potential bypass possibilities, ~~obviously `buildMaps`~~, and `CellBlockABC::forExternalProperties` for legacy reasons)
* This is interesting in terms of testing too, as an intermediate level between unit tests and integrated tests.
We can test mesh providers at module level: we only need to validate against the abstract contrat `CellBlockManagerABC`/`CellBlockABC`. What’s behind is not of our business.
* As part of the testing process, "_in the other direction_", it’s also possible to [mock](https://en.wikipedia.org/wiki/Mock_object) `CellBlockManagerABC`/`CellBlockABC`. It’s a nice technic that could help making GEOSX rock solid. Maybe cell block is not the best first candidate but, well, it’s there.
* In terms of readability, clients just have to focus on the API and shall not worry about the implementations behind them (if no bug!).
* In terms of features, it’s now easier not to compute mappings that your solver does not need.

There is now a copy for each mapping. If it’s not affordable, we could provide the output vector as argument of our accessors (like `cellBlockManager.getNodeToEdges( m_toEdgesRelation )`) or maybe use some `move` semantic.
But maybe we do not care; I do not know the price of copies w.r.t. our whole computation time.

Example:
```
NodeManager ─┬─> CellBlockManagerABC
EdgeManager ─┤              └── CellBlockManager <─┬── InternalMeshGenerator
FaceManager ─┘                                     └── PAMELAMeshGenerator
```
Right now, the `cmake` scripts does not impose this by design yet. This needs to de done so devs cannot bypass the abstraction.

If you want to invert the dependencies, move `*ABC` classes from on package to the other :wink:

The code which is building the mappings is mainly unchanged (or to some limited extent, in order to separate pure geometrical concepts to GEOSX business concepts, like `regions`).

Important, two functions implementations leak from the abstraction, they are in `CellBlockUtilities.hpp`:
- a `getFaceNodes` used twice (inside `isPointInsideElement` and inside `AcousticWaveEquationSEM::precomputeSourceAndReceiverTerm`, tests do not count).
- Some `buildEdgesMaps` ultimately called from `EmbeddedSurfaceGenerator`.

I do not know what to think about that. For the first point, it’s surely possible to rewrite using the standard mappings. For the second, I really do not know the subject.

Also, there is some direct plug from `Perforation` to `InternalWellGenerator`. And some things with `Partition`s.

Everything is not fully done, some elements should eventually be completed (in this PR or another)
- [x] Still some dependencies to `NodeManager`, dependency still should be inverted.
- [x] ~~The way to return the arrays must be improved (I’ll need help here).~~ Validated by code review. Improvements always possible.
- [x] ~~Create the `cmake` target with proper cut so we are neat and design can be enforced.~~ Automatic registration of products inside the factory makes a bit tricky to cut the libraries/targets. Plus `BLT` does not allow such flexibility when it comes to visibilities. Plus I'm no `cmake` expert. Thus, I won't be dealing with this item in the current PR.
- [x] ~~`getCellBlocks` still returns a `Group` instead of more strongly typed a container of `CellBlockABC`. I don’t know if we can do a lot here.~~ Nothing relevant to do at the scope of this PR.
- [x] Run integrated tests to be sure there’s no problem (done for CPU & GPU). Some order (not values) of nodes, faces... are different, which seems acceptable w.r.t. this PR modifications. Besides, other results are identical to the baselines: modifications of the mesh seem consistent and introduce to detectable error.
- [x] ~~Some questions about extra allocation space. Where should we consider the question?~~ Question still here, but not of the scope of this PR.
- [x] I’ll be presenting this next Thursday so I can answer your questions.

Sorry long post, most probably not to be read.
But that may be useful for further references after i've made the talk.

EDIT:  I've removed `CellBlockManagerABC::buildMaps` to reduce the footprint. Let's make it evolve smoothly with a clear use case.

Related to https://github.com/GEOSX/integratedTests/pull/184


src/coreComponents/mesh/generators/CellBlockABC.hpp
```diff
@@ -0,0 +1,131 @@
...
+   * @brief Get the number of elements.
+   * @return number of elements in the cell block
+   */
+  virtual localIndex numElements() const = 0;
+
+  /**
+   * @brief Get the element-to-node map.
+   * @return a reference to the element-to-node map
+   */
+  virtual array2d< localIndex, cells::NODE_MAP_PERMUTATION > const & getElemToNode() const = 0;
```
klevzoff, line 82: This contract seems to bind all implementations of `CellBlockABC` to a specific elem-to-node layout that is `array2d`. However "`CrazyCellBlock`" will need to use something like `ArrayOfArrays` to store its varying per-cell topology (same goes for other maps in this class), and trying to fit it into an `array2d` would be extremely memory-inefficient (think of one crazy cell that has 50 nodes in a block of otherwise "mildly" unstructured cells).

Two approaches come to mind:
1. Make the return type here the most flexible one (`ArrayOfArrays`) and make "structured" subregions like do the legwork to convert it into their internal storage format (`array2d`). This is runtime overhead, but maybe it's acceptable if it happens once per simulation.
2. Make two virtual functions (`getElemToNodeUniform()` and `getElemToNodeNonUniform()`) with different return types plus a discriminator (`bool isUniform()`), such that existing `CellElementSubRegion` will only call the former and new `CrazyElementSubRegion` will call the latter. This partially defeats the purpose of an abstract base class (?), but allows each subregion to call the implementation that is most efficient. 

Maybe my comment jumps ahead a bit, because we don't have `CrazyCellBlock` yet, and it's not decided how it's topology will be represented and stored in a subregion (I just assumed it will be `ArrayOfArrays`, but who knows). But since this whole refactoring is motivated by the need to prepare for unstructured mesh readers, I find it useful to think in advance what it's going to look like and how it can implement the ABC.

TotoGaz:
@corbett5 The most impacting changes w.r.t. your implementations are about the `FaceBuilder` struct.
It was mixing geometry (cells, indices...) and physics (regions index), so I had to split it (and therefore modify it a bit).

I renamed `FaceBuilder` into `NodesAndElementOfFace` and it stores the node information directly.

Also, I've modified `operator==` and `operator<`. I now only rely on the sorted nodes that I sort once for all and store.
You were working on the subset of three nodes, which is OK to define a unique face.
I was worried that this should not be enough for arbitrary non conformal cells, so I decided to modify it by using all the nodes. For most of our cells, faces only have 3 or 4 nodes so I don't expect anything big here. What do you think about it?
In order not to add any bugs, I relied on `std` algorithms to do the duty for me (it's the same as `std::vector::operator==` and `std::vector::operator<`).

And (e.g.) in the `CellBlockManager::populateFaceMaps`, I changed code like
```c++
    localIndex j = 0;
    for(; j < numFaces - 1; ++j )
    {
      // If two subsequent FaceBuilders compare equal then they describe an interior face.
      if( facesByLowestNode( nodeID, j ) == facesByLowestNode( nodeID, j + 1 ) )
      {
        addInteriorFace( elementManager, curFaceID, facesByLowestNode( nodeID, j ), facesByLowestNode( nodeID, j + 1 ), elemRegionList, elemSubRegionList,
                         elemList, nodeList );
        ++j;
      }
      // Otherwise it's a boundary face.
      else
      {
        addBoundaryFace( elementManager, curFaceID, facesByLowestNode( nodeID, j ), elemRegionList, elemSubRegionList, elemList, nodeList );
      }

      ++curFaceID;
    }

    if( j == numFaces - 1 )
    {
      addBoundaryFace( elementManager, curFaceID, facesByLowestNode( nodeID, j ), elemRegionList, elemSubRegionList, elemList, nodeList );
    }
```
into
```c++
    for( localIndex j = 0, curFaceID = uniqueFaceOffsets[nodeID]; j < numFaces; ++j, ++curFaceID )
    {
      // If two subsequent `NodesAndElementOfFace` compare equal then they describe an interior face.
      // It must therefore be considered "twice".
      // Otherwise it's a boundary face, and "once" is enough.
      NodesAndElementOfFace const & f0 = lowestNodeToFaces( nodeID, j );
      insertFaceToNodesEntry( curFaceID, f0, faceToNodes );
      faceToElements( curFaceID, 0 ) = f0.element;
      faceToElements( curFaceID, 1 ) = -1; // TODO Make a constant

      // This is where we check for the two subsequent faces (when they exist).
      if( j < numFaces - 1 )
      {
        NodesAndElementOfFace const & f1 = lowestNodeToFaces( nodeID, j + 1 );
        if( f0 == f1 )
        {
          faceToElements( curFaceID, 1 ) = f1.element;
          ++j;
        }
      }
    }
```
Which is the same algorithm, with a slightly different implementation (Also I did not want to have `curFaceID` 
 and `for loop` index `j` leak out from the loop, it's always complicated to understand.). I do not know if this matters; we can set it back anyway.

Ohterwise, there may be some slight modification here and there, but nothing worth pinpointing I think.

src/coreComponents/mesh/generators/CellBlockABC.hpp
TotoGaz: Hi @klevzoff thx for the comments, I appreciate.

You’re right, I split the code and kept the types as they are right now.
This is indeed questionable for the long term.
There are three elements I’d like to mention as an answer.

- First, I noticed that there are some extra allocations here and there (see the usage of `CellBlockManagerABC::getEdgeMapOverallocation` alike function for example).
I like `CellBlock*` being considered as a raw geometric part of the code; thus I do not know if this is the proper location for having this extra allocations for computations.
(Also, we could define a policy about this overallocation in the input file not to impose the extra allocation for solvers which do not need it. But where in the code?)
This is somehow related to your point: how can we know which memory layout is the best for client code? There is surely a way, but what would be the cost ? (ease of use, readibility, maintenance, performance…)

- Second, imho the root cause of the issue you are pointing is that we are somehow mixing concepts (`CellBlockABC`, `CellBlockManagerABC`) and implementations (`array2d`, `ArraysOfArrays`) here.
For example `CellBLockABC::getElemToEdges` returns an `array2d< localIndex >` but the concept behind this “_matrix_” is in fact an “_element to edges mapping, as a matter of fact implemented under the form of an `array2d< localIndex >`_”.
In an ideal world we could use some kind of our `FixedOneToManyRelation` or `OrderedVariableToManyElementRelation`… which would be abstract too. The actual underlying storage could possibly be prescribed based on the solver, the user... If performances are at stake, maybe down casting the underlying data structure could be considered?

- Last but important point related to your architecture concern.
So far my split binds to GEOSX’s capabilities, which is `CellElementSubRegion` has one unique type of cells.
When we have crazy cells, we probably have to transform some of the core and solvers of GEOSX too; and the `CellBlock*ABC` will follow.
Maybe will we have `Heterogeneous` and `Homogeneous` `CellElementSubRegions`? _I do not know yet, that’s the critical upcoming question_.
And when this question is answered, I’m convinced that `CellBlock*ABC` will follow the same path consistently. Look: `CellBlockABC::numFacesPerElement()` has no meaning for non homogeneous sub-regions. This is even worse 😨 😱 

And [as Joel says](https://www.joelonsoftware.com/2002/11/11/the-law-of-leaky-abstractions/) “_All non-trivial abstractions, to some degree, are leaky_.” We have to find the best balance!

src/coreComponents/mesh/FaceManager.cpp
```diff
@@ -69,497 +70,76 @@ void FaceManager::resize( localIndex const newSize )
...
-  resize( numFaces );
-
-  populateMaps( elementManager,
-                facesByLowestNode.toViewConst(),
-                uniqueFaceOffsets,
-                m_toElements.m_toElementRegion,
-                m_toElements.m_toElementSubRegion,
-                m_toElements.m_toElementIndex,
-                nodeList() );
+  m_toElements.m_toElementIndex = cellBlockManager.getFaceToElements(); // TODO do better
```
AntoineMazuyer, line 141: Why do better ? Do you think of a better alternative ?

src/coreComponents/mesh/generators/CellBlockManager.hpp
```diff
@@ -0,0 +1,236 @@
...
+#include "mesh/generators/CellBlock.hpp"
+
+namespace geosx
+{
+
+/**
+ * @class CellBlockManager
+ * @brief The CellBlockManager class provides an interface to ObjectManagerBase in order to manage CellBlock data.
+ */
+class CellBlockManager : public CellBlockManagerABC
```
AntoineMazuyer, line 32: In the next PR when we will have a `CellBlockManager` dedicated to arbitrary polyhedral grids, we will have to define another name for that :)

src/coreComponents/mesh/generators/CellBlockManagerABC.hpp
```diff
@@ -0,0 +1,187 @@
...
+
+#include "dataRepository/Group.hpp"
+
+#include <map>
+
+namespace geosx
+{
+
+/**
+ * @brief Abstract base class for CellBlockManager.
```
AntoineMazuyer, line 26: I think we need to have more details here to what is a `CellBlockManager` in GEOSX

src/coreComponents/mesh/generators/CellBlockManagerABC.hpp
```diff
@@ -0,0 +1,187 @@
...
+   *
+   * @note The mapping is computed on the fly and returned. It is not stored in the instance.
+   */
+  virtual ArrayOfArrays< localIndex > getNodeToElements() const = 0;
+
+  /**
+   * @brief Returns the edge to nodes mapping.
+   * @return A 1 to 2 relationship. The result is meant to have size (numEdges, 2).
+   */
+  virtual array2d< geosx::localIndex > const & getEdgeToNodes() const = 0;
```
AntoineMazuyer, line 145: We already talk about that but I found kinda disturbing that we have sometime `array2d` and sometime `ArrayOfArrays`

src/coreComponents/mesh/generators/CellBlockABC.hpp
```diff
@@ -0,0 +1,131 @@
...
+
+  /**
+   * @brief Get the element-to-edges map.
+   * @return A view of the mapping.
```
AntoineMazuyer, line 86: is it a view ?

AntoineMazuyer (changes requested):
First reading, I will deep further into the code in the next days

src/coreComponents/mesh/generators/CellBlockABC.hpp
```diff
@@ -0,0 +1,131 @@
...
+ * Mainly element type related information (number of faces per element...),
+ * some element to nodes, faces and edges mappings, a local to global mapping and
+ * some kind of accessors to external properties.
+ *
+ * The cells of CellBlockABC are all of the same kind.
+ *
+ * It's noteworthy that the CellBlockABC is immutable oriented.
+ * The derived implementations need to have the modification/creation capabilities.
+ */
+class CellBlockABC : public dataRepository::Group
```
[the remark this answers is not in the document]
TotoGaz: Neither am I; and I would prefer a simple `CellBlock`.
You do not have to know that it's a base, that it's abstract, whatever. 
Just that's it's a `CellBlock`.

But I did like the discussion open for the team and I'd follow the decision.
With this `ABC` suffix I wanted to emphasize that this is another role w.r.t. what we are used to.

src/coreComponents/mainInterface/ProblemManager.cpp
```diff
@@ -491,7 +491,7 @@ void ProblemManager::generateMesh()
 
   MeshManager & meshManager = this->getGroup< MeshManager >( groupKeys.meshManager );
   meshManager.generateMeshes( domain );
-  Group & cellBlockManager = domain.getGroup( keys::cellManager );
+  CellBlockManagerABC & cellBlockManager = domain.getGroup< CellBlockManagerABC >( keys::cellManager );
```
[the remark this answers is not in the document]
TotoGaz: I think this could be interesting indeed.
To be noted that this induces a change of dependencies, since now `DomainPartition` does not depend on `CellBlockManager`. I'll have a look.
Also, during the review/merge phase, I wanted to reduce the footprint of the PR to minimize conflicts.

src/coreComponents/mesh/NodeManager.cpp
```diff
@@ -65,192 +65,99 @@ void NodeManager::resize( localIndex const newSize )
...
-        localIndex const nodeIndex = elemToNodeMap( k, a );
-        RAJA::atomicInc< parallelHostAtomic >( &elemsPerNode[ nodeIndex ] );
-      }
-    } );
-  } );
+  toElementList = cellBlockManager.getNodeToElements();
 
   // Resize the node to elem map.
   toElementRegionList.resize( 0 );
   toElementSubRegionList.resize( 0 );
```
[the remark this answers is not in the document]
TotoGaz: Actually something's happening lines 243/244/245?!

src/coreComponents/mesh/FaceManager.cpp
TotoGaz: Because for the moment I'm just copying data with `operator=` and maybe something better is possible with move semantics.

Also, this is surely possible to define an overallocation policy, and this is potentially the place to do this.
We can do it directly in the `CellBlockManager*` region too. To be discussed.

src/coreComponents/mesh/generators/CellBlockManager.hpp
TotoGaz: Fatalitas!

src/coreComponents/mesh/generators/CellBlockManagerABC.hpp
TotoGaz: Absolutely.

src/coreComponents/mesh/generators/CellBlockManagerABC.hpp
TotoGaz: Here, it's kind of simple to have on edge with two nodes. Hence the `array2d` providing a fixed size in both dimensions.

You may be interested by https://github.com/GEOSX/GEOSX/pull/1418#discussion_r631991811

src/coreComponents/mesh/generators/CellBlockABC.hpp
TotoGaz: This is part of the elements that I need to check.

Do I want to return a view and copy it?
If so, I could have the overallocation after.

Do I want to `move` the data (no more a view I think), and then the overallocation policy could be done in the very beginning.

The way how to return the vectors needs to be discussed in a next meeting. 😉

src/coreComponents/mesh/NodeManager.cpp
```diff
@@ -65,192 +65,99 @@ void NodeManager::resize( localIndex const newSize )
...
 
-
-void NodeManager::setElementMaps( ElementRegionManager const & elementRegionManager )
+void NodeManager::setElementMaps( CellBlockManagerABC const & cellBlockManager, ElementRegionManager const & elementRegionManager )
 {
+  const localIndex totalNodeElems = cellBlockManager.numNodes();
+
   GEOSX_MARK_FUNCTION;
 
   ArrayOfArrays< localIndex > & toElementRegionList = m_toElements.m_toElementRegion;
```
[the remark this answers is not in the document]
TotoGaz: done

src/coreComponents/mesh/NodeManager.cpp
```diff
@@ -65,192 +65,99 @@ void NodeManager::resize( localIndex const newSize )
...
-                                                            ( localIndex const er, localIndex const esr, ElementRegionBase const &,
-                                                            CellElementSubRegion const & subRegion )
+  elementRegionManager.forElementSubRegionsComplete< CellElementSubRegion >(
+    [&toElementRegionList, &toElementSubRegionList] ( localIndex const er, localIndex const esr, ElementRegionBase const &, CellElementSubRegion const & subRegion )
   {
     arrayView2d< localIndex const, cells::NODE_MAP_USD > const elemToNodeMap = subRegion.nodeList();
     for( localIndex k = 0; k < subRegion.size(); ++k )
     {
-      for( localIndex a=0; a<subRegion.numIndependentNodesPerElement(); ++a )
+      for( localIndex a = 0; a < subRegion.numNodesPerElement(); ++a )
```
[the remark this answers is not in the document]
TotoGaz: done

src/coreComponents/mesh/NodeManager.cpp
```diff
@@ -65,192 +65,99 @@ void NodeManager::resize( localIndex const newSize )
...
-  // If this becomes a bottleneck create a temporary ArrayOfArrays of tuples and insert into that first then copy over.
-  elementRegionManager.
-    forElementSubRegionsComplete< CellElementSubRegion >( [&toElementRegionList, &toElementSubRegionList, &toElementList]
-                                                            ( localIndex const er, localIndex const esr, ElementRegionBase const &,
-                                                            CellElementSubRegion const & subRegion )
+  elementRegionManager.forElementSubRegionsComplete< CellElementSubRegion >(
+    [&toElementRegionList, &toElementSubRegionList] ( localIndex const er, localIndex const esr, ElementRegionBase const &, CellElementSubRegion const & subRegion )
   {
     arrayView2d< localIndex const, cells::NODE_MAP_USD > const elemToNodeMap = subRegion.nodeList();
     for( localIndex k = 0; k < subRegion.size(); ++k )
```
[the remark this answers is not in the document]
TotoGaz: done

src/coreComponents/mesh/CellElementSubRegion.cpp
```diff
@@ -42,24 +47,71 @@ CellElementSubRegion::CellElementSubRegion( string const & name, Group * const p
...
+    {
+      // Pyramid
+      m_numNodesPerElement = 5;
+      m_numEdgesPerElement = 8;
+      m_numFacesPerElement = 5;
+      break;
+    }
+    default:
+    {
+      GEOSX_ERROR( "Error.  Don't know what kind of element this is." );
```
francoishamon, line 92: Do we want to have a more explicit error message, with the name of the problematic subregion, etc?

src/coreComponents/mesh/CellElementSubRegion.hpp
```diff
@@ -16,24 +16,50 @@
...
+  using FaceMapType = FixedOneToManyRelation;
   /// Type of map between cell blocks and embedded elements
   using EmbSurfMapType = InterObjectRelation< ArrayOfArrays< localIndex > >;
 
+  /**
+   * @brief Const getter for the catalog name.
+   * @return the name of this type in the catalog
+   */
+  static const string catalogName()
+  { return "CellElementSubRegion"; } // FIXME I don't know what I am doing
```
francoishamon, line 55: Maybe remove the comment, because this return value looks fine?

src/coreComponents/mesh/CellElementSubRegion.hpp
```diff
@@ -247,13 +329,76 @@ class CellElementSubRegion : public CellBlock
...
+   * @brief Compute the volume of the k-th element in the subregion.
+   * @param[in] k the index of the element in the subregion
+   * @param[in] X an arrayView of (const) node positions
+   */
+  inline void calculateCellVolumesKernel( localIndex const k,
+                                          arrayView2d< real64 const, nodes::REFERENCE_POSITION_USD > const & X ) const
+  {
+    LvArray::tensorOps::fill< 3 >( m_elementCenter[ k ], 0 );
+
+    real64 Xlocal[10][3];
```
francoishamon, line 365: Does that contain an assumption on the max number of nodes per element? Do you think we could remove the hardcoded value of 10 and replace it with a `constexpr` defined at the top of the class to make the transition to polyhedral grids easier?

src/coreComponents/mesh/CellElementSubRegion.hpp
```diff
@@ -247,13 +329,76 @@ class CellElementSubRegion : public CellBlock
...
+        break;
+      }
+      case ElementType::Pyramid:
+      {
+        m_elementVolume[k] = computationalGeometry::PyramidVolume( Xlocal );
+        break;
+      }
+      default:
+      {
+        GEOSX_ERROR( "Volume calculation not supported for element type: " << m_elementType );
```
francoishamon, line 398: Maybe add a more explicit error message specifying the class and the name of the object?

src/coreComponents/mesh/EdgeManager.cpp
```diff
@@ -64,371 +66,29 @@ EdgeManager::~EdgeManager()
 
 void EdgeManager::resize( localIndex const newSize )
 {
+  // FIXME My tests say that this line could be commented out.
+  //       But those tests are not exhaustive.
```
francoishamon, line 70: Remove the comment if no longer useful, or write something more specific about what you mean

src/coreComponents/mesh/CellElementSubRegion.cpp
```diff
@@ -42,24 +47,71 @@ CellElementSubRegion::CellElementSubRegion( string const & name, Group * const p
 
 CellElementSubRegion::~CellElementSubRegion()
 {
-  // TODO Auto-generated destructor stub
+  // Left blank
 }
 
-void CellElementSubRegion::copyFromCellBlock( CellBlock & source )
+void CellElementSubRegion::setElementType( ElementType elementType )
```
francoishamon, line 53: Since this function is very similar to `CellBlock::setElementType`, is it really needed? Can you just directly copy the `m_elementType`, `m_numNodesPerElement`, `m_numEdgesPerElement`, `m_numFacesPerElement` from the `CellBlock` to the `CellElementSubRegion`?

src/coreComponents/mesh/generators/CellBlockManager.hpp
```diff
@@ -0,0 +1,234 @@
...
+
+  array1d< globalIndex > getNodeLocalToGlobal() const override;
+
+  /**
+   * @brief Returns a mutable reference to vector holding the node to global mapping.
+   * @return The reference
+   *
+   * @note This is meant to be used as a setter. Do not resize the vector yourself.
+   */
+  array1d< globalIndex > & getNodeLocalToGlobal();
```
francoishamon, line 99: Should we do something to prevent resizing? Like returning an `arrayView1d` instead of the reference to the `array1d`?

src/coreComponents/mesh/generators/CellBlockManagerABC.hpp
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  virtual ~CellBlockManagerABC() = default;
+
+  /**
+   * @brief Maximum number of edges allowed (in memory) per each node.
+   * @return The number as an integer.
+   */
+  static constexpr int maxEdgesPerNode()
+  { return 200; }
```
francoishamon, line 50: What is the motivation between putting `maxEdgesPerNode()` and `maxFacesPerNodes()` in two separate files (`CellBlockManagerABC.hpp` and `CellBlockManager.hpp`?

src/coreComponents/unitTests/constitutiveTests/testTriaxial_druckerPragerExtended.xml
```diff
@@ -60,4 +60,12 @@
       nz="{ 1 }"
       cellBlockNames="{ cellBlock01 }"/>
   </Mesh>
+
+  <ElementRegions>
+    <CellElementRegion
+      name="dummy"
+      cellBlocks="{ cellBlock01 }"
+      materialList="{ dummy }"/>
+  </ElementRegions>
```
francoishamon, line 69: Do we really want to initialize a `CellElementRegion` in the triaxial driver unit tests? I am not sure, but my impression was that we do not want to do that in this case?

francoishamon (approved):
This PR looks very good to me! Thanks @TotoGaz

src/coreComponents/mesh/generators/CellBlockManager.hpp
corbett5: An `array1d< globalIndex > const &` would do the trick too. But I feel like we probably shouldn't be modifying this outside of the class anyways and certainly 95% of the use cases are read only so I'd be inclined to just make it an `arrayView1d< globalIndex const >`.

src/coreComponents/mesh/CellBlock.cpp
```diff
@@ -1,392 +0,0 @@
-/*
```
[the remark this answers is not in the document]
TotoGaz: Surely a `merge` error on my side. Removing.

src/coreComponents/mesh/generators/CellBlockManager.hpp
TotoGaz: For this precise case, the original code from `ObjectManagerBase` does
```c++
  arrayView1d< globalIndex > localToGlobalMap()
  { return m_localToGlobalMap; }
```
The pattern (from `InternalMeshGenerator` and `PAMELAMeshGenerator`) is to fill it through `arrayView1d< globalIndex > const &` which I can reproduce.

But more important is the API at `CellBlockManagerABC`. This is actually the second tick box of this PR description (still unticked 😉 ): "_The way to return the arrays must be improved (I’ll need help here)._".

My way of returning the arrays is quite inconsistent, and therefore surely incorrect at some point.

I think (correct me please) that simply returning a `array1d< globalIndex >` allows `move` semantics which was a nice effect (while far from being mandatory, specially if we deallocate the `CellBlockManager` instance). But what about `const` correctness?

src/coreComponents/mesh/generators/CellBlockManagerABC.hpp
TotoGaz: `maxEdgesPerNode` is called from `EdgeManager` and `CellBlock*` stuff. That is, both sides of the abstraction.
While `maxFacesPerNode` is only called from `CellBlockManager`.
I wanted to reduce the API to the minimum, but it's true that some symmetry is sacrificed here. What's your opinion? Would you prefer symmetry back?

However, a followup is surely to build an allocation policy, so wave equation do not have to pay the extra cost of our tricky reservoir grids. When we do this (when?... never? 😉 ), this will move!

src/coreComponents/unitTests/constitutiveTests/testTriaxial_druckerPragerExtended.xml
TotoGaz: The problem here is that we do have an `InternalMesh` that builds one unique  `cellblock` named `cellBlock01`.
The new design break the _generator_ part (`InternalMesh` and `CellBlock*`) from the `ElementRegions` `CellElement(Sub)Region`. And it happens that `geosx` is not working without `ElementRegions`/`CellElementRegion`.

It happened that I had to force the creation.

But I kind of prefer having one dummy `CellElementRegion` receiving the result of a dummy (too) generator `InternalMesh`/`CellBlock` that are now deleted, than keeping the dummy generators in the air.

PS: look at line 50 and do not worry, somebody will resole this soon 😉 !

TotoGaz (commented):
Thx @francoishamon !

src/coreComponents/mesh/CellBlock.cpp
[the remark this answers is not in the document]
TotoGaz: done

src/coreComponents/mesh/CellElementSubRegion.cpp
TotoGaz: I was wrong, since our `ElementType` does not distinguish between 1, 2 and 3 dimensional elements (for edges, faces, cells). It's therefore legit not to consider all the cases in this switch.
Maybe splitting `ElementType` to consider those dimensions could be a move.

src/coreComponents/mesh/CellElementSubRegion.hpp
TotoGaz: done

src/coreComponents/mesh/EdgeManager.cpp
TotoGaz: done, it's still useful!

src/coreComponents/mesh/CellElementSubRegion.cpp
TotoGaz: I simplified.

src/coreComponents/mesh/CellElementSubRegion.hpp
TotoGaz: done

src/coreComponents/mesh/generators/CellBlockABC.hpp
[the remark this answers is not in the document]
TotoGaz: I've open discussion #1555 about this question.

src/coreComponents/mainInterface/ProblemManager.cpp
```diff
@@ -509,17 +509,19 @@ void ProblemManager::generateMesh()
 
       GeometricObjectManager & geometricObjects = this->getGroup< GeometricObjectManager >( groupKeys.geometricObjectManager );
 
+      nodeManager.setNodesInformation( cellBlockManager );
+
```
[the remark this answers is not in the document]
TotoGaz: done

AntoineMazuyer (approved):

src/coreComponents/mainInterface/ProblemManager.cpp
```diff
@@ -511,17 +511,22 @@ void ProblemManager::generateMesh()
 
       GeometricObjectManager & geometricObjects = this->getGroup< GeometricObjectManager >( groupKeys.geometricObjectManager );
 
+      // Nodes
+      nodeManager.setNodesInformation( cellBlockManager );
       MeshUtilities::generateNodesets( geometricObjects, nodeManager );
```
CusiniM, line 516:
```suggestion
      MeshUtilities::generateNodeSets( geometricObjects, nodeManager );
```

TotoGaz: Do you mean that I should correct the typo?
(I'm OK, just to be sure to fully understand your patch/comment)

CusiniM: ah, I guess you ll have to change the name of the function as well. Just coz everywhere else we use `nodeSets`. Not a particularly useful comment :)

TotoGaz: It's OK, let's take all the improvements everywhere!

PR #1426, opened 2021-05-15
Cleanup TractionBoundaryCondition

klevzoff:
I was annoyed by a bogus warning:
```
m_inputStress is specified, but m_tractionType!=2 so value of m_inputStress is unused.
```
Turned out the fix was actually in `xmlWrapper` (and already made in https://github.com/GEOSX/GEOSX/pull/1424), but while I was looking for it, I cleaned up input validation in `TractionBoundaryCondition` (along with `EnumStrings`... which caused a few other file changes). This is a small change, but I don't have any other current PR to sneak it into :)

In particular, this PR replaces integer values of `tractionType` with enum mnemonics that are easier (?) to remember:
 * `vector` (previously 0)
 * `normal` (previously 1)
 * `stress` (previously 2)

Let me know you have better names in mind... or if you'd actually prefer to keep integers.


rrsettgast (approved):
Looks good to me!! Thanks!!

Issue #1421, opened 2021-05-06
[Bug] Compilation broken when GEOSX_BUILD_OBJ_LIBS=OFF


klevzoff:
**Describe the bug**
I tried turning object libraries off (since they slow down a full build considerably for me) and got an early build error.

**To Reproduce**
1. Add `set(GEOSX_BUILD_OBJ_LIBS OFF CACHE BOOL "")` to your host-config
2. Configure a clean build
3. try `make common`
4. Observe the error:
```
[ 50%] Building CXX object coreComponents/common/CMakeFiles/common.dir/initializeEnvironment.cpp.o
In file included from /home/klevtsov/work/GEOSX-develop/src/coreComponents/common/initializeEnvironment.cpp:15:
/home/klevtsov/work/GEOSX-develop/src/coreComponents/common/initializeEnvironment.hpp:24:10: fatal error: adiak.hpp: No such file or directory
 #include <adiak.hpp>
          ^~~~~~~~~~~
compilation terminated.

```

**Expected behavior**
Should build with object libraries disabled.

**Platform (please complete the following information):**
- Machine: my laptop (ubuntu 20.04)
- Compiler: gcc-8.4.0 (but it's compiler independent I think)
- GEOSX version: most recent `develop` (`4ad3852`)

**Additional context**
Here's what compilation line for `initializeEnvironment.cpp` looks like with object libraries **enabled** (I isolated just the includes)
```
-I/home/klevtsov/work/GEOSX-develop/build-gcc8-debug/include
-I/home/klevtsov/work/thirdPartyLibs/install-gcc8-debug/pugixml/include
-I/home/klevtsov/work/thirdPartyLibs/install-gcc8-debug/conduit/include
-I/home/klevtsov/work/thirdPartyLibs/install-gcc8-debug/conduit/include/conduit
-I/home/klevtsov/work/GEOSX-develop/src/coreComponents
-isystem /home/klevtsov/work/thirdPartyLibs/install-gcc8-debug/raja/include
-isystem /home/klevtsov/work/thirdPartyLibs/install-gcc8-debug/chai/include
-isystem /home/klevtsov/work/thirdPartyLibs/install-gcc8-debug/caliper/include
-isystem /opt/openmpi/openmpi-4.1.0-gcc-8.4.0/include
-isystem /home/klevtsov/work/thirdPartyLibs/install-gcc8-debug/adiak/include 
```
and **disabled**
```
[...]

Issue #1389, opened 2021-04-13
[Cleanup] Collapse all the mpi, cuda, openmp, etc dependency inclusions at the lowest level possible, and remove from all upper level components.


rrsettgast:
**Describe the issue**
There are redundant dependency includes of some environment and TPL components. this looks like:
```
if ( ENABLE_MKL )
  set( dependencyList ${dependencyList} mkl )
endif()

if( ENABLE_OPENMP )
  set( dependencyList ${dependencyList} openmp )
endif( )

if( ENABLE_CUDA )
  set( dependencyList ${dependencyList} cuda )
endif( )

if( ENABLE_MPI )
  set( dependencyList ${dependencyList} mpi )
endif()

if( ENABLE_CALIPER )
  set( dependencyList ${dependencyList} caliper adiak )
endif()
```

**Proposed cleanup**
These dependencies should be inherited. I.e. if something is dependent on `common`, then it should have these already. Remove them all if dependency components include these in their dependencies.

PR #1433, opened 2021-05-20
CMake build cleanup

bmhan12:
This PR:
- Fixes compilation when building with GEOSX_BUILD__OBJ_LIBS=OFF (Relates to #1421)
- Removes repetitive mpi, openmp, etc. dependencies from higher level components (Relates to #1389)
  - cuda dependency could not be removed, related issue posted at llnl/blt#490
- Added a script `circular_dependencies.sh` that outputs:
  -  a matrix of coreComponents and if they depend on another coreComponent (dependency_matrix.txt)
  - any circular dependencies and the header includes between the two circular coreComponents (circular_dependencies.txt)

### DIFF
--- a/scripts/circular_dependencies.sh
+++ b/scripts/circular_dependencies.sh
@@ -1,0 +1,99 @@
+#!/bin/bash
+
+# Writes to a dependency_matrix.txt file in the current directory.
+# Each row has the component followed by a "yes" or "no" if the component
+# has a header/dependency to the other component named at the top of the 
+# column.
+# Tab delimited, can be exported to Excel (Data --> From Test/CSV)
+
+declare -ar coreComponents=(
+ common
+ codingUtilities
+ dataRepository
+ schema
+ functions
+ constitutive
+ mesh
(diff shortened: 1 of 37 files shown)


TotoGaz:
Hi @bmhan12 I've pushed an additional python scripts that uses the information provided directly by the `gcc` preprocessor to build the include graph and find the cycles.

One interest is that it uses the preprocessing options to precisely find the proper includes.

Then, it happens `coreComponents/fieldSpecification/FieldSpecificationManager.hpp` is including itself! That's some big cycle there! 😉 

Another interest is that we could also use the python module `networkx` (which is used to find the cycles) to export some graph and analyze it using, _e.g._ [gephi](https://gephi.org/).

TotoGaz (approved):
Great cleaning @bmhan12 !

sytuannguyen:
@bmhan12 You should profit this PR to add the missing header files in CMakeLists, there are many of them, PhaseFieldDamageFEMKernels.hpp as an example.

PR #1441, opened 2021-05-24
Corbett/hypre cuda fix

corbett5:


src/coreComponents/linearAlgebra/interfaces/hypre/HypreMatrix.cpp
```diff
@@ -162,38 +162,45 @@ void HypreMatrix::createWithGlobalSize( globalIndex const globalRows,
...
-  array1d< HYPRE_Int > sizes( localMatrix.numRows() );
-  array1d< HYPRE_Int > offsets( localMatrix.numRows() );
-  localIndex const * const pSizes = localMatrix.getSizes();
-  localIndex const * const pOffsets = localMatrix.getOffsets();
+  array1d< globalIndex > rows;
+  rows.resizeWithoutInitializationOrDestruction( LvArray::MemorySpace::cuda, localMatrix.numRows() );
+  arrayView1d< globalIndex > const rowsV = rows;
 
-  // Note: this relies on the fact the localMatrix has correct sparsity on the host
-  // (e.g. it was built or synced there) because we don't move sizes/offsets to host here.
```
corbett5, line 185: The bug is that not only is `localMatrix` expected to be current (at least the sizes and offsets) on host, but it's also supposed to _be_ on host. The change from `setValues` to `zero` meant that `localMatrix` itself is on device so the pointers are device pointers.

rrsettgast (approved):

klevzoff (approved):

PR #1445, opened 2021-05-27
Adds support for hypre-cuda via spack.

corbett5:


scripts/uberenv/spack_configs/config.yaml
```diff
@@ -23,8 +23,8 @@ config:
   template_dirs:
     - $spack/templates
 
-  # install directory layout
-  install_path_scheme: "${ARCHITECTURE}-${COMPILERNAME}@${COMPILERVER}/${PACKAGE}@${VERSION}"
```
rrsettgast, line 27: should this be commented out?

rrsettgast (approved):

scripts/uberenv/spack_configs/config.yaml
corbett5: Yes the default install path scheme is better, but I want to keep here as "documentation".

PR #1455, opened 2021-06-04
Creation of a CoupleSolidModel to handle Poro-Solid-Damage models.

CusiniM:
This PR is related to https://github.com/GEOSX/GEOSX/discussions/1422

Rebaseline: https://github.com/GEOSX/integratedTests/pull/141

Integrated tests work on Lassen.

In my opinion the design works quite well. I have added the permeability models to this PR although they will only be used in #1311.  
In a few words, I have created a base class template called `CoupledSolid` which owns method to access 
- a porosity model
- a solid model
- a permeability model.

```c++
template< typename SOLID_TYPE, typename PORO_TYPE, typename PERM_TYPE >
class CoupledSolid : public ConstitutiveBase
{
public:
        ...
protected:
  SOLID_TYPE const & getSolidModel() const
  { return this->getParent().template getGroup< SOLID_TYPE >( m_solidModelName ); }

  PORO_TYPE const & getPorosityModel() const
  { return this->getParent().template getGroup< PORO_TYPE >( m_porosityModelName ); }

  PERM_TYPE const & getPermModel() const
  { return this->getParent().template getGroup< PERM_TYPE >( m_permeabilityModelName ); }

  // the name of the solid model
  string m_solidModelName;

  // the name of the porosity model
  string m_porosityModelName;
  
  // the name of the permeability model
  string m_permeabilityModelName;
 };
 ```
Then, I have created 2 specialized classes:

- `PorousSolid` which always has a `BiotPorosity` and is used solely for Poromechanics
- `CompressibleSolid` which can have various porosity models and, for now, assumes a nullModel for the solid Model so that all mechanics data (e.g. stresses) are not allocated. This model is what we would use for simulations without mechanics. 

For now, the damage can stay as it is ( I think) and the SOLID_TYPE  can simply be a Damage<ElasticIsotropic> etc...

Thus both the `PorousSolid` and the `CompressibleSolid` would be

```c++
template< typename SOLID_TYPE >
class PorousSolid : public CoupledSolid< SOLID_TYPE, BiotPorosity, StrainDependentPermeability >

template< typename PORO_TYPE,  typename PERM_TYPE >
class CompressibleSolid : public CoupledSolid< NullModel, PORO_TYPE, PERM_TYPE >
```
I am registering a `PorousSolid` for each type of `SolidModel` so that the proper solid model gets instantiated based on the input, i.e.

```xml
</PoroElasticIsotropic
   name="porousSolidName"
   solidName="solidModelName"
   porosityName="poroName"
   permeabilityName="permName"
   />
   
</ElasticIsotropic
  name="solidName"
  ...
   />
   
 </BiotPorosity
  name="poroName"
  ...
   />
  </StrainDependentPermeability
  name="permName"
  ...
   />
````
will automatically instantiate a `PorousSolid< ElasticIsotropic > ` which will be a `CoupledSolid<ElasticIsotropic, BiotPorosity, StrainDependentPermeability>`.  

The same thing could be done for the `CompressibleSolid` with various porosity and permeability models (I think special model could be used for very specific physics, e.g., chemical reactions that change the porosity).

### CODE REVIEW

File: src/coreComponents/constitutive/solid/ElasticIsotropic.hpp [hunk 1]
```diff
@@ -126,6 +126,13 @@ class ElasticIsotropicUpdates : public SolidBaseUpdates
                                  localIndex const q,
                                  real64 ( &elasticStrain )[6] ) const override final;
 
+  GEOSX_HOST_DEVICE
+  virtual real64 getBulkModulus( localIndex const k ) const override final
```
TotoGaz, line 130: Just `final`.

src/coreComponents/constitutive/solid/PorousSolid.cpp
```diff
@@ -0,0 +1,66 @@
...
+{
+
+template< typename SOLID_TYPE >
+PorousSolid< SOLID_TYPE >::PorousSolid( string const & name, Group * const parent ):
+  CoupledSolid< SOLID_TYPE, BiotPorosity >( name, parent )
+{}
+
+template< typename SOLID_TYPE >
+PorousSolid< SOLID_TYPE >::~PorousSolid()
+{}
```
TotoGaz, line 46: Same, could you use `default` impl?

src/coreComponents/constitutive/solid/PorousSolid.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+// Register all PorousSolid model types.
+typedef PorousSolid< ElasticIsotropic > PorousElasticIsotropic;
+typedef PorousSolid< ElasticTransverseIsotropic > PorousElasticTransverseIsotropic;
+typedef PorousSolid< DruckerPrager > PorousDruckerPrager;
+typedef PorousSolid< DruckerPragerExtended > PorousDruckerPragerExtended;
+typedef PorousSolid< Damage< ElasticIsotropic > > PorousDamageElasticIsotropic;
+typedef PorousSolid< DamageSpectral< ElasticIsotropic > > PorousDamageSpectralElasticIsotropic;
+typedef PorousSolid< DamageVolDev< ElasticIsotropic > > PorousDamageVolDevElasticIsotropic;
```
TotoGaz, line 55: Same, do you need those typedefs?

src/coreComponents/constitutive/solid/porosity/PorosityBase.hpp
```diff
@@ -0,0 +1,205 @@
...
+  virtual void allocateConstitutiveData( dataRepository::Group & parent,
+                                         localIndex const numConstitutivePointsPerParentIndex ) override;
+
+  static string catalogName() { return "PorosityBase"; }
+
+  virtual string getCatalogName() const override { return catalogName(); }
+
+  struct viewKeyStruct : public ConstitutiveBase::viewKeyStruct
+  {
+    static constexpr char const * newPorosityString() { return "porosity"; }
```
TotoGaz, line 136: This should not not be `newPorosityString` but `porosityString` to stay consistent.

src/coreComponents/constitutive/solid/porosity/PressurePorosity.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+PressurePorosity::~PressurePorosity() = default;
+
+std::unique_ptr< ConstitutiveBase >
+PressurePorosity::deliverClone( string const & name,
+                                Group * const parent ) const
+{
+  std::unique_ptr< ConstitutiveBase > clone = ConstitutiveBase::deliverClone( name, parent );
+
+  return clone;
```
TotoGaz, line 51: direct `return`

src/coreComponents/constitutive/solid/porosity/PressurePorosity.hpp
```diff
@@ -0,0 +1,147 @@
...
+    return KernelWrapper( m_newPorosity,
+                          m_oldPorosity,
+                          m_dPorosity_dPressure,
+                          m_referencePorosity,
+                          m_referencePressure,
+                          m_compressibility );
+  }
+
+
+protected:
```
TotoGaz, line 132: Since no class inherit from `PressurePorosity`, use `private` instead.

src/coreComponents/constitutive/solid/CoupledSolid.cpp
```diff
@@ -0,0 +1,29 @@
...
+#include "CoupledSolid.hpp"
+
+namespace geosx
+{
+
+using namespace dataRepository;
+
+namespace constitutive
+{}
+} /* namespace geosx */
```
TotoGaz, line 29: Do you really need this file?

TotoGaz (commented):
Nothing really smart, just small C++ comments, nothing about architecture.

src/coreComponents/constitutive/solid/CoupledSolid.hpp
```diff
@@ -0,0 +1,188 @@
...
+
+  /// Deleted move assignment operator
+  CoupledSolidUpdates & operator=( CoupledSolidUpdates && ) =  delete;
+
+  GEOSX_HOST_DEVICE
+  GEOSX_FORCE_INLINE
+  typename SOLID_TYPE::KernelWrapper const getSolidModel() const
+  {
+    return m_solidUpdate;
+  }
```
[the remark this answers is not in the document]
CusiniM: no, not yet but because the `CompressibleSolid` is not used yet. I think it will be useful. I will remove it if not.

src/coreComponents/constitutive/solid/CoupledSolid.hpp
```diff
@@ -0,0 +1,188 @@
...
+
+  GEOSX_HOST_DEVICE
+  GEOSX_FORCE_INLINE
+  typename PORO_TYPE::KernelWrapper const getPorosityModel() const
+  {
+    return m_porosityUpdate;
+  }
```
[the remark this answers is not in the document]
CusiniM: I think the getter will be need in my other PR once I start using the compressible solid in the flow solvers but what do you mean by "not exposing so much" ? You mean just fwd each method that is needed? I will probably need a few more getters in the flow solvers which is why I thought it would be good to give access to the object but I can certainly change it.

src/coreComponents/constitutive/solid/PorousSolid.cpp
CusiniM: I really think I do. It's to register each type in the catalog so that they can be defined in the xml.

src/coreComponents/constitutive/solid/CoupledSolid.cpp
CusiniM: most llikely not.

src/coreComponents/constitutive/solid/porosity/PorosityBase.hpp
CusiniM: I followed what was done for the stress but I am happy to change it.

src/coreComponents/constitutive/permeability/permeabilitySelector.hpp
```diff
@@ -0,0 +1,69 @@
...
+namespace constitutive
+{
+
+#ifndef PASSTHROUGH_HANDLE_CASE
+#define PASSTHROUGH_HANDLE_CASE( MODEL ) \
+  if( dynamicCast< MODEL * >( &perm ) ) \
+  { \
+    lambda( static_cast< MODEL & >( perm ) ); \
+    return true; \
+  }
```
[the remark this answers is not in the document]
CusiniM: so, I created this in the other PR in which the permeability models are already used. I think we will need a dispatch but it really depends on what the answer to my question about the permeability is. It may very well be that the entire file is useless.

src/coreComponents/constitutive/solid/CoupledSolid.hpp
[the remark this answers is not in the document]
TotoGaz: You should maybe remove it now (or at least comment the code).
It will be easy to put it back when you need it.
Because it's really possible that you forget to remove it in the end if useless.

src/coreComponents/constitutive/solid/CoupledSolid.hpp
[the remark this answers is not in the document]
TotoGaz: "_Not exposing too much_" means that you give access to the entire `PORO_TYPE::KernelWrapper` while you only use one single function. It's then more likely that the devs using your code may use the `kernelWrapper` in ways that break your logic. After all, they can 😉

src/coreComponents/constitutive/permeability/permeabilitySelector.hpp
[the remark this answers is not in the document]
TotoGaz: Actually, the `PASSTHROUGH_HANDLE_CASE` macro seems unused? Am I right?

src/coreComponents/constitutive/permeability/PoroMechanicsPermeabilityKernel.hpp
```diff
@@ -0,0 +1,149 @@
...
+
+template< typename SUBREGION_TYPE,
+          typename FE_TYPE >
+class PoroMechanicsPermeabilityKernel
+{
+public:
+
+  static constexpr int numNodesPerElem = FE_TYPE::numNodes;
+
+  static constexpr int numQuadraturePointsPerElem = FE_TYPE::numQuadraturePoints;
```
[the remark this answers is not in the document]
CusiniM: If we do what I suggested for the permeabiity model we probably won't even need this file.

src/coreComponents/constitutive/solid/CompressibleSolid.cpp
```diff
@@ -0,0 +1,46 @@
...
+{}
+
+template< typename PORO_TYPE >
+CompressibleSolid< PORO_TYPE >::~CompressibleSolid()
+{}
+
+// Register all CoupleSolid model types.
+typedef CompressibleSolid< PressurePorosity > CompressibleRock;
+
+REGISTER_CATALOG_ENTRY( ConstitutiveBase, CompressibleRock, string const &, Group * const )
```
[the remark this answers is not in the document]
CusiniM: yes, I just followed what was done in other files in the code. If you think it's better without the `typedef` I can certainly remove it both in this file and in the other.

src/coreComponents/constitutive/solid/CoupledSolid.hpp
```diff
@@ -0,0 +1,188 @@
...
+
+  registerWrapper( viewKeyStruct::porosityModelNameString(), &m_porosityModelName ).
+    setInputFlag( dataRepository::InputFlags::REQUIRED ).
+    setDescription( "Name of the porosity model." );
+}
+
+template< typename SOLID_TYPE,
+          typename PORO_TYPE >
+CoupledSolid< SOLID_TYPE, PORO_TYPE >::~CoupledSolid()
+{}
```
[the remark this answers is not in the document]
CusiniM: Sure, although it would be interesting to know why it was done for a lot of the classes in geosx. If we should always use default then we should probably to a proper clean up.

src/coreComponents/constitutive/permeability/permeabilitySelector.hpp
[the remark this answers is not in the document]
klevzoff: It's a remnant of older switch code we had in fluid models. It was replaced by `ConstitutivePassThruHandler<>` template (which I recommend using, unless you like typing if/else branches) and was left in `relativePermeabilitySelector.hpp` by accident.

src/coreComponents/constitutive/solid/PorousSolid.cpp
CusiniM: I have really just followed an existing pattern. It may very well be that it's not necessary though so if that's the case we should clean it up everywhere.

src/coreComponents/constitutive/permeability/CarmanKozenyPermeability.hpp
```diff
@@ -0,0 +1,168 @@
...
+  CarmanKozenyPermeabilityUpdate( CarmanKozenyPermeabilityUpdate const & ) = default;
+
+  /// Default move constructor
+  CarmanKozenyPermeabilityUpdate( CarmanKozenyPermeabilityUpdate && ) = default;
+
+  /// Deleted copy assignment operator
+  CarmanKozenyPermeabilityUpdate & operator=( CarmanKozenyPermeabilityUpdate const & ) = delete;
+
+  /// Deleted move assignment operator
+  CarmanKozenyPermeabilityUpdate & operator=( CarmanKozenyPermeabilityUpdate && ) = delete;
```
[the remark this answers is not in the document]
klevzoff: I'm actually deleting these in all fluid models in another PR. There is no real reason to disallow assignment for these (not that anyone ever does it), rule of 0 works just fine.

src/coreComponents/constitutive/solid/ElasticIsotropic.hpp
CusiniM: should I remove `override`?

src/coreComponents/constitutive/solid/CompressibleSolid.cpp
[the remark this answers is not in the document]
klevzoff: It won't work without a typedef (although I'd prefer `using`), because the derived type name is used by `REGISTER_CATALOG_ENTRY` macro as part of the name of static variable it creates.

src/coreComponents/constitutive/solid/CoupledSolid.hpp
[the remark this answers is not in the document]
klevzoff: No real reason & I'm replacing these with `= default` for fluid models. Most models can also place `= default` directly in the class declaration.

src/coreComponents/constitutive/solid/CoupledSolid.hpp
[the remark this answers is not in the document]
CusiniM: sounds good.

src/coreComponents/constitutive/solid/CoupledSolid.hpp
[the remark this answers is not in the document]
CusiniM: I see, makes sense. I will only expose the getters then.

src/coreComponents/constitutive/permeability/CarmanKozenyPermeability.hpp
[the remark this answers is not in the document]
CusiniM: Should I start getting rid of them as well?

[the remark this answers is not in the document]
klevzoff: Yeah, I'd recommend at least not adding them to new code. We have enough boilerplate as it is :)

src/coreComponents/constitutive/solid/ElasticIsotropic.hpp
TotoGaz: Hmm, maybe not. You may need `override` indeed.

src/coreComponents/constitutive/solid/CoupledSolid.hpp
```diff
@@ -57,18 +57,16 @@ class CoupledSolidUpdates
...
 
-  GEOSX_HOST_DEVICE
-  GEOSX_FORCE_INLINE
-  typename PORO_TYPE::KernelWrapper const getPorosityModel() const
+  real64 getPorosity( localIndex const k,
+                      localIndex const q ) const
   {
-    return m_porosityUpdate;
+    return m_porosityUpdate.getPorosity( k, q);
   }
```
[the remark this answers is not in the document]
CusiniM: which is probably why some tests have failed. Those macros are needed for cuda builds.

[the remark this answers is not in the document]
corbett5: Do you actually need `GEOSX_FORCE_INLINE`? I don't really think it should be used anywhere.

src/coreComponents/constitutive/permeability/CarmanKozenyPermeability.hpp
```diff
@@ -0,0 +1,154 @@
...
+
+#include "constitutive/permeability/PermeabilityBase.hpp"
+
+
+namespace geosx
+{
+namespace constitutive
+{
+
+class CarmanKozenyPermeabilityUpdate : public PermeabilityBaseUpdate
```
sytuannguyen, line 30: I propose to use KozenyCarman instead of CarmanKozeny to make it more consistent with literature.

src/coreComponents/constitutive/permeability/ParallelPlatesPermeability.hpp
```diff
@@ -0,0 +1,125 @@
...
+
+#include "constitutive/permeability/PermeabilityBase.hpp"
+
+
+namespace geosx
+{
+namespace constitutive
+{
+
+class ParallelPlatesPermeabilityUpdate : public PermeabilityBaseUpdate
```
sytuannguyen, line 30: I propose to call it "conductivity" or "fluid conductivity" instead of permeability.

src/coreComponents/constitutive/permeability/ParallelPlatesPermeability.hpp
```diff
@@ -0,0 +1,125 @@
...
+
+
+GEOSX_HOST_DEVICE
+void ParallelPlatesPermeabilityUpdate::compute( real64 const & effectiveAperture,
+                                                arraySlice1d< real64 > const & permeability,
+                                                arraySlice1d< real64 > const & dPerm_dAperture ) const
+{
+  // Technically this is not a permeability but it's convenient to definite like this.
+  real64 const perm  = effectiveAperture*effectiveAperture*effectiveAperture / 12.0;
+  real64 const dPerm = 3*effectiveAperture*effectiveAperture / 12.0;
```
sytuannguyen, line 109:
```suggestion
  real64 const dPerm = effectiveAperture*effectiveAperture / 4.0;
```

src/coreComponents/constitutive/permeability/ParallelPlatesPermeability.hpp
```diff
@@ -0,0 +1,125 @@
...
+  {} viewKeys;
+
+private:
+  array3d< real64 > m_dPerm_dAperture;
+
+};
+
+
+GEOSX_HOST_DEVICE
+void ParallelPlatesPermeabilityUpdate::compute( real64 const & effectiveAperture,
```
sytuannguyen, line 103: There are mechanical aperture and hydraulic aperture that are well distinguished in literature. I would propose to call this effectiveHydraulicAperture or simply hydraulicAperture

src/coreComponents/constitutive/permeability/ParallelPlatesPermeability.hpp
```diff
@@ -0,0 +1,125 @@
...
+                                                arraySlice1d< real64 > const & permeability,
+                                                arraySlice1d< real64 > const & dPerm_dAperture ) const
+{
+  // Technically this is not a permeability but it's convenient to definite like this.
+  real64 const perm  = effectiveAperture*effectiveAperture*effectiveAperture / 12.0;
```
sytuannguyen, line 108: We may want to extend this with consideration of fracture roughness and other microstructure parameters

src/coreComponents/constitutive/solid/porosity/BiotPorosity.hpp
```diff
@@ -0,0 +1,157 @@
...
+                       real64 const & deltaPressure,
+                       real64 const ( &strainIncrement )[6],
+                       real64 & dPorosity_dPressure,
+                       real64 & dPorosity_dVolStrain,
+                       real64 & dTotalStress_dPressure ) const
+  {
+    real64 const biotSkeletonModulusInverse = ( m_biotCoefficient[k] - m_referencePorosity[k] ) / m_grainBulkModulus;
+
+    real64 const porosity = m_oldPorosity[k][q] +
+                            +m_biotCoefficient[k] * LvArray::tensorOps::symTrace< 3 >( strainIncrement )
```
sytuannguyen, line 75: Do we really need to call LvArray for computing a simple math: strainIncrement[0] + strainIncrement[1] + strainIncrement[2]

src/coreComponents/constitutive/solid/porosity/BiotPorosity.hpp
```diff
@@ -0,0 +1,157 @@
...
+    dPorosity_dVolStrain = m_biotCoefficient[k];
+
+    savePorosity( k, q, porosity, biotSkeletonModulusInverse );
+
+    dTotalStress_dPressure = m_biotCoefficient[k];
+  }
+
+
+  GEOSX_HOST_DEVICE
+  void updateBiotCoefficient( localIndex const k,
```
sytuannguyen, line 89: In case of transversely isotropic material, the equivalent drained bulk modulus can be computed from the components of the drained stiffness tensor as: 

drainedBulkModulus = stiffness_ii_jj / 9 (with Einstein's summation for repeated indexes)

You can check this from a relevant reference of Castaneda and Willis 1995 in J. Mech. Phys. Solids (Eq. 4.13).

However, I think that it is proper to compute the Biot's tensor from the drained stiffness tensor. Then if we want to work with an approximated isotropic Biot's coefficient we can take the average of the Biot's tensor diagonal.

src/coreComponents/constitutive/solid/porosity/BiotPorosity.hpp
```diff
@@ -0,0 +1,157 @@
...
+                              real64 const bulkModulus ) const
```
sytuannguyen, line 90: the bulk modulus here is the drained bulk modulus (in isotropic case). I propose to name it as drainedBulkModulus to avoid any confusion

sytuannguyen (commented):
@CusiniM Just some reviews of the mathematical/physical aspects in the constitutive models. Hope it helps.

src/coreComponents/constitutive/permeability/CarmanKozenyPermeability.hpp
```diff
@@ -0,0 +1,154 @@
...
+
+GEOSX_HOST_DEVICE
+GEOSX_FORCE_INLINE
+void CarmanKozenyPermeabilityUpdate::compute( real64 const & porosity,
+                                              arraySlice1d< real64 > const & permeability,
+                                              arraySlice1d< real64 > const & dPerm_dPorosity ) const
+{
+  real64 const permValue = pow( m_sphericity*m_particleDiameter, 2 ) * pow( porosity, 3 )
+                           / (150 * pow( (1 - porosity), 2 ) );
+  real64 const dPerm_dPorValue = pow( m_sphericity*m_particleDiameter, 2 ) * pow( porosity, 3 );
```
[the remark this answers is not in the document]
sytuannguyen: Also, it is well-known that Kozeny-Carman is not relevant for rock. We may want to increase a little bit the bottom line by considering for example a relevant adaptation of Kozeny-Carman theory for rock by Mavko and Nur 1997 Geophysics?

src/coreComponents/constitutive/permeability/CarmanKozenyPermeability.hpp
CusiniM: I think my Petrophysics professor always used CarmanKozeny and that 's why I am used to that but I am fine with changing it.

src/coreComponents/constitutive/permeability/ParallelPlatesPermeability.hpp
CusiniM: I would keep this as is. It is a permeability model after all and it will be stored as permeability. It just makes it a lot easier to handle if it's the same field as for the other regions. I do not like that we are computing it as aperture^3 instead of aperture^2 but it makes it a lot more convenient to use it the it has to be used.

src/coreComponents/constitutive/permeability/ParallelPlatesPermeability.hpp
CusiniM: I agree but we would have to change it everywhere in the code. For now I would keep it consistent.

src/coreComponents/constitutive/solid/porosity/BiotPorosity.hpp
CusiniM: yes, whenever an LvArray function exists that's what you wanna use.

src/coreComponents/constitutive/solid/porosity/BiotPorosity.hpp
CusiniM: we had a long discussion yesterday about this. I think we have a plan but it will be done in a separate PR.

src/coreComponents/constitutive/solid/CompressibleSolid.hpp
```diff
@@ -0,0 +1,121 @@
...
+
+/**
+ * @brief Provides kernel-callable constitutive update routines
+ *
+ *
+ * @tparam SOLID_TYPE
+ */
+template< typename PORO_TYPE,
+          typename PERM_TYPE >
+class CompressibleSolidUpdates : public CoupledSolidUpdates< NullModel, PORO_TYPE, PERM_TYPE >
```
rrsettgast, line 39: Is PoreCompressibility a better name?

src/coreComponents/constitutive/solid/CoupledSolid.hpp
```diff
@@ -0,0 +1,180 @@
...
+  CoupledSolid( string const & name, dataRepository::Group * const parent );
+
+  /// Destructor
+  virtual ~CoupledSolid() override;
+
+  /**
+   * @brief Catalog name
+   * @return Static catalog string
+   */
+  static string catalogName() { return string( "Poro" ) + SOLID_TYPE::m_catalogNameString; }
```
rrsettgast, line 96: "Poro" should be changed to something else?

src/coreComponents/constitutive/solid/CoupledSolid.hpp
```diff
@@ -0,0 +1,180 @@
...
+  virtual string getCatalogName() const override { return catalogName(); }
+
+//  /// Post-process XML input
+//  virtual void postProcessInput() override;
+
+  struct viewKeyStruct
+  {
+    static constexpr char const * solidModelNameString() { return "solidName"; }
+    static constexpr char const * porosityModelNameString() { return "porosityName"; }
+    static constexpr char const * permeabilityModelNameString() { return "permeabilityName"; }
```
rrsettgast, line 111:
```suggestion
    static constexpr char const * solidModelNameString() { return "solidModelName"; }
    static constexpr char const * porosityModelNameString() { return "porosityModelName"; }
    static constexpr char const * permeabilityModelNameString() { return "permeabilityModelName"; }
```

src/coreComponents/constitutive/solid/ElasticIsotropic.hpp
rrsettgast: https://stackoverflow.com/questions/29412412/does-final-imply-override

src/coreComponents/constitutive/solid/PoroElastic.hpp
```diff
@@ -112,7 +112,7 @@ class PoroElastic : public BASE
    * @brief Catalog name
    * @return Static catalog string
    */
-  static string catalogName() { return string( "Poro" ) + BASE::m_catalogNameString; }
+  static string catalogName() { return string( "OldPoro" ) + BASE::m_catalogNameString; }
```
rrsettgast, line 115: This class will be deprecated in a follow on, or as part of this PR?

src/coreComponents/constitutive/solid/PorousSolid.hpp
```diff
@@ -0,0 +1,173 @@
...
+                          real64 const ( &strainIncrement )[6],
+                          real64 ( & stress )[6],
+                          real64 & dPorosity_dPressure,
+                          real64 & dPorosity_dVolStrain,
+                          real64 & dTotalStress_dPressure,
+                          DiscretizationOps & stiffness ) const
+  {
+    m_solidUpdate.smallStrainUpdate( k, q, strainIncrement, stress, stiffness );
+
+    updateBiotCoefficient( k );
```
rrsettgast, line 68: Using stiffness in a follow on PR..correct?

rrsettgast (approved):
This is a great step in the right direction. Excellent work.

src/coreComponents/constitutive/ConstitutivePassThru.hpp
```diff
@@ -126,70 +164,84 @@ struct ConstitutivePassThru< NullModel >
...
     {
       lambda( *ptr1 );
     }
-    else if( auto * const ptr2 = dynamic_cast< PoroElastic< DruckerPrager > * >( &constitutiveRelation ) )
+    else if( auto * const ptr2 = dynamic_cast< DamageVolDev< ElasticIsotropic > * >( &constitutiveRelation ) )
     {
       lambda( *ptr2 );
     }
-    else if( auto * const ptr3 = dynamic_cast< PoroElastic< ElasticIsotropic > * >( &constitutiveRelation ) )
+    else if( auto * const ptr3 = dynamic_cast< Damage< ElasticIsotropic > * >( &constitutiveRelation ) )
```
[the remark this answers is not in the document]
klevzoff: You can replace `if`/`else` with
```c++
#include "constitutive/ConstitutivePassThruHandler.hpp"
...
ConstitutivePassThruHandler< DamageSpectral< ElasticIsotropic >,
                             DamageVolDev< ElasticIsotropic >,
                             Damage< ElasticIsotropic > >::execute( constitutiveRelation, 
                                                                    std::forward< LAMBDA >( lambda ) );
```

Other than increased call stack depth when debugging, I don't think there's any drawbacks

src/coreComponents/constitutive/solid/PoroElastic.hpp
CusiniM: I am going to try to deprecate it in here. Let's see how many tests failures I get....

src/coreComponents/constitutive/solid/PorousSolid.hpp
CusiniM: yes

src/coreComponents/constitutive/solid/CoupledSolid.hpp
CusiniM: yes, it should. Although this is class is never instantiated so it does not really matter.

src/coreComponents/constitutive/solid/CompressibleSolid.hpp
CusiniM: I don't know. I wanted a name that included the fact that it also has a permeability model...

src/coreComponents/constitutive/permeability/ParallelPlatesPermeability.cpp
```diff
@@ -0,0 +1,55 @@
...
+namespace geosx
+{
+
+using namespace dataRepository;
+
+namespace constitutive
+{
+
+
+ParallelPlatesPermeability::ParallelPlatesPermeability( string const & name, Group * const parent ):
```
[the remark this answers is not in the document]
CusiniM: Do we prefer `FracturePermeability` ?

PR #1458, opened 2021-06-11
Fixed bug in compositional residual norm calculation with multiple regions

francoishamon:
This PR fixes a bug in the residual norm of the compositional solver (that I introduced I think, sorry). The `localResidualNorm` needs to be incremented for each subregion, but it was reset to zero for each subregion.

This PR requires the rebaseline of the staircase examples of the compositional solver because the change results in a small change in the number of nonlinear iterations.

Related to https://github.com/GEOSX/integratedTests/pull/135


klevzoff (approved):
We should probably make it just return the norm and do accumulation over subregions in the calling code (but no need to change this PR as this fix is also sufficient)

PR #1459, opened 2021-06-11
Consistify temperature units for CO2-brine input

francoishamon:
Until now, the CO2-brine model was expecting a temperature in degrees Celsius, while the solver takes a temperature in Kelvin. This PR changes the CO2-brine model input to Kelvin for consistency. The implementation of the CO2-brine remains the same.

The PR also updates the documentation of the CO2-brine model to reflect the change.

Does not require a rebaseline of the integrated tests.

### DIFF
--- a/src/coreComponents/constitutive/fluid/PVTFunctions/PVTFunctionHelpers.hpp
+++ b/src/coreComponents/constitutive/fluid/PVTFunctions/PVTFunctionHelpers.hpp
@@ -99,16 +99,22 @@ struct PVTFunctionHelpers
       real64 const PEnd = stod( inputParameters[3] );
       real64 const dP = stod( inputParameters[4] );
 
+      GEOSX_THROW_IF( PStart >= PEnd, "PStart must be strictly smaller than PEnd",
+                      InputError );
+
       constexpr real64 T_K = 273.15;
 
       real64 const TStart = stod( inputParameters[5] ) - T_K;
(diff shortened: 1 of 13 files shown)


src/coreComponents/constitutive/fluid/PVTFunctions/PVTFunctionHelpers.hpp
```diff
@@ -99,10 +99,17 @@ struct PVTFunctionHelpers
...
+      constexpr real64 T_K = 273.15;
+
+      real64 const TStart = stod( inputParameters[5] ) - T_K;
+      real64 const TEnd = stod( inputParameters[6] ) - T_K;
       real64 const dT = stod( inputParameters[7] );
 
+      GEOSX_THROW_IF( TStart < 0, "Temperature must be in Kelvin and must be positive",
+                      InputError );
+      GEOSX_THROW_IF( TEnd < 0, "Temperature must be in Kelvin and must be positive",
+                      InputError );
```
TotoGaz, line 111: You are not consistent with `The temperature must be between 283.15 and 623.15 Kelvin.`. Is it OK?

francoishamon: Yes you are right, I added a check to make sure that the temperature are within the range expected for the CO2-brine model, and we will update later if other models are added.

TotoGaz (approved):

PR #1460, opened 2021-06-11
Fixed the bug related to underscores in PAMELA region/surface names

### LINKED ISSUE #1425: [Bug] PAMELA does not handle region/set names containing underscores
**Describe the bug**
PAMELA does not appear to correctly handle external meshes that contain regions/nodesets whose names include underscores.  For example, I have a simple test mesh that contains a couple of nodesets:

```
$MeshFormat
2.2 0 8
$EndMeshFormat
$PhysicalNames
8
2 2 "xpos"
2 3 "xneg"
2 4 "zpos"
2 5 "fault"
2 6 "zneg"
2 7 "ypos"
2 8 "yneg"
3 1 "ALL"
$EndPhysicalNames
...
```

Running the problem with the base mesh, everything looks OK.  However, if I were to change "fault" to "fault_a", then PAMELA appears to silently split the full nodeset name on the "_" character and take the first sub-string as the name.  Looking at this section of code, I couldn't figure out why we chose to do this:

```
static string retrieveSurfaceOrRegionName( string const & pamelaLabel )
    {
      string_array const splitLabel = stringutilities::tokenize( pamelaLabel, m_separator );
      return splitLabel[splitLabel.size() -2 ];
    }
```

The obvious issue is that if a user were to provide a mesh with such a name, they would expect to use the full name in the xml file.  If this occurs for a region name, then the code will throw an error since it cannot the appropriate name in the data structure.  However, if this occurs for a nodeset, then the error can go undetected by the user because we are not throwing an error/warning for field specifications applied to nodesets that don't exist.  For example, if I were to try to use "fault_a" to split the mesh, then nothing would happen, and no error would be produced:

```
<FieldSpecification
      name="frac"
      initialCondition="1"
      setNames="{ fault_a }"
      objectPath="faceManager"
      fieldName="ruptureState"
      scale="1"/>
```

**Expected behavior**
1) PAMELA should probably handle names that include underscores.  Otherwise, if there is some reason why we want to keep this behavior, the code should warn the users
2) We should throw an error and/or print a warning if a FieldSpecification is applied to a nodeset that doesn't exist.  Are there any issues with this @rrsettgast ?

francoishamon:
Until now, it was not easy to figure out how the `PAMELAMeshGenerator` was setting the names of the physical entities in the mesh. For instance, a surface called `Ovbd1_Ovbd2` in the msh file was becoming just `Ovbd2` in GEOSX without notification. 

This PR makes sure that the name used in GEOSX is the same as the name used in the msh file: `Ovbd1_Ovbd2` in the msh file is still `Ovbd1_Ovbd2` in GEOSX.

Fixes #1425

### DIFF
--- a/src/coreComponents/mesh/generators/PAMELAMeshGenerator.hpp
+++ b/src/coreComponents/mesh/generators/PAMELAMeshGenerator.hpp
@@ -152,14 +152,12 @@ public:
       // But we only want to keep Ovbd1_Ovbd2
       // So we find the word GROUP, and then we keep what is after, except the last piece
 
-      using std::begin;
-      using std::end;
-      auto it = std::find( begin( splitLabel ), end( splitLabel ), "GROUP" );
+      auto it = std::find( std::begin( splitLabel ), std::end( splitLabel ), "GROUP" );
 
-      GEOSX_THROW_IF( it == end( splitLabel ),
+      GEOSX_THROW_IF( it == std::end( splitLabel ),
                       "GEOSX assumes that PAMELA places the word GROUP before the region/surface name", InputError );
 
-      localIndex const id = std::distance( begin( splitLabel ), it );
+      localIndex const id = std::distance( std::begin( splitLabel ), it );
       string name = "";
       for( localIndex i = id+1; i < splitLabel.size()-1; ++i )
       {
--- a/src/coreComponents/mesh/generators/PAMELAMeshGenerator.hpp
+++ b/src/coreComponents/mesh/generators/PAMELAMeshGenerator.hpp
@@ -147,7 +147,29 @@ public:
     static string retrieveSurfaceOrRegionName( string const & pamelaLabel )
     {
       string_array const splitLabel = stringutilities::tokenize( pamelaLabel, m_separator );
-      return splitLabel[splitLabel.size() -2 ];
+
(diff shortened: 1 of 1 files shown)


src/coreComponents/mesh/generators/PAMELAMeshGenerator.hpp
```diff
@@ -147,7 +147,29 @@ class PAMELAMeshGenerator : public MeshGeneratorBase
...
     {
       string_array const splitLabel = stringutilities::tokenize( pamelaLabel, m_separator );
-      return splitLabel[splitLabel.size() -2 ];
+
+      // The PAMELA label looks like: PART00002_POLYGON_POLYGON_GROUP_Ovbd1_Ovbd2_14
+      // But we only want to keep Ovbd1_Ovbd2
+      // So we find the word GROUP, and then we keep what is after, except the last piece
+
+      using std::begin;
+      using std::end;
```
AntoineMazuyer, line 156: I am not sure about the usage of the `using` keyword, can't we just put `std` in front all the `begin` and `end ?`

AntoineMazuyer (approved):
Thanks for fixing this bug. For me it's ok

src/coreComponents/mesh/generators/PAMELAMeshGenerator.hpp
francoishamon: Yes, I removed the `using` keyword in the last commit and just used `std::begin` and `std::end`

PR #1464, opened 2021-06-14
Output 4D fields in VTK

### LINKED ISSUE #1430: [Feature] Output PhaseCompFraction (array4d) to VTK
It would be great to bring back the output to file of `phaseCompFraction` as this is the variable that is most convenient to visualize the amount of CO2 component dissolved in brine. The reason why it is not output to the VTK file right now is that we removed the `PlotLevel::LEVEL_O` from here:
https://github.com/GEOSX/GEOSX/blob/310492d084cf66fb6cee730f5cd25e5be22eb03e/src/coreComponents/constitutive/fluid/MultiFluidBase.cpp#L84
This is a temporary fix made necessary by the fact that the VTK writer does not currently support the output of `array4d`. See https://github.com/GEOSX/GEOSX/issues/1112

This is an important feature for CO2-brine simulation

Output the `phaseCompFraction` `array4d` to the vtk file and visualize it with Paraview. I can try to add the `array4d` in the list of accepted types here:
https://github.com/GEOSX/GEOSX/blob/310492d084cf66fb6cee730f5cd25e5be22eb03e/src/coreComponents/common/DataTypes.hpp#L555
 and see if this fixes the problem.

Currently, as a temporary hack when someone needs it, we can bring back the `PlotLevel::LEVEL_0` and use Silo/Visit.

francoishamon:
This PR allows VTK to output fields stored in 4D arrays. This can be quite useful for the output to file of `phaseCompFraction` as this is the variable that is most convenient to visualize the amount of CO2 component dissolved in brine.

To do that, I adopted the approach used in `SiloFile.cpp` for 4D arrays: the second and third dimensions of the array are output component by component. I checked in Paraview and it seems to work as expected.

This PR requires rebaseline.

Related to https://github.com/GEOSX/integratedTests/pull/137

Fixes #1430


src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.hpp
```diff
@@ -251,6 +251,31 @@ class VTKPolyDataWriterInterface
   void writeElementFields( vtkSmartPointer< vtkCellData > celldata,
                            ElementRegionBase const & er ) const;
 
+  /*!
```
TotoGaz, line 254: Everywhere else in the code doxygen's comment start with `/**`. Can you make all this file more consistent please?

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp
```diff
@@ -332,21 +331,127 @@ void VTKPolyDataWriterInterface::writeElementFields( vtkSmartPointer< vtkCellDat
...
     er.forElementSubRegions< SUBREGION >( [&]( auto const & esr )
     {
       WrapperBase const & wrapper = esr.getWrapperBase( field );
-      writeField( wrapper, data, esr.size(), count );
+      std::type_info const & typeID = wrapper.getTypeId();
+      if( typeID==typeid(array4d< real64 >) )
+      {
+        is4DField = true;
+      }
     } );
```
TotoGaz, line 354: Is there a way to extract the dimension of the field instead? (like `array1d -> 1`, `array2d -> 2`...).
That way you could inform the user that there is a non supported type.

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.hpp
```diff
@@ -251,6 +251,31 @@ class VTKPolyDataWriterInterface
...
+   * @brief Writes all the fields with up to 3 dimensions associated to the elements of \p er if their plotlevel is <= m_plotLevel
```
TotoGaz, line 255: What about scalar (1D) data? Your doc lets me assume it should be OK, but not the name of the function.

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.hpp
```diff
@@ -251,6 +251,31 @@ class VTKPolyDataWriterInterface
...
+  template< class SUBREGION >
+  void writeElementField2D3D( vtkSmartPointer< vtkCellData > const celldata,
+                              ElementRegionBase const & er,
+                              string const & field ) const;
+
+  /*!
+   * @brief Writes all the 4D fields associated to the elements of \p er if their plotlevel is <= m_plotLevel
+   *        Considering that the second dimension has size N and the third dimension has size M, this
+   *        function will write MxN fields to VTK
+   * @param[in] celldata a VTK object containing all the fields associated with the elements
```
TotoGaz, line 269: Can you elaborate more on what you are currently doing? On what will the output be?

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp
```diff
@@ -332,21 +331,127 @@ void VTKPolyDataWriterInterface::writeElementFields( vtkSmartPointer< vtkCellDat
...
+    if( is4DField )
+    {
+      // If the field has four dimensions, we apply it a special treatment in the following function
+      writeElementField4D< SUBREGION >( celldata, er, field );
+    }
+    else
+    {
+      // If the field has three dimensions or less, it can be output in a generic fashion
+      writeElementField2D3D< SUBREGION >( celldata, er, field );
+    }
```
TotoGaz, line 367: So here we may be dealing with 1D?

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp
```diff
@@ -332,21 +331,127 @@ void VTKPolyDataWriterInterface::writeElementFields( vtkSmartPointer< vtkCellDat
...
+        data->SetName( (field+"_"+std::to_string( iComp2 )+"_"+std::to_string( iComp3 )).c_str() );
+      }
+
+      // Step 2.2: for each subregion, only output to VTK the data corresponding to the pair of components (iComp2,iComp3)
+      er.forElementSubRegions< SUBREGION >( [&]( auto const & esr )
+      {
+        WrapperBase const & wrapper = esr.getWrapperBase( field );
+        Wrapper< array4d< real64 > > const & wrapperT = dynamicCast< Wrapper< array4d< real64 > > const & >( wrapper );
+        traits::ViewTypeConst< array4d< real64 > > const sourceArray = wrapperT.reference().toViewConst();
+
```
TotoGaz, line 438: So you do not need and type check here?

TotoGaz (commented):
I do not understand the traits stuff.

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.hpp
francoishamon: Done, I replaced /*! with /**

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.hpp
francoishamon: I added more comments on what people will be able to click on in Paraview. The Paraview fields will look like:
```
            field_0_0
              |
            field_0_(M-1)
              |
            field_(N-1)_0
              |
            field_(N-1)_(M-1)
```
and if the dimension of the 2nd dimension is one (which is the case for fluid constitutive data for flow-only simulations), then the first _0 is skipped. For a CO2-brine simulation, we will be able to click for instance:
```    
          phaseCompFraction_0
          phaseCompFraction_1
```
so this just mimics what is done in Silo/Visit.

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp
francoishamon: I added a check here

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp
francoishamon: I renamed the function `writeElementField1D2D3D` to make it clear that this function is going to output the fields stored in `array1d`, `array2d`, and `array3d`.

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp
francoishamon: I do not know if there is a way to extract the dimension of the field from the wrapper, other than using `typeid`. In the current code and in this PR, if the dimension of the field is not supported, an error message will be displayed and stop the simulation:
```
***** LOCATION: /g/g90/hamon1/geosx/GEOSX/src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp:442                                     
***** ERROR                                                                                                                                    
***** LOCATION: /g/g90/hamon1/geosx/GEOSX/src/coreComponents/common/DataTypes.hpp:917                                                          
***** Rank 0: TypeID not recognized. 
```
Even if it may not be very generic, I reused the logic of SiloFile.cpp, for instance here: 
https://github.com/GEOSX/GEOSX/blob/64bba253b61010d2810415c892b5227a2e3dc14f/src/coreComponents/fileIO/silo/SiloFile.cpp#L916-L921

klevzoff: We're going to have to change it very soon (both in Silo and VTK outputs), because the `typeid` check does not work with layout permutations (and so will fail or work incorrectly on CUDA builds).

The easiest "generic" mechanics is to add a new methods to `WrapperBase` and `Wrapper`. Call it something like `numArrayDims()` and have it return NDIM for `LvArray::Array` and perhaps 0 for all other types (using template specialization in `wrapperHelpers`, similar to other specialized methods). The mechanism is a bit overused and keeps inflating the API surface of `Wrapper` class, but it's quite reliable.

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp
```diff
@@ -332,21 +331,131 @@ void VTKPolyDataWriterInterface::writeElementFields( vtkSmartPointer< vtkCellDat
...
+  localIndex nCompInSecondDim = 0;
+  localIndex nCompInThirdDim = 0;
+  er.forElementSubRegions< SUBREGION >( [&]( auto const & esr )
+  {
+    WrapperBase const & wrapper = esr.getWrapperBase( field );
+    std::type_info const & typeID = wrapper.getTypeId();
+    GEOSX_THROW_IF( typeID!=typeid(array4d< real64 >),
+                    "This function is specific to array4d< real64 >", InputError );
+
+    Wrapper< array4d< real64 > > const & wrapperT = dynamicCast< Wrapper< array4d< real64 > > const & >( wrapper );
```
francoishamon, line 405: @klevzoff Is there a good way to make this piece of code handle all the possible permutations of the array? Should I test of the possible permutations of `array4d`, or is there a way to make this more generic?

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp
francoishamon: Thanks for the suggestion, I implemented what you suggested (but it is worth double checking what I have done :) )

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp
klevzoff: I don't actually have any simple ideas at the moment. The way we used to do it was with `arrayTypeLambda()`, but that clearly is not scalable - we'd need to add 24 switch cases just for 4D arrays of one data type (`real64`). Although if we're pretty sure about never going above 4D, we could just bite the bullet and do it once and for all (with the caveat that all code that currently relies on `arrayTypeLambda()` needs to be checked/updated to make sure it can handle 4D).

Short of that, I can only think of the following:
* `reinterpret_cast` the wrapper to type `Wrapper< array4d< real64 > >` (note, not `dynamicCast` since we want to avoid the actual `typeid` check)
* move data to host
* convert it to an `arraySlice4d< real64 const >` by calling `toSliceConst()`
* `reinterpret_cast` the result to `arraySlice4d< real64 const, -1 >` (i.e. forcing `USD=-1`). 

This is technically undefined behavior, but it should work in practice. It relies on the fact that array objects have exact same binary representation regardless of layout - only the computed stride values differ. It slightly pessimizes offset computation (since it has to multiply all strides in absence of single stride dim), but it's largely inconsequential for output. Still, this is brittle and might break in the future - imagine we "optimized" arrays by squashing `m_strides` (i.e. not physically store the unit stride).

@rrsettgast @corbett5 opinions?

francoishamon: Thanks @klevzoff. For now I implemented what you suggested with the `reinterpret_cast` and I will backtrack if we find a better solution. The current version is working, although I have not tested on GPU yet, and I don't have a good test for a permuted `array4d` (but I guess your PR will do that soon).

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp
```diff
@@ -332,21 +331,130 @@ void VTKPolyDataWriterInterface::writeElementFields( vtkSmartPointer< vtkCellDat
...
+{
+  // Step 1: we need to figure out the size of the two dimensions of the array that we are going to output component by component
+  localIndex nCompInSecondDim = 0;
+  localIndex nCompInThirdDim = 0;
+  er.forElementSubRegions< SUBREGION >( [&]( auto const & esr )
+  {
+    WrapperBase const & wrapper = esr.getWrapperBase( field );
+    Wrapper< array4d< real64 > > const & wrapperT = reinterpret_cast< Wrapper< array4d< real64 > > const & >( wrapper );
+    arrayView4d< real64 const > const & sourceArray = wrapperT.reference();
+    sourceArray.move( LvArray::MemorySpace::host, true );
```
klevzoff, line 403:
```suggestion
    sourceArray.move( LvArray::MemorySpace::host, false );
```

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp
```diff
@@ -332,21 +331,130 @@ void VTKPolyDataWriterInterface::writeElementFields( vtkSmartPointer< vtkCellDat
...
+    arraySlice4d< real64 const > const & tmp = sourceArray.toSliceConst();
+    arraySlice4d< real64 const, -1 > const & sourceSlice = reinterpret_cast< arraySlice4d< real64 const, -1 > const & >( tmp );
```
klevzoff, line 405: Would you mind putting this whole sequence in a function (just a top-level function in this cpp would be fine, perhaps in an unnamed namespace) and adding a warning comment that this is a temporary hack to get around the lack of type dispatching for 4D arrays?

francoishamon: Hi @klevzoff I don't mind doing what you suggest but I got stuck with the C++ casting syntax for the "return by value":

I can do something like that:
```
arraySlice4d< real64 const, -1 >                                                                                                                                                         
get4DSliceFromWrapper( WrapperBase const & wrapper )                                                                                                                                     
{                                                                                                                                                                                        
  Wrapper< array4d< real64 > > const & wrapperT = reinterpret_cast< Wrapper< array4d< real64 > > const & >( wrapper );                                                                   
  arrayView4d< real64 const > const & sourceArray = wrapperT.reference();                                                                                                                
  sourceArray.move( LvArray::MemorySpace::host, false );                                                                                                                                 
  arraySlice4d< real64 const > const & tmp = sourceArray.toSliceConst();                                                                                                                 
  arraySlice4d< real64 const, -1 > const & sourceSlice = reinterpret_cast< arraySlice4d< real64 const, -1 > const & >( tmp );                                                                                                                                                                                         
  return sourceSlice;                                                                                                                                                                    
}
```
but it does not work (it crashes when I access the returned slice) because I assume my `sourceSlice` is a reference to a temporary (?). Ideally I would like to do something like that:
```
arraySlice4d< real64 const, -1 >                                                                                                                                                         
get4DSliceFromWrapper( WrapperBase const & wrapper )                                                                                                                                     
{                                                                                                                                                                                        
  Wrapper< array4d< real64 > > const & wrapperT = reinterpret_cast< Wrapper< array4d< real64 > > const & >( wrapper );                                                                   
  arrayView4d< real64 const > const & sourceArray = wrapperT.reference();                                                                                                                
  sourceArray.move( LvArray::MemorySpace::host, false );                                                                                                                                 
  arraySlice4d< real64 const > const tmp = sourceArray.toSliceConst();                                                                                                                 
  arraySlice4d< real64 const, -1 > const sourceSlice = reinterpret_cast< arraySlice4d< real64 const, -1 > const >( tmp );                                                                                                                                                                                         
  return sourceSlice;                                                                                                                                                                    
}
```
but this does not compile:
```
reinterpret_cast from 'const arraySlice4d<const geosx::real64>' (aka 'const ArraySlice<const double, 4, 3, long>') to 'const arraySlice4d<const geosx::real64, -1>' (aka 'const ArraySlice<const double, 4, -1, long>') is not allowed 
```
I have to say I am a little bit confused by the syntax in this piece of code. Is there a better way to safely return the slice by value?

klevzoff: The first attempt seems like it should work. You can try removing reference from `tmp`, and also initializing a local variable of type `arraySlice4d< real64 const, -1 >` which you then return (instead of having the compiler copy-construct the return value for you). Try printing the slice object before returning to make sure it's initialized there.

If none of it works, we've likely been hit by the compiler optimizing around UB. Specifically, it likely just deletes all the code following an illegal `reinterpret_cast` and skips initializing the return value (it is allowed to do so by the standard). I see two ways of working around it:
1. Declare a function taking a `void *` and returning a `void *` and pass the slice through it, using two `reinterpret_cast`s for the argument and the return value. The function's definition must be in a separate translation unit. This approach keeps the UB, but prevents the compiler from seeing through it. I don't like this any better than I did the original solution, but it's an option.
2. (Preferred) Add a (conditionally enabled) explicit conversion constructor to `ArraySlice`, to allow constructing a slice with negative `USD` from a slice with any other `USD`. This is the route we should've probably taken from the start, but it's a bit longer.

francoishamon: Thanks! I tried the first part of what you suggested (removing the reference and adding the local variable), I see that the slice is well initialized in the function doing the cast (`get4DSliceFromWrapper`), but the slice that I get after the function has returned is garbage. The code only works properly when the two `reinterpret_cast` are in the body of the main function that writes the field (`writeElementField4D`). 

It would be great to have option 2, but I am not sure I can write it on my own. I can try to backtrack and add the `array4d` + permutations to the supported datatypes. I started doing that once, it requires the addition of a few functions in `FieldSpecifications` as you said, but maybe it is going to be a better strategy.

klevzoff: That sounds good to me. I can also help write the `ArraySlice` constructor, but extending the switchyard has always been the best option (as long as you don't encounter some existing call site that hard-assumed no more than 3 dims).

By the way, I looked again at the code above... I think the culprit was `sourceArray` is a local temporary, and any slice produced from it contained pointers to that temporary, which went out of scope. I thought initially it was a reference to the original (permanent) array object, but there is a non-`const` to `const` conversion involved which now produces a new temporary view. In any case, I'm actually relieved we won't be going the UB route in the end.

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp
corbett5: I think adding in a constructor for a -1 strided slice from one with a known slice would be a great call. I'll make an issue, I think it should be pretty straightforward but I doubt I'll get around to it in time for this PR.

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp
corbett5: I've tried to get rid of as many instances of this bug as possible (creating a slice from a temporary view) and I've managed to make it a compilation error for rvalues, but I don't think preventing this bug is possible. Improving documentation would help.

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp
francoishamon: Thanks a lot @corbett5. I ended up using a different approach based on `applyArrayTypeLambda2` to figure out the type of the `array4d`, and now I have something relatively clean that should support all the permutations without requiring the `reinterpret_cast`. So I think there is no longer an immediate need for this constructor in the present PR.

corbett5: Did you just add all the types (or at least all the types we're currently using) to the giant switch yard? If so I'd much rather have the generic slice and get rid of them. I think most of those dispatchers are used for IO anyways.

francoishamon: Yes in the current version of the PR, all the permutations of `array4d` are added to the dispatch in `applyArrayTypeLambda2`, which requires some changes in `common/FieldSpecificationOps.hpp`. But if the generic slice is a cleaner option, that is perfectly fine with me.

corbett5: Well the issue is that now when you pass something to `applyArrayTypeLambda2` it has to instantiate all 24 different variants, which really bloats compile time. God forbid we ever do 5D...

corbett5: I'm not at all saying that you should wait, I think this is a good solution. I've just been trying to ditch those type dispatchers for ever, I think they're quite nasty.

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp
```diff
@@ -91,6 +91,48 @@ gatherNbElementsInRegion( ElementRegionBase const & er,
...
+template< typename T >
+typename std::enable_if_t< !traits::is_array_type< T > || T::NDIM != 4 >
+insertValueFrom4DField( vtkSmartPointer< VTKGEOSXData > const data,
+                        T const & x,
+                        localIndex & count,
+                        localIndex i, localIndex j, localIndex k, localIndex l )
+{
+  GEOSX_UNUSED_VAR( data, x, count, i, j, k, l );
+  GEOSX_ERROR( "This function should not be called" );
+}
```
klevzoff, line 133: I don't think you need to handle non-array types here (and they wouldn't compile anyway, because `T::NDIM` wouldn't be a valid expression for anything other than an `ArrayView` or `ArraySlice`). So you could "simplify" this to:
```c++
template< typename T, int NDIM, int USD >
void insertValueFrom4DField( vtkSmartPointer< VTKGEOSXData > const data,
                             ArrayView< T, NDIM, USD > const & field,
                             ... )
{
  GEOSX_UNUSED_VAR( data, field, count, i, j, k, l );
  GEOSX_ERROR( "This function should not be called" );
}

template< typename T, int USD >
void insertValueFrom4DField( vtkSmartPointer< VTKGEOSXData > const data,
                             ArrayView< T, 4, USD > const & field,
                             ... )
{
  data->customInsertValue( count++, field( i, j, k, l ) );
}
```

This isn't any shorter than what you have, but the lack of `std::enable_if` is a small win for readability in my books :)
(On a side note, how I wish we could use C++17's `if constexpr`...)

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp
```diff
@@ -332,21 +373,154 @@ void VTKPolyDataWriterInterface::writeElementFields( vtkSmartPointer< vtkCellDat
...
+  localIndex nCompInThirdDim = 0;
+  er.forElementSubRegions< SUBREGION >( [&]( auto const & esr )
+  {
+    WrapperBase const & wrapperBase = esr.getWrapperBase( field );
+    std::type_info const & typeID = wrapperBase.getTypeId();
+    rtTypes::applyArrayTypeLambda2( rtTypes::typeID( typeID ),
+                                    true,
+                                    [&]( auto array, auto GEOSX_UNUSED_PARAM( Type ) )
+    {
+      typedef decltype( array ) arrayType;
```
klevzoff, line 448:
```suggestion
      using arrayType = decltype( array );
```
(sorry to nitpick, we probably still have `typedef` in a bunch of places, but I'm trying to root them out)

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp
```diff
@@ -332,21 +373,154 @@ void VTKPolyDataWriterInterface::writeElementFields( vtkSmartPointer< vtkCellDat
...
+    {
+      localIndex count = 0;
+
+      // Step 2.1: write the name of the specific component iComp we consider here
+      vtkSmartPointer< VTKGEOSXData > data = VTKGEOSXData::New();
+      data->SetNumberOfValues( er.getNumberOfElements< SUBREGION >() );
+
+      if( nCompInSecondDim == 1 )
+      {
+        data->SetName( (field+"_"+std::to_string( iComp3 )).c_str() ); // special case for nice display of constitutive fields
```
klevzoff, line 479: At some point we will need to handle constitutive fields entirely separately like Silo does... because we have to average over quadrature points - important for mechanics quantities like stress. Just a comment, this is fine for now :)

klevzoff (approved):
Looks great!

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp
klevzoff: On a second look, please also remove the passing of `VTKGEOSXData` via smart pointer. This involves a reference count increment/decrement, which, depending on the implementation, might be somewhat expensive (e.g. acquiring a mutex lock). Best to just pass by reference (`VTKGEOSXData & data`).

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp
francoishamon: Done, I also removed the `typedef` from the original function `writeField`.

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp
francoishamon: Done, I modified the function as suggested in your two comments above.

PR #1465, opened 2021-06-16
Specificy multiple input xml files from command line

cssherman:
This PR implements a minimal approach for handling multiple input xml files via the command line:

```
geosx -i input_a.xml -i input_b.xml -i input_c.xml
```

To do this, I am writing out an additional xml file (for now assumed to be "composite_input.xml"), which contains the include instructions and serves as a record of options.  This file is then passed to the existing parser.

In places where the code depends on the input filename, such as the default problem name, I am using the first file.


cssherman:
@rrsettgast - thoughts on this approach?  Alternatively, we could keep the push the inputFileNames list further into ProblemManager, and skip writing the composite xml file (this would require a integrated test rebase).

CusiniM:
What is the use case for this? Running multiple problems?

rrsettgast (approved):

PR #1466, opened 2021-06-17
Implemented CFL numbers in compositional solver

francoishamon:
This PR adds an option to compute the CFL numbers in `CompositionalMultiphaseFVM`. The calculation is based on the usual formulas (Hui Cao's thesis for 2-phase flow, Coats' paper for 3-phase flow). 

The immediate need is to run a weak scaling study at constant CFL number across refinement levels, but this will be useful more generally to know more about the time scales of our flow simulations.

Current status:
- [X] Component CFL numbers
- [X] Phase CFL numbers (2-phase flow)
- [X] Phase CFL numbers (3-phase flow)

Related to https://github.com/GEOSX/integratedTests/pull/138


src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVMKernels.cpp
```diff
@@ -663,6 +663,339 @@ INST_FluxKernel( 5, FaceElementStencil );
...
+    real64 const mobility = phaseRelPerm[er_up][esr_up][ei_up][0][ip] / phaseVisc[er_up][esr_up][ei_up][0][ip];
+
+    // increment the phase (volumetric) outflux of the upstream cell
+    real64 const absPhaseFlux = fabs( dt * mobility * potGrad );
+    phaseOutflux[er_up][esr_up][ei_up][ip] += absPhaseFlux;
+
+    // increment the component (mass/molar) outflux of the upstream cell
+    for( localIndex ic = 0; ic < NC; ++ic )
+    {
+      compOutflux[er_up][esr_up][ei_up][ic] += phaseCompFrac[er_up][esr_up][ei_up][0][ip][ic]
```
klevzoff, line 747: Do these sums need to be atomic?

francoishamon: Yes! It is a mistake, I have not started testing this with threads yet. Thanks for the catch!

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.cpp
```diff
@@ -175,6 +175,110 @@ void CompositionalMultiphaseFVM::assembleFluxTerms( real64 const dt,
...
+                                        compOutflux,
+                                        localMaxPhaseCFLNumber,
+                                        localMaxCompCFLNumber );
+  } );
+
+  real64 const globalMaxPhaseCFLNumber = MpiWrapper::max( localMaxPhaseCFLNumber );
+  real64 const globalMaxCompCFLNumber = MpiWrapper::max( localMaxCompCFLNumber );
+
+  GEOSX_LOG_RANK_0( "Max phase CFL number: " << globalMaxPhaseCFLNumber );
+  GEOSX_LOG_RANK_0( "Max component CFL number: " << globalMaxCompCFLNumber );
```
klevzoff, line 279: Not sure if we want these logged at verbosity level 0 (which is normally just time stepping progress)... level 1 usually includes all the details of a time step. Although one could argue that this information is relevant to analyze time stepping behavior (cuts, etc.). Just something to consider.

Also, since we can have multiple physics solvers in a simulation, maybe consider prefixing this output with solver name:
```suggestion
  GEOSX_LOG_RANK_0( getName() << ": Max phase CFL number: " << globalMaxPhaseCFLNumber );
  GEOSX_LOG_RANK_0( getName() << ": Max component CFL number: " << globalMaxCompCFLNumber );
```

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVMKernels.cpp
```diff
@@ -663,6 +663,356 @@ INST_FluxKernel( 5, FaceElementStencil );
...
+                   real64 & phaseCFLNumber )
+{
+  // first, check which phases are mobile in the cell
+  real64 mob[NP]{};
+  localIndex mobilePhases[NP]{};
+  localIndex numMobilePhases = 0;
+  for( localIndex ip = 0; ip < NP; ++ip )
+  {
+    mob[ip] = phaseRelPerm[ip] / phaseVisc[ip];
+    if( mob[ip] > 1e-12 )
```
klevzoff, line 851: Maybe make it a class-level constant? It's not really used anywhere else, just tells the reader we consider this value somewhat important.
```c++
static constexpr real64 minPhaseMobility = 1e-12;
```

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVMKernels.cpp
```diff
@@ -663,6 +663,356 @@ INST_FluxKernel( 5, FaceElementStencil );
...
+                  arraySlice1d< real64 const > compOutflux,
+                  real64 & compCFLNumber )
+{
+
+
+  compCFLNumber = 0.0;
+  for( localIndex ic = 0; ic < NC; ++ic )
+  {
+    real64 const compMoles = compDens[ic] * poreVol;
+    if( compMoles > 1e-12 )
```
klevzoff, line 929: Same here I suppose. Also this threshold seems rather scale-dependent, I guess we could use component fraction as an "intrinsic" quantity to detect present components, which doesn't protect against near-zero pore volume, but that would be a problem for other reasons...

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVMKernels.cpp
```diff
@@ -663,6 +663,356 @@ INST_FluxKernel( 5, FaceElementStencil );
...
+          arrayView1d< real64 const > const & porosityRef,
+          arrayView2d< real64 const > const & pvMult,
+          arrayView2d< real64 const > const & compDens,
+          arrayView3d< real64 const > const & phaseRelPerm,
+          arrayView4d< real64 const > const & dPhaseRelPerm_dPhaseVolFrac,
+          arrayView3d< real64 const > const & phaseVisc,
+          arrayView2d< real64 const > const & phaseOutflux,
+          arrayView2d< real64 const > const & compOutflux,
+          real64 & maxPhaseCFLNumber,
+          real64 & maxCompCFLNumber )
```
klevzoff, line 954: Would it also be of interest to compute and _store_ cell-wise CFL (so one could plot and analyze it in post-processing over the entire mesh)?

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVMKernels.cpp
```diff
@@ -663,6 +663,356 @@ INST_FluxKernel( 5, FaceElementStencil );
...
+    computeCompCFL< NC >( poreVol,
+                          compDens[ei],
+                          compOutflux[ei],
+                          cellCompCFLNumber );
+    regionCompCFLNumber.max( cellCompCFLNumber );
+
+  } );
+
+  maxPhaseCFLNumber = LvArray::math::max( maxPhaseCFLNumber, regionPhaseCFLNumber.get() );
+  maxCompCFLNumber = LvArray::math::max( maxCompCFLNumber, regionCompCFLNumber.get() );
```
klevzoff, line 984: I'm not a big fan of in-out parameters, they are too easy to misuse (accidentally provide uninitialized value, expecting the function to overwrite it, and get garbage out). Maybe you could just return the subregion max CFL via `maxPhaseCFLNumber` and `maxCompCFLNumber` and have the calling code do the max over (sub)regions. It's a little more verbose (two extra local variables in the calling code), but the code is easier to reason about IMO, since looping over subregions and taking the max are both in one place.

klevzoff (approved):
LGTM, just minor nitpicks/suggestions

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.cpp
francoishamon: I agree, I added the `getName()` and I changed the log level to 1 for the CFL numbers.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVMKernels.cpp
francoishamon: Yes it makes sense to try to visualize the map of CFL numbers, I added this option (cell-wise CFL numbers are now stored).

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVMKernels.cpp
francoishamon: Now I just return the subRegion max CFL and then the calling code takes the max over the subRegions. I made the same change in `calculateResidualNorm` for both `CompositionalMultiphaseFVM` and `CompositionalMultiphaseHybridFVM` to avoid the bug that I had introduced a few month ago by wrongly setting `localResidualNorm` to 0 for each subRegion.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVMKernels.cpp
francoishamon: I added this class-level constant to check the zero mobilities.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVMKernels.cpp
francoishamon: Now I check the component fraction and not anymore the moles of the components in this threshold, and I added a class-level constant as you suggest.

PR #1467, opened 2021-06-19
Add new cluster target and clean CI

TotoGaz:
Main purpose of this PR is to add TotalEnergies' cluster GPU Pecan, with automatic delivery (like already done for Pangea2).
This PR is related to PR https://github.com/GEOSX/thirdPartyLibs/pull/146 that builds the TPL Pecan look alike container.

To achieve the result, some refactoring was mandatory.

- It is now possible to select another `host-config` file in `.travis.yml`.
- A `host-config` file was developed (and validated) for _Pecan_.
- The usage of environmental variables is better centralized in `host-configs/environment.cmake` file. This file is now entirely responsible for transforming environment variables into `cmake` variables. It now allows `CUDA`.
- Improved management of linear algebra backend selection. `ENABLE_TRILINOS` and `ENABLE_HYPRE` **must** now be defined exclusively. Ambiguous selection result in an error (like both `ENABLE_TRILINOS=ON` and `ENABLE_HYPRE=ON`)

To be noted

- Same deployment code and pattern could be used for both Pangea2 and Pecan.
- Travis' targets have the same linear algebra backend as previously.
- Some weird error about `abs` in some unit test which had to be changed to `std::abs`. I do not know why...
- The build for Pecan is quite time consuming: _1h36_, like other `cuda` `release` builds. Maybe we could replace `Centos7.6_gcc8.3.1_cuda10.1.243_release` by Pecan since they do approximately the same compilation validation. Or maybe could we remove automatic builds for both Pangea and Pecan and only build them on request (a tag in the commit message when triggering the branch build on travis _e.g._?) @rrsettgast @joshua-white @joshua-white


host-configs/TOTAL/pecan.cmake
```diff
@@ -0,0 +1,47 @@
...
+  set(CMAKE_CUDA_HOST_COMPILER ${CMAKE_CXX_COMPILER} CACHE STRING "")
+  set(CMAKE_CUDA_COMPILER ${CUDA_TOOLKIT_ROOT_DIR}/bin/nvcc CACHE STRING "")
+  set(CUDA_ARCH sm_75 CACHE STRING "")
+  set(CMAKE_CUDA_STANDARD 14 CACHE STRING "")
+  ### The inclusion of -std=c++14 is a workaround for a cuda10/gcc8 bug ###
+  set(CMAKE_CUDA_FLAGS "-restrict -arch ${CUDA_ARCH} --expt-relaxed-constexpr --expt-extended-lambda -Werror cross-execution-space-call,reorder,deprecated-declarations -Xcompiler -std=c++14" CACHE STRING "")
+  set(CMAKE_CUDA_FLAGS_RELEASE "-O3 -DNDEBUG -Xcompiler -DNDEBUG -Xcompiler -O3" CACHE STRING "")
+  set(CMAKE_CUDA_FLAGS_RELWITHDEBINFO "-g -lineinfo ${CMAKE_CUDA_FLAGS_RELEASE}" CACHE STRING "")
+  set(CMAKE_CUDA_FLAGS_DEBUG "-g -G -O0 -Xcompiler -O0" CACHE STRING "")
+endif()
```
rrsettgast, line 29: ENABLE_HYPRE_CUDA ??

host-configs/environment.cmake
```diff
@@ -13,5 +13,59 @@ set(MPIEXEC_EXECUTABLE "$ENV{MPIEXEC}" CACHE PATH "" FORCE)
...
+
+  set(CUDA_ARCH "$ENV{CUDA_ARCH}" CACHE STRING "" FORCE)
+  if(NOT CUDA_ARCH)
+    set(CUDA_ARCH "sm_70" CACHE STRING "" FORCE)
+  endif()
+
+  set(ENABLE_HYPRE_CUDA "$ENV{ENABLE_HYPRE_CUDA}" CACHE BOOL "" FORCE)
+  if(NOT ENABLE_HYPRE_CUDA)
+    set(ENABLE_HYPRE_CUDA OFF CACHE BOOL "" FORCE)
+  endif()
```
rrsettgast, line 66: is this if block redundant

src/coreComponents/unitTests/virtualElementTests/testVirtualElementBase.cpp
```diff
@@ -40,7 +40,7 @@ static void checkIntegralMeanConsistency( VEMBASISDATATYPE const & basisData )
   {
     sum += basisFunctionsIntegralMean[iBasisFun];
   }
-  EXPECT_TRUE( abs( sum-1 ) < 1e-15 )
+  EXPECT_TRUE( std::abs( sum-1 ) < 1e-15 )
```
rrsettgast, line 43: is `std::abs` available in cuda? I want to say I had always used `fabs`.

rrsettgast (approved):

src/coreComponents/unitTests/virtualElementTests/testVirtualElementBase.cpp
corbett5: Use `LvArray::math::abs`, it's the best of both worlds.

host-configs/TOTAL/pecan.cmake
TotoGaz: What do you mean? I do not think that I understand.
Do you think that I should activate `HYPRE_CUDA` by default?

host-configs/environment.cmake
TotoGaz: Do you mean redundant w.r.t. the `if(ENABLE_CUDA)` it lives in?
I had seen that there was some special care about having `hypre` work with `CUDA`.
And this is what I wanted to reproduce here.

host-configs/TOTAL/pecan.cmake
rrsettgast: yes

host-configs/environment.cmake
TotoGaz: Here also, `ENABLE_HYPRE_CUDA` should be `ON` if we use `CUDA` and `HYPRE`, no need to check for the user input?

src/coreComponents/unitTests/virtualElementTests/testVirtualElementBase.cpp
TotoGaz: Done.

host-configs/TOTAL/pecan.cmake
TotoGaz: I've tested the compilation of `hypre` with `cuda` but it's not working on pecan...
But a more recent version of `hypre` is working.
I would like to have the pecan build like this and deal with the GPU stuff in another PR (maybe wait for `GEOSX` to update `hypre`).

Is it OK for you?

PR #1470, opened 2021-06-25
Structure of the wave propagator

mndiaye24:
In this PR we propose a new structure for the wave solver in GEOSX.

image

### DIFF
--- a/src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEM.cpp
+++ b/src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEM.cpp
@@ -18,6 +18,7 @@
  */
 
 #include "AcousticWaveEquationSEM.hpp"
+#include "AcousticWaveEquationSEMKernel.hpp"
 
 #include "dataRepository/KeyNames.hpp"
 #include "finiteElement/FiniteElementDiscretization.hpp"
@@ -580,7 +581,22 @@ real64 AcousticWaveEquationSEM::explicitStep( real64 const & time_n,
 
   arrayView1d< real64 > const rhs = nodeManager.getExtrinsicData< extrinsicMeshData::ForcingRHS >();
 
-  forTargetRegionsComplete( mesh, [&]( localIndex const,
+
+  
+  auto kernelFactory = AcousticWaveEquationSEMKernels::ExplicitAcousticSEMFactory( dt );
+
+  finiteElement::
+    regionBasedKernelApplication< parallelDevicePolicy< 32 >,
+                                  constitutive::NullModel,
+                                  CellElementSubRegion >( mesh,
+                                                          targetRegionNames(),
+                                                          getDiscretizationName(),
+                                                          arrayView1d< string const >(),
+                                                          kernelFactory );
+
+  
+  
+  /*forTargetRegionsComplete( mesh, [&]( localIndex const,
                                        localIndex const,
                                        ElementRegionBase & elemRegion )
   {
@@ -632,7 +648,8 @@ real64 AcousticWaveEquationSEM::explicitStep( real64 const & time_n,
       } );
     } );
   } );
-
+  */
+  
   addSourceToRightHandSide( time_n, rhs );
 
   /// Calculate your time integrators
--- a/src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEM.cpp
+++ b/src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEM.cpp
@@ -32,14 +32,9 @@ using namespace dataRepository;
 
 AcousticWaveEquationSEM::AcousticWaveEquationSEM( const std::string & name,
                                                   Group * const parent ):
-  SolverBase( name,
-              parent )
+  WaveSolverBase( name,
+		  parent )
 {
-
-  registerWrapper( viewKeyStruct::sourceCoordinatesString(), &m_sourceCoordinates ).
-    setInputFlag( InputFlags::REQUIRED ).
-    setSizedFromParent( 0 ).
-    setDescription( "Coordinates (x,y,z) of the sources" );
 
   registerWrapper( viewKeyStruct::sourceNodeIdsString(), &m_sourceNodeIds ).
     setInputFlag( InputFlags::FALSE ).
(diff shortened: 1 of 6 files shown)


src/coreComponents/mesh/CellBlock.hpp
```diff
@@ -145,6 +145,20 @@ class CellBlock : public ElementSubRegionBase
                      localIndex_array & nodeIndices ) const;
 
 
+  /**
```
corbett5, line 148: This shouldn't be a member. It should ideally be a free function or a member of the finite element/geometry space or something like that. Either way we should absolutely not be capturing a `CellBlock` on device.

Same goes for the `getFaceNodes` above, but you didn't add those.

corbett5: Also since this is templated on the `ELEM_TYPE`, `nodeIndices` could be statically sized. For example for a `ElementType::Hexahedron`, `nodeIndices` could be `localIndex[ 6 ][ 4 ]`. Also you're assuming that they user has properly resized `nodeIndices` prior to this call since no checking is done to make sure it is the correct size.

src/coreComponents/mesh/utilities/ComputationalGeometry.hpp
```diff
@@ -361,18 +361,18 @@ int sign( T const val )
...
 {
-  localIndex const numFaces = faceNodeIndicies.size( 0 );
+  localIndex const numFaces = faceNodeIndices.size( 0 );
   R1Tensor faceCenter, faceNormal;
   int prev_sign = 0;
 
   for( localIndex kf = 0; kf < numFaces; ++kf )
   {
-    Centroid_3DPolygon( faceNodeIndicies[kf], nodeCoordinates, faceCenter, faceNormal, areaTolerance );
+    centroid_3DPolygon( faceNodeIndices[kf], nodeCoordinates, faceCenter, faceNormal, areaTolerance );
```
corbett5, line 375: Nice fixes.

src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEM.cpp
```diff
@@ -163,6 +134,33 @@ void AcousticWaveEquationSEM::registerDataOnMesh( Group & meshBodies )
...
+
+  GEOSX_THROW_IF( m_receiverCoordinates.size( 1 ) != 3,
+                  "Invalid number of physical coordinates for the receivers",
+                  InputError );
+
+  localIndex const numNodesPerElem = 8;
+
+  localIndex const numSourcesGlobal = m_sourceCoordinates.size( 0 );
+  m_sourceNodeIds.resizeDimension< 0, 1 >( numSourcesGlobal, numNodesPerElem );
+  m_sourceConstants.resizeDimension< 0, 1 >( numSourcesGlobal, numNodesPerElem );
```
corbett5, line 152: These are 2D arrays right? In that case you can just do `m_sourceNodeIds.resize( numSourcesGlobal, numNodesPerElem )` and call it good.

src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEM.cpp
```diff
@@ -209,72 +207,33 @@ void AcousticWaveEquationSEM::precomputeSourceAndReceiverTerm( MeshLevel & mesh
 
         constexpr localIndex numNodesPerElem = FE_TYPE::numNodes;
         localIndex const numFacesPerElem = elementSubRegion.numFacesPerElement();
-        array1d< array1d< localIndex > > faceNodes( numFacesPerElem );
 
-        forAll< serialPolicy >( elementSubRegion.size(), [=, &elementSubRegion] ( localIndex const k )
+        // TODO remove
```
corbett5, line 211: This is going to be INCREDIBLY slow. I wouldn't commit this.

src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEMKernel.hpp
```diff
@@ -0,0 +1,500 @@
...
+                                        arrayView2d< localIndex const, cells::NODE_MAP_USD > const elemsToNodes,
+                                        arrayView2d< real64 const, nodes::REFERENCE_POSITION_USD > const X )
+  {
+    if( computationalGeometry::isPointInsidePolyhedron( X, faceNodeIndices, coords ) )
+    {
+      constexpr localIndex numNodesPerElem = FE_TYPE::numNodes;
+      real64 xLocal[numNodesPerElem][3];
+      for( localIndex a = 0; a < numNodesPerElem; ++a )
+      {
+        for( localIndex i = 0; i < 3; ++i )
```
corbett5, line 61: `LvArray::tensorOps::copy< 3 >( xLocal[ a ], X[ elemsToNodes( elementIndex, a ) ] )`

src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEMKernel.hpp
```diff
@@ -0,0 +1,500 @@
...
+          arrayView1d< localIndex > const sourceIsLocal,
+          arrayView2d< localIndex > const sourceNodeIds,
+          arrayView2d< real64 > const sourceConstants,
+          arrayView2d< real64 const > const receiverCoordinates,
+          arrayView1d< localIndex > const receiverIsLocal,
+          arrayView2d< localIndex > const receiverNodeIds,
+          arrayView2d< real64 > const receiverConstants )
+  {
+
+    forAll< EXEC_POLICY >( size, [=] GEOSX_HOST_DEVICE ( localIndex const k )
```
corbett5, line 129: Just so I understand, this is looping over every element, and then for each element it checks if each source/receiver is contained inside?

src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEMKernel.hpp
```diff
@@ -0,0 +1,500 @@
...
+            FE_TYPE::calcN( q, N );
+            real64 const detJ = m_finiteElement.template getGradN< FE_TYPE >( k, q, xLocal, gradN );
+
+            real64 invJ[3][3] = {{ 0 }};
+            FE_TYPE::invJacobianTransformation( q, xLocal, invJ );
+
+            for( localIndex a = 0; a < numNodesPerFace; ++a )
+            {
+              // compute ds = || detJ*invJ*normalFace_{kfe} ||
+              real64 tmp[3] = { 0 };
```
corbett5, line 296: You don't need three temporary variables here, just one.

PR #1471, opened 2021-06-25
Implemented MGR recipes for flow solvers (hybrid FVM, wells)

francoishamon:
This PR implements four MGR recipes that we have used over the past months, but that have not been merged into `develop` yet:
- [X] SinglePhaseFVM with wells
- [X] SinglePhaseHybridFVM
- [X] SinglePhaseHybridFVM with wells
- [X] CompositionalMultiphaseHybridFVM with wells

and the PR also creates a separate file for the hydrofracture MGR recipe.

I will add the recipes for (wells + poromechanics) in another PR because they require more changes in the physics solvers.

No rebaseline necessary.


castelletto1 (approved):
This looks great! Just a suggestion. With the current MGR strategies organization, I think it is cleaner to have a dedicated file for each solver. At the moment, both the `SinglePhasePoromechanicsSolver` and the `HydrofractureSolver` share the same strategy: [`SinglePhasePoromechanics`](https://github.com/GEOSX/GEOSX/blob/develop/src/coreComponents/linearAlgebra/interfaces/hypre/mgrStrategies/SinglePhasePoromechanics.hpp) (the only difference being the way the coarse grid is computed). I propose we introduce the `HydroFracture` strategy in another file. This will give us more flexibility if we need additional tuning of the MGR algorithm.

PR #1472, opened 2021-06-25
Improved well error/warning messages 

### LINKED ISSUE #1469: [Bug] Perforation transmissibility and perforation moving in parallel run
Please find the attached test dataset with one horizontal well(3 perforations)
[test_hwells.zip](https://github.com/GEOSX/GEOSX/files/6713585/test_hwells.zip)

test1: non parallel run geosx -i test_hwells1.xml 
image
image
CO2 is not supposed to be injected through perforation 2 because second perforation has transmissiblity=0

test2: same dataset with transmissiblity =0 for perforation 2  and run it in parallel:  mpirun -np 9 geosx -i test_hwells1.xml -x 3 -y 3 -z 1
paraview reports error in calculation and no results found. It seems the "transmissiblity =0" crashed vtk results in parallel run mode

test3: remove transmissibility  from perforation 2 and run it in parallel:  mpirun -np 9 geosx -i test_hwells1.xml -x 3 -y 3 -z 1
image
Geosx log shows
"Moving perforation #1 of well wellInjector1 from { 5025, 4824, 28.82 } to { 5025, 5024, 28.82 } to make sure that no well element is shared between two MPI ranks
Moving perforation #2 of well wellInjector1 from { 5025, 4624, 28.82 } to { 5025, 5024, 28.82 } to make sure that no well element is shared between two MPI ranks
".
and only one perforation left in reservoir
image

This is an unexpected behavior. Perforations are not supposed to be moved(merged?) because the whole well doesn't cross partitions.
image

Please review and fix it in future release if it's confirmed as bug.

francoishamon:
Fixes #1469 

This PR clarifies the warning that is output when a user tries to define a well with multiple perforations in the same well element message. The current message is a little bit cryptic, so I propose the following one:
```
The GEOSX wells currently have the following limitation in parallel:                                                                                                                                            
We cannot allow an element of the well mesh to have two or more perforations associated with it.                                                                                                                
So, in the present simulation, perforation #2 of well wellInjector1 is moved from { 5025, 4624, 28.82 } to { 5025, 5024, 28.82 } to make sure that no element of the well mesh has two perforations associated \
with it.                                                                                                                                                                                                        
To circumvent this issue, please increase the value of "numElementsPerSegment" that controls the number of (uniformly distributed) well mesh elements per segment of the well polyline.                         
Our recommendation is to choose "numElementsPerSegment" such that each well mesh element has at most one perforation.
```
The reason for not allowing two perforations in the same well mesh element is that these two perforations could be connected to two reservoir cells on different MPI partitions, which would break the logic of the well mesh partitioning.

This PR also clarifies the behavior of the well generator when the user provides a well transmissibility equal to zero (which we do not encourage). Currently, if a user sets the transmissibility to zero, GEOSX is silently recomputing the transmissibility using the Peaceman formula. With this PR, GEOSX will output the following warning:
```
Warning! A perforation is defined with a zero transmissibility!                                                                                                                                                
The simulation is going to proceed with this zero transmissibility,                                                                                                                                             
but a better strategy to shut down a perforation is to remove to <Perforation> block from the XML
```
and proceed with the simulation.

### DIFF
--- a/src/coreComponents/mesh/PerforationData.cpp
+++ b/src/coreComponents/mesh/PerforationData.cpp
@@ -126,8 +126,12 @@ void PerforationData::computeWellTransmissibility( MeshLevel const & mesh,
 
     // if the well transmissibility has been read from the XML
     // then skip the computation of the well transmissibility carried out below
-    if( m_wellTransmissibility[iperf] > 0 )
+    if( m_wellTransmissibility[iperf] >= 0 )
     {
+      GEOSX_LOG_RANK_IF( isZero( m_wellTransmissibility[iperf] ),
+                         "\n \nWarning! A perforation is defined with a zero transmissibility! \n"
+                         << "The simulation is going to proceed with this zero transmissibility,\n"
+                         << "but a better strategy to shut down a perforation is to remove the <Perforation> block from the XML\n \n" );
       continue;
     }
 
--- a/src/coreComponents/mesh/PerforationData.cpp
+++ b/src/coreComponents/mesh/PerforationData.cpp
@@ -128,10 +128,11 @@ void PerforationData::computeWellTransmissibility( MeshLevel const & mesh,
     // then skip the computation of the well transmissibility carried out below
     if( m_wellTransmissibility[iperf] >= 0 )
     {
+      WellElementRegion const & wellRegion = dynamicCast< WellElementRegion const & >( wellElemSubRegion.getParent().getParent() );
       GEOSX_LOG_RANK_IF( isZero( m_wellTransmissibility[iperf] ),
-                         "\n \nWarning! A perforation is defined with a zero transmissibility! \n"
-                         << "The simulation is going to proceed with this zero transmissibility,\n"
-                         << "but a better strategy to shut down a perforation is to remove the <Perforation> block from the XML\n \n" );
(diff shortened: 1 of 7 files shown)


src/coreComponents/mesh/PerforationData.cpp
```diff
@@ -126,8 +126,12 @@ void PerforationData::computeWellTransmissibility( MeshLevel const & mesh,
 
     // if the well transmissibility has been read from the XML
     // then skip the computation of the well transmissibility carried out below
-    if( m_wellTransmissibility[iperf] > 0 )
+    if( m_wellTransmissibility[iperf] >= 0 )
     {
+      GEOSX_LOG_RANK_IF( isZero( m_wellTransmissibility[iperf] ),
+                         "\n \nWarning! A perforation is defined with a zero transmissibility! \n"
```
TotoGaz, line 132: Is there a way to tell which one?
If not, it's OK.

TotoGaz (approved):

src/coreComponents/mesh/PerforationData.cpp
francoishamon: Yes, I modified this error message to show which well has the zero transmissibility. I also updated the error messages in various places in the well construction code.

francoishamon:
I am merging this PR since it is primarily about error messages and does not break the integrated tests.

PR #1486, opened 2021-07-09
Consider confining stress on SIF calculation and fracture propagation

WuHuiLLNL:
In the presence of confining stress, a negative aperture might be possible in the SIF calculation of the SurfaceGenerator.cpp. Since the SIF equation (tipNodeSIF = pow( (fabs...) always give a positive value, the SIF will be extremely large under high confining stress, and the fracture will propagate, which is unphysical.

We manually multiply the SIF by -1 in case of negative aperture to avoid fracture propagation under very high confining stress.


rrsettgast (approved):

src/coreComponents/physicsSolvers/surfaceGeneration/SurfaceGenerator.cpp
```diff
@@ -3042,6 +3042,11 @@ void SurfaceGenerator::calculateNodeAndFaceSif( DomainPartition & domain,
           tipNodeSIF = pow( (fabs( tipNodeForce[0] * trailingNodeDisp[0] / 2.0 / tipArea ) + fabs( tipNodeForce[1] * trailingNodeDisp[1] / 2.0 / tipArea )
                              + fabs( tipNodeForce[2] * trailingNodeDisp[2] / 2.0 / tipArea )), 0.5 );
 
+          if ( LvArray::tensorOps::AiBi< 3 >( trailingNodeDisp, faceNormalVector ) < 0.0)  //In case the aperture is negative with the presence of confining stress.
```
rrsettgast, line 3045:
```suggestion
          if( LvArray::tensorOps::AiBi< 3 >( trailingNodeDisp, faceNormalVector ) < 0.0 )  //In case the aperture is negative with the presence of confining stress.
```

PR #1487, opened 2021-07-11
Proposal for type dispatch

klevzoff:
This PR proposes a replacement for `rtTypes::ArrayTypeLambda[1,2]`. It doesn't go as far as https://github.com/GEOSX/GEOSX/issues/844 (inverting the dispatch from types to operations), but gets rid of a single point of dispatch in favor of allowing callers to customize their dispatch type lists however they want. This approach inspired in part by [VTK array dispatch](https://vtk.org/doc/nightly/html/VTK-7-1-ArrayDispatch.html). It also gets rid of `rtTypes::TypeIDs` enum.

In short, it consists of two parts:
  * A generic dispatch function that accepts a list of types to dispatch on, a runtime `std::type_index` and a lambda. I've implemented two methods of dispatch: a function pointer table approach (similar to `std::visit` implementations, but simpler because we only have one type to dispatch on), and a simple recursive approach with nested `if` checks; we can decide which method works better for us.
  * Facilities for callers to easily construct type lists to provide to the dispatch function, as well as some "standard" type lists matching what `ArrayTypeLambda` used to have. To that end, I use `camp`'s metaprogramming tools to construct type lists like `camp::list< Array< real64, 2, RAJA::PERM_IJ >, ... >` from just a list of desired value types and dimension numbers. Most callers will only ever use "standard" type list aliases like `types::RealArrays` (e.g. for boundary conditions) or `types::StandardArrays` (for most other generic array handling), but this leaves opportunity for customization.

Typical usage is not much different from current, aside from providing a type list as an extra argument:
```c++
types::dispatch( types::StandardArrays{}, wrapper.getTypeId(), true, [&]( auto array )
{
  using ArrayType = decltype( array );
  // cast and do things with typed wrapper/array
}
```

Compile times (`time make -j8 geosx` from clean build dir):
  * develop
```
real	3m32.800s
user	22m50.415s
sys	0m47.037s
```
  * this PR
```
real	3m16.258s
user	21m27.504s
sys	0m45.222s
```

Resolves https://github.com/GEOSX/GEOSX/issues/756


src/coreComponents/common/TypeDispatch.hpp
```diff
@@ -0,0 +1,341 @@
...
+  return false;
+}
+
+template< typename LAMBDA >
+bool dispatchViaTable( TypeList<>,
+                       std::type_index const type,
+                       LAMBDA && lambda )
+{
+  // Needed to avoid constructing an zero-size array
+  GEOSX_UNUSED_VAR( type, lambda )
```
corbett5, line 279: Do we actually use this somewhere?

src/coreComponents/common/TypeDispatch.hpp
```diff
@@ -0,0 +1,341 @@
...
+  return false;
+}
+
+template< typename ... Ts, typename LAMBDA >
+std::enable_if_t< ( sizeof...( Ts ) <= 32 ), bool >
```
corbett5, line 284: I feel like we should pick one and just use it. I don't really care which one.

src/coreComponents/fieldSpecification/FieldSpecificationBase.hpp
```diff
@@ -676,261 +562,19 @@ void FieldSpecificationBase::applyFieldValue( SortedArrayView< localIndex const
...
-                                                             typename LAI::ParallelMatrix & matrix,
-                                                             typename LAI::ParallelVector & rhs ) const
-{
-  dataRepository::WrapperBase const & wrapperBase = dataGroup.getWrapperBase( fieldName );
-  std::type_index typeIndex = std::type_index( wrapperBase.getTypeId());
-  arrayView1d< globalIndex const > const & dofMap = dataGroup.getReference< array1d< globalIndex > >( dofMapName );
-  integer const component = getComponent();
 
-  rtTypes::applyArrayTypeLambda1( rtTypes::typeID( typeIndex ), [&]( auto type )
+  // This function is used in setting boundary/initial conditions on simulation fields.
```
corbett5, line 566: This is so much better.

corbett5 (approved):
This is awesome! Now I just gotta read up on all those nifty camp features...

rrsettgast (approved):
@klevzoff wins infrastructure PR of the month.

src/coreComponents/common/TypeDispatch.hpp
```diff
@@ -0,0 +1,341 @@
...
+ *
+ * Collection of tools to help dispatch templated functions on types
+ */
+
+#ifndef GEOSX_COMMON_TYPEDISPATCH_HPP
+#define GEOSX_COMMON_TYPEDISPATCH_HPP
+
+#include "common/DataTypes.hpp"
+
+#include <camp/camp.hpp>
```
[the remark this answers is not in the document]
corbett5: TPL's should be with `<>`. AFAIK the only real difference is that warnings are silenced for them.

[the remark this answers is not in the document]
klevzoff: We seem to have a mix of both styles sprinkled around various parts of the code (even within the same component, e.g. `linearAlgebra`). If we were to standardize/cleanup these, I would vote for `<>` because it searches in `-I` directories only which seems appropriate for TPLs, whereas `""` also searches in current and parent directories.
https://gcc.gnu.org/onlinedocs/cpp/Include-Syntax.html

src/coreComponents/common/TypeDispatch.hpp
klevzoff: I would pick table dispatch then, since it has more predictable runtime (and probably compile-time) cost and better stack traces.

src/coreComponents/common/TypeDispatch.hpp
klevzoff: Only in unit tests :)

corbett5: If we don't see a need for it in the main code I'd replace the body of the function with a static_assert, just so that you get a decent compilation error instead of something gross.

src/coreComponents/common/TypeDispatch.hpp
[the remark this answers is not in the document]
corbett5: The one thing is that sometimes it takes a bit of manipulation to force a TPL added from CMake's `find_package` to treat it's includes as system. I think it's worth the effort.

PR #1489, opened 2021-07-13
Unify interfaces of FEM and VEM classes

andrea-borio:


src/coreComponents/finiteElement/elementFormulations/FiniteElementBase.hpp
```diff
@@ -79,18 +82,117 @@ class FiniteElementBase
...
+   */
+  struct Initialization
+  {
+    /**
+     * Constructor
+     */
+    GEOSX_HOST_DEVICE
+    Initialization()
+    {}
+  };
```
francoishamon, line 114: Given what you are going to store in this `struct` (geometry and topology of all the cells) for VEM, something like `MeshData` or `MeshInfo` would be a more descriptive name. And then you would keep the `initialize` for the function that fills this mesh data. Or do you think this `struct` could later also be used for non-mesh data in the future?

src/coreComponents/finiteElement/elementFormulations/FiniteElementBase.hpp
```diff
@@ -79,18 +82,117 @@ class FiniteElementBase
...
+  /**
+   * @struct Initialization
+   * @brief Variables used to initialize the class.
+   */
+  struct Initialization
+  {
+    /**
+     * Constructor
+     */
+    GEOSX_HOST_DEVICE
```
francoishamon, line 111: Is this flag needed on the constructor? Do you plan to instantiate this object in a kernel? If not you can probably remove it

src/coreComponents/finiteElement/elementFormulations/H1_Hexahedron_Lagrange1_GaussLegendre2.hpp
```diff
@@ -86,16 +88,69 @@ class H1_Hexahedron_Lagrange1_GaussLegendre2 final : public FiniteElementBase
...
+
+  /**
+   * @brief Empty initialization method.
+   * @param nodeManager The node manager.
+   * @param edgeManager The edge manager.
+   * @param faceManager The face manager.
+   * @param cellSubRegion The cell sub-region for which the element has to be initialized.
+   * @param initialization Object that holds initialization properties.
+   */
+  GEOSX_HOST_DEVICE
```
francoishamon, line 135: Here, if I understood well how the initialization `struct` is going to be used, you may not need this flag either.

src/coreComponents/unitTests/virtualElementTests/testConformingVirtualElementOrder1.cpp
```diff
@@ -176,51 +205,33 @@ static void testCellsInMeshLevel( MeshLevel const & mesh )
...
-                                                cellToNodeMap, cellCenters,
-                                                basisData );
-    checkSumOfQuadratureWeights< VEM >( cellVolumes[cellIndex],
-                                        basisData );
-  }
+    checkIntegralMeanConsistency< VEM >( virtualElement, stack );
+    checkIntegralMeanDerivativesConsistency< VEM >( virtualElement, stack );
+    checkStabilizationMatrixConsistency< VEM >( nodesCoords, cellIndex, cellToNodeMap,
+                                                cellCenters, virtualElement, stack );
+    checkSumOfQuadratureWeights< VEM >( cellVolumes[cellIndex], stack );
```
francoishamon, line 230: If these checks fail when CUDA is on, is this unit test actually going to report a failure? I am asking because I noticed that you `ifndef`'ed out all the `EXPECT_TRUE` in these functions. I am not sure, it is just something to double-check.

src/coreComponents/virtualElement/ConformingVirtualElementOrder1.hpp
```diff
@@ -169,8 +307,8 @@ class ConformingVirtualElementOrder1 final : public VirtualElementBase
           diameter = candidateDiameter;
         }
       }
-      diameter = LvArray::math::sqrt< real64 >( diameter );
     }
+    diameter = LvArray::math::sqrt< real64 >( diameter );
     return diameter;
```
francoishamon, line 312:
```suggestion
    return LvArray::math::sqrt< real64 >( diameter );
```

src/coreComponents/virtualElement/ConformingVirtualElementOrder1.hpp
```diff
@@ -197,11 +335,64 @@ class ConformingVirtualElementOrder1 final : public VirtualElementBase
...
```
francoishamon, line 340:
```suggestion
    return LvArray::math::sqrt< real64 >( diameter );
```

francoishamon (approved):
These changes look good to me (but I only have a superficial knowledge of the finite-element library). Thanks @andrea-borio

PR #1497, opened 2021-07-16
XML Redundancy Tool

cssherman:


cssherman:
The updated xml files in this PR need to be checked to see if we want to add any additional attributes to the redundancy whitelist:

- component (this shows up a lot where we apply a BC to the 0th component of a vector)

rrsettgast:
Nice!

sytuannguyen:
This xml tool is so nice! Is it possible to generate xml from simple geometrical/physical description of a given problem?

joshua-white (approved):
This is awesome.

PR #1499, opened 2021-07-17
Replace element type strings with enum

klevzoff:
Resolves https://github.com/GEOSX/GEOSX/issues/1088


src/coreComponents/fileIO/Outputs/BlueprintOutput.cpp
```diff
@@ -38,17 +38,54 @@ namespace internal
...
+    { return { 1, 0, 2, 3 };}
+    case ElementType::Pyramid:
+    { return { 0, 3, 2, 1, 4, 0, 0, 0 };}
+    case ElementType::Prism:
+    { return { 0, 4, 2, 1, 5, 3, 0, 0 };}
+    case ElementType::Hexahedron:
+    { return { 0, 1, 3, 2, 4, 5, 7, 6 };}
+    case ElementType::Polyhedron:
+    { return { 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12 };                             // TODO
+    }
```
klevzoff, line 82: Uncrustify is absolutely ruining it.. objections to allowing un-braced case statements? E.g.
```c++
case A: return 1;
case B: return 2;
```

src/coreComponents/mesh/ElementSubRegionBase.hpp
```diff
@@ -19,19 +19,49 @@
...
+              "BEAM",
+              "C2D3",
+              "C2D4",
+              "Polygon",
+              "C3D4",
+              "C3D5",
+              "C3D6",
+              "C3D8",
+              "Polyhedron" );
+
```
rrsettgast, line 64: Do you think this would be better in its own header? I don't know that this information would ever be useful outside the context of the subregion?

rrsettgast (approved):

src/coreComponents/mesh/ElementSubRegionBase.hpp
klevzoff: I moved it into its own header. It can probably reduce the include graph in a few places... for example in `FiniteElementDiscretization.hpp`

src/coreComponents/fileIO/Outputs/BlueprintOutput.cpp
rrsettgast: I think I was trying to avoid unbraced multi-line case statements. I don't think there is any sort of real issue with relaxing the formatting for this...I recall wanting to be consistent with always having if-statments braced. I have no objection to what you have done.

Issue #1510, opened 2021-07-22
[Bug] Time history output for multi-component quantities (array2d, etc)


francoishamon:
**Describe the bug**
I am trying to use the time history output feature to output the saturation (`phaseVolumeFraction`, stored in an array2d) into an hdf5 file. GEOSX crashes with the following stack trace message:
```
** StackTrace of 12 frames **                                                                                                                  
Frame 0: LvArray::system::stackTraceHandler(int, bool) /g/g90/hamon1/geosx/GEOSX/src/coreComponents/LvArray/src/system.cpp:478 (discriminator 1)                                                                                                                                             
Frame 1: /g/g90/hamon1/geosx/GEOSX/build-quartz-gcc@8.1.0-debug/bin/geosx /g/g90/hamon1/geosx/GEOSX/src/coreComponents/common/initializeEnvironment.cpp:82                                                                                                                                   
Frame 2: /g/g90/hamon1/geosx/GEOSX/build-quartz-gcc@8.1.0-debug/bin/geosx /g/g90/hamon1/geosx/GEOSX/src/coreComponents/common/initializeEnvironment.cpp:82                                                                                                                                   
Frame 3: /lib64/libc.so.6                                                                                                                      
Frame 4: geosx::bufferOps::PackDataByIndexDevice<true, double, 2, 1, LvArray::ArrayView<long const, 1, 0, long, LvArray::ChaiBuffer> >(signed char*&, LvArray::ArrayView<double const, 2, 1, long, LvArray::ChaiBuffer> const&, LvArray::ArrayView<long const, 1, 0, long, LvArray::ChaiBuffer> const&, std::vector<camp::resources::v1::HostEvent, std::allocator<camp::resources::v1::HostEvent> >&)::{lambda(long)#1}::operator()(long) const::{lambda(double const&)#1}::operator()(double const) const /g/g90/hamon1/geosx/GEOSX/src/coreComponents/dataRepository/BufferOpsDevice.cpp:173                                                                                                                                        
[...]

PR #1512, opened 2021-07-23
Moved resizeFields to registerDataOnMesh in compositional solvers

francoishamon:
This PR moves the `resizeFields` function to `registerDataOnMesh` in the compositional solvers (flow and well) to make sure that the data is already resized when the time history output kicks in. 

It solves #1510.

This PR does not break the integrated tests, so there is no rebaseline necessary.


klevzoff (approved):
As an option, we could also merge `resizeFields()` with `registerDataOnMesh()` so we don't have to get the wrapper twice (just add `.reference().resizeDimension<>()` to existing call chains). Obviously, we'd need to extract dimensions from fluid model beforehand. This would be more work, but the code would get a bit smaller.

wrtobin (approved):
I'll see about adding an assert to the time history setup to explicitly catch this issue in case it comes up again.

francoishamon:
> As an option, we could also merge `resizeFields()` with `registerDataOnMesh()` so we don't have to get the wrapper twice (just add `.reference().resizeDimension<>()` to existing call chains). Obviously, we'd need to extract dimensions from fluid model beforehand. This would be more work, but the code would get a bit smaller.

Done, I merged these two functions and removed `resizeFields`.

Issue #1508, opened 2021-07-21
Outstanding issues following introduction of CoupleSolid, PorousSolid and CompressibleSolid.


CusiniM:
The introduction of  `CoupledSolid`, `PorousSolid` and `CompressibleSolid` has changed the way physics solvers interact (or should interact) with the constitutive models. The change of behaviour has only partially been addressed in PR #1455 and some outstanding issues remain:

- [x]  force flow solvers to interface with a `CompressibleSolid` to get porosity, permeability etc and remove the old wrappers for porosity, oldPorosity and referencePorosity.
- [x] Modify flow kernels to use porosity instead of `pvMultiplier` (addressed in #1311)
- [x] Create a new assembly kernel for the `PoromechanicsEmebeddedFracturesSolver` which assembles all cell-based terms, uses the `PorousSolid` and reactivate the corresponding integrated test (addressed in #1311)
- [x] make `biotCoefficient` a property varying cell by cell (addressed in #3050).
- [x] remove `PoreVolumeCompressibleSolid`
- [x] remove the old `PoroElastic` model which is still used by some solvers (i.e., `hydrofractureSolver`)
- [x] Everything new will have to be documented.

TotoGaz:
@CusiniM is this finally done? If not, can you manage its impact/urgency/...?

paveltomin:
@CusiniM @joshua-white @jhuang2601 can this be closed now?

jhuang2601:
> @CusiniM @joshua-white @jhuang2601 can this be closed now?

I updated the task on `biotCoefficient` and it can be closed.

PR #1521, opened 2021-07-28
Biot's tensor

sytuannguyen:
This PR adds a function for computing Biot's tensor from the drained tiffness of the porous rock and the elastic moduli of the solid grain. This address an issues listed in #1508.  An approximation for Biot's coefficient of transversely isotropic rock is also given.

### DIFF
--- a/src/coreComponents/constitutive/solid/PorousSolid.hpp
+++ b/src/coreComponents/constitutive/solid/PorousSolid.hpp
@@ -91,7 +91,7 @@ public:
     m_porosityUpdate.updateBiotCoefficient( k, bulkModulus );
   }
 
-private:
+protected:
 
   using CoupledSolidUpdates< SOLID_TYPE, BiotPorosity, ConstantPermeability >::m_solidUpdate;
   using CoupledSolidUpdates< SOLID_TYPE, BiotPorosity, ConstantPermeability >::m_porosityUpdate;
--- a/src/coreComponents/constitutive/solid/PorousSolid.hpp
+++ b/src/coreComponents/constitutive/solid/PorousSolid.hpp
@@ -85,26 +85,10 @@ public:
   GEOSX_HOST_DEVICE
   void updateBiotCoefficient( localIndex const k ) const
   {
+    // This call is not general like this.
+    real64 const bulkModulus = m_solidUpdate.getBulkModulus( k );
 
-    string solidType = SOLID_TYPE::m_catalogNameString;
-
-    if( solidType == "ElasticIsotropic" )
-    {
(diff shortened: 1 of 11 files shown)


src/coreComponents/constitutive/solid/PorousSolid.hpp
```diff
@@ -85,10 +85,26 @@ class PorousSolidUpdates : public CoupledSolidUpdates< SOLID_TYPE, BiotPorosity,
...
+    {
+      real64 const c11 = m_solidUpdate.getC11( k );
+      real64 const c13 = m_solidUpdate.getC13( k );
+      real64 const c33 = m_solidUpdate.getC33( k );
+      real64 const c66 = m_solidUpdate.getC66( k );
+
+      real64 const c12 = c11 - 2.0 * c66;
+
+      m_porosityUpdate.updateBiotCoefficient( k, c11, c12, c13, c33 );
+    }
```
CusiniM, line 107: This functions gets called inside the poromechanics kernels in a loop over quadrature points nested inside a loop over elements. We need to find a way to implement this without using this `if statement`.

src/coreComponents/constitutive/solid/porosity/BiotPorosity.hpp
```diff
@@ -89,13 +92,74 @@ class BiotPorosityUpdates : public PorosityBaseUpdates
...
+
   arrayView1d< real64 > m_biotCoefficient;
 
   real64 m_grainBulkModulus;
+
+  real64 m_grainShearModulus;
+
+  arrayView1d< real64 > m_b11; ///< (1,1) element of TI Biot's matrix
+  arrayView1d< real64 > m_b22; ///< (2,2) element of TI Biot's matrix
+  arrayView1d< real64 > m_b33; ///< (3,3) element of TI Biot's matrix
```
CusiniM, line 162: this biot's matrix is only used to compute the biotCoefficient. I don't think we need to store it.

src/coreComponents/constitutive/solid/PorousSolid.hpp
sytuannguyen: @CusiniM We may need to add a separated class for the PorousElasticTransverseIsotropic model. This is a simple fix but the code will be douplicated. I will try to implement this option, please take a look and give me your opinion

src/coreComponents/constitutive/solid/porosity/BiotPorosity.hpp
sytuannguyen: @CusiniM For transversely isotropic or orthotropic rock, we should use the anisotropic Biot's tensor to update stress and porosity

PR #1532, opened 2021-08-06
Refactor PAMELA field import

klevzoff:
This PR changes the order of field data import from external meshes. Instead of importing data together with the mesh into CellBlocks and creating new wrappers for each imported field, the data import step is delayed until after the entire data repository has been initialized (see `ProblemManager::problemSetup() and `ProblemManager::importFields()`). This means each imported field must match a field already existing in GEOSX (typically allocated by a physics solver or a constitutive model), and creation of new fields is not allowed at this point.

I've also somewhat refactored the field import code in `PAMELAMeshGenerator` using type dispatch, which is necessary to handle array layout permutations correctly. The import is also now aware of the distinction between normal and material fields - the latter have an extra quadrature point dimension, and we copy the same value into each quadrature point when importing.

Finally, the PR adds the quadrature point dimension to permeability, which is a material field but did not have it previously. This is necessary for field cases and benchmarks that import permeability (egg, SPE10, etc.) to work.

@AntoineMazuyer this adds more work to you VTK mesh reader, although you don't have to tackle it all in the same PR. It's ok for https://github.com/GEOSX/GEOSX/pull/1453 to get merged with just the mesh generation capability, and refactor the property import later.

@francoishamon and @AntoineMazuyer this should fix external mesh workflows for users, but be aware that they need to specify target field for import as `rockPerm_permeability` where `rockPerm` is the name of permeability model.

Related to https://github.com/GEOSX/integratedTests/pull/142

### DIFF
--- a/src/coreComponents/constitutive/permeability/CarmanKozenyPermeability.cpp
+++ b/src/coreComponents/constitutive/permeability/CarmanKozenyPermeability.cpp
@@ -51,12 +51,10 @@ CarmanKozenyPermeability::deliverClone( string const & name,
 void CarmanKozenyPermeability::allocateConstitutiveData( dataRepository::Group & parent,
                                                          localIndex const numConstitutivePointsPerParentIndex )
 {
-  m_dPerm_dPorosity.resize( 0, 3 );
+  // NOTE: enforcing 1 quadrature point
+  m_dPerm_dPorosity.resize( 0, 1, 3 );
   PermeabilityBase::allocateConstitutiveData( parent, numConstitutivePointsPerParentIndex );
 }
-
-void CarmanKozenyPermeability::postProcessInput()
-{}
 
 REGISTER_CATALOG_ENTRY( ConstitutiveBase, CarmanKozenyPermeability, string const &, Group * const )
 
--- a/src/coreComponents/mainInterface/ProblemManager.cpp
+++ b/src/coreComponents/mainInterface/ProblemManager.cpp
@@ -164,6 +164,8 @@ void ProblemManager::problemSetup()
   registerDataOnMeshRecursive( getDomainPartition().getMeshBodies() );
 
   initialize();
+
+  importFields();
 }
 
 
@@ -558,6 +560,19 @@ void ProblemManager::generateMesh()
 }
 
 
+void ProblemManager::importFields()
+{
+  GEOSX_MARK_FUNCTION;
+  DomainPartition & domain = getDomainPartition();
+  MeshManager & meshManager = this->getGroup< MeshManager >( groupKeys.meshManager );
+
+  meshManager.forSubGroups< MeshGeneratorBase >( [&]( MeshGeneratorBase & generator )
+  {
+    generator.importFields( domain );
+    generator.freeResources();
+  } );
+}
+
 void ProblemManager::applyNumericalMethods()
 {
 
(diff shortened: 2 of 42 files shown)


src/coreComponents/mesh/generators/PAMELAMeshGenerator.cpp
```diff
@@ -61,319 +56,435 @@ PAMELAMeshGenerator::PAMELAMeshGenerator( string const & name, Group * const par
...
+
+      // Find source
+      PAMELA::VariableDouble * const meshProperty = regionPtr->FindVariableByName( sourceName );
+      GEOSX_ASSERT( meshProperty != nullptr );
+
+      // Find destination
+      WrapperBase & wrapper = subRegion.getWrapperBase( wrapperName );
+
+      // Decide if field is constitutive or not
+      if( materialWrapperNames.count( wrapperName ) > 0 )
```
CusiniM, line 478: Do we maybe want to throw an error if the property does not already exist in GEOSX?

CusiniM (approved):
Looks great to me!

src/coreComponents/mesh/generators/PAMELAMeshGenerator.cpp
klevzoff: `getWrapperBase()` will throw an error here, do you mean throwing a more user-friendly error? That sounds like a good idea.

src/coreComponents/constitutive/permeability/CarmanKozenyPermeability.cpp
```diff
@@ -51,13 +51,11 @@ CarmanKozenyPermeability::deliverClone( string const & name,
...
                                                          localIndex const numConstitutivePointsPerParentIndex )
 {
-  m_dPerm_dPorosity.resize( 0, 3 );
+  // NOTE: enforcing 1 quadrature point
+  m_dPerm_dPorosity.resize( 0, 1, 3 );
   PermeabilityBase::allocateConstitutiveData( parent, numConstitutivePointsPerParentIndex );
 }
 
-void CarmanKozenyPermeability::postProcessInput()
-{}
```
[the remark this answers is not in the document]
CusiniM: it's fine. Many of these things have changed in my other PR.

src/coreComponents/mainInterface/ProblemManager.cpp
```diff
@@ -558,6 +560,19 @@ void ProblemManager::generateMesh()
...
+void ProblemManager::importFields()
+{
+  GEOSX_MARK_FUNCTION;
+  DomainPartition & domain = getDomainPartition();
+  MeshManager & meshManager = this->getGroup< MeshManager >( groupKeys.meshManager );
+
+  meshManager.forSubGroups< MeshGeneratorBase >( [&]( MeshGeneratorBase & generator )
+  {
+    generator.importFields( domain );
+    generator.freeResources();
```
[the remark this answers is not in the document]
CusiniM: I think that ideally we would like to completely delete the generator but this is probably good enough for now.

src/coreComponents/mesh/generators/PAMELAMeshGenerator.cpp
CusiniM: Yeah, I meant like a more user-friendly error than what comes out of `getWrapperBase()`.

PR #1468, opened 2021-06-23
Implemented time-dependent well rate and bhp tables

tang39:
This PR ended up doing more than the original plan. Hewei modified the well controls and relevant files to accept phase rate, total rate and BHP as tables. Then a clean up in the well kernels was done to write them in the style of the flow kernels (launch + compute format). A lot of functions have been moved in the process, but there is not a lot of new code.

- [X] Implement time-dependent well rate / BHP
- [X] Modify the well solver to allow for zero well rate (addition of accumulation term + fixes in the residual normalization)
- [X] Cleanup in the well kernels (no new code, just moving to the format "launch + compute", with implementations in a cpp file)

Also fixes `examples/AdaptiveTimeStepping/deadoil_3ph_baker_1d.xml` that is crashing in the current `develop`

This PR requires rebaseline of the integrated tests in https://github.com/GEOSX/integratedTests/pull/149

PR #1537, opened 2021-08-20
Implement Black-Oil model on GPU

### LINKED ISSUE #1350: [Feature] Avoid division by zero when a fluid phase is absent
Working on the Black-Oil model, we found the following issue: if the gas phase is absent, its density and viscosity are not computed, which leads to divisions by zero in some places, including the constitutive model:
https://github.com/GEOSX/GEOSX/blob/80ab48fb27b53db20977903ea4743d31dcd0d501/src/coreComponents/constitutive/fluid/MultiFluidPVTPackageWrapper.cpp#L468
 as well as the flow solver:
https://github.com/GEOSX/GEOSX/blob/80ab48fb27b53db20977903ea4743d31dcd0d501/src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBaseKernels.cpp#L153
and in the well solver: 
https://github.com/GEOSX/GEOSX/blob/80ab48fb27b53db20977903ea4743d31dcd0d501/src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp#L660
This is could also happen in our current CO2-brine model, and is likely to happen when we implement a more complex CO2-brine model, or in true compositional simulations.

At least three possibilities to fix this:
1.  Force all the constitutive models to define a non-zero (potentially meaningless) value for all the phases, even if the phase fraction is zero in a given cell.
2. Set the density and viscosity of absent phases to zero in the constitutive model, and add checks on the phase density or phase fractions in the solver to make sure that we don't use them and that we don't divide by zero. 
3. Set the density and viscosity of absent phases to zero in the constitutive model, and pass around a `phaseStatus` vector (one value per cell) that encodes the phase state and can be used to deduce the existence/absence of a phase using a small `phaseCombinations` vector listing all the possibilities (as in AD-GPRS). 

@klevzoff @CusiniM

francoishamon:
This PR implements the Black-Oil model on GPU. It replaces the previous call to PVTPackage and implements all the compute functions inside GEOSX. For now, the water and gas phase updates use our native TableFunctions for interpolation, but the more complex oil phase updates are still based on "custom" tables.

- [X] Validation with Intersect
- [X] Works on GPU
- [x] Add unit test
- [x] Add integrated tests

The PR also slightly touches the flow/well solvers to avoid divisions by zero when the gas phase is absent. Fixes #1350 

Will require a rebaseline for the two Black-Oil integrated tests that will be added. Rebaselined in PR https://github.com/GEOSX/integratedTests/pull/146


src/coreComponents/constitutive/fluid/BlackOilFluidBase.hpp
```diff
@@ -0,0 +1,152 @@
...
+
+  /**
+   * @brief Create all the table kernel wrappers
+   */
+  virtual void createAllKernelWrappers();
+
+  /**
+   * @brief Use the TableFunctions provided by the user to get the PVT data
+   */
+  virtual void useProvidedTableFunctions() = 0;
```
TotoGaz, line 77: Maybe `buildPvtFromTables`?

TotoGaz (commented):
First wave

src/coreComponents/constitutive/fluid/BlackOilFluid.hpp
```diff
@@ -19,57 +19,1281 @@
...
+  // 4. Update phase fraction and phase component fractions
+
+  // 4.1 The gas phase is present
+  if( ( gasPhaseFraction > 0 ) && ( gasPhaseFraction < 1 ) )
+  {
+
+    // phase fractions
+    phaseFraction[PT::OIL] = 1. - gasPhaseFraction - zw;
+    phaseFraction[PT::GAS] = gasPhaseFraction;
+    phaseFraction[PT::WATER] =  zw;
```
klevzoff, line 710: Why is `phaseFraction` not using the same phase ordering as other output quantities (i.e. indexing though `ipGas`, `ipOil`, `ipWater`)?

src/coreComponents/constitutive/fluid/BlackOilFluid.hpp
```diff
@@ -19,57 +19,1281 @@
...
+  real64 const m_waterRefPressure;
+
+  /// Water formation volume factor
+  real64 const m_waterFormationVolFactor;
+
+  /// Water compressibility
+  real64 const m_waterCompressibility;
+
+  /// Water viscosity
+  real64 const m_waterViscosity;
```
klevzoff, line 320: There's no need to make member variables `const`. All the compute/update functions are already `const` so there's no danger of accidentally modifying a member.

src/coreComponents/constitutive/fluid/BlackOilFluid.hpp
```diff
@@ -19,57 +19,1281 @@
...
 
 namespace constitutive
 {
 
-class BlackOilFluid : public MultiFluidPVTPackageWrapper
+static constexpr real64 minForPhasePresence = 1e-10;
+
+static constexpr localIndex NC_BO = 3;
+static constexpr localIndex NP_BO = 3;
+static constexpr localIndex HNC_BO = NC_BO - 1;
```
klevzoff, line 37: I would prefer for these to be scoped in `BlackOilFluid` if they are primarily used in inside that class.

src/coreComponents/constitutive/fluid/BlackOilFluid.cpp
```diff
@@ -31,76 +26,443 @@ using namespace dataRepository;
...
+  std::sort( refineP.begin(), refineP.end() );
+  refineP.erase( std::unique( refineP.begin(), refineP.end() ), refineP.end() );
+
+  // linspace
+  real64 const maxVal = *std::max_element( refineP.begin(), refineP.end() );
+  array1d< real64 > refinePP( refineP.size() );
+  for( localIndex i = 0; i < localIndex( refineP.size() ); ++i )
+  {
+    refinePP[i] = refineP[i];
+  }
```
klevzoff, line 375: Is this part even needed? It looks like `refinePP` is being overwritten entirely by the code that follows.

src/coreComponents/constitutive/fluid/BlackOilFluid.cpp
```diff
@@ -31,76 +26,443 @@ using namespace dataRepository;
...
+  {
 
-  localIndex const NP = numFluidPhases();
+    // Bo
+    interpolation1( m_PVTO.m_undersaturatedPressure[i], m_PVTO.m_undersaturatedBo[i], refinePP, propRefined );
+    m_PVTO.m_undersaturatedBo[i].resize( ppSize );
+    for( localIndex j = 0; j < ppSize; ++j )
+    {
+      m_PVTO.m_undersaturatedBo[i][j] = propRefined[j];
+    }
```
klevzoff, line 394: I think loops like this can be replaced with just array assignment, i.e.
```
m_PVTO.m_undersaturatedBo[i] = propRefined;
```
or even using `m_PVTO.m_undersaturatedBo[i]` directly as the argument to `interpolation1`

src/coreComponents/constitutive/fluid/BlackOilFluid.cpp
```diff
@@ -31,76 +26,443 @@ using namespace dataRepository;
...
+    m_PVTO.m_undersaturatedViscosity[i].resize( ppSize );
+    for( localIndex j = 0; j < ppSize; ++j )
+    {
+      m_PVTO.m_undersaturatedViscosity[i][j] = propRefined[j];
+    }
+
+    for( localIndex j = 0; j < ppSize; ++j )
+    {
+      m_PVTO.m_undersaturatedPressure[i].resize( ppSize );
+      m_PVTO.m_undersaturatedPressure[i][j] = refinePP[j];
```
klevzoff, line 407: Is this resizing `m_PVTO.m_undersaturatedPressure[i]` on inner each loop iteration?

src/coreComponents/constitutive/fluid/BlackOilFluid.cpp
```diff
@@ -31,76 +26,443 @@ using namespace dataRepository;
...
+void BlackOilFluid::createUndersaturatedProperties()
 {
-  GEOSX_THROW_IF_NE_MSG( array.size(), expected,
-                         "BlackOilFluid: invalid number of entries in " << attr << " attribute",
-                         InputError );
+  localIndex upperBranchIndex = m_PVTO.m_nSaturatedPoints - 1;
+
+  // construct undersaturated branches by interpolation, start from second highest Rs
+
+  for( localIndex iCurrent = m_PVTO.m_nSaturatedPoints - 2; iCurrent-- > 0; )
```
klevzoff, line 235: Ah, the famous ["goes to" operator](https://stackoverflow.com/questions/1642028/what-is-the-operator-in-c-c)! I find this hard to process personally, something like
```suggestion
  for( localIndex iCurrent = m_PVTO.m_nSaturatedPoints - 2; iCurrent >= 0; --iCurrent )
```
is more natural :)

src/coreComponents/codingUtilities/Utilities.hpp
```diff
@@ -161,6 +161,160 @@ void applyChainRuleInPlace( localIndex const N,
...
+  for( iplus = 1; iplus < x.size() - 1 && x[iplus] < xval; ++iplus )
+  { }
+  iminus = iplus - 1;
+}
+
+template< typename T >
+void interpolation1( array1d< T > const & xin,
+                     array1d< T > const & yin,
+                     array1d< T > const & xout,
+                     array1d< T > & yout )
```
klevzoff, line 216: These should all be `arrayView1d< T (const) > const &` at least

src/coreComponents/codingUtilities/Utilities.hpp
```diff
@@ -161,6 +161,160 @@ void applyChainRuleInPlace( localIndex const N,
...
+
+    if( i_minus == i_plus )
+    {
+      yout[n] = yin[i_minus];
+    }
+    else
+    {
+      yout[n] = yin[i_minus] * ( xin[i_plus] - x ) / ( xin[i_plus] - xin[i_minus] )
+                + yin[i_plus] * ( ( x - xin[i_minus] ) / ( xin[i_plus] - xin[i_minus] ) );
+      GEOSX_THROW_IF( yout[n] < 0.0, "interpolation1: negative value", InputError );
```
klevzoff, line 237: The assumption here is that all `yin` are non-negative and therefore you can't have a negative interpolated value? This is something specific to the call site, not the interpolation routine itself. I recommend checking output values at the call site if desired, or not check them at all.

src/coreComponents/codingUtilities/Utilities.hpp
```diff
@@ -161,6 +161,160 @@ void applyChainRuleInPlace( localIndex const N,
...
+}
+
+template< typename T >
+void interpolation1( array1d< T > const & xin,
+                     array1d< T > const & yin,
+                     array1d< T > const & xout,
+                     array1d< T > & yout )
+{
+  GEOSX_THROW_IF( xin.size() != yin.size(), "interpolation1: size mismatch!", InputError );
+  GEOSX_THROW_IF( xout.size() != yout.size(), "interpolation1: size mismatch!", InputError );
```
klevzoff, line 219: I would make these non-throwing assertions (using `GEOSX_ASSERT_EQ` to get the size info if it fails). Sizes of user input are checked at runtime by the appropriate classes, it's not this function's job.

klevzoff (commented):
Nice, lots of work! Main just needs tidying up some stuff brought over from PVTPackage. Some of my comments might repeat @TotoGaz as I wrote most of them earlier this morning.

src/coreComponents/constitutive/fluid/BlackOilFluidBase.hpp
```diff
@@ -0,0 +1,152 @@
...
+
+  using exec_policy = parallelDevicePolicy<>;
+
+  struct PhaseType
+  {
+    static constexpr integer OIL            = 0;
+    static constexpr integer GAS            = 1;
+    static constexpr integer WATER          = 2;
+    static constexpr integer MAX_NUM_PHASES = 3;
+  };
```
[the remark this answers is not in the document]
klevzoff: We don't want a strongly typed enum because these constants are actually mainly used as indicies into arrays. On the other hand, we (almost?) never switch on them. But I do agree with the rest. I originally proposed `static constexpr integer` approach, but later I realized `enum` is a little better (especially the part about not requiring a definition). I've done a similar transformation in `RelativePermeabilityBase`, in case you want to keep it similar.

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWellKernels.hpp
```diff
@@ -941,40 +946,46 @@ struct PerforationKernel
...
-          resTotalMob     += resRelPerm / resVisc;
-          dResTotalMob_dP += ( dResRelPerm_dP * resVisc - resRelPerm * dResVisc_dP )
-                             / ( resVisc * resVisc );
-          for( localIndex ic = 0; ic < NC; ++ic )
-          {
-            dResTotalMob_dC[ic] += ( dRelPerm_dC[ic] * resVisc - resRelPerm * dVisc_dC[ic] )
-                                   / ( resVisc * resVisc );
+            for( localIndex jp = 0; jp < NP; ++jp )
+            {
+              real64 const dResRelPerm_dS = dResPhaseRelPerm_dPhaseVolFrac[er][esr][ei][0][ip][jp];
```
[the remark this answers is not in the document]
klevzoff: Yeah, sadly, the "influence" of FEM-based infrastructure (two of those dimensions are anyway).
One way to alleviate this slightly is to create local `[er][esr][ei][0]`-slices outside of inner (`ip`,`jp`) loops.

src/coreComponents/constitutive/fluid/BlackOilFluid.hpp
```diff
@@ -19,57 +19,1281 @@
...
+/**
+ * @brief Kernel wrapper class for BlackOilFluid
+ *        This kernel can be called on the GPU
+ */
+class BlackOilFluidUpdate final : public MultiFluidBaseUpdate
 {
 public:
 
-  using exec_policy = serialPolicy;
+  BlackOilFluidUpdate( arrayView1d< integer const > const & phaseTypes,
```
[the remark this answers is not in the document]
klevzoff: We could consider composition instead of inheritance for `Update` classes. I.e. `BlackOilFluidUpdate` would store an object of `MultiFluidBaseUpdate` and accept it in constructor. Thus all the members of that class (which make up 26 of the 36 parameters) would not need to be repeated here. We'd lose compiler-enforced virtual function signatures, but I feel more and more that we should get rid of `virtual` functions here anyway. We don't break these signatures often enough to justify `virtual` IMO.

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWellKernels.hpp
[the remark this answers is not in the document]
francoishamon: Now this is addressed in PR #1468, in which I replaced these 6-dimensional arrays with local slices.

src/coreComponents/constitutive/fluid/PVTFunctions/PVTFunctionHelpers.cpp
```diff
@@ -0,0 +1,163 @@
...
+    // Add and read a new line entry
+    data.emplace_back( 0 );
+    array1d< real64 > newLine = stringutilities::fromStringToArray< real64 >( str );
+    data.back() = newLine;
+
+    // Remove line entry of no data read
+    if( data.back().empty() )
+    {
+      data.pop_back();
+    }
```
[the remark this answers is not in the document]
francoishamon: Done

src/coreComponents/codingUtilities/StringUtilities.cpp
```diff
@@ -88,5 +88,42 @@ string_array tokenize( const string & str, const string & delimiters )
...
+    std::size_t const first = newStr.find_first_not_of( charsToRemove );
+    std::size_t const last = newStr.find_last_not_of( charsToRemove );
+    if( first != string::npos )
+    {
+      newStr = newStr.substr( first, ( last - first + 1 ) );
+    }
+    else
+    {
+      newStr.resize( 0 );
+    }
```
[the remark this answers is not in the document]
francoishamon: Done

src/coreComponents/constitutive/fluid/PVTFunctions/PVTFunctionHelpers.cpp
```diff
@@ -0,0 +1,163 @@
...
+  for( localIndex i = 0; i < data.size(); ++i )
+  {
+    GEOSX_ERROR_IF( data[i].size() < minRowLength,
+                    "BlackOilTables: too few entries in row " << i << " of table " << fileName
+                                                              << ", minimum " << std::to_string( minRowLength ) << " required" );
+  }
+}
+
+array1d< array1d< array1d< real64 > > >
+BlackOilTables::readAllTables( localIndex const ipWater,
```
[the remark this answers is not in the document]
francoishamon: I have not addressed this comment in the current PR, but I agree that it would be nice to consistify this at some point.

src/coreComponents/constitutive/fluid/BlackOilFluidBase.hpp
```diff
@@ -0,0 +1,152 @@
...
+    static constexpr char const * surfacePhaseMassDensitiesString() { return "surfaceDensities"; }
+    static constexpr char const * tableFilesString() { return "tableFiles"; }
+    static constexpr char const * formationVolumeFactorTableNamesString() { return "hydrocarbonFormationVolFactorTableNames"; }
+    static constexpr char const * viscosityTableNamesString() { return "hydrocarbonViscosityTableNames"; }
+    static constexpr char const * waterRefPressureString() { return "waterReferencePressure"; }
+    static constexpr char const * waterFormationVolumeFactorString() { return "waterFormationVolumeFactor"; }
+    static constexpr char const * waterCompressibilityString() { return "waterCompressibility"; }
+    static constexpr char const * waterViscosityString() { return "waterViscosity"; }
+
+  };
```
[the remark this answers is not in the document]
TotoGaz: While I respect your "_personal preference_", it's always interesting to get _facts_ in order to weigh your choices instead.

Let's consider the case of `surfacePhaseMassDensitiesString`.
* How do I know the type of the `surfacePhaseMassDensities`?
* Does it make sense to allow different types for `surfacePhaseMassDensities`?
* How do I prevent a developer to register some `surfacePhaseMassDensities` with different and/or inconsistent types?
* If we happen to change the main (unique?) type of `surfacePhaseMassDensities`, it will surely not break at compile time but at runtime with failed casts.

PR #1540, opened 2021-08-21
Prevent VTK counter FPE

### LINKED ISSUE #1529: [Bug] Crash when running FieldCaseTutorial1.xml
**Describe the bug**
MPI_ABORT was invoked on rank 0 in communicator MPI_COMM_WORLD
with errorcode 1.

NOTE: invoking MPI_ABORT causes Open MPI to kill all MPI processes.
You may or may not see output from other processes, depending on
exactly when Open MPI kills them.

**To Reproduce**
[error.pdf](https://github.com/GEOSX/GEOSX/files/6929089/error.pdf)

herve-gross:
This fixes #1529 by preventing a floating point exception in the `vtkXMLWriter::UpdateProgressDiscrete(float)` method. This behavior has only been observed on MacOS, hence the `#if`.

### DIFF
--- a/src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp
+++ b/src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp
@@ -760,7 +760,7 @@ void VTKPolyDataWriterInterface::write( real64 const time,
 #if defined(__APPLE__) && defined(__MACH__)
   LvArray::system::FloatingPointExceptionGuard guard;
 #endif
-  
+
   string const stepSubFolder = getTimeStepSubFolder( time );
   if( MpiWrapper::commRank( MPI_COMM_GEOSX ) == 0 )
   {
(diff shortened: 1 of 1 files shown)


klevzoff (approved):

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp
```diff
@@ -755,6 +755,12 @@ void VTKPolyDataWriterInterface::write( real64 const time,
                                         integer const cycle,
                                         DomainPartition const & domain )
 {
+  // This guard prevents crashes observed on MacOS due to a floating point exception
+  // triggered inside VTK by a progress indicator
+  #if defined(__APPLE__) && defined(__MACH__)
+    LvArray::system::FloatingPointExceptionGuard guard;
+  #endif
```
klevzoff, line 762:
```suggestion
#if defined(__APPLE__) && defined(__MACH__)
  LvArray::system::FloatingPointExceptionGuard guard;
#endif
```

PR #1543, opened 2021-08-24
Fix VTK crash with empty subregions

francoishamon:
This PR adds checks to avoid a division by zero when a subregion is empty on an MPI rank. This happens for example in the well subregions, and makes GEOSX crash. This is my bad, I should have added VTK output to the well integrated tests to catch errors like these. I propose to do it in this PR.

Requires a rebaseline for the integrated tests changed to VTK.

Related to https://github.com/GEOSX/integratedTests/pull/144

### DIFF
--- a/src/coreComponents/dataRepository/wrapperHelpers.hpp
+++ b/src/coreComponents/dataRepository/wrapperHelpers.hpp
@@ -683,7 +683,13 @@ int numArrayDims( T const & GEOSX_UNUSED_PARAM( var ) )
 template< typename T, int NDIM, int USD >
 localIndex numArrayComp( ArrayView< T const, NDIM, USD > const & var )
 {
-  return var.size() / var.size( 0 );
+  return LvArray::indexing::multiplyAll< NDIM - 1 >( var.dims() + 1 );
+}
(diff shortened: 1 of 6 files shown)


klevzoff (approved):
Hey @francoishamon , good catch! I found a couple more places with a similar pattern of dividing by `.size( 0 )`. I also wanted the code to set number of components and labels on all ranks - even though this doesn't seem to be required by paraview (your fix worked perfectly fine). I had to experiment a bit to find a solution, and took the liberty of just pushing the end result to this PR. If you're fine with it, then this PR is ready. My changes should not have affected any rebaselines.

francoishamon:
Great! Thank you @klevzoff. I will rebaseline and merge after #1311 to avoid interfering with a potential rebaseline in progress by @CusiniM

CusiniM:
> Great! Thank you @klevzoff. I will rebaseline and merge after #1311 to avoid interfering with a potential rebaseline in progress by @CusiniM

I am still fixing the last few things and I have not rebaselined yet. Feel free to go ahead and merge this first since it seems ready.

PR #1551, opened 2021-08-29
Updated LvArray.

corbett5:
https://github.com/GEOSX/LvArray/pull/242

### DIFF
--- a/src/coreComponents/constitutive/fluid/PVTFunctions/BrineCO2Density.cpp
+++ b/src/coreComponents/constitutive/fluid/PVTFunctions/BrineCO2Density.cpp
@@ -94,13 +94,20 @@ TableFunction const * makeDensityTable( string_array const & inputParams,
   array1d< real64 > densities( tableCoords.nPressures() * tableCoords.nTemperatures() );
   calculateBrineDensity( tableCoords, salinity, densities );
 
-  TableFunction * const densityTable = dynamicCast< TableFunction * >( functionManager.createChild( "TableFunction", "brineDensityTable" ) );
-  densityTable->setTableCoordinates( tableCoords.getCoords() );
-  densityTable->setTableValues( densities );
-  densityTable->reInitializeFunction();
-  densityTable->setInterpolationMethod( TableFunction::InterpolationType::Linear );
-
-  return densityTable;
+  // TODO: fix name/uniqueness
+  if( functionManager.hasGroup< TableFunction >( "brineDensityTable" ) )
+  {
+    return functionManager.getGroupPointer< TableFunction >( "brineDensityTable" );
+  }
+  else
+  {
+    TableFunction * const densityTable = dynamicCast< TableFunction * >( functionManager.createChild( "TableFunction", "brineDensityTable" ) );
+    densityTable->setTableCoordinates( tableCoords.getCoords() );
+    densityTable->setTableValues( densities );
+    densityTable->reInitializeFunction();
+    densityTable->setInterpolationMethod( TableFunction::InterpolationType::Linear );
+    return densityTable;
+  }
 }
 
 } // namespace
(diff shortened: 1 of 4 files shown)


rrsettgast (approved):

corbett5:
This uncovered a bug in how we're using the `TableFunction::KernelWrapper`. The reason `testMultiFluid` is failing is because we are using stale `TableFunction::KernelWrapper`s. First we create the wrapper (which creates a view into `m_values`) then update the table values via `setTableValues` which does the following `m_values = std::move( values )`. This changes the allocation that `m_values` has and when we use the old `KernelWrapper` it has a stale pointer. This wasn't popping up before because due to the leak we weren't actually freeing the old allocation.

I put in some print statements and you can see that `TableFunction::KernelWrapper::compute` is using old pointers.

```
   TableFunction: CO2DensityTable
In TableFunction::setTableValues, old pointer (will be free'd) = 0
In TableFunction::setTableValues, new pointer = 0x4dae1b0
   TableFunction: CO2ViscosityTable
In TableFunction::setTableValues, old pointer (will be free'd) = 0
In TableFunction::setTableValues, new pointer = 0x4db6db0
   TableFunction: brineDensityTable
In TableFunction::setTableValues, old pointer (will be free'd) = 0
In TableFunction::setTableValues, new pointer = 0x4dbc430
   TableFunction: CO2SolubilityTable
In TableFunction::setTableValues, old pointer (will be free'd) = 0
In TableFunction::setTableValues, new pointer = 0x4dc39a0
   TableFunction: CO2DensityTable
In TableFunction::setTableValues, old pointer (will be free'd) = 0x4dae1b0
In TableFunction::setTableValues, new pointer = 0x4dd40d0
   TableFunction: CO2ViscosityTable
In TableFunction::setTableValues, old pointer (will be free'd) = 0x4db6db0
In TableFunction::setTableValues, new pointer = 0x4dd7590
   TableFunction: brineDensityTable
In TableFunction::setTableValues, old pointer (will be free'd) = 0x4dbc430
In TableFunction::setTableValues, new pointer = 0x4dd0a50
   TableFunction: CO2SolubilityTable
In TableFunction::setTableValues, old pointer (will be free'd) = 0x4dc39a0
In TableFunction::setTableValues, new pointer = 0x4dbc430
In TableFunction::KernelWrapper::compute pointer = 0x4dc39a0
In TableFunction::KernelWrapper::compute pointer = 0x4dbc430
In TableFunction::KernelWrapper::compute pointer = 0x4dae1b0
In TableFunction::KernelWrapper::compute pointer = 0x4db6db0
In TableFunction::KernelWrapper::compute pointer = 0x4dbc430
In TableFunction::KernelWrapper::compute pointer = 0x4dae1b0
```

I would prefer not to fix this myself, so if someone else could step in that would be great!

klevzoff:
@corbett5 the last commit seems to fix the tests for me (both unit and integrated). It avoids repeatedly setting up tables, which was caused by copies of the fluid model being attached to subregions. This is a workaround, not a permanent fix, but it should be good enough for you to merge your changes assuming the CI passes (and you drop the debug print). I'm going to do some cleanup around `TableFunction` and PVT function wrappers in a separate PR.

rrsettgast:
@klevzoff In general don't know if it is the best idea for a constitutive model to carry around `TableFunction::KernelWrapper` objects as members. I had taken a glance at the problem and thought that this was the core issue that the `TableFunction` was being reset, but the `TableFunction::KernelWrapper` members inside the constitutive classes were not updated.

corbett5:
@klevzoff Excellent, thanks for the quick fix!

klevzoff:
@rrsettgast Yes, it's not a great idea for sure. We had to do this since we couldn't capture 4 `TableFunction::KernelWrapper`s in RAJA kernels because of their total size exceeding the 4kb CUDA limit. The solution was to pass them via global device memory, for which we chose to store them in `array1d`s, and for that the wrappers had to be pre-created (otherwise each array would be getting copied at every kernel launch). I think the full solution is going to involve reducing the size of `TableFunction::KernelWrapper` (there are some low hanging fruits there) and getting rid of `array1d` storage. Regardless of that, the fix I applied here is also useful, because we were attempting to reset the `TableFunction` from each copy of the constitutive model on every subregion, and that's clearly not what we should be doing.

PR #1554, opened 2021-09-01
Multiphase fluid model cleanup

klevzoff:
* Reduce size of `TableFunction::KernelWrapper` and re-enable capturing them in kernels, avoiding the storing them in array of size 1 (only in `MultiPhaseMultiComponentFluid`, black-oil models need some thinking as they may have a dynamic number of these)
* Group multiphase fluid variables into structs (of arrays, views or slices) to simplify passing them as function/constructor parameters
* Make all `...Update` wrappers be nested classes of the parent fluid model (all uniformly named `KernelWrapper` for lack of better name)
* Introduce `BlackOilFluidBase::KernelWrapper` to contain common variables of `DeadOilFluid::KernelWrapper` and `BlackOilFluid::KernelWrapper`
* Remove `deliverClone()` override from some models, instead registering all members via wrappers to enable auto-cloning
* Consistify all error messages across various models
* Fix naming uniqueness of the auto-generated `TableFunction`s by prepending the parent fluid model's name
* Replace all `localIndex` variables pertaining to components and phases by `integer` ones
* Enforce `;` after `GEOSX_UNUSED_VAR()` macro call (this has been bugging me for a while)
* Probably some miscallaneous stuff I forgot already

No numeric diffs, but rebaseline will be needed for all multiphase problems due to change in some wrappers/table names.
Rebaseline PR: https://github.com/GEOSX/integratedTests/pull/147

I apologize for the number of changes. These could've been 3-4 separate PRs, but I couldn't stop.


francoishamon (approved):
Really nice refactoring! I don't have specific comments, except that I like the shorter argument lists in the fluid constitutive models and the refactored table function.

In future PRs, should we start using `integer` instead of `localIndex` in the phase/component loops of the compositional flow/wells solvers as well?

Issue #1550, opened 2021-08-29
Develop doesn't compile on Quartz with GCC 8.1.0


corbett5:
```
/usr/WS2/corbett5/geosx/GEOSX2/src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVMKernels.cpp:567:12: error: ISO C++ forbids variable length array [-Wvla]
     real64 transmissibility[MAX_NUM_OF_CONNECTIONS][2], dTrans_dPres[MAX_NUM_OF_CONNECTIONS][2];
```

klevzoff:
Seems to be a compiler bug fixed in gcc 8.4: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=86429
We can probably work around by moving `constexpr` variable declarations inside the RAJA kernel here:
https://github.com/GEOSX/GEOSX/blob/a23824a6fc366c035a7a74f2adcd903882b8f12f/src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVMKernels.cpp#L550-L552

PR #1556, opened 2021-09-02
Bugfix for #1550

klevzoff:
Resolves https://github.com/GEOSX/GEOSX/issues/1550 (hopefully, as I don't have gcc-8.1.0 and can't check)

### DIFF
--- a/src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVMKernels.cpp
+++ b/src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVMKernels.cpp
@@ -567,12 +567,10 @@ FluxKernel::
   typename STENCILWRAPPER_TYPE::IndexContainerViewConstType const & sei = stencilWrapper.getElementIndices();
 
   constexpr localIndex MAX_NUM_ELEMS = STENCILWRAPPER_TYPE::NUM_POINT_IN_FLUX;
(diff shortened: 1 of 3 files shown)


CusiniM:
It still gives the same error. I guess a possible fix would be to use a `stackArray` instead but it requires some changes.

- Do we really need to support gcc 8.1.0? If supporting it is a must then we should probably add it to the CI.

rrsettgast (approved):

corbett5:
GCC 8.1.0 is the latest GCC available on Quartz so I think we should definitively try and support it.

CusiniM:
> GCC 8.1.0 is the latest GCC available on Quartz so I think we should definitively try and support it.

Then, I am not sure what we can do to fix this. I thought quartz had some newer versions of gcc...

klevzoff:
@CusiniM can you try building this branch on Quartz once again? I rebased it though, so don't pull but do something like:
```
git fetch origin
git reset --hard origin/bugfix/klevzoff/issue-1550
```

rrsettgast:
@klevzoff This compiles using gcc8.1 on quartz.

klevzoff:
@rrsettgast we'll consider it a win then :)

corbett5:
What a silly fix. How did you figure it out?

klevzoff:
> What a silly fix. How did you figure it out?

Yep. Since the compiler was clearly choosing to treat the local variable as non-constexpr (hence the VLA warning), I had to make sure it doesn't have a choice by removing the local variable.

PR #1557, opened 2021-09-02
Fix for reference porosity import

klevzoff:
Resolves https://github.com/GEOSX/GEOSX/issues/1553

Instead of making `referencePorosity` a pseudo-2D field (as I suggested in the issue), I just modified the logic in PAMELA field import to treat 1D fields as normal, non-material data.

This fixes the import in fluid flow benchmarks (Egg and SPE10), although both of these problems eventually fail due to a floating point error in the linear solver. But that is worth of another investigation.


CusiniM (approved):
This makes sense in my opinion.

PR #1558, opened 2021-09-03
Fix #1524

### LINKED ISSUE #1524: [Bug] R1Tensor input format not being enforced
**Describe the bug**
We are not enforcing the use of curly braces for R1Tensors in input files.  This can lead to odd looking inputs, which GEOSX will parse without any issues:

image

@klevzoff  pointed out that this is due to the following:

https://github.com/GEOSX/GEOSX/blob/6537dd0fd7d0b5d932c031b6cd706d05ab0b81a5/src/coreComponents/dataRepository/xmlWrapper.cpp#L32-L35

klevzoff:
This PR:
  * Resolves https://github.com/GEOSX/GEOSX/issues/1524 
  * Adds unit tests for tensor input parsing
  * Fixes problem with doxygen CTest under ninja (wasn't working due to ninja merging `stderr` with `stdout`)
  * Adds missing GEOSX_THROW macro variants


cssherman (approved):
Looks great!

PR #1568, opened 2021-09-11
Use {fmt} for message formatting

klevzoff:
Resolves https://github.com/GEOSX/GEOSX/issues/852

This PR includes only some examples of using new formatting macros. If the general direction is approved, I (we) can go ahead and use them in more places.

It's explicitly a non-goal to be able to use this in kernels. If portable kernel debug printing is desired, we can keep using `printf` or some wrapper around it, but it shouldn't prevent is from using nicer facilities for non-debugging output (progress, diagnostics, etc.).

TPL update PR: https://github.com/GEOSX/thirdPartyLibs/pull/161
LvArray PR for updating host-configs: https://github.com/GEOSX/LvArray/pull/245


scripts/uberenv/packages/geosx/package.py
```diff
@@ -119,6 +119,8 @@ class Geosx(CMakePackage, CudaPackage):
 
     depends_on('pugixml@1.8: +shared')
 
+    depends_on('fmt@8.0: +cxxstd=14 +pic')
```
corbett5, line 122: Nice, thank you!

corbett5 (approved):

src/coreComponents/constitutive/fluid/CompositionalMultiphaseFluid.cpp
```diff
@@ -89,11 +72,21 @@ template< typename ARRAY >
 void checkInputSize( ARRAY const & array, integer const expected, string const & attr, string const & name )
 {
   GEOSX_THROW_IF_NE_MSG( array.size(), expected,
-                         name << ": invalid number of entries in " << attr << " attribute",
+                         GEOSX_FMT( "{}: invalid number of entries in attribute '{}'", name, attr ),
```
TotoGaz, line 75: Here for example, I'm not convinced that `fmt` brings a lot _w.r.t._ `<<`.

I've been using this kind of approach on python for a while and I also think that's it's often more convenient to write strings with the `fmt` approach.
But when it comes to reading, I'm less convinced because your eyes have to swing here and there...

It's not as great as python's `fstring`!

src/coreComponents/dataRepository/WrapperBase.cpp
```diff
@@ -69,23 +69,15 @@ string WrapperBase::getPath() const
...
-  if( getInputFlag() == InputFlags::OPTIONAL ||
-      getInputFlag() == InputFlags::REQUIRED )
+  if( getInputFlag() == InputFlags::OPTIONAL || getInputFlag() == InputFlags::REQUIRED )
   {
-    sprintf( temp, "  | %20s | %9s | %s \n",
-             getName().c_str(),
-             InputFlagToString( getInputFlag()).c_str(),
-             getDescription().c_str() );
-    rval.append( temp );
+    rval.append( GEOSX_FMT( "  | {:20} | {:9} | {} \n", getName(), InputFlagToString( getInputFlag() ), getDescription() ) );
```
TotoGaz, line 80: Well, it looks (a little) better than `printf`. But what doesn't? 😉

src/coreComponents/mainInterface/GeosxState.cpp
```diff
@@ -74,12 +74,10 @@ class Timer
...
 {
+  // If we want to print HH::MM::SS (maybe in addition to seconds-only):
+  // return GEOSX_FMT( "{:%T}", duration );
   double const seconds = std::chrono::duration_cast< std::chrono::milliseconds >( duration ).count() / 1000.0;
-
-  char buffer[ 32 ];
-  std::snprintf( buffer, 32, "%20.3fs", seconds );
-
-  return buffer;
+  return GEOSX_FMT( "{:>20.3f}s", seconds );
```
TotoGaz, line 80: This looks better indeed.
But moreover removes the issue of the fixed sized buffer that may eventually burst someday.

src/coreComponents/physicsSolvers/SolverBase.cpp
```diff
@@ -113,15 +113,16 @@ bool SolverBase::checkModelNames( array1d< string > & modelNames,
   }
 
   GEOSX_ERROR_IF_NE_MSG( modelNames.size(), m_targetRegionNames.size(),
-                         getName() << ": invalid number of values in " << attribute << " attribute "
-                                                                                       "(expected one model name per target region, or one value for all regions)" );
+                         GEOSX_FMT( "{}: invalid number of values in attribute '{}' "
+                                    "(expected one model name per target region, or one value for all regions)",
+                                    getName(), attribute ) );
```
TotoGaz, line 118: You see here, I think that the `GEOSX_FMT` is more difficult to read than the `std` version.
On the left, there was an issue with the indentation, for sure. But we could have fixed it.

TotoGaz (commented):
> I (we) can go ahead and use them in more places.

While useful in some places where we want to control with care the output format, `fmt` may not bring something in all cases.

Also, how does this lib compares with the incoming `std::format`? Could it be replaced in the end?

rrsettgast (approved):
Looks good to me.

src/coreComponents/constitutive/fluid/CompositionalMultiphaseFluid.cpp
klevzoff: I think it still brings value, in the form of decoupling message structure from arguments, allowing you to change the message itself more easily. In this PR I specifically didn't change any of the messages in order to show 1-to-1 old/new comparison, but one can imagine wanting to rewrite messages at some point, and having format strings just makes it easier.

For example, one might want to standardize messages that check for number of arguments in an attribute across the code; you can define a `constexpr` format string in one header and reuse it everywhere. Not to mention that it's much more `grep`-able if you're searching for the source of a user-reported error in code. I do agree that some of the bigger advantages of `[std|fmt]::format` - e.g. type safety/awareness, flexible formats specifiers - don't shine in this particular example where we're just dealing with strings.

As for f-strings, I love them in Python, but it's unlikely to ever happen in C++, sadly.

src/coreComponents/mainInterface/GeosxState.cpp
klevzoff: Indeed. My first goal is to remove all mentions of `sprintf` from code. Not only could they overflow and overwrite the stack (e.g. if the user chose a particularly long name for their solver/region/whatever), but even those places that (correctly) used `snprintf` would print truncated messages for no good reason. For performance-sensitive code that can't afford allocation of a string (like file output in `HypreMatrix`) I left the ability to reuse a local buffer via `GEOSX_FMT_TO` with a mandatory size argument.

klevzoff:
> Also, how does this lib compares with the incoming `std::format`? Could it be replaced in the end?

`std::format` is the standardized subset of `{fmt}`'s functionality/syntax, and I'm sticking to just that subset on purpose. At some later point if we want to raise minimum standard support, it will be a matter of replacing `fmt::` with `std::` in macro definitions and removing the TPL.

src/coreComponents/physicsSolvers/SolverBase.cpp
klevzoff: I sort of disagree, but without a strong/technical argument. It all comes down to personal preference, I guess. I prefer to scan the entire message first to get an idea of what's being reported, then look at the specific arguments. Or maybe this preference is reinforced by the prevalence of `printf` descendants in programming languages; I don't know at this point.

rrsettgast:
> > Also, how does this lib compares with the incoming `std::format`? Could it be replaced in the end?
> 
> `std::format` is the standardized subset of `{fmt}`'s functionality/syntax, and I'm sticking to just that subset on purpose. At some later point if we want to raise minimum standard support, it will be a matter of replacing `fmt::` with `std::` in macro definitions and removing the TPL.
> 
> Edit: I was even thinking of defining
> 
> ```c++
> #if __cplusplus >= 202002L
>   #define FMT_NAMESPACE std
> #else
>   #define FMT_NAMESPACE fmt
> #endif
> ```
> 
> but as I don't have a c++20 conformant compiler on my hands, I can't verify correctness so there's no point.

what about something like this?
```
#if !defined(USE_FMT)
#if __cplusplus >= 202002L
  #define fmt std
#else
  #error "USE_FMT is undefined, must enable c++20"
#endif
#endif

```

src/coreComponents/physicsSolvers/SolverBase.cpp
TotoGaz: > I prefer to scan the entire message first to get an idea of what's being reported, then look at the specific arguments.

The idea is that you rarely cannot really understand a sentence when words are missing. Language is not a function! 😉 
https://en.wikipedia.org/wiki/The_Smurfs

src/coreComponents/constitutive/fluid/CompositionalMultiphaseFluid.cpp
TotoGaz: > ...one might want to standardize messages that check for ...

This is a relevant argument.
I would nevertheless say that "_same error message_" -> "_same error_" -> "_single function_".

src/coreComponents/physicsSolvers/SolverBase.cpp
```diff
@@ -203,20 +201,21 @@ void SolverBase::setNextDtBasedOnNewtonIter( real64 const & currentDt,
...
-    GEOSX_LOG_LEVEL_RANK_0( 1, getName() << ": Newton solver converged in less than " << iterIncLimit << " iterations, time-step required will be doubled." );
+    GEOSX_LOG_LEVEL_RANK_0( 1, GEOSX_FMT( msg, getName(), iterIncLimit, "doubled" ) );
   }
-  else if( newtonIter >  iterCutLimit )
+  else if( newtonIter > iterCutLimit )
   {
     // Tough convergence let us make the time-step smaller!
     nextDt = currentDt/2;
-    GEOSX_LOG_LEVEL_RANK_0( 1, getName() << ": Newton solver converged in more than " << iterCutLimit << " iterations, time-step required will be halved." );
+    GEOSX_LOG_LEVEL_RANK_0( 1, GEOSX_FMT( msg, getName(), iterCutLimit, "halved" ) );
```
TotoGaz, line 218: I did not spot it at first. But reading your answer about factorisation, I noticed it.
Here first line says "_less_" and second says "_more_".
While you took care of "_doubled_" and "_halved_", your refactoring was a bit harsh 😉 

About the searchability of error messages, who will really search for `Newton solver converged in {} than` instead of `Newton solver converged in less than`?

klevzoff: Thanks for catching this! Perhaps factoring out the common part of these messages is not worth it, since they are actually quite different (and may need to be worded independently).

TotoGaz: No pb! In this precise case, I think simplicity outperforms factorisation.

scripts/uberenv/packages/geosx/package.py
klevzoff: @corbett5 This will probably actually break the spack build, because the target branch in your fork does not have `8.x` version of `fmt`. How much effort would it be to rebase your branch onto latest upstream spack/develop and can I help?

corbett5: A non-trivial amount. I tried it before and encountered a bug with Silo, and maybe some other things I don't remember. My path (hack) forward has been to just update the individual package files. This isn't sustainable, but no-one besides me really uses Spack anyways :(

https://github.com/spack/spack/issues/22464

PR #1572, opened 2021-09-14
Added calls to reserve in transmissibilities precomputation

francoishamon:
This small PR adds three calls to `reserve` for the arrays stored in the stencil classes. 

This is to make sure that we are able run large one-rank models without stalling in the precomputation of the transmissibilities. The PR solves the problem and does not seem to break anything else.


CusiniM (approved):
Looks good. Can you also make sure that the other sencils allocate what they need? I am not sure the call is even made.

francoishamon:
> Looks good. Can you also make sure that the other sencils allocate what they need? I am not sure the call is even made.

Yes you're right! I added some calls to `reserve` for the other stencils. Thanks

Issue #1533, opened 2021-08-12
[Other] Throw on invalid XML


klevzoff:
GEOSX should probably throw/terminate if XML input was not parsed correctly. Currently we just log the error and attempt to proceed. I'm not sure what the use case for this behavior is. Usually it just leads to much less user-friendly errors down the line.

In addition, `ProblemManager` member variables `xmlDocument`, `xmlResult` and `xmlProblemNode` don't seem to be used outside of a single function, `ProblemManager::parseInputFile()`. As such, they may not need to be members, and keeping the whole document around is pointless. We should verify that no part of the code is keeping around xml node handles, and dispose of the document once it's fully processed.

joshua-white:
I second both points.  Adding @cssherman as our all-things-xml guru.

rrsettgast:
@klevzoff Do you have an example xml that is not parsed properly, but continues to run?

klevzoff:
@rrsettgast Adding malformed XML to any input file, for example adding an extra `<` before `<Problem>` gets you this:
```
Max threads: 12
XML parsed with errors!
Error description: Could not determine tag type
Error offset: 25

***** LOCATION: /home/klevtsov/work/GEOSX-develop/src/coreComponents/dataRepository/Group.hpp:325
***** Controlling expression (should be false): child == nullptr
***** Rank 0: Group /domain/MeshBodies doesn't have a child 0
```
Of course, it doesn't continue to run, but not because we rejected invalid XML, but because of a later error that reveals no information about the cause. My point is there seems to be no reason to continue beyond these lines:
https://github.com/GEOSX/GEOSX/blob/d3e262301b7dd124135d9797f118b86b20acde57/src/coreComponents/mainInterface/ProblemManager.cpp#L382-L388
i.e. we should throw here instead of just logging. Unless you think there is a case for "recoverable" errors?

klevzoff:
For a more convincing example, I added the same malformed `<` later in the file, before `<Outputs>`, which leads to
```
Max threads: 12
XML parsed with errors!
Error description: Could not determine tag type
Error offset: 3066
Adding Solver of type SinglePhaseFVM, named SinglePhaseFlow
Adding Mesh: InternalMesh, mesh1
Adding Geometric Object: Box, source
Adding Geometric Object: Box, sink
Adding Event: PeriodicEvent, outputs
Adding Event: PeriodicEvent, solverApplications
Adding Event: PeriodicEvent, restarts
[...]

PR #1580, opened 2021-09-17
Updating the input xml parsing, adding two xml unit tests

cssherman:


cssherman:
While adding a unit test to check the xml parsing errors, I added a method to parse an input string that was inspired by what we do in some of the flow tests.  We should be able to remove some redundant code in the remaining tests with this approach.  Also, I expect that this would be useful for pygeosx.

cssherman:
Addresses #1533

klevzoff (approved):
Nice!

PR #1581, opened 2021-09-18
Add version output

klevzoff:
This PR adds printing of GEOSX version at the beginning of each run (including git branch and commit hash, if applicable, i.e. when building from git clone as opposed to a tarball). This is helpful for resolving user-reported issues.

An alternative would be to add a command-line option for printing version. While it's a bit nicer, it potentially adds an additional communication round-trip (asking the user to run with `--version`). I can implement that if desired. Any other suggestions are welcome too.

### DIFF
--- a/src/docs/doxygen/GeosxConfig.hpp
+++ b/src/docs/doxygen/GeosxConfig.hpp
@@ -11,13 +11,15 @@
 /// GEOSX major version number
 #define GEOSX_VERSION_MAJOR 0
 /// GEOSX minor version number
-#define GEOSX_VERSION_MINOR 0
+#define GEOSX_VERSION_MINOR 2
 /// GEOSX patch version number
-#define GEOSX_VERSION_PATCH 31
+#define GEOSX_VERSION_PATCH 0
 /// GEOSX full version number string
-#define GEOSX_VERSION_FULL  "0.0.31"
+#define GEOSX_VERSION_FULL  "0.2.0"
+/// GEOSX development version string
+/* #undef GEOSX_VERSION_DEV */
 
-/// Enables floating point execptions
+/// Enables floating point exceptions
 #define GEOSX_USE_FPE
 
 /// Enables bounds check in LvArray classes (CMake option ARRAY_BOUNDS_CHECK)
--- a/src/VERSION
+++ b/src/VERSION
@@ -1,1 +1,1 @@
-VERSION_ID = v0.0.31
+VERSION_ID = v0.2.0
--- a/src/cmake/GeosxConfig.cmake
+++ b/src/cmake/GeosxConfig.cmake
@@ -8,8 +8,18 @@ string( REPLACE "." ";" GEOSX_VERSION_LIST ${GEOSX_VERSION_FULL} )
 list( GET GEOSX_VERSION_LIST  0 GEOSX_VERSION_MAJOR )
 list( GET GEOSX_VERSION_LIST  1 GEOSX_VERSION_MINOR )
 list( GET GEOSX_VERSION_LIST  2 GEOSX_VERSION_PATCH )
-message( STATUS "Configuring GEOSX version ${GEOSX_VERSION_FULL}" )
 
+if( GIT_FOUND )
+    blt_is_git_repo( OUTPUT_STATE is_git_repo )
+    if( is_git_repo )
+        # the other option is to use blt_git_tag() but we want branch info too
+        blt_git_branch( BRANCH_NAME git_branch RETURN_CODE git_branch_rc )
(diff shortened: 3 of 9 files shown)


src/docs/doxygen/GeosxConfig.hpp
```diff
@@ -11,13 +11,15 @@
...
-#define GEOSX_VERSION_MINOR 0
+#define GEOSX_VERSION_MINOR 2
 /// GEOSX patch version number
-#define GEOSX_VERSION_PATCH 31
+#define GEOSX_VERSION_PATCH 0
 /// GEOSX full version number string
-#define GEOSX_VERSION_FULL  "0.0.31"
+#define GEOSX_VERSION_FULL  "0.2.0"
+/// GEOSX full version number string
+#define GEOSX_VERSION_DEV  "feature/klevzoff/version-bccb96b9f"
```
klevzoff, line 20: Oops, that is definitely a problem. Will add a workaround.

rrsettgast: After you merge, just run on `develop`, then checkout a new branch to merge into develop....but it isn't really a big deal since this isn't really used.

rrsettgast (approved):

TotoGaz:
Does this imply that every time there is a new commit, all the files that include `GeosxConfig.hpp` require a rebuilt? (I think `GeosxConfig.hpp` is quite high in the hierarchy).

Do I have to rerun `cmake` to get this working? Or is it part of the `make` process which reruns the underlying `git` commands "each time".
If rerunning `cmake` is required and `cmake` is not run again, is it possible that the hash is not the good one?

klevzoff:
> Does this imply that every time there is a new commit, all the files that include `GeosxConfig.hpp` require a rebuilt? (I think `GeosxConfig.hpp` is quite high in the hierarchy).

Great question! Yes, and it is a problem. I should have placed the "development" version macro in a separate header that is included only where it is used (the definition of `getVersion()`).

> Do I have to rerun `cmake` to get this working? Or is it part of the `make` process which reruns the underlying `git` commands "each time".
> If rerunning `cmake` is required and `cmake` is not run again, is it possible that the hash is not the good one?

Another great question. It's part of `cmake` run in the current implementation. The reason for that is `blt` provides nice macro wrappers around raw git calls, and these are not callable at build time. It is thus possible that if a user pulls a new version and the build does not invoke `cmake` again, they will get stale value. I will look into fixing this as well.

I apologize for rushing this PR :( Both issues raised are very important to the intended functioning of this feature. Will follow up with a fix soon.

TotoGaz:
> > Does this imply that every time there is a new commit, all the files that include `GeosxConfig.hpp` require a rebuilt? (I think `GeosxConfig.hpp` is quite high in the hierarchy).
> 
> Great question! Yes, and it is a problem. I should have placed the "development" version macro in a separate header that is included only where it is used (the definition of `getVersion()`).

Yes, source code can surely be split differently such that you only have one file to build and a few links to edit.
This should not be an unsurmountable problem.

> > Do I have to rerun `cmake` to get this working? Or is it part of the `make` process which reruns the underlying `git` commands "each time".
> > If rerunning `cmake` is required and `cmake` is not run again, is it possible that the hash is not the good one?
> 
> Another great question. It's part of `cmake` run in the current implementation. The reason for that is `blt` provides nice macro wrappers around raw `git` calls, and these are not callable at build time. It is thus possible that if a user pulls a new version and the build does not invoke `cmake` again, they will get stale value. I will look into fixing this as well.

Thanks for the precisions.
That's what I was expecting, but it's always possible to link another build target to the `geosx` target and let `make` do its work. Maybe this is the solution that may come up next, who knows?

Also (and you did it properly), consider the case where sources are not a `git` repository.
Typically my case when remote developing with `cLion`.
Or probably when you click on the "_Download_" button from github.
And maybe code releases too.
I'm sure this is a feature which is quite used in companies with complicated proxies.

> I apologize for rushing this PR :( Both issues raised are very important to the intended functioning of this feature. Will follow up with a fix soon.

Do not worry, it's part of life 😉 
We are not supporting _New York Stock Exchange_ or the _gmail_ homepage, we'll go through it!

If you want a personal production IT advice: _never put yourself at stake_.
After this PR, the GEOSX version can be wrong by design (and _will therefore be_ wrong for some of our users).
According to my personal priority scale, providing wrong information to the user is always critical.

The current situation is that a solution needs to come up relatively fast, and it's uncomfortable.
Except if you have a smart patch that can pop fast, one good solution could be to simply revert the PR and calmly come in the next days with a nice solution carefully validated.

Now, as I mentioned: we're not _Wall Street_ we can probably live with it for a while 😉

PR #1583, opened 2021-09-21
Improve version header generation

klevzoff:
This PR addresses two issues noticed by @TotoGaz in https://github.com/GEOSX/GEOSX/pull/1581 :
1. Each time the development version changed, an almost-full rebuild would ensue as a result of changes to `GeosxConfig.hpp`. The solution is to put all version information macros into a separate header, `GeosxVersion.hpp`, that is only included in a single translation unit that is fast to rebuild. The header is placed as low as possible in the dependency tree - in `coreComponents/mainInterface` (however not in `main`, so that it remains accessible also through python in the future).
2. Version would only be updated upon re-configure, meaning that if one pulled new commits and did not rerun `cmake`, they could end up with stale version data baked into the binary. The solution is to generate version header at build time (instead of or in addition to config time). This is now done via new CMake script, `GeosxVersion.cmake` that runs both at config time (and generates a new build-time target) and then later at build time. There are some small tricks to avoid triggering partial rebuild if nothing has changed (using a temporary header and CMake's `copy_if_different` command), but all in all the end solution is fairly neat. 

I've tested it on both "dirty" and "clean" in-repo builds, as well as a non-repo build (deleting `.git` from a repository, which is more or less equivalent to a release tarball) and did not observe any problems.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -65,6 +65,14 @@ endif()
 ################################
 install( DIRECTORY ${PROJECT_BINARY_DIR}/include DESTINATION . )
 
+
+################################
+# Generate version information
+################################
+include( cmake/GeosxVersion.cmake )
+message( STATUS "Configuring GEOSX version ${GEOSX_VERSION_FULL}" )
+
+
 ################################
 # Create header of configuration options
 ################################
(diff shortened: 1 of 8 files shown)


TotoGaz (approved):
I could test on my non repo version and it was OK.
I think your solution will work well now 😉

Issue #1528, opened 2021-08-03
Automatic allocation of compound constitutive relations


klevzoff:
@CusiniM, I could see the need to specify all of the sub-materials independently becoming a usability issue.  Users may perhaps forget to add rock when they define porousRock.  Just something to think about.

_Originally posted by @joshua-white in https://github.com/GEOSX/GEOSX/pull/1475#discussion_r682035369_

This was in reference to the following XML input:
```
materialList="{ rock, water, porousRock, rockPerm, rockPorosity}"
```

klevzoff:
Simple proposal (subject to debate):
1. In `ConstitutiveBase`, add `virtual std::vector< string > getSubRelationNames() { return {}; }` (de-uglify the name)
2. Override in `CoupledSolid` to `return { m_solidModelName, m_porosityModelName, m_permeabilityModelName };`
3. Modify `ConstitutiveManager::hangConstitutiveRelation()` to additionally loop over this list from the compound relation and automatically register all sub-relations with the same quadrature.
4. Modify all XML files to only include the top-level model.

We can also consider doing this for multiphase fluids, where fluid, relative permeability and capillary pressure can be combined into a meta-model.

CusiniM:
I think this is the way to go and we should also do it for the Fluid models as you suggested. Although capillary pressure and relative permeability are these weird properties which I don't know whether should belong to the fluid or to the porous material since they kind of depend on both. In any case, this will also allow to avoid specifying all the material names in the solvers which I think will make our xmls easier to handle.

joshua-white:
I like this suggestion a lot.

CusiniM:
I would allow flexibility in terms of number of constitutive points though. For example, I think one may want the permeability to be piecewise constant and the porosity to vary at each quadrature point.

rrsettgast:
@klevzoff This is a good suggestion. 

@CusiniM Do we want a composite model mixing different quadratures for sub-models?

CusiniM:
I think that for certain cases we do. For example, if we choose to run a problem with a permeability model that ignores all permeability dependencies on the state variables it feels a bit of a waste allocating it at each quadrature point. This though, does not necessarily mean that the porosity will also be constant.

CusiniM:
I will start doing this.

PR #1584, opened 2021-09-22
Automatic allocation of compound constitutive relations 

CusiniM:
close issue #1528 

- [x] implement allocation
- [x] modify multiphysics xmls
- [x] modify fluidFlow xmls.
- [x] rebaseline integratedTests (https://github.com/GEOSX/integratedTests/pull/158)


CusiniM:
I am only not sure about what will happen if someone specifies both the compound and the standalone model on the same Region. Will it work? Should we throw an error?

src/coreComponents/constitutive/ConstitutiveManager.cpp
```diff
@@ -67,6 +67,19 @@ ConstitutiveManager::hangConstitutiveRelation( string const & constitutiveRelati
...
                                       numConstitutivePointsPerParentIndex );
 
+
+  std::vector<string> const subRelationNames = material->getSubRelationNames();
+
+  for( string const & subRelationName : subRelationNames )
+  {
+    ConstitutiveBase const & subRelation = getConstitutiveRelation( subRelationName );
+
+    std::unique_ptr< ConstitutiveBase > constitutiveModel = subRelation.deliverClone( subRelationName, parent );
```
klevzoff, line 77: `constitutiveModel` goes out of scope and destructs at the end of the loop body. It needs to be moved a wrapper on the new `constitutiveGroup`, similar to what happens to `material` some lines below.

CusiniM: should work fine now. Should I add a check that if the wrapper already exists we don't register it?

src/coreComponents/constitutive/ConstitutiveManager.cpp
```diff
@@ -60,25 +60,42 @@ ConstitutiveManager::hangConstitutiveRelation( string const & constitutiveRelati
...
+      group = constitutiveGroup->registerGroup< ConstitutiveBase >( subRelationName, std::move( constitutiveModel ) );
+      group.setSizedFromParent( 1 );
+      group.resize( constitutiveGroup->size() );
+  }
 
   ConstitutiveBase &
   rval = constitutiveGroup->registerGroup< ConstitutiveBase >( constitutiveRelationInstanceName, std::move( material ) );
   rval.setSizedFromParent( 1 );
   rval.resize( constitutiveGroup->size() );
+  // Why do we need to return this here? Couldn't this be just void?
```
CusiniM, line 98: I was wondering why we need this function to return that reference. It does not seem to be used.

sytuannguyen:
Please modify also the multiphysics/benchmark XMLs

src/coreComponents/constitutive/ConstitutiveManager.cpp
```diff
@@ -55,33 +55,59 @@ void ConstitutiveManager::expandObjectCatalogs()
...
-  ConstitutiveBase const & constitutiveRelation = getConstitutiveRelation( constitutiveRelationInstanceName );
-
-  std::unique_ptr< ConstitutiveBase > material = constitutiveRelation.deliverClone( constitutiveRelationInstanceName, parent );
-
-  material->allocateConstitutiveData( *parent,
-                                      numConstitutivePointsPerParentIndex );
-
   dataRepository::Group * constitutiveGroup = parent->getGroupPointer( groupKeyStruct::constitutiveModelsString() );
-  if( constitutiveGroup == nullptr )
+  if( constitutiveGroup == nullptr )   // Why do we need to do this check? shouldn't it always be there?
```
klevzoff, line 64: The way we currently use this by passing a descendant of `ElementSubRegionBase` - yes, it always has a constitutive subgroup in it. Bit the function signature takes an arbitrary `Group`, which may or may not have it. If we dropped the check, someone could unwittingly try to put constitutive models on something that is not a subregion and get segfaults. We could change the function signature to take `ElementSubRegionBase`, but that would require taking a dependency on `mesh` which I believe is not desirable and possibly circular.

src/coreComponents/constitutive/ConstitutiveManager.cpp
```diff
@@ -55,33 +55,59 @@ void ConstitutiveManager::expandObjectCatalogs()
...
+    ConstitutiveBase &
+    rval = constitutiveGroup->registerGroup< ConstitutiveBase >( constitutiveRelationInstanceName, std::move( material ) );
+    rval.setSizedFromParent( 1 );
+    rval.resize( constitutiveGroup->size() );
+
+    // 2. Allocate subrelations (for compound models)
+    std::vector< string > const subRelationNames = constitutiveRelation.getSubRelationNames();
+    for( string const & subRelationName : subRelationNames )
+    {
+      std::cout << "nothing to do" << std::endl;
```
klevzoff, line 90:
```suggestion
```

src/coreComponents/constitutive/ConstitutiveManager.cpp
```diff
@@ -55,33 +55,59 @@ void ConstitutiveManager::expandObjectCatalogs()
...
+      // we only register the group if the instance has not been registered yet.
+      if( !constitutiveGroup->hasGroup( subRelationName ) )
+      {
+        ConstitutiveBase &
+        group = constitutiveGroup->registerGroup< ConstitutiveBase >( subRelationName, std::move( constitutiveModel ) );
+        group.setSizedFromParent( 1 );
+        group.resize( constitutiveGroup->size() );
+      }
+    }
+    std::cout << "done" << std::endl;
```
klevzoff, line 107:
```suggestion
```

src/coreComponents/constitutive/ConstitutiveManager.cpp
```diff
@@ -55,33 +55,59 @@ void ConstitutiveManager::expandObjectCatalogs()
...
+      std::cout << "nothing to do" << std::endl;
+      ConstitutiveBase const & subRelation = getConstitutiveRelation( subRelationName );
+
+      std::unique_ptr< ConstitutiveBase > constitutiveModel = subRelation.deliverClone( subRelationName, parent );
+
+      constitutiveModel->allocateConstitutiveData( *parent,
+                                                   numConstitutivePointsPerParentIndex );
+
+      // we only register the group if the instance has not been registered yet.
+      if( !constitutiveGroup->hasGroup( subRelationName ) )
```
klevzoff, line 99: I am not 100% certain, but if the sub-relation has already been registered (presumably by another compound model, or explicitly by the user by listing it in region's material list), should we also avoid calling `deliverClone` and `allocateConstitutiveData` above, i.e. move entire loop body under this branch?

klevzoff (approved):
Nice! The XMLs are looking leaner now.

src/coreComponents/constitutive/ConstitutiveManager.cpp
CusiniM: yes, I think you are right. I don't see why we would not want to do it.

src/coreComponents/constitutive/ConstitutiveManager.cpp
```diff
@@ -55,33 +55,58 @@ void ConstitutiveManager::expandObjectCatalogs()
...
   dataRepository::Group * constitutiveGroup = parent->getGroupPointer( groupKeyStruct::constitutiveModelsString() );
   if( constitutiveGroup == nullptr )
   {
     constitutiveGroup = &parent->registerGroup( groupKeyStruct::constitutiveModelsString() ).setSizedFromParent( 1 );
     constitutiveGroup->resize( parent->size() );
   }
 
+  // 1. Allocate constitutive relation
+  // we only register the constitutive relation if it has not been registered yet.
+  if( !constitutiveGroup->hasGroup( constitutiveRelationInstanceName ) )
```
rrsettgast, line 72: hard error?

src/coreComponents/constitutive/ConstitutiveManager.cpp
```diff
@@ -55,33 +55,58 @@ void ConstitutiveManager::expandObjectCatalogs()
...
+    materialGroup = constitutiveGroup->registerGroup< ConstitutiveBase >( constitutiveRelationInstanceName, std::move( material ) );
+    materialGroup.setSizedFromParent( 1 );
+    materialGroup.resize( constitutiveGroup->size() );
 
+    // 2. Allocate subrelations (for compound models)
+    std::vector< string > const subRelationNames = constitutiveRelation.getSubRelationNames();
+    for( string const & subRelationName : subRelationNames )
+    {
+      // we only want to register the subRelation if it has not been registered yet.
+      if( !constitutiveGroup->hasGroup( subRelationName ) )
```
rrsettgast, line 91: hard error?

rrsettgast (approved):

src/coreComponents/constitutive/ConstitutiveManager.cpp
CusiniM: added.

src/coreComponents/constitutive/ConstitutiveManager.cpp
CusiniM: added.

src/coreComponents/constitutive/ConstitutiveManager.cpp
```diff
@@ -55,33 +55,64 @@ void ConstitutiveManager::expandObjectCatalogs()
 }
 
 
-ConstitutiveBase &
+void
 ConstitutiveManager::hangConstitutiveRelation( string const & constitutiveRelationInstanceName,
                                                dataRepository::Group * const parent,
```
sytuannguyen, line 60:
```suggestion
                                               Group * const parent,
```

src/coreComponents/constitutive/ConstitutiveManager.cpp
```diff
@@ -55,33 +55,64 @@ void ConstitutiveManager::expandObjectCatalogs()
...
                                                localIndex const numConstitutivePointsPerParentIndex ) const
 {
+  dataRepository::Group * constitutiveGroup = parent->getGroupPointer( groupKeyStruct::constitutiveModelsString() );
```
sytuannguyen, line 63:
```suggestion
  Group * constitutiveGroup = parent->getGroupPointer( groupKeyStruct::constitutiveModelsString() );
```

src/coreComponents/constitutive/ConstitutiveManager.hpp
```diff
@@ -52,7 +52,7 @@ class ConstitutiveManager : public dataRepository::Group
   /// This function is used to expand any catalogs in the data structure
   virtual void expandObjectCatalogs() override;
 
-  ConstitutiveBase &
+  void
   hangConstitutiveRelation( string const & constitutiveRelationInstanceName,
                             dataRepository::Group * const parent,
```
sytuannguyen, line 57:
```suggestion
                            Group * const parent,
```

sytuannguyen (approved):

PR #1589, opened 2021-09-22
Fixed negative cell volume with reverseZ on CPG

francoishamon:
In the `PAMELAMeshGenerator`, we have an option called `reverseZ` that can be used to convert depth (typically used in the GRDECL format) to elevation (that we use internally in GEOSX). However, I noticed with @wdmhouston 's case that using `reverseZ` to go from depth to elevation leads to negative cell volumes because of our node ordering assumption:
```
*                  6                   7                       ____________________
 *                   o-----------------o                       |Node   xi0  xi1  xi2|
 *                  /.                /|                       |=====  ===  ===  ===|
 *                 / .               / |                       | 0     -1   -1   -1 |
 *              4 o-----------------o 5|                       | 1      1   -1   -1 |
 *                |  .              |  |                       | 2     -1    1   -1 |
 *                |  .              |  |                       | 3      1    1   -1 |
 *                |  .              |  |                       | 4     -1   -1    1 |
 *                |  .              |  |                       | 5      1   -1    1 |
 *                |2 o..............|..o 3       xi2           | 6     -1    1    1 |
 *                | ,               | /          |             | 7      1    1    1 |
 *                |,                |/           | / xi1       |____________________|
 *                o-----------------o            |/
 *               0                   1           ------ xi0
```
This PR fixes that by switching the nodes {0,1,2,3} and {4,5,6,7} if `reverseZ` is equal to 1.

The other option would be to just add an absolute value [here](https://github.com/GEOSX/GEOSX/blob/a8711cad373dd1328ece6a51713a81732714a01d/src/coreComponents/mesh/utilities/ComputationalGeometry.hpp#L474). But, given that negative volumes have been a very good indicator for meshing issues over the past years, I would prefer not to add the absolute value. Let me know what you think.

### DIFF
--- a/src/coreComponents/mesh/generators/PAMELAMeshGenerator.cpp
+++ b/src/coreComponents/mesh/generators/PAMELAMeshGenerator.cpp
@@ -157,7 +157,8 @@ PAMELA::ELEMENTS::TYPE toPamelaElementType( ElementType const type )
   }
 }
 
-std::vector< int > getPamelaNodeOrder( PAMELA::ELEMENTS::TYPE const type )
+std::vector< int > getPamelaNodeOrder( PAMELA::ELEMENTS::TYPE const type,
+                                       bool const isZReverse )
 {
   switch( type )
   {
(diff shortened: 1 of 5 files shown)


src/coreComponents/mesh/generators/PAMELAMeshGenerator.cpp
```diff
@@ -167,7 +168,19 @@ std::vector< int > getPamelaNodeOrder( PAMELA::ELEMENTS::TYPE const type )
...
+      // (top nodes become bottom nodes) to make sure that the hex volume is positive
+      if( isZReverse )
+      {
+        return { 4, 5, 7, 6, 0, 1, 3, 2 };
+      }
+      else
+      {
+        return { 0, 1, 3, 2, 4, 5, 7, 6 };
+      }
+    }
```
TotoGaz, line 183:
```suggestion
return izReverse ? { 4, 5, 7, 6, 0, 1, 3, 2 } : { 0, 1, 3, 2, 4, 5, 7, 6 };
```
?

TotoGaz (approved):

src/coreComponents/mesh/generators/PAMELAMeshGenerator.cpp
francoishamon: Thanks for the suggestion, I like it, but the compiler says:
```
/g/g90/hamon1/geosx/GEOSX/src/coreComponents/mesh/generators/PAMELAMeshGenerator.cpp:175:26: error:                                                                                                             
      initializer list cannot be used on the right hand side of operator '?'                                                                                                                                    
      return izReverse ? { 4, 5, 7, 6, 0, 1, 3, 2 } : { 0, 1, 3, 2, 4, 5, 7, 6 };                                                                                                                               
                         ^~~~~~~~~~~~~~~~~~~~~~~~~~                                        
```

PR #1591, opened 2021-09-23
[Doc] Fixed some inconsistencies and misprints in the updated tutorial

mkhait:
Noticed a few things while I walked through the updated tutorial, so fixed them along the way.


mkhait:
ready for review

(I don`t have permission to apply the corresponding label here)

francoishamon (approved):
Looks great to me, thanks @mkhait

Issue #1587, opened 2021-09-22
[Bug] Issue of poroelastic wellbore problem in tutorial


jhuang2601:
**Describe the bug**
Rerun the tutorial example for poroelastic wellbore problem with recent develop repo; however, the simulation results cannot match analytical solutions anymore. Similarly, Sy-Tuan observed the same issue with a different wellbore problem.

**To Reproduce**
1. Run the case from `src/coreComponents/physicsSolvers/multiphysics/benchmarks/WellboreProblem_PoroElastic.xml`
2. Use the recent develop repo `cea27cd`

**Expected behavior**
This tutorial example previously worked with develop repo (`496f5b2`) and correlated well with the analytical solutions. Some recent changes cause the unexpected issue of the poromechanics solver and this kind of degradation should be fixed.

**Screenshots**
![image](https://user-images.githubusercontent.com/53012159/134391459-993645b5-1824-4d14-ae6f-e762544d7337.png)

**Platform (please complete the following information):**
- Machine: Pangea2 
- Compiler: gcc 8.3.0
- GEOSX Version: cea27cd

joshua-white:
Grrr, @shabnamjs has added this problem as an integrated test on her branch so we would catch this type of regression.  It looks like we are too late.

@jhuang2601, can you step through the commit history and figure out more precisely where this broke?

jhuang2601:
@joshua-white By tracking commit history and testing the example, the reported regression takes place from commit `a23824a`. The commit `d6347e0` is the last good one.

![image](https://user-images.githubusercontent.com/53012159/134405710-7f943d99-9bbb-49c9-be2c-844dc11806db.png)

joshua-white:
Okay, this PR updates the constitutive models, so that makes sense, but @CusiniM had to pass all of the poroelastic integrated tests to get it in, so those must still work.  Are you sure it is not an issue with a slight re-definition of the XML parameters?  What happens with the other wellbore examples?

@CusiniM, thoughts?

CusiniM:
Hmm Can I see the input files and have a few more details about the test case? The wellbore integratedTests work fine so hopefully it's just something with the xml files.

As a separate note, these test cases should have some coarse version as integratedTests to avoid issues like this one.

joshua-white:
I can confirm the overlap between tutorial and integratedTest coverage is basically zero.  Not good.

CusiniM:
[...]

PR #1592, opened 2021-09-23
Fix porosity initialization for advanced poromechanics wellbore problem

jhuang2601:
If using `BiotPorosity` model, porosity should be initialized; otherwise, inaccurate results will be obtained (#1587). 

In this PR, porosity initialization is added to the poroelastic and proplastic wellbore examples (including xml files for both the benchmark and integrated test).

To avoid similar issue, following three cases should be added to the integrated tests :

`src/coreComponents/physicsSolvers/multiphysics/integratedTests/WellboreProblem_PoroElastic.xml`

`src/coreComponents/physicsSolvers/multiphysics/integratedTests/WellboreProblem_PoroDruckerPrager.xml`

`src/coreComponents/physicsSolvers/solidMechanics/integratedTests/ExtendedDruckerPrager_Wellbore.xml`

### DIFF
--- a/src/coreComponents/physicsSolvers/multiphysics/benchmarks/WellboreProblem_Base.xml
+++ b/src/coreComponents/physicsSolvers/multiphysics/benchmarks/WellboreProblem_Base.xml
@@ -124,6 +124,15 @@
     scale="10e6"
     functionName="timeFunction"   
   />
+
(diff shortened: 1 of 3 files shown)


jhuang2601:
The reported issue has been fixed in PR(#1595).

PR #1596, opened 2021-09-23
Poromechanics kernels refactoring

castelletto1:
This PR moves the computation of the total stress from the poromechanics solver kernels to the constitutive update. Note that the total stress derivative with respect to pressure (`dTotalStress_dPressure`) is now treated as a symmetric second order tensor, which removes the assumption of scalar Biot's coefficient. 

New baselines are not required.

### DIFF
--- a/src/coreComponents/constitutive/solid/PorousSolid.hpp
+++ b/src/coreComponents/constitutive/solid/PorousSolid.hpp
@@ -80,12 +80,12 @@ public:
     dTotalStress_dPressure[4] = 0;
     dTotalStress_dPressure[5] = 0;
 
-    m_porosityUpdate.updatePorosity( k,
-                                     q,
-                                     deltaPressure,
-                                     strainIncrement,
-                                     dPorosity_dPressure,
-                                     dPorosity_dVolStrain );
+    m_porosityUpdate.updateFromPressureAndStrain( k,
+                                                  q,
+                                                  deltaPressure,
+                                                  strainIncrement,
+                                                  dPorosity_dPressure,
+                                                  dPorosity_dVolStrain );
 
 // TODO uncomment once we start using permeability model in flow.
 //    m_permUpdate.updateFromPressureStrain( k,
--- a/src/coreComponents/constitutive/solid/PorousSolid.hpp
+++ b/src/coreComponents/constitutive/solid/PorousSolid.hpp
@@ -61,7 +61,7 @@ public:
                           real64 ( & totalStress )[6],
                           real64 & dPorosity_dPressure,
                           real64 & dPorosity_dVolStrain,
-                          real64 ( &dTotalStress_dPressure )[6],
+                          real64 ( & dTotalStress_dPressure )[6],
                           DiscretizationOps & stiffness ) const
   {
     // Compute effective stress and store in totalStress
@@ -69,24 +69,23 @@ public:
 
     updateBiotCoefficient( k );
 
-   // Compute  total stress
-   real64 const biotCoefficient = m_porosityUpdate.getBiotCoefficient( k );
-   LvArray::tensorOps::symAddIdentity< 3 >( totalStress, - biotCoefficient * ( pressure + deltaPressure ) );
+    // Compute  total stress
+    real64 const biotCoefficient = m_porosityUpdate.getBiotCoefficient( k );
+    LvArray::tensorOps::symAddIdentity< 3 >( totalStress, -biotCoefficient * ( pressure + deltaPressure ) );
 
-   dTotalStress_dPressure[0] = biotCoefficient;
-   dTotalStress_dPressure[1] = biotCoefficient;
-   dTotalStress_dPressure[2] = biotCoefficient;
-   dTotalStress_dPressure[3] = 0;
-   dTotalStress_dPressure[4] = 0;
-   dTotalStress_dPressure[5] = 0;
+    dTotalStress_dPressure[0] = biotCoefficient;
+    dTotalStress_dPressure[1] = biotCoefficient;
+    dTotalStress_dPressure[2] = biotCoefficient;
+    dTotalStress_dPressure[3] = 0;
+    dTotalStress_dPressure[4] = 0;
+    dTotalStress_dPressure[5] = 0;
 
-   m_porosityUpdate.updatePorosity( k,
-                                    q,
-				    pressure,
-                                    deltaPressure,
-                                    strainIncrement,
-                                    dPorosity_dPressure,
-                                    dPorosity_dVolStrain );
+    m_porosityUpdate.updatePorosity( k,
+                                     q,
+                                     deltaPressure,
+                                     strainIncrement,
+                                     dPorosity_dPressure,
+                                     dPorosity_dVolStrain );
 
 // TODO uncomment once we start using permeability model in flow.
 //    m_permUpdate.updateFromPressureStrain( k,
--- a/src/coreComponents/constitutive/solid/PorousSolid.hpp
+++ b/src/coreComponents/constitutive/solid/PorousSolid.hpp
@@ -55,25 +55,38 @@ public:
   GEOSX_HOST_DEVICE
   void smallStrainUpdate( localIndex const k,
                           localIndex const q,
+                          real64 const & pressure,
                           real64 const & deltaPressure,
                           real64 const ( &strainIncrement )[6],
-                          real64 ( & stress )[6],
+                          real64 ( & totalStress )[6],
                           real64 & dPorosity_dPressure,
                           real64 & dPorosity_dVolStrain,
-                          real64 & dTotalStress_dPressure,
+                          real64 ( &dTotalStress_dPressure )[6],
                           DiscretizationOps & stiffness ) const
   {
-    m_solidUpdate.smallStrainUpdate( k, q, strainIncrement, stress, stiffness );
+    // Compute effective stress and store in totalStress
+    m_solidUpdate.smallStrainUpdate( k, q, strainIncrement, totalStress, stiffness );
 
     updateBiotCoefficient( k );
 
-    m_porosityUpdate.updatePorosity( k,
-                                     q,
-                                     deltaPressure,
-                                     strainIncrement,
-                                     dPorosity_dPressure,
-                                     dPorosity_dVolStrain,
-                                     dTotalStress_dPressure );
+   // Compute  total stress
+   real64 const biotCoefficient = m_porosityUpdate.getBiotCoefficient( k );
+   LvArray::tensorOps::symAddIdentity< 3 >( totalStress, - biotCoefficient * ( pressure + deltaPressure ) );
+
+   dTotalStress_dPressure[0] = biotCoefficient;
+   dTotalStress_dPressure[1] = biotCoefficient;
+   dTotalStress_dPressure[2] = biotCoefficient;
+   dTotalStress_dPressure[3] = 0;
+   dTotalStress_dPressure[4] = 0;
+   dTotalStress_dPressure[5] = 0;
+
+   m_porosityUpdate.updatePorosity( k,
+                                    q,
+				    pressure,
+                                    deltaPressure,
+                                    strainIncrement,
+                                    dPorosity_dPressure,
+                                    dPorosity_dVolStrain );
 
 // TODO uncomment once we start using permeability model in flow.
 //    m_permUpdate.updateFromPressureStrain( k,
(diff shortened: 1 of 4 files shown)


src/coreComponents/constitutive/solid/porosity/BiotPorosity.hpp
```diff
@@ -59,15 +59,16 @@ class BiotPorosityUpdates : public PorosityBaseUpdates
...
 
   GEOSX_HOST_DEVICE
   void updatePorosity( localIndex const k,
                        localIndex const q,
                        real64 const & deltaPressure,
                        real64 const ( &strainIncrement )[6],
                        real64 & dPorosity_dPressure,
-                       real64 & dPorosity_dVolStrain,
-                       real64 & dTotalStress_dPressure ) const
+                       real64 & dPorosity_dVolStrain ) const
```
CusiniM, line 71: Since you are doing this can you rename this function `updatePorosityFromPressureAndStrain` and then also rename the one in `PressurePorosity`  `updatePorosityFromPressure`. This way they are more similar to each other.

CusiniM (approved):
This is much cleaner!

src/coreComponents/constitutive/solid/PorousSolid.hpp
```diff
@@ -55,25 +55,37 @@ class PorousSolidUpdates : public CoupledSolidUpdates< SOLID_TYPE, BiotPorosity,
...
   {
-    m_solidUpdate.smallStrainUpdate( k, q, strainIncrement, stress, stiffness );
+    // Compute effective stress and store in totalStress
+    m_solidUpdate.smallStrainUpdate( k, q, strainIncrement, totalStress, stiffness );
 
     updateBiotCoefficient( k );
 
+    // Compute  total stress
+    real64 const biotCoefficient = m_porosityUpdate.getBiotCoefficient( k );
+    LvArray::tensorOps::symAddIdentity< 3 >( totalStress, -biotCoefficient * ( pressure + deltaPressure ) );
```
joshua-white, line 74: Is there a reason not to just pass newPressure as the argument to this function, instead of both pressure and deltaPressure?

joshua-white (approved):

src/coreComponents/constitutive/solid/PorousSolid.hpp
castelletto1: Do you mean the having the following?

```c++
// Compute  total stress
real64 const biotCoefficient = m_porosityUpdate.getBiotCoefficient( k );
real64 const newPressure = pressure + deltaPressure;
LvArray::tensorOps::symAddIdentity< 3 >( totalStress, -biotCoefficient * newPressure );
```

CusiniM: I think he means to only give `newPressure` as input to the `smallStrainUpdate` but, we only have the `strainIncrement` so the porosity is update as a function of the strain and the pressure increments which is why we need `deltaPressure`.

joshua-white: Sorry, I meant in the argument list at line 58.  We never actually use the two separately here.

joshua-white: Ohh, I see, deltaPressure is actually used below.  Nevermind.

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoromechanicsKernel.hpp
```diff
@@ -282,7 +282,7 @@ class Multiphase :
     real64 totalStress[6];
     real64 dPorosity_dPressure;
     real64 dPorosity_dVolStrainIncrement;
-    real64 dTotalStress_dPressure;
+    real64 dTotalStress_dPressure[6] = {0};
```
CusiniM, line 285:
```suggestion
    real64 dTotalStress_dPressure[6] = {0.0};
```

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoromechanicsKernel.hpp
```diff
@@ -227,14 +227,15 @@ class SinglePhase :
     real64 totalStress[6];
     real64 dPorosity_dPressure;
     real64 dPorosity_dVolStrainIncrement;
-    real64 dTotalStress_dPressure;
+    real64 dTotalStress_dPressure[6] = {0};
```
CusiniM, line 230:
```suggestion
    real64 dTotalStress_dPressure[6] = {0.0};
```

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoromechanicsKernel.hpp
castelletto1: I assume you'd like to see also 
```c++
real64 strainIncrement[6] = {0.0};
```

src/coreComponents/constitutive/solid/PorousSolid.hpp
```diff
@@ -55,25 +55,37 @@ class PorousSolidUpdates : public CoupledSolidUpdates< SOLID_TYPE, BiotPorosity,
...
+
+    dTotalStress_dPressure[0] = biotCoefficient;
+    dTotalStress_dPressure[1] = biotCoefficient;
+    dTotalStress_dPressure[2] = biotCoefficient;
```
sytuannguyen, line 78: Regarding the negative sign in line 74, can we explain the positive sign here?

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoromechanicsKernel.hpp
```diff
@@ -281,9 +278,17 @@ class SinglePhase :
...
+      stack.localDispFlowJacobian[a*3+0][0] += dNdX[a][2] * dTotalStress_dPressure[4] * detJxW;
+      stack.localDispFlowJacobian[a*3+0][0] += dNdX[a][1] * dTotalStress_dPressure[5] * detJxW;
+
+      stack.localDispFlowJacobian[a*3+1][0] += dNdX[a][1] * dTotalStress_dPressure[1] * detJxW;
+      stack.localDispFlowJacobian[a*3+1][0] += dNdX[a][2] * dTotalStress_dPressure[3] * detJxW;
+      stack.localDispFlowJacobian[a*3+1][0] += dNdX[a][0] * dTotalStress_dPressure[5] * detJxW;
+
+      stack.localDispFlowJacobian[a*3+2][0] += dNdX[a][2] * dTotalStress_dPressure[2] * detJxW;
+      stack.localDispFlowJacobian[a*3+2][0] += dNdX[a][1] * dTotalStress_dPressure[3] * detJxW;
+      stack.localDispFlowJacobian[a*3+2][0] += dNdX[a][0] * dTotalStress_dPressure[4] * detJxW;
```
sytuannguyen, line 291: `dTotalStress_dPressure[i]` with `i=3,4,5` are always nil. So we can simplify to:

```suggestion
      stack.localDispFlowJacobian[a*3+0][0] += dNdX[a][0] * dTotalStress_dPressure[0] * detJxW;
      stack.localDispFlowJacobian[a*3+1][0] += dNdX[a][1] * dTotalStress_dPressure[1] * detJxW;
      stack.localDispFlowJacobian[a*3+2][0] += dNdX[a][2] * dTotalStress_dPressure[2] * detJxW;
```

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoromechanicsKernel.hpp
```diff
@@ -344,9 +341,17 @@ class Multiphase :
...
+      stack.localDispFlowJacobian[a*3+0][0] += dNdX[a][2] * dTotalStress_dPressure[4] * detJxW;
+      stack.localDispFlowJacobian[a*3+0][0] += dNdX[a][1] * dTotalStress_dPressure[5] * detJxW;
+
+      stack.localDispFlowJacobian[a*3+1][0] += dNdX[a][1] * dTotalStress_dPressure[1] * detJxW;
+      stack.localDispFlowJacobian[a*3+1][0] += dNdX[a][2] * dTotalStress_dPressure[3] * detJxW;
+      stack.localDispFlowJacobian[a*3+1][0] += dNdX[a][0] * dTotalStress_dPressure[5] * detJxW;
+
+      stack.localDispFlowJacobian[a*3+2][0] += dNdX[a][2] * dTotalStress_dPressure[2] * detJxW;
+      stack.localDispFlowJacobian[a*3+2][0] += dNdX[a][1] * dTotalStress_dPressure[3] * detJxW;
+      stack.localDispFlowJacobian[a*3+2][0] += dNdX[a][0] * dTotalStress_dPressure[4] * detJxW;
```
sytuannguyen, line 354: `dTotalStress_dPressure[i]` with `i=3,4,5` are always nil. So we can simplify to:
```suggestion
      stack.localDispFlowJacobian[a*3+0][0] += dNdX[a][0] * dTotalStress_dPressure[0] * detJxW;
      stack.localDispFlowJacobian[a*3+1][0] += dNdX[a][1] * dTotalStress_dPressure[1] * detJxW;
      stack.localDispFlowJacobian[a*3+2][0] += dNdX[a][2] * dTotalStress_dPressure[2] * detJxW;
```

src/coreComponents/constitutive/solid/PorousSolid.hpp
sytuannguyen: Besides, do you want to replace the Biot coefficient by the Biot tensor?

sytuannguyen (approved):
Nice! Do you want to consider the Biot tensor in place of the Biot coefficient (as suggested by #1508)

src/coreComponents/constitutive/solid/PorousSolid.hpp
castelletto1: Yes, it is appropriate to have a minus sign here. This requires a sign change also here:
https://github.com/GEOSX/GEOSX/blob/6ab47177cbbebe1eefb6553da7ebf03d4dea8c0e/src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoromechanicsKernel.hpp#L281-L291
(Likewise in `MultiphasePoromechanicsKernel`).

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoromechanicsKernel.hpp
castelletto1: This PR is motivated by the fact that I would like the poromechanics kernels (single-phase and multiphase) to be agnostic on the specific constitutive assumptions. Irrespective of the current capabilities available in GEOSX, I would prefer to keep the general implementation so that the kernels won't need to be changed as we add new models for the Biot's coefficient.

joshua-white: yes, we may have a rotated basis and an anisotropic Biot coefficient that would fill these entries.

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoromechanicsKernel.hpp
castelletto1: Done

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoromechanicsKernel.hpp
castelletto1: Done

src/coreComponents/constitutive/solid/PorousSolid.hpp
castelletto1: Done

src/coreComponents/constitutive/solid/porosity/BiotPorosity.hpp
castelletto1: So to update porosity function will become:

- `updatePorosityFromPressure` instead of `updateFromPressure` in `PressurePorosity`
- `updatePorosityFromPressureAndStrain` instead of `updatePorosity` in `BiotPorosity`
- `updatePorosityFromProppantVolumeFraction` instead of `updateFromProppantVolumeFraction` in `ProppantPorosity`

How about just having `updatePorosity` everywhere?

CusiniM: If I m not mistaking the compilers don't love having virtual methods with the same name and just a different signature. I am almost sure I was getting warnings.

castelletto1: I'd be tempted to have a non-virtual method `update` defined in each derived class. But I know people may be opposed to this solution.

PR #1599, opened 2021-09-25
Keep neighbors with one-way ghost exchange

klevzoff:
This PR fixes a small issue in neighbor exchange shown in the image below. Here, rank 2 (rightmost) must receive data on ghosted nodes (highlighted) from rank 0 (leftmost) that owns them, but doesn't have anything to send in return. In current develop however, rank 0 is removed from rank 2's neighbor list because of the requirement for the 2-way exchange to exist, and data on ghosted nodes is never sync'ed.

ghost_nodes

Rebaseline PR: https://github.com/GEOSX/integratedTests/pull/154


rrsettgast (approved):
oof. that is a nice bug find.

PR #1609, opened 2021-09-29
Add pygeosx tutorials and update advanced xml features

cssherman:
This PR does the following:

- Implements two pygeosx tutorials (an in-situ data monitor, and initial condition modification)
- Updates the advanced xml feature documentation
- Updates the xml preprocessor to allow for multiple command line inputs and parameter overrides
- Fixes PEP8 issues with the geosx_xml_tools_package
- Adds a wrapper script that will automatically apply the xml preprocessor to input files


src/coreComponents/python/modules/geosx_xml_tools_package/geosx_xml_tools/__main__.py
```diff
@@ -5,32 +5,34 @@
 
 
 def main():
-  """Entry point for the geosx_xml_tools console script
+    """Entry point for the geosx_xml_tools console script
```
cssherman, line 8: Most of the changes below are due to changing the indentation level to match PEP8 standards.  Hiding whitespace changes in the diff is highly recommended.

scripts/automatic_xml_preprocess.sh
```diff
@@ -0,0 +1,79 @@
+#!/bin/bash
```
cssherman, line 1: This script will be linked to the build/bin directory when the user runs make geosx_xml_tools.  If you substitute geosx for geosx_preprocessed, then this script will intercept the preprocessor arguments, compile the final input xml file, and then forward to GEOSX.

Right now, it's not MPI-aware, so each process will do this in parallel.  I'm not sure if this will be an issue, or how to best avoid this yet.  We could require mpi4py as a pre-requisite, which is already going to be in the pygeosx distribution, or we could do some simple checks to look if there is a recently written output file.

corbett5 (approved):
Looks good to me, thanks for adding more examples!

examples/pygeosxExamples/hydraulicFractureWithMonitor/hydraulicFractureWithMonitor.py
```diff
@@ -9,27 +10,33 @@
 # Get the MPI rank
 comm = MPI.COMM_WORLD
 rank = comm.Get_rank()
```
TotoGaz, line 12: I believe we already discussed this, but at least finding `rank` in the `if __name__ == '__main__':` part and explicitly giving it to `run_problem` would be cleaner (less side effects).

examples/pygeosxExamples/hydraulicFractureWithMonitor/hydraulicFractureWithMonitor.py
```diff
@@ -42,11 +49,18 @@ def run_problem():
                'time': {'label': 'Time (min)',
                         'scale': 1.0 / 60.0,
                         'history': []}}
+    # PYGEOSX_QUERY_SETUP_END
 
+    # PYGEOSX_MAIN_LOOP
     # Run the code
+    query_count = 0
+    plot_frequency = 4
```
TotoGaz, line 57: Provide this `plot_frequency = 4` value as an optional value of `run_problem`?

examples/pygeosxExamples/sedovWithStressFunction/run_sedov_problem.py
```diff
@@ -18,15 +19,18 @@ def stress_fn(x):
...
     R = x[:, 0]**2 + x[:, 1]**2 + x[:, 2]**2
     return np.sin(2.0 * np.pi * R / np.amax(R))
+# PYGEOSX_STRESS_FN_END
 
 
 def run_problem():
     """
     Run the GEOSX problem
     """
+    # PYGEOSX_INITIALIZATION
```
TotoGaz, line 29: In this file, `PYGEOSX_INITIALIZATION` encloses the mpi rank stuff while in `hydraulicFractureWithMonitor.py` it is done outside `run_problem`. 
You documentation effort would surely benefit from an homogeneous way of doing things.

scripts/automatic_xml_preprocess.sh
TotoGaz: > Right now, it's not MPI-aware, so each process will do this in parallel. I'm not sure if this will be an issue, or how to best avoid this yet.

This surely can be an issue when running on a lot of cores: the filesystem could be clogged by a lot of request in the same time at the same spot (I've experienced it). Now it may be ok on fs with good caching systems I imagine.

src/coreComponents/python/modules/geosx_xml_tools_package/geosx_xml_tools/tests/test_manager.py
```diff
@@ -12,295 +13,309 @@
...
+        self.checkUnits(['1.234', 'meter'], 1.234)
 
-  def test_scale_c(self):
-    self.checkUnits(['1.234e1', 'meter'], 12.34)
+    def test_scale_c(self):
+        self.checkUnits(['1.234e1', 'meter'], 12.34)
 
-  def test_scale_d(self):
-    self.checkUnits(['1.234E1', 'meter'], 12.34)
+    def test_scale_d(self):
```
TotoGaz, line 38: I do not know which testing fwk you are using, but there probably is a parametrization on data feature that you could use. Like https://docs.pytest.org/en/6.2.x/example/parametrize.html e.g.

TotoGaz (approved):

src/coreComponents/python/modules/geosx_xml_tools_package/geosx_xml_tools/tests/test_manager.py
cssherman: Excellent idea - since I was already using unittests, I've used the parameterized package to greatly decrease the redundant code in the tests.

src/coreComponents/python/modules/geosx_xml_tools_package/geosx_xml_tools/tests/test_manager.py
```diff
@@ -115,35 +67,20 @@ def setUpClass(cls):
...
-        self.evaluateRegex('$foo$*1.234/$bar$', '4.56e7*1.234/4.56e7')
+    @parameterized.expand([['parameter_regex', '$:foo*1.234', '1.23*1.234'],
+                           ['parameter_regex', '$:foo*1.234/$:bar', '1.23*1.234/4.56e7'],
+                           ['parameter_regex', '$:foo*1.234/($:bar*$:foo)', '1.23*1.234/(4.56e7*1.23)'],
+                           ['parameter_regex', '$foo*1.234/$bar', '1.23*1.234/4.56e7'],
+                           ['parameter_regex', '$foo$*1.234/$bar', '1.23*1.234/4.56e7'],
+                           ['parameter_regex', '$foo$*1.234/$bar$', '1.23*1.234/4.56e7'],
+                           ['parameter_regex', '$blah$*1.234/$bar$', '1.23*1.234/4.56e7', True],
+                           ['parameter_regex', '$foo$*1.234/$bar$', '4.56e7*1.234/4.56e7', True]])
+    def test_sequence(self, name, parameterInput, expectedValue, expect_fail=False):
```
TotoGaz, line 78: `name` being unused here (and in other test member functions), I would simply rename `test_sequence` into `test_parameter_regex`. That would break a little of symmetry; but I really do not think it's important here.
It's always disturbing to have a parameter which is not used.

src/coreComponents/python/modules/geosx_xml_tools_package/geosx_xml_tools/tests/test_manager.py
```diff
@@ -235,40 +138,20 @@ class TestXMLProcessor(unittest.TestCase):
...
-                                    outputFile='./symbolic_parameters_processed.xml',
-                                    verbose=0,
-                                    keep_parameters=False,
-                                    keep_includes=False)
-        self.diff_xml(tmp, './symbolic_parameters_target.xml')
+    @parameterized.expand([['no_advanced_features', './no_advanced_features_input.xml', './no_advanced_features_target.xml'],
+                           ['parameters', './parameters_input.xml', './parameters_target.xml'],
+                           ['included', './included_input.xml', './included_target.xml'],
+                           ['symbolic', './symbolic_parameters_input.xml', './symbolic_parameters_target.xml']])
+    def test_sequence(self, name, input_file, target_file, expect_fail=False):
```
TotoGaz, line 145: Here too, `name` is not a parameter of the test.
Would relying on the `xml` file name be enough?
Or simply a comment at the end of the line?

src/coreComponents/python/modules/geosx_xml_tools_package/geosx_xml_tools/tests/test_manager.py
cssherman: The paremeterize decorator takes the first input and adds it to the test name, so the example I was using set it manually.  Removing this in all cases, works just fine.  It can create some odd looking names when I throw the regexes at them (e.g. test_symbolic_regex_06__1_0_2_0_2_), but they are numbered anyway so it doesn't matter.

TotoGaz: What's critical is that you are able to get the problematic parameters if it fails (which I'm quite sure should be OK).
If it's a common practice in this fwk, it's OK. 😉 Up to you!

PR #1617, opened 2021-10-02
Allow nested XML includes

klevzoff:
This PR mainly does two things:
   * Enables nested XML includes, e.g. when `a.xml` includes `b.xml` which includes `c.xml`
   * Adds a unit test that utilizes XML includes

The rest is just some cleanup along the way: 
   * some variable renaming and `const`
   * some message formatting
   * made `xmlWrapper` and `MeshUtilities` namespaces instead of classes
   * got rid of `"Problem"`, `"Included"` and `"File"` appearing as a string in multiple places
   * removed empty `Wrapper.cpp`


src/coreComponents/dataRepository/xmlWrapper.cpp
```diff
@@ -71,58 +70,78 @@ template void xmlWrapper::stringToInputVariable( Tensor< real64, 6 > & target, s
...
+      // Currently, schema only allows <Included> tags at the top level (inside <Problem>).
+      // We then proceed to merge each nested node from included file with the one in main.
+
+      xmlNode includedRootNode = includedXmlDocument.first_child();
+      GEOSX_THROW_IF_NE_MSG( string( includedRootNode.name() ), string( targetNode.name() ),
+                             "Included document root does not match the including XML node", InputError );
+
+      // Process potential includes in the included file to allow nesting
+      includedXmlDocument.append_child( filePathString ).append_attribute( filePathString ).set_value( includedFilePath.c_str() );
+      addIncludedXML( includedRootNode );
```
klevzoff, line 108: This is the essential change in this file (and PR). Everything else is just a touch up

rrsettgast (approved):

rrsettgast:
Passes the integrated tests that develop passes.

cssherman (approved):
Looks good.  Somewhere in the documentation, we should remind users that the relative paths of included files are with respect to their parent (this makes sense to me, but I could imagine someone messing it up).

PR #1624, opened 2021-10-05
Updating geosx_xml_tools installation method

cssherman:
This PR should help to harden the 'make geosx_xml_tools' step.

In the new version, if the user isn't manually requesting that a virtual environment be created, the setup script will attempt to install the required packages directly in the target environment (and then use the virtual method as a backup).  This is useful for cases where pygeosx is enabled, as the use is already creating a virtual environment.


src/coreComponents/python/CMakeLists.txt
```diff
@@ -29,11 +34,7 @@ set( GEOSX_XML_TOOLS_BINS
 
 
 add_custom_command( OUTPUT ${GEOSX_XML_TOOLS_BINS}
-                    COMMAND bash ${CMAKE_SOURCE_DIR}/../scripts/setupVirtualPythonEnvironment.bash -p ${PYTHON_POST_EXECUTABLE} -o ${CMAKE_BINARY_DIR}/python
-                    COMMAND ln -s ${CMAKE_BINARY_DIR}/python/geosx/bin/preprocess_xml ${CMAKE_BINARY_DIR}/bin/preprocess_xml
-                    COMMAND ln -s ${CMAKE_BINARY_DIR}/python/geosx/bin/format_xml ${CMAKE_BINARY_DIR}/bin/format_xml
-                    COMMAND ln -s ${CMAKE_BINARY_DIR}/python/geosx/bin/python ${CMAKE_BINARY_DIR}/bin/python
-                    COMMAND ln -s ${CMAKE_SOURCE_DIR}/../scripts/automatic_xml_preprocess.sh ${CMAKE_BINARY_DIR}/bin/geosx_preprocessed
+                    COMMAND bash ${CMAKE_SOURCE_DIR}/../scripts/setupPythonEnvironment.bash -p ${PYTHON_POST_EXECUTABLE} -b ${CMAKE_BINARY_DIR}/bin
```
klevzoff, line 37: Since the setup script is accepting `-m` and `-v` options, should there be a way for a user to pass them through a host-config?

klevzoff (approved):
I would like to have this merged. Any objections?

I'm unable to use the current version of `geosx_xml_tools` on a cluster where the `virtualenv` script is installed in a different location from python executable. This new version seems to fix it.

cssherman:
@klevzoff - this should be fine to merge.  @TotoGaz - to your question (sorry for not replying), this script in question was derived from a legacy script, which has more functionality than is probably needed.  If we were to lock down the python requirements, then some of the additional checks/build options could be removed.

src/coreComponents/python/CMakeLists.txt
cssherman: Sure, but since we are moving into a default-pygeosx direction, which is already virtual, these would be largely obsolete.

TotoGaz:
@cssherman I could read that `This branch is 4 commits ahead, 69 commits behind develop.`. Could your merge with `develop` and let this do?

TotoGaz (approved):

PR #1625, opened 2021-10-05
Adding a Pecan CPU target alongside the GPU one.

TotoGaz:
_Use case_: Some users are now developing on `Pecan` and not only launching computations.
They would greatly benefit from _both_ `CPU` _and_ `GPU` pre-compiled TPLs.

This PR (and https://github.com/GEOSX/thirdPartyLibs/pull/165) adds a target for `Pecan-CPU`.
Therefore, automatic deployments will be available for both `Pecan-GPU` and `Pecan-CPU`.

Note that
- As part of the refactoring process, the `Pecan` config files are split into `pecan-GPU.cmake` and `pecan-CPU.cmake`.
- We would have 11 builds on travis. The `Pecan-CPU` build lasts 54 minutes.
- If this PR gets accepted, I can reorder the builds such that the longest starts first (I do not want to do it now for code review purposes).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: minimal
 
 env:
   global:
-  - GEOSX_TPL_TAG=161-674
+  - GEOSX_TPL_TAG=165-679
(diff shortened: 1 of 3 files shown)


bmhan12 (approved):
LGTM!

rrsettgast:
The list of CI targets is getting rather long. @TotoGaz How many more platforms are you looking to support?

TotoGaz:
For the moment, `Pangea 2` and `Pecan` (`GPU` + `CPU`).
Ultimately, `Pangea 3` which is `power` and maybe some `ARM`, but with travis this seems complicated, so it won't be soon.

PR #1628, opened 2021-10-06
Remove memory leak when creating already created HypreVector.

TotoGaz:
Solves https://github.com/GEOSX/GEOSX/issues/1629


castelletto1 (approved):
Thank you

PR #1635, opened 2021-10-07
[Doc] advanced example of Kirsch wellbore problem

jhuang2601:
- [x] This PR documents the verification example with Kirsch wellbore problem.
- [x] Instead of using Visit, TimeHistory function and python script are employed to output and postprocess multi-dimensional data (e.g., stress and displacement).


jhuang2601:
@wrtobin Many thanks for fixing the reported timeHistory issues. Now, there is no need for querying multi-dimensional data with Visit anymore. TimeHistory functions and python scripts can do the job, as demonstrated in this example.

PR #1607, opened 2021-09-29
Implemented hydrostatic equilibrium initial condition

francoishamon:
This PR implements the initial condition needed to simulate the Northern Lights problem:
- Non-homogeneous temperature field specified as a function of depth
- Non-homogeneous component fractions specified as a function of depth
- Hydrostatic pressure computed using a fixed-point iteration scheme

Current status: hydrostatic pressure computed/accounted for in
- [X] Single-phase flow 
- [X] Compositional flow 
- [X] Poromechanical kernels 

In this PR, the implementation assumes single-phase flow. Phase contacts can be implemented later.

Documentation in #1640 

Will require rebaseline in https://github.com/GEOSX/integratedTests/pull/171

PR #1640, opened 2021-10-12
Added hydrostatic equilibrium documentation

francoishamon:
This PR adds the documentation about the hydrostatic equilibrium method (PR #1607).

It summarizes some of the discussions that took place there (comparison with the initialization with x-y-z tables; which initialization method has the priority over the other for flow simulations, etc).

This is done in a new section of the user guide called "Initial and boundary conditions", where I also plan to document the aquifer boundary condition, and where we could also mention traction.

There is no need to review now, it can wait until after the main PR is merged.


herve-gross (approved):
Great! Thanks for documenting the hydrostatic equilibrium initialization procedure.

PR #1647, opened 2021-10-18
Adding CO2 brine constitutives: brine density and viscosity based on Ezrokhi`s method 

mkhait:
Following Francois`s advice, a new fluid model is created.

- [x] Rename BrineDensity, BrineViscosity to PhillipsBrineDensity, PhilipsBrineViscosity
- [x] Rename CO2BrineFluid to PhillipsCO2BrineFluid
- [x] Add EzrokhiBrineDensity, EzrokhiBrineViscosity
- [x] Add unit tests
- [x] Implement P,T-dependent pure water density and viscosity, used in Ezrokhi, instead of constant.
- [x] Add documentation
- [x] Validate against IX
- [x]  Convert molar phase composition to mass phase composition

On the second thought, naming Phillips/Ezrokhi CO2BrineFluid only because 2 out of 5 properties are computed according to Phillips/Ezrokhi, while the rest are the same, does not seem like a good idea anymore. Ideas are welcome :)

https://github.com/GEOSX/integratedTests/pull/174


src/coreComponents/constitutive/fluid/PVTFunctions/EzrokhiBrineDensity.cpp
```diff
@@ -0,0 +1,263 @@
...
+  densities[2] = 998.21;
+  densities[3] = 997.05;
+  densities[4] = 995.65;
+  densities[5] = 995.65;
+  densities[6] = 988.04;
+  densities[7] = 983.2;
+  densities[8] = 977.76;
+  densities[9] = 971.79;
+  densities[10] = 965.31;
+  densities[11] = 958.35; // this value is used as referense density with referense pressure 100 kPa.
```
CusiniM, line 87:
```suggestion
  densities[11] = 958.35; // this value is used as reference density with reference pressure 100 kPa.
```

src/coreComponents/constitutive/fluid/PVTFunctions/EzrokhiBrineDensity.cpp
```diff
@@ -0,0 +1,263 @@
...
+  pressures[17] = 1550000;
+  pressures[18] = 2320000;
+  pressures[19] = 3350000;
+  pressures[20] = 4690000;
+  pressures[21] = 6420000;
+  pressures[22] = 8590000;
+  pressures[23] = 11300000;
+  pressures[24] = 14600000;
+  pressures[25] = 18700000;
+
```
CusiniM, line 182: The table spans quite a wide parameters space but what happens if the user inputs initial conditions that are not within the limits? Will it get an error when computing the property? Is the error easy to understand? is it Worth adding a check?

CusiniM (approved):
This looks good to me.

src/coreComponents/constitutive/fluid/PVTFunctions/EzrokhiBrineDensity.cpp
mkhait: Thanks for looking into it @CusiniM!
Actually, this whole comment is irrelevant after b9535f7bdf92304b9ae476ec90599da9b1d8d9da
I`m removing it.

src/coreComponents/constitutive/fluid/PVTFunctions/EzrokhiBrineDensity.cpp
mkhait: Great question! I`ve checked - no error occurs in this case, but what happens is that the table just silently takes the closest value (instead of extrapolation, for example). This is not obvious, and should probably be reported, but somewhere in 

https://github.com/GEOSX/GEOSX/blob/72670ad80d856bb9d2f2e286bdcad0f523c2a52c/src/coreComponents/functions/TableFunction.hpp#L356

, not in the actual property code. And there will be a lot of such reports since we`ll compute properties for all grid blocks in similar conditions...

inputFiles/compositionalMultiphaseWell/simpleCo2InjTutorial_base.xml
```diff
@@ -85,11 +85,12 @@
 
   <!-- SPHINX_FIELD_CASE_Co2_CONSTITUTIVE -->
   <Constitutive>
-    <CO2BrineFluid
+    <CO2BrinePhillipsFluid
```
francoishamon, line 88:
```suggestion
    <CO2BrineEzrokhiFluid
```
if I understood well based on the `pvtliquid.txt` in `compositionalMultiphaseWell` (but we can check all this when we do the rebaseline anyway)

francoishamon (approved):
Looks great, thanks for this work @mkhait! And we have an immediate use case for the new Ezrokhi models, so this is perfect

inputFiles/compositionalMultiphaseWell/simpleCo2InjTutorial_base.xml
mkhait: Thanks for noticing this @francoishamon! I intended to keep the previous behaviour, so `pvtliquid.txt` should have been with PhillipsBrineDensity. My bad, I`m going to restore the previous file.

src/coreComponents/constitutive/fluid/PVTFunctions/EzrokhiBrineDensity.cpp
rrsettgast: @francoishamon We can add the option to add a point on both ends of the table. This is generally what I did when creating the apertureTables. That way the extrapolation has bounds. I don't like the idea of indefinite extrapolation..if someone gives a nonsense input, it should throw an error if it goes off the end of the table....which it doesn't do now...but it should imo.

rrsettgast (commented):
This is a good example of why we need a specialized evenly spaced table type. The lookup could potentially be done without any complex logic...ok it isn't complex, but it is more complex than just integer arithmetic and a dereference.

mkhait:
> This is a good example of why we need a specialized evenly spaced table type. The lookup could potentially be done without any complex logic...ok it isn't complex, but it is more complex than just integer arithmetic and a dereference.

@rrsettgast that`s exactly what we do in OBL! Going a step further, if one finds an evenly spaced table not accurate enough for certain subspaces, he can go with a kind of LGR, which will cost only two integer arithmetic and two dereferences - still cheap.

Another thing that crossed my mind when I wrote compute method for these properties - all those difficult computations, involving exp, log, inversion - could be moved into the pre-computed tables, by extending their dimensionality from [P,T] to [P,T, y_CO2]. Some accuracy would be lost of course, since the correlation is nonlinear, but instead, it`ll probably be faster (interpolation instead of correlation evaluation) and more general (no need to derive partial derivatives).

rrsettgast:
> @rrsettgast that`s exactly what we do in OBL! Going a step further, if one finds an evenly spaced table not accurate enough for certain subspaces, he can go with a kind of LGR, which will cost only two integer arithmetic and two dereferences - still cheap.
> 
> Another thing that crossed my mind when I wrote compute method for these properties - all those difficult computations, involving exp, log, inversion - could be moved into the pre-computed tables, by extending their dimensionality from [P,T] to [P,T, y_CO2]. Some accuracy would be lost of course, since the correlation is nonlinear, but instead, it`ll probably be faster (interpolation instead of correlation evaluation) and more general (no need to derive partial derivatives). This is the OBL method, basically.

@mkhait I think we should look at all of these types of things when we get to the tuning portion of the program. I am all for neat tricks, but I will say that storing data in tables might result in more register pressure....but IMO it really is hard to say until we actually get in there and tune it. If you have a different opinion on this, it would be great to learn from your point of view.

mkhait:
> > @rrsettgast that`s exactly what we do in OBL! Going a step further, if one finds an evenly spaced table not accurate enough for certain subspaces, he can go with a kind of LGR, which will cost only two integer arithmetic and two dereferences - still cheap.
> > Another thing that crossed my mind when I wrote compute method for these properties - all those difficult computations, involving exp, log, inversion - could be moved into the pre-computed tables, by extending their dimensionality from [P,T] to [P,T, y_CO2]. Some accuracy would be lost of course, since the correlation is nonlinear, but instead, it`ll probably be faster (interpolation instead of correlation evaluation) and more general (no need to derive partial derivatives). This is the OBL method, basically.
> 
> @mkhait I think we should look at all of these types of things when we get to the tuning portion of the program. I am all for neat tricks, but I will say that storing data in tables might result in more register pressure....but IMO it really is hard to say until we actually get in there and tune it. If you have a different opinion on this, it would be great to learn from your point of view.

@rrsettgast Indeed you are right about register pressure, it's true for high-dimensional cases (4-5 DOF and higher), but also depends on the interpolation method. I`ll try to compare assembly on the SPE10 test (we already discussed this with @francoishamon) and let you know about the results.

rrsettgast:
> @rrsettgast Indeed you are right about register pressure, it's true for high-dimensional cases (4-5 DOF and higher), but also depends on the interpolation method. I`ll try to compare assembly on the SPE10 test (we already discussed this with @francoishamon) and let you know about the results.

One disclaimer. I am not overly concerned with full assembly kernels...However, when it comes to residual assembly in the case of matrix-free methods, it would probably matter a little more...and of course we have a lot less pressure for those cases, but we would really want to maximize the optimization of those kernels.

PR #1662, opened 2021-10-26
Contact relationships interface refactoring

CusiniM:
https://github.com/GEOSX/integratedTests/pull/175


src/coreComponents/constitutive/contact/ContactBase.cpp
```diff
@@ -29,16 +31,122 @@ namespace constitutive
 ContactBase::ContactBase( string const & name,
                           Group * const parent ):
   ConstitutiveBase( name, parent ),
-  m_penaltyStiffness( 0.0 )
+  m_penaltyStiffness( 0.0 ),
+  m_shearStiffness( 0.0 ),
```
andrea-franceschini, line 35: Do we want to generalize this class for non-penalty approaches in the future?
The name `ContactBase` is very general ...

src/coreComponents/physicsSolvers/multiphysics/LagrangianContactSolver.cpp
```diff
@@ -443,8 +443,8 @@ void LagrangianContactSolver::resetStateToBeginningOfStep( DomainPartition & dom
     {
       arrayView2d< real64 > const & traction = subRegion.getReference< array2d< real64 > >( viewKeyStruct::tractionString() );
       arrayView2d< real64 > const & deltaTraction = subRegion.getReference< array2d< real64 > >( viewKeyStruct::deltaTractionString() );
-      arrayView2d< real64 > const & localJump = subRegion.getReference< array2d< real64 > >( viewKeyStruct::localJumpString() );
-      arrayView2d< real64 const > const & previousLocalJump = subRegion.getReference< array2d< real64 > >( viewKeyStruct::previousLocalJumpString() );
+      arrayView2d< real64 > const & dispJump = subRegion.getReference< array2d< real64 > >( viewKeyStruct::dispJumpString() );
```
andrea-franceschini, line 446: Changing `localJump` to `dispJump` is fine!
Thanks.

andrea-franceschini (approved):
Thanks Matteo for this refactoring.

src/coreComponents/constitutive/contact/ContactBase.cpp
joshua-white: I agree we should maybe have a hierarchy `ContactBase->PenaltyBase->SomePenaltyModel` to move these specific member variables somewhere more appropriate.

src/coreComponents/constitutive/contact/ContactBase.hpp
```diff
@@ -132,28 +151,113 @@ class ContactBase : public ConstitutiveBase
...
+
+
+GEOSX_HOST_DEVICE
+void ContactBaseUpdates::addPressureToTraction( real64 const & pressure,
+                                                bool const isOpen,
+                                                arraySlice1d< real64 > const & tractionVector,
+                                                real64 & dTraction_dPressure ) const
+{
+  tractionVector[0] -= pressure;
+  dTraction_dPressure = isOpen ? -1.0 : 0.0;
```
joshua-white, line 250: the value and derivative aren't consistent with one another. They must either both depend on `isOpen` or both  be independent of `isOpen`. Physically I don't think it should depend on `isOpen` at all.  If there is pressure in the fracture, it will exert a traction, and its derivative should just be `-1` always.

src/coreComponents/constitutive/contact/ContactBase.hpp
```diff
@@ -132,28 +151,113 @@ class ContactBase : public ConstitutiveBase
...
+GEOSX_HOST_DEVICE
+void ContactBaseUpdates::computeTraction( localIndex const k,
+                                          arraySlice1d< real64 const > const & oldDispJump,
+                                          arraySlice1d< real64 const > const & dispJump,
+                                          arraySlice1d< real64 > const & tractionVector,
+                                          arraySlice2d< real64 > const & dTractionVector_dJump ) const
+{
+  GEOSX_UNUSED_VAR( k, oldDispJump );
+
+  tractionVector[0] = dispJump[0] >= 0 ? 0.0 : m_penaltyStiffness * dispJump[0];
```
joshua-white, line 234: Personally I would identify this as a specific model (`NormalPenaltyFrictionlessContact`) rather than make it the base implementation.

src/coreComponents/constitutive/contact/CoulombContact.cpp
```diff
@@ -47,6 +48,11 @@ CoulombContact::CoulombContact( string const & name, Group * const parent ):
     setApplyDefaultValue( -1 ).
     setInputFlag( InputFlags::OPTIONAL ).
```
joshua-white, line 49: The trouble with letting users provide either the angle or the coefficient directly is that both are optional.  I might pick one and make it required.  The user can trivially do the conversion externally.

src/coreComponents/constitutive/contact/ContactBase.cpp
rrsettgast: I agree with @CusiniM . I would like to se the constitutive relation be just that. The constraint enforcement method should indeed be a separate/PhysicsSolver type thing.

src/coreComponents/constitutive/contact/ContactBase.hpp
```diff
@@ -60,8 +70,7 @@ class ContactBaseUpdates
   GEOSX_HOST_DEVICE
   inline
   virtual real64 computeEffectiveAperture( real64 const aperture,
-                                           real64 & dEffectiveAperture_dAperture ) const
-  { GEOSX_UNUSED_VAR( aperture, dEffectiveAperture_dAperture ); return 0.0; }
+                                           real64 & dEffectiveAperture_dAperture ) const;
```
rrsettgast, line 73: should we be using the term "aperture" any longer?

rrsettgast (approved):
The only question if have is whether or not the stiffness should be a parameter of the constitutive relation in all cases?

src/coreComponents/constitutive/contact/ContactBase.hpp
CusiniM: I think `hydraulicAperture` may be a good term. I also don't love the fact that this `apertureTable` is now owned by the contact model but, at the moment, I can't think of a better location for it. This has more to do with how flow properties of the fractures are computed than with the contact model.

PR #1670, opened 2021-10-30
Fixed VEM unit test on Lassen

francoishamon:
This PR fixes the VEM unit test that started failing on Lassen after the recent merge of https://github.com/GEOSX/GEOSX/pull/1489. It is my bad, I was convinced that unit tests were checked automatically by Travis, but it is not the case for GPU platforms.


castelletto1 (approved):
Thank you!

PR #1676, opened 2021-11-02
Added logging for boundary conditions in flow solvers

francoishamon:
I propose to add some logging in the functions in charge of boundary condition application in the solvers (for now, only done in the flow solvers in this PR). The motivation for this PR is that it is in general difficult to know whether a boundary condition is actually applied, and if applied, whether the `targetSet` actually contains some elements/faces (I am struggling with this for an imported mesh).

To solve this problem, I added a `logLevel` in the `FieldSpecificationBase` class, and if the `logLevel` is sufficient large (larger than 2), then the solver outputs a message when the boundary condition is applied (with the size of the `targetSet`, the `setName`, the `targetGroup` name, the time, etc) at the first Newton iteration. 

This solution works for me, but I wonder if this is ok for everyone and can be propagated elsewhere (as we had a discussion on this in a dev meeting, and @klevzoff posted https://github.com/GEOSX/GEOSX/issues/1613 on the topic).

Requires a rebaseline of all the integrated tests if everyone likes this proposal.

Rebaseline done in https://github.com/GEOSX/integratedTests/pull/176


src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -890,11 +891,19 @@ void CompositionalMultiphaseBase::applySourceFluxBC( real64 const time,
...
+                        string const & setName,
                         SortedArrayView< localIndex const > const & targetSet,
                         Group & subRegion,
                         string const & )
   {
+    if( fs.getLogLevel() >= 2 && m_nonlinearSolverParameters.m_numNewtonIterations == 0 )
+    {
+      localIndex const numTargetElems = MpiWrapper::sum( targetSet.size() );
+      GEOSX_LOG_RANK_0( GEOSX_FMT(
+                          "CompositionalMultiphaseBase {}: at time {}s, the <{}> boundary condition named \"{}\" is applied to the element set named \"{}\" in subRegion \"{}\". \nThe total number of target elements (including ghost elements) is {}. \nNote that if this number is equal to zero for all subRegions, the boundary condition will not be applied on this element set.",
```
TotoGaz, line 903: One of the advantages of `fmt` is to be able to factorise some error messages.
Please refactor the full call or at least the main error string.

rrsettgast: Would it be useful to include the actual flux value?

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseFVM.cpp
```diff
@@ -444,6 +444,14 @@ void SinglePhaseFVM< BASE >::applyFaceDirichletBC( real64 const time_n,
                          string const & )
   {
     BoundaryStencil const & stencil = fluxApprox.getStencil< BoundaryStencil >( mesh, setName );
+    if( fs.getLogLevel() >= 2 && m_nonlinearSolverParameters.m_numNewtonIterations == 0 )
+    {
+      localIndex const numTargetFaces = MpiWrapper::sum( stencil.size() );
+      GEOSX_LOG_RANK_0( GEOSX_FMT(
+                          "SinglePhaseFVM {}: at time {}s, the <{}> boundary condition named \"{}\" is applied to the face set named \"{}\" in \"{}\". \nThe total number of target faces (including ghost faces) is {}. \nNote that if this number is equal to zero, the boundary condition will not be applied on this face set.",
+                          this->getName(), time_n+dt, AquiferBoundaryCondition::catalogName(), fs.getName(), setName, targetGroup.getName(), numTargetFaces ) );
```
rrsettgast, line 452: Same comment here about outputting the value of the BC.

rrsettgast (approved):

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -890,11 +891,19 @@ void CompositionalMultiphaseBase::applySourceFluxBC( real64 const time,
...
                    [&]( FieldSpecificationBase const & fs,
-                        string const &,
+                        string const & setName,
                         SortedArrayView< localIndex const > const & targetSet,
                         Group & subRegion,
                         string const & )
   {
+    if( fs.getLogLevel() >= 2 && m_nonlinearSolverParameters.m_numNewtonIterations == 0 )
+    {
+      localIndex const numTargetElems = MpiWrapper::sum( targetSet.size() );
```
klevzoff, line 901: Not that it matters for face BCs realistically, but for full correctness we should use the appropriate type to refer to global mesh quantities (we're making this mistake in a few more places I think): 
```suggestion
      globalIndex const numTargetElems = MpiWrapper::sum< globalIndex >( targetSet.size() );
```

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -1086,11 +1095,19 @@ void CompositionalMultiphaseBase::applyDirichletBC( real64 const time,
...
+      GEOSX_LOG_RANK_0( GEOSX_FMT(
+                          "CompositionalMultiphaseBase {}: at time {}s, the <{}> boundary condition named \"{}\" is applied to the element set named \"{}\" in subRegion \"{}\". \nThe total number of target elements (including ghost elements) is {}. \nNote that if this number is equal to zero for all subRegions, the boundary condition will not be applied on this element set.",
```
klevzoff, line 1107: Not that there is a definitive rule on GEOSX on this, but I've always preferred single quotes to refer to named entities in error messages, if not for any other reason, then at least so we don't have to keep escaping double quotes in format strings. I also suggested removing "named" because the following quotes already signal that it's referring to a name (at least in my opinion).
```suggestion
                          "CompositionalMultiphaseBase {}: at time {}s, the <{}> boundary condition '{}' is applied to the element set '{}' in subRegion '{}'. \nThe total number of target elements (including ghost elements) is {}. \nNote that if this number is equal to zero for all subRegions, the boundary condition will not be applied on this element set.",
```

klevzoff (approved):
It would be great if the debug messaging was centralized in one place, e.g. `FieldSpecificationManager::apply()` (right before invoking the user lambda), but this would also result in losing some context (like which solver is triggering the BC) and might also trigger in unexpected/undesired places (e.g. in `TwoPointFluxApproximation` when constructing the boundary stencil). So I don't have a better proposition.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
francoishamon: I refactored the error messages and factorized them a little more. Now I reuse the error messages in multiple locations, and it seems to work well.

Now, I also output the scale of the boundary condition. It is not exactly the value of the flux if a function is used to get a time-dependent flux, but it is a good start. The actual flux value is hidden in the `applyBoundaryConditionToSystem` and then multiplied by dt, so it is hard to get access to it from the solver. That is another good reason to try to move these log messages inside `FieldSpecificationBase` at some point.

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseFVM.cpp
francoishamon: In single-phase flow as well, I output the scale of the boundary condition for SourceFlux and Dirichlet BCs. 

For Aquifer, the flux computation is quite convoluted, but yes it would definitely be useful to output the cumulative aquifer flux from [here](https://github.com/GEOSX/GEOSX/blob/bb63df344b130034593a4163a58be6a55184a955/src/coreComponents/physicsSolvers/fluidFlow/FlowSolverBase.cpp#L488) for debugging purposes. I will add that.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
francoishamon: Good point, I made the modification that you suggested

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
francoishamon: Done, I made the modifications here and propagated it across the solver

PR #1677, opened 2021-11-03
Introduce ContactSolverBase

CusiniM:
https://github.com/GEOSX/LvArray/pull/252

https://github.com/GEOSX/integratedTests/pull/207


src/coreComponents/physicsSolvers/SolverBase.hpp
```diff
@@ -213,6 +213,32 @@ class SolverBase : public ExecutableGroup
...
+   * @param dofManager degree-of-freedom manager associated with the linear system
+   * @param matrix the system matrix
+   * @param rhs the system right-hand side vector
+   * @param solution the solution vector
+   * @param lastResidual (in) target value below which to reduce residual norm, (out) achieved residual norm
+   * @return return true if line search succeeded, false otherwise
+   *
+   */
+  virtual bool
+  lineSearchWithParabolicInterpolation ( real64 const & time_n,
```
andrea-franceschini, line 232: Thanks for keeping the parabolic version! It is slightly better than the simple linear one.
For further details, see
Kelley, Carl T. _Solving nonlinear equations with Newton's method_. Society for Industrial and Applied Mathematics, 2003.

src/coreComponents/physicsSolvers/contact/ContactSolverBase.cpp
```diff
@@ -0,0 +1,300 @@
...
+      elemManager.forElementRegions< SurfaceElementRegion >( [&] ( SurfaceElementRegion & region )
+      {
+        region.forElementSubRegions< EmbeddedSurfaceSubRegion >( [&]( EmbeddedSurfaceSubRegion & subRegion )
+        {
+
+          subRegion.registerWrapper< array2d< real64 > >( viewKeyStruct::dispJumpString() ).
+            setPlotLevel( PlotLevel::LEVEL_0 ).
+            reference().resizeDimension< 1 >( 3 );
+
+          subRegion.registerWrapper< array2d< real64 > >( viewKeyStruct::deltaDispJumpString() ).
```
andrea-franceschini, line 91: Why do we have `deltaDispJump` but not `deltaTraction`?
Shouldn't either or both make more sense?
I mean, I don't see why to distinguish. Do you have a specific reason?

andrea-franceschini (approved):
Thank you for this very big unification work!

src/coreComponents/physicsSolvers/contact/ContactSolverBase.cpp
CusiniM: I think it's only because the `deltaDispJump`  was used by the `SolidMechanicsEmbeddedFracturesSolver` trying to be consistent with what's done for primary variables in the flow solvers with `deltaPressure`, etc. We could add `deltaTraction` as well. I think that there is still some work to do to "consistify" the way variables are updated across the solvers though.

src/coreComponents/physicsSolvers/SolverBase.cpp
```diff
@@ -458,162 +562,223 @@ real64 SolverBase::nonlinearImplicitStep( real64 const & time_n,
     if( dtAttempt > 0 )
     {
       resetStateToBeginningOfStep( domain );
+      resetConfigurationToBeginningOfStep( domain );
```
TotoGaz, line 565: Why don't you call `resetConfigurationToBeginningOfStep` from the already existing `resetStateToBeginningOfStep`?
There seems to be no need to add another method for `SolverBase`?

src/coreComponents/physicsSolvers/SolverBase.cpp
```diff
@@ -458,162 +562,223 @@ real64 SolverBase::nonlinearImplicitStep( real64 const & time_n,
...
+    // Configuration loop
+    for( configurationLoopIter = 0; configurationLoopIter < maxConfigurationIter; ++configurationLoopIter )
     {
-      GEOSX_LOG_LEVEL_RANK_0( 1, GEOSX_FMT( "    Attempt: {:2}, NewtonIter: {:2}", dtAttempt, newtonIter ) );
-
-      // zero out matrix/rhs before assembly
-      m_localMatrix.zero();
-      m_rhs.zero();
-
+      if( getLogLevel() >= 1 )
```
TotoGaz, line 574: Maybe simply call `outputConfigurationStatistics( domain );` and let each solver decide what it wants to display depending on the appropriate `logLevel`?

src/coreComponents/physicsSolvers/SolverBase.cpp
```diff
@@ -458,162 +562,223 @@ real64 SolverBase::nonlinearImplicitStep( real64 const & time_n,
...
+          break;
+        }
+
+        // do line search in case residual has increased
+        if( m_nonlinearSolverParameters.m_lineSearchAction != NonlinearSolverParameters::LineSearchAction::None
+            && residualNorm > lastResidual )
         {
-          if( m_nonlinearSolverParameters.m_lineSearchAction == NonlinearSolverParameters::LineSearchAction::Attempt )
+          bool lineSearchSuccess = false;
+          if( m_nonlinearSolverParameters.m_lineSearchInterpType == NonlinearSolverParameters::LineSearchInterpolationType::Linear )
```
TotoGaz, line 652: Maybe build a `NonlinearSolverParameters::LineSearchInterpolationType -> callable` map so you do not have to these long lines there?

src/coreComponents/physicsSolvers/SolverBase.cpp
```diff
@@ -458,162 +562,223 @@ real64 SolverBase::nonlinearImplicitStep( real64 const & time_n,
...
           {
-            GEOSX_LOG_LEVEL_RANK_0( 1, "        Line search failed to produce reduced residual. Accepting iteration." );
+            residualNorm = lastResidual;
+            lineSearchSuccess = lineSearch( time_n,
```
TotoGaz, line 655: It's more frequent to see `error codes` returned.
It's richer than a boolean because you can have information based on the value of the returned error code.

src/coreComponents/physicsSolvers/SolverBase.cpp
```diff
@@ -458,162 +562,223 @@ real64 SolverBase::nonlinearImplicitStep( real64 const & time_n,
...
+        {
+          break;
+        }
+      }
+      else
+      {
+        break; // get out of configuration loop and cut the time-step if you can.
+      }
+
+    }  // end of configuration loop
```
TotoGaz, line 767: We can imagine to have a policy to make convergence easier, but I think that this policy could be `Solver` based and not imposed to all solvers.

Maybe a code like
```c++
if( isNewtonConverged ) simplify( domain );
```
which lets all the solvers implement their own versions of `simplify` could be a good balance.
If this removes `resetStateToBeginningOfStep`, `setSimplestConfigurationState` or `updateConfiguration` from the `public` interface of `SolverBase` which is already bloat, that could be great.

src/coreComponents/physicsSolvers/SolverBase.hpp
```diff
@@ -455,6 +482,32 @@ class SolverBase : public ExecutableGroup
...
+   * @brief resets the configuration to the beginning of the time-step.
+   * @param domain the domain containing the mesh and fields
+   */
+  virtual void resetConfigurationToBeginningOfStep( DomainPartition & domain );
+
+  /**
+   * @brief set the simplest configuration state.
+   * @param domain the domain containing the mesh and fields
+   */
+  virtual bool setSimplestConfigurationState( DomainPartition & domain ) const;
```
TotoGaz, line 508: Does this really make sense to all the solvers that inherit from `SolverBase`.
If not, I would argue that this should go down to `ContactSolverBase`, and replaced by a wide `simplify` (or any better name), see other comment.

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.cpp
```diff
@@ -616,14 +616,14 @@ void SolidMechanicsLagrangianFEM::applyDisplacementBCImplicit( real64 const time
...
     bc.applyBoundaryConditionToSystem< FieldSpecificationEqual,
-                                       parallelDevicePolicy< 32 > >( targetSet,
-                                                                     time,
-                                                                     targetGroup,
-                                                                     fieldName,
-                                                                     dofKey,
-                                                                     dofManager.rankOffset(),
-                                                                     localMatrix,
-                                                                     localRhs );
+                                       serialPolicy >( targetSet,
```
TotoGaz, line 619: Why do you need to change this?

src/coreComponents/physicsSolvers/contact/ContactSolverBase.cpp
```diff
@@ -0,0 +1,296 @@
...
+    setDescription( "Name of contact relation to enforce constraints on fracture boundary." );
+
+  this->getWrapper< string >( viewKeyStruct::discretizationString() ).
+    setInputFlag( InputFlags::FALSE );
+}
+
+ContactSolverBase::~ContactSolverBase()
+{
+  // TODO Auto-generated destructor stub
+}
```
TotoGaz, line 62: Ue the `= default` feature in the header please.

src/coreComponents/physicsSolvers/contact/ContactSolverBase.hpp
```diff
@@ -0,0 +1,171 @@
...
+  struct FractureState
+  {
+    static constexpr integer STICK = 0;    ///< element is closed: no jump across the discontinuity.
+    static constexpr integer SLIP = 1;     ///< element is sliding: no normal jump across the discontinuity, but sliding is allowed.
+    static constexpr integer NEW_SLIP = 2; ///< element just starts sliding: no normal jump across the discontinuity, but sliding is
+                                           ///< allowed.
+    static constexpr integer OPEN = 3;     ///< element is open: no constraints are imposed.
+  };
+
+  string fractureStateToString( integer const & state ) const
```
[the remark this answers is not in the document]
klevzoff: We can use `ENUM_STRINGS` to automate this, but only if we make this an actual `enum` or `enum class`. In the former case, enumeration labels will leak to `ContactBase` scope (which many not be too bad), but will implicitly convert to `integer` so we can keep using `array1d<integer>` to store fracture states. In the latter case, labels will remain properly scoped, but will not implicitly convert - which means we either need to use explicit casts for every comparison/update, or change storage to `array1d<FractureState>`, which may or may not work (in particular, things like restart and plot outputs are likely to break). 

A compromise between the two could be
```c++
struct FractureState
{
  enum State : integer
  {
    Stick = 0,
    Slip = 1,
    NewSlip = 2,
    Open = 3
  }
};
...
ENUM_STRINGS( ContactSolverBase::FractureState::State, "stick", ... );
```
so we can get all three: scoping (`FractureState::Open`), implicit conversion and string conversion.

src/coreComponents/physicsSolvers/SolverBase.cpp
```diff
@@ -855,11 +1020,32 @@ void SolverBase::updateState( DomainPartition & GEOSX_UNUSED_PARAM( domain ) )
...
+
+void SolverBase::resetConfigurationToBeginningOfStep( DomainPartition & GEOSX_UNUSED_PARAM( domain ) )
+{
+  // For most solvers there is nothing to do.
+}
+
 void SolverBase::resetStateToBeginningOfStep( DomainPartition & GEOSX_UNUSED_PARAM( domain ) )
 {
   GEOSX_ERROR( "SolverBase::ResetStateToBeginningOfStep called!. Should be overridden." );
 }
```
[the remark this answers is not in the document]
CusiniM: Honestly I have followed the pattern used for other virtual functions in `SolverBase`. I thought there was a specific reason for not having pure virtual methods in `SolverBase` but I cnanot remember what it was. If we can have pure virtuals I am in favour of it and I think we should change a number of them.

src/coreComponents/physicsSolvers/SolverBase.cpp
CusiniM: I guess something of that sort would clean up `SolverBase` which is certainly needed. I am not sure `simplify` is the best name for it though. Does anyone have another suggestion?

src/coreComponents/physicsSolvers/SolverBase.cpp
CusiniM: Yeah, this is probably a good idea.

src/coreComponents/physicsSolvers/SolverBase.cpp
CusiniM: I think in some cases we want to reset the state without resetting the configuration.

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.cpp
CusiniM: I don't actually. I am not sure why they are there.

src/coreComponents/physicsSolvers/contact/ContactSolverBase.hpp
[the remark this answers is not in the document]
CusiniM: I have done as suggested by @klevzoff

src/coreComponents/physicsSolvers/SolverBase.cpp
```diff
@@ -494,162 +598,221 @@ real64 SolverBase::nonlinearImplicitStep( real64 const & time_n,
...
+      {
+        isConfigurationLoopConverged = updateConfiguration( domain );
+
+        if( isConfigurationLoopConverged )
+        {
+          break; // get out of configuration loop coz everything converged.
+        }
+        else
+        {
+          GEOSX_LOG_LEVEL_RANK_0( 1, "  --------Configuration did not converge. Testing new configuration." );
```
klevzoff, line 783: I've opened a whole issue about this, our logging is very inconsistent (indentation, padding characters, verbosity levels). The configuration loop will be the latest addition :) This is general comment, not something you should try to fix in this PR.

src/coreComponents/physicsSolvers/contact/LagrangianContactSolver.cpp
```diff
@@ -52,43 +51,22 @@ namespace geosx
...
 using namespace finiteElement;
 
-constexpr integer LagrangianContactSolver::FractureState::STICK;
-constexpr integer LagrangianContactSolver::FractureState::SLIP;
-constexpr integer LagrangianContactSolver::FractureState::NEW_SLIP;
-constexpr integer LagrangianContactSolver::FractureState::OPEN;
+// constexpr integer LagrangianContactSolver::FractureState::Stick;
+// constexpr integer LagrangianContactSolver::FractureState::Slip;
+// constexpr integer LagrangianContactSolver::FractureState::NewSlip;
+// constexpr integer LagrangianContactSolver::FractureState::Open;
```
klevzoff, line 59:
```suggestion
```

klevzoff (approved):
Good job! Comments are mostly minor and addressable in future PRs.

src/coreComponents/mesh/EmbeddedSurfaceSubRegion.hpp
```diff
@@ -212,7 +212,7 @@ class EmbeddedSurfaceSubRegion : public SurfaceElementSubRegion
     static constexpr char const * deltaDispJumpString()     { return "deltaDisplacementJump"; }
 
     /// @return Fracture traction string
-    static constexpr char const * fractureTractionString()  { return "fractureTraction"; }
+    static constexpr char const * fractureTractionString()  { return "traction"; }
```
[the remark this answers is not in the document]
CusiniM: Yes, I should finish migrating this to `ContactExtrinsicData` and then get rid of these keys.

src/coreComponents/physicsSolvers/SolverBase.cpp
```diff
@@ -494,162 +598,223 @@ real64 SolverBase::nonlinearImplicitStep( real64 const & time_n,
...
 
-      // update non-primary variables (constitutive models)
-      updateState( domain );
+        // update non-primary variables (constitutive models)
+        updateState( domain );
 
-      lastResidual = residualNorm;
-    }
+        lastResidual = residualNorm;
+      } // end of Newton's loop
```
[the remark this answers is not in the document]
CusiniM: I don't know how hard it would be. It would certainly make this function a lot more readable. I am going to try to have a look into it.

src/coreComponents/physicsSolvers/SolverBase.cpp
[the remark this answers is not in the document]
CusiniM: I think some functions could be easily made virtual and others would create the issue described by Sergey. I think this requires rethinking a bit whether having a single base class for all solvers is or not the best choice and it's outside the scope of this PR.

src/coreComponents/physicsSolvers/contact/ContactSolverBase.cpp
```diff
@@ -0,0 +1,297 @@
...
+            setPlotLevel( PlotLevel::LEVEL_0 ).
+            setRegisteringObjects( this->getName()).
+            setDescription( "An array that holds the tractions on the fracture." ).
+            reference().resizeDimension< 1 >( 3 );
+
+          subRegion.registerWrapper< array1d< integer > >( viewKeyStruct::fractureStateString() ).
+            setPlotLevel( PlotLevel::LEVEL_0 ).
+            setRegisteringObjects( this->getName()).
+            setDescription( "An array that holds the fracture state." );
+          initializeFractureState( meshLevel, viewKeyStruct::fractureStateString() );
```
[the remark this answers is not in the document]
CusiniM: ah, yeah, let me fix this functions.

src/coreComponents/physicsSolvers/contact/ContactSolverBase.cpp
```diff
@@ -0,0 +1,297 @@
...
+
+void ContactSolverBase::initializeFractureState( MeshLevel & mesh,
+                                                 string const & fieldName ) const
+{
+  GEOSX_MARK_FUNCTION;
+  ElementRegionManager & elemManager = mesh.getElemManager();
+
+  elemManager.forElementSubRegions< FaceElementSubRegion >( [&]( FaceElementSubRegion & subRegion )
+  {
+    if( subRegion.hasWrapper( viewKeyStruct::tractionString() ) )
```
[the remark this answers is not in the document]
CusiniM: I don't think so. This must be an old check. It's not really for the solver I wrote though so I will check with @andrea-franceschini ?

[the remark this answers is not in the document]
andrea-franceschini: This wrapper was used to identify the `faceElementSubRegions` where I am solving Lagrangian contact mechanics, that is I want to avoid `faceElementSubRegions` of other solvers. Is there the possibility of having `faceElementSubRegions` with no `FractureState` defined on them?

src/coreComponents/physicsSolvers/contact/ContactSolverBase.cpp
```diff
@@ -0,0 +1,297 @@
...
+
+  return dtReturn;
+}
+
+void ContactSolverBase::computeFractureStateStatistics( DomainPartition const & domain,
+                                                        globalIndex & numStick,
+                                                        globalIndex & numSlip,
+                                                        globalIndex & numOpen ) const
+{
+  MeshLevel const & mesh = domain.getMeshBody( 0 ).getMeshLevel( 0 );
```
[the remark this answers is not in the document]
andrea-franceschini: I would vote for the per-body statistics. IMO it's not so significant the global statistics.

src/coreComponents/physicsSolvers/SolverBase.hpp
andrea-franceschini: I think `defaultConfiguration` works fine.

andrea-franceschini (approved):

PR #1678, opened 2021-11-04
Added warning and error messages in fluid and wells

francoishamon:
This PR implements some warning/error messages suggested by a Total user that had a problematic input file:
- Error when `useMass` is not the same in the flow and well solvers
- Error when the formation volume factor (resp. the viscosity) is not a decreasing (resp. increasing) function of pressure
- Error when the trajectory of the well is invalid (going down and then going up)
- Less logging when the well segments are small (seems to happen more than what I initially thought)

New baselines in https://github.com/GEOSX/integratedTests/pull/186


src/coreComponents/constitutive/fluid/BlackOilFluidBase.hpp
```diff
@@ -155,10 +155,17 @@ class BlackOilFluidBase : public MultiFluidBase
...
    */
-  void validateTable( TableFunction const & table ) const;
+  virtual void validateFormationVolumeFactorTable( TableFunction const & table ) const;
+
+  /**
+   * @brief Check that the viscosity table values make sense
+   * @param[in] table the values in the oil or gas table
+   */
+  virtual void validateViscosityTable( TableFunction const & table ) const;
+
```
TotoGaz, line 168: Please document that these functions `\throw` an exception in case of error.

src/coreComponents/mesh/generators/InternalWellGenerator.cpp
```diff
@@ -220,28 +223,45 @@ void InternalWellGenerator::constructPolylineNodeToSegmentMap()
     real64 const segmentLength = LvArray::tensorOps::l2Norm< 3 >( vSeg );
 
     // various checks and warnings on the segment and element length
-    GEOSX_THROW_IF( segmentLength < 1e-2,
-                    "Error in the topology of well " << getName() <<
-                    ": we detected a polyline segment measuring less than 1 cm",
+    GEOSX_THROW_IF( segmentLength < 1e-5,
```
TotoGaz, line 226: I'm afraid that (here and below) you're just hard coding other values that will fail in other contexts.
You should define default values but let the user change this if needed.

src/coreComponents/mesh/generators/InternalWellGenerator.cpp
```diff
@@ -220,28 +223,45 @@ void InternalWellGenerator::constructPolylineNodeToSegmentMap()
...
+                    "Error in the topology of well '" << getName() <<
+                    "': we detected a polyline segment measuring less than 1 cm",
```
TotoGaz, line 228: Please update the error message accordingly (`1 cm` -> `segmentLength m`).

src/coreComponents/mesh/generators/InternalWellGenerator.cpp
```diff
@@ -220,28 +223,45 @@ void InternalWellGenerator::constructPolylineNodeToSegmentMap()
...
 
     // map the polyline node ids to the polyline segment ids
     m_polyNodeToSegmentMap[ipolyNode_a].insert( iseg );
     m_polyNodeToSegmentMap[ipolyNode_b].insert( iseg );
   }
+
+  if( foundSmallSegment )
+  {
+    GEOSX_LOG_RANK_0(
+      "\nWarning: we detected a segment measuring less than 1 m in the topology of well '" << getName() << "'\n"
```
TotoGaz, line 255: Same, please update the error message accordingly (`1 m` -> `minSegmentLength m`).

src/coreComponents/mesh/generators/InternalWellGenerator.cpp
```diff
@@ -220,28 +223,45 @@ void InternalWellGenerator::constructPolylineNodeToSegmentMap()
...
+                                                                                           << "The simulation can proceed like that, but small segments may cause numerical issues, so it is something to keep an eye on.\n" );
+  }
+
+  if( foundSmallWellElement )
+  {
+    GEOSX_LOG_RANK_0( "\nWarning: the chosen number of well elements per polyline segment (" << m_numElemsPerSegment <<
+                      ") leads to well elements measuring less than 1 cm in the topology of well '" << getName() << "'.\n" <<
```
TotoGaz, line 262: And this `1 cm` will fail soon.

src/coreComponents/constitutive/fluid/BlackOilFluidBase.hpp
francoishamon: I added a line in the Doxygen to say that these functions are throwing an exception when the tables are invalid.

src/coreComponents/mesh/generators/InternalWellGenerator.cpp
francoishamon: I added an input parameter to let the user choose the value.

src/coreComponents/mesh/generators/InternalWellGenerator.cpp
francoishamon: I updated the error message.

src/coreComponents/mesh/generators/InternalWellGenerator.cpp
francoishamon: I updated the error message

src/coreComponents/mesh/generators/InternalWellGenerator.cpp
francoishamon: I added an input parameter to let the user choose the value.

TotoGaz (approved):

PR #1682, opened 2021-11-08
Enabling multiple mesh bodies

rrsettgast:
This PR implements the concept of multiple mesh bodies. The changes are summarized as follows:

1. In the input XML, each entry in the `Mesh` block is a `MeshBody`
2. As was previously the case, `MeshBody` contains `MeshLevel`s . No changes to this.
3. `SolverBase::m_targetRegions` is no longer directly used. Instead, a `map<string,array1d<string>> m_meshTargets` is generated from the input of targetRegions and this is used to iterate over the targets.
4. If multiple `MeshBody` are defined, the specification of `targetRegions` must include the specification of what `MeshBody` it refers to. For instance, `meshBody1/reservoir1, meshBody2/reservoir2`. If there is only a single `MeshBody` defined, then no specification is required.
5. Constitutive relations are no longer required to be specified for each `PhysicsSolver`. These are implicitly set by the physics solver. For instance, `SinglePhaseFVM` no longer requires specification of `fluidNames` or `solidNames`, etc. As a result, only one sub-type of constitutive model may be defined in an `ElementRegion`. i.e. one `SolidBase`, one `SingleFluidBase`, etc.
6. Stencil objects no longer require specification of `targetRegions`, or `fieldName`, or `coefficientName`. These are now set by the physics solvers that refer to those stencils. This needs modification in the future for different physics packages referring to the same stencil.
7. As a side note, we found a bug when passing a reference to a `const char *` into the `KernelFactory`. This results in the creation of a temporary `string`, which is a reference in the `KernelFactory`, and passed to the `Kernel` where it is a reference. If this temporary is destroyed prior to launching the constructor to the `Kernel` then we have a reference to a destroyed object. As a solution we no longer pass references to `string` or any type of `View`.

Extension of #1679 

integrated test branch https://github.com/GEOSX/integratedTests/pull/187


src/coreComponents/mainInterface/ProblemManager.cpp
```diff
@@ -621,55 +646,57 @@ map< std::pair< string, string >, localIndex > ProblemManager::calculateRegionQu
       for( localIndex a = 0; a < meshBodies.getSubGroups().size(); ++a )
```
CusiniM, line 646: So, for now, the solver defines the discretization type and then applies it to all meshBodies? I wonder if at some point the same solver could work on different mesh bodies using different discretizations.

src/coreComponents/mainInterface/ProblemManager.cpp
```diff
@@ -502,16 +522,18 @@ void ProblemManager::generateMesh()
 
   MeshManager & meshManager = this->getGroup< MeshManager >( groupKeys.meshManager );
   meshManager.generateMeshes( domain );
-  Group & cellBlockManager = domain.getGroup( keys::cellManager );
 
   Group & meshBodies = domain.getMeshBodies();
 
   for( localIndex a = 0; a < meshBodies.numSubGroups(); ++a )
```
CusiniM, line 528: you could use `forMeshBodies` here since you have created it.

CusiniM (commented):
are there more solvers that need to be changed to allow for multiple meshBodies? Will the dofManager know what to do?

src/coreComponents/mainInterface/ProblemManager.cpp
```diff
@@ -411,10 +411,30 @@ void ProblemManager::parseXMLDocument( xmlWrapper::xmlDocument const & xmlDocume
...
+      ElementRegionManager & elementManager = meshBody.getMeshLevel( 0 ).getElemManager();
+
+      xmlWrapper::xmlNode elementRegionsNode = xmlProblemNode.child( elementManager.getName().c_str());
+      for( xmlWrapper::xmlNode regionNode : elementRegionsNode.children() )
+      {
+        string const regionName = regionNode.attribute( "name" ).value();
+        string const regionMeshBodyName = regionNode.attribute( "meshBody" ).value();
+        string const cellBlocks = regionNode.attribute( "cellBlocks" ).value();
+
+        if( regionMeshBodyName==meshBodyName )
```
povolny1, line 428: Element regions will be assigned to a mesh body iff this is true. For users using a single mesh body this may feel unnecessary when configuring their input files - perhaps when there's a single mesh body, all element regions are added to it by default?

src/coreComponents/mesh/MeshBody.hpp
```diff
@@ -55,21 +55,54 @@ class MeshBody : public dataRepository::Group
...
+  void forMeshLevels( FUNCTION && function ) const
+  {
+    m_meshLevels.forSubGroups< MeshLevel >( std::forward< FUNCTION >( function ) );
+  }
+
+  template< typename FUNCTION >
+  void forMeshLevels( FUNCTION && function )
+  {
+    m_meshLevels.forSubGroups< MeshLevel >( std::forward< FUNCTION >( function ) );
+  }
```
rrsettgast, line 105: dox

src/coreComponents/mesh/MeshBody.hpp
```diff
@@ -90,23 +123,26 @@ class MeshBody : public dataRepository::Group
...
    * @brief Group keys
    */
   struct groupStructKeys
-  {} groupKeys; ///< groupKeys
+  {
+    static constexpr char const * meshLevelsString() {return "meshLevels"; }
+  } groupKeys; ///< groupKeys
 
 private:
+  Group & m_meshLevels;
```
rrsettgast, line 137: should be just a group

src/coreComponents/physicsSolvers/SolverBase.hpp
```diff
@@ -571,78 +575,99 @@ class SolverBase : public ExecutableGroup
...
+//    mesh.getElemManager().
+//      template forElementSubRegionsComplete< SUBREGIONTYPE, SUBREGIONTYPES... >( targetRegionNames(), std::forward< LAMBDA >( lambda ) );
+//  }
+//
+//  template< typename SUBREGIONTYPE = ElementSubRegionBase, typename ... SUBREGIONTYPES, typename LAMBDA >
+//  void forTargetSubRegionsComplete( MeshLevel & mesh, LAMBDA && lambda ) const
+//  {
+//    mesh.getElemManager().
+//      template forElementSubRegionsComplete< SUBREGIONTYPE, SUBREGIONTYPES... >( targetRegionNames(), std::forward< LAMBDA >( lambda ) );
+//  }
```
rrsettgast, line 671:
```suggestion
```

src/coreComponents/physicsSolvers/SolverBase.hpp
```diff
@@ -654,43 +679,23 @@ class SolverBase : public ExecutableGroup
                                  real64 const oldNewtonNorm,
                                  real64 const weakestTol );
 
+  template< typename CONSTITUTIVE_BASE_TYPE >
+  static string getConstitutiveName( ElementSubRegionBase const & );
```
rrsettgast, line 683: dox

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -1462,81 +1541,95 @@ void CompositionalMultiphaseBase::implicitStepComplete( real64 const & time,
...
+        {
+          compDens[ei][ic] += dCompDens[ei][ic];
+        }
+      } );
 
-    // Step 3: save the converged solid state (porosity, solid internal energy, etc)
-    CoupledSolidBase const & porousMaterial = getConstitutiveModel< CoupledSolidBase >( subRegion, m_solidModelNames[targetIndex] );
-    porousMaterial.saveConvergedState();
+      string const & solidName = subRegion.getReference< string >( viewKeyStruct::solidNamesString() );
+      CoupledSolidBase const & porousMaterial = getConstitutiveModel< CoupledSolidBase >( subRegion, solidName );
```
rrsettgast, line 1578: We should make a convenience function for this pattern. So we replace 
```
string const & solidName = subRegion.getReference< string >( viewKeyStruct::solidNamesString() );
CoupledSolidBase const & porousMaterial = getConstitutiveModel< CoupledSolidBase >( subRegion, solidName );
```

with something like:

```
CoupledSolidBase const & porousMaterial = getConstitutiveModel< CoupledSolidBase >( viewKeyStruct::solidNamesString(), subRegion );
```

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.hpp
```diff
@@ -240,7 +234,7 @@ class CompositionalMultiphaseBase : public FlowSolverBase
    * from prescribed intermediate values (i.e. global densities from global fractions)
    * and any applicable hydrostatic equilibration of the domain
    */
-  void initializeFluidState( MeshLevel & mesh );
+  void initializeFluidState( MeshLevel & mesh, arrayView1d< string const > const & regionNames );
```
rrsettgast, line 237: dox

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.hpp
```diff
@@ -251,7 +245,7 @@ class CompositionalMultiphaseBase : public FlowSolverBase
    * @brief Backup current values of all constitutive fields that participate in the accumulation term
    * @param domain the domain containing the mesh and fields
    */
-  void backupFields( MeshLevel & mesh ) const;
+  void backupFields( MeshLevel & mesh, arrayView1d< string const > const & regionNames ) const;
```
rrsettgast, line 248: dox

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.hpp
```diff
@@ -315,23 +309,13 @@ class CompositionalMultiphaseBase : public FlowSolverBase
...
-   * @param[in] cm reference to the global constitutive model manager
-   */
-  void validateAquiferBC( constitutive::ConstitutiveManager const & cm ) const;
 
   /**
    * @brief Initialize the aquifer boundary condition (gravity vector, water phase index)
    * @param[in] cm reference to the global constitutive model manager
    */
-  void initializeAquiferBC( constitutive::ConstitutiveManager const & cm ) const;
+  void initializeAquiferBC( Group const & meshBodies, constitutive::ConstitutiveManager const & cm ) const;
```
rrsettgast, line 317: dox

src/coreComponents/physicsSolvers/wavePropagation/WaveSolverBase.hpp
```diff
@@ -81,7 +81,7 @@ class WaveSolverBase : public SolverBase
    * corresponding elements nodes.
    * @param mesh mesh of the computational domain
    */
-  virtual void precomputeSourceAndReceiverTerm( MeshLevel & mesh ) = 0;
+  virtual void precomputeSourceAndReceiverTerm( MeshLevel & mesh, arrayView1d< string const > const & regionNames ) = 0;
```
rrsettgast: dox

src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEM.hpp
```diff
@@ -116,7 +116,7 @@ class AcousticWaveEquationSEM : public WaveSolverBase
    * corresponding elements nodes.
    * @param mesh mesh of the computational domain
    */
-  virtual void precomputeSourceAndReceiverTerm( MeshLevel & mesh ) override;
+  virtual void precomputeSourceAndReceiverTerm( MeshLevel & mesh, arrayView1d< string const > const & regionNames ) override;
```
rrsettgast: dox

src/coreComponents/physicsSolvers/surfaceGeneration/SurfaceGenerator.hpp
```diff
@@ -529,10 +529,10 @@ class SurfaceGenerator : public SolverBase
   /// choice of failure criterion
   integer m_failCriterion=1;
 
-  // solid solver name
-  array1d< string > m_solidMaterialNames;
+//  // solid solver name
+//  array1d< string > m_solidMaterialNames;
```
rrsettgast:
```suggestion
```

src/coreComponents/physicsSolvers/surfaceGeneration/SurfaceGenerator.cpp
```diff
@@ -179,9 +179,9 @@ SurfaceGenerator::SurfaceGenerator( const string & name,
 {
   this->registerWrapper( viewKeyStruct::failCriterionString(), &this->m_failCriterion );
 
-  registerWrapper( viewKeyStruct::solidMaterialNameString(), &m_solidMaterialNames ).
-    setInputFlag( InputFlags::REQUIRED ).
-    setDescription( "Name of the solid material used in solid mechanic solver" );
+//  registerWrapper( viewKeyStruct::solidMaterialNameString(), &m_solidMaterialNames ).
+//    setInputFlag( InputFlags::REQUIRED ).
+//    setDescription( "Name of the solid material used in solid mechanic solver" );
```
rrsettgast:
```suggestion
```

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.hpp
```diff
@@ -251,7 +255,7 @@ class SolidMechanicsLagrangianFEM : public SolverBase
     dataRepository::ViewKey timeIntegrationOption = { timeIntegrationOptionString() };
   } solidMechanicsViewKeys;
 
-  arrayView1d< string const > solidMaterialNames() const { return m_solidMaterialNames; }
+//  arrayView1d< string const > solidMaterialNames() const { return m_solidMaterialNames; }
```
rrsettgast:
```suggestion
```

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.hpp
```diff
@@ -289,16 +295,19 @@ class SolidMechanicsLagrangianFEM : public SolverBase
   real64 m_maxForce = 0.0;
   integer m_maxNumResolves;
   integer m_strainTheory;
-  array1d< string > m_solidMaterialNames;
+//  array1d< string > m_solidMaterialNames;
```
rrsettgast:
```suggestion
```

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.hpp
```diff
@@ -289,16 +295,19 @@ class SolidMechanicsLagrangianFEM : public SolverBase
...
   string m_contactRelationName;
-  SortedArray< localIndex > m_sendOrReceiveNodes;
-  SortedArray< localIndex > m_nonSendOrReceiveNodes;
-  SortedArray< localIndex > m_targetNodes;
+//  SortedArray< localIndex > m_sendOrReceiveNodes;
+//  SortedArray< localIndex > m_nonSendOrReceiveNodes;
+//  SortedArray< localIndex > m_targetNodes;
```
rrsettgast:
```suggestion
```

src/coreComponents/constitutive/fluid/SingleFluidBase.hpp
```diff
@@ -182,16 +182,13 @@ class SingleFluidBase : public ConstitutiveBase
   arrayView2d< real64 > dViscosity_dPressure() { return m_dViscosity_dPressure; }
   arrayView2d< real64 const > dViscosity_dPressure() const { return m_dViscosity_dPressure; }
 
-  real64 defaultDensity() const { return m_defaultDensity; }
-  real64 defaultViscosity() const { return m_defaultViscosity; }
+  virtual real64 defaultDensity() const { return 1.0; }
+  virtual real64 defaultViscosity() const { return 1.0; }
```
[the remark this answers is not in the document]
CusiniM: I have a feeling that it's not going to compile like this. I think those functions cannot be pure virtual coz the `SlurryFluidBase` does not implement them.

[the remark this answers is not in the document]
rrsettgast: yeah...well..it shouldn't have a default of 1.0

src/coreComponents/finiteVolume/TwoPointFluxApproximation.cpp
```diff
@@ -215,10 +215,14 @@ void TwoPointFluxApproximation::addToFractureStencil( MeshLevel & mesh,
...
-                                                                 m_targetRegions,
-                                                                 m_coefficientModelNames );
+  // TODO: can we look this up better?
+  string const & meshBodyName = mesh.getParent().getParent().getName();
+  string const coeffModelNames = m_coefficientModelNames.at( meshBodyName )[0];
+
+//  ElementRegionManager::ElementViewAccessor< arrayView3d< real64 const > > const coefficient =
+//    elemManager.constructMaterialArrayViewAccessor< real64, 3 >( m_coeffName,
+//                                                                 targetRegions,R
+//                                                                 coeffModelNames );
```
rrsettgast, line 225:
```suggestion
```

src/coreComponents/physicsSolvers/SolverBase.hpp
rrsettgast, line 671:
```suggestion
```

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
```diff
@@ -258,172 +298,186 @@ void compareMulticomponentModels( MODEL1_TYPE const & lhs, MODEL2_TYPE const & r
...
+//         compareMultiphaseModels( wellFluid, resFluid );
+//         compareMulticomponentModels( wellFluid, resFluid );
+
+//         RelativePermeabilityBase const & resRelPerm = cm.getConstitutiveRelation< RelativePermeabilityBase >(
+// flowRelPermModels[resTargetIndex] );
+//         compareMultiphaseModels( wellRelPerm, resRelPerm );
+//       }
+//     }
+//   } );
+// }
```
rrsettgast:
```suggestion
```

src/coreComponents/constitutive/fluid/SingleFluidBase.hpp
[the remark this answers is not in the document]
CusiniM: we should not have a public accessors for these IMO.

src/coreComponents/dataRepository/Group.hpp
```diff
@@ -492,6 +492,47 @@ class Group
...
+                                                          [&]( auto & castedSubGroup )
+      {
+        lambda( counter, castedSubGroup );
+      } );
+      ++counter;
+    }
+  }
+
+  /**
+   * @copydoc forSubGroups(LAMBDA &&)
```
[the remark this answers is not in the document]
CusiniM: I think it is wrong.

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoromechanicsEFEMKernel.hpp
```diff
@@ -95,7 +95,7 @@ class SinglePhase :
                CRSMatrixView< real64, globalIndex const > const & inputMatrix,
                arrayView1d< real64 > const & inputRhs,
                real64 const (&inputGravityVector)[3],
-               arrayView1d< string const > const & fluidModelNames ):
+               string const & fluidModelKey ):
```
CusiniM, line 98: I thought those had to be passed by value and not by reference.

src/coreComponents/dataRepository/xmlWrapper.hpp
```diff
@@ -204,7 +208,7 @@ readAttributeAsType( T &, string const &, xmlNode const &, U const & )
  * @param[in] name       the name of the xml attribute to process
  * @param[in] targetNode the xml node that should contain the attribute
  * @param[in] defVal     default value of @p rval (or entries of @p rval, if it is an array)
- * @return boolean value indicating whether the value was successfully read from XML.
+ * @return true
```
CusiniM, line 211: Why this change to the doc? The previous one seemed correct to me.

rrsettgast: It is always true for this SFINAE implementation.

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoromechanicsEFEMKernel.hpp
rrsettgast: They do!!

src/coreComponents/constitutive/fluid/SingleFluidBase.cpp
```diff
@@ -31,14 +31,6 @@ namespace constitutive
...
   : ConstitutiveBase( name, parent )
 {
-  registerWrapper( viewKeyStruct::defaultDensityString(), &m_defaultDensity ).
-    setInputFlag( InputFlags::REQUIRED ).
-    setDescription( "Default value for density." );
-
-  registerWrapper( viewKeyStruct::defaultViscosityString(), &m_defaultViscosity ).
-    setInputFlag( InputFlags::REQUIRED ).
-    setDescription( "Default value for viscosity." );
-
```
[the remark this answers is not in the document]
rrsettgast: When a new cell is added, and the arrays are resized, what is the values are the new cell initialized to?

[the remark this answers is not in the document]
CusiniM: - the default value can be set inside the constitutive model and there should be no need to access it outside. right now these are used in `SinglePhaseBase` in a couple of places but it's unclear to me why they are needed.
 - the `ProppantSlurryFluid` has hardcoded values for a few parameters and does not really have just a density and a viscosity default.  

However, density and viscosity are not primary variables so I don't see why they cannot be computed for new cells. I think that, when creating a new element, we should only assign values to primary variables and let the model compute the others.

src/coreComponents/physicsSolvers/fluidFlow/StencilAccessors.hpp
```diff
@@ -62,55 +63,65 @@ class StencilAccessors
...
+      forEachArgInTuple( std::tuple< TRAITS ... >{}, [&]( auto t, auto idx )
+      {
+        GEOSX_UNUSED_VAR( t );
+        using TRAIT = TYPEOFREF( t );
+
+        auto & acc = std::get< idx() >( m_accessors );
+        acc = elemManager.constructExtrinsicAccessor< TRAIT >();
+        acc.setName( solverName + "/accessors/" + TRAIT::key() );
+      } );
+    }
```
francoishamon, line 80: What is the role of the `fill` argument that is passed here?

src/coreComponents/constitutive/fluid/SingleFluidBase.cpp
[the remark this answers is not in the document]
rrsettgast: So when we create a new fracture element, how do we determine the state of the fluid in that new element?

src/coreComponents/physicsSolvers/multiphysics/LagrangianContactSolver.cpp
```diff
@@ -97,86 +97,109 @@ LagrangianContactSolver::LagrangianContactSolver( const string & name,
...
 
+void LagrangianContactSolver::setConstitutiveNames( ElementSubRegionBase & subRegion ) const
+{
+  subRegion.registerWrapper< string >( viewKeyStruct::contactRelationNameString() ).
+    setPlotLevel( PlotLevel::NOPLOT ).
+    setRestartFlags( RestartFlags::NO_WRITE ).
+    setSizedFromParent( 0 );
+
+  string & contactRelationName = subRegion.getReference< string >( viewKeyStruct::contactRelationNameString() );
+//      solidName = getConstitutiveName< CoupledSolidBase >( subRegion );
```
francoishamon:
```suggestion
```

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoromechanicsKernel.hpp
```diff
@@ -89,15 +89,15 @@ class Multiphase :
...
               globalIndex const rankOffset,
               real64 const (&inputGravityVector)[3],
               localIndex const numComponents,
               localIndex const numPhases,
-              arrayView1d< string const > const fluidModelNames,
-              CRSMatrixView< real64, globalIndex const > const & inputMatrix,
-              arrayView1d< real64 > const & inputRhs ):
+              string const fluidModelKey,
+              CRSMatrixView< real64, globalIndex const > const inputMatrix,
+              arrayView1d< real64 > const inputRhs ):
```
francoishamon, line 100: I have two questions:
- What was wrong with passing the `inputMatrix` and `inputRhs` by reference?
- After this PR, should we start removing references where they are still present (for instance, in `CompositionalMultiphaseBaseKernels` and possibly in the wells?

src/coreComponents/physicsSolvers/simplePDE/PhaseFieldDamageFEM.hpp
```diff
@@ -131,7 +131,6 @@ class PhaseFieldDamageFEM : public SolverBase
 
   struct viewKeyStruct : public SolverBase::viewKeyStruct
   {
-//    static constexpr auto coeffFieldName = "coeffFieldName";
     static constexpr char const * coeffNameString() { return "coeffField"; }
     static constexpr char const * localDissipationOptionString() { return "localDissipation"; }
     static constexpr char const * solidModelNamesString() { return "solidMaterialNames"; }
```
francoishamon: Maybe remove?
```suggestion
```

francoishamon (commented):
Work in progress in the review. Looks great so far

src/coreComponents/codingUtilities/Utilities.hpp
```diff
@@ -225,6 +225,9 @@ void applyChainRuleInPlace( integer const N,
   copy( N, work, df_dxy );
 }
 
+template< typename T >
+struct typeTag {};
+
```
francoishamon, line 230: Just checking that this is used somewhere?

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.hpp
```diff
@@ -241,6 +240,11 @@ class SolidMechanicsLagrangianFEM : public SolverBase
     static constexpr char const * elemsAttachedToSendOrReceiveNodesString() { return "elemsAttachedToSendOrReceiveNodes"; }
     static constexpr char const * elemsNotAttachedToSendOrReceiveNodesString() { return "elemsNotAttachedToSendOrReceiveNodes"; }
 
+    static constexpr char const * sendOrRecieveNodesString() { return "sendOrReceiveNodes";}
```
francoishamon:
```suggestion
    static constexpr char const * sendOrReceiveNodesString() { return "sendOrReceiveNodes";}
```

src/coreComponents/physicsSolvers/surfaceGeneration/SurfaceGenerator.hpp
```diff
@@ -529,10 +529,7 @@ class SurfaceGenerator : public SolverBase
   /// choice of failure criterion
   integer m_failCriterion=1;
 
-  // solid solver name
-  array1d< string > m_solidMaterialNames;
-
-  localIndex m_solidMaterialFullIndex;
+  array1d< localIndex > m_solidMaterialFullIndex;
```
francoishamon: There is a `solidMaterialNamesString` above in this file that may need to be removed as well, I think

francoishamon (approved):
Looks great, thanks for all this work! 

If that is ok, I am just going to quickly check that everything still works with the combination (imported mesh + wells), because we do not have a very good integrated tests coverage for this.

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoromechanicsKernel.hpp
klevzoff: In this particular case, probably nothing wrong. In fact, I'd say the change to kernel constructor signature is not that important, it's just there to match the change to `MultiphaseKernelFactory` alias below. @rrsettgast found a subtle bug: instance of a factory class keeps the values given to it in a tuple for some amount of time before the instance of a kernel is created from those values; if you happen to pass a temporary to the factory (such as a `string` created from a `const char *`, or a `CRSMatrixView` created from a `CRSMatrix` on the fly), it'll store a reference to that temporary which becomes dangling as soon as the factory constructor is finished executing, so when we later call `.createKernel()` it passes dangling references to the kernel constructor. This hasn't happened specifically to `inputMatrix` and `inputRhs` so far because those temporary view object generally have lifetimes for the entire duration of assembly, but as a safety policy we shouldn't be passing references to factories unless they refer to really persistent objects (such as various mesh object managers or other high-level components).

src/coreComponents/constitutive/fluid/SingleFluidBase.cpp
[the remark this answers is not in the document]
CusiniM: Well, I think we should determine whatever the primary variables are and then compute the remaining ones based on what we have done for those. Basically I think we should assign a pressure (and compositions if multiphase) and then compute all secondary variables based on the constitutive models used. Now, which pressure to choose is a good question. I am guessing the same as the trailing element one?

src/coreComponents/physicsSolvers/fluidFlow/StencilAccessors.hpp
CusiniM: Well, when constructing the material version of this `stencilAccessor` the construction of the accessor has to occur differently (need to call `constructExtrinsicMaterialAccessor`) and it's defined in the constructor of the `StencilMaterialAccessor`. However, the constructor of this base class still gets called so I had to add this boolean to avoid building the `ViewAccessor`.

src/coreComponents/finiteVolume/FluxApproximationBase.hpp
```diff
@@ -245,10 +248,10 @@ class FluxApproximationBase : public dataRepository::Group
   string m_coeffName;
 
   /// names of coefficient models to build the stencil for
-  string_array m_coefficientModelNames;
+  map< string, array1d< string > > m_coefficientModelNames; // TODO: remove
```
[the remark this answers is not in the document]
klevzoff: I think the only part of this class that still uses both `coefficientName` and `coefficientModelNames` is the boundary stencil, which has not been updated to use the new permeability model and still assumes constant values. If we still think that having face-based Dirichlet boundary conditions for flow is valuable, it would be great to have a separate PR to update that kernel (it's going to be somewhat tricky because the kernel already uses fluid dispatch to compute fluid properties at the face, adding permeability dispatch will make it quite nasty).

src/coreComponents/physicsSolvers/fluidFlow/StencilAccessors.hpp
klevzoff: Sorry for engaging to late on this... not a super important thing, but I think it would be better to have a `protected` default constructor (which takes no arguments and does not populate accessors) and call that from the derived class, and not use `fill` flags.

src/coreComponents/finiteVolume/FluxApproximationBase.hpp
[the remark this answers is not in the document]
francoishamon: Ok thanks, I see. To me having the faced-based Dirichlet boundary conditions is very valuable, so I agree with this plan of updating the face-based BC kernel in a separate PR.

francoishamon:
> If that is ok, I am just going to quickly check that everything still works with the combination (imported mesh + wells), because we do not have a very good integrated tests coverage for this.

Just to confirm, I tried the `deadOilEgg_benchmark.xml` (which has wells and imported mesh) on this branch, and it works fine. I think this is the only test for this combination, and it is not checked automatically yet. Thanks for all this work

src/coreComponents/dataRepository/Group.hpp
```diff
@@ -509,11 +550,14 @@ class Group
     localIndex counter = 0;
     for( auto const & subgroup : subGroupKeys )
     {
-      applyLambdaToContainer< GROUPTYPE, GROUPTYPES... >( getGroup( subgroup ), [&]( auto & castedSubGroup )
+      if( hasGroup( subgroup ) )
```
TotoGaz, line 553: I'm must confess I'm a bit confused by this new check.
We use to have a `Group` (and therefore the whole `GEOSX`) failing in case a `Group` instance "_was not there_" (the call to `getGroup` would `throw`/`exit`).

Now we silently accept the sub-group not being there.
There surely is a reason, but wouldn't this open more space for vicious bugs?

src/coreComponents/dataRepository/Group.cpp
```diff
@@ -268,6 +268,18 @@ void Group::initializationOrder( string_array & order )
   }
 }
 
+void Group::initialize_postMeshGeneration()
+{
+  string_array initOrder;
```
TotoGaz, line 273: So what's our policy about `string_array`? Should we use it or `array1d< string >`?

src/coreComponents/finiteVolume/TwoPointFluxApproximation.cpp
```diff
@@ -862,16 +861,21 @@ void TwoPointFluxApproximation::computeBoundaryStencil( MeshLevel & mesh,
   ElementRegionManager::ElementViewAccessor< arrayView1d< integer const > > const elemGhostRank =
     elemManager.constructArrayViewAccessor< integer, 1 >( ObjectManagerBase::viewKeyStruct::ghostRankString() );
 
+  // TODO: can we look this up better?
+  string const & meshBodyName = mesh.getParent().getParent().getName();
+  arrayView1d< string const > const targetRegions = m_targetRegions.at( meshBodyName );
+  string const coeffModelNames = m_coefficientModelNames.at( meshBodyName )[0];
```
TotoGaz, line 867: Why do you only consider the first element of the list?

src/coreComponents/linearAlgebra/DofManager.cpp
```diff
@@ -140,59 +138,119 @@ DofManager::Location DofManager::location( string const & fieldName ) const
...
+}
+
 namespace
 {
 
-std::vector< string > processFieldRegionList( MeshLevel const & mesh,
-                                              std::vector< string > const & inputList )
+std::vector< string >
+processFieldRegionList( MeshLevel const & mesh,
+                        std::vector< string > inputList )
```
TotoGaz, line 251: is `inputList` a `regionNames`?

src/coreComponents/linearAlgebra/DofManager.cpp
```diff
@@ -525,226 +666,83 @@ void DofManager::setSparsityPatternFromStencil( SparsityPatternView< globalIndex
...
-    typename StenciType::IndexContainerViewConstType const & seri = stencil.getElementRegionIndices();
-    typename StenciType::IndexContainerViewConstType const & sesri = stencil.getElementSubRegionIndices();
-    typename StenciType::IndexContainerViewConstType const & sei = stencil.getElementIndices();
+    ElementRegionManager::ElementViewAccessor< arrayView1d< globalIndex const > > const dofNumber =
+      mesh.getElemManager().constructArrayViewAccessor< globalIndex, 1 >( field.key );
 
-    rowDofIndices.reserve( maxNumFluxElems );
-    colDofIndices.reserve( maxStencilSize * numComp );
+    array1d< globalIndex > rowDofIndices( numComp );
+    array1d< globalIndex > colDofIndices( numComp );
```
TotoGaz, line 675: Do you need to already set a size here since it will be `reserved` and `resized` again hereafter?

TotoGaz (commented):
First comments.

src/coreComponents/dataRepository/Group.hpp
rrsettgast: oof. I think I put this in as a temporary workaround...and was supposed to remove this. nice catch.

src/coreComponents/dataRepository/Group.cpp
rrsettgast: we have no policy. I typically just use `array1d<string>` nowadays. I think it makes more sense due to the use of `arrayView`.

src/coreComponents/finiteVolume/TwoPointFluxApproximation.cpp
rrsettgast: They are all the same value....in the case of flow, it will be a string named "permeabilityNames", which is actually set in `FlowSolverBase`.  We will need to clean this concept up/out in a later PR.

src/coreComponents/linearAlgebra/DofManager.cpp
```diff
@@ -364,10 +487,28 @@ void DofManager::addCoupling( string const & fieldName,
...
+                              bool symmetric )
+{
+  // Convert input into internal format
+  std::vector< Regions > support;
+  for( auto const & p : regions )
+  {
+    MeshBody const & meshBody = m_domain->getMeshBody( p.first );
+    MeshLevel const & mesh = meshBody.getMeshLevel( 0 ); // TODO by name?
+    std::vector< string > regionNames( p.second.begin(), p.second.end() );
+    support.push_back( { meshBody.getName(), mesh.getName(), std::move( regionNames ) } );
```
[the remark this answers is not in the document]
rrsettgast: @klevzoff ?? I don't know that anything bad can happen by using the initialization list? I will concede the constructor is more obvious, but I don't think either way would save us if parameters are swapped.

src/coreComponents/dataRepository/Group.hpp
```diff
@@ -534,11 +575,14 @@ class Group
     localIndex counter = 0;
     for( auto const & subgroup : subGroupKeys )
     {
-      applyLambdaToContainer< GROUPTYPE, GROUPTYPES... >( getGroup( subgroup ), [&]( auto const & castedSubGroup )
+      if( hasGroup( subgroup ) )
```
klevzoff, line 578: @rrsettgast this one too

src/coreComponents/linearAlgebra/DofManager.cpp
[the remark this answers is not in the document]
klevzoff: Regardless of whether `Regions` has a constructor or not, the two lines do exactly the same thing (call the constructor if one exists, or aggregate-initialize if not). There's also not a way (that I know of) to prevent the user from accidentally mixing up two consecutive parameters of the same type, short of using a separate "strong type" for each parameter, i.e. 
```c++
struct MeshBodyName : public std::string { using std::string::string; };
```
(but note that this can also be done with an aggregate struct without constructors).

The case @TotoGaz mentions where someone changes the layout of `Regions` could indeed be helped by having a constructor, provided that the same person doesn't simultaneously change the order of constructor arguments. However, purely on philosophical grounds, I believe a struct with all public members that does not maintain any invariants (in other words, serves simply to group related data, like a `tuple`) _should not_ have constructors. If we decided `Regions` should have invariants or other useful logic (say, check that given names correspond to an existing mesh body/level), I'd fully agree to adding a constructor, but here it's really used as a more expressive and compile-time-friendly replacement for `tuple`.

src/coreComponents/linearAlgebra/DofManager.cpp
```diff
@@ -364,10 +487,28 @@ void DofManager::addCoupling( string const & fieldName,
...
+                              map< string, array1d< string > > const & regions,
+                              bool symmetric )
+{
+  // Convert input into internal format
+  std::vector< Regions > support;
+  for( auto const & p : regions )
+  {
+    MeshBody const & meshBody = m_domain->getMeshBody( p.first );
+    MeshLevel const & mesh = meshBody.getMeshLevel( 0 ); // TODO by name?
+    std::vector< string > regionNames( p.second.begin(), p.second.end() );
```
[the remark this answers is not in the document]
klevzoff: I don't think I fully understand, could you explain a bit more? Note that I'm converting here from `array1d< string >` to `std::vector< string >` using the latter's range constructor. I don't think I can do more simple?

src/coreComponents/linearAlgebra/DofManager.cpp
klevzoff: Indeed, it's a list of region names. If not empty, the function checks that every name corresponds to an existing region and returns it unchanged. If empty, it builds and returns a list of all region names, which allows the user to skip listing all regions when there's many of them.

src/coreComponents/linearAlgebra/DofManager.cpp
klevzoff: No, doesn't look like it. I'm guessing this carried over accidentally from some refactoring a long time ago. Should be removed.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -130,24 +112,68 @@ void CompositionalMultiphaseBase::registerDataOnMesh( Group & meshBodies )
...
   ConstitutiveManager const & cm = domain.getConstitutiveManager();
 
+  // 0. Find a "reference" fluid model name (at this point, models are already attached to subregions)
+  forMeshTargets( meshBodies, [&]( string const &,
+                                   MeshLevel & mesh,
+                                   arrayView1d< string const > const & regionNames )
+  {
+    mesh.getElemManager().forElementSubRegions( regionNames,
+                                                [&]( localIndex const,
+                                                     ElementSubRegionBase & subRegion )
```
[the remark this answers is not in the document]
klevzoff: I'm not sure if it's useful for subregion loops in physics solvers, but in general when iterating over a set of objects from a given list, it might occasionally be useful to get an index back into the list, such as when there's other data associated with that list.

To reduce boilerplate, I've used the following tactic in a similar situation (it's on a [branch](https://github.com/GEOSX/GEOSX/blob/70df502864c39dc79888493a62be4ef233a32b5f/src/coreComponents/linearAlgebra/multiscale/MeshUtils.hpp#L182-L215) not yet merged): add a type trait (using our `IS_VALID_EXPRESSION` macro) to check whether user-provided function is callable with an index and object, or just an object, and call it accordingly. Not sure how it would affect compile times if added to such a ubiquitous class as `Group` though.

src/coreComponents/finiteVolume/TwoPointFluxApproximation.cpp
TotoGaz: OK, great. Maybe should be have a list of the small pebbles in our shoes that we need to solve after this PR is merged?

src/coreComponents/linearAlgebra/DofManager.cpp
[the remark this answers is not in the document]
TotoGaz: _In that context_, I believe that the difference between a `tuple` and a `struct` is that
- the first relies on explicit _indexing_ to get the proper information
- while the latter relies on explicit "_keys_" (the attribute names).

That's fair and the contracts are clear.
But by using the "_aggregate-initialize_" for `Regions` we now rely on both (adding _implicit_ indexing).
So we have to be careful for both.

_Nobody_ will ever think and see that reordering the parameters (some renaming + sorting in alphabetical order, _e.g._) will break the code in another file.
But _everybody_ knows that changing the signature of a function needs to be  propagated. Tools provided by the IDEs will even do this for you.

To face this issue, I've just discovered that c++20 introduces `designated initializers` (see [here](https://en.cppreference.com/w/cpp/language/aggregate_initialization)):
```c++
struct A { int x; int y; int z; };
A a{.y = 2, .x = 1}; // error; designator order does not match declaration order
A b{.x = 1, .z = 2}; // ok, b.y initialized to 0
```
Is acts like some on-the-fly constructor generation I presume.

It's a matter of risks. No approach will save us for sure. But one is really risky, the other is really safe.

src/coreComponents/linearAlgebra/DofManager.cpp
TotoGaz: OK; then I presume `inputList` should be renamed into `regionNames` 😉

src/coreComponents/linearAlgebra/DofManager.cpp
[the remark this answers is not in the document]
TotoGaz: Oh, sorry, I thought you could simply `std::vector< string > regionNames( p.second )` but I was confused.

Also, is there a reason why we are mixing `std::vector< string >` and `array1d< string >`?

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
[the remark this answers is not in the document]
TotoGaz: If we want to reduce `boilerplate`, could it be possible to have something like
```c++
for( ElementSubRegionBase & subRegion: elemMgr.getElementSubRegions( regionNames ) ) {
   ...
}
```
or if we need an index
```c++
for( int i = 0; i<  elemMgr.numElementSubRegions( regionNames ); ++i ) {
   ElementSubRegionBase & subRegion = elemMgr.getElementSubRegion( i );
   ...
}
```
No (nested) `lambda`, no `traits`, no `callback`. Classical `for` loops.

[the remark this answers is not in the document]
rrsettgast: @klevzoff That is a neat idea!!

@TotoGaz In retrospect, there isn't much benefit to some of the `forBlahBlah` functions. Perhaps we can discuss the merit of them in a dev meeting.

src/coreComponents/linearAlgebra/DofManager.cpp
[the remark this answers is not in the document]
klevzoff: My reason is that `array1d` is missing some features of `vector`... mainly initializer-list construction which is very useful when writing [tests](https://github.com/GEOSX/GEOSX/blob/d34ec3645cbb0e88b1c4ff6436e041c80d45b0c0/src/coreComponents/unitTests/linearAlgebraTests/testDofManager.cpp#L659) and range constructor (granted, I could add those to `Array`... but I don't have time to). But also because `array1d<>` brings too many additional features and associated implementation complexity (e.g. dual memory space semantics) that aren't needed for a simple host-only linear container of strings. This isn't a super strong opinion... I could change internal storage to be `array1d` and add a compatibility API that accepts vector to keep tests simple. It's one way or the other.

On a related side note, if we keep using `LvArray::Array` to hold strings, we should at least use it with a `MallocBuffer` instead of `ChaiBuffer` to reduce unneeded overhead.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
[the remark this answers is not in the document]
TotoGaz: Actually, this _systematic_ `forXXX` pattern is very impactful since it hinders testability and makes abstraction much more complicated (if not impossible).
If you want I can prepare some presentation on what we can do so we can exchange on the subject when I come at LLNL.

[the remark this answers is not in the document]
klevzoff: Ranges seems to be the most composable abstraction these days. With respect to the above two patterns, given a library like range-v3, it would be a matter of writing
```c++
elementManager.subregions() | for_each( [&]( ElementSubRegionBase & subRegion ) { ... } );
```
vs
```c++
elementManager.subregions() | enumerate | for_each( [&]( localIndex i, ElementSubRegionBase & subRegion ) { ... } );
```
(provided we implement the range returned by `elementManager.subregions()`, which shouldn't be too difficult)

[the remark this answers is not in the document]
TotoGaz: Well, as always it's not so simple. 🤷 

Obviously `ranges` look really convenient, and would at least end up to `chained actions` instead of `nested actions`.

My main concern would be: _"What will the effective type returned by `elementManager::subregions` be?"_
Because after that call, we can chain how we want or need to. Would it be a `lambda`, or even standard code...

The effective `range` types can become insane.
We would not like this returned type to reflect the internal implementation too much.
```c++
namespace rv = ranges::views;
auto r = rv::iota(0, 20)
       | rv::filter([](int i) { return i % 2 == 0; })
       | rv::transform([](int i){ return i * i; });
std::cout << abi::__cxa_demangle(typeid(r).name(), nullptr, nullptr, nullptr) << '\n';
```
prints `ranges::transform_view<ranges::filter_view<ranges::iota_view<int, int>, main::{lambda(int)#1}>, main::{lambda(int)#2}>`. What are we concretely going to return in our interfaces? If we want to support multiple implementations (or `mock`), we would need `ranges` that do not depend on the implementation.
I do not know if `ranges` are meant to be returned.

We could try to return the `any_view` range which does some type erasure. (It has a cost, but we'll never face the bottleneck for the kind of problems we are considering.) But, `any_view` is not in `c++20` and is not even proposed in `c++23` 😨 . So what are we going to decide? Are we going to commit to `ranges-v3` if we do not know if we'll be able to eventually replace it by the `std` implementations?

Maybe returning a good old `std::vector` of pointers is the best solution? 🤣 

(Plus I've read that compilation times of ranges are dumb. I've never tested in real projects anyway.)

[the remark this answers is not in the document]
klevzoff: I have more thoughts on the subject, but I think the discussion has outgrown the PR comment :nerd_face: Too bad there is no magic button to create a new issue from a thread and automatically copy all the comments...

Back the the original comment, in short term I think we can consider removing the index parameter if there are no actual usages anywhere. Maybe in a follow-up PR though?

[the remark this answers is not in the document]
TotoGaz: We may want want to revive https://github.com/GEOSX/GEOSX/issues/1089, in which some elements (like `ranges`) were already mentioned?
Not exactly the same context, but I think it's the same idea.

src/coreComponents/finiteVolume/TwoPointFluxApproximation.cpp
rrsettgast: @francoishamon @CusiniM  are aware of this, and I think they have a plan.

src/coreComponents/linearAlgebra/DofManager.cpp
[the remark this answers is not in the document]
rrsettgast: @TotoGaz c++20 it is then!!

src/coreComponents/physicsSolvers/SolverBase.hpp
```diff
@@ -654,42 +623,29 @@ class SolverBase : public ExecutableGroup
...
-   * Looks up each model by name and type in each subregion of target regions.
+   * @brief This function sets constitutive name fields on an
+   *  ElementSubRegionBase, and calls the base function it overrides.
+   * @param subRegion The ElementSubRegionBase that will have constitutive
+   *  names set.
    */
-  template< typename MODEL_TYPE = constitutive::ConstitutiveBase >
-  void validateModelMapping( ElementRegionManager const & elemRegionManager,
-                             arrayView1d< string const > const & modelNames ) const;
+  virtual void setConstitutiveNamesCallSuper( ElementSubRegionBase & subRegion ) const { GEOSX_UNUSED_VAR( subRegion ); }
```
TotoGaz, line 642: I must confess that I do not fully understand what you are trying to achieve here. (I know you talked about it during the dev meeting, but I did not catch it...)
It looks like you are willing to enforce the derived class to call the upper function which does nothing?

TotoGaz (commented):
Second batch of commentsl

src/coreComponents/physicsSolvers/SolverBase.hpp
CusiniM: The SolverBase implementation does nothing (for now) but the physicsSolvers hierarchy is deeper. All, flow solvers, for example, will call `FlowSolverBase` implementaiton of `setConstitutiveNamesCallsuper`.

src/coreComponents/linearAlgebra/DofManager.cpp
```diff
@@ -1292,32 +1270,38 @@ void DofManager::reorderByRank()
     dofOffset += field.numLocalDof;
   }
 
-  std::map< string, string_array > fieldToSync;
+  std::map< std::pair< string, string >, std::map< string, string_array > > fieldsToSync;
```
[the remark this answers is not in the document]
rrsettgast: what...it isn't obvious??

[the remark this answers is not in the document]
TotoGaz: I'm getting old 👴

src/coreComponents/physicsSolvers/SolverBase.hpp
TotoGaz: What do you expect to do (in the future) with this architecture then?

src/coreComponents/mainInterface/ProblemManager.cpp
```diff
@@ -501,16 +529,17 @@ void ProblemManager::generateMesh()
...
 
   for( localIndex a = 0; a < meshBodies.numSubGroups(); ++a )
   {
     MeshBody & meshBody = meshBodies.getGroup< MeshBody >( a );
-    for( localIndex b = 0; b < meshBody.numSubGroups(); ++b )
+    Group & meshLevels = meshBody.getMeshLevels();
+    for( localIndex b = 0; b < meshLevels.numSubGroups(); ++b )
     {
-      MeshLevel & meshLevel = meshBody.getGroup< MeshLevel >( b );
+      CellBlockManagerABC & cellBlockManager = meshBody.getGroup< CellBlockManagerABC >( keys::cellManager );
```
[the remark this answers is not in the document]
rrsettgast: because it was merged into the wrong place.

src/coreComponents/physicsSolvers/SolverBase.hpp
CusiniM: Every time there is a constitutive model with a base time common to all (or many) solvers can be set in the base class. This is already used by the Flow Solvers in which, for example, the solid model is set in `FlowSolveBase` but the fluid model is specific to the solve type and it's set  in the private function.  
I am not sure whether we will ever have something set in `SolverBase` or not. If that's not the case, eventually, we could just make this a pure virtual.

src/coreComponents/physicsSolvers/fluidFlow/wells/WellSolverBase.cpp
```diff
@@ -75,42 +71,68 @@ WellSolverBase::~WellSolverBase() = default;
...
+        setRestartFlags( RestartFlags::NO_WRITE ).
+        setSizedFromParent( 0 );
+
+      PerforationData * const perforationData = subRegion.getPerforationData();
+      perforationData->registerExtrinsicData< extrinsicMeshData::well::gravityCoefficient >( getName() );
+    } );
   } );
 }
 
+void WellSolverBase::setConstitutiveNamesCallSuper( ElementSubRegionBase & subRegion ) const
```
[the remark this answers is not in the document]
CusiniM: I don't think it's intended.

src/coreComponents/physicsSolvers/surfaceGeneration/SurfaceGenerator.cpp
```diff
@@ -3662,11 +3663,15 @@ int SurfaceGenerator::calculateElementForcesOnEdge( DomainPartition const & doma
...
-  ConstitutiveManager const & cm = domain.getConstitutiveManager();
-  ConstitutiveBase const & solid = cm.getConstitutiveRelation< ConstitutiveBase >( m_solidMaterialNames[0] );
-  m_solidMaterialFullIndex = solid.getIndexInParent();
-
   ConstitutiveManager const & constitutiveManager = domain.getConstitutiveManager();
+  m_solidMaterialFullIndex.resize( elementManager.numRegions() );
+  elementManager.forElementRegionsComplete< CellElementRegion >( [&]( localIndex regionIndex,
+                                                                      CellElementRegion const & region )
+  {
+    string const & solidMaterialName = region.getSubRegion( 0 ).getReference< string >( viewKeyStruct::solidMaterialNameString() );
```
TotoGaz: I'm a bit surprised by this `getSubRegion( 0 )`.
I suppose that we take the first one because they're all the same across the `sub regions`.
So why don't we store this business information in the `region`?

TotoGaz (approved):
I've reviewed the c++ files only.

src/coreComponents/linearAlgebra/DofManager.cpp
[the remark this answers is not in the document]
rrsettgast: I added a description of what the key and values are.

src/coreComponents/mesh/generators/InternalMeshGenerator.cpp
```diff
@@ -567,13 +567,11 @@ void InternalMeshGenerator::generateMesh( DomainPartition & domain )
...
 
-  Group & meshBodies = domain.getGroup( string( "MeshBodies" ) );
-  MeshBody & meshBody = meshBodies.registerGroup< MeshBody >( this->getName() );
-  meshBody.registerGroup< MeshLevel >( string( "Level0" ) );
+  MeshBody & meshBody = domain.getMeshBody( this->getName() );
 
   // Make sure that the node manager fields are initialized
 
-  CellBlockManager & cellBlockManager = domain.registerGroup< CellBlockManager >( keys::cellManager );
+  CellBlockManager & cellBlockManager = meshBody.getGroup< CellBlockManager >( keys::cellManager );
```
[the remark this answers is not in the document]
rrsettgast: @TotoGaz I didn't find the reasons above. Can you repeat?

src/coreComponents/mainInterface/ProblemManager.cpp
```diff
@@ -605,18 +638,18 @@ void ProblemManager::applyNumericalMethods()
   ConstitutiveManager & constitutiveManager = domain.getGroup< ConstitutiveManager >( keys::ConstitutiveManager );
   Group & meshBodies = domain.getMeshBodies();
 
-  map< std::pair< string, string >, localIndex > const regionQuadrature = calculateRegionQuadrature( meshBodies );
+  map< std::tuple< string, string, string >, localIndex > const regionQuadrature = calculateRegionQuadrature( meshBodies );
```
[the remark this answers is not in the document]
rrsettgast: Point taken, but I don't know that this warrants a class definition. I think it is pretty obvious what the 3 strings represent here?

src/coreComponents/mesh/generators/InternalMeshGenerator.cpp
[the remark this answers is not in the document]
TotoGaz: Oops, github fooled me. It's _"below"_ -> https://github.com/GEOSX/GEOSX/pull/1682#discussion_r803027854.

src/coreComponents/mainInterface/ProblemManager.cpp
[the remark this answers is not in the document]
TotoGaz: > I think it is pretty obvious what the 3 strings represent here?

I'm afraid I have to respectfully disagree.

Plus a nested `struct` could do the job (like `DofManager::Regions`) if possible.

src/coreComponents/mesh/ElementRegionBase.hpp
```diff
@@ -68,6 +68,19 @@ class ElementRegionBase : public ObjectManagerBase
...
 
+
+  /**
+   * @brief verify that the meshBody name specified exists in the meshBodies group. If there is only one meshBody
+   * it returns the name of the only existing mesh body.
+   *
+   * @param[in] meshBodies the meshBodies group.
+   * @param[in] meshBodyBlockName name of the meshbody to be verified
+   * @return the name of the only mesh body present if there is only one
+   * and the name specified by @p  meshBodyName if the meshBody is found.
```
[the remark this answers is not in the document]
rrsettgast: This is just testing to see that the `MeshBody` name specified in the `ElementRegion` xml block exists, and returns it if it does...throws an error if not.

src/coreComponents/physicsSolvers/surfaceGeneration/SurfaceGenerator.cpp
rrsettgast: We should be storing a lot of things in Region. Follow up PR.

src/coreComponents/mesh/MeshBody.cpp
```diff
@@ -26,9 +27,10 @@ using namespace dataRepository;
 MeshBody::MeshBody( string const & name,
                     Group * const parent ):
   Group( name, parent ),
+  m_meshLevels( registerGroup( groupStructKeys::meshLevelsString() ) ),
   m_globalLengthScale( 0 )
 {
-  registerWrapper< integer >( viewKeys.meshLevels );
+  registerGroup< CellBlockManager >( keys::cellManager );
```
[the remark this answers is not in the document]
rrsettgast: This isn't clear to me. I interpret the relationship such that every `MeshBody` needs to have a `CellBlockManager` in order to build its base `MeshLevel`. So then it wouldn't be breaking anything if `MeshBody` knows about the existence of `CellBlockManager`.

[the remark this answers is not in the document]
TotoGaz: > `MeshBody` needs to have a `CellBlockManager` [...]

...but `MeshBody` does not have to know which implementation.
Everything that `MeshBody` needs to know is that an implementation of the abstraction `CellBlockManagerABC` will provides the meshing information. No more, no less.

- The implementation `CellBlockManager` provides more capabilities, and so it's too much for `MeshBody`.
- More important, `MeshBody` is not able to (and therefore does not have to) select which `CellBlockManager` should be used behind the scenes.
For instance, in a future refactoring we'll have an optimized `CellBlockManager` dedicated to structured `InternalMeshGenerator` (for wave propagation _e.g._), another for conformal meshes, a third one for `CornerPointGrid`...

In a nutshell, `MeshBody` uses a service (the `abstract base class`) but does not provide the service (the `implementation`).

As such, this patch would still work because we're at the beginning of the "_meshing journey_" and we have one unique implementation `CellBlockManager`.
But it hinders future refactoring actions and we'll eventually have to revert it.

PR #1684, opened 2021-11-11
[Doc] Mandel's poroelastic problem

jhuang2601:
This PR

- [x]  documents the verification of Mandel's 2D consolidation problem with a poroelastic solver;
- [x]  uses TimeHistory functions and provides python script to output and postprocess pore pressure and displacement field;
- [x]  adds `PoroElastic_Mandel_smoke` to the integratedTests and checks the running on Pecan
[GEOSX/integratedTests#173](https://github.com/GEOSX/integratedTests/pull/173)
- [x]  removes redundant kgdViscosityDominated xml files


src/docs/sphinx/advancedExamples/validationStudies/mandel/Example.rst
```diff
@@ -0,0 +1,607 @@
...
+  
+        pplist = np.zeros([len(tl), len(xlist)]) 
+        for i in range(len(xlist)): 
+            ind = indlist[i]
+            pplist[:,i]= pl[:,ind]
+
+        zlist = [] 
+        indlist = [] 
+        for i in range(0,len(zcord_node)):
+            if xcord_node[i]==0.5 and ycord_node[i]==0.0:
```
castelletto1, line 521: We should be a little cautious in using the current input files for Mandel's problem (`PoroElastic_Mandel_*`) as a validation example. We're missing the rigid constraint at the top/bottom surface, namely where the compressive loading is applied, in the problem definition. As such we cannot reproduce Mandel's analytical solution.

For example, the displacement in the `z`-direction (with reference to the sketch figure [here](https://github.com/GEOSX/GEOSX/blob/6dba08a3236d683fd5805c24411bae1e815429e5/src/docs/sphinx/advancedExamples/validationStudies/mandel/Example.rst)) should be a function of `z` and `time` only, with a zero gradient in the `x`-direction. Here, the script plotting the analytical solution extracts the numerical solution at nodes such that:
```python
if xcord_node[i]==0.5 and ycord_node[i]==0.0:
```
If `0.5` is changed to a different value the analytical solution will only be matched towards steady state -- the numerical solution being _behind_ (largest effect at `x = 0.0`) and _ahead_ (largest effect at `x = 1.0`) of the analytical solution for values lower and larger than `0.5` respectively , respectively. Similar comments apply to the displacement in the `x`-direction and `pressure`, both depending on `x` and `time` with a zero gradient in `z`. 

An easy fix would be to prescribe the analytical `z`-displacement at the loaded boundary instead of traction. This would require computing the analytical solution at `z = 1.0` for different time levels. The drawback of this approach is that if a user wants to modify hydromechanical parameters, such prescribed displacement condition needs to be changed accordingly. However, the python script that computes the analytical solution is already available.

If we all agree on the above strategy, we may also want to prescribe the exact initial solution corresponding to the undrained response. The pressure would be constant everywhere in the domain. The `x`-displacement would be linear in `x`, and the `z`-displacement would be linear in `z`.

jhuang2601: @castelletto1 Already apply these comments and please review this PR again.

castelletto1 (approved):
Thank you @jhuang2601, this looks very good to me! I just left a couple of suggestions for the python script used to compute Mandel's analytical solution.

src/docs/sphinx/advancedExamples/validationStudies/mandel/Example.rst
```diff
@@ -0,0 +1,576 @@
...
+            if x1 != None:
+                a = x2
+                root = bisect(f,x1,x2,1)              
+                if root != None:
+                    pass
+                    out.append(root)            
+            else:
+                print ('Done')
+                break     
+        return out
```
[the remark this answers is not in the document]
jhuang2601: Thanks for your suggestion on getting the roots of `alpha_n` with Newton's method.

herve-gross (approved):
Nice example, Jian!

PR #1686, opened 2021-11-14
LAI vector refactor - zero copy assembly

klevzoff:
This PR is a step towards https://github.com/GEOSX/GEOSX/issues/1675
Before addressing matrices, I'm taking a stab at a much easier problem - vectors.

This PR has 3 goals:
1. Remove all element-wise methods from vector classes (add/set/insert/get)
2. Zero-copy assembly interface (LA packages wrap and operate directly on GEOSX-managed memory)
3. Clean up implementation of `HypreVector` by removing `IJVector` object and always operate directly on the `ParVector` object.

The new vector API functions as follows:
* All vector objects contain an `array1d< real64 > m_values;` which replaces `m_localRhs` and `m_localSolution`
* The physics solver calls `rhs.create( localSize, comm );` (now the only creation method) which allocates the array and creates the internal LA object wrapping array's memory in the correct space (i.e. device pointer in case of hypre-gpu, host pointer for all other)
* The physics solver calls `arrayView1d< real64 > const localRhs = rhs.open();` which returns a mutable view that can be used in assembly kernels as before
* The physics solver calls `rhs.close();`, which moves the assembled values to the correct memory space if needed
* The linear solver solves the system, operating directly on the rhs/solution pointers in the appropriate space
* The linear solver calls `solution.touch()` which calls `m_values.registerTouch( space )` and notifies it of value changes
* The physics solver gets a const view to solution values by calling `arrayView1d< real64 const > localSol = solution.values();` and proceeds to apply them as before

This PR passes unit tests and integrated tests, including in hypre-gpu builds (with the usual diffs for int32/int64 types).

I also ran a couple benchmark inputs that use Krylov solvers and AMG (`beamBending_benchmark.xml` and `SSLE-QS-small.xml`) with hypre-gpu build. They seem to run fine without segfaulting (and without any ATS magic involved). This means that at least for rhs/solution vectors hypre does not seem to require UM, though it still needs more verification - I'd welcome more cases that I can run with hypre-gpu, especially using MGR - the ones I tried like SPE10 and Egg all fail similarly with FP exceptions on `develop` and this PR.

I wasn't able to measure any statistically significant speedup/slowdown compared to `develop`, but I don't think much speedup should've been expected - the linear solve time dwarfs everything else, including memory transfers, even on my PCI-E machine.

### DIFF
--- a/src/coreComponents/linearAlgebra/interfaces/hypre/HypreExport.cpp
+++ b/src/coreComponents/linearAlgebra/interfaces/hypre/HypreExport.cpp
@@ -36,10 +36,10 @@ HypreExport::HypreExport( HypreMatrix const & mat,
   : m_targetRank( targetRank )
 {
   // make a sub-communicator for scatter and ensure target rank is mapped to 0 in new comm
-  int const rank = MpiWrapper::commRank( mat.getComm() );
+  int const rank = MpiWrapper::commRank( mat.comm() );
   int const color = ( mat.numLocalRows() > 0 ) ? 0 : MPI_UNDEFINED;
   int const key = ( rank == m_targetRank ) ? 0 : ( rank < m_targetRank ) ? rank + 1 : rank;
-  m_subComm = MpiWrapper::commSplit( mat.getComm(), color, key );
+  m_subComm = MpiWrapper::commSplit( mat.comm(), color, key );
 }
 
 HypreExport::~HypreExport()
@@ -122,7 +122,7 @@ void HypreExport::exportCRS( HypreMatrix const & mat,
                              arrayView1d< COLUMN_TYPE > const & colIndices,
                              arrayView1d< real64 > const & values ) const
 {
-  int const rank = MpiWrapper::commRank( mat.getComm() );
+  int const rank = MpiWrapper::commRank( mat.comm() );
 
   // import on target rank if needed, or extract diag+offd part in each rank
   hypre_CSRMatrix * const localMatrix = m_targetRank < 0
@@ -168,34 +168,15 @@ void HypreExport::exportCRS( HypreMatrix const & mat,
   GEOSX_LAI_CHECK_ERROR( hypre_CSRMatrixDestroy( localMatrix ) );
 }
 
-namespace hypre
-{
-
-hypre_Vector * parVectorToVectorAll( hypre_ParVector * const vec )
-{
-  if( hypre_ParVectorMemoryLocation( vec ) == HYPRE_MEMORY_HOST )
-  {
-    return hypre_ParVectorToVectorAll( vec );
-  }
-
-  // Create a host-based clone
-  hypre_ParVector * const hostVec = hypre_ParVectorCloneDeep_v2( vec, HYPRE_MEMORY_HOST );
-  hypre_Vector * const fullVec = hypre_ParVectorToVectorAll( hostVec );
-  hypre_ParVectorDestroy( hostVec );
-  return fullVec;
-}
-
-}
-
 void HypreExport::exportVector( HypreVector const & vec,
                                 arrayView1d< real64 > const & values ) const
 {
-  int const rank = MpiWrapper::commRank( vec.getComm() );
+  int const rank = MpiWrapper::commRank( vec.comm() );
 
   // Gather vector on target rank, or just get the local part
   hypre_Vector * const localVector = m_targetRank < 0
                                    ? hypre_ParVectorLocalVector( vec.unwrapped() )
-                                   : hypre::parVectorToVectorAll( vec.unwrapped() );
+                                   : (hypre_Vector *)hypre::parVectorToVectorAll( vec.unwrapped() );
   GEOSX_ERROR_IF( rank == m_targetRank && !localVector, "HypreExport: vector is empty on target rank" );
 
   if( m_targetRank < 0 || m_targetRank == rank )
@@ -219,7 +200,7 @@ void HypreExport::importVector( arrayView1d< real64 const > const & values,
   if( m_subComm != MPI_COMM_NULL )
   {
     hypre_Vector * wrapperVector{};
-    if( MpiWrapper::commRank( vec.getComm() ) == m_targetRank )
+    if( MpiWrapper::commRank( vec.comm() ) == m_targetRank )
     {
       GEOSX_LAI_ASSERT_EQ( values.size(), vec.globalSize() );
       values.move( LvArray::MemorySpace::host, false );
--- a/src/coreComponents/linearAlgebra/interfaces/hypre/HypreExport.cpp
+++ b/src/coreComponents/linearAlgebra/interfaces/hypre/HypreExport.cpp
@@ -193,9 +193,10 @@ void HypreExport::exportVector( HypreVector const & vec,
 void HypreExport::importVector( arrayView1d< real64 const > const & values,
                                 HypreVector & vec ) const
 {
+  hypre_Vector * const localVector = hypre_ParVectorLocalVector( vec.unwrapped() );
   if( m_subComm != MPI_COMM_NULL )
   {
-    hypre_Vector * localVector{};
+    hypre_Vector * wrapperVector{};
     if( MpiWrapper::commRank( vec.getComm() ) == m_targetRank )
     {
       GEOSX_LAI_ASSERT_EQ( values.size(), vec.globalSize() );
@@ -203,26 +204,26 @@ void HypreExport::importVector( arrayView1d< real64 const > const & values,
 
       // HACK: create a hypre vector that points to local data; we have to use const_cast,
       //       but this is ok because we don't modify the values, only scatter the vector.
-      localVector = hypre_SeqVectorCreate( LvArray::integerConversion< HYPRE_Int >( values.size() ) );
-      hypre_VectorOwnsData( localVector ) = false;
-      hypre_VectorData( localVector ) = const_cast< real64 * >( values.data() );
-      hypre_SeqVectorInitialize_v2( localVector, HYPRE_MEMORY_HOST );
+      wrapperVector = hypre_SeqVectorCreate( LvArray::integerConversion< HYPRE_Int >( values.size() ) );
+      hypre_VectorOwnsData( wrapperVector ) = false;
+      hypre_VectorData( wrapperVector ) = const_cast< real64 * >( values.data() );
+      hypre_SeqVectorInitialize_v2( wrapperVector, HYPRE_MEMORY_HOST );
     }
 
     // scatter the data
     hypre_ParVector * const parVector = hypre_VectorToParVector( m_subComm,
-                                                                 localVector,
+                                                                 wrapperVector,
                                                                  hypre_ParVectorPartitioning( vec.unwrapped() ) );
     // copy local part of the data over to the output vector
     HYPRE_Real const * const parVectorData = hypre_VectorData( hypre_ParVectorLocalVector( parVector ) );
-    if( hypre_ParVectorMemoryLocation( vec.unwrapped() ) == HYPRE_MEMORY_HOST )
+    if( hypre_VectorMemoryLocation( localVector ) == HYPRE_MEMORY_HOST )
     {
-      std::copy( parVectorData, parVectorData + vec.localSize(), vec.extractLocalVector() );
+      std::copy( parVectorData, parVectorData + vec.localSize(), hypre_VectorData( localVector ) );
     }
     else
     {
 #ifdef GEOSX_USE_HYPRE_CUDA
-      cudaMemcpy( vec.extractLocalVector(),
+      cudaMemcpy( hypre_VectorData( localVector ),
                   parVectorData,
                   vec.localSize() * sizeof( HYPRE_Real ),
                   cudaMemcpyHostToDevice );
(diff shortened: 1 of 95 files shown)


src/coreComponents/linearAlgebra/interfaces/hypre/HypreExport.cpp
```diff
@@ -165,6 +168,25 @@ void HypreExport::exportCRS( HypreMatrix const & mat,
...
+  {
+    return hypre_ParVectorToVectorAll( vec );
+  }
+
+  // Create a host-based clone
+  hypre_ParVector * const hostVec = hypre_ParVectorCloneDeep_v2( vec, HYPRE_MEMORY_HOST );
+  hypre_Vector * const fullVec = hypre_ParVectorToVectorAll( hostVec );
+  hypre_ParVectorDestroy( hostVec );
+  return fullVec;
+}
```
klevzoff, line 186: `hypre_ParVectorToVectorAll` is not GPU-adapted and relied on vector memory being UM. Now that vector memory is device-only, the workaround is to explicitly copy everything to host before gathering.

src/coreComponents/linearAlgebra/interfaces/hypre/HypreVector.cpp
```diff
@@ -116,216 +87,80 @@ HypreVector::~HypreVector()
...
+  m_values.registerTouch( hypre::memorySpace );
+}
+
 void HypreVector::scale( real64 const scalingFactor )
 {
   GEOSX_LAI_ASSERT( ready() );
   if( !isEqual( scalingFactor, 1.0 ) )
   {
-    GEOSX_LAI_CHECK_ERROR( HYPRE_ParVectorScale( scalingFactor, m_par_vector ) );
+    GEOSX_LAI_CHECK_ERROR( HYPRE_ParVectorScale( scalingFactor, m_vec ) );
```
rrsettgast, line 155: wouldn't this be easier to do on our own?

src/coreComponents/linearAlgebra/interfaces/hypre/HypreVector.cpp
```diff
@@ -396,20 +233,20 @@ void HypreVector::pointwiseProduct( HypreVector const & x,
   GEOSX_LAI_ASSERT_EQ( localSize(), x.localSize() );
   GEOSX_LAI_ASSERT_EQ( localSize(), y.localSize() );
 
-  real64 const * const data = extractLocalVector();
-  real64 const * const x_data = x.extractLocalVector();
-  real64 * const y_data = y.extractLocalVector();
-  forAll< hypre::execPolicy >( localSize(), [y_data, data, x_data] GEOSX_HYPRE_DEVICE ( localIndex const i )
+  arrayView1d< real64 const > my_values = m_values.toViewConst();
```
rrsettgast, line 236: All of these can be `const`
```suggestion
  arrayView1d< real64 const > const my_values = m_values.toViewConst();
```

src/coreComponents/linearAlgebra/interfaces/hypre/HypreVector.cpp
```diff
@@ -459,16 +320,16 @@ void HypreVector::print( std::ostream & os ) const
     MpiWrapper::barrier( getComm() );
     if( rank == myRank )
     {
-      real64 const * const local_data = extractLocalVector();
+      arrayView1d< real64 const > const data = values();
       globalIndex const firstRowID = ilower();
-      for( localIndex i = 0; i < localSize(); ++i )
+      forAll< serialPolicy >( localSize(), [&, data]( localIndex const i )
```
rrsettgast, line 325: I think this would be more useful if it printed out in the desired memory space. We would of course have to use `printf` for this....but I often want to see the values on device rather than the host values.

The thing is that you have to launch a bunch of single threaded kernels in sequence to get the output in order when on device....unless there is a better way to do this.

src/coreComponents/linearAlgebra/interfaces/trilinos/EpetraMatrix.cpp
```diff
@@ -376,6 +376,7 @@ void EpetraMatrix::apply( EpetraVector const & src,
   GEOSX_LAI_ASSERT_EQ( numGlobalCols(), src.globalSize() );
 
   GEOSX_LAI_CHECK_ERROR( m_matrix->Multiply( false, src.unwrapped(), dst.unwrapped() ) );
+  dst.touch();
```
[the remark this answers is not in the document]
rrsettgast: Not a fan of registering a touch in that condition.

One thought I had would be to create an interface function for all of these virtual functions you are having to call `touch()` on. The interface function would call the virtual function, and apply the touch from the base class. The virtual functions become private....etc.

src/coreComponents/linearAlgebra/unitTests/testKrylovSolvers.cpp
```diff
@@ -127,9 +127,9 @@ class KrylovSolverTest : public KrylovSolverTestBase< typename LAI::ParallelMatr
     this->precond.setup( this->matrix );
 
     // Set up vectors
-    this->sol_true.createWithGlobalSize( this->matrix.numGlobalCols(), MPI_COMM_GEOSX );
-    this->sol_comp.createWithGlobalSize( this->matrix.numGlobalCols(), MPI_COMM_GEOSX );
-    this->rhs_true.createWithGlobalSize( this->matrix.numGlobalRows(), MPI_COMM_GEOSX );
+    this->sol_true.create( this->matrix.numLocalCols(), MPI_COMM_GEOSX );
```
rrsettgast, line 130: This seems like a behavior change?

src/coreComponents/physicsSolvers/SolverBase.cpp
```diff
@@ -235,27 +234,32 @@ real64 SolverBase::linearImplicitStep( real64 const & time_n,
 
   // zero out matrix/rhs before assembly
   m_localMatrix.zero();
-  m_localRhs.zero();
+  m_rhs.zero();
+
+  arrayView1d< real64 > localRhs = m_rhs.open();
```
rrsettgast, line 239:
```suggestion
  arrayView1d< real64 > const localRhs = m_rhs.open();
```

src/coreComponents/physicsSolvers/SolverBase.cpp
```diff
@@ -467,34 +468,41 @@ real64 SolverBase::nonlinearImplicitStep( real64 const & time_n,
...
+                                 localRhs );
+
+        m_rhs.close();
+      }
 
       if( m_assemblyCallback )
       {
-        m_assemblyCallback( m_localMatrix, m_localRhs );
+        // TODO: figure out why callback takes LA objects by value, not by view
+        // m_assemblyCallback( m_localMatrix, localRhs );
```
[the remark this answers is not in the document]
rrsettgast: @corbett5 ??

rrsettgast (approved):

src/coreComponents/physicsSolvers/SolverBase.cpp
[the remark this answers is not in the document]
corbett5: To give the user the ability to insert and modify the sparsity without worrying if there might not be enough capacity. Not sure how useful it is, but there's also no way of exporting a `CRSMatrixView` to Python, nor is there a way of exporting an `ArrayView`.

src/coreComponents/linearAlgebra/interfaces/hypre/HypreVector.cpp
klevzoff: The intention is that view semantics always does the right thing and gets you the most recently modified values. What would be the use case for printing directly from device? Device values will only be different if we forget to call `touch()` in one of the LAI methods... I think? But in that case seeing outdated values printed might even help in discovering the bug IMO.

There's also the fact that current API prints to an iostream, which we can't do with `printf`... which if course we could change and make it always go to `stdout`. Still, not a fan of either `printf` (`%d` vs `%lld` thing especially) or launching a cuda kernel per element.

src/coreComponents/linearAlgebra/interfaces/hypre/HypreVector.cpp
klevzoff: Yeah, it's a super trivial kernel to write. In general I try to only write a custom kernel when a hypre method doesn't exist or doesn't work properly on device, the idea being that hypre folks might spend a bit more time optimizing it (e.g. CUDA block sizes). For this simple kernel it's unlikely to matter.

src/coreComponents/linearAlgebra/unitTests/testKrylovSolvers.cpp
klevzoff: Yes, I removed the ability to auto-partition a vector, it just didn't seem that useful to have - we pretty much always want out vectors partitioned in a very specific way (aligned with mesh/dofs/matrix rows).

src/coreComponents/linearAlgebra/interfaces/hypre/HypreVector.hpp
```diff
@@ -244,29 +191,25 @@ class HypreVector final : private VectorBase< HypreVector >
...
-  /**
-   * Pointer to underlying HYPRE_IJVector type.
-   */
-  HYPRE_IJVector m_ij_vector;
-
   /**
    * Pointer to underlying HYPRE_ParVector type.
    */
-  HYPRE_ParVector m_par_vector;
+  HYPRE_ParVector m_vec;
```
castelletto1, line 212: Is there a specific reason to keep using `HYPRE_ParVector`? Given the ongoing refactoring, would it make sense to use directly `hypre_ParVector`? Also, can we now declare `m_vec` as unique pointer?  

```c++
std::unique_ptr< hypre_ParVector > m_vec;
```

castelletto1, line 212: Is there a specific reason to keep using `HYPRE_ParVector`? Given the ongoing refactoring, would it make sense to use directly `hypre_ParVector`?

klevzoff: No particular technical reason, but in theory clients of `HypreVector` class might not be including "private" hypre headers, and the only definitions they would have access to from hypre are these:
https://github.com/hypre-space/hypre/blob/66e1f2df45331ddc01a424586ad86716d8d75367/src/parcsr_mv/HYPRE_parcsr_mv.h#L30-L31
```c
struct hypre_ParVector_struct;
typedef struct hypre_ParVector_struct *HYPRE_ParVector;
```
`HYPRE_ParVector` is also the name of the type in most other hypre function signatures (e.g. all BoomerAMG/Krylov calls) so I would want `unwrapped()` to return exactly that type to avoid confusion (even though it's just an alias for `hypre_ParVector *`). This is also how we do things in `PetscVector` where `Vec` is the "opaque handle" type stored and returned, not `_p_Vec *`.

src/coreComponents/linearAlgebra/interfaces/hypre/HypreVector.cpp
rrsettgast: I view it as a diagnostic tool to make sure you have what you think you have in the place you think it should be. Since the underlying object is managed by `LvArray`, but hypre will not respect the touching/moving semantics. It will just do what it wants. So being able to print both the host and device copies has been useful for me at times.

src/coreComponents/linearAlgebra/interfaces/hypre/HypreVector.cpp
```diff
@@ -481,169 +342,70 @@ void HypreVector::write( string const & filename,
...
+
+      if( globalSize() > 0 )
       {
-        // Copy distributed parVector in a local vector on every process
-        // with at least one component
+        // Copy distributed parVector in a local vector on every process with at least one component
         // Warning: works for a parVector that is smaller than 2^31-1
-        hypre_Vector *vector;
-        vector = hypre_ParVectorToVectorAll( m_par_vector );
+        hypre_Vector * const fullVector = hypre_ParVectorToVectorAll( m_vec );
```
castelletto1, line 366: Don't we need here something as in `HypreExport`? 

https://github.com/GEOSX/GEOSX/blob/a924d56d2887c449ab40b82220dc531785fdd9c7/src/coreComponents/linearAlgebra/interfaces/hypre/HypreExport.cpp#L171-L186

klevzoff: We do (and if only we had a unit test for `write()`, I would have found that out myself)

src/coreComponents/linearAlgebra/interfaces/hypre/HypreExport.cpp
```diff
@@ -193,54 +215,45 @@ void HypreExport::exportVector( HypreVector const & vec,
...
   }
   else
   {
-    hypre_Vector * const localVector = hypre_ParVectorLocalVector( vec.unwrapped() );
-    importArray( hypre_VectorMemoryLocation( localVector ),
+    exportArray( hypre_VectorMemoryLocation( localVector ),
                  values,
                  hypre_VectorData( localVector ) );
   }
+  vec.touch();
```
castelletto1, line 256: The current `touch()` logic should be explained in detail somewhere (not here).

castelletto1:
https://github.com/GEOSX/GEOSX/blob/a924d56d2887c449ab40b82220dc531785fdd9c7/src/coreComponents/linearAlgebra/utilities/LAIHelperFunctions.hpp#L173-L201

Missing return value in both functions.

src/coreComponents/linearAlgebra/interfaces/hypre/HypreExport.cpp
klevzoff: I've documented the usage and intended behavior in doxygen comment for `VectorBase::touch()`, but I don't know where else to put it. It's an implementation detail that is mostly relevant to a group of classes wrapping the same LA package, not external users of these classes.

PR #1689, opened 2021-11-15
Refactor enthalpy and internal energy for CO2 brine

mkhait:
- [x] Compile old-style properties (with derivative computation) to verify refactored properties
- [x] Refactor them using new platform-agnostic style
- [x] Verify that results match
- [x] Delete old-style properties

PR is ready for review&merge, no rebaseline needed


src/coreComponents/constitutive/fluid/PVTFunctions/BrineEnthalpy.hpp
```diff
@@ -160,9 +160,7 @@ void BrineEnthalpyUpdate::compute( real64 const & pressure,
 
   if( useMass )
   {
-    // Do we really need to convert here from mole fraction to mass fraction?
-    real64 const X = C * waterMW / (C * waterMW + (1.0 - C) * CO2MW);
-    value = (1.0 - X ) * CO2Enthalpy + X * brineEnthalpy;
+    value = (1.0 - C ) * CO2Enthalpy + C * brineEnthalpy;
```
mkhait, line 163: After discussion with @francoishamon we decided that this conversion from molar fraction to mass fraction was a mistake and removed it in both old and new implementation.

src/coreComponents/constitutive/fluid/PVTFunctions/BrineEnthalpy.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+  dh = 0.0;
+
+  for( localIndex i = 0; i < 4; ++i )
+    for( localIndex j = 0; j < 3; ++j )
+    {
+
+      dh += a[i][j] * pow( T, real64( i )) * pow( m, real64( j ));
+
+    }
```
francoishamon, line 67: It is just a small detail, but in other places in the code, we started replacing `localIndex` with `integer` for quantities that do not scale with the local mesh size (like the fluid phases, the fluid components, or here) to keep `localIndex` for quantities that really scale with the local mesh size (like the pressure field, etc). But no need to change this in your PR, it is good as is.

src/coreComponents/constitutive/fluid/PVTFunctions/BrineInternalEnergy.hpp
```diff
@@ -0,0 +1,160 @@
...
+
+  BrineInternalEnergy( string const & name,
+                       string_array const & inputParams,
+                       string_array const & componentNames,
+                       array1d< real64 > const & componentMolarWeight )
+    : PVTFunctionBase( name,
+                       componentNames,
+                       componentMolarWeight )
+  {
+    // reserve for future: more accurate internal energy model should probably have some parameters
```
francoishamon, line 87: I agree with you that at some point we should use a more accurate phase internal energy model, because right now it is quite simple :)

francoishamon (approved):
Looks good to me, thanks a lot @mkhait and sorry for the delayed review. That's going to be very useful for thermal. I suggest we try to merge your PR as soon as possible, and then we can connect it to the rest of the CO2Brine fluid package in #1225

src/coreComponents/constitutive/fluid/PVTFunctions/BrineEnthalpy.cpp
mkhait: Ok, thanks for explaining, understood!

mkhait:
Thanks @francoishamon for reviewing! I`m merging develop into this PR now and when/if Travis succeeds, will merge the PR.

PR #1691, opened 2021-11-18
Check that PR is assigned

TotoGaz:
Defines a `check` step that fails (at end) if the PR is assigned to no one (like we do for the `documentation` or `code_style` for instance).
This patch is meant to enforce part of or merge policy.
 
I've checked that it works by assigning and unassigning myself (on this PR).
It is enough to assign anyone and run again the proper steps (2 times 20 seconds) in order to get the global validation to green (no need of a full replay).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -185,7 +185,7 @@ jobs:
     script: scripts/test_submodule_updated.sh
   - stage: checks
     name: check_pr_is_assigned
-    <<: *assign_script
+    <<: *assigned_script
   - stage: builds
     name: Ubuntu18.04_clang8.0.0_cuda10.1.243_debug
     <<: *geosx_linux_build
--- a/.travis.yml
+++ b/.travis.yml
@@ -129,7 +129,7 @@ draft_script: &draft_script
 
 # PR must be assigned to be merged.
 # This script will fail if this is not the case.
-assign_script: &assign_script
+assigned_script: &assigned_script
   script:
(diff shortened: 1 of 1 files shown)


joshua-white (approved):
Looks good to me.  We'll need to update our documentation maybe so new folks will know how to pass this check.

TotoGaz:
Yes, I will update the doc.
Do we have already an appropriate section on `readthedocs`?
That does not appear obvious to me.

rrsettgast:
> Looks good to me.  We'll need to update our documentation maybe so new folks will know how to pass this check.

I think the name of the test 'check_pr_is_assigned' is sufficient?

.travis.yml
```diff
@@ -127,6 +127,18 @@ draft_script: &draft_script
       exit 0
     fi
 
+# PR must be assigned to be merged.
+# This script will fail if this is not the case.
+assign_script: &assign_script
```
rrsettgast, line 132:
```suggestion
assigned_script: &assigned_script
```

.travis.yml
```diff
@@ -170,6 +183,9 @@ jobs:
   - stage: checks
     name: check_submodules
     script: scripts/test_submodule_updated.sh
+  - stage: checks
+    name: check_pr_is_assigned
+    <<: *assign_script
```
rrsettgast, line 188:
```suggestion
    <<: *assigned_script
```

rrsettgast (approved):
Just a suggestion that we call the test assigned instead of assign.

PR #1695, opened 2021-11-22
Fix #1694

### LINKED ISSUE #1694: [Bug] norm2 with PETSc LAI
**Describe the bug**
Using PETSc as linear algebra interface, the rhs has always 0 norm2.

**To Reproduce**
1. Compile geosx with
```
GEOSX_LA_INTERFACE:STRING=Petsc
GEOSX_LA_INTERFACE_PETSC:BOOL=ON
```
2. run any test case, e.g., `inputFiles/simplePDE/10x10x10_LaplaceFEM.xml`
3. check the visit/vtk output: zero everywhere! The linear system is not solved because the residual has zero norm2, even if it is different from the zero vector.

**Expected behavior**
To compute the right norm2.

- Machine: standard Ubuntu 20.04
- Compiler: gcc 9.3.0
- GEOSX current develop (8577777)

klevzoff:
Resolves https://github.com/GEOSX/GEOSX/issues/1694

PETSc caches previously computed norm values, so it needs a little hint to drop them after we've modified the values externally.

### DIFF
--- a/src/coreComponents/linearAlgebra/interfaces/petsc/PetscVector.cpp
+++ b/src/coreComponents/linearAlgebra/interfaces/petsc/PetscVector.cpp
@@ -127,6 +127,8 @@ void PetscVector::close()
   GEOSX_LAI_ASSERT( !closed() );
   m_values.move( LvArray::MemorySpace::host, false );
   m_closed = true;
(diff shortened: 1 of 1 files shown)


andrea-franceschini (approved):
It works. Thanks!

PR #1702, opened 2021-12-03
Extrinsic mesh data for flow solvers

### LINKED ISSUE #1693: [Docs] A small mistake in the Multiphase Flow example documentation
In the [Multiphase Flow example documentation](https://geosx-geosx.readthedocs-hosted.com/en/latest/docs/sphinx/basicExamples/multiphaseFlow/Example.html#events):

> "Then, if the solver converges in **more** than a certain number of nonlinear iterations..."

I think "**more**" should be "**less**".

francoishamon:
This PR implements the extrinsic mesh data in the flow solvers. Each quantity is declared using the macro `EXTRINSIC_MESH_DATA_TRAIT` outside the solver class, in a file named `SolverNameExtrinsicData.hpp` (say, `SinglePhaseBaseExtrinsicData.hpp` for instance).

Then in the solver, the following lines are used to register data and get reference to data, respectively:
```
subRegion.registerExtrinsicData< extrinsicMeshData::pressure >( solverName );
subRegion.getExtrinsicData< extrinsicMeshData::pressure >();
```

This will make the flow solver consistent with the approach used in `SolidMechanicsLagrangianFEM` and will help remove the circular dependency introduced by the new flow kernel interface (see [here](https://github.com/GEOSX/GEOSX/blob/72916b43af99065320c23e365973c63a77c6e5cf/src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBaseKernels.hpp#L346) for instance). Also will help implement the tuple-based approached proposed by @klevzoff for the creation of `ElementViewAccessors`.

Current status
- [X] Single phase flow
- [x] Compositional multiphase flow

If this is not going in the right direction, please let me know!

Fixes #1693


src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -451,8 +449,8 @@ void CompositionalMultiphaseBase::updateFluidModel( Group & dataGroup, localInde
 {
   GEOSX_MARK_FUNCTION;
 
-  arrayView1d< real64 const > const pres = dataGroup.getReference< array1d< real64 > >( viewKeyStruct::pressureString() );
-  arrayView1d< real64 const > const dPres = dataGroup.getReference< array1d< real64 > >( viewKeyStruct::deltaPressureString() );
+  arrayView1d< real64 const > const pres = dataGroup.getReference< array1d< real64 > >( extrinsicMeshData::pressure::key() );
+  arrayView1d< real64 const > const dPres = dataGroup.getReference< array1d< real64 > >( extrinsicMeshData::deltaPressure::key() );
```
TotoGaz, line 453: Is there a reason why you do not use `getExtrinsicData` here (and in other places of the code) while you use in other places?

francoishamon: Yes there is a good reason, sorry, I am not done with the compositional multiphase solvers :) I am working on them now

TotoGaz: HA HA, sorry I should have waited!

francoishamon: It should be good now. I went through the single-phase and compositional solvers, and replaced as many `getReference` as possible (I may have forgotten a few). I will stop here, and do the wells in a separate PR.

TotoGaz (commented):
Thank you @francoishamon and welcome to the club.
Get in the merge line fast!

TotoGaz (approved):

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -367,24 +346,24 @@ void CompositionalMultiphaseBase::initializePreSubGroups()
...
-void CompositionalMultiphaseBase::updateComponentFraction( Group & dataGroup ) const
+void CompositionalMultiphaseBase::updateComponentFraction( ObjectManagerBase & dataGroup ) const
 {
   GEOSX_MARK_FUNCTION;
 
   // outputs
 
   arrayView2d< real64, compflow::USD_COMP > const & compFrac =
-    dataGroup.getReference< array2d< real64, compflow::LAYOUT_COMP > >( viewKeyStruct::globalCompFractionString() );
+    dataGroup.getExtrinsicData< extrinsicMeshData::globalCompFraction >();
```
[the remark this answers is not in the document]
francoishamon: Done, I put this on one line

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoromechanicsKernel.hpp
```diff
@@ -141,33 +143,29 @@ class Multiphase :
...
+      m_deltaFluidPressure = elementSubRegion.template getExtrinsicData< extrinsicMeshData::deltaPressure >();
     }
 
     // extract views into multiphase solver data
     {
-      using keys = CompositionalMultiphaseBase::viewKeyStruct;
-
       m_fluidPhaseDensityOld =
-        elementSubRegion.template getReference< array2d< real64, compflow::LAYOUT_PHASE > >( keys::phaseDensityOldString() );
+        elementSubRegion.template getExtrinsicData< extrinsicMeshData::phaseDensityOld >();
```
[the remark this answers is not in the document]
francoishamon: Yes I agree, I put this on one line

klevzoff (approved):
Great, I'm glad this is moving forward! Been on my todo list for a long time as well.
Line-by-line diff looks good to me.

As a general comment, I want to invite opinions on the following topics:
1. Scoping. Given that we've mostly moved away from a giant unstructured pile of strings in `dataRepository::keys` and made each key scoped to the solver that "owns" the corresponding field, do we want to now put all traits back into the same namespace? I.e. `extrinsicMeshData::pressure` instead of `extrinsicMeshData::flow::pressure` or some? Should we come up with a consistent scoping strategy for field traits before we move ahead with applying the pattern universally?
2. Looking through the PR, I'm growing less and less fond of the repeated term "extrinsic". I know we've debated over the name, and I don't intend to engage in bikeshedding, but have we considered something simpler and more familiar, like `fields::pressure` instead of `extrinsicMeshData::pressure`?
3. I think I prefer
```c++
group.getExtrinsicData( extrinsicMeshData::pressure{} );
```
over
```c++
group.getExtrinsicData< extrinsicMeshData::pressure >();
```
It's the same amount of typing, but gives us a little more flexibility, for example some trait struct could be made stateful and have a dynamic key, e.g.
```c++
group.registerExtrinsicData( extrinsicMeshData::myAwesomeField( this->getName() + "_awesome" ) );
```
Same thing could be done with e.g. default value - it could be made runtime instead of a compile-time constant. This would apply on a field-by-field basis, most fields traits would keep their current definition with static traits. This former pattern seems to be commented out in `ObjectManagerBase.hpp` but we could revisit it.

Perhaps all of these questions warrant a separate issue/discussion. Let me know if you have any thoughts @francoishamon @TotoGaz @rrsettgast @joshua-white .

TotoGaz:
Hi @klevzoff thanks for your comments.

For 1, as stated in the zen of python `Namespaces are one honking great idea -- let's do more of those!`.

For 2, as part of the initial discussion, I was wondering if "_Do we have something like a Field concept (ScalarField, VectorField)? Is it close to what you mean by extrinsic information?_" (see https://github.com/GEOSX/GEOSX/pull/938#issuecomment-626821734, and the whole initial discussion if you're interested). How does it behave with finite elements and their discretization support?

For 3, I am personally not in favor of using templates _when it can be done with standard functions_. I do no think there is any performance or `virtual` calls issues here (isn't it?).
That helps new comers (and I'm sure old comers too 😉 !)!
It helps refactoring. IDE are very powerful at helping in those cases.

Furthermore, I think that the most critical point is `python`.
If `GEOSX` wants to move forward to `python`, then we may eventually allow python users to register data/fields.
How could we do this in the current situation?
The `registerExtrinsicData` is not a function since it's a template function and needs to be instantiated with a concrete type. That will be impossible to do from `python` and as such, by design, the current `registerExtrinsicData< T >` is really `C++` oriented and _compile time_. If we want to offer the same solutions in `python` (which is really _runtime_), we may have to duplicate/reimplement.

But first, have all the "_extrinsic stuff_" been migrated or do we still have some to do?
If we decide to change the pattern again, it's always better to move one pattern to another and not a mix of pattern to another an even greater mix 😉 🤣

francoishamon:
I like the suggestions above:
- I am going to implement the suggested namespace. I am planning to do only do `flow` in this PR, and leave `wavePropagation` and `solidMechanics` (which also have extrinsic data) for separate PRs.
- If there is no objection, I will change from `extrinsicMeshData` to `fields`.
- I will change to `{}`.

and then I will wrap up the PR.

TotoGaz:
@francoishamon Is the total migration over or will we end up with 3 ways of doing things in the codebase?

francoishamon:
We can also do wave propagation (see [here](https://github.com/GEOSX/GEOSX/blob/1ac5032a4544f8d7d7828b10df1a9142ba7496a2/src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEM.hpp#L161)) and take care of [this](https://github.com/GEOSX/GEOSX/blob/1ac5032a4544f8d7d7828b10df1a9142ba7496a2/src/coreComponents/mesh/ExtrinsicMeshData.hpp#L93), but that will be a larger PR. What do you think? My preference is to stick to the flow solvers and do small, separate PRs for the other solvers.

TotoGaz:
> We can also do wave propagation (see [here](https://github.com/GEOSX/GEOSX/blob/1ac5032a4544f8d7d7828b10df1a9142ba7496a2/src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEM.hpp#L161)) and take care of [this](https://github.com/GEOSX/GEOSX/blob/1ac5032a4544f8d7d7828b10df1a9142ba7496a2/src/coreComponents/mesh/ExtrinsicMeshData.hpp#L93), but that will be a larger PR. What do you think? My preference is to stick to the flow solvers and do small, separate PRs for the other solvers.

By doing this will we end up with

- The old `registerWrapper` way
- The now old `registerExtrinsicData< T >()` way
- The new `registerField( ... )` way

in the same time in the same fwk?

I would not personally begin implementation right away (unless you have a real need for this, obviously).

klevzoff:
Just to be clear, I think the PR is good to go as is, without additional changes. 
> * I am going to implement the suggested namespace. I am planning to do only do `flow` in this PR, and leave `wavePropagation` and `solidMechanics` (which also have extrinsic data) for separate PRs.

This a fairly local decision that affects only changes in this PR, so completely up to you. It's also not difficult to revert if the decision to keep everything in one namespace comes in later.

> * If there is no objection, I will change from `extrinsicMeshData` to `fields`.

I would definitely make sure we have consensus first, even though renaming a namespace is a quick and simple refactoring. The current name was the subject of some debate, but it stayed in the end as the best alternative. I only suggested to re-evaluate it, but not necessarily change right away. Also, if we changed it, it'd be also reasonable to rename methods like `getExtrinsicData()`, so it's a actually an even bigger decision.

> * I will change to `{}`.

Feel free to uncomment code for this variant of `get`/`register` functions in `ObjectManager.hpp` and see what happens. I've seen some comments there about pre-C++17 usage of static constexpr data, let's make sure we don't trip over some linking issues. Again, this is a refactoring that can be done later.

klevzoff:
> For 3, I am personally not in favor of using templates _when it can be done with standard functions_. I do no think there is any performance or `virtual` calls issues here (isn't it?). That helps new comers (and I'm sure old comers too wink !)! It helps refactoring. IDE are very powerful at helping in those cases.

Well... generic registration of user data _cannot_ be done with non-templated functions. Even if we limit ourselves to only `Array<>` types (not the only one, but by far the most common type to register on the mesh), the sheer number of possible combinations of element type, dimension (we have use cases for up to 5 dims) and layout makes it a huge combinatorial explosion problem if we were to try and dispatch these 3 runtime arguments into compile time (as needed in order to instantiate the typed `Wrapper<>` under the hood). I don't think there is a good workaround (just in case, any approach that involves macros or codegen is the same as templates but worse IMO).

> Furthermore, I think that the most critical point is `python`. If `GEOSX` wants to move forward to `python`, then we may eventually allow python users to register data/fields. How could we do this in the current situation? The `registerExtrinsicData` is not a function since it's a template function and needs to be instantiated with a concrete type. That will be impossible to do from `python` and as such, by design, the current `registerExtrinsicData< T >` is really `C++` oriented and _compile time_. If we want to offer the same solutions in `python` (which is really _runtime_), we may have to duplicate/reimplement.

I imagine that for Python we might actually provide a limited subset of field types to register (say, limited to arrays with `integer`/`real64` data, 1-3 dimensions and default layout). I'd personally prefer if it was the Python interface class that was doing the dispatch on runtime args and generating all the calls to `registerWrapper<>` with compile-time args, while the C++ code may stay idiomatic (i.e. templated).

francoishamon:
> Just to be clear, I think the PR is good to go as is, without additional changes.
> 
> > * I am going to implement the suggested namespace. I am planning to do only do `flow` in this PR, and leave `wavePropagation` and `solidMechanics` (which also have extrinsic data) for separate PRs.
> 
> This a fairly local decision that affects only changes in this PR, so completely up to you. It's also not difficult to revert if the decision to keep everything in one namespace comes in later.

In the last commit, I introduced the namespace `extrinsicMeshData::flow` for the flow fields. It is now used in the flow solvers.

> 
> > * If there is no objection, I will change from `extrinsicMeshData` to `fields`.
> 
> I would definitely make sure we have consensus first, even though renaming a namespace is a quick and simple refactoring. The current name was the subject of some debate, but it stayed in the end as the best alternative. I only suggested to re-evaluate it, but not necessarily change right away. Also, if we changed it, it'd be also reasonable to rename methods like `getExtrinsicData()`, so it's a actually an even bigger decision.

Ok sounds good. I did not change anything in the current PR (no change from `extrinsicMeshData` to `fields`).

> 
> > * I will change to `{}`.
> 
> Feel free to uncomment code for this variant of `get`/`register` functions in `ObjectManager.hpp` and see what happens. I saw some comments there about pre-C++17 usage of static constexpr data, let's make sure we don't trip over some linking issues. Again, this is a refactoring that can be done later.

I gave it a try. I uncommented the code that was ifdef'ed out in `ObjectManagerBase.hpp` and updated a few things there (change from `Type` to `type`, etc). When I tried to call it this new `getExtrinsicData` from the solver, I got:

```
/g/g90/hamon1/geosx/francois/GEOSX/src/coreComponents/mesh/ObjectManagerBase.hpp:693:12: error: non-const lvalue reference to type 'ArrayView<...>' cannot bind to a temporary of type '\
ArrayView<...>'                                                                                                                                                                          
    return this->getWrapper< typename MESH_DATA_TRAIT::type >( extrinsicDataTrait.key() ).referenceAsView();                                                                             
           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
on Quartz and 
```
/g/g90/hamon1/geosx/francois/GEOSX/src/coreComponents/mesh/ObjectManagerBase.hpp(693): error: initial value of reference to non-const must be an lvalue                                 \
   detected during instantiation of "auto &geosx::ObjectManagerBase::getExtrinsicData(const MESH_DATA_TRAIT &) [with MESH_DATA_TRAIT=geosx::extrinsicMeshData::flow::pressure]" 
```
on Lassen. I did not try anything further than that, and ended up putting back the `#if 0` to comment out the code in `ObjectManagerBase.hpp`. 

I think the PR is ready to merge.

francoishamon:
In passing, I am changing one word in the documentation to fix #1693

PR #1704, opened 2021-12-04
Feature/besset pygeosx

labesse40:
This PR is about new structures for pygeosx which would allow the python user to do multi-realization and have more flexibility to create complex workflow.
- Direct access to methods of SolverBase
- Creation of custom "reinit" method for each solver to re-initialize variables that would not be accessible from PyWrapper
- Time loop control from python
- Control over TimeHistoryCollection and TimeHistoryOutput from python


src/coreComponents/events/EventBase.hpp
```diff
@@ -304,8 +310,8 @@ class EventBase : public ExecutableGroup
    * @brief Get the target of this event.
    * @return The target of this event.
    */
-  ExecutableGroup * getEventTarget() const
-  { return m_target; }
+  // ExecutableGroup * getEventTarget() const
```
corbett5, line 313: Remove.

src/coreComponents/events/EventBase.hpp
```diff
@@ -261,6 +261,12 @@ class EventBase : public ExecutableGroup
   bool isIdle() const
   { return m_eventForecast > 1; }
 
+  ExecutableGroup * getEventTarget() const
```
corbett5, line 264: Document.

src/coreComponents/mainInterface/ProblemManager.hpp
```diff
@@ -168,6 +168,11 @@ class ProblemManager : public dataRepository::Group
    * @brief Returns a pointer to the DomainPartition
    * @return Pointer to the DomainPartition
    */
+
+  EventManager & getEventManager()
```
corbett5, line 172: Document.

src/coreComponents/physicsSolvers/SolverBase.hpp
```diff
@@ -648,6 +648,44 @@ class SolverBase : public ExecutableGroup
...
+
+  /**
+   * @brief Performs initialization required after reading from a restart file.
+   */
+  virtual void postRestartInitialization() override {}
+
+  /**
+   * @brief Performs re-initialization of certain variable depending on the solver being used.
+   */
+  virtual void reinit() override {}
```
corbett5, line 686: What's the reason to add these?

src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEM.cpp
```diff
@@ -146,6 +145,11 @@ void AcousticWaveEquationSEM::postProcessInput()
                   "Invalid number of physical coordinates for the receivers",
                   InputError );
 
+  EventManager const & event = this->getGroupByPath< EventManager >( "/Problem/Events" );
+  real64 const & maxTime = event.getReference< real64 >( EventManager::viewKeyStruct::maxTimeString() );
```
corbett5, line 149: Not that this should be fixed in this PR, but we should really not be accessing things like this by traversing the data repository.

src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEM.cpp
```diff
@@ -260,40 +264,65 @@ void AcousticWaveEquationSEM::addSourceToRightHandSide( real64 const & time_n, a
...
+      {
+        ptmp_np1 += pressure_np1[receiverNodeIds[ircv][inode]] * receiverConstants[ircv][inode];
+        ptmp_n += pressure_n[receiverNodeIds[ircv][inode]] * receiverConstants[ircv][inode];
+      }
+      p_rcvs[iSeismo][ircv] = ((time_np1 - time_seismo)*ptmp_n+(time_seismo - time_n)*ptmp_np1)/dt;
+    }
+  } );
+
+  /*
+     arrayView2d< localIndex const > const receiverNodeIds = m_receiverNodeIds.toViewConst();
```
corbett5, line 294: Delete.

src/coreComponents/physicsSolvers/wavePropagation/WaveSolverBase.cpp
```diff
@@ -68,6 +78,15 @@ WaveSolverBase::~WaveSolverBase()
 }
 
 
+void WaveSolverBase::reinit()
+{
+  DomainPartition & domain = this->getGroupByPath< DomainPartition >( "/Problem/domain" );
```
corbett5, line 83: `getGlobalState().getProblemManger().getDomainPartition()`

src/pygeosx/pygeosx.cpp
```diff
@@ -347,11 +360,15 @@ PyInit_pygeosx()
 {
   import_array();
 
-  LvArray::python::PyObjectRef<> module{ PyModule_Create( &pygeosxModuleFunctions ) };
-  if( module == nullptr )
-  {
-    return nullptr;
-  }
+  PyObject * module = PyModule_Create( &pygeosxModuleFunctions );
```
corbett5, line 363: Can you replace these with `PyObjectRef`, similar to the previous implementation?  And try and catch the errors from module creation? Otherwise you risk leaks.

src/pygeosx/PyGroup.hpp
```diff
@@ -26,6 +27,77 @@ namespace python
...
+  "Returns\n"
+  "_______\n"
+  "Wrapper\n"
+  "    The wrapper at the relative path.";
+
+
+template< typename T >
+static PyObject * PyGroup_getWrapper( T * const self, PyObject * const args ) noexcept
+{
+  //VERIFY_NON_NULL_SELF( self );
```
corbett5, line 53: Replace these with the macro expansions.

src/pygeosx/pygeosx.cpp
```diff
@@ -347,11 +360,15 @@ PyInit_pygeosx()
...
+  PyObject * submodule1 = geosx::python::PyInit_pysolver();
+  PyObject * submodule2 = geosx::python::PyInit_pyhdf5();
+
+  Py_INCREF( submodule1 );
+  PyModule_AddObject( module, "pysolver", submodule1 );
+
+  Py_INCREF( submodule2 );
+  PyModule_AddObject( module, "pyhdf5", submodule2 );
```
corbett5, line 371: Any particular reason these are added as new sub-modules?

src/pygeosx/pygeosx.cpp
```diff
@@ -376,12 +393,22 @@ PyInit_pygeosx()
     return nullptr;
   }
 
+  if( !LvArray::python::addTypeToModule( module, geosx::python::getPySolver2Type(), "Solver2" ) )
```
corbett5, line 396: What's `Solver2`? I see it here but nowhere else.

src/pygeosx/pysolver/PySolver.cpp
```diff
@@ -0,0 +1,303 @@
...
+#define VERIFY_NON_NULL_SELF( self ) \
+  PYTHON_ERROR_IF( self == nullptr, PyExc_RuntimeError, "Passed a nullptr as self.", nullptr )
+
+#define VERIFY_INITIALIZED( self ) \
+  PYTHON_ERROR_IF( self->group == nullptr, PyExc_RuntimeError, "The PySolver is not initialized.", nullptr )
+
+
+namespace geosx
+{
+extern std::unique_ptr< GeosxState > g_state;
```
corbett5, line 18: I don't really like the `extern`. Could we have a function in `pygeosx.hpp` to return these instead?

src/pygeosx/pysolver/PySolver.cpp
```diff
@@ -0,0 +1,303 @@
...
+
+struct PySolver
+{
+  PyObject_HEAD
+
+  static constexpr char const * docString =
+    "A Python interface to geosx::SolverBase.";
+
+  geosx::SolverBase * group;
+  geosx::ProblemManager * pb_manager;
```
corbett5, line 33: What does adding the `ProblemManager` here help? I'd imagine it might be better to just access everything through `GeosxState`.

src/pygeosx/PyGroup.cpp
```diff
@@ -250,14 +254,31 @@ static PyObject * PyGroup_getWrapper( PyGroup * const self, PyObject * const arg
...
 
-  string groupPath, wrapperName;
-  std::tie( groupPath, wrapperName ) = splitPath( path );
-
   try
   {
-    dataRepository::Group & group = self->group->getGroupByPath( groupPath );
-    dataRepository::WrapperBase & wrapper = group.getWrapperBase( wrapperName );
-    return createNewPyWrapper( wrapper );
+    geosx::ProblemManager * pbManager = static_cast< geosx::ProblemManager * >(self->group);
```
corbett5, line 259: So this is a function that's defined on `PyGroup` but is only valid if it's called on the `PyGroup` wrapping the problem manager?

src/pygeosx/PyGroup.cpp
```diff
@@ -250,14 +254,31 @@ static PyObject * PyGroup_getWrapper( PyGroup * const self, PyObject * const arg
...
+        break;
+      }
+      else
+      {
+        subEvent = nullptr;
+      }
+    }
+    PYTHON_ERROR_IF( subEvent == nullptr, PyExc_RuntimeError, "Target not found", nullptr );
+
+    return createNewPySolver2( subEvent, pbManager );
```
corbett5, line 281: Where is this defined?

src/pygeosx/PyGroup.cpp
```diff
@@ -210,23 +212,25 @@ static PyObject * PyGroup_getGroup( PyGroup * const self, PyObject * const args
   }
 }
 
-static constexpr char const * PyGroup_getWrapperDocString =
-  "get_wrapper(self, path, default, /)\n"
+
+static constexpr char const * PyGroup_getSolverDocString =
+  "get_solver(self, path, default, /)\n"
```
corbett5, line 217: The documentation needs to be updated.

src/pygeosx/PyGroup.cpp
```diff
@@ -250,14 +254,31 @@ static PyObject * PyGroup_getWrapper( PyGroup * const self, PyObject * const arg
...
+    eventManager.forSubGroups< geosx::EventBase >( [&]( geosx::EventBase & subEvent )
+    {
+      subEvent.getTargetReferences();
+    } );
+
+    geosx::EventBase * subEvent = nullptr;
+    for( int currentSubEvent = 0; currentSubEvent<eventManager.numSubGroups(); ++currentSubEvent )
+    {
+      subEvent = static_cast< geosx::EventBase * >( eventManager.getSubGroups()[currentSubEvent]);
+      if( subEvent->getEventName() == path )
```
corbett5, line 270: So here you really just pass the name of the solver right, and not the full path? And then if you input something that is an `Event` but not a `Solver` what happens? Is there not a better way to loop over the physics solvers than to loop over all events?

src/pygeosx/pyhdf5/PyHDF5.cpp
```diff
@@ -0,0 +1,294 @@
...
+struct PyHDF5
+{
+  PyObject_HEAD
+
+  static constexpr char const * docString =
+    "A Python interface to HDF5.";
+
+  std::map< string, geosx::PackCollection * > collection;
+  std::map< string, geosx::TimeHistoryOutput * > output;
+  geosx::ProblemManager * pb_manager;
```
corbett5, line 36: Again I'm not sure having the problem manager here is the best idea. Also why not have a pointer to the `TimeHIstoryCollection` group?

src/pygeosx/pyhdf5/PyHDF5.cpp
```diff
@@ -0,0 +1,294 @@
...
+namespace geosx
+{
+extern std::unique_ptr< GeosxState > g_state;
+
+extern bool g_alreadyInitialized;
+
+namespace python
+{
+
+struct PyHDF5
```
corbett5, line 27: I'd also rename this to something related to time history and not HDF5.

src/pygeosx/pyhdf5/PyHDF5.cpp
```diff
@@ -0,0 +1,294 @@
...
+  PyObject * module = PyModule_Create( &pyhdf5module );
+
+  if( PyType_Ready( &PyHDF5Type ) < 0 )
+    return nullptr;
+
+  if( module == nullptr )
+    return nullptr;
+
+  Py_INCREF( &PyHDF5Type );
+  PyModule_AddObject( module, "HDF5", (PyObject *)&PyHDF5Type );
```
corbett5, line 285: I'd rewrite these using `PyObectRef` to eliminate leaks.

src/pygeosx/pysolver/PySolver.cpp
```diff
@@ -0,0 +1,303 @@
...
+  }
+
+  geosx::EventManager & eventManager = self->pb_manager->getEventManager();
+
+  eventManager.forSubGroups< geosx::EventBase >( [&]( geosx::EventBase & subEvent )
+  {
+    subEvent.getTargetReferences();
+  } );
+
+  geosx::EventBase * subEvent = nullptr;
```
corbett5, line 101: This looks very similar to `PyGroup::getSolver`. Also not a fan of the two `static_cast`s below.

src/pygeosx/pysolver/PySolver.cpp
```diff
@@ -0,0 +1,303 @@
...
+  double time;
+  double dt;
+  if( !PyArg_ParseTuple( args, "dd", &time, &dt ) )
+  {
+    return nullptr;
+  }
+
+  geosx::DomainPartition & domain = self->pb_manager->getDomainPartition();
+
+  self->group->explicitStep( time, dt, 0, domain );
```
corbett5, line 181: What happens if you've got an implicit solver but you call this function?

src/pygeosx/pysolver/PySolver.cpp
```diff
@@ -0,0 +1,303 @@
...
+  "pysolver module for SolverBase",
+  -1,
+  NULL, NULL, NULL, NULL, NULL
+};
+
+
+PyMODINIT_FUNC
+PyInit_pysolver( void )
+{
+  PyObject * module = PyModule_Create( &pysolvermodule );
```
corbett5, line 285: Same leak stuff.

corbett5 (changes requested):
Very cool, super glad someone is improving this! It's probably just because I'm out of the loop but some examples of the new functionality would be nice too.

src/coreComponents/physicsSolvers/SolverBase.hpp
labesse40: The reason originally was to be able to access all these reinitialization routines from the PySolver to be able to call them from python and then simplify multi-realization. 
We had a discussion with Josh and Randy and it has been decided to create the "reinit" routine in Group.hpp and access it in the PySolver interface, then this routine is overriden by each solver dependeing on the variables to re-initialize and how all this other initialization routines (postProcessInput etc.) are overriden.

src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEM.cpp
labesse40: Yes we've discussed this with François, but since this is done in multiple places over GEOSX, we decided to put it this way for now

src/coreComponents/events/EventBase.hpp
labesse40: Done.

src/coreComponents/events/EventBase.hpp
labesse40: Done.

src/coreComponents/mainInterface/ProblemManager.hpp
labesse40: Done.

src/coreComponents/events/EventBase.hpp
labesse40: Done.

src/pygeosx/pysolver/PySolver.cpp
labesse40: Sorry the getSolver routine was for the second version I had of the PySolver Structure. I decided to not put it in this PR

labesse40: I took the `static_cast` logic directly from the `EventManager::run`

src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEM.cpp
labesse40: Done.

src/coreComponents/physicsSolvers/wavePropagation/WaveSolverBase.cpp
labesse40: Done.

src/pygeosx/pygeosx.cpp
labesse40: Solver2 is another way to create PySolver structure in the way you created PyWrapper and PyGroup

src/pygeosx/pysolver/PySolver.cpp
labesse40: I rathered having a pointer to ProblemManager as I didn't want to go through GeosxState each time, I don't really remember why though...

src/pygeosx/PyGroup.cpp
labesse40: In the second version of creating a PySolver, this was kind of an assumption that it should be done from a ProblemManager PyGroup since I found bizarre to create a Solver from a Group that would not be related to Solver

src/pygeosx/PyGroup.cpp
labesse40: Other version I didn't put in this PR

src/pygeosx/pygeosx.cpp
labesse40: The idea is that pygeosx is the module to the whole "GEOSX machine" (especially initialization part of the machine), pysolver and pyhdf5 (and possibly other features) are components of this machine that have their own utilities. I don't know if I'm clear, but this is why I structured the things this way.

src/pygeosx/pysolver/PySolver.cpp
labesse40: I tried in a first time but I think I had some issues. I will try this again.

src/pygeosx/PyGroup.cpp
labesse40: Removed.

src/pygeosx/PyGroup.cpp
labesse40: That is right, the user just gives the name of the Solver not the full path. 
Yes you're right this can be an issue. I don't know if there is a way to loop only over the Solver I will check that. Could we also imagine checking the dynamic type of the event ?

src/coreComponents/physicsSolvers/wavePropagation/WaveSolverBase.cpp
labesse40: If I'm not mistaking, I think I can't do this because the `getGlobalState()` method is private within `GeosxState.hpp`.

src/coreComponents/physicsSolvers/SolverBase.hpp
corbett5: I guess for example `postProcessInput()` already has that exact definition in `Group.hpp` https://github.com/GEOSX/GEOSX/blob/1ac5032a4544f8d7d7828b10df1a9142ba7496a2/src/coreComponents/dataRepository/Group.hpp#L1300

So what's the point of re-defining it again?

src/coreComponents/physicsSolvers/wavePropagation/WaveSolverBase.cpp
corbett5: Nope, `getGlobalState()` is public.

src/pygeosx/pygeosx.cpp
corbett5: Cool, just curious.

src/pygeosx/pygeosx.cpp
corbett5: So then all references to it should be removed right?

src/pygeosx/pysolver/PySolver.cpp
corbett5: Yeah those `static_casts` should be `dynamic_casts`, but this is even more dangerous than that because here you're casting to `SolverBase` instead of `EventBase`.

src/pygeosx/PyGroup.hpp
labesse40: I did something using macro expansion but I don't know if this is what you had in mind

src/pygeosx/PyGroup.cpp
corbett5: I'd like to see how you get one of these solver objects from Python. My understanding is that it's something like this

``` python
problem = geosx.initialize(...)
solver = geosx.PySolver.solver('solverName')
```

My suggestion is that it is done as follows

``` python
problem = geosx.initialize(...)
solver = geosx.getGroup('path/to/solvers/solverName')
```

Then internally the C++ function returns a `geosx.PySolver.solver` type (which inherits from `geosx.Group`) instead of the `geosx.Group` object. This could easily be extended to support whatever arbitrary group types we want to export to python (including the time history stuff). Yes this requires that you specify the full path to the solver, but when combined with @cssherman's existing pure-python `pygeosx` helpers this is not a problem.

src/coreComponents/physicsSolvers/SolverBase.hpp
labesse40: The re-definition of these was to make them `Public` for the `Solvers` since they are `Protected` in `Group.hpp`. But since we now use the `reinit()` I just need this one function to be public in the different solvers. I can now remove all of the other re-definition.

corbett5: Ahh that makes sense, thanks!

src/pygeosx/pygeosx.cpp
labesse40: Done in new commit. Sorry for that.

src/pygeosx/PyGroup.hpp
corbett5: I meant replace `VERIFY_NON_NULL_SELF(self)` with `PYTHON_ERROR_IF( self == nullptr, PyExc_RuntimeError, "Passed a nullptr as self.", nullptr )`, which is the macro expansion, and then include the necessary headers.

labesse40: Aaah ok that makes sense.

src/coreComponents/dataRepository/Group.hpp
```diff
@@ -1329,6 +1329,14 @@ class Group
   virtual void postRestartInitialization()
   {}
 
+  /**
+   * @brief Performs re-initialization of certain variable depending on the solver being used.
+   */
+  virtual void reinit()
+  {}
```
TotoGaz, line 1336: Why do you have to add a member function to nearly all the classes of `GEOSX` when only the `WaveSolver*` family use them (I think)? Is it because of technical consideration in `python`?

src/coreComponents/physicsSolvers/wavePropagation/WaveSolverBase.hpp
```diff
@@ -54,11 +54,28 @@ class WaveSolverBase : public SolverBase
...
   };
 
+
+
 protected:
 
+  /**
+   * @brief Re-initialize source and receivers positions in the mesh, and resize the pressureNp1_at_receivers array
+   */
+  virtual void reinit() override final;
```
TotoGaz, line 70: BTW, you do not need both `override` and `virtual`. You can just keep `override`.

src/coreComponents/events/EventBase.hpp
```diff
@@ -261,6 +261,20 @@ class EventBase : public ExecutableGroup
...
+   * @return The target of this event.
+   */
+  ExecutableGroup * getEventTarget() const
+  { return m_target; }
+
+  /**
+   * @brief Get the name of this event target.
+   * @return The name of the target of this event.
+   */
+  string getEventName() const
```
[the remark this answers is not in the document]
labesse40: Ok I will change it to `getTargetName` then. Thanks

src/coreComponents/dataRepository/Group.hpp
labesse40: Yes we had a discussion with Randy and Josh about how we should re-initialize the variables from python, and it has been decided to create a `reini()` method that would be overriden in each Solver depending on the variables subject to re-initialization. I only override it the WaveSolverBase because I don't know how the other Solver should rewrite this function since I don't know how the other solvers work

labesse40: I guess I could declare this method in `SolverBase` though.

src/coreComponents/physicsSolvers/wavePropagation/WaveSolverBase.hpp
labesse40: Done.

src/coreComponents/dataRepository/Group.hpp
TotoGaz: OK. But why don't you implement a protected `reinit` member function only in `SolverBase` then?
Why do you have to put it on top in `Group` if you only expect `SolverBase` implementations to use it?

src/coreComponents/events/EventBase.hpp
```diff
@@ -261,6 +261,20 @@ class EventBase : public ExecutableGroup
   bool isIdle() const
   { return m_eventForecast > 1; }
 
+  /**
+   * @brief Get the target of this event.
+   * @return The target of this event.
+   */
+  ExecutableGroup * getEventTarget() const
+  { return m_target; }
```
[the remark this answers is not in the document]
labesse40: For this I just followed the logic of all the `getter` define above.

src/coreComponents/dataRepository/Group.hpp
labesse40: I need it to be `Public` because the `reinit()` is called from `PySolver` files.

TotoGaz: I do not know all the details (specially the way you do in `python`) but in `C++`, if you explicitly call a `WaveSolver*` instance, you could have a `public` `reinit` member in `WaveSolver*` classes and `protected` (even `private` I think) in `SolverBase`.

labesse40: Ok I see what you mean, I can make it `Protected` in `SolverBase` and `Public` in `WaveSolver` then, it should be sufficient.

labesse40: Actually not, I really need it to be `Public` in `SolverBase` because I am manipulating `Solverbase` Object in `PySolver.cpp` when calling `reinit()`

src/pygeosx/PyGroup.cpp
labesse40: You're right this is the way it is done on `Python` side.
I see what you mean and this is kind of what I did in the second version of `PySolver` `(PySolver2)` except that I had a different routine `getSolver` which creates the `PySolver` object. What you propose is that `getGroup` does actually all sort of `PyStructure` creation. In my understanding this would mean that we would have some sort of mechanism to know which `PyStructure` to build based on the `path` of the `group` right ?

src/pygeosx/pysolver/PySolver.cpp
labesse40:
```
***** ERROR
***** LOCATION: /home/julien/codes/GEOSX/src/coreComponents/physicsSolvers/SolverBase.cpp:647
***** Rank 0: SolverBase::ImplicitStepSetup called!. Should be overridden.
```
Is there a way to `catch` a wrong call method ?

corbett5: Is this raised as a Python exception or does it kill the program? The first is fine, but the second isn't (just change it from a `GEOSX_ERROR_IF` to `GEOSX_THROW_IF`, and then wrap it in a `try-except` block. But what about if you call it on a solver that can do both, like the `SolidMechanics`?

src/pygeosx/PyGroup.cpp
corbett5: You could add a function
```c++
PyTypeObject * getPythonType() virtual const
{ return getPyGroupType(); }
```

to `Group` and then derived types could override it with their own type (should inherit from the Python `Group` class), then all you'd have to do is in `createNewPyGroup` replace `getPyGroupType()` with `group.getPythonType()`. It'd take a bit of trickery to avoid circular references, but I think something like that would be best. If we're going to have more of these python types we'll be exporting perhaps we should move them out of the `pygeosx` directory and put them under `dataRepository/python` for example.

src/coreComponents/physicsSolvers/wavePropagation/WaveSolverBase.cpp
labesse40: Right I just forgot to include the `GeosxState.hpp` file. Done

src/pygeosx/pyhdf5/PyHDF5.cpp
labesse40: Right, I changed it to point to `HistoryCollection`

src/pygeosx/pyhdf5/PyHDF5.cpp
labesse40: I renamed it as `PyHistory`

src/pygeosx/pysolver/PySolver.cpp
labesse40: Done.
For the case where a `Solver` can do both, could we add flags in the solvers` that would tell if it's using `linear` or `non linear` and `explicit` or `implicit` depending on what is defined in the XML ? Then we would rely on that on `PySolver` side

klevzoff: What's the reason we're not simply exposing `solverStep` as the main "driver" function for a solver? GEOSX solvers already implement switching logic for different step types, so why shift the burden on user? (I'm a little out of context on this PR, so apologies if the answer is obvious).

labesse40: Actually I think you're absolutely right @klevzoff. It should be sufficient to just expose `solverStep`

src/pygeosx/pygeosx.cpp
labesse40: Done.

src/pygeosx/pyhdf5/PyHDF5.cpp
labesse40: Done

src/pygeosx/pysolver/PySolver.cpp
labesse40: Done.

src/pygeosx/pysolver/PySolver.cpp
corbett5: 10/10 solution.

src/pygeosx/pyhdf5/PyHDF5.cpp
labesse40: Actually when I try to call `PyModule_Create()` using the `PyObjectRef` for `PySolver` and `PyHistory` in `PySolver.cpp` and `PyHistory.cpp` (ex :   LvArray::python::PyObjectRef<> module{ PyModule_Create( &pysolvermodule ) }; ) 
This is causing a `core dumped` at execution.

corbett5: You have to be careful with how you do it, if I had to guess you're missing a call to `release` and the module is getting deleted when it shouldn't be. Have you pushed the breaking change?

labesse40: Yes I pushed it

labesse40: I think I should not `Py_XINCREF` in `pygeosx.cpp` since you're doing it `PyObjectRef` ?

corbett5: I don't see it in the changes, I don't even see `PyHDF5.cpp` or `HistoryCollection.cpp`.

labesse40: It's strange because I can see it on GitHub. I will push something now.

labesse40: I pushed something with the breaking change and the "fix" line commented. I also pushed a dummy example with the xml associated with it. It's all in the src/pygeosx directory.

src/pygeosx/pysolver/PySolver.cpp
labesse40: Done.

src/pygeosx/pysolver/PySolver.cpp
```diff
@@ -0,0 +1,226 @@
...
+  {
+    return nullptr;
+  }
+
+  if( !LvArray::python::addTypeToModule( module, geosx::python::getPySolverType(), "Solver" ) )
+  {
+    return nullptr;
+  }
+
+  return module;
```
corbett5, line 219: I'm still not sure why I can't see the time history file. But if you look at how I do it for the LvArray module it should be pretty similar.

https://github.com/GEOSX/LvArray/blob/3ad0795f10023bd9a2fd8a41f0de0caa376845c0/src/python/python.cpp#L162

Specifically you'll see that you return `module.release()`
https://github.com/GEOSX/LvArray/blob/3ad0795f10023bd9a2fd8a41f0de0caa376845c0/src/python/python.cpp#L134

corbett5: You removed `pyhdf5` in 2b6c10cfbe487f46053bc06d1152a063cf59f01f, I imagine you just forgot to add the renamed file.

labesse40: `return module.release()` instead of `return module`, that's exactly what I was missing. It's working now, thanks !

labesse40: Also you're right I forgot to add the new pyhistory directory in the commit... Sorry for that.

src/pygeosx/PyGroup.cpp
labesse40: Hi Ben, first I wish you a happy new year and hope you had good holidays ! 
I just pushed a commit that integrates this modification for creation of the different `PyStructure` (including the `PyHistory` one that I broke in 2 different structures being part of the new `PyHistory` `submodule` of `pygeosx` `module)`. We could easily imagine pure `Python` object taking those `PyStructures` to make it easier for users, and as you said, combined with @cssherman existing pure-python, makes it even more easy. 
Please let me know what you think of that new version. You also have a dummy test file in the `pygeosx` directory with `xml` associated with it if you want to see how it works.

src/pygeosx/pysolver/PySolver.cpp
labesse40: No more `static_cast` with the last commit.

src/pygeosx/pygeosx_AcousticTest.xml
```diff
@@ -0,0 +1,161 @@
...
+    maxTime="0.2">
+    <!-- trigger the application of the solver -->
+    <!-- control the timestepping here with forceDt -->
+    <!-- (there are other ways to control the time step size) -->
+
+    <PeriodicEvent
+      name="solverApplications"
+      forceDt="0.05"
+      target="/Solvers/acousticSolver"/>
+    <!-- generate an output that can be read from VTK -->
```
cssherman, line 62: Is this comment left over from something?  VTK outputs don't seem to be active with this file

src/pygeosx/pygeosx_AcousticTest.xml
```diff
@@ -0,0 +1,161 @@
...
+
+  <Events
+    maxTime="0.2">
+    <!-- trigger the application of the solver -->
+    <!-- control the timestepping here with forceDt -->
+    <!-- (there are other ways to control the time step size) -->
+
+    <PeriodicEvent
+      name="solverApplications"
+      forceDt="0.05"
```
cssherman, line 60: If you do want to respect the timestep requests from the solvers, then you should consider using maxEventDt instead

src/pygeosx/pygeosx_AcousticTest.xml
```diff
@@ -0,0 +1,161 @@
...
+      fieldName="pressure_n"
+      scale="0.0"/>
+
+    <FieldSpecification
+      name="initialPressure"
+      initialCondition="1"
+      setNames="{ all }"
+      objectPath="nodeManager"
+      fieldName="pressure_nm1"
+      scale="0.0"/>
```
cssherman, line 120: xml blocks should have unique names to avoid unexpected behavior during execution

src/pygeosx/pygeosx_AcousticTest.xml
labesse40: This xml is a copy from an other xml, and then I just modified the strict necessary for dummy tests, there might be some minor mistakes in it

src/pygeosx/pygeosx_AcousticTest.xml
labesse40: This actual timestep that is given in the xml is optionnal since we handle it from the python script. But the difference between forceDt and maxEventDt is good to know thanks.

src/pygeosx/pygeosx_AcousticTest.xml
labesse40: Also good to know thanks.

src/pygeosx/pygeosx_AcousticTest.xml
labesse40: Note that this XML is just to illustrate/test the modifications, it won't be part of the final version for the PR

src/pygeosx/PyGroup.cpp
corbett5: Thanks, you too!

I'm bit busy at the moment, but I'm looking forward to checking it out when I have some free time.

cssherman:
Are you planning to further specialize PySolver, PyHistoryOutput, and PyHistoryColelction?   Both of the history objects seem to be wrappers to execute (aside from their call), so why not have a single approach to call group.execute?

The PySolver.solverStep does call a separate method, but if `m_nonlinearSolverParameters.m_maxSubSteps=1`, then SolverBase.execute will effectively behave as if you called solverStep directly.  If you were to combine the additional methods defined for PySolver with a general group method, would that cause any issues?

labesse40:
What do you mean by specialize ? 
I imagined that we could have a pure python structure wrapping the 2 Pyhistory structures with its own methods that behind the scenes use the right pystructure with the right call. In that way we keep similar pystructure and stay general for their creation through createNewPyGroup(). We can always create pure python stucture to wrap them as we want.

I guess I can call execute on pysolver instead of solverStep.

cssherman:
By specialize, I meant whether there are going to be any additional C/C++ extension code for either object.  If the extension is only wrapping ExecutableGroup::Execute (and perhaps some other common Group methods), then we may as well be general here.  This would allow us to trigger other non-history/solver related processes from python as well.

cssherman:
And with regards to the name of the call, if all we are doing is running execute, I would rather have a common name (run, execute, etc.) rather than specialized names (collect, output, etc.)

labesse40:
For the `solvers,` there is also something to `reinitialize` parameters related to the solver. We will also add something to calculate `gradients,` `cost functions`, `residual`...
For `history` structures not sure for now, we may add stuff to have more control on data.
In that context I could template the `execute` call (like the `get_wrapper()` ). 
I gave a different name to be more explicit on what is happening instead of just "executing" something, I can change it to "execute".

src/coreComponents/dataRepository/Group.hpp
```diff
@@ -1283,6 +1284,13 @@ class Group
   integer getLogLevel() const { return m_logLevel; }
   ///@}
 
+  /**
+   * @brief Return PyGroup type.
+   * @return Return PyGroup type.
+   */
+  virtual PyTypeObject * getPythonType() const
```
corbett5, line 1291: I don't really like how `Group.hpp` depends on stuff in `pygeosx`, but I suspect this is the type of stuff we will talk about in the meeting.

Off the top of my head I have two solutions
1. Leave things where they are but move the implementation into the `cpp` file so that things that depend on `Group.hpp` (everything) don't depend on `pygeosx`.
2. Move all of they python wrapper classes out of `pygeosx` and next to the classes that they wrap. Then we can leave `pygeosx` to contain all the `C++` glue as well as whatever pure python stuff we want to build on top.

src/coreComponents/physicsSolvers/SolverBase.cpp
```diff
@@ -636,15 +636,15 @@ real64 SolverBase::explicitStep( real64 const & GEOSX_UNUSED_PARAM( time_n ),
                                  integer const GEOSX_UNUSED_PARAM( cycleNumber ),
                                  DomainPartition & GEOSX_UNUSED_PARAM( domain ) )
 {
-  GEOSX_ERROR( "SolverBase::ExplicitStep called!. Should be overridden." );
+  GEOSX_THROW( "SolverBase::ExplicitStep called!. Should be overridden.", std::runtime_error );
```
corbett5, line 639: Nice change!

src/pygeosx/PyGroup.cpp
```diff
@@ -46,6 +47,16 @@ struct PyGroup
...
 
+struct PySolver
+{
+  PyObject_HEAD
+
+  static constexpr char const * docString =
+    "A Python interface to geosx::SolverBase.";
+
+  geosx::SolverBase * group;
+};
```
corbett5, line 58: This seems a little backwards to me. I feel like `PyGroup.cpp` shouldn't have to include anything from `pysolver`. It also feels strange to define `PySolver` in this file.

src/pygeosx/pygeosx.cpp
```diff
@@ -353,6 +364,20 @@ PyInit_pygeosx()
     return nullptr;
   }
 
+  LvArray::python::PyObjectRef<> pysolverModule = geosx::python::PyInit_pysolver();
+  Py_XINCREF( pysolverModule );
```
corbett5, line 368: I think it would be clearer here if you got rid of the `Py_XINCREF` and replaced it with `pysolverModule.release()` in the next line, since we're transferring ownership of the module to the interpreter.

src/pygeosx/pygeosx.cpp
```diff
@@ -383,5 +408,9 @@ PyInit_pygeosx()
   }
 
   // Since we return module we don't want to decrease the reference count.
-  return module.release();
+  return module;
```
corbett5, line 411: Why this change?

src/pygeosx/pyhistory/PyHistory.cpp
```diff
@@ -0,0 +1,60 @@
...
+// Source includes
+#include "../pygeosx.hpp"
+#include "PyHistory.hpp"
+
+#include "PyHistoryCollectionType.hpp"
+#include "PyHistoryOutputType.hpp"
+
+namespace geosx
+{
+extern std::unique_ptr< GeosxState > g_state;
```
corbett5, line 13: Hmm, I'm really not a fan of `extern` variables. Could you not just call `getGlobalState` in this file instead of using `g_state` directly?

Also if we'll need to check if things are already initialized we should make a function in `pygeosx.hpp` instead of directly accessing the variable.

src/pygeosx/pyhistory/PyHistory.cpp
```diff
@@ -0,0 +1,60 @@
...
+  {
+    return nullptr;
+  }
+
+  if( !LvArray::python::addTypeToModule( module, geosx::python::getPyHistoryOutputType(), "HistoryOutput" ) )
+  {
+    return nullptr;
+  }
+
+  return module.release();
```
corbett5, line 56: Here however you do have the `release()`.

src/pygeosx/pysolver/PySolver.cpp
```diff
@@ -0,0 +1,169 @@
...
+};
+
+
+/**
+ * Initialize the module object for Python with the exported functions
+ */
+
+BEGIN_ALLOW_DESIGNATED_INITIALIZERS
+
+static PyTypeObject PySolverType = {
```
corbett5, line 118: It seems like a lot of work to go through to make this a module, especially since it only contains one type. I probably asked this before, but why not make all these new classes live in the `pygeosx` module directly?

src/pygeosx/testget.py
```diff
@@ -0,0 +1,42 @@
...
+
+collection = problem.get_group("/Tasks/waveFieldCollection")
+output = problem.get_group("Outputs/waveFieldOutput")
+
+pygeosx.apply_initial_conditions()
+time = 0
+dt = 0.005
+for i in range(100):
+    solver.solverStep(time, dt)
+    if i==999 :
```
corbett5, line 22: Doesn't `i` range from `0` to `99`?

src/pygeosx/PyGroup.cpp
labesse40: Yes that's actually something I noticed yesterday and removed. It was from a previous version.

src/pygeosx/testget.py
labesse40: Right ahah

src/pygeosx/pyhistory/PyHistory.cpp
labesse40: Yes I forgot to remove this in this file. Thanks for catching those mistakes...

src/pygeosx/pysolver/PySolver.cpp
labesse40: In the top of my head, I was seeing those stuff as `module` inside the `pygeosx` "machine". But you're right, it would be simpler to just make them classes in `pygeosx` module.

src/pygeosx/pygeosx.cpp
labesse40: Good question... That actually could explain those PyModule cleaning error at the end of my simulations !

src/pygeosx/pygeosx.cpp
labesse40: Changed it to be `Types` in the `pygeosx` module

src/coreComponents/dataRepository/Group.hpp
labesse40: As discussed during the meeting, I added a subdirectory `python` under the different class we are wrapping with python, and modified/created the `CMakeLists` associated (including the ones for the `unit_tests`). All circular references are now fixed.

labesse40:
When executing the code I keep getting the following message at the end of the simulation : 

```
corrupted size vs. prev_size in fastbins
[julien-Latitude-5520:645965] *** Process received signal ***
[julien-Latitude-5520:645965] Signal: Aborted (6)
[julien-Latitude-5520:645965] Associated errno: Unknown error 65537 (65537)
[julien-Latitude-5520:645965] Signal code:  (1548913968)
[julien-Latitude-5520:645965] [ 0] /lib/x86_64-linux-gnu/libc.so.6(+0x46210)[0x7f696b8af210]
[julien-Latitude-5520:645965] [ 1] /lib/x86_64-linux-gnu/libc.so.6(gsignal+0xcb)[0x7f696b8af18b]
[julien-Latitude-5520:645965] [ 2] /lib/x86_64-linux-gnu/libc.so.6(abort+0x12b)[0x7f696b88e859]
[julien-Latitude-5520:645965] [ 3] /lib/x86_64-linux-gnu/libc.so.6(+0x903ee)[0x7f696b8f93ee]
[julien-Latitude-5520:645965] [ 4] /lib/x86_64-linux-gnu/libc.so.6(+0x9847c)[0x7f696b90147c]
[julien-Latitude-5520:645965] [ 5] /lib/x86_64-linux-gnu/libc.so.6(+0x98c4c)[0x7f696b901c4c]
[julien-Latitude-5520:645965] [ 6] /lib/x86_64-linux-gnu/libc.so.6(+0x9a160)[0x7f696b903160]
[julien-Latitude-5520:645965] [ 7] /home/julien/codes/GEOSX/build-configUbuntu20-release/lib/PYGEOSX/lib/python3.8/site-packages/pygeosx.so(judy_close+0x1b)[0x7f696a2b0e0b]
[julien-Latitude-5520:645965] [ 8] /home/julien/codes/GEOSX/build-configUbuntu20-release/lib/PYGEOSX/lib/python3.8/site-packages/pygeosx.so(_ZN4chai12ArrayManagerD1Ev+0x80)[0x7f696a1fb6b0]
[julien-Latitude-5520:645965] [ 9] /lib/x86_64-linux-gnu/libc.so.6(+0x49a27)[0x7f696b8b2a27]
[julien-Latitude-5520:645965] [10] /lib/x86_64-linux-gnu/libc.so.6(on_exit+0x0)[0x7f696b8b2be0]
[julien-Latitude-5520:645965] [11] /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xfa)[0x7f696b8900ba]
[julien-Latitude-5520:645965] [12] python3(_start+0x2e)[0x55e8823ef09e]
[julien-Latitude-5520:645965] *** End of error message ***
Aborted (core dumped)
```

I can't figure out where it comes from. 
To reproduce this output, you can execute the `testget.py` under `src/pygeosx` using the following xml : `pygeosx_AcousticTest.xml` in the same directory.

labesse40:
> Very cool, super glad someone is improving this! It's probably just because I'm out of the loop but some examples of the new functionality would be nice too.

labesse40 (commented):
All changes have been done

corbett5 (approved):

corbett5 (approved):
Looks great!

src/coreComponents/dataRepository/Group.hpp
```diff
@@ -1329,6 +1330,19 @@ class Group
...
+   */
+  virtual void reinit() {}
+
+  /**
+   * @brief Return PyGroup type.
+   * @return Return PyGroup type.
+   */
+#if defined(GEOSX_USE_PYGEOSX)
+  virtual PyTypeObject * getPythonType() const;
+#endif
```
TotoGaz, line 1344: So I'm definitely late about this PR, but why do we add a `python` dependency here in `Group`?
From there, "_everything_" in `geosx` will depend on `python`.

I know that "_everything_" being a `Group`, it's convenient that `python` gets access to `Groups`.
- But it will eventually break encapsulation. And it's `geosx`'s internal implementation; I doubt a wrapper should access it.
- We should be able to compile `geosx`, with no `python` dependency, even if `ENABLE_PYGEOSX` is `ON`. This `Group::getPythonType` just kills it.

TotoGaz: @sframba

corbett5: >
> why do we add a python dependency here in Group
> 

It's already included from `View.hpp` https://github.com/GEOSX/GEOSX/blob/01f4b624504ec0ee915a6f457ef8cc09d054b168/src/coreComponents/dataRepository/wrapperHelpers.hpp#L37

But also if you actually follow the includes I'm pretty sure that we don't actually `#include <python.h>` in any headers. I think we got away with forward declarations only.

https://github.com/GEOSX/LvArray/blob/ab0d3025d2962aad10457b7bf7f5b6d266232276/src/python/pythonForwardDeclarations.hpp

https://github.com/GEOSX/LvArray/search?l=C%2B%2B&p=3&q=python.h

> But it will eventually break encapsulation. And it's `geosx`'s internal implementation; I doubt an external 
wrapper should access it.

I disagree. We need to be able to access and modify arbitrary data from Python doing it without exposing `Group` and `Wrapper` sounds convoluted. Not to mention that the data repository isn't really an implementation detail, it's mirrored in the XML files we currently use.


You could imagine a solution where instead of putting virtual functions in `Group` and `Wrapper` we had a massive type switching block inside `pygeosx` that would get the `typeid` of the object and then downcast and export it to Python. But I've been in favor of removing that code pattern from GEOSX for a long time.

TotoGaz: > But also if you actually follow the includes I'm pretty sure that we don't actually `#include <python.h>` in any headers. I think we got away with forward declarations only.

I’m not sure to understand entirely. 
If we need to forward declare, it’s because we need something.
So we depend on this something.
In this case we’ll depend on `PyTypeObject` which will be found in `python` (most probably during link edition?), (unless the name is misleading).

> I disagree. We need to be able to access and modify arbitrary data from Python doing it without exposing `Group` and `Wrapper` sounds convoluted. 

It depends what you call convoluted. This surely is a matter of costs and benefits w.r.t. the global goal.

> Not to mention that the data repository isn't really an implementation detail, it's mirrored in the XML files we currently use.

Yes, I know.
- But for the end users, the software architecture should be an implementation detail. The input file should make their life easy, and there is not strong reason that a C++ hierarchy is the most convenient for them to fill.
- And as a developer, the fact that I’m not able to _reorganize_ my code freely without potentially impacting the end user is not the most comfortable situation. To my eyes, this breaks encapsulation.

> You could imagine a solution where instead of putting virtual functions in Group and Wrapper we had a massive type switching block inside pygeosx that would get the typeid of the object and then downcast and export it to Python.

It could be a way to invert dependencies, which can be helpful. So while apparently tedious, maybe this kind of pattern could help us for tricky situations?

corbett5: > I’m not sure to understand entirely.
> If we need to forward declare, it’s because we need something.
> So we depend on this something.
> In this case we’ll depend on `PyTypeObject` which will be found in `python` (most probably during link edition?), (unless the name is misleading).

Yes it requires that we link the main geosx library against Python, but it does not require that we include the python headers. This seems like a pretty big difference to me. Take for example conduit, which is something that is directly included in Wrapper to make writing restarts possible. This is directly analogous to Python in that it exports data stored in any Wrapper to another environment. Except that for conduit we need not only need to link but compile against the headers as well. As far as I can tell this hasn't produced any problems for GEOSX. Conduit is a third party library and if all of GEOSX depends on it then so be it. It's not even a question of compile time either because 99% of the time is spent compiling the CUDA kernels.

> It could be a way to invert dependencies, which can be helpful. So while apparently tedious, maybe this kind of pattern could help us for tricky situations?

This pattern is a pain in the ass to work with. It makes already tedious compilation errors 40 times harder to deal with, it leads to code that is hard to read, it is hard to maintain and extend. I also don't think this situation is particularly tricky. When building pygeosx we add Python as another top level dependency. If you don't want this then don't build with Python.

TotoGaz: > Yes it requires that we link the main geosx library against Python, but it does not require that we include the python headers. This seems like a pretty big difference to me.

Can you elaborate on this difference?
I’m more on the “_you need it, whatever the form_“ side. 

> Take for example conduit, which is something that is directly included in Wrapper to make writing restarts possible. This is directly analogous to Python in that it exports data stored in any Wrapper to another environment.

`conduit` looks mainly like a smart serializer/deserializer. It helps at exchanging data.
Using `python`, we could exchange processes and implementations too.

> As far as I can tell this hasn't produced any problems for GEOSX.

Actually, having those implementations exposed in the heart of the whole code base indeed produces problems.
Because it binds our questions to one unique solution.
It is imposed as the source of information, so it’s impossible to get a different one.

An `Event`, a `MeshGenerator` or a customized `Output` could be implemented in another language (e.g. `python`) because it could be more convenient.
It’s not possible because we have an implementation centric architecture.

And therefore whatever I want to do in GEOSX I have to deal with `conduit` of `xml` or `...`, even if that should be out of scope.
How can I mock the inputs of my `VTK` code and validate that this key feature is rock solid? I cannot, so there will be more bugs.

corbett5: > Can you elaborate on this difference?
> I’m more on the “_you need it, whatever the form_“ side.

If you try and call some Python function in a physics solver it will produce a compilation error because the header wasn't included. And your IDE doesn't get clogged up.

> Actually, having those implementations exposed in the heart of the whole code base indeed produces problems.
> Because it binds our questions to one unique solution.
> It is imposed as the source of information, so it’s impossible to get a different one.

How does optionally linking to Python impose anything as a source for information?

>An `Event`, a `MeshGenerator` or a customized `Output` could be implemented in another language (e.g. `python`)  because it could be more convenient.
> It’s not possible because we have an implementation centric architecture.

Yes one can infer from the dependencies and source of data repository that we write our restarts with Conduit. But I don't see how that would make it any harder to create an alternate restart implementation that used something else entirely or was perhaps in a different language.

> And therefore whatever I want to do in GEOSX I have to deal with conduit of xml or ..., even if that should be out of scope.

This sounds like it would involve changes way beyond the scope of this PR, basically a wholesale rewrite of the data repository. Again though I don't see the value. I can't speak for others, but when I'm working on something for a physics solver, I never even think about Conduit or any of the other TPL's that I'm not actively using.

> How can I mock the inputs of my VTK code and validate that this key feature is rock solid? I cannot, so there will be more bugs.

Can you be specific as to what it is that you'd like to do, and how having GEOSX link to Python prevents you from doing it?

TotoGaz: Hi @corbett5 sorry for the delay and thank you for the fruitful discussion, I appreciate! 😉 

> If you try and call some Python function in a physics solver it will produce a compilation error because the header wasn't included. And your IDE doesn't get clogged up.

So in a nutshell, you trim to the maximum what can and cannot be used.
If anyone wants more, they will have to explicitly add the `include` (or the forward declaration) to make it work.
This makes sense, meanwhile I personally prefer to be extremely strict in the build: when a dependency is added to the `CMakeLists.txt` it's really easy to detect.

For the rest of the discussion, let's consider the following simplified example.
I have a `Solver` which generates a `ComplicatedOutput` that I want to `WriteOnDisk`.
This operation being critical for my users, I want to perform extensive testing.
Today in `GEOSX` I may have to create a lot of simulations, such that all my `Solvers` will be able to create a lot of different `ComplicatedOutput` and I'll test `WriteOnDisk` with all these data coming from real workflows.
I have to abide by the `GEOSX` organization (`xml`, `conduit`, running the simulation on multiple ranks, etc.).

But instead, I'd like to be able to implement "from scratch" my own versions of `ComplicatedOutput` dedicated to testing.
I could be able to generate batches of cases without running any simulation (therefore in the unit tests and not in the integrated tests).
I could even be able to generate "_wrong_" data or dangerous corner cases, simulate a `NaN` quite conveniently and see how my `WriteOnDisk` behaves.

This is what I mean by "_different source of data_": I'm bound to the current organization and implementation and it's complicated to escape. It's not `xml` or `conduit` themselves, but more about our implementation centric architecture.

> This sounds like it would involve changes way beyond the scope of this PR

Absolutely, this is where the discussion takes place, but not where actions (if any) would happen.

> I can't speak for others, but when I'm working on something for a physics solver, I never even think about Conduit or any of the other TPL's that I'm not actively using.

Absolutely, when you are working in `GEOSX` _à la_ `GEOSX`, there is no limitation and things flow seamlessly.
But things may be a little different when you try to plug other components for whatever reason (I mentioned testing _e.g._).

corbett5: > I could be able to generate batches of cases without running any simulation (therefore in the unit tests and not in the integrated tests)

I certainly agree this would be super useful and God knows we're really lacking on the unit test front for a lot of our capabilities. But I don't really see how depending on `conduit` or `xml` or `Python` really complicates things here. We should have some generic unit test helpers that can create a bunch of bare bones GEOSX data structures (basically a mesh) and then it would be easy enough for your unit tests to create and populate some dummy fields that could then be written out. Some of the physics solver tests basically do this. Or if you write `WriteOnDisk` in a modular enough way then maybe you could just manufacture the test data without the Data Repository.

TotoGaz: > But I don't really see how depending on conduit or xml or Python really complicates things here. We should have some generic unit test helpers that can create a bunch of bare bones GEOSX data structures (basically a mesh) and then it would be easy enough for your unit tests to create and populate some dummy fields that could then be written out.

The main idea I have in mind, in a nutshell, is that because of the current construction (not specially `xml` or `conduit`, it's a bit more general), one surely has to implement more than the shallow layer that could be required.

Like you said, if I want to test my `WriteOnDisk` for all kinds of inputs, I'd like to be able to do so without having to create any of the `GEOSX` infrastructure: I just want a `ComplicatedOutput`. If I just want the skins, I do not want the bare bones! 😉 And today in `GEOSX` this is not straightforward.
Consider the `testTriaxial` example. This test looks simple, but we nevertheless have to deal with `FunctionManager`, `ConstitutiveManager`, `xml` input file, `Group` and `conduit::node`, `GeosxState`. It's more a mini `GEOSX` simulation than a unit test: it's complicated to test at the "_module_" level.

I'll be in LLNL on a regular basis now, so I'll hope we can find some time to interact on concrete cases, that will be more efficient!

corbett5: Yeah I'd be down to talk about what we could do to make unit testing more accessible. Unfortunately though I'm now living in New Mexico so...

TotoGaz: I'm working on "propositions" about this, I'll be doing a presentation with (hopefully) concrete examples.
It's with great pleasure that I would get you involved!

PR #1711, opened 2021-12-15
PVTDriver

joshua-white:
This PR creates a PVTDriver (comparable to the existing TriaxialDriver) that provides an easy mechanism to test out PVT (and other fluid property) behavior for a given set of pressure/temperature conditions.  The code will work for any fluid model that conforms to `MultiFluidBase`.  This driver also provides a convenient mechanism for unit testing PVT models.

Sphinx documentation can be live viewed here:
https://geosx-geosx.readthedocs-hosted.com/en/feature-white-compositionaldriver/coreComponents/constitutive/docs/PVTDriver.html#id7


src/coreComponents/constitutive/fluid/PVTDriver.cpp
```diff
@@ -0,0 +1,332 @@
...
+  {
+    compositionValues[0][i] /= sum;
+  }
+
+  arraySlice1d< real64 const, compflow::USD_COMP - 1 > const composition = compositionValues[0];
+
+  // perform fluid update using table (P,T) and save resulting total density, etc.
+  // note: column indexing should be kept consistent with output file header below.
+
+  forAll< parallelDevicePolicy<> >( 1, [=]  GEOSX_HOST_DEVICE ( integer const ei )
```
francoishamon, line 229:
```suggestion 
  using ExecPolicy = typename FLUID_TYPE::exec_policy;
  forAll< ExecPolicy >( 1, [=]  GEOSX_HOST_DEVICE ( integer const ei )
```
This is what we do in the flow solvers since PVTPackage is not ready for the GPU, and its `compute` function is disabled if used with `parallelDevicePolicy` on a GPU platform (see [here](https://github.com/GEOSX/GEOSX/blob/0ebaf87e3f7a4b171e1b0e03d5213d75c91aa31d/src/coreComponents/constitutive/fluid/CompositionalMultiphaseFluid.hpp#L163-L164)). So using `parallelDevicePolicy` here may result in a failed unit test on Lassen if `CompositionalMultiphaseFluid` is selected (to be confirmed).

francoishamon (approved):
Looks great!

PR #1713, opened 2021-12-15
Fixed Pamela parallel properties import (inside GEOSX)

francoishamon:
If we want to (partially) fix the parallel import of properties using PAMELA, the call to `synchronizeFields` that was added to VTK is also needed in `PAMELAMeshGenerator.cpp`. 

There are other fixes needed in the PAMELA submodule to allow for the parallel import of properties, but they are hacks that only work in the one-region case (what I need for the test case I am running now), so I have been reluctant to merging them and always used GEOSX tables instead.


klevzoff (approved):
Thanks!

I guess I never tested the parallel case. Which reminds me that we need an integrated test with PAMELA (not in this PR :slightly_smiling_face: )

PR #1717, opened 2021-12-21
VTK output for multiple mesh bodies

povolny1:
**Only looking to merge in VTK output changes:**
- src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp
- src/coreComponents/fileIO/vtk/VTKVTMWriter.cpp
- src/coreComponents/fileIO/vtk/VTKVTMWriter.hpp

Please ignore other changes.


src/coreComponents/mesh/ParticleRegion.hpp
```diff
@@ -0,0 +1,124 @@
...
+
+  /**
+   * @brief Deleted default constructor.
+   */
+  ParticleRegion() = delete;
+
+  /**
+   * @brief Destructor.
+   */
+  virtual ~ParticleRegion() override;
```
TotoGaz, line 63: Use `= default;` instead.

src/coreComponents/mesh/ParticleRegion.hpp
```diff
@@ -0,0 +1,124 @@
...
+  /**
+   * @brief Add a cellBlockRegion name to the list.
+   * @param cellBlockName string containing the cell block region name.
+   */
+  void addCellBlockName( string const & cellBlockName )
+  {
+    m_cellBlockNames.emplace_back( cellBlockName );
+  }
+
+  virtual void generateMesh( Group & cellBlocks ) override;
```
TotoGaz, line 99: For info, you do not need both `virtual` and `override`.

src/coreComponents/fileIO/vtk/VTKVTMWriter.cpp
```diff
@@ -59,11 +59,18 @@ void VTKVTMWriter::addSubBlock( string const & blockName, string const & subBloc
...
 
-void VTKVTMWriter::addDataToSubBlock( string const & blockName, string const & subBlockName, string const & filePath, int mpiRank ) const
+void VTKVTMWriter::addSubSubBlock( string const & blockName, string const & subBlockName, string const & subSubBlockName ) const
 {
-  auto blockNode = m_vtmFile.child( "VTKFile" ).child( "vtkMultiBlockDataSet" ).find_child_by_attribute( "Block", "name", blockName.c_str() );
-  auto subBlockNode = blockNode.find_child_by_attribute( "Block", "name", subBlockName.c_str() );
-  auto dataNode = subBlockNode.append_child( "DataSet" );
+  auto subBlockNode = m_vtmFile.child( "VTKFile" ).child( "vtkMultiBlockDataSet" ).find_child_by_attribute( "Block", "name", blockName.c_str() ).find_child_by_attribute( "Block", "name", subBlockName.c_str() );
+  auto subSubBlockNode = subBlockNode.append_child( "Block" );
+  subSubBlockNode.append_attribute( "name" ) = (blockName + "_" + subSubBlockName).c_str();
```
TotoGaz, line 66: The same value `blockName + "_" + subSubBlockName` is used in both `addSubSubBlock` and `addDataToSubSubBlock`.
If it's supposed to be the same, could you build a unique function that builds the name from `blockName` and `subSubBlockName`.

src/coreComponents/fileIO/vtk/VTKVTMWriter.hpp
```diff
@@ -42,29 +42,40 @@ class VTKVTMWriter
...
    * @details The final level : paths to the vtu file per rank
-   * @param[in] blockName Name of the parent block
-   * @param[in] subBlockName Name of the subBlock (usually the name of the Region)
+   * @param[in] subBlockName Name of the parent block (will be one of the ElementRegion types)
+   * @param[in] subSubBlockName Name of the subSubBlock (usually the name of the Region)
    * @param[in] filePath path to the vtu file containing the unstructured mesh
    * @param[in] mpiRank the mpi rank.
    */
-  void addDataToSubBlock( string const & blockName, string const & subBlockName, string const & filePath, int mpiRank ) const;
+  void addDataToSubSubBlock( string const & blockName, string const & subBlockName, string const & subSubBlockName, string const & filePath, int mpiRank ) const;
```
TotoGaz, line 76: Your doxygen parameters are out of sync

src/coreComponents/mesh/ParticleRegion.cpp
```diff
@@ -0,0 +1,51 @@
...
+//{
+//  Group & ParticleSubRegions = this->getGroup( viewKeyStruct::particleSubRegions() );
+//
+//  for( string const & cellBlockName : this->m_cellBlockNames )
+//  {
+//    ParticleSubRegion & subRegion = particleSubRegions.registerGroup< ParticleSubRegion >( cellBlockName );
+//    CellBlock & source = cellBlocks.getGroup< CellBlock >( subRegion.getName() );
+//    subRegion.copyFromCellBlock( source );
+//  }
+//}
```
TotoGaz, line 47: Please remove commented code.

src/coreComponents/fileIO/vtk/VTKVTMWriter.cpp
```diff
@@ -59,11 +59,18 @@ void VTKVTMWriter::addSubBlock( string const & blockName, string const & subBloc
...
+}
+
+void VTKVTMWriter::addDataToSubSubBlock( string const & blockName, string const & subBlockName, string const & subSubBlockName, string const & filePath, int mpiRank ) const
+{
+  auto subBlockNode = m_vtmFile.child( "VTKFile" ).child( "vtkMultiBlockDataSet" ).find_child_by_attribute( "Block", "name", blockName.c_str() ).find_child_by_attribute( "Block", "name", subBlockName.c_str() );
```
TotoGaz, line 71: You could refactor with some `getOrCreateSubSubBlock`?
`vtmWriter.addSubSubBlock( meshBodyName, region.getCatalogName(), region.getName() );` would therefore be useless.

src/coreComponents/mesh/ParticleRegion.hpp
```diff
@@ -0,0 +1,124 @@
...
+
+  /**
+   * @brief A struct to serve as a container for variable strings and keys.
+   * @struct viewKeyStruct
+   */
+  struct viewKeyStruct : public ElementRegionBase::viewKeyStruct
+  {
+    /// @return String key for the cell block names
+    static constexpr char const * sourceCellBlockNamesString() {return "cellBlocks"; }
+  };
```
TotoGaz, line 112: Do you need this to be 
* `public`
* inherit from `ElementRegionBase::viewKeyStruct`

?

src/coreComponents/mesh/ParticleRegion.hpp
```diff
@@ -0,0 +1,124 @@
...
+
+#ifndef GEOSX_MESH_PARTICLEREGION_HPP_
+#define GEOSX_MESH_PARTICLEREGION_HPP_
+
+#include "ElementRegionBase.hpp"
+
+namespace geosx
+{
+class EdgeManager;
+class EmbeddedSurfaceGenerator;
```
TotoGaz, line 28: Why do you need this?

src/coreComponents/fileIO/vtk/VTKVTMWriter.cpp
povolny1: I'm not sure I understand - are you suggesting to merge those two functions into a single function?

src/coreComponents/fileIO/vtk/VTKVTMWriter.hpp
povolny1: Fixed

povolny1:
I'm realizing I have too many incomplete things going on in this branch and it will complicate merging later, so I'm going to make a new PR based on a clean branch off of the most current version of the multibody branch containing only the changes to VTK output. Thank you @TotoGaz for your suggestions on this PR, I will carry them over to the new PR and retract this one.

src/coreComponents/fileIO/vtk/VTKVTMWriter.cpp
TotoGaz: The core issue is that your code relies on the naming pattern of your sub-block to be `blockName + "_" + subSubBlockName`.
So this pattern generation must be unique (if you change it, you change it everywhere, and the code still works).

In your current implementation, you can change the first `_` to `-`, forgetting to modify the second by design. It will compile but the code won't work anymore.

* A solution is to centralize this naming pattern generation in its own function.
* Another solution could be (i'm not 100% sure) to make one unique function `getOrCreateSubSubBlock` that could help the factorisation. I think that the naming pattern generation would be used only once in this function and so you would not need an extra function just for this. Up to you 😉

src/coreComponents/mesh/ParticleSubRegion.hpp
```diff
@@ -0,0 +1,299 @@
...
+  map< localIndex, array1d< globalIndex > > m_unmappedGlobalIndicesInEdgelist;
+
+  /// Map of unmapped global indices in the element-to-face map
+  map< localIndex, array1d< globalIndex > > m_unmappedGlobalIndicesInFacelist;
+
+  /// List of the fractured elements on this rank
+  SortedArray< localIndex > m_fracturedCells;
+
+  /// Map from local Cell Elements to Embedded Surfaces
+  EmbSurfMapType m_toEmbeddedSurfaces;
```
CusiniM, line 277: Do you really need these for the `ParticleSubRegion`? Are there going to be fractured particles? I am not sure I can picture this.  `m_toEmbeddedSurfaces` is a list of the embeddedSurface elements inside each cellblock. I am not sure it has a meaning for particles.

src/coreComponents/mesh/ParticleRegion.hpp
CusiniM: I think that, in general, we do want the `viewKeyStruct` to be public. We don't have a getter for it and it is often used to get the keys inside solvers. This is a region though so I am not 100% sure it is actually useful.

src/coreComponents/mesh/ParticleSubRegion.cpp
```diff
@@ -0,0 +1,270 @@
...
+  registerWrapper( viewKeyStruct::constitutivePointVolumeFractionString(), &m_constitutivePointVolumeFraction );
+
+  registerWrapper( viewKeyStruct::dNdXString(), &m_dNdX ).setSizedFromParent( 1 ).reference().resizeDimension< 3 >( 3 );
+
+  registerWrapper( viewKeyStruct::detJString(), &m_detJ ).setSizedFromParent( 1 ).reference();
+
+  registerWrapper( viewKeyStruct::toEmbSurfString(), &m_toEmbeddedSurfaces ).setSizedFromParent( 1 );
+
+  registerWrapper( viewKeyStruct::fracturedCellsString(), &m_fracturedCells ).setSizedFromParent( 1 );
+}
```
CusiniM, line 41: I am not sure all these wrappers make sense for this subRegion type.

src/coreComponents/mesh/ParticleRegion.hpp
TotoGaz: In this case, `viewKeyStruct::sourceCellBlockNamesString` was never used outside of the `ParticleRegion` class.
Exposing it `public`ly would in the end make the `cellBlocks` instances `public` too, while they are used as a `private` members in the present case.
If we have to keys in `viewKeyStruct`, one for a `public` attribute, the other for a `private` attribute, then the `private` attribute would become `public` even though it's not the desired pattern.

Having `viewKeyStruct ` `public` actually means that any registered `Wrapper`/`Group` could be accessed by anyone. This can beak encapsulation sometimes (like this time).

PR #1723, opened 2022-01-07
Added validation example for Problem 1.2 of Class (2009)

francoishamon:
This PR adds a validation case for the second problem (Problem 1.2) proposed in the benchmark paper: "A benchmark study on problems related to CO 2 storage in geologic formations", Class et al, 2009.

I got the following result, which is the best I could do given the differences in the other codes' results and the information provided in the paper.

match_problem_1_2

This PR requires some code fixes in the CO2-brine fluid package that will be done in a separate PR.

This PR will require a rebaseline to add a smoke test based on a coarse version of the problem. Rebaseline done in https://github.com/GEOSX/integratedTests/pull/274.

PR #1729, opened 2022-01-11
Improved the Newton solver for solubility and density in CO2-brine fluid

francoishamon:
This PR adds some improvements to the local Newton solver used in the CO2-brine fluid package to compute the CO2 solubility and density. Previously, the local solver would not converge below about 300 K, which was problematic for some use cases (PR #1723 and surface conditions in wells). 

I unified the local solver code (previously duplicated) and added some simple safeguards (backtrackings, chopping, etc) to make sure it converges in the full (pressure, temperature) range. This is checked in `testCO2BrinePVTModels`. 

The CO2-brine fluid package would benefit from more refactoring (remove hard-coded constants, unify some functions, etc) but this can be done in a separate PR.

This PR does not need a rebaseline.

### DIFF
--- a/src/coreComponents/constitutive/fluid/PVTFunctions/CO2Solubility.cpp
+++ b/src/coreComponents/constitutive/fluid/PVTFunctions/CO2Solubility.cpp
@@ -18,6 +18,7 @@
 
 #include "constitutive/fluid/PVTFunctions/CO2Solubility.hpp"
 
+#include "constitutive/fluid/PVTFunctions/CO2EOSSolver.hpp"
 #include "constitutive/fluid/PVTFunctions/PVTFunctionHelpers.hpp"
 #include "functions/FunctionManager.hpp"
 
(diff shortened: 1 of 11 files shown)


src/coreComponents/constitutive/fluid/PVTFunctions/CO2Solubility.cpp
```diff
@@ -125,40 +126,42 @@ real64 Par( real64 const & T, real64 const & P, real64 const * cc )
...
-                  "CO2Solubility NR convergence fails! " << "dre = " << dre << ", tolerance = " << tolerance,
-                  InputError );
-  return V_r;
+  // compute the initial guess for Newton's method
+  real64 const initialReducedVolume = 0.75*Rgas*(T_K_f+T)/(P*P_Pa_f)*(1/V_c);
+
+  // define the local solver parameters
+  // for now, this is hard-coded, but we may want to let the user access the parameters at some point
+  integer const maxNumNewtonIter = 500;
+  integer const maxNumBacktrackIter = 20;
```
joshua-white, line 141: This may be an extreme number of backtracks (0.5^20 ~ 1e-6 ~ zero update step).  It may be better to quit at 5-10 and just hope the next Newton step takes it to a better starting point.  Perhaps you can test on the challenging example to see if 20 are needed?

joshua-white (approved):
Looks good.

src/coreComponents/constitutive/fluid/PVTFunctions/CO2Solubility.cpp
francoishamon: You are right, thanks for catching this, that was too much backtracking. I reduced this number to 8 steps (max) and will give the user access to these values in a future PR.

PR #1732, opened 2022-01-14
Implemented thermal conductivity models

francoishamon:
This PR adds thermal conductivity constitutive models that I would like to use in the flux term of the thermal compositional solver here:
https://github.com/GEOSX/GEOSX/blob/d33a78f0c7003c590e5fa7e5002c7a877a97238a/src/coreComponents/physicsSolvers/fluidFlow/ThermalCompositionalMultiphaseFVMKernels.hpp#L334-L336
by doing something like:
```
m_stencilWrapper.computeWeights( iconn,
                                 m_thermalConductivity,
                                 m_dThermalConductivity_dPhaseVolFrac,
                                 stack.thermalTransmissibility,
                                 stack.dThermalTransmissibility_dPhaseVolFrac );
```
This call is not yet implemented and will require some small changes in the stencil code done in a separate PR.

In this PR I implemented two simple conductivity models, that, I think, are the most often used: 
- [X] constant conductivity
- [X] volume-weighted conductivity

The PR is ready for review and will require a rebaseline done in https://github.com/GEOSX/integratedTests/pull/183


src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.hpp
```diff
@@ -223,6 +231,8 @@ class CompositionalMultiphaseBase : public FlowSolverBase
 
     static constexpr char const * capPressureNamesString() { return "capPressureNames"; }
 
+    static constexpr char const * thermalConductivityNamesString() { return "thermalConductivityNames"; }
+
```
CusiniM, line 235: Keep in mind that we are trying to remove the need to specify model names in the solvers. Although it's all done in #1682 so I guess this is fine for now.

CusiniM (approved):
I think this is fine. My only concer/comment is that the way it is implemented here thermal conductivity is sort of a property of the saturated porous medium as a whole. I wonder if it would make sense to have 2 separate properties, one for the fluid and one for the rock, and then do the volume averaging in the assembly kernel. It really depends on how complex of a model we want for the conductivity.

francoishamon:
> I think this is fine. My only concern/comment is that the way it is implemented here thermal conductivity is sort of a property of the saturated porous medium as a whole. I wonder if it would make sense to have 2 separate properties, one for the fluid and one for the rock, and then do the volume averaging in the assembly kernel. It really depends on how complex of a model we want for the conductivity.

Yes, great point! It is true that the logic implemented in this PR breaks the logic of some other models by considering fluid and rock properties as a whole. 

I struggled with the that in some conductivity models the "rock-part of conductivity" and the "fluid-part of conductivity" are not always blended in an additive fashion, which makes it hard to break conductivity as two separate properties. 

For instance in [Coats (1980)](https://onepetro.org/spejournal/article-abstract/20/06/533/121944), the fluid part is just a multiplier for the rock conductivity. In [Krupiczka (1967) ](https://scholar.google.com/scholar?lookup=0&q=Krupiczka+Analysis+of+thermal+conductivity+in+granular+materials&hl=en&as_sdt=0,5) it is really messy, you have to solve a nonlinear equation that involves both fluid and rock conductivity. I don't know if we will ever need these models (I hope we won't to be honest), but since they are both implemented in Eclipse/IX, we never know.

For now I propose to keep this organization because it is simple, but if at some point another solver needs a conductivity parameter for the rock only, then I will revisit and see if we can break the model into two pieces.

src/coreComponents/constitutive/thermalConductivity/ConstantThermalConductivity.cpp
```diff
@@ -0,0 +1,72 @@
...
+}
+
+std::unique_ptr< ConstitutiveBase >
+ConstantThermalConductivity::deliverClone( string const & name,
+                                           Group * const parent ) const
+{
+  return ThermalConductivityBase::deliverClone( name, parent );
+}
+
+void ConstantThermalConductivity::allocateConstitutiveData( dataRepository::Group & parent,
```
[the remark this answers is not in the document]
francoishamon: I will investigate, it may be possible to do as you suggest.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.hpp
francoishamon: Yes, you are right, I kept the approach used in the current `develop`, I will update this depending on who merges first.

francoishamon:
Following a discussion with @CusiniM , the latest changes in this PR are:
- Thermal conductivity is a diagonal tensor
- Porosity and saturation dependences are treated explicitly for now

This will make it easier to use the existing stencil wrapper in the flux kernel

PR #1734, opened 2022-01-18
Implemented J-function capillary pressure

francoishamon:
This PR implements the Leverett J-function capillary pressure model, as requested by one of our users. 

In this model, the Leverett J-function is read from a table (J vs saturation). During the simulation, capillary pressure is obtained by interpolating in the J-function table, and then scaling it by a multiplier that explicitly depends on surface tension, (cell-wise) porosity, and (cell-wise) permeability. The converged porosity and permeability are used at the end of each time step to recompute the J-function multiplier.

The PR is ready for review:
- [X] Unit test 
- [X] Integrated test

Rebaseline done in https://github.com/GEOSX/integratedTests/pull/181


src/coreComponents/constitutive/capillaryPressure/TableCapillaryPressure.cpp
```diff
@@ -84,11 +85,12 @@ void TableCapillaryPressure::postProcessInput()
...
   {
     GEOSX_THROW_IF( m_wettingIntermediateCapPresTableName.empty() || m_nonWettingIntermediateCapPresTableName.empty(),
-                    GEOSX_FMT(
-                      "{}: for a three-phase flow simulation, we must use {} to specify the capillary pressure table for the pair (wetting phase, intermediate phase), and {} to specify the capillary pressure table for the pair (non-wetting phase, intermediate phase)",
-                      getFullName(),
-                      viewKeyStruct::wettingIntermediateCapPresTableNameString(),
-                      viewKeyStruct::nonWettingIntermediateCapPresTableNameString()  ),
+                    GEOSX_FMT( string( "{}: for a three-phase flow simulation, we must use {} to specify the capillary pressure table " ) +
+                               string( "for the pair (wetting phase, intermediate phase), and {} to specify the capillary pressure table " ) +
+                               string( "for the pair (non-wetting phase, intermediate phase)" ),
```
klevzoff, line 90: No need to break format string literal apart, you should be able to format it like this (I hope uncrustify doesn't mess it up): 
```suggestion
                    GEOSX_FMT( "{}: for a three-phase flow simulation, we must use {} to specify the capillary pressure table "
                               "for the pair (wetting phase, intermediate phase), and {} to specify the capillary pressure table "
                               "for the pair (non-wetting phase, intermediate phase)" ,
```

klevzoff (approved):
Looks great!

src/coreComponents/constitutive/capillaryPressure/TableCapillaryPressure.cpp
francoishamon: I modified these error messages as you suggested.

PR #1735, opened 2022-01-19
Implemented output of reservoir statistics

### LINKED ISSUE #1968: [Feature] Proposal for output of aggregate statistics from GEOSX
Very often over the last months, we have been asked to output "aggregate statistics" that require some (simple or not so simple) processing of the solver variables such as:
- Min, max, pore-volume averaged pressure per region
- Min, max, volume averaged temperature per region
- Total pore volume of a region
- Phase pore volume per region
- Mobile phase mass per region
- Immobile (trapped) phase mass per region
- Dissolved component mass per region
- Min, max displacement in each direction
- Min, max diagonal component of effective stress in each direction
- Phase CFL numbers
- Component CFL numbers
- Cumulative linear iterations
- Cumulative nonlinear iterations

Some of these aggregate statistics can be difficult to compute as a post-processing step outside GEOSX because:
- They would require a lot of data output to file (7 arrays to compute dissolved component mass)
- They may require data that is difficult to access from outside (i.e., residual gas saturation to compute trapped phase mass)
- Many users don't know how to compute these aggregate values from the GEOSX compositional variables (i.e., dissolved component mass)

So it may be easier to do this processing inside the code and just output the aggregate (scalar) statistics, I think.

To avoid polluting the solver code with the computation of the statistics, it would be nice to compute them in separate physics-dependent `Tasks` that can be called only when needed, at the frequency needed by the users. These `Tasks` are responsible for the registration of the data and drive the statistics kernel launches. Additionally, when `TimeHistory` can output scalars, we will use it to output the aggregate statistics (currently, we just output to terminal).

In PR #1735 is the code that we have used up to now (Northern Lights, etc)
It implements three `Tasks` computing most of the statistics mentioned above:
- `CompositionalMultiphaseStatistics`
- `SinglePhaseStatistics`
- `SolidMechanicsStatistics`

When we have a good pre-processor (to automatically setup all the `TimeHistory` xml tags to output the appropriate variables) and post-processor (to compute the aggregates) for GEOSX, we may re-evaluate all this. It will be easy to just remove the `Tasks` if they become useless.

francoishamon:
This PR adds some output for quantities that we have been asked to report for comparison and analysis purposes. These quantities are computed for each `ElementRegions` and can be accessed using `TimeHistory`.

The PR is ready for review, and will require a rebaseline. Rebaseline done in https://github.com/GEOSX/integratedTests/pull/243

Resolves #1968


src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -859,8 +868,161 @@ void CompositionalMultiphaseBase::initializePostInitialConditionsPreSubGroups()
...
+  for( integer ip = 0; ip < m_numPhases; ++ip )
+  {
+    m_phasePoreVolume[ip] = MpiWrapper::sum( m_phasePoreVolume[ip] );
+    m_totalPoreVolume += m_phasePoreVolume[ip];
+  }
+
+  m_averagePressure /= totalUncompactedPoreVolume;
+  m_averageTemperature /= totalUncompactedPoreVolume;
+
+  if( outputToTerminal )
```
TotoGaz, line 1015: Are you computing all the above even if you do not display?
Are you using the information somewhere else?
If  possible, I would make this computation iff `outputToTerminal` is `true`.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -106,6 +100,18 @@ CompositionalMultiphaseBase::CompositionalMultiphaseBase( const string & name,
...
+
+  this->registerWrapper( viewKeyStruct::averagePressureString(), &m_averagePressure );
+  this->registerWrapper( viewKeyStruct::minimumPressureString(), &m_minimumPressure );
+  this->registerWrapper( viewKeyStruct::maximumPressureString(), &m_maximumPressure );
+  this->registerWrapper( viewKeyStruct::averageTemperatureString(), &m_averageTemperature );
+  this->registerWrapper( viewKeyStruct::minimumTemperatureString(), &m_minimumTemperature );
+  this->registerWrapper( viewKeyStruct::maximumTemperatureString(), &m_maximumTemperature );
+  this->registerWrapper( viewKeyStruct::totalPoreVolumeString(), &m_totalPoreVolume );
+  this->registerWrapper( viewKeyStruct::phasePoreVolumeString(), &m_phasePoreVolume ).
+    setSizedFromParent( 0 );
```
TotoGaz, line 113: You are attaching/registering/giving a name to 8 member to your solver just to display some convergence information. That sounds heavy.

I would preferably build a `struct` with the proper fields
```c++
struct{
  real64 m_averagePressure;
  real64 m_minimumPressure;
  real64 m_maximumPressure;
  real64 m_averageTemperature;
  real64 m_minimumTemperature;
  real64 m_maximumTemperature;
  real64 m_totalPoreVolume;
  array1d< real64 > m_phasePoreVolume;
}
```
and get your `computeReservoirStatistics` build and return it. Then you could display it in your `computeStatistics`.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -859,8 +868,161 @@ void CompositionalMultiphaseBase::initializePostInitialConditionsPreSubGroups()
...
+  if( outputStatisticsToScreen && m_nonlinearSolverParameters.m_totalSuccessfulNewtonNumIterations > 0 )
+  {
+    GEOSX_LOG_LEVEL_RANK_0( 1, getName()
+                            << ": Total number of time steps = " << cycleNumber+1
+                            << ", successful nonlinear iterations = " << m_nonlinearSolverParameters.m_totalSuccessfulNewtonNumIterations
+                            << ", wasted nonlinear iterations = " << m_nonlinearSolverParameters.m_totalWastedNewtonNumIterations );
+  }
+
+  // output the average pressure, average temperature, pore volumes
+  computeReservoirStatistics( mesh, outputStatisticsToScreen );
```
TotoGaz, line 895: Move this `computeReservoirStatistics( mesh, outputStatisticsToScreen );` into the `if ( outputStatisticsToScreen...` scope (I think).

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -188,7 +194,7 @@ void CompositionalMultiphaseBase::registerDataOnMesh( Group & meshBodies )
       subRegion.registerExtrinsicData< dPhaseMobility_dGlobalCompDensity >( getName() ).
         reference().resizeDimension< 1, 2 >( m_numPhases, m_numComponents );
 
-      if( m_computeCFLNumbers )
+      if( m_computeStatistics )
```
TotoGaz, line 197: Does it make sense that `phaseOutflux`, `componentOutflux`, `phaseCFLNumber`, `componentCFLNumber` are conditionally register on `if( m_computeStatistics )`?
I could spot no usage of them related to the statistics display.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -859,8 +868,161 @@ void CompositionalMultiphaseBase::initializePostInitialConditionsPreSubGroups()
...
+  m_totalPoreVolume = 0.0;
+  totalUncompactedPoreVolume = MpiWrapper::sum( totalUncompactedPoreVolume );
+  for( integer ip = 0; ip < m_numPhases; ++ip )
+  {
+    m_phasePoreVolume[ip] = MpiWrapper::sum( m_phasePoreVolume[ip] );
+    m_totalPoreVolume += m_phasePoreVolume[ip];
+  }
+
+  m_averagePressure /= totalUncompactedPoreVolume;
+  m_averageTemperature /= totalUncompactedPoreVolume;
```
TotoGaz, line 1013: Could you regroup `min` and `max` on top, separated form the averages computations (which is more complicated)?

More important, are you sure that you are properly computing the average?
How are you considering, _e.g._ the weighted sum for each domain?
This is not so clear to me, but I may be wrong.

TotoGaz (commented):
First review.
Definitely conflicting with https://github.com/GEOSX/GEOSX/pull/1677.
It would be better to call the `displayStatistics` directly from the `SolverBase` loop on top (like https://github.com/GEOSX/GEOSX/pull/1677 does IIRC).

(and maybe rename to `dislpayStatistics` also, but we should discuss about it.)

francoishamon:
> First review. Definitely conflicting with #1677. It would be better to call the `displayStatistics` directly from the `SolverBase` loop on top (like #1677 does IIRC).
> 
> (and maybe rename to `dislpayStatistics` also, but we should discuss about it.)

@TotoGaz I changed my mind for this PR. Initially I was planning to modify `SolverBase` and have a mechanism to output the nonlinear iterations to the terminal, which was implemented in the first version of the PR. Now, to finish this quickier, I am going to downsize the PR to just compute the average quantities in `CompositionalMultiphaseBase`, because this is the most urgent request. I no longer plan to modify `SolverBase`. I am going to start implementing your suggestions now

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
francoishamon: I am no longer going to output anything to terminal (average pressure, nonlinear iterations, etc), so I got rid of `outputStatisticsToScreen`.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
francoishamon: Same here, the information will be always computed when `computeStatistics` is `true`, but only accessible through `TimeHistory`

src/coreComponents/physicsSolvers/SolverBase.cpp
```diff
@@ -75,6 +75,16 @@ SolverBase::SolverBase( string const & name,
...
 
+  registerWrapper( viewKeyStruct::computeStatisticsString(), &m_computeStatistics ).
+    setApplyDefaultValue( 0 ). // do nothing by default
+    setInputFlag( InputFlags::OPTIONAL ).
+    setDescription( "Flag indicating whether statistics are computed or not" );
+
+  registerWrapper( viewKeyStruct::statisticsOutputFrequencyString(), &m_statisticsOutputFrequency ).
+    setApplyDefaultValue( 10 ).
+    setInputFlag( InputFlags::OPTIONAL ).
+    setDescription( "Flag indicating the frequency of the output of statistics to screen" );
```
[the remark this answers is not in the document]
francoishamon: I got rid of this output frequency (which was used to output to file), the output frequency will now be decided by `TimeHistory`.

src/coreComponents/physicsSolvers/SolverBase.hpp
```diff
@@ -494,6 +494,24 @@ class SolverBase : public ExecutableGroup
...
+   *
+   * This function performs whatever tasks are needed to provide the user with a description of the dynamics
+   * taking place in the domain, as well as some statistics on the behavior of the solver
+   */
+  virtual void
+  computeStatistics( real64 const & time,
+                     real64 const & dt,
+                     integer cycleNumber,
+                     DomainPartition & domain,
+                     bool outputStatisticsToScreen );
```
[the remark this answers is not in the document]
francoishamon: I added the `const` keyword here

src/coreComponents/physicsSolvers/SolverBase.cpp
```diff
@@ -75,6 +75,16 @@ SolverBase::SolverBase( string const & name,
     setRestartFlags( RestartFlags::WRITE_AND_READ ).
     setDescription( "Initial time-step value required by the solver to the event manager." );
 
+  registerWrapper( viewKeyStruct::computeStatisticsString(), &m_computeStatistics ).
+    setApplyDefaultValue( 0 ). // do nothing by default
+    setInputFlag( InputFlags::OPTIONAL ).
+    setDescription( "Flag indicating whether statistics are computed or not" );
```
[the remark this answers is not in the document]
francoishamon: In the new version of the PR, I would like to use `TimeHistory` to output the statistics over time (i.e., no output to terminal, no output to vtk). So, I think it would be better not to use `plotLevel` or `logLevel` to control the behavior from the XML file.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
francoishamon: Yes, I think it makes sense. Now, the CFL numbers are computed only when `computeStatistics` is true, in which case we have to register these quantities (`phaseOutflux`, `componentOutflux`, etc). If `computeStatistics` is `false`, the CFL numbers are not computed and these quantities are not needed.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
francoishamon: I separated the `min` and `max` from the averages as you suggested.

Yes, I think I am properly computing the averages. The average is computed to reproduce the `FPRP` keyword in our references. Specifically, in the average, the pressure is weighted by the uncompacted pore volume (`referencePorosity * volume`), and in my tests, the GEOSX average pressure matches well `FPRP`.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
francoishamon: To simplify the code, I removed the `averageTemperature`, `minTemperature`, and `maxTemperature` so there are not fewer quantities to register. Also, I removed all the member variables `m_X` used for statistics (I just register data on the mesh, without using member variables).

francoishamon:
@TotoGaz I updated this PR based on your comments, and it is ready for another round of reviews when you have some time. I modified the PR to compute the average, min, and max quantities on a per-region basis, which is what our use case requires. Therefore, you will see that the statistics are now registered by the solver on the `ElementRegions`. Let me know what you think.

src/coreComponents/physicsSolvers/NonlinearSolverParameters.hpp
```diff
@@ -172,14 +175,23 @@ class NonlinearSolverParameters : public dataRepository::Group
...
   integer m_maxNumConfigurationAttempts;
+
+  // Simulation statistics
+
+  /// Total number of successful Newton iterations
+  integer m_totalSuccessfulNewtonNumIterations;
+
+  /// Total number of wasted Newton iterations
+  integer m_totalWastedNewtonNumIterations;
+
```
TotoGaz, line 194: This should (at least) go in its own `class`.
Here we're mixing input/configuration parameters and meta-results.
As a side effect you could wrap a `LvArray< integer >`, one index for each information, so you should be able to use it in `TimeHistory` too (I know it's a workaround 😉 )

src/coreComponents/physicsSolvers/NonlinearSolverParameters.cpp
```diff
@@ -111,6 +113,8 @@ NonlinearSolverParameters::NonlinearSolverParameters( string const & name,
     setInputFlag( InputFlags::OPTIONAL ).
     setDescription( "Max number of times that the configuration can be changed" );
 
+  registerWrapper( viewKeysStruct::totalSuccessfulNewtonNumIterationsString, &m_totalSuccessfulNewtonNumIterations );
+  registerWrapper( viewKeysStruct::totalWastedNewtonNumIterationsString, &m_totalWastedNewtonNumIterations );
```
TotoGaz, line 117: If you use one class instead of multiple integers you'll be able to register it once.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -245,19 +247,32 @@ void CompositionalMultiphaseBase::registerDataOnMesh( Group & meshBodies )
...
+        ElementRegionBase & region = elemManager.getRegion( regionNames[i] );
+
+        region.registerWrapper< real64 >( viewKeyStruct::averagePressureString() );
+        region.registerWrapper< real64 >( viewKeyStruct::minimumPressureString() );
+        region.registerWrapper< real64 >( viewKeyStruct::maximumPressureString() );
+        region.registerWrapper< real64 >( viewKeyStruct::totalPoreVolumeString() );
+        region.registerWrapper< real64 >( viewKeyStruct::totalUncompactedPoreVolumeString() );
+        region.registerWrapper< array1d< real64 > >( viewKeyStruct::phasePoreVolumeString() ).
+          setSizedFromParent( 0 ).reference().
+          resizeDimension< 0 >( m_numPhases );
```
TotoGaz, line 273: Same here, could you try to use a utility `class/struct` instead?

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.hpp
```diff
@@ -133,21 +126,20 @@ class CompositionalMultiphaseFVM : public CompositionalMultiphaseBase
...
-   * @brief Compute the largest CFL number in the domain
-   * @param dt the time step size
-   * @param domain the domain containing the mesh and fields
-   */
-  void
-  computeCFLNumbers( real64 const & dt, DomainPartition & domain );
-
+  virtual void
+  computeStatistics( real64 const & dt,
+                     DomainPartition & domain ) const override;
```
TotoGaz, line 133: This member functions conditionally computes the statistics if `m_computeStatistics` is `true` (I think that the `doxygen` makes no mention of it).
I would compute the statistics whatever `m_computeStatistics` because it can be clearer.
And I would let the client code check that `m_computeStatistics` is `true` before calling.
If it happens that we have a lot of `if ( m_computeStatistics ) computeStatistics( dt, domain )`, we can make things evolve.

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBaseKernels.hpp
```diff
@@ -566,6 +566,58 @@ struct SolutionCheckKernel
...
+          arrayView2d< real64 const > const & porosity,
+          real64 & minPres,
+          real64 & avgPresNumerator,
+          real64 & maxPres,
+          real64 & totalUncompactedPoreVol,
+          real64 & totalPoreVol )
+  {
+    RAJA::ReduceMin< parallelDeviceReduce, real64 > subRegionMinPres( LvArray::NumericLimits< real64 >::max );
+    RAJA::ReduceSum< parallelDeviceReduce, real64 > subRegionAvgPresNumerator( 0.0 );
+    RAJA::ReduceMax< parallelDeviceReduce, real64 > subRegionMaxPres( 0.0 );
```
TotoGaz, line 589: This is meant to be positive or can you have negative values too?
In the latter case, starting value `0.0` needs to be changed.
And in the first case maybe too (do we intend this value to be unchanged?)...

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBaseKernels.hpp
```diff
@@ -566,6 +566,58 @@ struct SolutionCheckKernel
...
+    forAll< parallelDevicePolicy<> >( size, [=] GEOSX_HOST_DEVICE ( localIndex const ei )
+    {
+      if( elemGhostRank[ei] >= 0 )
+      {
+        return;
+      }
+
+      // To match our "reference", we have to use reference porosity here, not the actual porosity when we compute averages
+      real64 const uncompactedPoreVol = volume[ei] * refPorosity[ei];
+      real64 const dynamicPoreVol = volume[ei] * porosity[ei][0];
```
TotoGaz, line 603: Why index 0 in `porosity[ei][0]`? Could you document it?

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.cpp
```diff
@@ -262,9 +279,114 @@ void SinglePhaseBase::initializePostInitialConditionsPreSubGroups()
...
+    string const & solidName = subRegion.getReference< string >( viewKeyStruct::solidNamesString() );
+    CoupledSolidBase const & solid = getConstitutiveModel< CoupledSolidBase >( subRegion, solidName );
+    arrayView1d< real64 const > const refPorosity = solid.getReferencePorosity();
+    arrayView2d< real64 const > const porosity = solid.getPorosity();
+
+    real64 subRegionAvgPresNumerator = 0.0;
+    real64 subRegionMinPres = 0.0;
+    real64 subRegionMaxPres = 0.0;
+    real64 subRegionTotalUncompactedPoreVol = 0.0;
+    real64 subRegionTotalPoreVol = 0.0;
```
TotoGaz, line 331: Do these values really make sense?

src/coreComponents/physicsSolvers/NonlinearSolverParameters.hpp
francoishamon: The PR no longer deals with the solver parameters, I thought it would be easier to do something separately.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.hpp
francoishamon: I changed my approach to control the output of the statistics. Now they are computed as part of `Tasks`, so we don't have to pollute the solvers with statistics calculations anymore. In particular, this `computeStatistics` function is now gone.

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBaseKernels.hpp
francoishamon: This is because the data living in constitutive models has a dimension for the quadrature point index (always equal to zero in the FV solvers). It shows up everywhere in the solvers, so it would be a little overwhelming to comment on this every time.

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.cpp
francoishamon: Yes, it makes sense to initialize the values like this before computing the min and the max over the sub-region

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBaseKernels.hpp
francoishamon: You are right, I changed the min value initialization to `-LvArray::NumericLimits< real64 >::max` because the single-phase solver allows for negative values.

TotoGaz (commented):
It's surely possible to come up with better `Statistics` classes managemenent.

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsStatistics.hpp
```diff
@@ -0,0 +1,110 @@
...
+    /// String for the node-based statistics
+    constexpr static char const * nodeStatisticsString() { return "nodeStatistics"; }
+  };
+
+  struct NodeStatistics
+  {
+    /// minimum displacement in each direction
+    array1d< real64 > minDisplacement;
+    /// maximum displacement in each direction
+    array1d< real64 > maxDisplacement;
```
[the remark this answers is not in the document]
francoishamon: I added a comment to say that it means in the x, y, and z directions

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseStatistics.hpp
```diff
@@ -0,0 +1,120 @@
...
+{
+
+class SinglePhaseBase;
+
+/**
+ * @class SinglePhaseStatistics
+ *
+ * Task class allowing for the computation of aggregate statistics in single-phase simulations
+ */
+class SinglePhaseStatistics : public TaskBase
```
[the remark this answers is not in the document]
francoishamon: I implemented the `FieldStatisticsBase` class that you suggested. It is used to automate the registration of the solver as we discussed.

src/coreComponents/physicsSolvers/fluidFlow/FlowSolverBase.hpp
```diff
@@ -144,16 +149,18 @@ class FlowSolverBase : public SolverBase
   /// flag to determine whether or not this is a thermal simulation
   integer m_isThermal;
 
+  /// flux estimate used for normalization in single-phase flow
   real64 m_fluxEstimate;
 
+  /// flag to decide whether statistics (avg pressure, CFL) are computed or not
+  integer m_computeStatistics;
```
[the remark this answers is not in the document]
francoishamon: Sorry, that was remnant of the old implementation without `Task`. Now this flag is not needed anymore (I removed it) and I use events to trigger the statistics tasks.

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsStatistics.hpp
```diff
@@ -0,0 +1,110 @@
...
+                        real64 const eventProgress,
+                        DomainPartition & domain ) override;
+
+  /**@}*/
+
+  /**
+   * @brief Compute node-based statistics on the reservoir
+   * @param[in] mesh the mesh level object
+   */
+  void computeNodeStatistics( MeshLevel & mesh ) const;
```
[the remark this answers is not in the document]
francoishamon: I made this function `private`

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsStatistics.cpp
```diff
@@ -0,0 +1,166 @@
...
+void SolidMechanicsStatistics::initializePostInitialConditionsPreSubGroups()
+{
+  DomainPartition & domain = this->getGroupByPath< DomainPartition >( "/Problem/domain" );
+
+  m_solidMechanicsSolver->forMeshTargets( domain.getMeshBodies(), [&] ( string const &,
+                                                                        MeshLevel & mesh,
+                                                                        arrayView1d< string const > const & )
+  {
+    NodeManager & nodeManager = mesh.getNodeManager();
+    nodeManager.registerWrapper< NodeStatistics >( viewKeyStruct::nodeStatisticsString() ).
```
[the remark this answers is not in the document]
francoishamon: Here, I prefer registering the statistics on the `NodeManager` rather than on the `Task` in case there are multiple mesh bodies in the simulation,

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsStatistics.hpp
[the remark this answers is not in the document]
francoishamon: In fact, I had forgotten why I put this function `public` in the first place: it has a kernel launch, so the code does not compile if the function is `private`.

src/coreComponents/physicsSolvers/FieldStatisticsBase.hpp
```diff
@@ -0,0 +1,99 @@
...
+
+    m_solver = physicsSolverManager.getGroupPointer< SOLVER >( m_solverName );
+    GEOSX_THROW_IF( m_solver == nullptr,
+                    GEOSX_FMT( "Could not find solver '{}' of type {}",
+                               m_solverName, LvArray::system::demangleType< SOLVER >() ),
+                    InputError );
+  }
+
+  /// Name of the solver
+  string m_solverName;
```
TotoGaz, line 89: The `m_solverName` is used to retrieve the `m_solver`.
Once done, the we could ditch it (while I think that unfortunately, it's not possible in our current pattern). And if not, `m_solver->getName()` will do the trick in a more consistent way.

Anyhow, `m_solverName` can surely be declared `private`.
And maybe `postProcessInput` as well.

src/coreComponents/physicsSolvers/FieldStatisticsBase.hpp
```diff
@@ -0,0 +1,99 @@
...
+  /**@{*/
+
+  virtual bool execute( real64 const time_n,
+                        real64 const dt,
+                        integer const cycleNumber,
+                        integer const eventCounter,
+                        real64 const eventProgress,
+                        DomainPartition & domain ) override = 0;
+
+  /**@}*/
```
TotoGaz, line 72: I'm not sure this really is mandatory, after all... 🤷

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsStatistics.hpp
```diff
@@ -0,0 +1,98 @@
...
+/**
+ * @class SolidMechanicsStatistics
+ *
+ * Task class allowing for the computation of aggregate statistics in solid mechanics simulations
+ */
+class SolidMechanicsStatistics : public FieldStatisticsBase< SolidMechanicsLagrangianFEM >
+{
+public:
+
+  using Base = FieldStatisticsBase< SolidMechanicsLagrangianFEM >;
```
TotoGaz, line 38: Do you need this? And do you need it `public`?

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseStatistics.hpp
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * @class SinglePhaseStatistics
+ *
+ * Task class allowing for the computation of aggregate statistics in single-phase simulations
+ */
+class SinglePhaseStatistics : public FieldStatisticsBase< SinglePhaseBase >
+{
+public:
+
+  using Base = FieldStatisticsBase< SinglePhaseBase >;
```
TotoGaz, line 38: Do you need this? And do you need it public?

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseStatistics.hpp
```diff
@@ -0,0 +1,109 @@
...
+
+#ifndef SRC_CORECOMPONENTS_PHYSICSSOLVERS_FLUIDFLOW_SINGLEPHASESTATISTICS_HPP_
+#define SRC_CORECOMPONENTS_PHYSICSSOLVERS_FLUIDFLOW_SINGLEPHASESTATISTICS_HPP_
+
+#include "physicsSolvers/FieldStatisticsBase.hpp"
+
+namespace geosx
+{
+
+class SinglePhaseBase;
```
TotoGaz, line 27: Does importing this class result in some link error?

TotoGaz (approved):
LGTM

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsStatistics.hpp
francoishamon: It is needed in the constructor of the class for instance. I made it `private`.

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseStatistics.hpp
francoishamon: I made it `private`.

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseStatistics.hpp
francoishamon: I did not get a link error with this forward declaration.

src/coreComponents/physicsSolvers/FieldStatisticsBase.hpp
francoishamon: I made the `m_solverName` `private`. The `postProcessInput` function cannot be `private` because it is used in the derived classes.

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsStatistics.hpp
TotoGaz: I was able to spot its usage, but since you use it only once in the ctor, I was thinking it was a bit overkill. Anyway, `private` is better.

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseStatistics.hpp
TotoGaz: same

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseStatistics.hpp
TotoGaz: Indeed, but what about using a proper `#include`?

Issue #1725, opened 2022-01-10
[Cleanup] Fix issues found while working on flow validation cases


francoishamon:
**Describe the issue**
Some limitations/issues were found while working on benchmarks and need fixing.

**Proposed cleanup**
For https://github.com/GEOSX/GEOSX/pull/1721, make sure that the case can run smoothly with phases named `water` and `gas` (i.e., no `oil`) with the two input options (tables provided with external PVT files or with `TableFunction`s). Update the phase names in the integrated tests accordingly when this is done.

For https://github.com/GEOSX/GEOSX/pull/1723, two issues need to be fixed before merging: 
- Nonlinear convergence improvement in `CO2Solubility` and `SpanWagnerDensity` [here](https://github.com/GEOSX/GEOSX/blob/cd33485c842afa0754933dfd04c0a017610b2613/src/coreComponents/constitutive/fluid/PVTFunctions/CO2Solubility.cpp#L142) and [here](https://github.com/GEOSX/GEOSX/blob/cd33485c842afa0754933dfd04c0a017610b2613/src/coreComponents/constitutive/fluid/PVTFunctions/SpanWagnerCO2Density.cpp#L195) (the current nonlinear solver does not converge for the values of temperature used in the benchmark). This is fixed in https://github.com/GEOSX/GEOSX/pull/1729. Also, give the user access to the nonlinear solver parameters (max number of iterations, etc).
- Make pure water viscosity dependent on temperature in the Phillips correlation (currently hard-coded [here](https://github.com/GEOSX/GEOSX/blob/cd33485c842afa0754933dfd04c0a017610b2613/src/coreComponents/constitutive/fluid/PVTFunctions/PhillipsBrineViscosity.cpp#L51)). This is fixed in https://github.com/GEOSX/GEOSX/pull/1739

francoishamon:
Resolved by #1729 and #1739

PR #1739, opened 2022-01-21
Fixed pure water viscosity computation in CO2BrinePhillipsFluid

francoishamon:
This PR fixes the computation of the brine viscosity in `CO2BrinePhillipsFluid` as requested in the second bullet point of  #1725. 

This model was previously relying on a hard-coded value of pure water viscosity (0.00089 Pa.s = viscosity of pure water at 25 degC), which was producing inaccurate results in some reservoir conditions (such as Northern Lights or the Class validation case of PR #1723). 

In the PR, the pure water viscosity is now computed as a function of temperature using a table lookup, as in the Ezrokhi model. We can now reproduce very well the results of the Phillips et al. (1981) paper. 

This PR requires a rebaseline that will introduce numerical diffs. Rebaseline done in https://github.com/GEOSX/integratedTests/pull/185


src/coreComponents/constitutive/fluid/PVTFunctions/PureWaterProperties.cpp
```diff
@@ -0,0 +1,276 @@
...
+using namespace stringutilities;
+
+namespace constitutive
+{
+
+namespace PVTProps
+{
+
+TableFunction const *
+PureWaterProperties::makeViscosityTable( string const & functionName,
```
mkhait, line 33: Just for the sake of consistency - maybe name this function also mentioning that it's saturated water?
```suggestion
PureWaterProperties::makeSaturationViscosityTable( string const & functionName,
```

mkhait:
Thanks, @francoishamon for asking me to review this! Looks great to me - pure water properties can definitely be used by many models, so it should have been done in a separate file right from the start. Also thanks for fixing my grammar mistake )

Everything looks good, I only have a single suggestion, but it's up to you.

mkhait (approved):

src/coreComponents/constitutive/fluid/PVTFunctions/PureWaterProperties.cpp
francoishamon: Thanks @mkhait, I modified the code as you suggested.

Issue #114, opened 2018-08-15
EventManager time stepping issues.


corbett5:
When writing out plot files at every step the EventManager winds up taking minuscule time steps and never makes any progress. Below is an example output.

Time: 0s, dt:0s, Cycle: 0
Time: 0s, dt:1e-05s, Cycle: 1
Time: 1e-05s, dt:1e-05s, Cycle: 2
Time: 2e-05s, dt:1e-05s, Cycle: 3
Time: 3e-05s, dt:1e-05s, Cycle: 4
Time: 4e-05s, dt:1e-05s, Cycle: 5
Time: 5e-05s, dt:1.69407e-21s, Cycle: 6
Time: 5e-05s, dt:1.69407e-21s, Cycle: 7
Time: 5e-05s, dt:1.69407e-21s, Cycle: 8
Time: 5e-05s, dt:1.69407e-21s, Cycle: 9
Time: 5e-05s, dt:1.69407e-21s, Cycle: 10
And so on.

1. Go to sedov.xml
2. Change
`<PeriodicEvent name="outputs"
                   timeFrequency="5e-4"
                   targetExactTimestep="1"
                   target="/Outputs/siloOutput" />`
to
`<PeriodicEvent name="outputs"
                   timeFrequency="1e-5"
                   targetExactTimestep="1"
                   target="/Outputs/siloOutput" />`
3. Run the code

PR #1741, opened 2022-01-21
buildrootdir/installrootdir options in config-build.py

XL64:
Adds a `build root dir` / `build path` distinction.
The first option will keep the auto-generated folder name. The latter is unchanged.
Two options overlap somehow: the script fails if both options are provided.

https://github.com/GEOSX/thirdPartyLibs/pull/175 makes the same changes to the TPL version of the script.

### DIFF
--- a/scripts/config-build.py
+++ b/scripts/config-build.py
@@ -50,16 +50,34 @@ def setup_ats(scriptsdir, buildpath):
 
 parser = argparse.ArgumentParser(description="Configure cmake build.")
 
-parser.add_argument("-bp",
-                    "--buildpath",
-                    type=str,
-                    default="",
-                    help="specify path for build directory.  If not specified, will create in current directory.")
+group = parser.add_mutually_exclusive_group()
+group.add_argument("-bp",
+                   "--buildpath",
+                   type=str,
+                   default="",
+                   help="Specify path for build directory. If both `--buildpath` and `--buildrootdir` are not specified, the build directory will be created in current directory, in a subdirectory named `build-<config-file-name>-<buildtype>`. The `--buildpath` option is not compatible with `--buildrootdir`.")
 
-parser.add_argument("-ip",
-                    "--installpath", 
-                    type=str, default="",
-                    help="specify path for installation directory.  If not specified, will create in current directory.")
+
+group.add_argument("-br",
+                   "--buildrootdir",
+                   type=str,
+                   default="",
+                   metavar="BUILD_ROOT_DIR",
+                   help="Specify path for root of build directory. The build directory will be created as `BUILD_ROOT_DIR/build-<config-file-name>-<buildtype>`. The `--buildrootdir` option is not compatible with `--buildpath`.")
+
+group = parser.add_mutually_exclusive_group()
+group.add_argument("-ip",
+                   "--installpath",
+                   type=str,
+                   default="",
+                   help="Specify path for installation directory. If both `--installpath` and `--buildpathdir` are not specified, the install directory will be created in current directory, in a subdirectory named `install-<config-file-name>-<buildtype>`. The `--installpath` option is not compatible with `--installrootdir`.")
+
+group.add_argument("-ir",
+                   "--installrootdir",
+                   type=str,
+                   default="",
+                   metavar="INSTALL_ROOT_DIR",
+                   help="Specify path for root of install directory. The install directory will be created as `INSTALL_ROOT_DIR/build-<config-file-name>-<buildtype>`. The `--installrootdir` option is not compatible with `--installpath`.")
 
 parser.add_argument("--noinstall",
                     action="store_true",
@@ -85,18 +103,13 @@ parser.add_argument("-x",
 parser.add_argument("-ecc",
                     "--exportcompilercommands",
                     action='store_true',
-	                help="generate a compilation database.  Can be used by the clang tools such as clang-modernize.  Will create a file called 'compile_commands.json' in build directory.")
+                    help="generate a compilation database.  Can be used by the clang tools such as clang-modernize.  Will create a file called 'compile_commands.json' in build directory.")
 
 parser.add_argument("-hc",
                     "--hostconfig",
                     required=True,
                     type=str,
                     help="select a specific host-config file to initalize CMake's cache")
-
-parser.add_argument("-tpl",
-                    "--thirdpartylib",
-                    action='store_true',
-                    help="build third party libraries")
 
 parser.add_argument("-gvz",
                     "--graphviz",
@@ -106,8 +119,6 @@ parser.add_argument("-gvz",
 args, unknown_args = parser.parse_known_args()
 if unknown_args:
     print("[config-build]: Passing the following unknown arguments directly to cmake... %s" % unknown_args)
-
-
 
 ########################
 # Find CMake Cache File
@@ -131,10 +142,10 @@ if args.buildpath != "":
     buildpath = args.buildpath
 else:
     # use platform info & build type
-    if args.thirdpartylib:
-        buildpath = "-".join(["../thirdPartyLibs/build",platform_info,args.buildtype.lower()])        
-    else:
-        buildpath = "-".join(["build",platform_info,args.buildtype.lower()])
+    buildpath = "-".join(["build",platform_info,args.buildtype.lower()])
+    if args.buildrootdir != "":
+        buildpath = os.path.join(args.buildrootdir, buildpath)
+
 print("buildpath = ", buildpath)
 
 buildpath = os.path.abspath(buildpath)
@@ -158,10 +169,9 @@ if not args.noinstall:
         installpath = os.path.abspath(args.installpath)
     else:
         # use platform info & build type
-        if args.thirdpartylib:
-            installpath = "-".join(["../thirdPartyLibs/install",platform_info,args.buildtype.lower()])        
-        else:
-            installpath = "-".join(["install",platform_info,args.buildtype.lower()])
+        installpath = "-".join(["install",platform_info,args.buildtype.lower()])
+        if args.installrootdir != "":
+            installpath = os.path.join(args.installrootdir, installpath)
 
     installpath = os.path.abspath(installpath)
 
@@ -209,10 +219,7 @@ if args.graphviz:
 if unknown_args:
     cmakeline += " " + " ".join( unknown_args )
 
-if args.thirdpartylib:
-    cmakeline += " %s/../thirdPartyLibs " % scriptsdir
-else:
-    cmakeline += " %s/../src " % scriptsdir
+cmakeline += " %s/../src " % scriptsdir
 
 # Dump the cmake command to file for convenience
 cmdfile = open("%s/cmake_cmd" % buildpath, "w")
--- a/src/docs/sphinx/buildGuide/BuildProcess.rst
+++ b/src/docs/sphinx/buildGuide/BuildProcess.rst
@@ -38,6 +38,12 @@ Build steps
   * ``--hostconfig`` or ``-hc`` is a path to host-config file.
   * all unrecognized options are passed to CMake.
 
+  If ``--buildpath`` is not used, build directory is automatically named ``build-<config-filename-without-extension>-<buildtype>``.
+  It is possible to keep automatic naming and change the build root directory with ``--buildrootdir``.
+  In that case, build path will be set to ``<buildrootdir>/<config-filename-without-extension>-<buildtype>``.
+  Both ``--buildpath`` and ``--buildrootdir`` are incompatible and cannot be used in the same time.
+  Same pattern is applicable to install path, with ``--installpath`` and ``--installrootdir`` options.
+
 - Run the build:
 
   .. code-block:: console


TotoGaz:
Hello @XL64 thanks for your contribution.

Could you change the title such that is fits, please?

Also,
- What are you trying to do with additional `buildrootdir` and `installrootdir` that you cannot do with `buildpath` and `installpath`?
- How do you deal with potentially conflicting `buildrootdir`/`buildpath` and `installrootdir`/`installpath`?

`config-build.py` is extremely important file they everybody uses (including non devs), it has to be solid.

Also, we unfortunately have a duplicate here https://github.com/GEOSX/thirdPartyLibs/blob/master/scripts/config-build.py 
If this gets merged, do you mind patching there too?

XL64:
Ok I'll do the requested changes. 
To answer the questions:
- buildpath sets the entire buildpath, i just wante to keep the automatic naming ""-".join(["build",platform_info,args.buildtype.lower()])" which is lost if you use buildpath, buildrootdir gives the path before the build directory
- This option is only appicated if buildpath is empty, thus it won't conflict.

I also proposed the change on TPL concerning the duplicate if I am right.

XL64:
I think I should change the help too, to be more explicit on what does each options.

XL64:
See : https://github.com/GEOSX/thirdPartyLibs/pull/172

TotoGaz:
BTW, you can remove the `--thirdpartylib` option since no one uses it. It will make the code you modify clearer.

 > This option is only applicated if `buildpath` is empty, thus it won't conflict.

The main issue is not that you are dealing with it correctly in your codebase, it's that
- `python config-build-py --buildrootdir=path0` makes obvious sense,
- `python config-build-py --buildpath=path1` makes obvious sense,
- but `python config-build-py --buildrootdir=path0 --buildpath=path1` makes no _obvious_ sense: they _somehow_ conflict. Your script discards `path0`, but `path0/path1` could be a sensible move too. Since `config-build.py` is the script everyone is asked to use (even the users with absolutely no dev experience), it has to be rock solid. I would prefer the script to fail and tell the user to make an explicit choice.

Also, there are (at least) https://geosx-geosx.readthedocs-hosted.com/en/latest/docs/sphinx/buildGuide/BuildProcess.html and https://geosx-geosx.readthedocs-hosted.com/en/latest/coreComponents/LvArray/docs/sphinx/buildGuide.html to update with the appropriate options.

XL64:
Ok I'll update this, for the LvArray doc section I think it is linked to another copy of the config-build.py file that I didn't touch.
Should I ?

TotoGaz:
Hello @XL64 !

Let's focus on the _"user interface"_ parameters first, which is not obvious.
Maybe `--buildpath=...` and `--generate-build-folder-name` could be used such that we minimize the risk of wrong interaction there?

Then we can see about doc and duplicates, this should be straightforward once this is done.

XL64:
You mean remove buildrootdir and add a boolean option to jeep the generated name inside buildpath ? 
Why not but in that case the buildpath option would be full path to build directory in case the option is false and build directory parent directory in the other case. Which is slightly different. What do you think is best ?

TotoGaz:
The problem here is that the `config-build.py` is an extremely sensitive script (as I mentioned on top).
So we must pay extra care that end-users do not face any consequences with your patch.

I bet we could have two solutions here
- Have `buildroot` and `buildpath`, failing (with proper error message) in case both are used
- Use a boolean as I last mentioned.

Personally I think that the first one is best, but second could also be a solution.

TotoGaz:
The problem here is that the `config-build.py` is an extremely sensitive script (as I mentioned on top).
So we must pay extra care that end-users do not face any consequences with your patch.

I bet we could have two solutions here
- Have `buildroot` and `buildpath`, failing (with proper error message) in case both are used
- Use a boolean as I last mentioned.

Personally I think that the first one is best, but second could also be a solution.

XL64:
I understand, so what do you think of m'y error message ? Is it clear
enough? Do we need another opinion on that maybe?

Thanks.

Le mar. 25 janv. 2022, 03:03, Thomas Gazzola ***@***.***> a
écrit :

scripts/config-build.py
```diff
@@ -54,12 +54,24 @@ def setup_ats(scriptsdir, buildpath):
...
                     default="",
-                    help="specify path for build directory.  If not specified, will create in current directory.")
+                    help="specify path for build directory.  If not specified, will create in current directory (in a subdirectory named build-<config-file-name>-<buildtype>). Not compatible with -br --buildrootdir, --buildpath should be sufficient.")
+
+
+parser.add_argument("-br",
+                    "--buildrootdir",
+                    type=str,
+                    default="",
+                    help="specify path for root of build directory.  Build path will be set to <BUILDROOTDIR>/build-<config-file-name>-<buildtype> . Not compatible with -br --buildpath.")
```
TotoGaz, line 64:
```suggestion
                    metavar="BUILD_ROOT_DIR",
                    help="Specify path for root of build directory. The build directory will be created as `BUILD_ROOT_DIR/build-<config-file-name>-<buildtype>`. The `--buildrootdir` option is not compatible with `--buildpath`.")
```

scripts/config-build.py
```diff
@@ -54,12 +54,24 @@ def setup_ats(scriptsdir, buildpath):
...
-                    "--installpath", 
+                    "--installpath",
+                    type=str, default="",
+                    help="specify path for root of installation directory.  If not specified, will create in current directory (in a subdirectory named install-<config-file-name>-<buildtype>). Not compatible with -br --installrootdir, --installpath should be sufficient.")
+
+parser.add_argument("-ir",
+                    "--installrootdir",
                     type=str, default="",
-                    help="specify path for installation directory.  If not specified, will create in current directory.")
+                    help="specify path for installation directory.  Install path will be set to <INSTALLROOTDIR>/install-<config-file-name>-<buildtype>. Not compatible with -br --installpath.")
```
TotoGaz, line 74: Same patter for `install*` directories.

scripts/config-build.py
```diff
@@ -107,7 +114,11 @@ def setup_ats(scriptsdir, buildpath):
 if unknown_args:
     print("[config-build]: Passing the following unknown arguments directly to cmake... %s" % unknown_args)
 
+if args.buildpath != "" and args.buildrootdir != "" :
+    sys.exit("Buildpath and buildrootdir are not compatible, you may spectify full buildpath with --buildpath (-bp)")
 
+if args.installpath != "" and args.installrootdir != "" :
+    sys.exit("Installpath and installrootdir are not compatible, you may spectify full installpath with --installpath (-ip)")
```
TotoGaz, line 121: Couldn't https://docs.python.org/3/library/argparse.html#mutual-exclusion be used instead? It looks built-in the std package.

src/docs/sphinx/buildGuide/BuildProcess.rst
```diff
@@ -38,6 +38,11 @@ Build steps
   * ``--hostconfig`` or ``-hc`` is a path to host-config file.
   * all unrecognized options are passed to CMake.
 
+  If --buildpath is not used build directory is automatically named build-<config-filename-without-extension>-<buildtype>.
+  If one wants to keep automatic naming and change the build root directory with --buildrootdir.
+  In that case, build path will be set to <buildrootdir>/<config-filename-without-extension>-<buildtype>.
+  Same is applicable to install path.
```
TotoGaz, line 44:
```suggestion
  If ``--buildpath`` is not used, build directory is automatically named ``build-<config-filename-without-extension>-<buildtype>``.
  It is possible to keep automatic naming and change the build root directory with ``--buildrootdir``.
  In that case, build path will be set to ``<buildrootdir>/<config-filename-without-extension>-<buildtype>``.
  Both ``--buildpath`` and ``--buildrootdir`` are incompatible and cannot be used in the same time.
  Same pattern is applicable to install path, with ``--installpath`` and ``--installrootdir`` options.
```

XL64:
Thanks for the changes, i'll report on TPL file.

XL64:
I'll add the suggestion for exclusive options too.

XL64:
I did change the help + doc + used mutually excluded options, and also updated on TPL

TotoGaz (approved):
~~LGTM~~

scripts/config-build.py
```diff
@@ -50,16 +50,32 @@ def setup_ats(scriptsdir, buildpath):
...
+group.add_argument("-ip",
+                   "--installpath",
+                   type=str, default="",
+                   help="specify path for root of installation directory.  If not specified, will create in current directory (in a subdirectory named install-<config-file-name>-<buildtype>). Not compatible with -br --installrootdir, --installpath should be sufficient.")
+
+group.add_argument("-ir",
+                   "--installrootdir",
+                   type=str, default="",
+                   metavar="INSTALL_ROOT_DIR",
+                   help="specify path for installation directory.  Install path will be set to <INSTALLROOTDIR>/install-<config-file-name>-<buildtype>. Not compatible with -br --installpath.")
```
TotoGaz, line 78: Please the the `metavar` value (`INSTALL_ROOT_DIR`) in your helper for clearer doc.
Plus I think that you swapped `--installrootdir` and `--installpath`, didn't you?
```suggestion
                   type=str,
                   default="",
                   metavar="INSTALL_ROOT_DIR",
                   help="Specify path for root of install directory. The install directory will be created as `INSTALL_ROOT_DIR/build-<config-file-name>-<buildtype>`. The `--installrootdir` option is not compatible with `--installpath`.")
```

scripts/config-build.py
```diff
@@ -50,16 +50,32 @@ def setup_ats(scriptsdir, buildpath):
...
+                   type=str,
+                   default="",
+                   metavar="BUILD_ROOT_DIR",
+                   help="Specify path for root of build directory. The build directory will be created as `BUILD_ROOT_DIR/build-<config-file-name>-<buildtype>`. The `--buildrootdir` option is not compatible with `--buildpath`.")
+
+group = parser.add_mutually_exclusive_group()
+group.add_argument("-ip",
+                   "--installpath",
+                   type=str, default="",
+                   help="specify path for root of installation directory.  If not specified, will create in current directory (in a subdirectory named install-<config-file-name>-<buildtype>). Not compatible with -br --installrootdir, --installpath should be sufficient.")
```
TotoGaz, line 72:
```suggestion
                   type=str,
                   default="",
                   help="Specify path for installation directory. If both `--installpath` and `--buildpathdir` are not specified, the install directory will be created in current directory, in a subdirectory named `install-<config-file-name>-<buildtype>`. The `--installpath` option is not compatible with `--installrootdir`.")
```

TotoGaz (commented):
I think that you swapped --installrootdir and --installpath, didn't you?

scripts/config-build.py
XL64: Indeed, I'll fix it.
Sorry for removing the backquotes also, as they didn't have any effect on formattiong I though it was an error.

TotoGaz: Yes, the back quotes have no effect, other than indicating that we are dealing with some "code".
But more important, it's nice to get the doc homogeneous.

TotoGaz (approved):
LGTM.

Please, when review has started, do not rewrite history of the branch since it makes following the changes really complicated. 😉

XL64:
Ok I though squashing made history clearer, I'll now it for next changes.

Is see that failed jobs are check_pr_is_assigned and Stage 3 but I don't understand how to read it. I don't know where to find jobs that failed. I'll check the dock maybe.

TotoGaz:
> Ok I though squashing made history clearer, I'll now it for next changes.

When we merge, then we squash to get clearer history.
You were correct, but too soon 😉 

> Is see that failed jobs are check_pr_is_assigned and Stage 3 but I don't understand how to read it. I don't know where to find jobs that failed. I'll check the dock maybe.

Yes, `check_pr_is_assigned` checks that the current `pr` is indeed assigned. Which is not the case currently.
You can assign it to yourself and rerun the step (there is a small button on the right).
And you can trigger again the very last job that sums everything up. 🎉

TotoGaz:
> Ok found it. Got a timeout on a test here : https://app.travis-ci.com/github/GEOSX/GEOSX/jobs/556118863 Start 114: testLAIHelperFunctions
> 
> ... still building ...
> 
> ... still building ...
> 
> ... still building ...
> 
> ... still building ...
> 
> ... still building ...
> 
> 114/151 Test #114: testLAIHelperFunctions ...............................***Timeout 1500.03 sec

Actually, https://app.travis-ci.com/github/GEOSX/GEOSX/jobs/556118863 is not related to your PR.
You can click on the `checks` tab.

bmhan12 (commented):
LGTM! There seems to be an issue with adding these changes to thirdPartyLibs? It's failing the CI [here](https://app.travis-ci.com/github/GEOSX/thirdPartyLibs/builds/245475187)

TotoGaz:
> LGTM! There seems to be an issue with adding these changes to thirdPartyLibs? It's failing the CI [here](https://app.travis-ci.com/github/GEOSX/thirdPartyLibs/builds/245475187)

Thx @bmhan12 for the check. Yes, I've seen. It looks like an issue with docker credentials. @XL64 is using a fork for `thirdPartyLibs`, this surely is the reason.

bmhan12 (approved):

XL64:
I shouldn't fork? I shoud commit directly on main repo?

TotoGaz:
> I shouldn't fork? I shoud commit directly on main repo?

Yes (for the `thirdPartyLibs` only).
We have deployment credentials that are not available from the outside.

PR #1742, opened 2022-01-21
Feature/povolny1/multibody vtk

povolny1:
Adding in support for VTK output when using multiple mesh bodies.


src/coreComponents/fileIO/vtk/VTKVTMWriter.cpp
```diff
@@ -59,11 +59,18 @@ void VTKVTMWriter::addSubBlock( string const & blockName, string const & subBloc
...
-  auto dataNode = subBlockNode.append_child( "DataSet" );
+  auto subBlockNode = m_vtmFile.child( "VTKFile" ).child( "vtkMultiBlockDataSet" ).find_child_by_attribute( "Block", "name", blockName.c_str() ).find_child_by_attribute( "Block", "name", subBlockName.c_str() );
+  auto subSubBlockNode = subBlockNode.append_child( "Block" );
+  subSubBlockNode.append_attribute( "name" ) = (blockName + "_" + subSubBlockName).c_str();
+}
+
+void VTKVTMWriter::addDataToSubSubBlock( string const & blockName, string const & subBlockName, string const & subSubBlockName, string const & filePath, int mpiRank ) const
+{
+  auto subBlockNode = m_vtmFile.child( "VTKFile" ).child( "vtkMultiBlockDataSet" ).find_child_by_attribute( "Block", "name", blockName.c_str() ).find_child_by_attribute( "Block", "name", subBlockName.c_str() );
+  auto subSubBlockNode = subBlockNode.find_child_by_attribute( "Block", "name", (blockName + "_" + subSubBlockName).c_str() );
```
povolny1, line 72: It was pointed out in PR #1717 that this code should be refactored such that the naming convention `blockName + "_" + subSubBlockName` is established in a single scope, rather than split across two functions. Suggestions were to either group `addSubSubBlock` and `addDataToSubSubBlock` in a single function, or create a new additional function that is called by `addSubSubBlock` and `addDataToSubSubBlock` which returns `blockName + "_" + subSubBlockName.` Right now I'm leaning towards the latter since `addSubSubBlock` and `addDataToSubSubBlock` are called different amounts of times.

PR #1745, opened 2022-01-22
Updated tutorial 3 with correct depths

herve-gross:
Changes to the mesh file (correct negative depth values, meaningful pressures). Updated figures.


francoishamon (approved):
Looks very good to me, thanks @herve-gross for fixing this tutorial

joshua-white (approved):
Looks great.  My only suggestion might be to use a different colorscale on the pressure figure.  I think blue-white-red is best for symmetric scales with white at zero.  It looks okay as is though if you want to keep it.

herve-gross:
Thanks for the comment @joshua-white, it makes sense. I uploaded two new pressure plots with a "bright rainbow" colorscale instead of red/white/blue.

joshua-white:
Also, where are the wells?  Do the pressure hot spots make physical sense to you, particularly the one at the top ridge?  Just checking.  It is a bit hard for me to understand since this is total pressure, not over pressure.

TotoGaz:
@joshua-white @herve-gross This can be merged as is?

herve-gross:
Yes. Thanks!

PR #1746, opened 2022-01-22
Feature/jafranc/co table relperm hysteresis

jafranc:
This PR is implementing _Killough_ method for hysteresis on the wetting phase. It is implemented so that wetting relperm key points are not necessary derived from the non-wetting ones in `void computeImbibitionWettingRelPerm()`. To do that it introduce a new `m_imbibitionPhaseMaxVolFraction` for the end-point wetting saturation in imbibition and refactor `m_maxPhaseVolFraction` in `m_drainagePhaseMaxVolFraction`. It introduces also a `m_drainagePhaseRelPermEndPoint`, `m_imbibitionPhaseRelPermEndPoint`, and `m_alphaParam_2` to construct the scaled end-point for the wetting imbibition scanning curve.

The requirement for equal Land coefficient is realxed from `GEOSX_ERROR` to `GEOSX_WARNING`.

A unit-test `testRelPermKillough.cpp` is also added to test implementation versus register value of three 2-phases hysteretic cycles based on _Killough_ article provided data.
An extension to check cycles for three phases is work in progress.


francoishamon (approved):
@jafranc Thanks a lot for all this work on wetting-phase relperm hysteresis! Your branch looks great, I am merging it into the base hysteresis branch now, and I will ask you a review when I am done.

PR #1747, opened 2022-01-26
Correct traits and add template getter protection.

TotoGaz:
The was a bug in the `type_list_index` traits which was pointing to last element even when the type was not found.
I corrected by returning ~`end` like for STL algorithms.
I added an index check in `StencilAccessors` to prevent anyone from using an "unsupported" template argument.

### DIFF
--- a/src/coreComponents/codingUtilities/traits.hpp
+++ b/src/coreComponents/codingUtilities/traits.hpp
@@ -215,8 +215,8 @@ static constexpr bool hasCopyAssignmentOp = hasCopyAssignmentOperatorImpl< T >::
 namespace internal
 {
 
-template< typename T, typename U >
-constexpr std::size_t type_list_index( T, U ) { return -1ull; }
+template< typename T, template< typename ... > class L, typename ... Ts >
+constexpr std::size_t type_list_index( T, L<> ) { return 0; }
 
 template< typename T, template< typename ... > class L, typename ... Ts >
 constexpr std::size_t type_list_index( T, L< T, Ts ... > ) { return 0; }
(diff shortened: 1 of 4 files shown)


src/coreComponents/codingUtilities/traits.hpp
```diff
@@ -215,8 +215,8 @@ static constexpr bool hasCopyAssignmentOp = hasCopyAssignmentOperatorImpl< T >::
 namespace internal
 {
 
-template< typename T, typename U >
-constexpr std::size_t type_list_index( T, U ) { return -1ull; }
+template< typename T, template< typename ... > class L, typename ... Ts >
+constexpr std::size_t type_list_index( T, L<> ) { return 0; }
```
TotoGaz, line 219: Instead of using `T` and `U`, I preferred to use an "empty list".
Any concern?

klevzoff: That's fine by me, the only minor thing is you don't need the pack here
```suggestion
template< typename T, template< typename ... > class L >
constexpr std::size_t type_list_index( T, L<> ) { return 0; }
```

We could also consider a version that has static assertion built-in, so you don't have to check the return value at place of use, but this does slightly decrease the flexibility of this facility (no way to check a type list without hard-error), so I don't advocate for this.
```suggestion
template< typename T >
using always_false = std::false_type;

template< typename T, template< typename ... > class L >
constexpr std::size_t type_list_index( T, L<> )
{
  static_assert( always_false< T >::value, "Type not found in list!" );
  return 0;
}
```

klevzoff (approved):
Thanks!

Also @francoishamon you were right when you wanted to return 0 there, but I failed to understand your suggestion :(

src/coreComponents/codingUtilities/traits.hpp
TotoGaz: In terms of error message also, I think that relying on testing the index lets us provide an error message with more context. This can be useful when we spread the pattern and everyone needs to implement it.

Also, it's more similar to `std::find` which returns some "not found"!

francoishamon (approved):
Thanks @TotoGaz

PR #1748, opened 2022-01-27
config file for pangea3

XL64:
Add an update config file for Pangea3

### DIFF
--- a/host-configs/TOTAL/pangea3-gcc8.3-smpi-10.3.0.1.cmake
+++ b/host-configs/TOTAL/pangea3-gcc8.3-smpi-10.3.0.1.cmake
@@ -1,0 +1,121 @@
+# hostconfig for pangea3
+# module load cmake/3.18.2 gcc/8.3.0 cuda/11.3.0 smpi/10.3.0.1 openblas/0.3.8 lapack/3.9.0
+set(CONFIG_NAME "pangea3-gcc8.3-smpi-10.3.0.1" CACHE PATH "") 
+
+# Set up the tpls
+if (DEFINED ENV{GEOSX_TPL_DIR})
+  set(GEOSX_TPL_DIR ENV{GEOSX_TPL_DIR})
+else ()
+  message(FATAL_ERROR "You must set GEOSX_TPL_DIR")
+endif ()
+
+if ((DEFINED ENV{CC}) AND (DEFINED ENV{CXX}) AND (DEFINED ENV{FC}))
+  # C options
+  set(CMAKE_C_COMPILER $ENV{CC} CACHE PATH "")
+  set(CMAKE_C_FLAGS_RELEASE "-O3 -DNDEBUG -mcpu=power9 -mtune=power9" CACHE STRING "")
+  set(CMAKE_C_FLAGS_RELWITHDEBINFO "-g ${CMAKE_C_FLAGS_RELEASE}" CACHE STRING "")
+  set(CMAKE_C_FLAGS_DEBUG "-O0 -g" CACHE STRING "")
+
+  # C++ options
+  set(CMAKE_CXX_COMPILER $ENV{CXX} CACHE PATH "")
+  set(CMAKE_CXX_FLAGS_RELEASE "-O3 -DNDEBUG -mcpu=power9 -mtune=power9" CACHE STRING "")
+  set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "-g ${CMAKE_CXX_FLAGS_RELEASE}" CACHE STRING "")
+  set(CMAKE_CXX_FLAGS_DEBUG "-O0 -g" CACHE STRING "")
+  set(CMAKE_CXX_STANDARD 14 CACHE STRING "")
+
+  # Fortran options
+  set(CMAKE_Fortran_COMPILER $ENV{FC} CACHE PATH "")
+  set(CMAKE_Fortran_FLAGS_RELEASE "-O3 -DNDEBUG -mcpu=power9 -mtune=power9" CACHE STRING "")
+  #set(FORTRAN_MANGLE_NO_UNDERSCORE ON CACHE BOOL "")
+else()
+  message(FATAL_ERROR "You must load module gcc/8.3")
+endif()
+
+
+# OpenMP options
+set(ENABLE_OPENMP ON CACHE BOOL "" FORCE)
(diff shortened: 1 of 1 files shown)


TotoGaz:
Consider modifying `pangea3.cmake` instead?
Or remove it and keep yours if you prefer.

XL64:
Yes I though about removing it, noone uses it I assume, it didn't work for me.

host-configs/TOTAL/pangea3-gcc8.3-smpi-10.3.0.1.cmake
```diff
@@ -0,0 +1,121 @@
+# hostconfig for pangea3
+# module load cmake/3.18.2 gcc/8.3.0 cuda/11.3.0 smpi/10.3.0.1 openblas/0.3.8 lapack/3.9.0
+set(CONFIG_NAME "pangea3-gcc8.3-smpi-10.3.0.1" CACHE PATH "") 
+
+# Set up the tpls
+if (DEFINED ENV{GEOSX_TPL_DIR})
+  set(GEOSX_TPL_DIR ENV{GEOSX_TPL_DIR})
+else ()
+  message(FATAL_ERROR "You must set GEOSX_TPL_DIR")
+endif ()
```
TotoGaz, line 10: While you can surely work this way, I would encourage you to provide it with the command line.
```
python scripts/config-build.py -hc host-configs/your-platform.cmake -bt Release -D GEOSX_TPL_DIR=/full/path/to/thirdPartyLibs
```
see https://geosx-geosx.readthedocs-hosted.com/en/latest/docs/sphinx/QuickStart.html?highlight=GEOSX_TPL_DIR#compilation

Environment variables is surely acceptable for compilers, lapack... but for the `TPL` fixing it command line can be good (specially when you are playing with different versions and different `TPL` versions).

host-configs/TOTAL/pangea3-gcc8.3-smpi-10.3.0.1.cmake
```diff
@@ -0,0 +1,121 @@
...
+  set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "-g ${CMAKE_CXX_FLAGS_RELEASE}" CACHE STRING "")
+  set(CMAKE_CXX_FLAGS_DEBUG "-O0 -g" CACHE STRING "")
+  set(CMAKE_CXX_STANDARD 14 CACHE STRING "")
+
+  # Fortran options
+  set(CMAKE_Fortran_COMPILER $ENV{FC} CACHE PATH "")
+  set(CMAKE_Fortran_FLAGS_RELEASE "-O3 -DNDEBUG -mcpu=power9 -mtune=power9" CACHE STRING "")
+  #set(FORTRAN_MANGLE_NO_UNDERSCORE ON CACHE BOOL "")
+else()
+  message(FATAL_ERROR "You must load module gcc/8.3")
```
TotoGaz, line 31: Note that this will not ensure that the user has the proper `gcc` version loaded.

host-configs/TOTAL/pangea3-gcc8.3-smpi-10.3.0.1.cmake
XL64: Ok il remove the ENV part of this check.

host-configs/TOTAL/pangea3-gcc8.3-smpi-10.3.0.1.cmake
XL64: Indeed i just wanted to check that the ENV i use where defined. One Can use another one but my message indicate the version i tested. Different MPI Can generate hdf5 errors for example. But other might work.

XL64: Maybe i could change my message to "please define CC,CCX,FC variables. We recommand loading module...".

TotoGaz: So maybe comparing with the exact path of the `CC` you want could be a good way to enforce the version?

TotoGaz: > Maybe i could change my message to "please define CC,CCX,FC variables. We recommand loading module...".

It's your users 😉

host-configs/TOTAL/pangea3-gcc8.3-smpi-10.3.0.1.cmake
XL64: This does not work as the argument passed to cmake is not seen within the config file. I'll go back to ENV solution.

TotoGaz: Are you testing against
```
if (NOT DEFINED GEOSX_TPL_DIR})
```
Did you remove the trailing `}`?

I'm surprised that this is not working.
What is the content of the `cmake_cmd` which gets created by the `config-build.py` script?

XL64: Yes I first though it was the forgotten } but I had the same without it.

XL64: Hmm I didn't checkt the command, I'll retry.

TotoGaz: @XL64 What's the status of this PR?

XL64:
Two people used this file in our side last week. It is ready (i forgot to
push a commit, this is why it changed when people tried it).

XL

Le lun. 14 févr. 2022, 19:23, Thomas Gazzola ***@***.***> a
écrit :

TotoGaz:
> Two people used this file in our side last week. It is ready (i forgot to push a commit, this is why it changed when people tried it).

So you did not manage to make the `-DGEOSX_TPL_DIR=...` work (not `-D TPL_DIR=...` like I could read in your commit message)? I think that this will cause you trouble in the future. Do you want to search for the fix or di you want this to be merged as is?

PS: When you answer by mail, could you remove the automatic quoting from your mail (it makes the comments on github quite dirty).

XL64:
I'll give it a try next week and tell you.

XL

XL64:
It seems to work.

I'll fix it in my file then.

XL.

PR #1750, opened 2022-01-27
Wraps extrinsic mesh accessors.

TotoGaz:
An example of what we can do to simplify the usage of these accessors.

Strictly forwarding the `template` call hides any unwanted member function.

- Only the one relevant constructor of `StencilAccessor` is exposed (2 exist).
- Templated getters simply cannot be used anymore. Even if there is a check on wrong template types now, (see https://github.com/GEOSX/GEOSX/pull/1747), this should reduce compilation errors, specially for the people new to C++/GEOSX.

Also, users do not have to go check the type to understand the available data, since it's provided smoothly by the IDE.
Eventually, it's possible that the `extrinsicMeshData` `traits` classes end hidden to "_physics developers_", which I believe would decrease the entry ticked to GEOSX.


src/coreComponents/constitutive/fluid/CellBasedFluxSlurryFluidAccessors.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+#include "physicsSolvers/fluidFlow/StencilAccessors.hpp"
+
+namespace geosx
+{
+
+namespace ProppantTransportKernels
+{
+
+class CellBasedFluxSlurryFluidAccessorsImpl
```
TotoGaz, line 28: Naming still contains the `Impl` suffix, which I should remove.

src/coreComponents/constitutive/fluid/CellBasedFluxSlurryFluidAccessors.hpp
```diff
@@ -0,0 +1,57 @@
...
+{
+public:
+  CellBasedFluxSlurryFluidAccessorsImpl( ElementRegionManager const & elemManager,
+                                         string const & solverName,
+                                         arrayView1d< string const > const & regionNames,
+                                         arrayView1d< string const > const & materialNames )
+    : m_impl( elemManager, solverName, regionNames, materialNames )
+  { }
+
+  auto density()
```
TotoGaz, line 38: We could definitely argue on returning `auto`.
I would prefer some explicit types here for better understanding.
Also, we could "_doxygen_" the accessors with contextualized information.

src/coreComponents/physicsSolvers/fluidFlow/proppantTransport/ProppantTransport.cpp
```diff
@@ -914,13 +914,13 @@ void ProppantTransport::updateCellBasedFlux( real64 const GEOSX_UNUSED_PARAM( ti
...
-                                                permAccessors.get< extrinsicMeshData::permeability::permeability >(),
-                                                permAccessors.get< extrinsicMeshData::permeability::permeabilityMultiplier >(),
-                                                flowAccessors.get< extrinsicMeshData::elementAperture >(),
+                                                flowAccessors.pressure(),
+                                                flowAccessors.gravityCoefficient(),
+                                                slurryFluidAccessors.density(),
+                                                slurryFluidAccessors.viscosity(),
+                                                permAccessors.permeability(),
+                                                permAccessors.permeabilityMultiplier(),
+                                                flowAccessors.elementAperture(),
```
TotoGaz, line 923: This call (e.g.) gets definitely more readable.

src/coreComponents/constitutive/fluid/CellBasedFluxSlurryFluidAccessors.hpp
```diff
@@ -0,0 +1,57 @@
...
+ * See top level LICENSE, COPYRIGHT, CONTRIBUTORS, NOTICE, and ACKNOWLEDGEMENTS files for details.	
+ * ------------------------------------------------------------------------------------------------------------	
+ */
+
+#ifndef GEOSX_CELLBASEDFLUXSLURRYFLUIDACCESSORS_HPP
+#define GEOSX_CELLBASEDFLUXSLURRYFLUIDACCESSORS_HPP
+
+#include "SlurryFluidExtrinsicData.hpp"
+
+#include "physicsSolvers/fluidFlow/StencilAccessors.hpp"
```
TotoGaz, line 20: Is this tool really ment to live only for flow or is it broader than that?

francoishamon: The stencil accessors are used in the kernels that need to access non-local data (i.e., when you need to access data in the two cells adjacent to a face). For now, this pattern is only present in the flow part of the code for the assembly of the fluxes and the computation of flux-like quantities (aquifer, CFL) and it may stay like that for a while. So, I think it makes sense to keep these accessors in `physicsSolvers/fluidFlow`. We can always move them later if/when we have another use case.

src/coreComponents/constitutive/fluid/CellBasedFluxSlurryFluidAccessors.hpp
francoishamon: My preference is to keep the explicit type here, if possible, otherwise the code becomes more difficult to read. If you don't use a smart IDE that tells you what is behind this `auto`, you have to go the file defining the `struct`, read the type, and come back here.

Also, I have not checked that it is the case here, but we should make sure that these accessor names (here,`density`) have the same name as the accessor names defined in the constitutive model for consistency.

src/coreComponents/physicsSolvers/fluidFlow/proppantTransport/ProppantTransport.cpp
francoishamon: Yes, I agree that the kernel calls are lighter.

francoishamon (commented):
Nice work, thank @TotoGaz! I think it is a good approach that we can propagate if other people agree. I see some pros and cons:

Pro: 
- Very clean kernel calls with concise getters that have meaningful names. The solver code becomes easier to read. This is a big improvement.

Cons:
- Loss of flexibility: currently each kernel can use `StencilAccessors` to define exactly the fields it needs. For instance, in the compositional code, the flux kernel and the CFL kernel do not require the same fields. With this new approach, the fields become hard-coded in the `*Accessors.hpp` file that defines the getters, and we lose this flexibility.
- A lot of boilerplate code to define the getters (I don't know if this is a big issue, probably not).

src/coreComponents/constitutive/fluid/CellBasedFluxSlurryFluidAccessors.hpp
TotoGaz: > If you don't use a smart IDE

Then do! 😉 

The issue here is that the type is like `ElementRegion::ElementViewAccessor< array1d< ... > >`. And you add the `const`, and you add the `view`... But yes, having an explicit type is often better.

> we should make sure that these accessor names (here,`density`) have the same name as the accessor names defined in the constitutive model for consistency.

I'm not sure to fully understand. Here I took the name of the `extrinsicMeshData`.
Is there another check?

TotoGaz:
> Loss of flexibility: currently each kernel can use `StencilAccessors` to define exactly the fields it needs. For instance, in the compositional code, the flux kernel and the CFL kernel do not require the same fields. With this new approach, the fields become hard-coded in the `*Accessors.hpp` file that defines the getters, and we lose this flexibility.

I'm not sure to fully understand. E.g. in the given PR, you have two different `PermeabilityAccessorsImpl` with two diffrent sets of fields. One is used in `ProppantTransportKernels` the other in `CompositionalMultiphaseFVMKernels`.

> * A lot of boilerplate code to define the getters (I don't know if this is a big issue, probably not).

But, this is easy code. This could nearly get `macroed`.

Another benefit is that it centralizes those fields definition, and make them unique (if the need to be).
For example, in your case, `SinglePhaseFVMKernels` and `CompositionalMultiphaseFVMKernels` happen to be sharing the same `PermeabilityAccessors`. This gets centralized and made explicit.

src/coreComponents/constitutive/fluid/CellBasedFluxSlurryFluidAccessors.hpp
francoishamon: For the existing accessors, I was just saying that the new getter names you are defining should be consistent with these ones, but that will be easy to check/enforce:
https://github.com/GEOSX/GEOSX/blob/1f36c9006b8d39b4d8785712428b873900b8dc2e/src/coreComponents/constitutive/fluid/MultiFluidBase.hpp#L107
https://github.com/GEOSX/GEOSX/blob/1f36c9006b8d39b4d8785712428b873900b8dc2e/src/coreComponents/constitutive/relativePermeability/RelativePermeabilityBase.hpp#L134
https://github.com/GEOSX/GEOSX/blob/1f36c9006b8d39b4d8785712428b873900b8dc2e/src/coreComponents/constitutive/permeability/PermeabilityBase.hpp#L117

francoishamon:
> I'm not sure to fully understand. E.g. in the given PR, you have two different `PermeabilityAccessorsImpl` with two diffrent sets of fields. One is used in `ProppantTransportKernels` the other in `CompositionalMultiphaseFVMKernels`.

Ok sounds good. Yes, we may want to have multiple `FlowAccessorsImpl` in the same namespace for the compositional solver, but if that is still possible, then that's perfect.

src/coreComponents/constitutive/fluid/CellBasedFluxSlurryFluidAccessors.hpp
TotoGaz: Ah OK. 🕵️ 
I didn't check either.

Do you know if a reorganization such as presented would make those accessors disappear? I do not have the big picture on this.

src/coreComponents/constitutive/fluid/CellBasedFluxSlurryFluidAccessors.hpp
```diff
@@ -0,0 +1,57 @@
...
+{
+
+class CellBasedFluxSlurryFluidAccessorsImpl
+{
+public:
+  CellBasedFluxSlurryFluidAccessorsImpl( ElementRegionManager const & elemManager,
+                                         string const & solverName,
+                                         arrayView1d< string const > const & regionNames,
+                                         arrayView1d< string const > const & materialNames )
+    : m_impl( elemManager, solverName, regionNames, materialNames )
```
TotoGaz, line 35: Note that in this case, the 
```c++
  StencilAccessors( ElementRegionManager const & elemManager, string const & solverName )
```
constructor is not exposed, and cannot be used (by mistake, _e.g._).

CusiniM (commented):
I am on board with these changes and I think it's a good idea. However, wait for #1682 to go in before doing any more work coz you will have to do quite a few changes.

TotoGaz:
Hello @francoishamon @CusiniM @klevzoff @joshua-white @rrsettgast @castelletto1 !
Before I continue working on this, I'd like to state whether we want to push the code in this PR's direction or not (obviously some details need to be assessed).

francoishamon:
I like the changes made in this PR and I vote for propagating them to all the flow solvers.

src/coreComponents/constitutive/fluid/CellBasedFluxSlurryFluidAccessors.hpp
TotoGaz: Yes, I don't know about using `auto` here. I've used it for convenience while refactoring.
I guess we have multiple options
- Using `auto`
- Using some `decltype`/`result_of`/`etc` like in the proposition
- Reproducing the type like `Accesor< array2d< double > >`... This duplicates the type owned in the `extrinsicMeshData` trait, but it has the benefit of totally hiding this trait pattern (always interesting). If the trait changes, then the compilation would fail so this duplication _cannot go unnoticed_.

klevzoff:
I can't decide where I am on this change. I can see the benefits, but on the other hand the whole point of `StencilAccessors` class was that we avoid stamping out boilerplate-looking classes and use a more declarative approach with traits; all the code that currently uses `accessors.get< extrinsicMeshData::someLongName >()` is meant to be exclusively in kernel constructors (as soon as all FV solvers move to kernel class-based approach). 

I guess I'm ok with adding wrappers it as long as I'm not the one doing it :laughing:

TotoGaz:
> [...] the whole point of `StencilAccessors` class was that we avoid stamping out boilerplate-looking classes and use a more declarative approach with traits.

I think this is not lost. The variadic template pattern is sill used at hist best.
I would even say that `variadic templates` exposed is more boilerplate, but it's personnal.

> [...] to me `accessor.getPressure()` and `accessor.get< pressure >()` look roughly the same.

But it's not the case.

`accessor.getPressure()` takes _no_ argument.

`accessor.get< T >()` takes _one_ argument `T`, which needs to be consistent with the type of `accessor`. So you have to understand what's going on with the variadic pattern to properly use it. Moreover, before this patch https://github.com/GEOSX/GEOSX/blob/1dc46ce69fc126f0cd492be3f9f78661a434bb5e/src/coreComponents/physicsSolvers/fluidFlow/StencilAccessors.hpp#L48-L49 we could put anything as a template argument (we found a bug).
`accessor.getPressure()` hides the implementation and your IDE will provide you with the appropriate suggestions.

> Also all the code that currently uses `accessor.get< extrinsicMeshData::someLongName >()` is meant to be contained exclusively in kernel constructors.

This is a good point. That's why we're still in the appropriate `namespaces` (which were already there). Before `modules` save us all!

TotoGaz:
As a second thought, I don't think this is worth the price due to the very localized usage of wrapped variadic template classes.

PR #1756, opened 2022-02-02
Implementation of an MPM solver and Particle Manager

povolny1:
New implementations:
1. A particle manager and associated data structures (`particleRegion`, `particleSubRegion`) which is based on the cell-element data structures. `ParticleManager` is currently registered on `MeshLevel` by default, so this affects all integrated tests. Maybe we can filter this case out or refactor before merging.
2. A new mesh generator - `ParticleMeshGenerator` - which populates the particle manager and has associated `ParticleBlock` infrastructure.
3. A Material Point Method (MPM) solver which currently supports MPI parallelization and has loops in place to support GPUification.

Changes to existing code:
1. ~~Multi-meshbody VTK output.~~ This was somewhat superseded by https://github.com/GEOS-DEV/GEOS/pull/2027, but I still had to make particle-specific changes.
2. Added a '+=' operator for tensors.
3. Added a new argument to various unpacking functions to allow for up to 3 operations: replace, add, max. Replace is the default behavior when the argument is not specified.
4. Added 2 constitutive models: PerfectlyPlastic and CeramicDamage
5. Added a new type of small strain update whereby we rotate to the reference configuration, perform an incremental stress update there, and re-rotate the stress back to the current configuration (`hypoUpdate2`).

Integrated tests: https://github.com/GEOS-DEV/integratedTests/pull/6


povolny1:
This PR will require a rebaseline of all integrated tests due to the `ParticleManager` being registered by default, similar to all other managers.

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsMPM.cpp
```diff
@@ -0,0 +1,897 @@
...
+      } );
+    }
+
+  } );
+
+  forMeshTargets( meshBodies, [&] ( string const &,
+                                    MeshLevel & meshLevel,
+                                    arrayView1d<string const> const & regionNames )
+  {
+    MeshBody const & meshBody = dynamicCast< MeshBody const & >( meshLevel.getParent().getParent() );
```
TotoGaz, line 185: Could you iterate on the `meshBodies` in a classical way and then access the `meshLevels`?
Tricking with `getParent().getParent()` sounds like this is not the pattern awaited by the fwk.

src/coreComponents/physicsSolvers/SolverBase.hpp
```diff
@@ -777,6 +782,20 @@ string SolverBase::getConstitutiveName( ElementSubRegionBase const & subRegion )
   return validName;
 }
 
+template< typename CONSTITUTIVE_BASE_TYPE >
+string SolverBase::getConstitutiveName( ParticleSubRegionBase const & subRegion ) // particle overload
```
TotoGaz, line 786: We should be able to build up a proper class architecture such that we do not have to duplicate to much code and patterns.

src/coreComponents/mesh/generators/ParticleBlockABC.hpp
```diff
@@ -0,0 +1,133 @@
...
+   * @brief Get the type of particle in this subregion.
+   * @return a string specifying the type of particle in this subregion
+   *
+   * See class FiniteParticleBase for possible particle type.
+   */
+  virtual ParticleType getParticleType() const = 0;
+
+  virtual array1d< int > getParticleID() const = 0;
+
+  virtual void setParticleID(array1d< int > const particleID) = 0;
```
TotoGaz, line 62: In the current pattern with the `*ABC` classes, the abstract base class are meant to be consumed by some client code. Seeing setters here mean that your pattern is different. What was your need?
Also I spot no `ParticleBlockManagerABC` so you expose the implementation?

src/coreComponents/mesh/generators/ParticleBlockManager.hpp
```diff
@@ -0,0 +1,166 @@
...
+  {
+    return this->getGroup( viewKeyStruct::particleBlocks() ).getGroup< ParticleBlock >( name );
+  }
+
+private:
+
+  /**
+   * @brief Copy constructor.
+   */
+  ParticleBlockManager( const ParticleBlockManager & );
```
TotoGaz, line 119: What do you want to do here?
Could you use `= default` or `= delete` instead?

TotoGaz (commented):
I didn't go through all the files obviously.
So many architectural questions already.
I'm concerned by this `if/else` pattern popping everywhere.

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsMPM.cpp
povolny1: Now using `MeshBody const & meshBody = meshBodies.getGroup< MeshBody >( meshBodyName );` since `meshBodyName` is an input to this lambda (previously unused). This way I can keep the `forMeshTargets` loop.

src/coreComponents/mesh/generators/ParticleBlockABC.hpp
povolny1: I don't know enough about this to comment, I was just copying the architecture for the cell block manager. I did away with `ParticleBlockManagerABC` since there is currently only one kind of particle block. Should this be addressed now or fixed in a future PR?

TotoGaz: The idea is to have an abstraction (say `ParticleBlockManagerABC` like for `CellBlockManagerABC`) that is used outside of the "meshing" zone.

povolny1: I looked at `CellBlockManagerABC` again and now I remember why I didn't make a `ParticleBlockManagerABC` - `CellBlockManagerABC` doesn't contain any functions or variables that make sense for particles. That being said, I'll implement it anyways since it's possible something may go there someday, but for now it will be basically empty.

povolny1:
`integratedTests` have been rebaselined (here)[https://github.com/GEOSX/integratedTests/pull/226]

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsMPM.hpp
```diff
@@ -0,0 +1,272 @@
...
+  real64 m_stiffnessDamping;
+  TimeIntegrationOption m_timeIntegrationOption;
+  integer m_useVelocityEstimateForQS;
+  real64 m_maxForce = 0.0;
+  integer m_maxNumResolves;
+  integer m_strainTheory = 0;
+  string m_contactRelationName;
+  MPI_iCommData m_iComm;
+
+  std::array<real64, 3> m_hx = {DBL_MAX,DBL_MAX,DBL_MAX};         // Grid spacing in x-y-z
```
TotoGaz, line 237: Maybe use our `stackArray1d` instead?

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsMPM.cpp
```diff
@@ -0,0 +1,897 @@
...
+  forMeshTargets( meshBodies, [&] ( string const & meshBodyName,
+                                    MeshLevel & meshLevel,
+                                    arrayView1d< string const > const & regionNames )
+  {
+    ParticleManager & particleManager = meshLevel.getParticleManager();
+
+    MeshBody const & meshBody = meshBodies.getGroup< MeshBody >( meshBodyName );
+
+    // Set constitutive names on particles
+    if( meshBody.hasParticles() )
```
TotoGaz, line 168: I do not know your business, put is it legit to continue the simulation if you have no particle (maybe yes, just asking.)

src/coreComponents/mesh/ParticleManager.hpp
```diff
@@ -0,0 +1,1514 @@
...
+    this->getGroup( groupKeyStruct::particleRegionsGroup() ).forSubGroups< REGIONTYPE, REGIONTYPES... >( targetRegions, std::forward< LAMBDA >( lambda ) );
+  }
+
+  /**
+   * @brief This const function is used to launch kernel function over all the types of particle regions.
+   * @tparam LAMBDA type of the user-provided function
+   * @param lambda kernel function
+   */
+  template< typename LAMBDA >
+  void forParticleRegionsComplete( LAMBDA lambda ) const
```
TotoGaz, line 308: can you be a bit more strict on your visibilities?
It looks like `forParticleRegionsComplete` could be `private`?
And do you need all these flavors?

src/coreComponents/mesh/ParticleManager.hpp
```diff
@@ -0,0 +1,1514 @@
...
+
+  /**
+   * @brief This is a const function to construct a ParticleViewAccessor to access the data registered on the mesh.
+   * @tparam TRAIT data type
+   * @param neighborName neighbor data name
+   * @return ParticleViewAccessor that contains traits::ViewTypeConst< typename TRAIT::type > data
+   */
+  template< typename TRAIT >
+  ParticleViewAccessor< traits::ViewTypeConst< typename TRAIT::type > >
+  constructExtrinsicAccessor( string const & neighborName = string() ) const;
```
TotoGaz, line 725: It looks like you've built a lot of those helper functions, but I'm not sure you're actually using them.
My IDE does not return some really meaningful results.

TotoGaz (commented):
There are `particle` components which can live their lives in their world, which is fine.
But there are some components which clearly impact our architecture (`MeshBody`, `MeshLevel`, `Partition`, `SolverBase`), and for which it would be wise to have a plan (schedule + resources) before we more forward.

src/coreComponents/constitutive/solid/SolidBase.hpp
```diff
@@ -643,6 +643,15 @@ class SolidBase : public constitutive::ConstitutiveBase
     return m_density;
   }
 
+  /**
+   * @brief Full accessor for density.
+   * @return Accessor
+   */
+  array2d< real64 > & getDensityFull()
```
CusiniM, line 650: Why do you need this? I would avoid giving the possibility to access the array directly. I a resize is necessary it should be the case for the entire `objectManager` (probably a subRegion in this case) and all wrappers should be resized at once.

src/coreComponents/mesh/DomainPartition.cpp
```diff
@@ -156,41 +156,43 @@ void DomainPartition::setupCommunications( bool use_nonblocking )
 
   forMeshBodies( [&]( MeshBody & meshBody )
   {
-
-    MeshLevel & meshLevel = meshBody.getBaseDiscretization();
-
-    for( NeighborCommunicator const & neighbor : m_neighbors )
+    if( !meshBody.hasParticles() )
```
CusiniM, line 159: so, if the `meshBody` has particles, we basically don't need any of the other managers? We clearly need a better way to address this.

src/coreComponents/mesh/MeshManager.cpp
```diff
@@ -106,7 +106,13 @@ void MeshManager::importFields( DomainPartition & domain )
   forSubGroups< MeshGeneratorBase >( [&domain]( MeshGeneratorBase const & generator )
   {
     if( !domain.hasMeshBody( generator.getName() ) )
+    {
       return;
+    }
+    else if( domain.getMeshBody( generator.getName() ).hasParticles() )
```
CusiniM, line 112: I am not sure I understand this. how does having particle relate to fields import?

src/coreComponents/mesh/ObjectManagerBase.cpp
```diff
@@ -842,6 +843,18 @@ void ObjectManagerBase::copyObject( const localIndex source, const localIndex de
...
 }
 
+void ObjectManagerBase::eraseObject( std::set< localIndex > const & indicesToErase )
+{
+  for( auto & nameToWrapper: wrappers() )
+  {
+    WrapperBase * wrapper = nameToWrapper.second;
+    if( wrapper->sizedFromParent() )
+    {
+      wrapper->erase( indicesToErase );
```
CusiniM, line 853: am I understanding correctly that you are not really deleting the wrapper but just removing some entries?

src/coreComponents/mesh/ParticleManager.cpp
```diff
@@ -0,0 +1,347 @@
...
+    unpackedSize += bufferOps::Unpack( buffer, numSubRegionsRead );
+    packList[kReg].resize( numSubRegionsRead );
+    particleRegion.forParticleSubRegionsIndex< ParticleSubRegionBase >(
+      [&]( localIndex const esr, ParticleSubRegionBase & subRegion )
+    {
+      string subRegionName;
+      unpackedSize += bufferOps::Unpack( buffer, subRegionName );
+
+      /// THIS IS WRONG
+      localIndex_array & particleList = packList[kReg][esr].get();
```
CusiniM, line 329: what's wrong? I guess it should be fixed before merging this into develop. Or, at least, I would add a detailed comment of what is not as you would like it to be.

CusiniM (commented):
I have left a few comments. 

I still don't see any `integratedTests`. I would really work on adding those.

src/coreComponents/mesh/ParticleManager.cpp
povolny1: This just got copied over from https://github.com/GEOS-DEV/GEOS/blob/414008af660eacbc7836fd39faf2c9b116bec237/src/coreComponents/mesh/ElementRegionManager.cpp#L443, I figured it was there for a reason

src/coreComponents/mesh/ObjectManagerBase.cpp
povolny1: Yes, that's correct.

src/coreComponents/mesh/MeshManager.cpp
povolny1: The addition of VTK field importation broke my branch and I was not able to figure out why, so I put this is in to circumvent it. We will want to use VTK input for particles someday, but that's beyond the scope of this PR.

src/coreComponents/mesh/DomainPartition.cpp
povolny1: Yes, this all dates from when I had the impression that a mesh body was a particular discretization scheme. This will certainly change when the `MeshLevel` refactor happens, but I hope it's fine for now.

src/coreComponents/constitutive/solid/SolidBase.hpp
povolny1: I can't remember why I added this, so I'll remove it.

src/coreComponents/fileIO/timeHistory/HistoryCollectionBase.cpp
```diff
@@ -158,7 +158,7 @@ dataRepository::Group const * HistoryCollectionBase::getTargetObject( DomainPart
       }
       else if( !meshBody.getMeshLevels().hasGroup< MeshLevel >( targetTokens[1] ) )
       {
-        //GEOSX_LOG_RANK_0( "In TimeHistoryCollection.hpp, Mesh Level Discretization not specified, "
+        //GEOS_LOG_RANK_0( "In TimeHistoryCollection.hpp, Mesh Level Discretization not specified, "
```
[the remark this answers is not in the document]
povolny1: I didn't write this, it must have been deleted at some point but survived on my branch somehow. I will remove it.

src/coreComponents/constitutive/solid/CeramicDamage.hpp
homel1, line 1: This is a copy of a model in old geos.  it will eventually need strengthScale state variable to be passed (to enable weibull variability of strength) but this implementation has passed verification tests.

src/coreComponents/constitutive/solid/PerfectlyPlastic.hpp
homel1, line 1: This looks fine but we should add a plastic strain calc eventually

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp
```diff
@@ -107,6 +129,24 @@ getVtkPoints( NodeManager const & nodeManager,
...
+ * @brief Gets the particle coordinates as a VTK Object from @p particleManager
+ * @param[in] particleManager the ParticleManager associated with the domain being written
+ * @return a VTK object storing all particle coordinates
+ */
+vtkSmartPointer< vtkPoints >
+getVtkPoints( ParticleRegion const & particleRegion )
+{
+  vtkSmartPointer< vtkPoints > points = vtkPoints::New();
+  points->SetNumberOfPoints( 8*particleRegion.size() );
+  array2d< real64 > const coord = particleRegion.getParticleCorners();
```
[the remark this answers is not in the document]
homel1: In the cpdi method, the average of the background grid shape function over the particle domain is computed by sampling the shape function at corners of a parallelepiped particle domain that deforms with deformation gradient.  we will eventually add support for other particle types (tet particles, etc.) which will be plotted differently.

src/coreComponents/mesh/ParticleManager.cpp
homel1: That comment was added when the DEM solver and manager in old geos was being used as a basis for the particle data for the MPM code.  The MPM code didn't use global maps, but i recall during debugging when we were comparing packed and unpacked data that there was some inconsistency related to the global maps.  The issue may have related to the presence of ghost particles in the data structure.  

I've added a more descriptive comment in the code.  we may wish to create a unit test at some point to investigate.

src/coreComponents/mesh/generators/ParticleBlockManager.hpp
```diff
@@ -0,0 +1,113 @@
...
+  /**
+   * @brief Launch kernel function over all the sub-regions
+   * @tparam LAMBDA type of the user-provided function
+   * @param lambda kernel function
+   */
+  template< typename LAMBDA >
+  void forParticleSubRegions( LAMBDA lambda )
+  {
+    this->getGroup( viewKeyStruct::particleBlocks ).forSubGroups< ParticleBlock >( lambda );
+  }
```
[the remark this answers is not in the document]
homel1: Yes, this is used throughout the MPM solver.  while we currently only have full support for CPDI particle types, eventually the different particle type (parallelpiped CPDI, tet CPTI, hex CPDI2, and single point) will be in different subregions allowing for efficient multi-threading of their particle-grid mappings

src/coreComponents/physicsSolvers/SolverBase.hpp
homel1: It's a bit tricky because particles have connections to constitutive data (like elements) but also have velocity, acceleration, etc. (like nodal data).  We'd discussed ideas of having particle be some special kind of single-node element but the existing finite element data structure is not ideal for the MPM algorithm.

Having a constitutive model be something that could be attached to any object (an element, a gauss point, a node, a particle, etc.) might be a good direction to go.

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsMPM.cpp
homel1: It is common for there to me no particles within a single partition for some or all of a simulation.  Particles move through the grid and the entire domain is uniformly partitioned spatially.

homel1 (approved):
I've checked all the MPM specific code and it looks reasonable, much of the approach follows from the old geos-mpm implementation.

We'll need to create issues for the following

1.  address future changes in the mesh level/body data structure,
2. complete support for other particle types
3. Add support for periodic BC
4, port capability for mpm-specific events (healing surface, material swap, ect.)
5. Adopt the current infrastructure for boundary conditions rather than in-solver implementation
6. Make use of infrastructure for table inputs rather than the current use of FTable, BCtable, etc. files.
7. Port capabilities for on-the-fly data reduction from old geos
8. port other constittive models (polymer, arensica, porous, relax and recovery) from old geos
9. add plastic strain calculation to perfectly plastic materials.
10. implement use of strength-scale into brittle damage models
11. optimize performance and use of parallel_device policies where appropriate.

rrsettgast:
work moved to https://github.com/GEOS-DEV/GEOS/pull/2767

PR #1757, opened 2022-02-03
Added flag to disable crossflow in wells

francoishamon:
To reproduce a case with a multi-perforation well injecting very light CO2, we need to ability to detect/disable cross-flow in injectors. This PR:
- Adds a flag to disable crossflow in injectors (i.e., disallow flow from reservoir to well)
- Adds a warning when there is crossflow (otherwise the solver struggles to converge and it is impossible for the users to know why).

Requires a rebaseline in https://github.com/GEOSX/integratedTests/pull/188


src/coreComponents/physicsSolvers/fluidFlow/wells/WellControls.hpp
```diff
@@ -233,6 +233,12 @@ class WellControls : public dataRepository::Group
    */
   bool wellIsOpen( real64 const & currentTime ) const;
 
+  /**
+   * @brief Getter for the flag to disable crossflow in injectors
+   * @return the flag deciding whether crossflow is allowed or not
+   */
+  bool disableInjectorCrossflow() const { return m_disableInjectorCrossflow; }
```
TotoGaz, line 240: Is it relevant to get this flag for a `PRODUCER` well?
What happens in either case?

francoishamon: No, for now, it is not relevant to get this flag for a producer. If a user tries to define the field `disableInjectorCrossflow` in the xml, I put a check in `WellControls.cpp` that will stop the simulation.

So inside the solver, the flag can be used to disable reservoir-to-well flow for injectors, but if someone calls it for a producer (despite the name), it will not do anything and will be ignored.

jhuang2601: How it would function if the injector becomes a producer (or vice versa) after certain time period, such as Pump-in/Flowback Tests?

TotoGaz: @francoishamon What I mean is that at some moment you 
```c++
WellControls::Type const wellType = wellControls.getType();
  bool const disableReservoirToWellFlow =
    ( wellType == WellControls::Type::INJECTOR ) && wellControls.disableInjectorCrossflow();
```
So this is a double check (that you've already made).

What I would find clearer, is to get a `disableInjectorCrossflow` feature more generic, like `disableCrossflow`.
If you want to prevent from using it with `PRODUCER`, then you can do i t like you do.
Reading
```c++
wellControls.isInjector() and wellControls.isCrossFlowDisabled()
```
is more meaningfull.

Moreover, I'm uncomfortable with the wording `disableInjectorCrossflow`.
When first reading it, I thought that calling this function would effectively disable the injector crossflow.
While it does not seem to be the case.
I would rename into `isCrossFlowDisabled` if it's actually the meaning you want to give to this function.

Also, I would add some `wellControls.isInjector()` and `wellControls.isProducer()` as well, it makes the reading more  fluid.

francoishamon: I addressed the comments above: 

> What I would find clearer, is to get a `disableInjectorCrossflow` feature more generic, like `disableCrossflow`. If you want to prevent from using it with `PRODUCER`, then you can do i t like you do. Reading 
> ```c++
> wellControls.isInjector() and wellControls.isCrossFlowDisabled()
> ```

Done
 
> 
>  I would rename into `isCrossFlowDisabled` if it's actually the meaning you want to give to this function.

Done 

> Also, I would add some `wellControls.isInjector()` and `wellControls.isProducer()` as well, it makes the reading more fluid.

Done 

> EDIT: And maybe, at your convenience, rename `wellControls.wellIsOpen()` into `wellControls.isWellOpen()` for consistency.

Done

francoishamon: > How it would function if the injector becomes a producer (or vice versa) after certain time period, such as Pump-in/Flowback Tests?

@jhuang2601 In the cases that you mention, disabling the crossflow would not be a good option to use. The new option implemented in this PR is more an expert option to use with caution in very specific cases. For now, I suggest using it when only 1) when the warning about crossflow pops up, and 2) when the presence of crossflow cripples nonlinear convergence. Otherwise, we should stick to the default behavior (i.e., crossflow enabled).

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWellKernels.cpp
```diff
@@ -1681,7 +1682,7 @@ PresTempCompFracInitializationKernel::
     // note: the targetBHP is not used here because we sometimes set targetBHP to a very large (unrealistic) value
     //       to keep the well in rate control during the full simulation, and we don't want this large targetBHP to
     //       be used for initialization
-    pressureControl = ( wellType == WellControls::Type::PRODUCER )
+    pressureControl = ( isProducer )
                       ? 0.5 * pres
                       : 2.0 * pres;
```
TotoGaz, line 1687:
```suggestion
    pressureControl = isProducer ? 0.5 * pres : 2.0 * pres;
```

src/coreComponents/physicsSolvers/fluidFlow/wells/SinglePhaseWellKernels.cpp
```diff
@@ -611,7 +611,7 @@ PresInitializationKernel::
     // note: the targetBHP is not used here because we sometimes set targetBHP to a very large (unrealistic) value
     //       to keep the well in rate control during the full simulation, and we don't want this large targetBHP to
     //       be used for initialization
-    pressureControl = ( wellType == WellControls::Type::PRODUCER )
+    pressureControl = ( isProducer )
                     ? 0.5 * pres
                     : 2.0 * pres;
```
TotoGaz, line 616:
```suggestion
     pressureControl = isProducer ? 0.5 * pres : 2.0 * pres;
```

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
```diff
@@ -333,37 +333,36 @@ void CompositionalMultiphaseWell::validateWellConstraints( MeshLevel const & mes
...
     WellControls const & wellControls = getWellControls( subRegion );
-    WellControls::Type const wellType = wellControls.getType();
     WellControls::Control const currentControl = wellControls.getControl();
     real64 const & targetTotalRate = wellControls.getTargetTotalRate( m_currentTime + m_currentDt );
     real64 const & targetPhaseRate = wellControls.getTargetPhaseRate( m_currentTime + m_currentDt );
     integer const useSurfaceConditions = wellControls.useSurfaceConditions();
     real64 const & surfaceTemp = wellControls.getSurfaceTemperature();
 
-    GEOSX_THROW_IF( wellType == WellControls::Type::INJECTOR && currentControl == WellControls::Control::PHASEVOLRATE,
+    GEOSX_THROW_IF( wellControls.isInjector() && currentControl == WellControls::Control::PHASEVOLRATE,
```
TotoGaz, line 342: Depending on how you feel it, you may want to reproduce the pattern for `currentControl == WellControls::Control::PHASEVOLRATE` too 😉

src/coreComponents/physicsSolvers/fluidFlow/wells/SinglePhaseWellKernels.cpp
```diff
@@ -573,7 +573,7 @@ PresInitializationKernel::
   real64 const targetBHP = wellControls.getTargetBHP( currentTime );
   real64 const refWellElemGravCoef = wellControls.getReferenceGravityCoef();
   WellControls::Control const currentControl = wellControls.getControl();
-  WellControls::Type const wellType = wellControls.getType();
+  bool const isProducer = wellControls.isProducer();
```
TotoGaz, line 576: Personally I would have called `wellControls.isProducer()` twice instead of storing a variable. This would improve readability (imho).
Also I would have replaced `currentControl == WellControls::Control::BHP` with `wellControls.isBHP()`. But it's pure personal opinion/style.

src/coreComponents/physicsSolvers/fluidFlow/wells/WellControls.hpp
```diff
@@ -226,18 +220,30 @@ class WellControls : public dataRepository::Group
...
    * @brief Getter for the status of the well (open or shut)
    * @param[in] currentTime the current time
    * @return a flag equal to true if the well is open, and false otherwise
    */
-  bool wellIsOpen( real64 const & currentTime ) const;
+  bool isWellOpen( real64 const & currentTime ) const;
 
   /**
    * @brief Getter for the flag to disable crossflow in injectors
    * @return the flag deciding whether crossflow is allowed or not
```
TotoGaz, line 244:
```suggestion
   * @brief Is the crossflow disabled?
   * @return A boolean
```
Note that there is no more mention of `injector`.

src/coreComponents/physicsSolvers/fluidFlow/wells/WellControls.hpp
```diff
@@ -226,18 +220,30 @@ class WellControls : public dataRepository::Group
    */
   const real64 & getSurfaceTemperature() const { return m_surfaceTemp; }
 
+  /**
+   * @brief Getter for the well type
```
TotoGaz, line 224:
```suggestion
   * @brief Is the well an injector?
```

src/coreComponents/physicsSolvers/fluidFlow/wells/WellControls.hpp
```diff
@@ -226,18 +220,30 @@ class WellControls : public dataRepository::Group
...
+   * @return true if the well is an injector, false otherwise
+   */
+  bool isInjector() const { return ( m_type == Type::INJECTOR ); }
+
+  /**
+   * @brief Getter for the well type
```
TotoGaz, line 230:
```suggestion
   * @brief Is the well a producer?
```

src/coreComponents/physicsSolvers/fluidFlow/wells/WellControls.hpp
```diff
@@ -226,18 +220,30 @@ class WellControls : public dataRepository::Group
...
+   * @return true if the well is a producer, false otherwise
+   */
+  bool isProducer() const { return ( m_type == Type::PRODUCER ); }
+
   /**
    * @brief Getter for the status of the well (open or shut)
```
TotoGaz, line 236:
```suggestion
   * @brief Is the well open (or shut) at @p currentTime?
```

src/coreComponents/physicsSolvers/fluidFlow/wells/WellControls.hpp
```diff
@@ -226,18 +220,30 @@ class WellControls : public dataRepository::Group
...
    */
-  bool wellIsOpen( real64 const & currentTime ) const;
+  bool isWellOpen( real64 const & currentTime ) const;
 
   /**
    * @brief Getter for the flag to disable crossflow in injectors
    * @return the flag deciding whether crossflow is allowed or not
    */
-  bool disableInjectorCrossflow() const { return m_disableInjectorCrossflow; }
+  bool isCrossflowDisabled() const { return m_isCrossflowDisabled; }
```
TotoGaz, line 246: Just a detail, but it could be a good habit to use some
```suggestion
  bool isCrossflowEnabled() const { return m_isCrossflowEnabled; }
```
to prevent from some double negations like `wellControls.isInjector() && !wellControls.isCrossflowDisabled()`.
And would a `wellControls.hasCrossFlow()` would make sense? (not sure...)

src/coreComponents/physicsSolvers/fluidFlow/wells/WellControls.hpp
```diff
@@ -226,18 +220,30 @@ class WellControls : public dataRepository::Group
...
+   */
+  bool isProducer() const { return ( m_type == Type::PRODUCER ); }
+
   /**
    * @brief Getter for the status of the well (open or shut)
    * @param[in] currentTime the current time
    * @return a flag equal to true if the well is open, and false otherwise
    */
-  bool wellIsOpen( real64 const & currentTime ) const;
+  bool isWellOpen( real64 const & currentTime ) const;
```
TotoGaz, line 240: I looks better like this 🤓

TotoGaz (commented):
Bad luck that `WellControls::Type` is used in some `ENUM_STRINGS` pattern, otherwise it could be `private`. Do you need this?
And if I were you, I would write a unit test which validates that `WellControls::Type::PRODUCER` and `INJECTOR` are always in the same order and equal to 0 and 1. (`ENUM_STRINGS` seems to rely on this). Simply sorting alphabetically will cause a bug I think.

src/coreComponents/physicsSolvers/fluidFlow/wells/WellControls.hpp
francoishamon: Done, I changed to `isCrossflowEnabled()`.

For the `wellControls.hasCrossFlow()` getter, it makes less sense, because it is something that we detect on the fly, so there is no need for a getter.

src/coreComponents/physicsSolvers/fluidFlow/wells/WellControls.hpp
francoishamon: Done

src/coreComponents/physicsSolvers/fluidFlow/wells/WellControls.hpp
francoishamon: Done

src/coreComponents/physicsSolvers/fluidFlow/wells/SinglePhaseWellKernels.cpp
francoishamon: Done

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWellKernels.cpp
francoishamon: Done

src/coreComponents/physicsSolvers/fluidFlow/wells/SinglePhaseWellKernels.cpp
francoishamon: Done

src/coreComponents/physicsSolvers/fluidFlow/wells/WellControls.hpp
TotoGaz: Damned, we forgot about the doxygen 🙈

TotoGaz (approved):

PR #1758, opened 2022-02-04
Improve VisIt Compatibility

cssherman:
This PR does the following:

- Updates the names of vtk files to match the expected format of the VisIt grouping feature
- Adds an option to limit region types that are output in vtk file (VisIt cannot handle the multi-region format we are using)
- Adds a visit macro that can be used to convert existing vtk files
- Updates the output documentation


cssherman:
Visit appears to only populate its list of ranks when reading the initial file in a series.  This will require us to include empty ranks in the vtm files so that later time states can be visualized correctly

klevzoff (approved):

src/coreComponents/fileIO/doc/Outputs.rst
```diff
@@ -108,6 +108,63 @@ In VisIT :
...
+
+
+In VisIt, the prepared files can be visualized as follows:
+
+1. Click the Open icon (or select File/Open from the top menu).
+2. Enter the vtk output folder in the left-hand panel (these will be separate if you use option 2 above).
+3. Select the desired *.vtm group in the right-hand panel (by double-clicking or selecting OK).
+4. (Repeat for any additional datasets/regions.)
+5. In the main window, select the desired dataset from the Active Sources dropdown menu.
+6. Click on the Add icon, and select a parameter to visualize from the mesh, psuedocolor, or vector plot options
```
jhuang2601, line 161:
```suggestion
6. Click on the Add icon, and select a parameter to visualize from the mesh, pseudo color, or vector plot options
```

src/coreComponents/python/visitMacros/visitVTKConversion.py
```diff
@@ -0,0 +1,66 @@
...
+regions into separate files, which can then be normally read by the tool.
+To install the macros, do the following:
+  1) Select Controls/Command from the top menu in VisIt
+  2) In the newly opened Commands window, select the Macros tab
+  3) Copy the following scripts to the panel and click Update Macros
+  4) The macro will appear under the name 'Convert pvd' in the macros window
+
+If the macros window is not visible, select Controls/Macros from the top menu in VisIt.
+When you select 'Convert pvd', visit will attempt to convert any pvd files in the current
+directory (where the visit command was ran).  The macro will create a new set of .vtm files
```
jhuang2601, line 15:
```suggestion
directory (where the visit command was running).  The macro will create a new set of .vtm files
```

jhuang2601 (approved):

jhuang2601:
@cssherman Nice job and thanks for updating the outdated xml file. However, except for changing the output format of vtk files, the rest xml files are not running and covered in the integratedTests. I think rebaseling is not necessary for this PR.

cssherman:
> @cssherman Nice job and thanks for updating the outdated xml files. However, except for changing the output format of vtk files, the rest xml files are not running and covered in the integratedTests. I think rebaseline is not necessary for this PR.

Indeed, I found the issue with the xml validation while working on the integrated tests

src/coreComponents/fileIO/doc/Outputs.rst
cssherman: In VisIt, psuedocolor is written as a single word

TotoGaz: But it's still not spelled `psuedocolor` 😝

src/coreComponents/fileIO/Outputs/VTKOutput.cpp
```diff
@@ -47,10 +46,14 @@ VTKOutput::VTKOutput( string const & name,
     setDescription( "Level detail plot. Only fields with lower of equal plot level will be output." );
 
   registerWrapper( viewKeysStruct::binaryString, &m_writeBinaryData ).
-    setApplyDefaultValue( 1 ).
+    setApplyDefaultValue( m_writeBinaryData ).
     setInputFlag( InputFlags::OPTIONAL ).
-    setDescription( "Output the data in binary format" );
+    setDescription( "Output the data in binary format.  Valid options:\n" + EnumStrings< vtk::VTKOutputMode >::concat( "\n* " ) );
```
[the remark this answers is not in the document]
klevzoff: > I still do not think that attaching `concat` to `ENUM_STRINGS` is the best move. It should go in our `stringutilities` bucket.

Well, `EnumStrings<T>::concat()` is just a shortcut for `stringutilities::join()`, but yeah, we could just use that one directly.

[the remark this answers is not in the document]
cssherman: Sounds reasonable.  As a side-note, I really like this current approach with enum strings and how it integrates into the schema.  The validation tools that I'm using produce very human-readable errors with these attributes, and I expect that these will be very useful in constructing the gui

PR #1760, opened 2022-02-04
Unified mass/molar conversions in fluid models

francoishamon:
This PR unifies the mass/molar fractions conversions and the total density computations done in the fluid models, as they are currently duplicated. The PR adds the necessary functions `convertToMassFractions`, `convertToMoleFractions`, and `computeTotalDensity` in `MultiFluidBase.hpp`.

There would be more refactoring needed in the CO2-brine case (for instance to better use the `MultiFluidVars`in the PVT functions), but I left that for another PR.

This does not require a rebaseline.


src/coreComponents/constitutive/fluid/MultiFluidBase.hpp
```diff
@@ -254,10 +265,106 @@ class MultiFluidBase : public ConstitutiveBase
       m_totalDensity( std::move( totalDensity ) )
     { }
 
+    /**
+     * @brief Utility function to convert mass fractions to mole fractions
+     * @tparam OUT_ARRAY the type of array storing the component mole fractions
+     * @param[in] composition the component mass fractions
+     * @param[in] componentMolarWeight the component molar weight
+     * @param[out] compMoleFrac the newly converted component mole fractions
+     * @detail The template is needed because PVTPackage expects an std::vector
```
joshua-white, line 274:
```suggestion
     * @detail The template is needed because PVTPackage expects a std::vector
```

src/coreComponents/constitutive/fluid/MultiFluidBase.hpp
```diff
@@ -254,10 +265,106 @@ class MultiFluidBase : public ConstitutiveBase
...
+
+    /**
+     * @brief Utility function to convert mass fractions to mole fractions and keep derivatives
+     * @tparam maxNumPhase the max number of phases
+     * @tparam OUT_ARRAY the type of array storing the component mole fractions
+     * @param[in] composition the component mass fractions
+     * @param[in] componentMolarWeight the component molar weight
+     * @param[out] compMoleFrac the newly converted component mole fractions
+     * @param[out] dCompMoleFrac_dCompMassFrac the derivatives of the newly converted component mole fractions
+     * @detail The template is needed because PVTPackage expects an std::vector
```
joshua-white, line 290:
```suggestion
     * @detail The template is needed because PVTPackage expects a std::vector
```

src/coreComponents/constitutive/fluid/MultiFluidBase.hpp
```diff
@@ -339,6 +478,322 @@ class MultiFluidBase : public ConstitutiveBase
 
 };
 
+template< typename OUT_ARRAY >
+GEOSX_HOST_DEVICE
+void
+MultiFluidBase::KernelWrapper::
+  convertToMoleFractions( arraySlice1d< real64 const, compflow::USD_COMP - 1 > const composition,
```
joshua-white, line 485: If i understand this correctly, there are two version here, one returning derivs and one not.  I might have one call the other, to condense the actual implementation to one spot.  You could call the one with derivs from a non-deriv version with a junk deriv array, and then just discard the extra data.  It would look cleaner, at the cost of a negligeable overhead when you just want values and not derivs.  Just a suggestion.  Your preference.

joshua-white (approved):
Looks good to me.  This is great. 

We may want to have a separate discussions about using `PhaseProp::SliceType` and similar data types more consistently in the code, but that is another PR.

src/coreComponents/constitutive/fluid/MultiFluidBase.hpp
```diff
@@ -339,6 +478,322 @@ class MultiFluidBase : public ConstitutiveBase
...
+    phaseFrac.dPres[ip] = ( phaseFrac.dPres[ip] - phaseFrac.value[ip] * dTotalMass_dP ) * totalMassInv;
+    phaseFrac.dTemp[ip] = ( phaseFrac.dTemp[ip] - phaseFrac.value[ip] * dTotalMass_dT ) * totalMassInv;
+
+    for( integer jc = 0; jc < numComp; ++jc )
+    {
+      phaseFrac.dComp[ip][jc] = ( phaseFrac.dComp[ip][jc] - phaseFrac.value[ip] * dTotalMass_dC[jc] ) * totalMassInv;
+    }
+  }
+
+  // 2. Convert phase compositions
```
klevzoff, line 672: Same question, would it be better to make this into 3 smaller, more focused functions (at the cost of having to remember to call all 3 at in each model)?

klevzoff (approved):
Very nice! Just have some minor optional suggestions/questions

src/coreComponents/constitutive/fluid/MultiFluidBase.hpp
francoishamon: Done

src/coreComponents/constitutive/fluid/BlackOilFluid.hpp
```diff
@@ -329,55 +318,31 @@ BlackOilFluid::KernelWrapper::
...
-                   phaseViscosity.dComp );
+    convertToMassFractions< NC_BO, NP_BO >( dCompMoleFrac_dCompMassFrac,
+                                            phaseMolecularWeight,
+                                            dPhaseMolecularWeight_dPressure,
+                                            dPhaseMolecularWeight_dTemperature,
+                                            dPhaseMolecularWeight_dGlobalCompFraction,
+                                            phaseFraction,
+                                            phaseCompFraction,
+                                            phaseDensity.dComp,
+                                            phaseViscosity.dComp );
```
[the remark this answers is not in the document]
francoishamon: Done, I removed these templates parameters that were unnecessary.

src/coreComponents/constitutive/fluid/MultiFluidBase.hpp
francoishamon: Yes, I implemented what you suggested, with three more focused functions doing just one specific task.

src/coreComponents/constitutive/fluid/MultiFluidBase.hpp
```diff
@@ -339,6 +478,322 @@ class MultiFluidBase : public ConstitutiveBase
...
+
+    bool const phaseExists = (phaseFrac[ip] > 0);
+    if( !phaseExists )
+    {
+      continue;
+    }
+
+    phaseFrac[ip] *= phaseMolecularWeight[ip];
+    totalMass += phaseFrac[ip];
+  }
```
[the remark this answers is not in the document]
francoishamon: For your first question, you are right, the fractions for non-existent phases are just zero, so we can just skip them here. Specifically, the places where we set the phase fractions to zero for non-existent phases are:
- Black-Oil: https://github.com/GEOSX/GEOSX/blob/343d427a3eef852a50e7a73a372a0b9df0c24355/src/coreComponents/constitutive/fluid/BlackOilFluid.hpp#L524
- CO2-brine: https://github.com/GEOSX/GEOSX/blob/343d427a3eef852a50e7a73a372a0b9df0c24355/src/coreComponents/constitutive/fluid/PVTFunctions/CO2Solubility.hpp#L178
- Compositional: https://github.com/GEOSX/PVTPackage/blob/064526518b966fea1ae6f722a0903e934827ef4a/PVTPackage/source/MultiphaseSystem/PhaseSplitModel/CompositionalFlash.cpp#L99

Therefore, I removed these checks in the latest commits.

For your second question, yes I agree, I think it would be nice to have the `forExistingPhases` lambda in the compositional solver kernels where we divide by density. Much better than the `continue` keywords that I put in a few places. I have not implemented this in this PR to avoid touching other files, but it would be good for a future refactoring.

PR #1767, opened 2022-02-09
Update hypre, localIndex to int, add smoke and benchmark cases for ECP problem. 

CusiniM:
- [x] Add benchmark for ECP test problem
- [x] update hypre to fix bugs in MGR
- [x] try using mixedInt in hypre on GPU
- [x] switch `localIndex` to `int` for all builds

https://github.com/GEOSX/LvArray/pull/249
https://github.com/GEOSX/thirdPartyLibs/pull/177
https://github.com/GEOSX/integratedTests/pull/197


src/coreComponents/dataRepository/BufferOps_inline.hpp
```diff
@@ -1088,7 +1088,7 @@ Pack( buffer_unit_type * & buffer,
     else
     {
       sizeOfPackedChars += Pack< DO_PACKING >( buffer,
-                                               localIndex( -1 ) );
+                                               globalIndex( -1 ) );
```
rrsettgast, line 1091: These were getting packed as a `localIndex`, but read as a `globalIndex`, which doesn't work if they are not the same underlying type.

src/coreComponents/fileIO/Outputs/TimeHistoryOutput.hpp
```diff
@@ -110,7 +110,7 @@ class TimeHistoryOutput : public OutputBase
    * @return Return PyHistoryOutput type.
    */
 #if defined(GEOSX_USE_PYGEOSX)
-  virtual PyTypeObject * getPythonType() const;
+  virtual PyTypeObject * getPythonType() const override;
```
rrsettgast, line 113: We have a strict requirement for this...don't know why it doesn't show up on CI??

src/coreComponents/physicsSolvers/multiphysics/LagrangianContactSolver.cpp
```diff
@@ -2404,7 +2404,7 @@ void LagrangianContactSolver::computeFractureStateStatistics( DomainPartition co
                                                               bool printAll ) const
 {
 
-  array1d< localIndex > localCounter( 3 );
+  array1d< globalIndex > localCounter( 3 );
```
rrsettgast, line 2407: `mpiWrapper::gather` was sending `localIndex` and receiving `globalIndex`. Don't know how this is supposed to work...but it was causing problems, so I just made the sending type `globalIndex`. Perhaps we should disable send/receive of different types in our MPI wrappers?

klevzoff: Yeah, MPI_Gather won't work if send/receive types are of different size. That said, this function should probably use `MpiWrapper::allReduce()` instead.

src/cmake/GeosxOptions.cmake
```diff
@@ -108,7 +108,7 @@ message( "CMAKE_CXX_COMPILER_ID = ${CMAKE_CXX_COMPILER_ID}" )
 blt_append_custom_compiler_flag( FLAGS_VAR CMAKE_CXX_FLAGS DEFAULT "${OpenMP_CXX_FLAGS}" )
 blt_append_custom_compiler_flag( FLAGS_VAR CMAKE_CXX_FLAGS
                                  GNU   "-Wall -Wextra -Wpedantic -pedantic-errors -Wshadow -Wfloat-equal -Wcast-align -Wcast-qual"
-                                 CLANG "-Wall -Wextra -Wpedantic -pedantic-errors -Wshadow -Wfloat-equal -Wcast-align -Wcast-qual"
+                                 CLANG "-Wall -Wextra -Wpedantic -pedantic-errors -Wshadow -Wfloat-equal -Wno-cast-align -Wcast-qual"
```
klevzoff, line 111: This warning seems a bit dangerous to disable, although it is kept on gcc at least. What was the problem with it, are we casting something to/from `localIndex *` that we shouldn't?

src/coreComponents/schema/docs/FaceManager_other.rst
```diff
@@ -1,25 +1,25 @@
...
+edgeList                geosx_InterObjectRelation< LvArray_ArrayOfArrays< int, int, LvArray_ChaiBuffer > >                                                                                                  (no description available)                                
+elemList                integer_array2d                                                                                                                                                                     (no description available)                                
+elemRegionList          integer_array2d                                                                                                                                                                     (no description available)                                
+elemSubRegionList       integer_array2d                                                                                                                                                                     (no description available)
```
klevzoff, line 10: A lot of these changes are undesirable. On the other hand, it's impossible to distinguish aliases that map to the same type. Not sure if there is a solution to this.

src/coreComponents/schema/docs/AcousticSEM.rst
```diff
@@ -9,9 +9,9 @@ dtSeismoTrace             real64         0        Time step for output pressure
...
 receiverCoordinates       real64_array2d required Coordinates (x,y,z) of the receivers                                                                                                                                                                                                                                                                                     
-rickerOrder               localIndex     2        Flag that indicates the order of the Ricker to be used o, 1 or 2. Order 2 by default                                                                                                                                                                                                                                     
+rickerOrder               integer        2        Flag that indicates the order of the Ricker to be used o, 1 or 2. Order 2 by default
```
klevzoff, line 14: Would make sense for these to be actually changed to `integer` in code

klevzoff (approved):
:+1:

src/cmake/GeosxOptions.cmake
rrsettgast: We have to update python in the TPL's and then re-enable this. The error was:
```
/usr/gapps/GEOSX/thirdPartyLibs/python/linux-rhel7-x86_64-clang@10.0.0/python@3.8.5/include/python3.8/cpython/abstract.h:91:11: error: cast from 'char *' to 'vectorcallfunc *' (aka '_object *(**)(_object *, _object *const *, unsigned long, _object *)') increases required alignment from 1 to 8 [-Werror,-Wcast-align]
```

src/coreComponents/schema/docs/FaceManager_other.rst
rrsettgast: I didn't look at the rst changes...I don't understand why these have propagated to the rst generation.

src/coreComponents/schema/docs/AcousticSEM.rst
rrsettgast: Yes, in this case it would. Still don't understand what is prompting the changes. @cssherman ??

src/coreComponents/schema/docs/FaceManager_other.rst
klevzoff: The XML/RST name is derived from `typeid`, but now `typeid(integer_array2d) == typeid(localIndex_array2d)` so I guess it just picks the first one it finds

src/coreComponents/physicsSolvers/multiphysics/LagrangianContactSolver.cpp
rrsettgast: I think this was just copied from the residual norm evaluation which isn't a simple reduction.

src/cmake/GeosxOptions.cmake
TotoGaz: IIUC this error comes from the python code in `cpython/abstract.h`?
Warnings from TPL includes should not pop if they are included between `<>` and/or with the `-isystem` from the command line (in `gcc`).
It looks like it's working with `GNU` (since you did not update the line). But the same behavior should occur with `clang`?

PR #1770, opened 2022-02-12
Fixing Tutorial 3

herve-gross:
Minor changes based on the multibody PR


francoishamon (approved):
Thanks @herve-gross for fixing this

PR #1777, opened 2022-02-15
Add tpls include in dummy cmake for doc.

TotoGaz:


src/docs/sphinx/QuickStart.rst
```diff
@@ -285,6 +285,9 @@ A typical one may look like:
 
   # enable tests
   set(ENABLE_GTEST_DEATH_TESTS ON CACHE BOOL "" FORCE )
+  
+  # define some thrid party libraries information
```
joshua-white, line 289:
```suggestion
  # define some third party libraries information
```

joshua-white (approved):

TotoGaz:
Hello @karimifard can you check that this patch solves your issue?

karimifard:
It still does not work. 

Added the following line to the end of your-platform.cmake
include(${CMAKE_CURRENT_LIST_DIR}/tpls.cmake)

The TPLs were compiled fine (no error messages)

When compiling GEOSX the following command does not succeed
python scripts/config-build.py -hc host-configs/your-platform.cmake -bt Release

ERR: GEOSX requires hdf5, set HDF5_DIR to the hdf5 installation directory.

Note that if I add it manually using set() it will work (I have to do it for all of them)

Let me know if you need a more detailed log file.

TotoGaz:
Hmmm... And the `tpls.cmake` file is in the same folder as your `your-platform.cmake` file obviously? (otherwise `cmake` would yell that it doesn't find the `tpls.cmake` file).

When you run the `python` script, do you provide `-DGEOSX_TPL_DIR=...`?
What is the error log? What is the `GEOSX_TPL_DIR` `cmake` uses?

karimifard:
Yes, tpls.cmake is in the same folder.
No, when running python script I do not provide -DGEOSX_TPL_DIR=
I just tried -DGEOSX_TPL_DIR=path to my thirdPartyLib folder, no change
I tried path to my thirdPartyLib/install-your-platform-release, didn't work
I don't think GEOSX_TPL_DIR  is defined.

TotoGaz:
Hmmm, this is strange. No information from the logs?
However, I'll have a look when I'm in Stanford. 😉

karimifard:
I think the problem is that GEOSX_TPL_DIR is not defined. Where should this be defined? 
Is it automatic or must be added to  your-platform.cmake file?

TotoGaz:
> I think the problem is that GEOSX_TPL_DIR is not defined. Where should this be defined? Is it automatic or must be added to your-platform.cmake file?

If you do not define it, there is a fallback (something like `../thirdPartyLibs/install-your-platform-debug`).
If you've respected very precisely the tutorial, it should be OK. Otherwise you can define it with `python scripts/config-build.py -hc host-configs/your-platform.cmake -bt Release -DGEOSX_TPL_DIR=...`.
The log should tell you what `cmake` uses for `GEOSX_TPL_DIR`

karimifard:
I use the same folder structure as described by the tutorial.
GEOSX_TPL_DIR is not defined and adding it to the script via -D didn't work.
Let's check it when you are on campus.

TotoGaz:
@karimifard You can review the patch again 🎉

PR #1779, opened 2022-02-15
Combined multifluid derivatives to reduce fluid kernel wrapper size

francoishamon:
As suggested [here](https://github.com/GEOSX/GEOSX/pull/1225#discussion_r781755352), this PR combines the multifluid derivatives (pressure, temperature, component fractions) into a single array to reduce the size of the fluid kernel wrapper and make sure that we can compile on GPU platforms in the thermal PRs #1755 and #1225.

I am getting the following result [here](https://github.com/GEOSX/GEOSX/blob/1dc46ce69fc126f0cd492be3f9f78661a434bb5e/src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp#L442):
- develop: `sizeof(fluidWrapper)` = 3096 bytes
- this branch: `sizeof(fluidWrapper)` = 2136 bytes

So I am confident that I will be able to compile with the thermal option activated.

Requires rebaseline, but no numerical diffs: https://github.com/GEOSX/integratedTests/pull/193


TotoGaz:
> develop: sizeof(fluidWrapper) = 3096 bytes
> this branch: sizeof(fluidWrapper) = 2136 bytes

 😨 🎉 👍 🥇

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoromechanicsKernel.hpp
```diff
@@ -380,13 +380,14 @@ class Multiphase :
 
       real64 const dPhaseAmount_dP = dPorosity_dPressure * m_fluidPhaseSaturation( k, ip ) * m_fluidPhaseDensity( k, q, ip )
                                      + porosityNew * (m_dFluidPhaseSaturation_dPressure( k, ip ) * m_fluidPhaseDensity( k, q, ip )
-                                                      + m_fluidPhaseSaturation( k, ip ) * m_dFluidPhaseDensity_dPressure( k, q, ip ) );
+                                                      + m_fluidPhaseSaturation( k, ip ) * m_dFluidPhaseDensity( k, q, ip, Deriv::dP ) );
```
TotoGaz, line 383: In terms of memory layout, is having `Deriv::dP` as last component efficient?

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBaseKernels.hpp
```diff
@@ -344,12 +342,12 @@ class PhaseVolumeFractionKernel : public PropertyKernelBase< NUM_COMP >
...
 
       dPhaseVolFrac_dPres[ip] =
-        (dPhaseFrac_dPres[ip] - phaseVolFrac[ip] * dPhaseDens_dPres[ip]) * phaseDensInv;
+        (dPhaseFrac[ip][Deriv::dP] - phaseVolFrac[ip] * dPhaseDens[ip][Deriv::dP]) * phaseDensInv;
 
       for( integer jc = 0; jc < numComp; ++jc )
       {
         dPhaseVolFrac_dComp[ip][jc] =
-          (dPhaseFrac_dComp[ip][jc] - phaseVolFrac[ip] * dPhaseDens_dComp[ip][jc]) * phaseDensInv;
+          (dPhaseFrac[ip][Deriv::dC+jc] - phaseVolFrac[ip] * dPhaseDens[ip][Deriv::dC+jc]) * phaseDensInv;
```
TotoGaz, line 350: I bet this `Deriv::dC+jc` is a trick to put keep the same dimension for all your arrays?

TotoGaz (commented):
Good job so far, can't wait to see if anything can be done about https://github.com/GEOSX/GEOSX/pull/1225#discussion_r797828122 🤞

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoromechanicsKernel.hpp
francoishamon: Here I just kept what we have now in the code (automatic permutation of the layout on the GPU). On the GPU, we permute the data layout as requested [here](https://github.com/GEOSX/GEOSX/blob/1dc46ce69fc126f0cd492be3f9f78661a434bb5e/src/coreComponents/constitutive/fluid/layouts.hpp#L56) to make the cell number last. This PR by Sergey discussed (and did) the implementation: https://github.com/GEOSX/GEOSX/pull/1463

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBaseKernels.hpp
francoishamon: Yes, in the current version of the PR, the derivatives are grouped and ordered as follows:

- 0: derivative wrt pressure
- 1: derivative wrt temperature
- 2: first derivative wrt component fraction

With this approach, there is only one array storing derivatives for a given physical quantity. For instance, for the phase component fractions (of size: `numPhases x numComps`), there is one array of derivatives of size `numPhases x numComps x numDofs`, with `numDofs = numComps + 2`.

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoromechanicsKernel.hpp
klevzoff: On CPU platforms, this places all derivatives in a single cell in contiguous memory, which is optimal because the access pattern for this data form a single core is basically sequential and values will be cached (with the caveat that we don't do anything to align cell data on cacheline size, so [false sharing](https://en.wikipedia.org/wiki/False_sharing) can occur; but that's a different story).

With CUDA, as @francoishamon mentioned, we permute the layout such that the unit-stride (contiguous) dimension is the one corresponding to cell index `k`, which results in perfectly coalesced access pattern optimal for GPUs.

src/coreComponents/constitutive/fluid/MultiFluidUtils.hpp
```diff
@@ -55,47 +55,41 @@ template< typename T, int DIM, int USD, int USD_DC >
 struct MultiFluidVarSlice
 {
   internal::ArraySliceOrRef< T, DIM, USD > value;        /// variable value
-  internal::ArraySliceOrRef< T, DIM, USD > dPres;        /// derivative w.r.t. pressure
-  internal::ArraySliceOrRef< T, DIM, USD > dTemp;        /// derivative w.r.t. temperature
-  internal::ArraySliceOrRef< T, DIM + 1, USD_DC > dComp; /// derivative w.r.t. composition
+  internal::ArraySliceOrRef< T, DIM + 1, USD_DC > derivs; /// derivative w.r.t. pressure, temperature, compositions
```
klevzoff, line 58: This is where (along with `View` class below) where we would potentially add `dP`, `dT` and `dC` methods as described in https://github.com/GEOSX/GEOSX/pull/1225#discussion_r797828122 and https://github.com/GEOSX/GEOSX/pull/1225#discussion_r807554325 if we agree on that idea. The raw slice `derivs` need to remain accessible though, because it must be passed directly to `applyChainRule`.

This would somewhat decrease the flexibility of these classes however. For example, currently they are usable both with and without temperature derivatives - the owner decides whether or not it is present by resizing and then using the `derivs` array accordingly. By adding a `dT()` method we are forcing the temperature derivative to be always present. These classes could also be used outside of `MultiFluid` classes, anywhere one need to bundle values with a bunch of derivatives - for example in `RelativePermeabilityBase` and `CapillaryPressureBase`, physics solvers, etc.

For that reason I'd propose a refactoring in which this family of classes is moved into some shared place (`common` probably?) under a different name (`ValuesAndDerivatives[View|Slice]`?), while here we create MultiFluid-specific wrappers that either inherit from or contain the above classes and add the specialized access methods. I would propose maybe for this to be a separate PR.

I am also not opposed to just keeping the use of `Deriv::dP`, etc. as explicit offsets for the foreseeable future.

klevzoff (approved):
I like that that, in addition to solving a technical problem, this PR is a net decrease in code size, and readability is not affected much. In fact, I'm glad to get rid of `dX_dY` names - there was always something unsettling about those underscores violating our naming rules.

src/coreComponents/constitutive/fluid/MultiFluidBase.hpp
```diff
@@ -636,49 +581,41 @@ void
...
                                       phaseFrac,
                                       phaseCompFrac,
-                                      dPhaseDens_dGlobalCompFrac,
-                                      dPhaseVisc_dGlobalCompFrac );
+                                      dPhaseDens,
+                                      dPhaseVisc );
 }
 
-template< integer maxNumComp, integer maxNumPhase >
+template< integer maxNumDof, integer maxNumPhase >
```
TotoGaz, line 604: Here and after, I would definitely `doxygen` the functions.
Please do not forget the `@tparam maxNumDof` and specially which `dof` we are talking about.

src/coreComponents/codingUtilities/Utilities.hpp
```diff
@@ -188,11 +188,14 @@ void forEachArgInTuple( std::tuple< Ts ... > const & tuple, F && func )
 
 template< typename VEC1, typename VEC2 >
 GEOSX_HOST_DEVICE
-void copy( integer const N, VEC1 const & v1, VEC2 const & v2 )
+void copy( integer const N,
```
TotoGaz, line 191: Sorry to bother you with this, but documenting why we need this `firstDeriv` could be helpful (in which context we need this, _e.g._)

More important, instead of providing `firstDeriv`, couldn't `v2` be a `range`/`slice` instead? That would keep this function more _mathematical_ and less _implementation aware_.

src/coreComponents/unitTests/constitutiveTests/testCO2BrinePVTModels.cpp
```diff
@@ -1285,12 +1265,12 @@ TEST_F( CO2InternalEnergyTest, CO2InternalEnergyMassValuesAndDeriv )
...
-  localIndex counter = 0;
-  for( localIndex iComp = 0; iComp < 3; ++iComp )
+  integer counter = 0;
+  for( integer iComp = 0; iComp < 3; ++iComp )
   {
-    for( localIndex iPres = 0; iPres < 3; ++iPres )
+    for( integer iPres = 0; iPres < 3; ++iPres )
     {
-      for( localIndex iTemp = 0; iTemp < 3; ++iTemp )
+      for( integer iTemp = 0; iTemp < 3; ++iTemp )
```
TotoGaz, line 1273: It's nice that you are changing these `localIndex` into `integer` but I do think that you are wasting your energy.
If we really want to do this, then a carefully written `search`/`replace` on the whole code base will do the duty.

TotoGaz (commented):
Moving forward!

src/coreComponents/constitutive/fluid/MultiFluidUtils.hpp
TotoGaz: I perceive this pattern as really similar (in nature) to the `extrinsicMeshData` wrappings of https://github.com/GEOSX/GEOSX/pull/1750 (the current situation brings additional complexity).

There would be an extremely carefully designed implementation (in the current case, some `CompositionalVarView`, in the case of https://github.com/GEOSX/GEOSX/pull/1750, the `*StencilAccessors`), that may be "_specialized_" depending on the context. That way, if you do not want the `dT`, then don't ask your smart implementation for space to store `dT`, and don't provide the `dT` accessors. Therefore why would we lose flexibility?

For example, it's fine for the underlying implementation to provide variadic accessors `real64 & dT( INDICES ... indices )` but providing a more constrained `real64 & dT( int a, int b, int c )` member function will be simpler in the end.

src/coreComponents/unitTests/constitutiveTests/testCO2BrinePVTModels.cpp
TotoGaz: I heard the discussion this morning about `integer` loops but I still do not fully understand. Take this loop
```c++
for( integer ip = 0; ip < numPhase; ++ip )
{
  phaseMolecularWeight[ip] = props.getMolecularWeight( m_phaseTypes[ip] ).value;
  phaseMolecularWeight[ip][Deriv::dP] = props.getMolecularWeight( m_phaseTypes[ip] ).dP;
  [...]
```
I thing the indexing type of `phaseMolecularWeight` or `phaseMolecularWeight` is `localIndex`. So what is the point at using `integer` instead of `phaseMolecularWeight::IndexType` which appear to be a `localIndex`? Performances really get improved there?

klevzoff: For me this is not about performance but semantics. A `localIndex` is a counterpart to `globalIndex` where "local" and "global" refer, generally speaking, to parallel partition of the physical domain and mesh. Therefore these terms make sense when indexing elements, nodes, faces, etc which can be either local or global. Model properties like number of phases or components are in a different category - they are much smaller, and the notion of local/global does not apply to them. Ideally we could introduce separate aliases for them, e.g.
```
using phaseIndex = /* some type */;
using componentIndex = /* some type */;
```
but I thought this might be a bit overkill. Therefore I started changing, and suggesting to others in reviews (mainly @francoishamon who did most of the work), those indices to a generic `integer`. As a secondary concern, we also get to save some bytes in kernel wrapper objects, which sometimes need to carry phase/component indices as members. But it's a really minor saving compared to what this PR achieves at scale.

Now, `phaseMolecularWeight::IndexType` is indeed currently a `localIndex`. But that is also wrong in my opinion! For certain "small" arrays (e.g. `m_componentMolecularWeight`) we could be using a smaller type such as `integer` as index type, and it would also help a bit with reducing the size of kernel arguments. I don't advocate necessarily that we implement this change, because the increasing complexity in the type system has cognitive costs (there would be `array1d<T>` and some `array1dSmallIndex<T>` to keep in mind). But neither am I concerned about integer promotions taking place, just like I'd never force someone to change
```c++
std::vector< double > v;
v[0] ...
```
to
```c++
v[ std::vector< double >::size_type(0) ] ...
```

TotoGaz: OK, thanks for the explanations. Some kind of Hungarian stuff on types, in a nutshell?

src/coreComponents/constitutive/fluid/layouts.hpp
```diff
@@ -31,6 +32,17 @@ namespace constitutive
 namespace multifluid
 {
 
+/// indices of pressure, temperature, and composition derivatives
+struct Deriv
```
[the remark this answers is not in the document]
francoishamon: Done, I changed `struct Deriv` to `struct DerivativeOffset` and used the `using` where I needed.

src/coreComponents/constitutive/fluid/BlackOilFluid.hpp
```diff
@@ -570,36 +560,19 @@ BlackOilFluid::KernelWrapper::
...
-  }
+  LvArray::forValuesInSlice( phaseFraction.value, []( real64 & val ){ val = 0.0; } );
+  LvArray::forValuesInSlice( phaseFraction.derivs, []( real64 & val ){ val = 0.0; } );
+  LvArray::forValuesInSlice( phaseCompFraction.value, []( real64 & val ){ val = 0.0; } );
+  LvArray::forValuesInSlice( phaseCompFraction.derivs, []( real64 & val ){ val = 0.0; } );
 
   // 2. Check feed first, and if only water is present (e.g., water inj), then skip
 
-  if( zw >= 1. - minForPhasePresence )
+  if( zw >= 1.0 - minForPhasePresence )
```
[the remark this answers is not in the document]
francoishamon: Purely for personal convenience, I think it looks better like this.

src/coreComponents/constitutive/fluid/MultiFluidBase.hpp
francoishamon: I added the doxygen above in the class where the function is declared, with some information about which dof (pressure, temperature, and component fractions) we are talking about.

src/coreComponents/constitutive/fluid/BlackOilFluid.hpp
```diff
@@ -570,36 +560,19 @@ BlackOilFluid::KernelWrapper::
...
-        {
-          phaseCompFraction.dComp[ip][ic][jc] = 0.0;
-        }
-      }
-    }
-  }
+  LvArray::forValuesInSlice( phaseFraction.value, []( real64 & val ){ val = 0.0; } );
+  LvArray::forValuesInSlice( phaseFraction.derivs, []( real64 & val ){ val = 0.0; } );
+  LvArray::forValuesInSlice( phaseCompFraction.value, []( real64 & val ){ val = 0.0; } );
+  LvArray::forValuesInSlice( phaseCompFraction.derivs, []( real64 & val ){ val = 0.0; } );
```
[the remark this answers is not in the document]
klevzoff: Yes, that's the best we can do without making changes to `LvArray` (which is wise to avoid for this PR), as slices do not define assignment operators, intentionally or not.

src/coreComponents/codingUtilities/Utilities.hpp
francoishamon: Yes, as we discussed, it would be nice to passe a slice or range, but I don't know how to do it. 
Here for example:
https://github.com/GEOSX/GEOSX/blob/3e72aadf62b7083dd7f434052f3d90c8f3470230/src/coreComponents/constitutive/fluid/MultiFluidBase.hpp#L722
my `phaseFrac.derivs[ip]` is already a slice of size `numComp+2`, but I don't need first two entries (derivatives with respect to pressure and temperature). I would need a way to skip them, maybe by doing: 
```
applyChainRuleInPlace( numComp, dCompMoleFrac_dCompMassFrac, phaseFrac.derivs[ip][Deriv::dC], work );
```
Now that I am writing this, I should probably try that to see if it works.

For now, I added the Doxygen documentation in the `applyChainRule` function to explain why we have to pass `Deriv::dC`.

klevzoff: In this case `v2` represents a general single-subscriptable entity, which may be a local array or an `ArraySlice`. The latter may not be contiguous in memory, and in that case precomputing an offset pointer is not safe or correct. So we have to rely on the type's `operator[]` to do the job at all times.

Theoretically it's possible to write a `SubSlice` type which represents a sub-selection of an `ArraySlice` along one or more dimensions. The type would store the offset(s) internally and apply as needed. But we don't have it yet.

src/coreComponents/constitutive/fluid/MultiFluidUtils.hpp
francoishamon: I would vote for the following plan:
1. We merge this PR so that I can then finish and merge #1755 using the derivative format implemented in this PR
2. Once #1755 is merged, we work on implementing the `ValuesAndDerivatives` getters for the constitutive models and solvers

Even though this plan implies a transition period during which `Deriv::dP` etc will be used, it will allow me to start testing #1225 soon, which is important.

Regarding the idea of implementing getters (`dP()`, `dT()`, `dC()`) to get the derivatives in the kernels, will it slow down the code a little bit? Or not significantly?

klevzoff: > That way, if you do not want the dT, then don't ask your smart implementation for space to store dT, and don't provide the dT accessors. Therefore why would we lose flexibility?

What I meant is, if we just added `dT()` getter to this class, it would become unusable in context where there is no temperature dependence. But of course as you said, if we provide appropriately specialized wrappers for different contexts, there will be no issues :+1: 

> I would vote for the following plan:

Sounds good to me!

> Regarding the idea of implementing getters (`dP()`, `dT()`, `dC()`) to get the derivatives in the kernels, will it slow down the code a little bit? Or not significantly?

I expect them to be inlined 100% of the time, so no slowdown, if only a tiny compilation time hit.

TotoGaz: Let's go! 🚗 💨 

> I would vote for the following plan:

If I were you, I would open an issue with multiple tick boxes, references to meaningful comments, etc.

Issue #1776, opened 2022-02-15
[Feature] Make XML preprocessor MPI-aware


klevzoff:
**What is the requested feature?**
Would be great if XML preprocessor was safe for launching via `mpirun`.

**Is your request related to a specific problem?**
I started using preprocessor heavily (it's great for maintaining families of test cases!). One thing that popped up is that, while `geosx_preprocessed` is a safe replacement for `geosx` in serial, it's not safe for launching via `mpirun` because every rank tries to write the preprocessed file at the same time. It sort of still works fine in practice with small number of processes, but it's subject to a data race, where some ranks could already launch `geosx` and try to consume the file while others are still overwriting it. To be clear, I've yet to observe this in practice with `geosx` specifically, but the XML validation script is already throwing exceptions.

**Describe the solution you'd like**
Perhaps `xml_processor.py` could take an (optional) dependency on `mpi4py`, detect when being called from an MPI job and have only one rank do the preprocessing and output of resulting XML, locking other ranks on a barrier until completion.

**Describe alternatives you've considered**
One can run the preprocessor manually and then launch the MPI job with only `geosx`. That's slightly less convenient.

cssherman:
@klevzoff - I've just ran into this myself, and agree entirely.  I'm not sure how to do this under bash, but if we were to require mpi4py as a pre-requisite for geosx_xml_tools, I could limit the processing to the root rank.  I could probably make this an optional dependency if this package causes issues.

TotoGaz:
Dealing with this properly using `mpi4py` could be interesting _w.r.t._ to `pygeosx` too?!

cssherman:
@TotoGaz - Exactly.  This is what I've done in pygeosx_tools.xml  (https://github.com/GEOSX/GEOSX/blob/develop/src/coreComponents/python/modules/pygeosx_tools_package/pygeosx_tools/xml.py#L1)

cssherman:
While testing this approach out, I've found that importing mpi4py causes some odd issues.  After preprocessing completes, when I try to call GEOSX, I get errors that look like this:

srun: error: quartz2: tasks 0-35: Exited with exit code 141

Thsi I'm suspicious that this is due to the way python is handling the mpi environment.  Be default mpi4py attempts to initialize the environment when we request the problem rank, and tries to finalize it when it destructor is called.  Any thoughts @TotoGaz ?

klevzoff:
@cssherman I think the following way of importing might help:
```py
import mpi4py
[...]

PR #1781, opened 2022-02-17
Attempting to make the xml processing mpi aware

cssherman:
This attempts to address #1776


scripts/automatic_xml_preprocess.sh
```diff
@@ -1,13 +1,16 @@
...
-# Parse and modify input arguments
+
+# Parse input arguments
 INPUT_ARGS=""
 INPUT_COUNTER=0
 COMPILED_XML_NAME=""
 COMPILED_XML_NAME_OVERRIDE=""
 PARAMETER_ARGS=""
 NEW_ARGS=""
+USE_PYGEOSX=1
```
cssherman, line 11: If we would like to require pygeosx to be installed, then this script could be greatly simplified

klevzoff: I haven't been unable to build pygeosx so far, because it hard codes python version 3.8 and our cluster only has 3.6

https://github.com/GEOSX/GEOSX/blob/1414a7dc5d572b5852bcd5654f3c5e3954ffeec3/src/pygeosx/CMakeLists.txt#L25

src/coreComponents/python/modules/geosx_xml_tools_package/geosx_xml_tools/main.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+def preprocess_parallel():
+    """
+    MPI aware xml preprocesing
+    """
+    # Process the xml file
+    from mpi4py import MPI
+    comm = MPI.COMM_WORLD
+    rank = comm.Get_rank()
```
cssherman, line 40: We need to have separate serial and parallel (called using pygeosx) methods because using ``mpi4py`` seems to break the MPI environment for subsequent calls to GEOSX.  I haven't been able to track down why this is happening unfortunately.

scripts/automatic_xml_preprocess.sh
cssherman: For pygeosx to work properly, you need to build a version of python (and certain dependencies) with exact compiler / mpi matches.  I used spack to do so for the Lassen build

TotoGaz (commented):
Hello, I do not have a lot to say about this PR, but I do not know the key component to peruse.

src/coreComponents/python/modules/geosx_xml_tools_package/geosx_xml_tools/main.py
```diff
@@ -0,0 +1,128 @@
...
+                    if target_file_exists:
+                        if (os.path.getmtime(fname) > target_file_edit_time):
+                            break
+                    else:
+                        if os.path.isfile(fname):
+                            break
+                    time.sleep(0.1)
+                print('File detected')
+        return wait_for_file_write_rank_0_decorator
+    return wait_for_file_write_rank_0_inner
```
cssherman, line 61: @klevzoff - With some experimentation, I think I found a decent backup option.  This decorator function will check for the current rank by looking at environment variables, attempt to do the preprocessing on rank 0, and wait for the file to be created/updated on the remaining processes.  At worst, this should fall back to the current method, which produces (seemingly benign) file warnings in python.

klevzoff: Hey @cssherman, I'm testing this on my local machine, and with 4 or more ranks I occasionally have the job stall for the entire timeout value (100 s). I think this might happen when rank 0 manages to complete writing the file *before* one of the other ranks sets `target_file_edit_time` - in that case the modification time stamp doesn't ever change anymore, and that rank is stuck waiting.

cssherman: Interesting.  I think that the appropriate thing here is to add a bit of room for the time check.  This would add an implicit maximum frequency for writing these files, but I don't think that would be an issue.  Perhaps something like this:

`target_file_edit_time = os.path.getmtime(fname) - 10`

src/coreComponents/python/modules/geosx_xml_tools_package/geosx_xml_tools/main.py
```diff
@@ -0,0 +1,133 @@
...
+
+            target_file_exists = os.path.isfile(fname)
+            target_file_edit_time = 0
+            if target_file_exists:
+                target_file_edit_time = os.path.getmtime(fname)
+
+                # Variations in thread startup times may mean the file has already been processed
+                # If the last edit was done within the specified time, then allow the thread to proceed
+                if (abs(target_file_edit_time - time.time()) < max_startup_delay):
+                    target_file_edit_time = 0
```
cssherman, line 49: @klevzoff - After giving it some thought, this should work

klevzoff (approved):
This final version works great (at least on up to 64 ranks that I tested). Thanks for taking care of this!

scripts/automatic_xml_preprocess.sh
```diff
@@ -50,29 +53,38 @@ do
...
-   if [ ! -z "$COMPILED_XML_NAME_OVERRIDE" ]
+# Check for pygeosx
+if [ "$USE_PYGEOSX" -eq "1" ]
+then
+   if [ -f $PYGEOSX ]
    then
-      COMPILED_XML_NAME=$COMPILED_XML_NAME_OVERRIDE
+      echo "Using pygeosx to preprocess the xml file"
+   else
+      echo "Pygeosx installation not found... reverting to non-pygeosx version"
```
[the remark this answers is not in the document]
cssherman: Great! I'll work on cleaning up the messaging and add it to the queue

src/coreComponents/python/modules/geosx_xml_tools_package/geosx_xml_tools/main.py
```diff
@@ -0,0 +1,133 @@
...
+                              parameter_override=args.parameters)
+    if not compiled_name:
+        if args.compiled_name:
+            compiled_name = args.compiled_name
+        else:
+            raise Exception('When applying the preprocessor in parallel (outside of pygeosx), the --compiled_name argument is required')
+
+    # Note: the return value may be passed to sys.exit, and cause bash to report an error
+    # return format_geosx_arguments(compiled_name, unknown_args)
+    print(compiled_name, flush=True)
```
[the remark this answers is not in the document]
cssherman: I can modify these to use a proper logging function, which should help to avoid any versioning issues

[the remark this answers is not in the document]
cssherman: Although on second thought, this print statement is required for some workflows.  Dropping the flush argument should make it 2.x friendly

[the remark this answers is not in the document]
TotoGaz: Yes, if you can use std module `logging` https://docs.python.org/3/howto/logging.html

PR #1785, opened 2022-02-19
Fix face pressure BC with variable permeability

klevzoff:
This PR refactors `BoundaryStencil` and `FaceDirichletBCKernel` to follow what was done for regular TPFA flux computations: the stencil only stores the geometric weight (area/distance) for the cell-face pair, and the transmissibility computation happens in-kernel using updated permeability.

In particular, this fixes a bug I recently discovered: if face-based Dirichlet BC for `SinglePhaseFVM` were used with a non-constant permeability model (e.g. Carman-Kozeny), the entire model would not work because transmissibility precomputation used dummy values of permeability before they are computed for the first time.

To support the changes, I updated `staircase_3d` model to use `CarmanKozenyPermeability` instead of constant. It's set up to result in roughly the same values as previous constant ones (1.6e-16 vs 2e-16), but of course now slightly varying with the pressure distribution as it changes.

I also refactored and cleaned up the stencils a bit to reduce duplication and removed `coefficientModelNames` as it is no longer used. This will require a rebaseline of all tests using stencils.

Rebaseline PR: https://github.com/GEOSX/integratedTests/pull/191


src/coreComponents/finiteVolume/SurfaceElementStencil.hpp
```diff
@@ -44,48 +44,14 @@ namespace geosx
...
-  /// Number of points the flux is between (normally 2)
-  static localIndex constexpr NUM_POINT_IN_FLUX = 6;
-
-  /// Maximum number of points in a stencil
-  static localIndex constexpr MAX_STENCIL_SIZE = 6;
-
-  /// Maximum number of connections in a stencil
-  static localIndex constexpr MAX_NUM_OF_CONNECTIONS = MAX_STENCIL_SIZE * (MAX_STENCIL_SIZE - 1) / 2;
-};
+using SurfaceElementStencilTraits = StencilTraits< ArrayOfArrays, 6, 6, 6 * ( 6 - 1) / 2 >;
```
francoishamon, line 49: Maybe we should add a small comment here to remind future developers that `6 * ( 6 - 1) / 2` is really `maxStencilSize * (maxStencilSize - 1) / 2`. That was self-documented by the variable names in the previous implementation, but that this information may be lost with the new `localIndex` templates

src/coreComponents/finiteVolume/TwoPointFluxApproximation.cpp
```diff
@@ -855,6 +850,9 @@ void TwoPointFluxApproximation::computeBoundaryStencil( MeshLevel & mesh,
   arrayView2d< localIndex const > const & elemList           = faceManager.elementList();
   arrayView2d< real64 const, nodes::REFERENCE_POSITION_USD > const & nodePosition = nodeManager.referencePosition();
 
+  arrayView1d< real64 const > const & transMultiplier =
+    faceManager.getReference< array1d< real64 > >( m_coeffName + viewKeyStruct::transMultiplierString() );
```
francoishamon, line 854: At some point, we should also remove the `m_coeffName` member, but this can be done later (after the thermal work is merged maybe). 

If I remember well, this class data member is only used to register the `transMultiplier` and attach it to the physical coefficient it will multiply. But this  prevents us from easily reusing the existing `computeWeights` function if we ever want to apply the multiplier differently on different physics (not sure there is a good use case, but we never know). So I would vote for removing the `m_coeffName` member, moving the multiplier to the solver, and making the transmissibility multiplier an argument of `computeWeights`.

francoishamon (approved):
Great, thanks @klevzoff

src/coreComponents/finiteVolume/SurfaceElementStencil.hpp
klevzoff: Agreed, I'm expanding the documentation for this type alias.

src/coreComponents/finiteVolume/TwoPointFluxApproximation.cpp
klevzoff: Yes, we should do this. The only problem is that to access the multiplier from the physics solver kernel we need the original connector (e.g. face) index, which is currently not stored with the stencil, but it won't be a problem to add it.

PR #1790, opened 2022-02-23
Initial HIP/ROCM Support

### LINKED ISSUE #2135: Enforce dispatch policies used in generated kernels.
Change the policy at e.g. 

https://github.com/GEOSX/GEOSX/blob/7f543eaf2ae3c9ce9bead7678ade167fa7bfbd4e/src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.cpp#L268

And compile / link the code on a platform with gpu dispatch, and the code will fail to link with the easiest error message ever to parse.

This is due to our generated kernels having their policies explicitly specified in e.g. 

https://github.com/GEOSX/GEOSX/blob/7f543eaf2ae3c9ce9bead7678ade167fa7bfbd4e/src/coreComponents/physicsSolvers/solidMechanics/kernels/SolidMechanicsKernels.cmake#L7-L10

This probably won't be an immediate issue until / unless we work to change these depending on the platform. Since there is currently no enforcement that the generated kernels have policies matching the usage elsewhere in the codebase.

wrtobin:
Requires: https://github.com/GEOSX/LvArray/pull/248

Requires: https://github.com/GEOSX/hdf5_interface/pull/10

closes #2135


TotoGaz:
@wrtobin do we have an associated TPL PR?

wrtobin:
> @wrtobin do we have an associated TPL PR?

@TotoGaz Not yet as we're starting to move to Spack and spack environments to do our TPL build and install and host-file generation. 

@bmhan12 is working on that process, I might be stepping over to push that forward after this goes in. But once we have that in place it will include a crusher spack environment where installation of the TPLs should be possible via only:
```
spack env activate crusher
spack install -j 8
```

Here is a brief set of spack specs that allow the minimal install and usage of hip on a platform that support it:
```
conduit@0.7.2%cce@13.0.1~hdf5_compat+silo ^hdf5@1.12.1+hl

parmetis%cce@13.0.1,
metis%cce@13.0.1,

pugixml%cce@13.0.1,

fmt@8.0.1%cce@13.0.1,

suite-sparse@5.10.1%cce@13.0.1~cuda~graphblas~openmp+pic~tbb,

conduit@0.7.2%cce@13.0.1~hdf5_compat+silo ^hdf5@1.12.1+hl,

hypre@develop%cce@13.0.1~int64+rocm+superlu-dist amdgpu_target=gfx90a,
hypre@develop%cce@13.0.1+int64+superlu-dist,

chai@2022.03.0~examples+raja~rocm ^camp@0.2.2 ^raja@2022.03.0 ^umpire@2022.03.0,
chai@2022.03.0~examples+raja+rocm amdgpu_target=gfx90a ^camp@0.2.2 ^raja@2022.03.0 ^umpire@2022.03.0
```

You would need to change `amdgpu_target=[TARGET]` and possibly add `^hip@4.5.2 ^hsa-rocr-dev@4.5.2 ^llvm-amdgpu@4.5.2` (changing hip version numbers) as well. 

Really this build is going to be taking a while to mature as the platforms are still early and changing often.

src/coreComponents/common/CMakeLists.txt
```diff
@@ -33,24 +33,12 @@ set( common_sources
      initializeEnvironment.cpp
    )
 
-set( dependencyList lvarray pugixml RAJA chai conduit::conduit fmt::fmt )
+set( dependencyList ${commonDeps} lvarray pugixml::pugixml RAJA chai conduit::conduit fmt::fmt )
```
rrsettgast, line 36: what is up with the scoping here?

src/coreComponents/common/initializeEnvironment.cpp
```diff
@@ -201,6 +205,19 @@ void setupCaliper( cali::ConfigManager & caliperManager,
 #endif
   adiak::value( "CUDA runtime version", cudaRuntimeVersion );
   adiak::value( "CUDA driver version", cudaDriverVersion );
+
+  // HIP info
+  int hipRuntimeVersion = 0;
+  int hipDriverVersion = 0;
```
rrsettgast, line 211: Should these be set?

src/coreComponents/unitTests/finiteVolumeTests/CMakeLists.txt
```diff
@@ -6,18 +6,14 @@ set( gtest_geosx_tests
      testMimeticInnerProducts.cpp
    )
 
-set( dependencyList gtest )
+set( dependencyList ${commonDeps} gtest )
```
rrsettgast, line 9: acceleratorDeps?

src/coreComponents/physicsSolvers/surfaceGeneration/SurfaceGenerator.cpp
```diff
@@ -39,7 +39,7 @@ namespace geosx
 using namespace dataRepository;
 using namespace constitutive;
 
-constexpr real64 SurfaceGenerator::m_nonRuptureTime;
+//constexpr real64 SurfaceGenerator::m_nonRuptureTime;
```
rrsettgast, line 42:
```suggestion
```

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.cpp
```diff
@@ -628,6 +628,8 @@ real64 SolidMechanicsLagrangianFEM::explicitStep( real64 const & time_n,
 
     CommunicationTools::getInstance().asyncSendRecv( domain.getNeighbors(), m_iComm, true, packEvents );
 
+    waitAllDeviceEvents( packEvents );
```
rrsettgast, line 631: guard

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.cpp
```diff
@@ -433,7 +433,7 @@ void SinglePhaseBase::computeHydrostaticEquilibrium()
     arrayView1d< real64 > const pres =
       subRegion.getReference< array1d< real64 > >( extrinsicMeshData::flow::pressure::key() );
 
-    forAll< parallelDevicePolicy<> >( targetSet.size(), [=] GEOSX_HOST_DEVICE ( localIndex const i )
+    forAll< parallelHostPolicy >( targetSet.size(), [=] GEOSX_HOST_DEVICE ( localIndex const i )
```
rrsettgast, line 436: ??

src/coreComponents/mesh/generators/ParMETISInterface.hpp
```diff
@@ -27,6 +27,13 @@ namespace geosx
 namespace parmetis
 {
 
+#if defined(GEOSX_USE_DEVICE)
```
rrsettgast, line 30: more specific

rrsettgast (approved):

src/coreComponents/common/CMakeLists.txt
wrtobin: Between pugixml v1.8 (currently in our tpls) and v1.11.4 (installed by spack on crusher), they added scoping to heir exported targets, similar to conduit and fmt later in the line.

I think this is best handled in the tpl discovery scripts, so we can use our tpls uniformly throughout the rest of the codebase without scoping. I'll add logic there to our currently scoped tpl import targets to alias them as unscoped for consistency.

src/coreComponents/common/initializeEnvironment.cpp
wrtobin: This is just mimicking the same calls for cuda just above for consistency. If we want to change both we can address it in a quick follow-up.

src/coreComponents/constitutive/contact/ContactBase.hpp
```diff
@@ -71,9 +71,10 @@ class ContactBaseUpdates
    * @return The hydraulic aperture that is always > 0
    */
   GEOSX_HOST_DEVICE
-  inline
-  virtual real64 computeHydraulicAperture( real64 const aperture,
-                                           real64 & dHydraulicAperture_dAperture ) const;
+  real64 computeHydraulicAperture( real64 const aperture, real64 & dHydraulicAperture_dAperture ) const
```
wrtobin, line 74: This had to be made non-virtual to avoid an issue in the cce device-pass where it would inappropriately inline a virtual function in a context it isn't able to by the C++ standard, which caused a segfault when using this function in device kernels since the "output param" wasn't appropriately bound to a valid memory location.

src/coreComponents/constitutive/contact/ContactBase.hpp
```diff
@@ -116,10 +117,14 @@ class ContactBaseUpdates
    * @return the updated traction
    */
   GEOSX_HOST_DEVICE
-  inline
+  GEOSX_FORCE_INLINE
   virtual void addPressureToTraction( real64 const & pressure,
```
wrtobin, line 121: Not sure why we don't see the same exact issue here.. since it is literally the same pattern that caused the bug noted above, yay cce.

PR #1792, opened 2022-02-28
`*packByIndex*` functions of Wrapper depends on `sizedFromParent() == 1`.

TotoGaz:
This PR solves https://github.com/GEOSX/GEOSX/issues/1810.

I have
- moved the `if( wrapper.sizedFromParent() )` check outside the `Wrapper` class.
- refactored duplicated code in `Group` and `Wrapper`
- reorganized code to get a clearer separation between the `wrapper` selection and their packing.
- changed `ObjectManagerBase::viewPackingExclusionList` to `ObjectManagerBase::getPackingExclusionList` that returns a `std::set< string >`; then use `STL` algorithm to filter out the excluded wrappers.

Some more refactoring could be done about the `packing` but stopping there looks reasonnable.


src/coreComponents/codingUtilities/Utilities.hpp
```diff
@@ -142,19 +142,19 @@ VAL findOption( mapBase< KEY, VAL, SORTED > const & map,
...
+  * @return The container with the keys.
+  */
+template< typename MAP, typename R = std::vector< typename MAP::key_type > >
+R mapKeys( MAP const & map )
 {
-  std::vector< KEY > keys;
-  std::transform( map.begin(), map.end(), std::back_inserter( keys ), [=]( auto const & p ){ return p.first; } );
+  R keys;
+  std::transform( map.begin(), map.end(), std::inserter( keys, keys.end() ),
+                  [=]( auto const & p ) { return p.first; } );
```
TotoGaz, line 157: Hi @klevzoff I somehow needed to be able to get the keys as a `set` so I ended up with this implementation.

But I'm not happy with this implementation since optional `R` depends on `MAP` because of the nested type `key_type` (and optional parameters go at the end).

So when I call with anything different than `std::vector< typename MAP::key_type >` I need to define `MAP` in the template call. I get like `mapKeys< decltype( m ), std::set< string > >( m )` which upsets me.

Do you have any idea?

klevzoff: How about making it a template template parameter? :smile: 
```c++
template< template< typename > class R, typename MAP >
R< typename MAP::key_type > mapKeys( MAP const & map )
{
  R< typename MAP::key_type > keys;
  ...
}
...
auto keyVec = mapKeys< std::vector >( map );
auto keySet = mapKeys< std::set >( map );
```
(I haven't tested this code)

TotoGaz: I'm fine with something like what you proposed, but that means that there is no more default.
I'm OK with that nevertheless...

klevzoff: > I did not know that we could provide "incomplete types" like `std::vector`! Good!

Yes, I keep forgetting it too and get surprised by this fact from time to time :laughing:

TotoGaz: Hmm, it appears that this does not work with `-std=c++14` 😢

klevzoff: Hmm, it can be worked around by adding
```c++
template< typename T >
using set = std::set< T >;

template< typename T >
using vector = std::vector< T >;
```
and then using those as parameters. But it's not as nice :disappointed: .

klevzoff: Alternative solution (note the added `...`):
```c++
template< template< typename ... > class R = std::vector, typename MAP >
R< typename MAP::key_type > mapKeys( MAP const & map )
{
  R< typename MAP::key_type > keys;
  std::transform( map.begin(), map.end(), std::inserter( keys, keys.end() ),
                  [=]( auto const & p ) { return p.first; } );
  return keys;
}
```

TotoGaz: Brilliant! I've just pushed the patch!

rrsettgast:
This diff is unreadable.

francoishamon (approved):
Thanks @TotoGaz, this looks like a good refactoring. I am not really in a position to make specific comments on your PR because my knowledge of the packing functions is still very shallow. So, my review should be taken with caution ;)

TotoGaz:
Thx @francoishamon This patch passes unit tests and integrated tests.

rrsettgast (commented):
There are some naming changes that are not clear from just reviewing the code. Can you summarize the changes in the call flow of the pack/unpack functions?

src/coreComponents/dataRepository/Group.cpp
```diff
@@ -316,63 +317,84 @@ void Group::initializePostInitialConditions()
...
 }
 
-localIndex Group::packSize( string_array const & wrapperNames,
-                            arrayView1d< localIndex const > const & packList,
-                            integer const recursive,
-                            bool onDevice,
-                            parallelDeviceEvents & events ) const
+template< bool DO_PACK >
+localIndex Group::packPrivate( buffer_unit_type * & buffer,
+                               string_array const & wrapperNames,
```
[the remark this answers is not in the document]
TotoGaz: It was a direct replacement of the previous function signature. I'll update.

src/coreComponents/dataRepository/Wrapper.hpp
```diff
@@ -912,6 +800,137 @@ class Wrapper final : public WrapperBase
...
+   * @param[in] withMetadata Whether to pack string metadata with the underlying data.
+   * @param[in] onDevice Whether to use device-based packing functions
+   *                     (buffer must be either pinned or a device pointer)
+   * @param[out] events A collection of events to poll for completion of async
+   *                    packing kernels ( device packing is incomplete until all
+   *                    events are finalized )
+   * @return The packed size.
+   */
+  template< bool DO_PACKING >
+  localIndex packTemplate( buffer_unit_type * & buffer,
```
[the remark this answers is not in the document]
TotoGaz: The previous pattern for `pack` (and `unpack`, it's the same) was having pure virtual `WrapperBase::pack`/`WrapperBase::packSize` member functions, delegating the implementations to `Wrapper< T >`.

To keep the symmetry with other code and remove some duplicated code, I've added a `WrapperBase::pack< DO_PACKING >` as a `public` member function. To keep the delegation pattern, I have pure virtual `WrapperBase::packSizePrivate`/`WrapperBase::packPrivate`.

Then `Wrapper` has to implement `WrapperBase::packSizePrivate`/`WrapperBase::packPrivate`. To remove code duplication I went on implementing `private` `WrapperBase::packTemplate< DO_PACKING >`, because of name conflict, which could admittedly be called `WrapperBase::packImpl< DO_PACKING >` (or something else)?

src/coreComponents/dataRepository/Wrapper.hpp
```diff
@@ -927,22 +946,8 @@ class Wrapper final : public WrapperBase
...
-//  return TV_ttf_format_raw;
-//}
-//
-//template int TV_ttf_display_type( geosx::dataRepository::Wrapper<int> const * wrapper );
-//
-//template< typename T >
-//void geosx::dataRepository::Wrapper<T>::tvTemplateInstantiation()
-//{
-//  TV_ttf_display_type<T>(this);
-//}
```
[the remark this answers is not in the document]
TotoGaz: OK, then I'll put it back with explanations so people do not remove it again 😉

src/coreComponents/dataRepository/Group.cpp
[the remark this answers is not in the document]
TotoGaz: done

src/coreComponents/dataRepository/Wrapper.hpp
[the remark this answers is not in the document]
TotoGaz: I've pushed a version with `packImpl` with is a little better I think. Is it OK?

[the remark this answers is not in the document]
rrsettgast: @TotoGaz I think that is a little more accurate of a name, don't you? Perhaps we should adopt this naming convention for this pattern in general.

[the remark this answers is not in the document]
TotoGaz: I do not think that this exact pattern (double indirection) actually happens too often (maybe it's the only one?).
I think the idea was to be able to call the `pack/unpack` from the `base` class without using some `CRTP`.

Sometime we can see some `packPrivate` (with a single indirection), which could be renamed into `packImpl`.

PR #1793, opened 2022-03-01
Restored compatibility checks in compositional solvers

francoishamon:
This PR just restores the compatibility checks between constitutive models in the compositional solvers (flow and wells). The validation functions were still in the code, but just not called anymore.

I also changed the names of two flags to make them more meaningful:
- `isThermal` (still unused in `develop`)
- `hasCapPressure`


src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -397,6 +398,64 @@ void CompositionalMultiphaseBase::initializePreSubGroups()
...
+      constitutiveUpdatePassThru( fluid, [&] ( auto & castedFluid )
+      {
+        bool const isFluidModelThermal = castedFluid.isThermal();
+        GEOSX_THROW_IF( m_isThermal && !isFluidModelThermal,
+                        GEOSX_FMT( "CompositionalMultiphaseBase {}: the thermal option is enabled in solver `{}`, but the fluid model `{}` is incompatible with the thermal option",
+                                   getName(), getName(), fluid.getName() ),
+                        InputError );
+        GEOSX_THROW_IF( !m_isThermal && isFluidModelThermal,
+                        GEOSX_FMT( "CompositionalMultiphaseBase {}: the thermal option is enabled in fluid model `{}`, but the solver options in `{}` are incompatible with the thermal option",
+                                   getName(), fluid.getName(), getName() ),
```
klevzoff, line 433: Nitpick: these error messages repeat the solver name twice, which is probably unneeded

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -397,6 +398,64 @@ void CompositionalMultiphaseBase::initializePreSubGroups()
   initializeAquiferBC( cm );
 }
 
+void CompositionalMultiphaseBase::validateConstitutiveModels() const
+{
+  GEOSX_MARK_FUNCTION;
+
+  DomainPartition const & domain = this->getGroupByPath< DomainPartition >( "/Problem/domain" );
```
klevzoff, line 405: While we do this in quite a few places now, I think it's best to limit this access pattern when we can, to make replacing it easier if we ever decide to. Here the function is always called from `initializePreSubGroups()` which already extracts the domain in this way, so we should just pass it as a parameter.

klevzoff (approved):
:+1:

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
francoishamon: Done, I removed one occurrence of the solver name in the latest commit

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
francoishamon: Yes, I agree. I now pass the `DomainPartition` as a parameter to the validation function

Issue #1794, opened 2022-03-03
[Other]  Figures generated with embedded python scripts cannot be displayed in readthedocs


jhuang2601:
For the following three tutorial examples, figures generated with embedded python scripts cannot be created/displayed properly in readthedocs,
However, by running these cases locally, figures can be generated with these scripts, as expected.
A solution is required to fix this issue.

https://geosx-geosx.readthedocs-hosted.com/en/latest/docs/sphinx/advancedExamples/validationStudies/deviatedPoroElasticWellbore/Example.html
https://geosx-geosx.readthedocs-hosted.com/en/latest/docs/sphinx/advancedExamples/validationStudies/deviatedPoroElasticWellbore/Example2.html
https://geosx-geosx.readthedocs-hosted.com/en/latest/docs/sphinx/advancedExamples/validationStudies/isothermalLeakyWell/Example.html

cssherman:
It looks like we need to add scipy to the readthedocs yaml and to fix indentation issues in some scripts

PR #1796, opened 2022-03-03
Moving sphinx documentation scripts to their own files

cssherman:
Addresses #1794


cssherman:
Here is the documentation test branch:  https://geosx-geosx.readthedocs-hosted.com/en/bugfix-sherman-fixembeddedscriptsindocumentation/docs/sphinx/advancedExamples/validationStudies/deviatedPoroElasticWellbore/Example.html

There don't seem to be any python-related errors occurring, but there are a bunch of other rst-related warnings (mismatched tags, etc) that should also be cleaned up

jhuang2601 (approved):

PR #1797, opened 2022-03-04
Fixed the computation of the initial total mass density

francoishamon:
This PR fixes a bug that I introduced in the compositional solver for the computation of the initial total mass density (only used in the computation of the body force in `MultiphasePoromechanicsKernel.hpp`). 

[Here](https://github.com/GEOSX/GEOSX/blob/8ff5ccf3871a7f9528e3765e58ca9b55bce3e2ea/src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp#L724) I was computing the initial density only for the first quadrature point, and therefore [here](https://github.com/GEOSX/GEOSX/blob/8ff5ccf3871a7f9528e3765e58ca9b55bce3e2ea/src/coreComponents/physicsSolvers/multiphysics/MultiphasePoromechanicsKernel.hpp#L323) the value of the body force increment was wrong. This PR moves the computation of the initial total mass density to the fluid constitutive model (as already done in `SingleFluidBase`), and now initializes the value for all quadrature points. 

Requires rebaseline (because the total mass density is saved in the restart file). Rebaseline done in https://github.com/GEOSX/integratedTests/pull/201


castelletto1 (approved):
👍

PR #1803, opened 2022-03-08
`Group::m_wrappers[name]` can be different from `Group::m_wrappers[name]->getName()`

TotoGaz:
This patch solves #1804.
- It removes the possibility of registering a `wrapper` with another `name`, making the inconsistency (hopefully) unreproducible. (Calls to previous version are updated accordingly.)
- Passes the unit tests
- Require rebaseline because of obvious
```
Group has a child 'shale_stress' in the file to compare but not the baseline file.
Group has a child 'stress' in the baseline file but not the file to compare.
...
```

I need comments about:
- [x] I've used `static_assert` instead of `std::enable_if` to prevent using other functions with wrappers. Looks OK to me since we're not in some SFINAE implementation here.
- [x] In the `static_assert` I use `std::is_base_of` and not `std::is_convertible`. What's best?
- [x] ~~Is there a way to check that children `shale_stress` and `stress` are identical?~~ I have no automatic process to do this, so I've made the comparison by hand for a bunch of cases and results were identical.

integrated tests update : https://github.com/GEOSX/integratedTests/pull/203


jhuang2601:
> You are running too fast @jhuang2601 😉 I'm trying a clean fix 🎉

@TotoGaz Oh, I thought this is just a temporary solution to spotlight the issue. A sophisticated fix will be followed.

TotoGaz:
> A sophisticated fix will be followed.

Sophisticated... Time will tell! 🤣

src/coreComponents/dataRepository/Group.hpp
```diff
@@ -1492,6 +1492,7 @@ template< typename T >
 Wrapper< T > & Group::registerWrapper( string const & name,
                                        std::unique_ptr< T > newObject )
 {
+  static_assert( not std::is_base_of< WrapperBase, T >::value, "This function should not be used for `WrapperBase`. Use the dedicated `registerWrapper` instead." );
```
rrsettgast, line 1495: `not`??? what is this??? python? Haha.. I think we discussed "not" using the alternative tokens at some point. (pun intended). We can revisit this, but my preference would be to stick with one throughout the code....and that would be `! , && , ||`. I think that these are probably implemented as macro constants, which makes me like them less.

https://en.wikipedia.org/wiki/C_alternative_tokens

TotoGaz: Yes I will change this. I got this habit of using those english words instead.
Otherwise, is the patch OK?

rrsettgast (approved):

rrsettgast:
> * [ ]  I've used `static_assert` instead of `std::enable_if` to prevent using other functions with wrappers. Looks OK to me since we're not in some SFINAE implementation here.

I think this is the better way. SFINAE is clunky. 

> * [ ]  In the `static_assert` I use `std::is_base_of` and not `std::is_convertible`. What's best?

This is a good check as far as I can see. I don't think you would want to use is_convertible here since the the criteria should be that T is derived from WrapperBase.

> * [ ]  Is there a way to check that children `shale_stress` and `stress` are identical?

You would have to check the pointers to the underlying `Wrapper<T>::m_data` object.

klevzoff (approved):

src/coreComponents/dataRepository/Group.hpp
TotoGaz: Fun fact, in the Wikipedia article it's said that "_The above-mentioned identifiers are [operator keywords](https://en.wikipedia.org/wiki/Operators_in_C_and_C%2B%2B) in the [ISO C++ programming language](https://en.wikipedia.org/wiki/ISO_C%2B%2B_programming_language) and do not require the inclusion of a header file._".

And code like `int not = 3;` won't compile. Which I never noticed.
```
error: expected unqualified-id before 'not' token
```

PR #1807, opened 2022-03-08
fix clang10 issue by modifying dependency in unitTests CmakeLists

labesse40:
This PR is about fixing unitTest issue with Clang10.

This issue was coming from the dependency on pygeosx in the CMakeLists of the unitTests. It has been fixed by changing the dependencies on the different pystructures (pygroup, pyfileIO, pysolver) 

Note : I managed to fix the issue on my laptop with GEOSX compiled with clang10. Needs to be tested on Quartz but I don't have access to the platform.


sframba (approved):
It seems that all the impacted unit tests have been changed.

rrsettgast (approved):

PR #1809, opened 2022-03-08
Refactored compositional flow unit tests

francoishamon:
This PR refactors the unit tests: `testCompMultiphaseFlow` and `testCompMultiphaseFlowHybrid` by removing some duplicate code and moving unified functions to `testCompFlowUtils.hpp`.

The goal is to be able to add a thermal unit test in https://github.com/GEOSX/GEOSX/pull/1225 without adding much code. I am doing the present refactoring in a separate PR to avoid clutter in https://github.com/GEOSX/GEOSX/pull/1225.

Ready for review.


src/coreComponents/unitTests/fluidFlowTests/testCompFlowUtils.hpp
```diff
@@ -212,6 +219,574 @@ void setupProblemFromXML( ProblemManager & problemManager, char const * const xm
...
+                           components );
+        } );
+      }
+    } );
+  } );
+}
+
+void testPhaseVolumeFractionNumericalDerivatives( CompositionalMultiphaseFVM & solver,
+                                                  DomainPartition & domain,
+                                                  bool const isThermal,
```
TotoGaz, line 304: So the `isThermal` is currently never `true`?

TotoGaz (approved):

src/coreComponents/unitTests/fluidFlowTests/testCompFlowUtils.hpp
francoishamon: Yes, I confirm that the `isThermal` is currently never `true` (until PR #1225 is merged). I am trying to take small steps towards thermal to avoid doing everything in the same PR. So currently,
- There is already a flag `isThermal` in the solver, currently always `false`.
- There is already some extrinsic data available in the solver for temperature, currently only modified at initialization.
- The present PR adds the unit tests functions to check derivatives wrt temperature, but they are currently unused.

PR #1225 will put everything together, allow the flag `isThermal` to be `true`, allow the temperature field to change with time, and will add a unit test using the functions added in the present PR.

PR #1813, opened 2022-03-10
Fixed the double free corruption problem in the unit tests after integration of pygeosx

sframba:
This patch solves the issue with the unit tests encountered after the merge of pygeosx (PR Feature/besset pygeosx #1704). Main changes:
- The optional Python code in dataRepository, fileIO and physicsSolvers (only activated when ENABLE_PYGEOSX=TRUE) is compiled as part of the compilation targets, instead of an additional target;
- The dependency on the Python3 and pylvarray is handled at the level of these targets, and no explicit dependency is needed at the level of the unitTests themselves;
- A circular dependency introduced by pygeosx has been removed: dataRepository, fileIO and physicsSolver no longer depend on mainInterface.


src/coreComponents/linearAlgebra/unitTests/CMakeLists.txt
```diff
@@ -18,7 +18,7 @@ if ( ENABLE_CUDA )
 endif()
 
 if ( ENABLE_PYGEOSX )
-  set( dependencyList ${dependencyList} pygroup pyfileIO pysolver )
+	set( dependencyList ${dependencyList} pygroup pyfileIO pysolver Python3::Python Python3::NumPy pylvarray)
```
TotoGaz, line 21: I'm sorry to see that the situation is so tricky for this merge.

So I have a question, that - who knows - may help you.
I always considered that `pygeosx` should be a wrapper around `geosx`.
_Why would a C++ linear algebra unit test depend on `python`?_
I'm not speaking about _technical considerations_, but from a _macroscopic point of view_.

(Same question applies for all tests here, but I guess it's more striking for linear algebra.)

sframba:
Thank you @TotoGaz for the hints. No, ENABLE_PYGEOSX is always false on Travis. I activated it by force in my branch to do some testing, but the test platforms don't have Python3 installed --> no dice.
I agree completely on your high-level view of the subject. 
Today the dependency unitTest->python goes as follows: unitTest->Group->PyGroup->python. I agree that it is somewhat unnatural. With this PR I am trying to cleanup a bit the compilation chain, but to change this dependency we would need to follow some of the advice that you gave elsewhere about having a static structure instead of a virtual function in Group.cpp (with its associated conditional compilation)...

src/coreComponents/fileIO/python/PyHistoryCollection.cpp
```diff
@@ -19,7 +17,7 @@
 
 namespace geosx
 {
-
+using namespace dataRepository;
```
labesse40, line 20: Why do we need to add this namespace ?

src/coreComponents/fileIO/python/PyHistoryCollection.cpp
```diff
@@ -84,8 +82,7 @@ static PyObject * collect( PyHistoryCollection * self, PyObject * args )
     return nullptr;
   }
 
-  geosx::GeosxState * g_state = &getGlobalState();
-  geosx::DomainPartition & domain = g_state->getProblemManager().getDomainPartition();
+  geosx::DomainPartition & domain = self->group->getGroup< DomainPartition >( keys::domain );
```
labesse40, line 85: Nice and simpler !

src/coreComponents/fileIO/python/PyHistoryCollection.cpp
sframba: It was to access the constant geosx::dataRepository::keys::domain, but its is no longer required in the latest commit (I have removed this line)

francoishamon:
Progress on testing this PR:

Quartz:
- Clang 10 (Release and Debug): compiles, 170/170 unit tests passing, no double free corruption
- GCC 8.1 (Release): compiles, 170/170 unit tests passing, no double free corruption

Lassen on going

francoishamon:
Lassen:

- Clang (Release): compiles, 166/166 unit tests passing, no double free corruption
- GCC 8.3.1 (Release): compiles, 166/166 unit tests passing, no double free corruption

labesse40 (approved):

rrsettgast (approved):

francoishamon:
> @francoishamon This is first in line. Is this safe to merge?

When I checked last week, the PR was working well for unit tests and integrated tests, on Quartz and Lassen, for clang and gcc, so I would say that it is ok to merge

labesse40:
@TotoGaz Yes I confirm, this one is ready to be merged today.

PR #1817, opened 2022-03-11
Add Reactions for Geochemistry and a ReactiveMultiFluidModel

CusiniM:
In this PR we

- [x]  Create `ReactiveMultiFluid` model that:
    - [x] adds Equilibrium and KineticReactions and all reactions related quantities
    - [x] serves as base class to all reactive models
    - [x] owns `EquilibriumReactions` and  `KineticReactions` and interfaces with them.
- [x]  Create `ReactiveBrineFluid` model that:
   - [x] adds a single phase (brine) with its own constitutive laws for viscosity, density, etc.
   - [x]  handles conversion from standard compositional variables to what's needed for geochem (volumetric concentrations)
- [x] Add WaterDensity model to have a simple density model for the brine that does not depend on the phase composition.
- [x] Add  two type of reactions:
    - [x] Equilibrium reactions
    - [x] kinetic reactions 
- [x] Create a dedicated `ReactiveFluidDriver` to test Reactions.
- [x] Test implementation for a few specific examples:
    - [x] equilibrium reactions
    - [x] kinetic reactions 
- [x] Split `linearAlgebra` and `denseLinearAlgebra` into 2 separate components (needed to avoid having constitutive depending on `linearAlgebra` which would also create a circular dependency).

The ReactiveFluidDriver can be run from the build directory with

`./tests/testReactiveFluid -i ../src/coreComponents/unitTests/constitutiveTests/testReactiveFluid.xml `

Needs to go in after https://github.com/GEOSX/GEOSX/pull/2211/files but otherwise it's ready.


CusiniM:
@francoishamon, @klevzoff I think this is ready for a first round of comments while I am gone. Ignore the std::output which is there for @jiyer28 to validate the implementation against her matlab code.

src/coreComponents/constitutive/fluid/ReactiveBrineFluid.hpp
```diff
@@ -0,0 +1,358 @@
...
+
+namespace constitutive
+{
+
+template< typename PHASE >
+class ReactiveBrineFluid : public ReactiveMultiFluid
+{
+public:
+
+  using exec_policy = parallelDevicePolicy<>;
```
francoishamon, line 44: Maybe you could just have `exec_policy = serialPolicy;` in this PR, since it seems really close to be mergeable for the CPU case, and worry about the GPU case in a later PR. For instance with `serialPolicy`, you would not have to worry about the call to
```
BlasLapackLA::solveLinearSystem( matrix, rhs, solution );
```
and you could just merge the PR as is.

src/coreComponents/constitutive/fluid/ReactiveBrineFluid.hpp
```diff
@@ -0,0 +1,358 @@
...
+           temperature,
+           composition,
+           m_phaseFraction( k, q ),
+           m_phaseDensity( k, q ),
+           m_phaseMassDensity( k, q ),
+           m_phaseViscosity( k, q ),
+           m_phaseEnthalpy( k, q ),
+           m_phaseInternalEnergy( k, q ),
+           m_phaseCompFraction( k, q ),
+           m_totalDensity( k, q ) );
```
francoishamon, line 350: After the PR is merged, we can probably find ways to reduce the code duplication with `CO2BrineFluid` because it has a lot in common with `ReactiveMultiFluid` (the `compute` functions, etc). In the current state of the PR, the `ReactiveMultiFluid` could almost inherit from `CO2BrineFluid`. 

At the same time, if we want a generic reactive capability that would work with true compositional EOS, maybe what I am suggesting above is not the right path (unless `CO2BrineFluid` is refactored to become more general).

src/coreComponents/constitutive/fluid/chemicalReactions/EquilibriumReactions.cpp
```diff
@@ -0,0 +1,339 @@
...
+{
+  // Here we should either read the database or the input values.
+
+  // Hardcoding values for now
+
+  
+
+  // Stochiometric Matrix
+  // First index: 0 = OH-, 1 = CO2, 2 = CO3-2, 3 = H2CO3, 4 = CaHCO3+, 5 = CaCO3, 6 = CaSO4, 7 = CaCl+, 8 = CaCl2, 9 = MgSO4, 10 = NaSO4-
+  // Second index: 0 = H+, 1 = HCO3-, 2 = Ca+2, 3 = SO4-2, 4 = Cl-, 5 = Mg+2, 6 = Na+1
```
francoishamon, line 46: I don't know much about reactive transport, but maybe a third possibility (in addition to reading the database or the input values) would be to keep everything hardcoded in the equilibrium/kinetic reaction classes. 

You would write multiple `EquilibriumReactions` and `KineticReactions` classes with everything hardcoded inside (`numPrimarySpecies`, `numSecSpecies`, stochiometric coefficients), then you would select the right class/template from `ReactiveMultiFluid::createChemicalReactions`.

That could be a way to have a general implementation but still have `matrix`, `solution`, and `rhs` with the right size below. But I don't know enough about reactive transport to know if this is a viable solution.

src/coreComponents/constitutive/fluid/ReactiveMultiFluid.hpp
```diff
@@ -0,0 +1,224 @@
...
+                    arraySlice1d< real64, compflow::USD_COMP - 1 > const & primarySpeciesConcentration,
+                    arraySlice1d< real64, compflow::USD_COMP - 1 > const & secondarySpeciesConcentration,
+                    arraySlice1d< real64, compflow::USD_COMP - 1 > const & primarySpeciesTotalConcentration,
+                    arraySlice1d< real64, compflow::USD_COMP - 1 > const & kineticReactionRates ) const
+{
+  // I am assuming that the primary variable is the concentration of the primary species.
+  for( int i=0; i < m_numPrimarySpecies; i++ )
+  {
+    primarySpeciesTotalConcentration[i] = composition[i];
+  }
```
francoishamon, line 207: Just a small question: is the `composition` coming in the right unit here (when coming from the compositional solver)? 

I am just asking because I saw in the `ReactiveMultiFluid` driver:
```
feedComposition="{3.76e-1, 3.76e-1, 3.87e-2, 3.21e-2, 1.89, 1.65e-2, 1.09 }"
```
with compositions above 1.

src/coreComponents/constitutive/fluid/chemicalReactions/EquilibriumReactions.cpp
```diff
@@ -0,0 +1,339 @@
...
+  array1d<real64> rhs( m_numPrimarySpecies );
+  array1d<real64> solution( m_numPrimarySpecies );
+
+  setInitialGuess(primarySpeciesTotalConcentration, primarySpeciesConcentration);
+  std::cout<<"Initial guess " << primarySpeciesConcentration <<std::endl;
+  std::cout<<"Total concentration to solve for " << primarySpeciesTotalConcentration <<std::endl;
+
+
+  bool converged = false;
+  for( int iteration = 0; iteration < m_maxNumIterations; iteration++ )
```
francoishamon, line 119: There is also a Newton solver [here](https://github.com/GEOSX/GEOSX/blob/develop/src/coreComponents/constitutive/fluid/PVTFunctions/CO2EOSSolver.cpp). It seems a little bit difficult to use it in the new `EquilibriumReactions` class, but maybe later we can unify them, or at least group them in the same place.

francoishamon (approved):
Very nice, the PR seems close to ready and contains a lot of work. I just posted small suggestions/questions.

src/coreComponents/constitutive/fluid/ReactiveMultiFluid.hpp
CusiniM: yeah, I think these are `mol/L`.  unit conversion is kind of a pain which is why I have not taken care of it yet.

src/coreComponents/constitutive/fluid/chemicalReactions/EquilibriumReactions.cpp
CusiniM: ah, yeah, we should really write a generic nonllinear solver.

CusiniM:
@TotoGaz @rrsettgast @francoishamon has anyone seen  [this](https://app.travis-ci.com/github/GEOSX/GEOSX/jobs/590820506) issue with the unit tests on OSX before? I am not sure why those are failing.

PR #1820, opened 2022-03-12
Correct argument order inconsistency.

TotoGaz:
The parameters order of `EmbeddedSurfaceSubRegion::addNewEmbeddedSurface` was not consistent between header and definition. Quite dangerous!

I've left the header unchanged but changed the definition (and the calls) such that regions come before sub-regions.


CusiniM (approved):
Thanks for catching this.

PR #1821, opened 2022-03-14
Add another verification example of a radial frac developed in viscosity dominated regime

jhuang2601:
This PR 

- [x] documents the analytical verification of a radial fracture developed in viscosity-storage-dominated regime;

image

- [x] adds `pennyShapedViscosityDominated_smoke` to the integratedTests and checks the running on Pecan. 
       [GEOSX/integratedTests#202](https://github.com/GEOSX/integratedTests/pull/202)     

This case is running with the updated TPLs (https://github.com/GEOSX/thirdPartyLibs/pull/177)


herve-gross (approved):
Great example. Thank you!

src/docs/sphinx/advancedExamples/validationStudies/pennyFracViscosityDominated/pennyFracViscosityDominatedFigure.py
```diff
@@ -0,0 +1,143 @@
+import numpy as np
```
rrsettgast, line 1: looks like you didn't use the existing [HydrofractureSolutions.py](https://github.com/GEOSX/GEOSX/blob/develop/scripts/postProcessing/hydraulicFractureVerification/HydrofractureSolutions.py)

Was there a problem using this?

jhuang2601:
> @jhuang2601 Are these the parameters from the 2016 paper?

@rrsettgast Yes, the input parameters should be identical to the publication.

src/docs/sphinx/advancedExamples/validationStudies/pennyFracViscosityDominated/pennyFracViscosityDominatedFigure.py
jhuang2601: While testing this case, I coded up the python script separately, but it should be the same as the existing one.

rrsettgast: The point is that we have a single place where all the analytical solutions for KGD and Penny problems are contained rather than having it duplicated in multiple scripts. Could you use the existing capability?

rrsettgast:
> > @jhuang2601 Are these the parameters from the 2016 paper?
> 
> @rrsettgast Yes, the input parameters should be identical to the ones in publication.

Do you want to refer to the publication?

src/docs/sphinx/advancedExamples/validationStudies/pennyFracViscosityDominated/pennyFracViscosityDominatedFigure.py
jhuang2601: Use existing python script and update the examples for both KGD and Penny fractures

PR #1822, opened 2022-03-15
Cleaned up isPointInsidePolyhedron

francoishamon:
This PR cleans up the computational geometry functions in charge of locating a point inside an element (called `isPointInsidePolyhedron`). Before this PR, we had two versions of the function, but only one working robustly for prisms and pyramids. 

This PR removes entirely the not-robust version and makes the necessary changes to use the robust version in the code that matches a well perforation with a reservoir element. 

No need for a rebaseline

### DIFF
--- a/src/coreComponents/mesh/WellElementSubRegion.cpp
+++ b/src/coreComponents/mesh/WellElementSubRegion.cpp
@@ -115,29 +115,27 @@ void collectLocalAndBoundaryNodes( InternalWellGenerator const & wellGeometry,
 }
 
 /**
- * @brief Check if "location" is contained in reservoir element ei
- * @param[in] subRegion the subRegion of reservoir element ei
- * @param[in] ei the index of the reservoir element
- * @return true if "location" is contained in reservoir element ei, false otherwise
+ * @brief Check if "location" is contained in reservoir element
+ * @param[in] referencePosition the coordinates of the nodes
+ * @param[in] elemsToFaces the map from element to faces
+ * @param[in] facesToNodes the map from face to nodes
+ * @param[in] elemCenter the coordinates of the element center
+ * @param[in] location the target coordinates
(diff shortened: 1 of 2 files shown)


src/coreComponents/mesh/WellElementSubRegion.cpp
```diff
@@ -115,29 +115,27 @@ void collectLocalAndBoundaryNodes( InternalWellGenerator const & wellGeometry,
...
+ * @param[in] facesToNodes the map from face to nodes
+ * @param[in] elemCenter the coordinates of the element center
+ * @param[in] location the target coordinates
+ * @return true if "location" is contained in the reservoir element, false otherwise
  */
-bool isPointInsideElement( NodeManager const & nodeManager,
-                           R1Tensor const & location,
-                           CellElementSubRegion const & subRegion,
-                           localIndex ei )
+bool isPointInsideElement( arrayView2d< real64 const, nodes::REFERENCE_POSITION_USD > const referencePosition,
```
TotoGaz, line 126: Hasn't this function become an alias like
```c++
bool isPointInsideElement( arrayView2d< real64 const, nodes::REFERENCE_POSITION_USD > const referencePosition,
                           arraySlice1d< localIndex const > const elemsToFaces,
                           ArrayOfArraysView< localIndex const > const facesToNodes,
                           real64 const (&elemCenter)[3],
                           real64 const (&location)[3] )
{
  return computationalGeometry::isPointInsidePolyhedron( referencePosition,
                                                         elemsToFaces,
                                                         facesToNodes,
                                                         elemCenter,
                                                         location );
}
```
Consider simply removing it?

francoishamon: Yes, good idea, thanks!

TotoGaz (approved):

PR #1827, opened 2022-03-17
Moved fluid variables at time n to multifluid constitutive model 

francoishamon:
This PR addresses [this comment](https://github.com/GEOSX/GEOSX/pull/1225#discussion_r821257097) from @CusiniM. I prefer to address the comment in a separate PR mergeable before or after the thermal PR.

This PR moves the phase density, phase component fraction, phase internal energy, phase enthalpy evaluated at the previous time level (time `n`) to multifluid constitutive model, and add a method `saveConvergedState` to the `MultiFluidBase` class.

The PR does not introduce numerical diffs, but requires a rebaseline. Rebaseline done in https://github.com/GEOSX/integratedTests/pull/210


src/coreComponents/constitutive/fluid/MultiFluidBase.cpp
```diff
@@ -193,8 +203,52 @@ void MultiFluidBase::initializeState( arrayView2d< real64 const, compflow::USD_P
...
+  // initialize the "old" variables
+  saveConvergedState();
+}
+
+void MultiFluidBase::saveConvergedState() const
+{
+  localIndex const numElem = m_initialTotalMassDensity.size( 0 );
+  localIndex const numGauss = m_initialTotalMassDensity.size( 1 );
+  integer const numPhase = m_phaseMassDensity.value.size( 2 );
+  integer const numComp = m_phaseCompFraction.value.size( 3 );
```
TotoGaz, line 216: Some day we can get rid of these hard coded indices. 😉

src/coreComponents/constitutive/fluid/MultiFluidBase.cpp
```diff
@@ -193,8 +203,52 @@ void MultiFluidBase::initializeState( arrayView2d< real64 const, compflow::USD_P
...
+      totalDensityOld[k][q] = totalDensity.value[k][q];
+      for( integer ip = 0; ip < numPhase; ++ip )
+      {
+        phaseDensityOld[k][q][ip] = phaseDensity.value[k][q][ip];
+        phaseEnthalpyOld[k][q][ip] = phaseEnthalpy.value[k][q][ip];
+        phaseInternalEnergyOld[k][q][ip] = phaseInternalEnergy.value[k][q][ip];
+        for( integer ic = 0; ic < numComp; ++ic )
+        {
+          phaseCompFractionOld[k][q][ip][ic] = phaseCompFraction.value[k][q][ip][ic];
+        }
```
TotoGaz, line 243: So we do not store the derivatives, right?

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBaseKernels.hpp
```diff
@@ -741,13 +741,13 @@ class ElementBasedAssemblyKernel
     arraySlice1d< real64 const, compflow::USD_PHASE - 1 > dPhaseVolFrac_dPres = m_dPhaseVolFrac_dPres[ei];
     arraySlice2d< real64 const, compflow::USD_PHASE_DC - 1 > dPhaseVolFrac_dCompDens = m_dPhaseVolFrac_dCompDens[ei];
 
-    arraySlice1d< real64 const, compflow::USD_PHASE - 1 > phaseDensOld = m_phaseDensOld[ei];
+    arraySlice1d< real64 const, multifluid::USD_PHASE - 2 > phaseDensOld = m_phaseDensOld[ei][0];
```
TotoGaz, line 744: Why do you need this extra (dummy?) dimension? (The `[0]` at the end).
That bothers me here and there in the code.

TotoGaz (approved):

src/coreComponents/constitutive/fluid/MultiFluidBase.cpp
CusiniM: well, for the quantities at the previous time-step we only need the value non its derivatives.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBaseKernels.hpp
CusiniM: It's because the second dimension contains the quadrature point index. The thing is that our flow kernels assume a single integration point since we use piece-wise constant functions for all our FV relevant quantities. We could loop over quadrature points and avoid hard-coding the value 0 but it would add an extra loop that in these kernels would not make much sense. The problem is that the same constitutive model can be used with a discretization scheme that does need that second dimension.

PR #1831, opened 2022-03-18
Improve packing naming conventions

TotoGaz:
- `DO_PACKING` is now the uniformly used boolean template parameter name of `pack` and `unpack` functions.
- `(un)pack*Private` are renamed `(un)pack*Impl` for this pattern is more meaningful.

Follow-up from comments from https://github.com/GEOSX/GEOSX/pull/1792


francoishamon (approved):
Nice refactoring, thanks

PR #1832, opened 2022-03-19
Improve config-build.py and put cache option at the end of the command line.

### LINKED ISSUE #1795: `config-build.py` appends `-D` options after `-C`
The `-D` options may not be considered by the `-C` cache script, because the order is important.
Changing the order should be OK, but we need to be sure of what we are doing because this script is critical to a lot of users.

TotoGaz:
Could you please check that it's working on your box? @bmhan12 @XL64 
If you can test TPLs as well, that would be great.
Do not hesitate to put your own patch.

~~I'll update TPL accordingly after validation, do not merge before.~~ Third party libs updated here: https://github.com/GEOSX/thirdPartyLibs/pull/182

Resolves #1795


XL64 (approved):
Seems OK to me, I also used this script on TPLs (do not forget to change the source path and remove , "src" in TPLs.

bmhan12 (approved):
LGTM! I appreciate the clarity of the logged comments.

TotoGaz:
Thank you guys for you first check! I appreciate!

TotoGaz:
Hello @karimifard is it possible that you check this branch to see if option `-DGEOSX_TPL_DIR` can now be considered from the `config-build.py` command line? You should not need to write it in your `host-config` `cmake` file to compile `geosx`.

karimifard:
Just tested the configuration with the new config-build.py. Now it works with option "-D GEOSX_TPL_DIR=/full/path/to/thirdPartyLibs" without the need to define "GEOSX_TPL_DIR" in "your-platform.cmake".

XL64 (approved):

PR #1835, opened 2022-03-21
VEM support in small strain solid mechanics solvers

andrea-borio:
The purpose of this PR is to add support of Virtual Elements to `SolidMechanicsLagrangianFEM` and `SolidMechanicsSmallStrainQuasiStaticKernel` in order to solve linear elasticity with VEM.

### Changes to solid mechanics solver

- `solidMechanicsLagrangianFEMKernels::QuasiStatic` and `SolidMechanicsLagrangianFEM` modified introducing enhanced FEM syntax compatible with VEM.
- classes in `finiteElement/elementFormulations` modified to be able to add stabilization terms in case of vectorial problems.
- the stabilization matrix is scaled locally by a factor that is computed by `solidMechanicsLagrangianFEMKernels::QuasiStatic::computeStabilizationScaling`. The method is currently implemented for a linear elastic model.
- to avoid long compilation time which caused failure of some Travis builds, the `SolidMechanicsLagrangianFEM` solver is compiled only with VEM types that do not overlap with existing FEM types.

### Changes to `FiniteElementDiscretization::factory`

The factory always allocates VEM on thos shapes where we do not have FEM. It allocates VEM on standard shapes only if the flag `useVirtualElements` is set to 1 in the input xml file.

### Changes in `FiniteElementDispatch`

The behavior of dispatch methods was changed as follows: if the `GEOSX_DISPATCH_VEM` flag is defined, the class tries to dispatch also VEM types that overlap with standard FEM, otherwise it considers only the ones defined on non-standard shapes.


castelletto1:
Hi @andrea-borio, I have a question that is not directly related to this PR. However, if the concern is legitimate, it may be resolved here. The convenience typedefs introduced in `ConformingVirtualElementOrder1.hpp` (see below) are not quite clear to me. Specifically I don't understand the second template parameter.
https://github.com/GEOSX/GEOSX/blob/b8d4a23007add5677e65aaf9965e6ca9078b87f7/src/coreComponents/finiteElement/elementFormulations/ConformingVirtualElementOrder1.hpp#L567-L574
My understanding is that such parameter defines the maximum number of nodes per face for each element type. Hence, I would expect it to be `3` for the tetrahedra and `4` for the all the other elements, namely:

```c++
 /// Convenience typedef for VEM on tetrahedra. 
 using H1_Tetrahedron_VEM_Gauss1 = ConformingVirtualElementOrder1< 4, 3 >; 
 /// Convenience typedef for VEM on hexahedra. 
 using H1_Hexahedron_VEM_Gauss1 = ConformingVirtualElementOrder1< 8, 4 >; 
 /// Convenience typedef for VEM on pyramids. 
 using H1_Pyramid_VEM_Gauss1 = ConformingVirtualElementOrder1< 5, 4 >; 
 /// Convenience typedef for VEM on wedges. 
 using H1_Wedge_VEM_Gauss1 = ConformingVirtualElementOrder1< 6, 4 >; 
 ```

File: src/coreComponents/constitutive/solid/SolidBase.hpp [hunk 1]
```diff
@@ -132,6 +132,36 @@ class SolidBaseUpdates
...
+
+  /**
+   * @brief Get a parameter representative of the stiffness.
+   * @details This can be used when stabilizations need to be scaled locally by taking into account
+   * the order of magnitude of the stiffness matrix.
+   * @param[in] k Element index.
+   * @return A parameter representative of the stiffness matrix dstress/dstrain
+   */
+  GEOSX_HOST_DEVICE
+  virtual real64 getStiffnessParameter( localIndex const k ) const
```
joshua-white, line 157: I would remove this function, as the `getShearModulus()` function does nearly what you want.  Having a specific stabilization constant here breaks the API separation between constitutive models and physics solvers.  I would just compute `stiffnessParameter = 2*solid.getShearModulus()` in the assembly kernels (or whatever you need).  If you need other material constants for anisotropic models, though, we can add them here later (or perhaps just use the stiffness matrix directly to compute the appropriate scalings).

francoishamon (approved):
The PR looks good to me. It probably only needs to be brought up-to-date with `develop`. I can help with the integrated tests when it is a good time to merge

PR #1837, opened 2022-03-21
Fixed an input validation bug in well controls

francoishamon:
This PR fixes an input validation bug in the well controls related the use of a time-dependent BHP table. The user was forced to specify `targetBHP` in the xml, even when using a time-dependent BHP table, which was confusing.

Now, I changed the default value of `targetBHP` to make sure that the input validation step passes if the user does not specify a target BHP. 

I changed an xml to test this case, so the PR will require a rebaseline.
Rebaseline done in https://github.com/GEOSX/integratedTests/pull/209


src/coreComponents/physicsSolvers/fluidFlow/wells/WellControls.cpp
```diff
@@ -65,7 +65,7 @@ WellControls::WellControls( string const & name, Group * const parent )
   registerWrapper( viewKeyStruct::targetPhaseRateString(), &m_targetPhaseRate ).
     setDefaultValue( 0.0 ).
     setInputFlag( InputFlags::OPTIONAL ).
-    setDescription( "Target phase volumetric rate" );
+    setDescription( "Target phase volumetric rate [sm^3/s]" );
```
TotoGaz, line 68: Is this intended? `[sm^3/s]` -> `[m^3]`?

src/coreComponents/physicsSolvers/fluidFlow/wells/WellControls.cpp
```diff
@@ -53,7 +53,7 @@ WellControls::WellControls( string const & name, Group * const parent )
     setDescription( "Well control. Valid options:\n* " + EnumStrings< Control >::concat( "\n* " ) );
 
   registerWrapper( viewKeyStruct::targetBHPString(), &m_targetBHP ).
-    setDefaultValue( -1 ).
+    setDefaultValue( 0.0 ).
```
TotoGaz, line 56: Will this change the value of the `m_targetBHP <= 0.0` check? Same value for `-1` or `0.0`?

TotoGaz (commented):
> I changed an xml to test this case, so the PR will require a rebaseline

😢

src/coreComponents/physicsSolvers/fluidFlow/wells/WellControls.cpp
francoishamon: Yes, this can be confusing, [sm^3/s] means "surface m^3 per second", it is now written explicitly in the code

src/coreComponents/physicsSolvers/fluidFlow/wells/WellControls.cpp
francoishamon: This PR should not change the value of the `m_targetBHP <= 0.0`, and we don't want this to change for now.

TotoGaz (approved):

PR #1840, opened 2022-03-22
Enable Coulomb Friction for EFEM Solvers 

CusiniM:
https://github.com/GEOSX/integratedTests/pull/214


CusiniM:
@jhuang2601 you should now be able to run that test case of a slanted fracture under compression with friction. The smoke test I have created should already more or less be the test case we want.

@joshua-white, @castelletto1, @rrsettgast it would be nice if we could think of a way to test the Coulomb contact machinery well. Maybe loading and unloading a fracture and then keep track of the elastic and plastic slip? I d appreciate some help coming up with a meaningful case.

joshua-white:
Randy made a good suggestion for having something like our `TriaxialDriver` but for contact constitutive relationships (`ShearBoxDriver`).  This can unit test the Coulomb and other contact constitutive relationships directly.  We can then have a smoke test of a meshed shear box test.  The results should be identical to the material point driver, but now running through a full solver.

joshua-white:
To be more concrete, here's an image of a typical shear box apparatus (for a fracture--there are also soil shear box tests, which are similar but have a diffuse shear zone):

iu

rrsettgast:
I certainly think a constitutive driver for contact is in order. On a broader note, as we have more complex constitutive models, we should enable dumping of a material state to file in order to read it into the drivers. This will allow for easier debugging of code and model.

joshua-white:
Adding a ShearBoxDriver should be quick, as it will look nearly identical to the TriaxialDriver---just a slightly different input meaning and call to a different constitutive model type.  With a little templating or clever inheritance we can probably re-use 90% of the code.

CusiniM:
The only different is that, contrary to the `TriaxialDriver`, the enforcement of contact varies depending on the solver (formulation and discretization) and it would be nice to have a way to test all our discretizations. Unless the goal of the `ShearBox` is solely to test the constitutive model (I have not really looked at the `TriaxialDriver` tbh).

joshua-white:
I would say that goal of the ShearBoxDriver is precisely to test *only* the constitutive response, without going through the mesh / discretization / solver.  We then need integrated tests for the various discretizations, but one of them could look like the shear box test above.  The response should be uniform everywhere, but at least the meshed model can confirm that.

One challenge though is that the different discretizations call the constitutive models in slightly different ways (e.g. is the normal traction a known quantity or an output?).

CusiniM:
The `LagrangianSolver` has kind of the `CoulombContact` model embedded in itself and basically only asks for the maximum tau but the EFEM one is written in a general way and basically should be able to accommodate any contact model. I think we can start by testing that interface through a `ShearBoxDriver`. 

I think I will wrap up this PR and then work on the  `ShearBoxDriver` and on two integratedTest that mimic its behavior (one per discretization/solver).

src/coreComponents/physicsSolvers/contact/LagrangianContactSolver.cpp
```diff
@@ -1068,7 +1065,7 @@ void LagrangianContactSolver::
 
             switch( fractureState[kfe] )
             {
-              case FractureState::Stick:
+              case extrinsicMeshData::contact::FractureState::Stick:
```
TotoGaz, line 1068: If it works, maybe something like `using namespace extrinsicMeshData::contact;`?

src/coreComponents/physicsSolvers/contact/LagrangianContactSolver.cpp
```diff
@@ -1706,6 +1703,8 @@ bool LagrangianContactSolver::resetConfigurationToDefault( DomainPartition & dom
 {
   GEOSX_MARK_FUNCTION;
 
+  using namespace extrinsicMeshData::contact;
```
TotoGaz, line 1706: Arf, like there 🤣

src/coreComponents/physicsSolvers/contact/SolidMechanicsEmbeddedFractures.cpp
```diff
@@ -725,18 +758,72 @@ void SolidMechanicsEmbeddedFractures::updateState( DomainPartition & domain )
...
+
+  using namespace extrinsicMeshData::contact;
 
+  forMeshTargets( domain.getMeshBodies(), [&] ( string const &,
+                                                MeshLevel & mesh,
+                                                arrayView1d< string const > const & regionNames )
+  {
+    ElementRegionManager & elemManager = mesh.getElemManager();
+
+    elemManager.forElementSubRegions< EmbeddedSurfaceSubRegion >( regionNames, [&]( localIndex const,
```
TotoGaz, line 780: IIUC this loop checks old and new configuration state to see if it has changed or not?
If there is not change, then you conclude it's a convergence?
Maybe add a little comments about this?

TotoGaz (approved):

src/coreComponents/constitutive/contact/FrictionlessContact.hpp
```diff
@@ -145,18 +149,38 @@ void FrictionlessContactUpdates::computeTraction( localIndex const k,
...
 
+GEOSX_HOST_DEVICE
+void FrictionlessContactUpdates::updateFractureState( localIndex const k,
+                                                      arraySlice1d< real64 const > const & dispJump,
+                                                      arraySlice1d< real64 const > const & tractionVector,
+                                                      integer & fractureState ) const
+{
+  GEOSX_UNUSED_VAR( k, tractionVector );
+
+  if( dispJump[0] >  -LvArray::NumericLimits< real64 >::epsilon )
```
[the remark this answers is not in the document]
CusiniM: I think that, for non linear problems, since we solve to a tolerance it's a bit too stricht to ask for the jump to be strictly greater than 0.0. For example, especially for nearly closed fractures (with small apertures), we don't really want the state to change if we are just a little under 0.0. I know I tried to do so in the prototype code I had and it can lead to convergence issues.

src/coreComponents/constitutive/contact/CoulombContact.hpp
```diff
@@ -219,11 +232,78 @@ void CoulombContactUpdates::computeTraction( localIndex const k,
...
+      }
+    }
+  }
+}
+
+GEOSX_HOST_DEVICE
+void CoulombContactUpdates::updateFractureState( localIndex const k,
+                                                 arraySlice1d< real64 const > const & dispJump,
+                                                 arraySlice1d< real64 const > const & tractionVector,
+                                                 integer & fractureState ) const
```
CusiniM, line 295: @TotoGaz if done in scope this is fine even in an hpp file, right?

src/coreComponents/physicsSolvers/contact/LagrangianContactSolver.cpp
CusiniM: yeah, I don't understand why sometimes it doesn't work. I think that, in this function it was not working...

TotoGaz: fatalitas!

src/coreComponents/physicsSolvers/contact/SolidMechanicsEmbeddedFractures.cpp
CusiniM: yes, precisely.

src/coreComponents/constitutive/contact/CoulombContact.hpp
TotoGaz: I think it's OK if the `using namespace` dies when out of scope, yer.
I remember that @klevzoff mentioned something about it? Is there some specific care?

src/coreComponents/constitutive/contact/FrictionlessContact.hpp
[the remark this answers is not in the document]
TotoGaz: So this is not really an IT constraint but more something about your algorithm.
Maybe do you want to define this quantity as a `dispJumpApertureThreshold`, that can be `const` or specified by the user (or the dev to easily check numerical parameters of convergence)?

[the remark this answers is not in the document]
CusiniM: yes, indeed. Ah, yeah, I can do that. Maybe the numeriLitmits can be its default value.

src/coreComponents/physicsSolvers/SolverBase.cpp
```diff
@@ -618,7 +618,9 @@ real64 SolverBase::nonlinearImplicitStep( real64 const & time_n,
         }
         else
         {
-          GEOSX_LOG_LEVEL_RANK_0( 1, "  --------Configuration did not converge. Testing new configuration." );
+          GEOSX_LOG_LEVEL_RANK_0( 1, "   " );
+          GEOSX_LOG_LEVEL_RANK_0( 1, "---------- Configuration did not converge. Testing new configuration. ----------" );
+          GEOSX_LOG_LEVEL_RANK_0( 1, "   " );
```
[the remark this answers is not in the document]
CusiniM: I think the code I have only outputs the log if `logLevel() >=1` whereas the other one always prints it.

src/coreComponents/constitutive/contact/CoulombContact.hpp
klevzoff: I'm relatively certain you'll need `inline` on the function body (but not force inline), otherwise the linker will complain about ODR violations

src/coreComponents/physicsSolvers/SolverBase.cpp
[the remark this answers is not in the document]
TotoGaz: Oh sorry, I think I misread the code. Forget about it!

src/coreComponents/constitutive/contact/ContactBase.cpp
```diff
@@ -54,6 +51,12 @@ ContactBase::ContactBase( string const & name,
                     "to smooth out highly nonlinear behavior associated with 1/0 in addition to avoiding the "
                     "1/0 error." );
 
+  registerWrapper( viewKeyStruct::displacementJumpThresholdString(), &m_displacementJumpThreshold ).
+    setApplyDefaultValue( std::numeric_limits< real64 >::epsilon() ).
```
andrea-franceschini, line 55: I agree with the need to use tolerances for the fracture state computation according to the jump.
In the Lagrangian Solver I encountered the same problem and it turned out that, al least in that specific solver, the tolerances have to be problem dependent, i.e., considering the element size, the stiffness of the surrounding FE, and so on ...
I know this is completely beyond the scope of this PR, but perhaps to initialize this threshold to something larger, e.g. 10/100 * _tol_ (with _tol_ the default exit value for the linear solver), would help. Indeed, the linear system is usually solved iteratively up to _tol_ (_tol_ = eps only for a direct solver). Moreover, tol is reached by the residual, while we compare the solution (or derived quantities), thus I would suggest the multiplicative factor of at least 10.
I think setting a proper default initialization will help the end user a lot.

andrea-franceschini (approved):

src/coreComponents/constitutive/contact/CoulombContact.hpp
```diff
@@ -232,47 +315,9 @@ void CoulombContactUpdates::computeTraction( localIndex const k,
...
-      dTractionVector_dJump[2][2] = limitTau * pow( slip[0], 2 )  / pow( LvArray::tensorOps::l2NormSquared< 2 >( slip ), 1.5 );
-
-      // Compute elastic component of the slip for this case
-      real64 const plasticSlip[2] = { tractionVector[1] / m_shearStiffness,
-                                      tractionVector[2] / m_shearStiffness };
-
-      LvArray::tensorOps::copy< 2 >( m_elasticSlip[k], slip );
-      LvArray::tensorOps::subtract< 2 >( m_elasticSlip[k], plasticSlip );
-    }
+    fractureState = yield < 0 ? FractureState::Stick : FractureState::Slip;
```
jhuang2601, line 319: Four different states are defined. When updating the FractureState, it seems that the `new_slip` state is missing?

src/coreComponents/physicsSolvers/contact/ContactExtrinsicData.hpp
```diff
@@ -32,6 +32,22 @@ namespace extrinsicMeshData
...
+ * A struct for the fracture states
+ */
+struct FractureState
+{
+  enum State : integer
+  {
+    Stick = 0,   ///< element is closed: no jump across the discontinuity.
+    Slip = 1,   ///< element is sliding: no normal jump across the discontinuity, but sliding is allowed.
+    NewSlip = 2,   ///< element just starts sliding: no normal jump across the discontinuity, but sliding is allowed.
+    Open = 3   ///< element is open: no constraints are imposed.
```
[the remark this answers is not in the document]
jhuang2601: Vote for Andrea's suggestion. But, considering plastic/irreversible sliding, how to ensure monotonic change of the FractureState?

src/coreComponents/constitutive/contact/CoulombContact.hpp
CusiniM: the efem solver does not distinguish between `new_slip` and `slip`.

src/coreComponents/physicsSolvers/contact/ContactExtrinsicData.hpp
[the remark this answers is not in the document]
CusiniM: This should be a very easy change.

jhuang2601 (approved):

src/coreComponents/physicsSolvers/contact/LagrangianContactSolver.cpp
```diff
@@ -1133,7 +1125,7 @@ void LagrangianContactSolver::
 
 //                GEOSX_LOG_LEVEL_BY_RANK( 3, "element: " << kfe << " sliding: " << sliding[0] << " " << sliding[1] );
 
-                  if( !( ( m_nonlinearSolverParameters.m_numNewtonIterations == 0 ) && ( fractureState[kfe] == FractureState::NewSlip ) )
+                  if( !( ( m_nonlinearSolverParameters.m_numNewtonIterations == 0 ) && ( fractureState[kfe] == extrinsicMeshData::contact::FractureState::NewSlip ) )
```
CusiniM, line 1128: @rrsettgast, @klevzoff this is one of the places where I would like to use `using namespace extrinsicMeshData::contact` at the beginning of the function but, with nvcc I get compilation errors (basically saying that it does not know what `FractureState` is).

PR #1856, opened 2022-03-28
Update bug-report.md

herve-gross:
Remove the [Bug] in the title of the report.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -1,7 +1,7 @@
 ---
 name: Bug report
 about: Create a report to help us improve
-title: '[Bug] Your Title'
+title: 'Your Title'
 labels: new
 assignees: ''
 


joshua-white:
Is there a reason we are just removing [Bug] and not [Feature], etc, in other issues?  We should perhaps discuss preferred labeling during our dev mtg.

TotoGaz:
> Is there a reason we are just removing [Bug] and not [Feature], etc, in other issues? We should perhaps discuss preferred labeling during our dev mtg.

@herve-gross 's patch was surely meant to initiate (ignite?) discussion about it (quick web ui patch).
The statement being that the nature of a significant amount of "_issues_" cannot be determined at first glance.
Manipulating tags is convenient, but nobody never changes the title (nor filters from title).

I'd vote for complete title tag removal; in favor of a precise github tag management (even though github's tag system ia far from being great).

joshua-white:
I am on board with complete tag remove. @herve-gross, do you want to edit the other markdown templates as well and we can do it in this PR?

TotoGaz:
@joshua-white @herve-gross @castelletto1 I've finished the PR by removing all the tags in the title of the github issues. Using `labels` should be the preferred way to filter issues.

Plus sometimes bugs are features, which would require changing the title, which is not often done.

castelletto1 (approved):
Looks good to me, thank you!

joshua-white (approved):

PR #1857, opened 2022-03-28
Increased readibility for hypre interface

castelletto1:
This PR improves the readability of the preconditioner specialization for the hypre LAI. Maps for hypre's type identifier codes have been moved from `HyprePreconditioner.cpp` to `HypreUtilis.hpp` to allow for their use in the MGR strategies. Enumerations have also been added, in particular the MGR preconditioner.


src/coreComponents/linearAlgebra/interfaces/hypre/HypreMGR.cpp
```diff
@@ -151,16 +151,17 @@ void hypre::mgr::createMGR( LinearSolverParameters const & params,
     GEOSX_LAI_CHECK_ERROR( HYPRE_BoomerAMGSetTol( mgrData.mechSolver.ptr, 0.0 ) );
     GEOSX_LAI_CHECK_ERROR( HYPRE_BoomerAMGSetMaxIter( mgrData.mechSolver.ptr, 1 ) );
     GEOSX_LAI_CHECK_ERROR( HYPRE_BoomerAMGSetMaxRowSum( mgrData.mechSolver.ptr, 1.0 ) );
-    GEOSX_LAI_CHECK_ERROR( HYPRE_BoomerAMGSetStrongThreshold( mgrData.mechSolver.ptr, 0.6 ) );
+    GEOSX_LAI_CHECK_ERROR( HYPRE_BoomerAMGSetStrongThreshold( mgrData.mechSolver.ptr, 0.8 ) );
     GEOSX_LAI_CHECK_ERROR( HYPRE_BoomerAMGSetPrintLevel( mgrData.mechSolver.ptr, 0 ) );
 #ifdef GEOSX_USE_HYPRE_CUDA
-    GEOSX_LAI_CHECK_ERROR( HYPRE_BoomerAMGSetCoarsenType( mgrData.mechSolver.ptr, 8 ) ); // PMIS
-    GEOSX_LAI_CHECK_ERROR( HYPRE_BoomerAMGSetRelaxType( mgrData.mechSolver.ptr, 18 ) ); // l1-Jacobi
+    GEOSX_LAI_CHECK_ERROR( HYPRE_BoomerAMGSetCoarsenType( mgrData.mechSolver.ptr, hypre::getHypreAMGCoarseType( "PMIS" ) ) );
```
klevzoff, line 157: Wasn't this also renamed?
```suggestion
    GEOSX_LAI_CHECK_ERROR( HYPRE_BoomerAMGSetCoarsenType( mgrData.mechSolver.ptr, hypre::getAMGCoarseType( "PMIS" ) ) );
```

src/coreComponents/linearAlgebra/interfaces/hypre/mgrStrategies/CompositionalMultiphaseHybridFVM.hpp
```diff
@@ -70,9 +70,9 @@ class CompositionalMultiphaseHybridFVM : public MGRStrategyBase< 3 >
 
     setupLabels();
 
-    m_levelFRelaxMethod[0] = 0; // Jacobi
-    m_levelFRelaxMethod[1] = 0; // Jacobi
-    m_levelFRelaxMethod[2] = 0; // Jacobi
+    m_levelFRelaxMethod[0] = hypre::getAMGRelaxationType( LinearSolverParameters::AMG::SmootherType::jacobi );
+    m_levelFRelaxMethod[1] = hypre::getAMGRelaxationType( LinearSolverParameters::AMG::SmootherType::jacobi );
+    m_levelFRelaxMethod[2] = hypre::getAMGRelaxationType( LinearSolverParameters::AMG::SmootherType::jacobi );
```
klevzoff, line 75: I think you're assigning the wrong thing here, relaxation type (0..18) to relaxation method ({0,1,2}). The old comment `// Jacobi` was probably also incorrect?

src/coreComponents/linearAlgebra/interfaces/hypre/mgrStrategies/MultiphasePoromechanics.hpp
```diff
@@ -71,17 +71,17 @@ class MultiphasePoromechanics : public MGRStrategyBase< 3 >
...
-    m_levelInterpType[1] = 2;       // diagonal scaling (Jacobi)
-    m_levelCoarseGridMethod[1] = 0; // standard Galerkin
-    m_levelInterpType[2] = 2;       // diagonal scaling (Jacobi)
-    m_levelCoarseGridMethod[2] = 0; // standard Galerkin
+    m_levelInterpType[0] = toUnderlying( hypre::MGRLevelInterpolationType::jacobi );
+    m_levelCoarseGridMethod[0] =toUnderlying( hypre::MGRLevelCoarseGridMethod::nonGalerkin );
+    m_levelInterpType[1] = toUnderlying( hypre::MGRLevelInterpolationType::jacobi );
+    m_levelCoarseGridMethod[1] = toUnderlying( hypre::MGRLevelCoarseGridMethod::galerkin );
+    m_levelInterpType[2] = toUnderlying( hypre::MGRLevelInterpolationType::jacobi );
+    m_levelCoarseGridMethod[2] = toUnderlying( hypre::MGRLevelCoarseGridMethod::galerkin );
```
klevzoff, line 84: Would it be feasible to change the declaration of all these variables in `MGRStrategyBase` to actually be `enum`s, and only call `toUnderlying` when passing values into _hypre_ calls? Thus we can enforce stronger typing on our side and avoid potential errors.

This might not work if at any point a _hypre_ call requires us to pass a **pointer** to `HYPRE_Int`.

src/coreComponents/linearAlgebra/interfaces/hypre/HypreUtils.hpp
```diff
@@ -259,7 +262,247 @@ HYPRE_Int RelaxationSolve( HYPRE_Solver solver,
...
+  blockJacobi = 0, //!< block Jacobi (default)
+  jacobi = 1,      //!< Jacobi
+  ilu0 = 16        //!< incomplete LU factorization
+         // nameTBD = 2, //!< Gauss-Seidel, sequential (very slow!)
+         // nameTBD = 3, //!< Gauss-Seidel, interior points in parallel, boundary sequential (slow!)
+         // nameTBD = 4, //!< hybrid Gauss-Seidel or SOR, forward solve
+         // nameTBD = 5, //!< hybrid Gauss-Seidel or SOR, backward solve
+         // nameTBD = 6, //!< hybrid chaotic Gauss-Seidel (works only with OpenMP)
+         // nameTBD = 7, //!< hybrid symmetric Gauss-Seidel or SSOR
+         // nameTBD = 8, //!< Euclid (ILU)
```
klevzoff, line 504: Are these not defined because they are not yet supported by MGR?

klevzoff (approved):
Nice!

src/coreComponents/linearAlgebra/interfaces/hypre/mgrStrategies/CompositionalMultiphaseHybridFVM.hpp
castelletto1: You're correct. Here we need to specify the F-Relaxation method, i.e. `singleLevel`, `multilevel`, or `amgVCycle`. We would need to specify the F-Relaxation type -- in this case `jacobi`) -- introducing a new member array `m_levelFRelaxType` in `MGRStrategyBase`. However, we realized that such capability is currently missing in MGR. @oseikuffuor1 is working on it.

src/coreComponents/linearAlgebra/interfaces/hypre/HypreUtils.hpp
castelletto1: The reason for not defining all these options is that we won't use all of them -- for example definitely not the `very slow` ones. I want to have a closer look and ask @oseikuffuor1 for guidance of what to keep. In the meantime, it's probably better to remove them and add new options only when needed.

src/coreComponents/linearAlgebra/interfaces/hypre/mgrStrategies/CompositionalMultiphaseHybridFVM.hpp
```diff
@@ -70,15 +70,25 @@ class CompositionalMultiphaseHybridFVM : public MGRStrategyBase< 3 >
...
-    m_levelFRelaxMethod[1] = 0; // Jacobi
-    m_levelFRelaxMethod[2] = 0; // Jacobi
-
-    m_levelInterpType[0] = 2;
-    m_levelInterpType[1] = 2;
-    m_levelInterpType[2] = 2;
-
-    m_globalSmoothType = 16; // ILU(0)
+    // Level 0
+    m_levelFRelaxMethod[0]     = hypre::MGRFRelaxationMethod::singleLevel; //default, i.e. Jacobi (to be confirmed)
```
klevzoff, line 74: You can drop `hypre::` on these (if you want) since we're already inside `hypre` namespace

src/coreComponents/linearAlgebra/interfaces/hypre/mgrStrategies/MultiphasePoromechanics.hpp
castelletto1: Done. Following your suggestion, the **pointer** to `HYPRE_Int` has been addressed using `reinterpret_cast`, (see function [toUnderlyingPtr](https://github.com/GEOSX/GEOSX/blob/e52301b7dd5a60da6d65bde1bc74b29576f08069/src/coreComponents/codingUtilities/Utilities.hpp#L200-L210)).

PR #1859, opened 2022-03-29
Change LvArray submodule to target feature/deviceErrorMessage

XL64:


XL64:
Apparently my change on the assert leads to core dump during compilation in   Ubuntu CUDA debug (18.04, clang 8.0.0 + gcc 8.3.1, open-mpi 2.1.1, cuda-10.1.243), I have no idea how to debug that...

XL64:
I want to use printf format to have a decent error message on device. But as it is templated i don't know thé type of integer i have and which format i have to use in my printf(), but i know it should fit in an int64_t. Thus i cast and use PRId64.

XL64:
No it makes the error more readable, i was expecting it to be integrated (when the tests passe) but it is not essential.

XL64:
Any idea how I could debug that on the failling VMs ?

TotoGaz:
For `Pecan-CPU` the error message says that
```
In function 'void makeTetra(geosx::array2d<double>&, geosx::FaceManager::NodeMapType&, geosx::array1d<int>&, geosx::real64 (&)[3], geosx::real64&, geosx::real64 (&)[3], geosx::real64&, geosx::integer, geosx::arraySlice2d<double>&)':
cc1plus: error: 'void* __builtin_memset(void*, int, long unsigned int)' specified size between 18446744065119617048 and 18446744073709551612 exceeds maximum object size 9223372036854775807 [-Werror=stringop-overflow=]
```
which surely means that you have an integer conversion mismatch type pinpointed by this test.

For `Ubuntu CUDA debug` I do not know, it looks like a `Segmentation fault (core dumped)` which is complicated to debunk. You should download the image and try to build in it, that's what I would do.

XL64:
Thanks, I'll try to get the VM and work on it. Didn't know I could do that. Great.

XL64:
> [GEOSX/LvArray#250](https://github.com/GEOSX/LvArray/pull/250)

Thanks !

XL64:
My Bad, didn't build the correct branch. Now i just need to debug it. Thanks for the tips.

XL64:
Hello, I finally found that the memset error was linked to the use of localIndex in the ConditionalMultiply.
Changing this removed the error :
using localIndex  = std::ptrdiff_t; //GEOSX_LOCALINDEX_TYPE;
I don't know why my change reveals that but I think it is not linked to it.

XL.

src/cmake/GeosxOptions.cmake
```diff
@@ -141,7 +141,7 @@ if( ENABLE_HYPRE AND ENABLE_HYPRE_CUDA )
     set( GEOSX_LOCALINDEX_TYPE "int" CACHE STRING "" )
     set( GEOSX_GLOBALINDEX_TYPE "long long int" CACHE STRING "" )
 else()
-    set( GEOSX_LOCALINDEX_TYPE "int" CACHE STRING "" )
+    set( GEOSX_LOCALINDEX_TYPE "std::ptrdiff_t" CACHE STRING "" )
```
rrsettgast, line 144: we want this to be `int`

rrsettgast:
> Hello, I finally found that the memset error was linked to the use of localIndex in the ConditionalMultiply. Changing this in DataTypes.hpp removed the error : using localIndex = std::ptrdiff_t; //GEOSX_LOCALINDEX_TYPE; I don't know why my change reveals that but I think it is not linked to it. So maybe changing the localIndex to 32 bits integer was not so safe.
> 
> XL.

Sorry. Where is `ConditionalMultiply`?

XL64:
In indexing.hpp

src/cmake/GeosxOptions.cmake
XL64: Yes i know i just wanted to check Jenkins was ok with that. Maybe i Can add a cast in the multiplication routine.

XL64: Changed that back and used size_t when computing linear index, seems to be ok on pecan docker file.

XL64: Ok I broke something else. This needs some work.

XL64:
Ok it seems the std:size_t does not solve the issue completly, it depends on the number of tansMatrixRef(a,b) = c ; I comment in makeTetra... sorry for the noise.

XL.

PR #1861, opened 2022-03-29
Add Exponential Decay Permeability Model for Fault

jhuang2601:
This PR implements Exponential Decay model for fault permeability change
- [x] add Exponential Decay model [(Gutierrez et al., 2000)](https://www.sciencedirect.com/science/article/pii/S0264817200000271)
- [x] prepare documents for this permeability model
- [x] add a smoke test to the integratedTests and check the run on Pecan   
       [GEOSX/integratedTests#283](https://github.com/GEOSX/integratedTests/pull/283)


src/coreComponents/constitutive/permeability/ExponentialDecayPermeability.hpp
```diff
@@ -0,0 +1,163 @@
...
+    compute( traction,
+             m_initialPermeability,
+             m_permeability[k][0],
+             m_dPerm_dTraction[k][0] );
+  }
+
+private:
+
+  /// Derivative of fracture permeability to effective normal stress applied on the fracture surfaces
+  arrayView4d< real64 > m_dPerm_dTraction;
```
CusiniM, line 71: if you want to use this model as fracture permeability we will need to think how to pass this to the FVM kernel. Right now, you will get that `seg fault` because  we are not defining 
````c++
/// Derivative of fracture permeability w.r.t. displacement jump
  array4d< real64 > m_dPerm_dDispJump;
  ````
  
  but it tries to use it here: https://github.com/GEOSX/GEOSX/blob/e9469bf176dbe673e0edd0cb6fbce37d5a4e512c/src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseFVM.cpp#L458

**Review by herve-gross — APPROVED** [review]
Thanks! This looks good to me.

File: inputFiles/poromechanicsFractures/ExponentialDecayPermeability_EDFM_base.xml [hunk 2]
```diff
@@ -0,0 +1,267 @@
...
+      flowSolverName="flowSolver"
+      solidSolverName="matrixMechSolver"      
+      logLevel="1">
+      <NonlinearSolverParameters
+        newtonTol="1.0e-5"
+        newtonMaxIter="10"
+        maxTimeStepCuts="1"/>
+      <LinearSolverParameters
+        directParallel="0"/>
+    </SinglePhasePoromechanicsEmbeddedFractures>
```
CusiniM, line 20: We should add a test with the conforming fracture.

jhuang2601: @CusiniM A smoke test with conforming fracture is added and the running of integratedTests is checked on Pecan.

CusiniM (approved):
Please use `conforming` and `edfm` in the names of the xml files so that it's consistent with the other files. 

Otherwise this looks good to me. We should probably always compute both derivatives w.r.t to the jump and the traction and then let individual solvers take care of the chain rule based on the primary variable of choice.

CusiniM:
once it passes the checks we can press squash merge. IntegratedTests have been already merged.

PR #1862, opened 2022-03-30
Fix missing multibody loops in SolidMechanicsLagrangianFEM solver.

CusiniM:
These changes are important for the Multilevel EFEM-phase field work.


src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.cpp
```diff
@@ -1071,75 +1079,82 @@ SolidMechanicsLagrangianFEM::
 {
   GEOSX_MARK_FUNCTION;
 
-  MeshLevel const & mesh = domain.getMeshBody( 0 ).getMeshLevel( 0 );
-  NodeManager const & nodeManager = mesh.getNodeManager();
+  real64 totalResidualNorm = LvArray::NumericLimits< real64 >::max;
```
rrsettgast, line 1082: ??

rrsettgast (approved):

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.cpp
CusiniM: we need to do something with the residual when we have multiple bodies. I am taking the max across all mesh bodies.

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.cpp
```diff
@@ -1071,75 +1079,82 @@ SolidMechanicsLagrangianFEM::
...
 
-  MpiWrapper::bcast( globalResidualNorm, 2, 0, MPI_COMM_GEOSX );
+    MpiWrapper::bcast( globalResidualNorm, 2, 0, MPI_COMM_GEOSX );
 
 
-  real64 const residual = sqrt( globalResidualNorm[0] )/(globalResidualNorm[1]+1);  // the + 1 is for the first
-                                                                                    // time-step when maxForce = 0;
+    real64 const residual = sqrt( globalResidualNorm[0] )/(globalResidualNorm[1]+1); // the + 1 is for the first
+                                                                                     // time-step when maxForce = 0;
+    totalResidualNorm = std::max( residual, totalResidualNorm );
```
CusiniM, line 1149: max taken here.

klevzoff (approved):

CusiniM:
@rrsettgast, @klevzoff I had not noticed but, apparently, there is still one place where we assume a single `meshBody` and it's in https://github.com/GEOSX/GEOSX/blob/fa2011abd34505835fa1f0a1242a98ccfd4075dd/src/coreComponents/fieldSpecification/FieldSpecificationManager.hpp#L248. I think that, we need to change this or it won't be possible to work with multiple `meshBodies` or `meshLevels`. If I remember correctly the reason why we need this to work asap is that @andrembcosta needs to apply different b.c. in two different SolidMechanicsSolvers (basically a coarse global and a local refined one) on two different meshes but, right now, the mesh to which mechanics b.c. are applied is hardcoded to be the first one (body 0 and level 0). 

I propose that I change the` FieldspecificationManager::apply` method  so that it does not use domain but the specific `meshLevel` to which b.c. need to be applied. Any issue with that?

rrsettgast:
> @rrsettgast, @klevzoff I had not noticed but, apparently, there is still one place where we assume a single `meshBody` and it's in
> 
> https://github.com/GEOSX/GEOSX/blob/fa2011abd34505835fa1f0a1242a98ccfd4075dd/src/coreComponents/fieldSpecification/FieldSpecificationManager.hpp#L248
> 
> I think that, we need to change this or it won't be possible to work with multiple `meshBodies` or `meshLevels`. If I remember correctly the reason why we need this to work asap is that @andrembcosta needs to apply different b.c. in two different SolidMechanicsSolvers (basically a coarse global and a local refined one) on two different meshes but, right now, the mesh to which mechanics b.c. are applied is hardcoded to be the first one (body 0 and level 0).
> I propose that I change the` FieldspecificationManager::apply` method so that it does not use domain but the specific `meshLevel` to which b.c. need to be applied. Any issue with that?

sounds good to me.

PR #1864, opened 2022-03-31
Better warning message when wrapper is not here.

TotoGaz:
Improve error message.

### DIFF
--- a/src/coreComponents/mesh/ObjectManagerBase.cpp
+++ b/src/coreComponents/mesh/ObjectManagerBase.cpp
@@ -272,7 +272,7 @@ localIndex ObjectManagerBase::packImpl( buffer_unit_type * & buffer,
     std::set_difference( input.cbegin(), input.cend(), available.cbegin(), available.cend(), std::inserter( tmp1, tmp1.end() ) );
     if( !tmp1.empty() )
     {
-      GEOSX_LOG( "Wrappers \"" << stringutilities::join( tmp1, ", " ) << "\" were requested to \"" << getName() << "\" but are not available." );
+      GEOSX_LOG( "Wrappers \"" << stringutilities::join( tmp1, ", " ) << "\" were requested from \"" << getName() << "\" but are not available." );
     }
 
     // Not taking the wrappers that are excluded.
--- a/src/coreComponents/mesh/ObjectManagerBase.cpp
+++ b/src/coreComponents/mesh/ObjectManagerBase.cpp
@@ -272,7 +272,7 @@ localIndex ObjectManagerBase::packImpl( buffer_unit_type * & buffer,
     std::set_difference( input.cbegin(), input.cend(), available.cbegin(), available.cend(), std::inserter( tmp1, tmp1.end() ) );
     if( !tmp1.empty() )
     {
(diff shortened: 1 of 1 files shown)


src/coreComponents/mesh/ObjectManagerBase.cpp
```diff
@@ -272,7 +272,7 @@ localIndex ObjectManagerBase::packImpl( buffer_unit_type * & buffer,
     std::set_difference( input.cbegin(), input.cend(), available.cbegin(), available.cend(), std::inserter( tmp1, tmp1.end() ) );
     if( !tmp1.empty() )
     {
-      GEOSX_LOG( "Wrappers \"" << stringutilities::join( tmp1, ", " ) << "\" were requested and are not available." );
+      GEOSX_LOG( "Wrappers \"" << stringutilities::join( tmp1, ", " ) << "\" were requested to " << getName() << " but are not available." );
```
CusiniM, line 275: I think we could make it print only on rank 0. Unless it makes sense to know on which rank it failed.

CusiniM (approved):

src/coreComponents/mesh/ObjectManagerBase.cpp
```diff
@@ -272,7 +272,7 @@ localIndex ObjectManagerBase::packImpl( buffer_unit_type * & buffer,
     std::set_difference( input.cbegin(), input.cend(), available.cbegin(), available.cend(), std::inserter( tmp1, tmp1.end() ) );
     if( !tmp1.empty() )
     {
-      GEOSX_LOG( "Wrappers \"" << stringutilities::join( tmp1, ", " ) << "\" were requested and are not available." );
+      GEOSX_LOG( "Wrappers \"" << stringutilities::join( tmp1, ", " ) << "\" were requested from \"" << getName() << "\" but are not available." );
```
rrsettgast, line 275: should this be an error?

TotoGaz: I originally wanted this to be an error, but I realized that there were actually multiple of our integrated tests that were impacted. That would have meant a broader fix, and furthermore, actions on multiple locations of the code that I do not master at all.
I should open an issue pinpointing the impacted cases and let people help me with this!

jhuang2601 (approved):

src/coreComponents/mesh/ObjectManagerBase.cpp
klevzoff: Here's a legitimate use case where the warning is not desirable: when running simulation with a cell-centered field that is registered on part of the domain (e.g. `pressure` field defined only in reservoir zone), our current API for `synchronizeFields()` does not allow specifying regions to operate on, it attempts to to pack the field on every element region. Previously regions where the field does not exist would silently get ignored, but now the warnings (repeated from every rank, every newton iteration) render the simulation log unreadable.

TotoGaz: > our current API for `synchronizeFields()` does not allow specifying regions to sync, it attempts to to pack the field on every element region

Don't shoot the messenger.
I would blame the `synchronizeFields` that should evolve to allow the the definition of regions?
Or maybe at least try not to pack a field if it's not there?

PR #1865, opened 2022-04-01
Low order elastic wave equation

acitrain:
First order Spectral Element Method (SEM) for the second order elastic wave equation.

In this PR we propose the first version of the elastic wave solver using the SEM.

The space discretization is performed using the finite element implementation provide in GEOSX. For the time integration, we use first order central finite difference scheme.

As for the acoustic solver, we implements first order Absorbing Boundary Condition.

The source injection is done using a Ricker wavelet but the source coefficients are computed using the gradient of the basis functions and not just the basis functions evaluated at the source point.
This seems to generate "unwanted" S-waves in particular when the source point is not placed at the center of the element.
This behaviour is due to the numerical scheme and not to  a  bug inside the code.

P.R.: https://github.com/GEOSX/integratedTests/pull/265


sframba:
> This seems to generate "unwanted" S-waves in particular when the source point is not placed at the center of the element.

Could you please specify that this behavior is expected and due to the numerical scheme, not to a bug?

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEM.cpp
```diff
@@ -0,0 +1,746 @@
...
+                  InputError );
+
+
+}
+
+
+void ElasticWaveEquationSEM::registerDataOnMesh( Group & meshBodies )
+{
+
+  meshBodies.forSubGroups< MeshBody >( [&]( MeshBody & meshBody )
```
sframba, line 129: why not a forMeshTargets( ) loop like in the Acoustic case?

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEM.cpp
```diff
@@ -0,0 +1,746 @@
...
+    nodeManager.registerExtrinsicData< extrinsicMeshData::Displacementx_nm1,
+                                       extrinsicMeshData::Displacementy_nm1,
+                                       extrinsicMeshData::Displacementz_nm1,
+                                       extrinsicMeshData::Displacementx_n,
+                                       extrinsicMeshData::Displacementy_n,
+                                       extrinsicMeshData::Displacementz_n,
+                                       extrinsicMeshData::Displacementx_np1,
+                                       extrinsicMeshData::Displacementy_np1,
+                                       extrinsicMeshData::Displacementz_np1,
+                                       extrinsicMeshData::ForcingRHS_x,
```
sframba, line 145: I think for consistency in naming there should be no underscore before x, y or z (if @francoishamon agrees)

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEM.cpp
```diff
@@ -0,0 +1,746 @@
...
+  // Output will then only be done via the previous code.
+  if( iSeismo == m_nsamplesSeismoTrace - 1 )
+  {
+    forAll< serialPolicy >( receiverConstants.size( 0 ), [=] ( localIndex const ircv )
+    {
+      if( this->m_outputSeismoTrace == 1 )
+      {
+        if( receiverIsLocal[ircv] == 1 )
+        {
+          // Note: this "manual" output to file is temporary
```
sframba, line 401: You can remove these two comments, we already have a `TODO` above saying the same thing

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEM.cpp
```diff
@@ -0,0 +1,746 @@
...
+    arrayView2d< real64 const, nodes::REFERENCE_POSITION_USD > const X = nodeManager.referencePosition().toViewConst();
+
+    /// Get table containing all the face normals
+    arrayView2d< real64 const > const faceNormal  = faceManager.faceNormal();
+    ArrayOfArraysView< localIndex const > const facesToNodes = faceManager.nodeList().toViewConst();
+
+    arrayView1d< integer > const & facesDomainBoundaryIndicator = faceManager.getDomainBoundaryIndicator();
+    arrayView1d< localIndex const > const freeSurfaceFaceIndicator = faceManager.getExtrinsicData< extrinsicMeshData::FreeSurfaceFaceIndicator >();
+
+    arrayView1d< real64 > const mass = nodeManager.getExtrinsicData< extrinsicMeshData::MassVector >();
```
sframba, line 454: Please also add a reinitialization for the mass matrix `mass.zero()` -- cf. bug found by Xavier

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEM.cpp
```diff
@@ -0,0 +1,746 @@
...
+        localIndex const kf = targetSet[ i ];
+        freeSurfaceFaceIndicator[kf] = 1;
+
+        localIndex const numNodes = faceToNodeMap.sizeOfArray( kf );
+        for( localIndex a=0; a < numNodes; ++a )
+        {
+          localIndex const dof = faceToNodeMap( kf, a );
+          freeSurfaceNodeIndicator[dof] = 1;
+
+          ux_np1[dof] = value;
```
sframba, line 561: Shouldn't `ux_n`, `uy_n`, `uz_n`, `ux_nm1`, `uy_nm1`, `uz_nm1` also be updated? Cf. acoustic solver

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEMKernel.hpp
```diff
@@ -0,0 +1,740 @@
...
+              {
+                real64 const alpha_x = density[k] * velocityVp[k];
+                real64 const alpha_y = density[k] * velocityVs[k];
+                real64 const alpha_z = density[k] * velocityVs[k];
+                damping_x[numNodeGl] += alpha_x*detJ*ds*N[a];
+                damping_y[numNodeGl] += alpha_y*detJ*ds*N[a];
+                damping_z[numNodeGl] += alpha_z*detJ*ds*N[a];
+              }
+              // Damping in y=ypos and y=yneg direction
+              if( faceNormal[iface][1] < 0.0 || faceNormal[iface][1] > 0.0 )
```
sframba, line 484: Could it happen that the normal components are not _exactly_ 0?  In that case the behavior of this code could be indetermined. Would something like 
`
alpha_x = density[k] * (velocityVp[k]*abs(faceNormal[iface][0])+velocityVs[k]*sqrt(faceNormal[iface][1]*faceNormal[iface][1]*+faceNormal[iface][2]*faceNormal[iface][2]*))
`
be more appropriate?

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEMKernel.hpp
```diff
@@ -0,0 +1,740 @@
...
+        real64 const Rzy_ij =  detJ*(stack.mu * gradN[j][1]*gradN[i][2] + stack.lambda * gradN[j][2]*gradN[i][1]);
+
+        real64 const localIncrement_x = (Rxx_ij * m_ux_n[m_elemsToNodes[k][j]] + Rxy_ij*m_uy_n[m_elemsToNodes[k][j]] + Rxz_ij*m_uz_n[m_elemsToNodes[k][j]]);
+        real64 const localIncrement_y = (Ryx_ij * m_ux_n[m_elemsToNodes[k][j]] + Ryy_ij*m_uy_n[m_elemsToNodes[k][j]] + Ryz_ij*m_uz_n[m_elemsToNodes[k][j]]);
+        real64 const localIncrement_z = (Rzx_ij * m_ux_n[m_elemsToNodes[k][j]] + Rzy_ij*m_uy_n[m_elemsToNodes[k][j]] + Rzz_ij*m_uz_n[m_elemsToNodes[k][j]]);
+
+        RAJA::atomicAdd< parallelDeviceAtomic >( &m_stiffnessVector_x[m_elemsToNodes[k][i]], localIncrement_x );
+        RAJA::atomicAdd< parallelDeviceAtomic >( &m_stiffnessVector_y[m_elemsToNodes[k][i]], localIncrement_y );
+        RAJA::atomicAdd< parallelDeviceAtomic >( &m_stiffnessVector_z[m_elemsToNodes[k][i]], localIncrement_z );
+      }
```
sframba, line 687: This wasn't easy to check !! But it looks good...

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEMKernel.hpp
```diff
@@ -0,0 +1,740 @@
...
+    }
+    stack.mu = m_density[k] * m_velocityVs[k] * m_velocityVs[k];
+    stack.lambda = m_density[k] *m_velocityVp[k] * m_velocityVp[k] - 2.0*stack.mu;
+  }
+
+  /**
+   * @copydoc geosx::finiteElement::KernelBase::quadraturePointKernel
+   *
+   * ### ExplicitAcousticSEM Description
```
sframba, line 651: ExplicitElasticSEM

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEMKernel.hpp
```diff
@@ -0,0 +1,740 @@
...
+    GEOSX_UNUSED_VAR( targetRegionIndex );
+  }
+
+
+
+  //*****************************************************************************
+  /**
+   * @copydoc geosx::finiteElement::KernelBase::StackVariables
+   *
+   * ### ExplicitAcousticSEM Description
```
sframba, line 607: ExplicitElasticSEM

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEMKernel.hpp
```diff
@@ -0,0 +1,740 @@
...
+
+};
+
+/**
+ * @brief Implements kernels for solving the acoustic wave equations
+ *   explicit central FD method and SEM
+ * @copydoc geosx::finiteElement::KernelBase
+ * @tparam SUBREGION_TYPE The type of subregion that the kernel will act on.
+ *
+ * ### AcousticWaveEquationSEMKernel Description
```
sframba, line 522: ElasticWaveEquationSEMKernel

sframba (approved):

inputFiles/wavePropagation/benchmarks/elas3D_cube2kmx2km.xml
```diff
@@ -0,0 +1,250 @@
...
+  <!-- hexahedral mesh generated internally by GEOSX -->
+  <!-- there is also an option to import external meshes -->
+  <Mesh>
+    <InternalMesh
+      name="mesh"
+      elementTypes="{ C3D8 }"
+      xCoords="{ 0, 2000 }"
+      yCoords="{ 0, 2000 }"
+      zCoords="{ 0, 2000 }"
+      nx="{ 100 }"
```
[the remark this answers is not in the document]
acitrain: It is only for benchmark. The two integrated tests are in the folder just above this one.

inputFiles/wavePropagation/benchmarks/elas3D_cube2kmx2km.xml
```diff
@@ -0,0 +1,250 @@
...
+      fieldName="displacementx_n"
+      scale="1e-32"/>
+
+    <FieldSpecification
+      name="initialdisplacementny"
+      initialCondition="1"
+      setNames="{ all }"
+      objectPath="nodeManager"
+      fieldName="displacementy_n"
+      scale="1e-32"/>
```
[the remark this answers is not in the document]
acitrain: We can set it to zero. I put 1e-32 to avoid floating point underflow but is not necessary for the integrated tests.

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEMKernel.hpp
```diff
@@ -0,0 +1,723 @@
...
+    }
+  }
+
+  GEOSX_HOST_DEVICE
+  static real64
+  evaluateRicker( real64 const & time_n,
+                  real64 const & f0,
+                  localIndex const & order )
+  {
+    real64 const o_tpeak = 1.0/f0;
```
jeannepellerin, line 138: Are we sure that this can never ever be zero ?

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEMKernel.hpp
```diff
@@ -0,0 +1,723 @@
...
+   * @param[out] sourceNodeIds indices of the nodes of the element where the source is located
+   * @param[out] sourceConstantsx constant part of the source terms in x-direction
+   * @param[out] sourceConstantsy constant part of the source terms in y-direction
+   * @param[out] sourceConstantsz constant part of the source terms in z-direction
+   * @param[in] receiverCoordinates coordinates of the receiver terms
+   * @param[out] receiverIsLocal flag indicating whether the receiver is local or not
+   * @param[out] receiverNodeIds indices of the nodes of the element where the receiver is located
+   * @param[out] receiverNodeConstants constant part of the receiver term
+   * @param[out] sourceValue array containing the value of the time dependent source (Ricker for e.g)
+   * @param[in] dt time-step
```
jeannepellerin, line 195: Consider grouping the input parameters.

jeannepellerin: Would it be meaningful to create an Class or Structure to handle the source parameters and computations?

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEM.cpp
```diff
@@ -0,0 +1,746 @@
+/*
+ * ------------------------------------------------------------------------------------------------------------
+ * SPDX-License-Identifier: LGPL-2.1-only
+ *
+ * Copyright (c) 2018-2020 Lawrence Livermore National Security LLC
+ * Copyright (c) 2018-2020 The Board of Trustees of the Leland Stanford Junior University
+ * Copyright (c) 2018-2020 Total, S.A
```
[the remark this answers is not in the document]
jeannepellerin: Company name changed in 2021. Copyright is the same across all files of the software.

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEMKernel.hpp
acitrain: It can in the case someone does not know how to use the source stuff
BY the way all those kind of source stuff will be refactor later maybe a the same time we can put a test to check f0 value and threw out the user if it is set to 0 inside the XMl (the setting itself is mandatory by the way).

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEMKernel.hpp
acitrain: Same thing for the source: it will be refactor and this can indeed be a question (even if WaveSolverBase is supposed to handle the generic stuff for all wave solvers in particular the sources)

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEM.cpp
```diff
@@ -0,0 +1,758 @@
...
+  {
+    m_nsamplesSeismoTrace = int( maxTime / m_dtSeismoTrace) + 1;
+  }
+  else
+  {
+    m_nsamplesSeismoTrace = 0;
+  }
+  localIndex const nsamples = int(maxTime/dt) + 1;
+
+  localIndex numNodesPerElem = 8;
```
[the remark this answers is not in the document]
acitrain: For the hardcoded value, the issue is the same as before: we need some structure to get it from the discretization (and avoid a loop other region which in that case is useless) and those structure are in high order branch.


If we launch the seismic solver with tet, there a test on the elemntType at a point wich return an error saying that the solver is designed for hexaedral meshes.

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEM.cpp
```diff
@@ -0,0 +1,758 @@
...
+    computeAllSeismoTraces( time_n, 0, uz_np1, uz_n, uzReceivers );
+  } );
+}
+
+void ElasticWaveEquationSEM::computeAllSeismoTraces( real64 const time_n,
+                                                     real64 const dt,
+                                                     arrayView1d< real64 const > const var_np1,
+                                                     arrayView1d< real64 const > const var_n,
+                                                     arrayView2d< real64 > varAtReceivers )
+{
```
nmbader, line 747:
```suggestion
{
  localIndex nsamplesSeismoTrace = m_nsamplesSeismoTrace;
```

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEM.cpp
```diff
@@ -0,0 +1,758 @@
...
+  for( real64 timeSeismo;
+       (timeSeismo = m_dtSeismoTrace*m_indexSeismoTrace) <= (time_n + epsilonLoc) && m_indexSeismoTrace < m_nsamplesSeismoTrace;
+       m_indexSeismoTrace++ )
```
nmbader, line 750:
```suggestion
  localIndex indexSeismoTrace = m_indexSeismoTrace;
  for( real64 timeSeismo;
       (timeSeismo = m_dtSeismoTrace*indexSeismoTrace) <= (time_n + epsilonLoc) && indexSeismoTrace < m_nsamplesSeismoTrace;
       indexSeismoTrace++ )
```

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEM.cpp
```diff
@@ -0,0 +1,758 @@
...
+  {
+    computeSeismoTrace( time_n, dt, timeSeismo, m_indexSeismoTrace, var_np1, var_n, varAtReceivers );
```
nmbader, line 752:
```suggestion
    computeSeismoTrace( time_n, dt, timeSeismo, indexSeismoTrace, var_np1, var_n, varAtReceivers );
```

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEM.cpp
```diff
@@ -0,0 +1,758 @@
...
+      uy_n[a] = uy_np1[a];
+      uz_n[a] = uz_np1[a];
+
+      stiffnessVectorx[a] = 0.0;
+      stiffnessVectory[a] = 0.0;
+      stiffnessVectorz[a] = 0.0;
+      rhsx[a] = 0.0;
+      rhsy[a] = 0.0;
+      rhsz[a] = 0.0;
+    } );
```
nmbader, line 711:
```suggestion
    } );
    for( real64 timeSeismo;
     (timeSeismo = m_dtSeismoTrace*m_indexSeismoTrace) <= (time_n + epsilonLoc) && m_indexSeismoTrace < m_nsamplesSeismoTrace;
           m_indexSeismoTrace++ )
     {
        /// do nothing 
     }
```
See my comment below to why this is needed.

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEMKernel.hpp
```diff
@@ -0,0 +1,719 @@
...
+      // all dot products should be non-negative (we enforce outward normals)
+      if( prevSign * s < 0 )
+      {
+        return false;
+      }
+      prevSign = s;
+
+    }
+    return true;
+  }
```
nmbader, line 89:
```suggestion
      // all dot products should be non-negative (we enforce outward normals)
      if( s < 0 )
      {
        return false;
      }

    }
    return true;
  }
```
This part is probably copied from the function `computationalGeometry::isPointInsidePolyhedron` which is used in the acoustic solver. I don't see why `prevSign` is needed. It is causing a severe bug in some cases: when the point (source or receiver) is located on an element face, some elements on the mesh can still detect it as local even if those elements are very far from the true point location. I was able to produce a such bug, which naturally leads to an incorrect seismogram.
This issue may belong to a different PR but I just wanted to point it out here because I was testing this elastic solver and ran into the problem.

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEMKernel.hpp
```diff
@@ -0,0 +1,719 @@
...
+
+              real64 const alphay = density[k] * (velocityVs[k]*(faceNormal[iface][0]*faceNormal[iface][0]) + velocityVp[k]*(faceNormal[iface][1]*faceNormal[iface][1]) +
+                                                  velocityVs[k]*(faceNormal[iface][2]*faceNormal[iface][2]) );
+
+              real64 const alphaz = density[k] * (velocityVs[k]*(faceNormal[iface][0]*faceNormal[iface][0]) + velocityVs[k]*(faceNormal[iface][1]*faceNormal[iface][1]) +
+                                                  velocityVp[k]*(faceNormal[iface][2]*faceNormal[iface][2]) );
+
+              dampingx[numNodeGl] += alphax*detJ*ds*N[a];
+              dampingy[numNodeGl] += alphay*detJ*ds*N[a];
+              dampingz[numNodeGl] += alphaz*detJ*ds*N[a];
```
nmbader, line 481:
```suggestion
              RAJA::atomicAdd< ATOMIC_POLICY >( &dampingx[numNodeGl], alphax*detJ*ds*N[a] );
              RAJA::atomicAdd< ATOMIC_POLICY >( &dampingy[numNodeGl], alphay*detJ*ds*N[a] );
              RAJA::atomicAdd< ATOMIC_POLICY >( &dampingz[numNodeGl], alphaz*detJ*ds*N[a] );
```

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEM.cpp
acitrain: Ok I see I'll fix it.

For the receivers and the stuff to move in WaveSolverBase I totally agree. I think for the stuff to put in WaveSolverBase it will be be a future PR at least for the acoustic case and maybe update this one directly (or change both at the same time it's something we need to discuss more then we have already done).
For the receiverts I think it is not reakky urgent because we generate then usually using PyGeosx and we get directly the array pnp1AtReceiver so we do not go threw the writing stuff inside the code (which in fact I think is devoted to disappear because it was mainly for debug purpose)

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEMKernel.hpp
acitrain: I agree prevSign is an issue. 
In fact I copied the `computationalGeometry::isPointInsidePolyhedron` for future compatibility with High Order stuff (if I do not it implies changes in a high level of GEOSX with touch every solvers and broke integrated tests).
If we do an other PR, I think we will completely change the way to locate the pouint to make it more robust, I need to speak with Stefano about that.

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEMKernel.hpp
acitrain: Agree, I'll fix it.

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEM.cpp
```diff
@@ -0,0 +1,766 @@
...
+void ElasticWaveEquationSEM::computeAllSeismoTraces( real64 const time_n,
+                                                     real64 const dt,
+                                                     arrayView1d< real64 const > const var_np1,
+                                                     arrayView1d< real64 const > const var_n,
+                                                     arrayView2d< real64 > varAtReceivers )
+{
+  localIndex indexSeismoTrace = m_indexSeismoTrace;
+
+  for( real64 timeSeismo;
+       (timeSeismo = m_dtSeismoTrace*m_indexSeismoTrace) <= (time_n + epsilonLoc) && m_indexSeismoTrace < m_nsamplesSeismoTrace;
```
nmbader, line 757:
```suggestion
       (timeSeismo = m_dtSeismoTrace*indexSeismoTrace) <= (time_n + epsilonLoc) && indexSeismoTrace < m_nsamplesSeismoTrace;
```
Just replacing the `indexSeismoTrace`

nmbader (approved):
The PR looks good to me, at least from a results point of view (below is a modeled seismogram compared to an analytical solution)

image

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEM.hpp
```diff
@@ -0,0 +1,426 @@
...
+   * @brief Compute the sesimic traces for a given variable at each receiver coordinate at a given time, using the field values at the
+   * last two timesteps.
+   * @param time_n the time corresponding to the field values at iteration n
+   * @param dt the simulation timestep
+   * @param timeSeismo the time at which the seismogram is computed
+   * @param iSeismo the index of the seismogram time in the seismogram array
+   * @param var_np1 the field values at time_n + dt
+   * @param var_n the field values at time_n
+   * @param varAtreceivers the array holding the trace values, where the output is written
+   */
```
francoishamon, line 102: It is just a small detail, but you don't have to copy-paste the Doxygen documentation for functions that are overriding base class functions, because Doxygen figures things out automatically if you only document the base class functions.

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEM.hpp
```diff
@@ -0,0 +1,426 @@
...
+                           WRITE_AND_READ,
+                           "y-component of displacement at time n-1." );
+
+EXTRINSIC_MESH_DATA_TRAIT( Displacementz_nm1,
+                           "displacementz_nm1",
+                           array1d< real64 >,
+                           0,
+                           NOPLOT,
+                           WRITE_AND_READ,
+                           "z-component of displacement at time n-1." );
```
francoishamon, line 250: For my own understanding, why did you use 3 `array1d` for displacement, instead of just one `array2d` storing the 3 components of displacement? I am asking because we already define a displacement variable in the solid mechanics solvers as an `array2d`. See here:
https://github.com/GEOSX/GEOSX/blob/36f108a19fec1ffb2429b4e7316c8bcb8fddd999/src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.cpp#L151-L155
so I just wonder if there is a good reason to do things differently in the elastic solver?

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEM.hpp
```diff
@@ -0,0 +1,426 @@
...
+  /// Displacement_np1 at the receiver location for each time step for each receiver (y-component)
+  array2d< real64 > m_displacementyNp1AtReceivers;
+
+  /// Displacement_np1 at the receiver location for each time step for each receiver (z-component)
+  array2d< real64 > m_displacementzNp1AtReceivers;
+
+};
+
+
+namespace extrinsicMeshData
```
francoishamon, line 225: The treatment of the `extrinsicMeshData` in the wave propagation solvers (elastic and acoustic) has diverged a little bit from the rest of the code. We don't include the `EXTRINSIC_MESH_DATA_TRAIT` in the hpp file of the solver anymore because this tends to introduce circular dependencies in the code.

For instance, below you define `Displacementx_n` in `ElasticWaveEquationSEM.hpp`. Then in `ElasticWaveEquationSEMKernels.hpp` you use `Displacementx_n` in the kernel here:
https://github.com/GEOSX/GEOSX/blob/5ef6490a44bc9070c7db170c078521808687caaa/src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEMKernel.hpp#L562
This means that the kernel code has to know about the solver code, which is something that we trying to avoid.

So in the rest of the code, we create a separate file, named `NameOfTheSolverExtrinsicData.hpp` that contains all the macros that you have defined below. Then we include `NameOfTheSolverExtrinsicData.hpp` both in `NameOfTheSolver.hpp` and in `NameOfTheSolverKernels.hpp`. 

We also define a namespace for each type of physics, so we have for instance `extrinsicMeshData::flow::pressure`. You can find an example [in this file](https://github.com/GEOSX/GEOSX/blob/develop/src/coreComponents/physicsSolvers/fluidFlow/FlowSolverBaseExtrinsicData.hpp).

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEM.hpp
```diff
@@ -0,0 +1,412 @@
...
+  array2d< localIndex > m_receiverNodeIds;
+
+  /// Basis function evaluated at the receiver for the nodes listed in m_receiverNodeIds
+  array2d< real64 > m_receiverConstants;
+
+  /// Flag that indicates whether the receiver is local or not to the MPI rank
+  array1d< localIndex > m_receiverIsLocal;
+
+  /// Displacement_np1 at the receiver location for each time step for each receiver (x-component)
+  array2d< real64 > m_displacementxNp1AtReceivers;
```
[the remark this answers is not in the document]
francoishamon: @acitrain What was your decision on this question from @jeannepellerin?

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEMKernel.hpp
```diff
@@ -0,0 +1,723 @@
...
+        return false;
+      }
+      prevSign = s;
+
+    }
+    return true;
+  }
+
+  /**
+   * @brief Convert a mesh element point coordinate into a coordinate on the reference element
```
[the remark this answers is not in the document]
francoishamon: Yes, you are right, the return type in the documentation can be removed

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEM.cpp
```diff
@@ -0,0 +1,776 @@
...
+
+  registerWrapper( viewKeyStruct::sourceConstantsString(), &m_receiverConstants ).
+    setInputFlag( InputFlags::FALSE ).
+    setSizedFromParent( 0 ).
+    setDescription( "Constant part of the receiver for the nodes listed in m_receiverNodeIds" );
+
+  registerWrapper( viewKeyStruct::receiverIsLocalString(), &m_receiverIsLocal ).
+    setInputFlag( InputFlags::FALSE ).
+    setSizedFromParent( 0 ).
+    setDescription( "Flag that indicates whether the receiver is local to this MPI rank" );
```
francoishamon, line 80: This can be moved to the constructor of the base class, I think.

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEM.cpp
```diff
@@ -0,0 +1,776 @@
...
+
+  if( m_dtSeismoTrace > 0 )
+  {
+    m_nsamplesSeismoTrace = int( maxTime / m_dtSeismoTrace) + 1;
+  }
+  else
+  {
+    m_nsamplesSeismoTrace = 0;
+  }
+  localIndex const nsamples = int(maxTime/dt) + 1;
```
francoishamon, line 204: This can be moved to the `postProcessInput` function of the base class, and then you will be able to call `WaveSolverBase::postProcessInput` both here and in `AcousticWaveEquationSEM`, which will remove a lot of code duplication.

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEM.cpp
```diff
@@ -0,0 +1,776 @@
...
+  localIndex const numReceiversGlobal = m_receiverCoordinates.size( 0 );
+  m_receiverNodeIds.resize( numReceiversGlobal, numNodesPerElem );
+  m_receiverConstants.resize( numReceiversGlobal, numNodesPerElem );
+  m_receiverIsLocal.resize( numReceiversGlobal );
+
+  m_displacementxNp1AtReceivers.resize( m_nsamplesSeismoTrace, numReceiversGlobal );
+  m_displacementyNp1AtReceivers.resize( m_nsamplesSeismoTrace, numReceiversGlobal );
+  m_displacementzNp1AtReceivers.resize( m_nsamplesSeismoTrace, numReceiversGlobal );
+  m_sourceValue.resize( nsamples, numSourcesGlobal );
+
```
francoishamon, line 225: For my own understanding, I see that the `resize` is done in `postProcessInput` for `ElasticWaveEquationSEM`, but in `initializePreSubGroups` in `AcousticWaveEquationSEM`. What is the right place to do it in your opinion? Should you consistify it?

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEM.cpp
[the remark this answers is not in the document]
francoishamon: I see that there is a `getNumNodesPerElem` in `AcousticWaveEquationSEM` that would address @jeannepellerin 's comment if you use it here.

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEM.cpp
```diff
@@ -0,0 +1,776 @@
...
+                        Group &,
+                        string const & )
+  {
+    string const & functionName = bc.getFunctionName();
+
+    if( functionName.empty() || functionManager.getGroup< FunctionBase >( functionName ).isFunctionOfTime() == 2 )
+    {
+      real64 const value = bc.getScale();
+
+      for( localIndex i = 0; i < targetSet.size(); ++i )
```
francoishamon, line 555: Why not use:
```
forAll< EXEC_POLICY >( targetSet.size(), [=] GEOSX_HOST_DEVICE ( localIndex const i )
```
to make sure that the data is properly moved to/from the GPU?

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEM.cpp
```diff
@@ -0,0 +1,776 @@
...
+                                                            "",
+                                                            kernelFactory );
+
+
+    addSourceToRightHandSide( cycleNumber, rhsx, rhsy, rhsz );
+
+
+
+    real64 dt2 = dt*dt;
+    forAll< serialPolicy >( nodeManager.size(), [=] ( localIndex const a )
```
francoishamon, line 654: Are you sure about the `serialPolicy` here?

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEMKernel.hpp
```diff
@@ -0,0 +1,717 @@
...
+   */
+
+  GEOSX_HOST_DEVICE
+  static bool
+  locateSourceElement( real64 const numFacesPerElem,
+                       real64 const (&elemCenter)[3],
+                       arrayView2d< real64 const > const faceNormal,
+                       arrayView2d< real64 const > const faceCenter,
+                       arraySlice1d< localIndex const > const elemsToFaces,
+                       real64 const (&coords)[3] )
```
francoishamon, line 52: For my own understanding, why not using and fixing this function [here](https://github.com/GEOSX/GEOSX/blob/36f108a19fec1ffb2429b4e7316c8bcb8fddd999/src/coreComponents/mesh/utilities/ComputationalGeometry.hpp#L368) since it is doing the same thing. Or at least using the same code in `AcousticWaveEquationSEMKernel.hpp` and `ElasticWaveEquationSEMKernel.hpp`?

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEMKernel.hpp
```diff
@@ -0,0 +1,717 @@
...
+   * @param[out] coordsOnRefElem to contain the coordinate computed in the reference element
+   * @return true if coords is inside the element num index
+   */
+  template< typename FE_TYPE >
+  GEOSX_HOST_DEVICE
+  static void
+  computeCoordinatesOnReferenceElement( real64 const (&coords)[3],
+                                        arraySlice1d< localIndex const, cells::NODE_MAP_USD - 1 > const elemsToNodes,
+                                        arrayView2d< real64 const, nodes::REFERENCE_POSITION_USD > const X,
+                                        real64 (& coordsOnRefElem)[3] )
```
francoishamon, line 104: Same here, would it be possible to reuse the function also called `computeCoordinatesOnReferenceElement` in `AcousticWaveEquationSEMKernel.hpp`? If needed, you can create a file called `WaveEquationBaseKernel.hpp` that will contain the functions that are common to the two solvers.

francoishamon (approved):
Looks great, that's a lot of work! I just made a few suggestions to make the elastic solver more consistent with the existing solvers (the acoustic one and the others as well). Feel free to address only the ones you want.

inputFiles/wavePropagation/elas3D_abcAndFreeSurface.xml
```diff
@@ -0,0 +1,249 @@
...
+      name="mesh"
+      elementTypes="{ C3D8 }"
+      xCoords="{ 0, 20 }"
+      yCoords="{ 0, 20 }"
+      zCoords="{ 0, 20 }"
+      nx="{ 20 }"
+      ny="{ 20 }"
+      nz="{ 20 }"
+      cellBlockNames="{ cb }"/>
+  </Mesh>
```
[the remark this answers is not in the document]
acitrain: Thanks François, we will take a look at it.

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEM.hpp
```diff
@@ -0,0 +1,426 @@
...
+  array1d< localIndex > m_sourceIsLocal;
+
+  /// Indices of the element nodes (in the right order) for each receiver point
+  array2d< localIndex > m_receiverNodeIds;
+
+  /// Basis function evaluated at the receiver for the nodes listed in m_receiverNodeIds
+  array2d< real64 > m_receiverConstants;
+
+  /// Flag that indicates whether the receiver is local or not to the MPI rank
+  array1d< localIndex > m_receiverIsLocal;
```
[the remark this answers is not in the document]
acitrain: You're right. We discuss that with @sframba and we will do soon a refactoring PR for wave solver in particular for the second order solver which are kind of a "base" for us before going full on the first order solver which will be used more often.

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEM.hpp
```diff
@@ -0,0 +1,426 @@
...
+
+  struct viewKeyStruct : SolverBase::viewKeyStruct
+  {
+    static constexpr char const * sourceNodeIdsString() { return "sourceNodeIds"; }
+    static constexpr char const * sourceConstantsString() { return "sourceConstants"; }
+    static constexpr char const * sourceIsLocalString() { return "sourceIsLocal"; }
+
+    static constexpr char const * receiverNodeIdsString() { return "receiverNodeIds"; }
+    static constexpr char const * receiverConstantsString() {return "receiverConstants"; }
+    static constexpr char const * receiverIsLocalString() { return "receiverIsLocal"; }
```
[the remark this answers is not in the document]
acitrain: See comment just before about the refactoring

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEM.hpp
acitrain: Okay noted, than you !

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEM.hpp
acitrain: Indeed there is not... I think when, I did the first version of the solver ( a long time ago) my understanding of GEOSX was very low so I take the easy way by putting three vectors as I did not think about the resizeDimension stuff
Since the first order now I know it so we use it^^'

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEM.hpp
acitrain: Okay I note that for our refactor, thanks François

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEM.hpp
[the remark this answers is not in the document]
acitrain: In fact I was hesitating a long time and I finally forget it... 
I think I'll put upper case for those one at least which are in fact welle hidden

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEM.cpp
acitrain: HUm I checked and in fact it is duplicated in the Acoustic solver... I think the one we should kept is the one in postProcessInput. We will check it and refactor after.

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEM.cpp
acitrain: Good question, it is done without forAll in acoustic solver so I thought that you had a reason you and @mndiaye24 ti do it but it can be changed if there is not

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEM.cpp
acitrain: Nope it needs to be changed (again old version etc...) 
Thanks !

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEMKernel.hpp
acitrain: It is because of high order. centroid3D function needs to be changed to take in account the new nodes due to high order but if we change it it broke integrated test because of solver using tetrahedra. So to not have to modify a lot of stuff I changed it locally on our solvers (now that a part of high order is merged, both solvers should be the same)

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEMKernel.hpp
acitrain: I'll check it can be a part of our refactoring to

inputFiles/wavePropagation/elas3D_abcAndFreeSurface.xml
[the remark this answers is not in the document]
sframba: I have reduced the size of the integrated (now "smoke") tests to 5x5x5. I have used the structure you proposed, although we have 3 inequivalent tests for acoustic and 2 for elastic so I could not base them on the same xml file, so there is still a bit of redudnancy

CusiniM:
I am moving this out of the merge queue for now coz it seems that you are still missing an integratedTests PR. I saw there is a branch called `feature/lowOrderWaveEqnKernel` but I wasn't sure that was the one relative to this PR. I would also suggest to avoid merging baselines in the integratedTests repository but use a rebase instead.

sframba:
> I am moving this out of the merge queue for now coz it seems that you are still missing an integratedTests PR. I saw there is a branch called `feature/lowOrderWaveEqnKernel` but I wasn't sure that was the one relative to this PR. I would also suggest to avoid merging baselines in the integratedTests repository but use a rebase instead.

That is indeed the correct integrateTests branch (there is a small name mismatch). Sorry about the merge, we'll use a rebase next time!

PR #1866, opened 2022-04-01
Fixing ordering for pyramids

### LINKED ISSUE #1841: [Other] Crash when using imported mesh with only pyramid elements
Hello. I tried to use an imported mesh with xMin = (0,0,0) and xMax = (1,1,1) named pyramid_2x2x2_crash.msh. It is composed only of pyramids and is used to perform geomechanics simulation. I received an error message when running the file (see screenshot below).

To correct it, I had to add +1 to every coordinate of each point in order to get rid of this error. So the version working is with xMin =
(1,1,1) and xMax = (2,2,2) (mesh name is pyramid_2x2x2_no_crash.msh). There is no other modification.

I performed a similar test with a mesh composed of only 1 pyramid (1_pyra_crash.msh) with xMin = (0,0,0) and xMax = (0.5,0.5,0.5). Again, I observe the same error and to correct it, I added +1 to the X-coordinates of every node (1_pyra_no_crash.mesh).

It seems that the crash was in the Jacobian computation was due to the flat quad in the pyramids.
The link to the files is : https://filesender.renater.fr/?s=download&token=ef9a4905-131f-41cc-a491-b5385853ac00

image

castelletto1:
This PR resolves #1841. Pyramids were not correctly imported due to a wrong ordering assumption in `PAMELAMeshGenerator`. The same problem was also present in `VTKMeshGenerator`. I also noticed that the ordering for wedges/prisms in `PAMELAMeshGenerator` and `VTKMeshGenerator` is not consistent with the GEOSX ordering (see [H1_Wedge_Lagrange1_Gauss6.hpp](https://github.com/GEOSX/GEOSX/blob/develop/src/coreComponents/finiteElement/elementFormulations/H1_Wedge_Lagrange1_Gauss6.hpp)).

Rebaseline done in https://github.com/GEOSX/integratedTests/pull/212


castelletto1:
It is a unclear to me why tetrahedra are treated differently by `PAMELAMeshGenerator` and `VTKMeshGenerator`:

https://github.com/GEOSX/GEOSX/blob/583be30951e73c1a3d6702d298c7ddaba6ec72e0/src/coreComponents/mesh/generators/PAMELAMeshGenerator.cpp#L135

https://github.com/GEOSX/GEOSX/blob/583be30951e73c1a3d6702d298c7ddaba6ec72e0/src/coreComponents/mesh/generators/VTKMeshGenerator.cpp#L544

I would expect the same ordering. Moreover, the finite element implementation for tetrahedra ([H1_Tetrahedron_Lagrange1_Gauss1.hpp](https://github.com/GEOSX/GEOSX/blob/develop/src/coreComponents/finiteElement/elementFormulations/H1_Tetrahedron_Lagrange1_Gauss1.hpp)) assumes the standard VTK ordering. I don't understand why in the `VTKMeshGenerator.cpp` the first to vertices are flipped, as opposed to `PAMELAMeshGenerator`.

AntoineMazuyer (approved):
Ok for me, thanks Nicola. Seems consistent now

src/coreComponents/mesh/generators/CellBlockUtilities.cpp
```diff
@@ -192,9 +192,9 @@ void getFaceNodes( ElementType const & elementType,
         {
           nodeIndices.resize( 4 );
           nodeIndices[0] = elementToNodes[iElement][0];
-          nodeIndices[1] = elementToNodes[iElement][1];
-          nodeIndices[2] = elementToNodes[iElement][2];
-          nodeIndices[3] = elementToNodes[iElement][3];
+          nodeIndices[1] = elementToNodes[iElement][2];
+          nodeIndices[2] = elementToNodes[iElement][3];
+          nodeIndices[3] = elementToNodes[iElement][1];
```
TotoGaz, line 197: @klevzoff Be aware of this patch since your PR modifies this code too!

herve-gross (approved):
Thank you very much! Let's test this.

TotoGaz (approved):

src/coreComponents/mesh/generators/CellBlockUtilities.cpp
klevzoff: I decided to jump in first, so now it's a conflict for @castelletto1 to resolve :smile: 
But the change is simple, I just split this long function into sub-functions, one per element shape.

src/coreComponents/fileIO/silo/SiloFile.cpp
```diff
@@ -1331,8 +1331,8 @@ static std::vector< int > getSiloNodeOrdering( ElementType const elementType )
     case ElementType::Triangle:      return { 0, 1, 2 };
     case ElementType::Quadrilateral: return { 0, 1, 2, 3 }; // TODO check
     case ElementType::Polygon:       return { 0, 1, 2, 3, 4, 5, 6, 7, 8 }; // TODO
-    case ElementType::Tetrahedron:   return { 1, 0, 2, 3 };
-    case ElementType::Pyramid:       return { 0, 3, 2, 1, 4, 0, 0, 0 };
+    case ElementType::Tetrahedron:   return { 0, 1, 3, 2 };
+    case ElementType::Pyramid:       return { 0, 2, 3, 1, 4, 0, 0, 0 };
```
klevzoff, line 1335: This isn't the goal of your PR, but I'm curious if (and why) these extra 0s at the end are important. Does it have something to do with squashing nodes of a hex into pyramid shape, or or can they be safely removed?

klevzoff (approved):

src/coreComponents/fileIO/silo/SiloFile.cpp
rrsettgast: That is how we used to do it in geos.

castelletto1: Removed extra 0s.

Issue #1830, opened 2022-03-18
[Bug] Duplicate named xml blocks silently merge


cssherman:
**Describe the bug**
If the user fails to provide unique values for named xml entries, then odd silent errors can occur.  For example:

```xml
<InternalWell
      name="well_c"
      wellRegionName="well_c"
      wellControlsName="well_c"
      meshName="external_mesh"
      polylineNodeCoords="{ { 9392.7232, 14683.3110, 1248.0640 },
                                           { 9392.7232, 14683.3110, 708.1640 } }"
      polylineSegmentConn="{ { 0, 1 } }"
      radius="0.1"
      numElementsPerSegment="20">
      
      <Perforation
        name="producer1_perf"
        distanceFromHead="538.9000"/>

      <Perforation
        name="producer1_perf"
        distanceFromHead="488.5905"/>

      <Perforation
        name="producer1_perf"
        distanceFromHead="438.2810"/>

    </InternalWell>
```

Reading this, the code will only populate a single perforation, with properties set by the last entry encountered by the xml parser.
For a typical user, duplicate names likely indicate a mistake in the input and not the desire to merge values.

Note: in the python xml preprocessor, we allow multiple named xml blocks _across included files_, which we use to merge the structure.
A user could still accidentally use this feature, but at least there is a static record of the mistake in the compiled xml file.

**To Reproduce**

Run any xml file using duplicated named blocks.

**Expected behavior**

Throw an error (or warning) when the code detects duplicate named xml blocks

rrsettgast:
Duplicate xml block names should be an error.

klevzoff:
As far as I can tell, we're relying on current behavior. For example, when processing an `<Included>` tag, it allows us to merge multiple `FieldSpecifications` blocks coming in from different files. Changing this behavior might break some multi-file workflows (a workaround is to use the preprocessor which does the merging properly). It would be nice to enable implicitly merging unnamed nodes, while producing a diagnostic for named ones.

PR #1868, opened 2022-04-01
Adding check for duplicate named xml entries

cssherman:
#1830


cssherman:
FYI - running the tests, it looks like there are duplicated names that need to be cleaned up

src/coreComponents/dataRepository/Group.cpp
```diff
@@ -156,6 +157,12 @@ void Group::processInputFileRecursive( xmlWrapper::xmlNode & targetNode )
       childName = childNode.name();
     }
 
+    // Make sure child names are not duplicated
+    GEOSX_ERROR_IF( std::find( childNames.begin(), childNames.end(), childName ) != childNames.end(),
+                    GEOSX_FMT( "Error: An XML block cannot contain children with duplicated names ({}/{}). ",
+                               getPath(), childName ) );
+    childNames.emplace_back( childName );
```
klevzoff, line 164: I think this will error for unnamed nodes as well, since `childName` above is set to the tag name in that case?

cssherman: Agreed.  Also, in terms of merging named nodes, this is behavior that I've allowed in the xml preprocessor.  While it could lead to the same mistakes that we are trying to iron out, we do produce a compiled xml file that can be debugged

rrsettgast (approved):

PR #1869, opened 2022-04-05
Improve VTK mesh distribution via ParMETIS

klevzoff:
Resolves https://github.com/GEOSX/GEOSX/issues/1823

Rebaseline required for the one test currently using `VTKMesh`.

Issues remaining:
  - [x] How to update MPI neighbor list after redistribution - added neighbor detection via bounding box exchange between all ranks.
  - [x] Use of `PartKway` vs `RefineKway` - added option to control partition refinement.
  - [x] How to enforce partition contiguity - this is impossible with ParMETIS.

Rebaseline: https://github.com/GEOSX/integratedTests/pull/215
(one test only, no numeric diffs)

Before:
staircase-5ranks-default

After:
staircase-5ranks-parmetis


src/coreComponents/mesh/generators/VTKMeshGeneratorTools.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+/**
+ * @file VTKMeshGeneratorTools.hpp
+ */
+
+#include "VTKMeshGeneratorTools.hpp"
+
+#include "LvArray/src/Array.hpp"
+#include "LvArray/src/ArrayOfArraysView.hpp"
+#include "LvArray/src/ChaiBuffer.hpp"
```
klevzoff, line 23:
```suggestion
```

src/coreComponents/mesh/generators/VTKMeshGenerator.cpp
```diff
@@ -239,8 +312,48 @@ std::unordered_set< int > computeMPINeighborRanks( std::vector< vtkBoundingBox >
...
+  vtkNew< vtkGenerateGlobalIds > generator;
+  generator->SetInputDataObject( &loadedMesh );
+  generator->Update();
+
+  vtkIdType const minCellsOnAnyRank = MpiWrapper::min( loadedMesh.GetNumberOfCells() );
+  if( minCellsOnAnyRank > 0 )
+  {
+    // We read a properly pre-partitioned mesh, so there should be nothing to do
+    return vtkUnstructuredGrid::SafeDownCast( generator->GetOutputDataObject( 0 ) );
+  }
```
klevzoff, line 340: Does this criterion for skipping the redistribution seem reasonable to everyone? The idea is that if every rank already has at least some cells, then we've likely loaded an already properly partitioned `.pvtu` dataset.

klevzoff:
@francoishamon whenever you have some time, could you please test this branch with meshes you previously had problems with? While this branch fixes some of my own problems, I'd like to see if it's a substantial improvement in general.

francoishamon:
> @francoishamon whenever you have some time, could you please test this branch with meshes you previously had problems with? While this branch fixes some of my own problems, I'd like to see if it's a substantial improvement in general.
> 
> EDIT: ParMETIS manual says
> 
> ```
> The graph must be initially distributed among the processors such that each processor has at least one vertex.
> ```
> 
> This may be a problem for the cases you mentioned where some ranks are completely empty after the kd-tree distribution. You'll likely get a ParMETIS error reported in the log (let me know what it says). We might need a separate fix for that.

Thanks very much for working on that @klevzoff , I have not tried yet but I should be able to do it on Thursday (tomorrow) and report the results back to you.

src/coreComponents/mesh/generators/ParMETISInterface.cpp
```diff
@@ -0,0 +1,76 @@
...
+              MPI_Comm comm )
+{
+  int const numProcs = MpiWrapper::commSize( comm );
+  idx_t const numElems = elemToNodes.size();
+
+  array1d< int64_t > part( numElems ); // all 0 by default
+  if( numProcs > 1 )
+  {
+    // Compute `elemdist` parameter (element range owned by each rank)
+    array1d< idx_t > elemCounts( numProcs + 1 );
```
klevzoff, line 44:
```suggestion
    array1d< idx_t > elemCounts;
```

klevzoff:
Folks, I intend to merge this PR soon, can I get an (hopefully approving) review?

Current state is as good as it's gonna get for now:
  * Redistribution via ParMETIS works well for all meshes I've tested, it's enabled by default but optional, and additionally the number of partitioning refinements can be controlled by the user (via XML attribute `partitionRefinement`, 1 is default, 0 to disable ParMETIS, higher values for better quality).
  * MPI neighbor candidates are now detected via an all-to-all exchange of partition bounding boxes - this is much faster than building the whole ghosting layer via `vtkGhostCellsGenerator` filter. We then let GEOSX then figure out exact neighbors and remove candidates that do not share common nodes.
  * For pre-partitioned datasets (`.pvtu`), and only if `partitionRefinement="0"`, we now don't do anything (i.e. keep original partitions) as long as every rank has at least a piece of the mesh, otherwise still redistribute as usual.

There are some outstanding issues related to reading `.pvtu` datasets that I won't be able to fix immediately. Additional investigation needed for all of these:
  * When reading on fewer ranks than partitions in file, some heap corruption occurs. I've put an explicit user-friendly error trigger in place for now.
  * When reading on exact number of ranks, and with `partitionRefinement="0"`, there are spurious ghosting errors in GEOSX.
  * When reading on more ranks than partitions in file, and with `partitionRefinement="0"`, some heap corruption or segfaults occur.

src/coreComponents/mesh/generators/ParMETISInterface.cpp
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    for( int iter = 0; iter < numRefinements; ++iter )
+    {
+      int const res = ParMETIS_V3_RefineKway( elemDist.data(), xadj, adjncy,
+                                              nullptr, nullptr, &wgtflag,
+                                              &numflag, &ncon, &npart, tpwgts.data(),
+                                              &ubvec, options, &edgecut, part.data(), &comm );
+      checkReturnValue( res, "RefineKway" );
+    }
```
francoishamon, line 97: I went back to the tests that we discussed (~6M cells, 64 ranks) and tested with:
- `partitionRefinement="1"`
- `partitionRefinement="2"`
- `partitionRefinement="3"`

I think that with  `partitionRefinement="1"` the quality is already significantly improved compared to the native VTK partitions. As `partitionRefinement` is increased, we see the improvement of the partitions, although the shape of the original VTK partitions is still visible for 3 (but it is not an issue at all). When the PR is merged, I will further test to see what happens if I choose 10, 20, etc

src/coreComponents/mesh/generators/VTKMeshGenerator.cpp
```diff
@@ -148,101 +159,180 @@ loadVTKMesh( Path const & filePath )
...
- * @param[in] cuts the bounding boxes used by the VTK partitioner for all ranks
- * @details Compute the rank neighbors. The assumption is that 2 ranks are neighbors if
- * the corresponding bounding boxes intersect.
+ * @brief Compute the rank neighbor candidate list.
+ * @param[in] boundingBoxes the bounding boxes used by the VTK partitioner for all ranks
+ * @return the list of neighboring MPI ranks, will be updated
  */
-std::unordered_set< int > computeMPINeighborRanks( std::vector< vtkBoundingBox > cuts )
+std::vector< int >
+findNeighborRanks( std::vector< vtkBoundingBox > boundingBoxes )
```
francoishamon, line 274: With the latest commit of the PR, I confirm that the new approach to find the neighbor ranks works very well. Previously I had to disable the ghosting to reach the end of initialization, but now everything goes smoothly, the ghost cells seem to be exchanged without issue, and the simulation can proceed.

src/coreComponents/mesh/generators/VTKMeshGenerator.cpp
```diff
@@ -954,18 +1045,21 @@ void VTKMeshGenerator::generateMesh( DomainPartition & domain )
...
-    std::unordered_set< int > const neighbors = computeMPINeighborRanks( std::move( cuts ) );
+    GEOSX_LOG_LEVEL_RANK_0( 2, "  reading the dataset..." );
+    vtkSmartPointer< vtkUnstructuredGrid > loadedMesh = vtk::loadMesh( m_filePath );
+    GEOSX_LOG_LEVEL_RANK_0( 2, "  redistributing mesh..." );
+    m_vtkMesh = vtk::redistributeMesh( *loadedMesh, comm, m_partitionRefinement );
+    GEOSX_LOG_LEVEL_RANK_0( 2, "  finding neighbor ranks..." );
+    std::vector< vtkBoundingBox > boxes = vtk::exchangeBoundingBoxes( *m_vtkMesh, comm );
+    std::vector< int > const neighbors = vtk::findNeighborRanks( std::move( boxes ) );
     domain.getMetisNeighborList().insert( neighbors.begin(), neighbors.end() );
+    GEOSX_LOG_LEVEL_RANK_0( 2, "  done!" );
```
francoishamon, line 1062: Thanks for all the logging that has been added to the mesh generators over multiple PRs, it is saving a lot of debugging time

francoishamon (approved):
Thanks @klevzoff, this is great! For the case I tested, this is really a big improvement for the shape of the MPI partitions.

klevzoff:
> There are some outstanding issues related to reading `.pvtu` datasets that I won't be able to fix immediately. Additional investigation needed for all of these:
> 
> * When reading on fewer ranks than partitions in file, some heap corruption occurs. I've put an explicit user-friendly error trigger in place for now.
> * When reading on exact number of ranks, and with `partitionRefinement="0"`, there are spurious ghosting errors in GEOSX.
> * When reading on more ranks than partitions in file, and with `partitionRefinement="0"`, some heap corruption or segfaults occur.

**Update**: all of these are caused by duplicate points that VTK produces when it merges partitions (either in the PVTU reader or the redistribution filter). The next VTK release will include fixes for the redistribution filter, as well as a new filter that can be used to "clean up" the mesh and remove duplicate points. As soon as it becomes available (likely within a month or two), I will update the code.

For now, I implemented a simple duplicate point check using an `unordered_set`. It will detect the issue and terminate with a user-friendly error rather than segfault or heap corruption. The error message also includes a suggestion to pre-sanitize datasets using "Clean grid" filter in Paraview (which helps in some cases, but not universally).

src/coreComponents/mesh/generators/VTKMeshGeneratorTools.cpp
```diff
@@ -0,0 +1,176 @@
...
+namespace geosx
+{
+namespace vtk
+{
+
+vtkSmartPointer< vtkUnstructuredGrid >
+redistribute( vtkPartitionedDataSet & localParts,
+              MPI_Comm mpiComm )
+{
+  // The code below is modified from vtkDIYKdTreeUtilities::Exchange():
```
TotoGaz, line 37: Why don't you use the original code?

src/coreComponents/mesh/generators/VTKMeshGenerator.cpp
```diff
@@ -148,101 +159,180 @@ loadVTKMesh( Path const & filePath )
...
 {
-  std::vector< int > assignments;
-  assignments.reserve( numParts );
-  std::div_t const d = std::div( numParts, numRanks );
+  localIndex const numCells = LvArray::integerConversion< localIndex >( grid.GetNumberOfCells() );
+  array1d< INDEX_TYPE > nodeCounts( numCells );
 
-  for( int i = 0; i < numRanks; ++i )
+  // GetCell() is not thread-safe, use serial policy
+  forAll< serialPolicy >( numCells, [nodeCounts = nodeCounts.toView(), &grid] ( localIndex const cellIdx )
```
TotoGaz, line 170: If you're using no parallel feature, maybe doing simple `for` loops is OK?

src/coreComponents/mesh/generators/ParMETISInterface.cpp
klevzoff: Yes, I find that additional refinement iterations have very quickly diminishing returns. As soon as the imbalance tolerance limit is reached (hardcoded to 1.05 currently), ParMETIS does not seem eager to do too many changes. I do think that the partitions it produces look reasonable in most cases.

src/coreComponents/mesh/generators/VTKMeshGenerator.cpp
```diff
@@ -148,101 +159,180 @@ loadVTKMesh( Path const & filePath )
...
   return neighbors;
 }
 
+/**
+ * @brief Generate global point/cell IDs and redistribute the mesh among MPI ranks.
+ * @param[in] loadedMesh the mesh that was loaded on one or several MPI ranks
+ * @param[in] comm the MPI communicator
+ * @param[in] useGraphPartitioning whether to
+ * @param[in,out] mpiNeighbors the set of neighboring MPI ranks, will be updated
+ */
```
TotoGaz, line 304: doxygen is out of sync.

TotoGaz (approved):

src/coreComponents/mesh/generators/VTKMeshGeneratorTools.hpp
```diff
@@ -0,0 +1,65 @@
...
+redistribute( vtkPartitionedDataSet & localParts, MPI_Comm mpiComm );
+
+/**
+ * @brief All-to-all exchange of mesh partition bounding boxes between all ranks.
+ * @param dataSet the local part of the mesh
+ * @param mpiComm the MPI communicator
+ * @return a vector of bounding boxes, one per rank in @p mpiComm
+ */
+std::vector< vtkBoundingBox >
+exchangeBoundingBoxes( vtkDataSet & dataSet, MPI_Comm mpiComm );
```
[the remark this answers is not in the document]
klevzoff: Yes, most of VTK is that way, unfortunately :( Some of the newer template data structures have const-correct APIs, but a lot of the older core classes don't.

src/coreComponents/mesh/generators/ParMETISInterface.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+#include "common/DataTypes.hpp"
+
+namespace geosx
+{
+namespace parmetis
+{
+
+/**
+ * @brief Partition a finite element mesh according to its dual graph.
```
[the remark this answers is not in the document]
klevzoff: I believe I copied that phrase from ParMETIS manual. "Finite element" here refers to the fact that the mesh is represented by elem-node map which is typical for FE codes. I agree this description is misleading, I will probably just drop the words "finite element".

src/coreComponents/mesh/generators/ParMETISInterface.cpp
```diff
@@ -0,0 +1,107 @@
...
+partMeshKway( ArrayOfArraysView< int64_t const, int64_t > const & elemToNodes,
+              MPI_Comm comm,
+              int const minCommonNodes,
+              int const numRefinements )
+{
+  int const numProcs = MpiWrapper::commSize( comm );
+  idx_t const numElems = elemToNodes.size();
+
+  array1d< int64_t > part( numElems ); // all 0 by default
+  if( numProcs > 1 )
```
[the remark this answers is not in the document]
klevzoff: Sure, I like early returns and nesting reduction. This is how I would normally write it, not sure why I didn't.

src/coreComponents/mesh/generators/VTKMeshGeneratorTools.cpp
klevzoff: The VTK code is a bit more general and complex than needed here. For example, it takes an additional `assigner` parameter which specifies the distribution of partitions. Our use case is simpler, since every local partition goes directly to only one target rank, and respectively every rank receives exactly one partition from every other. Therefore we can simplify some data structures used in exchange (e.g. use `vector` instead of `vector<vector>` for storing blocks) and do strictly less work, even if it's just O(1) less.

src/coreComponents/mesh/generators/VTKMeshGenerator.cpp
klevzoff: Thanks, fixing

src/coreComponents/mesh/generators/VTKMeshGeneratorTools.cpp
TotoGaz: But could we call the function from the lib directly, even if we have to set up some info a little more complex?

klevzoff: Ah, I see what you mean. No, we can't call it directly (that would've been my choice too) because it's only exposed in a header `vtkDIYKdTreeUtilities.h` that VTK considers "private" and does not install.

https://gitlab.kitware.com/vtk/vtk/-/blob/master/Filters/ParallelDIY2/CMakeLists.txt#L37

src/coreComponents/mesh/generators/VTKMeshGenerator.cpp
klevzoff: I'm actually refactoring this piece a little now that I've taken a second look. I found that VTK has a runtime indicator function that tells whether certain APIs of interest are thread-safe or not (I know, it's rather weird that it's runtime and not compile-time knowledge, but they have an option for switching internal storage type for connectivity arrays at runtime (!)). So I will be dispatching with either a serial or a parallel policy based on that, so the loop will stay as a RAJA kernel, but with POLICY passed as a template parameter.

Issue #1851, opened 2022-03-26
[Bug] The entries reported by `TimeHistory` when collection and output async.


jafranc:
**Describe the bug**
The entries reported by `TimeHistory` when _collection_ event is asynchronized with _output_ event are erroneous.

**To Reproduce**
1. Unzip __sm_class09.zip__[sm_cases.zip](https://github.com/GEOSX/GEOSX/files/8447643/sm_cases.zip)  
2. Run with __class09_sync.xml__ with blocks as

```
 <Events
   <PeriodicEvent
      name="timeRhoHistoryCollection"
      timeFrequency=".1e7"
      targetExactTimestep="1"
      target="/Tasks/rhoCollection" />
   <PeriodicEvent
      name="timeRhoHistoryOutput"
      timeFrequency=".1e7"
      targetExactTimestep="1"
      target="/Outputs/timeRhoHistoryOutput" />
</Events>
```
3. Plot data using __postt_class09_3.py__ which reports for two component sum of cell entries for all three collected vars (saturation, densities, componentFraction)
5. Run with __class09_sync.xml__ with blocks as
```
 <Events
   <PeriodicEvent
      name="timeRhoHistoryCollection"
       timeFrequency=".1e7"
      targetExactTimestep="1"
      target="/Tasks/rhoCollection" />
   <PeriodicEvent
      name="timeRhoHistoryOutput"
      timeFrequency="1e7" 
      targetExactTimestep="1"
      target="/Outputs/timeRhoHistoryOutput" />
</Events>
```
6. Repeat 3.

**Expected behavior**
Both cases should produce the same **hdf5** collected solution, collected with different strategies regarding output.

**Platform (please complete the following information):**
- Machine Ubuntu 18.04
- Compiler: gcc 10.x.x
- GEOSX Version [877758e65]

**Notes**
~~In the `timeHistory` output values are reorganized depending on the variable, _e.g._ for _densities_ or _componentFraction_ are ordered colwise as [ _Time_, _coords_, _value_ ] while _phaseVolumeFraction_  starts with [ _value_ , ..etc]~~

TotoGaz:
@jafranc I cannot find `class09_sync.xml` you mention (it's not in the `zip`, for instance).
I expect it to be some tutorial or something like... Could you help me?

jafranc:
[...]

PR #1873, opened 2022-04-12
Little time history reorganization

TotoGaz:
- Created an abstract base class `HistoryCollection` with minimal interface
- Created a `HistoryCollectionBase` to receive the implementations extracted from `HistoryCollection`. Let's see if we want to send it to unique implementation `PackCollection`.
- Modified the buffer calls in `HistoryCollection` to get clearer responsibilities
- Modified `BufferedHistoryIO` to create minimal interface
- Refactor `PackCollection::updateSetsIndices` to limit side effects and improve readability (and correct a bug)
- Change `TimeHistoryOutput::m_recordCount` to `TimeHistoryOutput::m_hasRecordStarted` which is treated as a boolean.
- Implemented `TestingBuffer` for testing purposes. Still do not know where to put it in final version.

Fixes https://github.com/GEOSX/GEOSX/issues/1671
Fixes https://github.com/GEOSX/GEOSX/issues/1851


src/coreComponents/fileIO/Outputs/TimeHistoryOutput.cpp
```diff
@@ -154,12 +151,11 @@ bool TimeHistoryOutput::execute( real64 const GEOSX_UNUSED_PARAM( time_n ),
                                  DomainPartition & GEOSX_UNUSED_PARAM( domain ) )
 {
   GEOSX_MARK_FUNCTION;
-  localIndex newBuffered = m_io.front()->getBufferedCount( );
   for( auto & th_io : m_io )
   {
     th_io->write( );
   }
-  m_recordCount += newBuffered;
+  m_hasRecordStarted = true;
```
TotoGaz, line 158: @wrtobin can you double check this function?

src/coreComponents/fileIO/Outputs/TimeHistoryOutput.cpp
```diff
@@ -43,69 +45,64 @@ TimeHistoryOutput::TimeHistoryOutput( string const & name,
...
   // rank == 0 does time output for the collector
-  if( MpiWrapper::commRank( ) == 0 )
+  if( MpiWrapper::commRank() == 0 )
   {
-    HistoryMetadata timeMetadata = collector.getTimeMetadata( );
-    m_io.emplace_back( std::make_unique< HDFHistIO >( outputFile, timeMetadata, m_recordCount, 1, 2, MPI_COMM_SELF ) );
-    collector.registerTimeBufferCall( [this]() { return m_io.back()->getBufferHead( ); } );
-    m_io.back()->init( !freshInit );
+    HistoryMetadata timeMetadata = collector.getTimeMetaData();
+    m_io.emplace_back( std::make_unique< HDFHistIO >( outputFile, timeMetadata, MPI_COMM_SELF ) );
```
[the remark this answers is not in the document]
wrtobin: On restart it should be nonzero, which was used to set the write head in the file-writer (TimeHistHDF.cpp), which was required to append to previously collected time-history files on a restart. Is that accommodated in a different way now? 

We could just assume any preexisting time history data wrote to the correct row index and just start appending from there, but IIRC there was some amount of hdf dataspace dimensionality checking to ensure a preexisting file with preexisting data was structured correctly (don't 100% remember if that included a restart row check).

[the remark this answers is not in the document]
TotoGaz: Thx @wrtobin for the comments.

My main concern about this was to trying to remove `BufferedHistoryIO::getBufferCount`.
While `m_recordCount` surely makes sense in the in an _HDF5 context_ (and I did  not remove it from the class), I'm not sure that it should be exposed outside.

For instance, my other little impl `TestingBuffer` still has to expose `getBufferCount`.
Would there be a way to let this into `HDF5` (where it belongs I think)?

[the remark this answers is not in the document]
wrtobin: I absolutely agree that it really only makes sense at present in an HDF5 context. 

Unfortunately we have to add something to the DataRepository to be able to restart write/read, which this value necessarily requires.

I was hoping to keep the file writers (right now just HDF5, but maybe eventually additional formats as well) "stateless" across restarts, to allow the higher-level HistoryOutput to manage the state required for the restart. 

Adding some data from the file writers to the DataRepository (in this case just the 'row' count) to the data repo is really probably the correct solution long-term. In the initial development I just erred toward a "stateless" approach versus encapsulating the term in the appropriate context.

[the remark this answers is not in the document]
TotoGaz: Could the information be recomputed instead?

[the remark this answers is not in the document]
wrtobin: We can assume that the file has written to the correct row and append from there, since on successful shutdown the file is truncated to the correct number of rows. This doesn't allow us to check on restart that the old data is where we expect to be, but will successfully append data regardless, and that data will be correct. The most likely failure mode is that a file isn't successfully truncated for some reason and we wind up with zero rows in both the 'time' itself and the actual collected data.

I can't think of a straightforward mechanism to calculate the number of 'rows' collected on a previous run other than relying on the files themselves (particularly since I think collection cadence, etc. could change on restart) over simply counting them, which is what `m_recordCount` was for.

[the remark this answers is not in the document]
wrtobin: We could also add an additional value to the file itself telling us the expected number of rows. Not sure that would provide anything over just assuming the file truncated correctly though.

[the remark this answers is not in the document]
TotoGaz: I've rolled back because it seems more complicated than what my PR is supposed to be.
I've commented that something may be done about it nevertheless.
I think this is ready for final review. I've split some files and renamed some classes to reach more consistent naming. But I changed nothing from the code.

src/coreComponents/fileIO/timeHistory/PackCollection.cpp
```diff
@@ -57,99 +57,127 @@ void PackCollection::initializePostSubGroups( )
...
+  }
+
+  if( collectAll )
   {
-    // if sets are specified we retrieve the field only from those sets
-    Group const & setGroup = targetObject->getGroup( ObjectManagerBase::groupKeyStruct::setsString() );
-    m_setsIndices.resize( numSets );
-    localIndex setIdx = 0;
-    for( auto & setName : m_setNames )
+    // Here we only have one "all" field.
```
jafranc, line 132:
```suggestion
    // Here we only have one "all" set.
```

src/coreComponents/fileIO/timeHistory/PackCollection.cpp
```diff
@@ -158,28 +186,28 @@ void PackCollection::updateSetsIndices( DomainPartition & domain )
...
       {
-        m_setsIndices[ setIdx ][ idx ] = ownedIndices[ idx ];
+        setIndices[ idx ] = ownedIndices[ idx ];
       }
     }
   }
 }
 
-localIndex PackCollection::getNumMetaCollectors( ) const
+localIndex PackCollection::numMetaDataCollectors( ) const
```
jafranc, line 205: couldn't `numMetaDataCollectors` be confused for a method returning a size or numeration rather than `int`/`bool` like code ?

jafranc: btw isn't it the same as `return m_targetIsMeshObject && !m_disableCoordCollection` by precedence rules ?

src/coreComponents/fileIO/timeHistory/PackCollection.cpp
```diff
@@ -200,42 +228,39 @@ void PackCollection::buildMetaCollectors( )
...
-    coordCollector->getWrapper< string >( PackCollection::viewKeysStruct::fieldNameString() ).reference() = string( coordField );
-    coordCollector->getWrapper< string_array >( PackCollection::viewKeysStruct::setNamesString() ).reference() = m_setNames;
-    coordCollector->getWrapper< localIndex >( PackCollection::viewKeysStruct::onlyOnSetChangeString() ).reference() = 1;
-    // don't recursively keep creating coordinate metacollectors
-    coordCollector->disableCoordCollection( );
-    m_metaCollectors.push_back( std::move( coordCollector ) );
+    string metaName( "coordinates" ); // FIXME Name?
+    std::unique_ptr< PackCollection > collector = std::make_unique< PackCollection >( metaName, this );
+    collector->m_objectPath = m_objectPath;
+    collector->m_fieldName = coordField ? string( coordField ) : m_fieldName;
```
jafranc, line 239: Not sure of what should happen in the case `coordField` stays a `nullptr`

jafranc (commented):
Many thanks for the refactoring of this `TimeHistory` mechanisms. Structure is more clear now. :+1:

src/coreComponents/fileIO/timeHistory/PackCollection.cpp
TotoGaz: I still need to finish this kind of "details"!

src/coreComponents/fileIO/timeHistory/PackCollection.cpp
TotoGaz: Indeed, I was confused too.
It happens that the global contract seems to allow multiple `metaDataCollectors`, but the implementation allows only one (since the `bool` is 0 or 1).
I decided not to get too much in the details here and accept the pattern as it is.

Could you further explain @wrtobin because I do not have the full picture here?

> btw isn't it the same as return m_targetIsMeshObject && !m_disableCoordCollection by precedence rules ?

Indeed but I wanted to explicitly convert the `bool` in `localIndex` (this should/could be a `warning`/`error`).
I'll write a line of comment here.

src/coreComponents/fileIO/timeHistory/HistoryCollectionBase.hpp
```diff
@@ -0,0 +1,120 @@
...
+#endif
+
+namespace geosx
+{
+
+/**
+ * @brief Technical code factorisation after refactoring. Feel free to challenge this class.
+ */
+class HistoryCollectionBase : public HistoryCollection
+{
```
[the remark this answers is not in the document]
TotoGaz: Yes and no.

In GEOSX, except `CellBlockManagerABC` or `CellBlockABC` (and maybe a very few others that I'm not aware of)"_on top classes_" are not 100% abstract like `HistoryCollection` is.

To be consistent, I could rename it `HistoryCollectionABC`.
While I would prefer to rename `CellBlockManagerABC` `CellBlockManager` 🤣

src/coreComponents/fileIO/timeHistory/PackCollection.cpp
wrtobin: At present yes it seems there is only going to be one type of meta-collector, when we are collecting from a mesh object and want to collect coordinates for each object automatically. Previously we also added a separate 'time' collector, but that functionality got rolled into the primary collection object, so really this naming is an artifact of that combined with code intertia (in the absence of additional reasons to change it, time wasn't given to it). 

Also leaving it unspecialized for the particular case we have is a nod towards future proofing if we develop additional / specialized history collectors (other than pack), and they require different metacollection semantics (not happening anytime soon or even planned, but leaving the pattern of potentially multiple metacollectors also doesn't detract).

src/coreComponents/fileIO/timeHistory/HDFFile.hpp
```diff
@@ -0,0 +1,71 @@
...
+{
+public:
+  /**
+   * @brief Constructor -- this creates/opens the target file for read/write.
+   * @param fnm The filename.
+   * @param deleteExisting Whether to remove/recreate if a file with the same name exists.
+   * @param parallelAccess Whether to access one file in parallel or one file per rank in the comm.
+   * @param comm An MPI communicator where each rank in the communicator will be accesing the target file.
+   */
+  HDFFile( string const & fnm, bool deleteExisting, bool parallelAccess, struct ompi_communicator_t * comm );
```
klevzoff, line 37: @TotoGaz here

TotoGaz:
Hello @wrtobin 

I'm having hard time fixing the bug described by https://github.com/GEOSX/GEOSX/issues/1851
Basically my `TestingBuffer` class emphasizes that _the data provided are correct_, but it's the way we write them to disk which is surely faulty here.

In other words, I think that `HDFHistoryIO` most likely has a problem when it comes to dumping the data in the `hdf5` file.
But I really do not know how to compose with this `hdf5` library.

From what I had to do with `TestingBuffer`, I expect that for `hdf5` too, the `m_dims` are not in sync in with the provided data when the collection and the output are not in sync (see https://github.com/GEOSX/GEOSX/issues/1851).
In `TestingBuffer` I fixed this mismatch by storing `m_dims` every time new data are stored, even if `m_dims` do not change.

I do not quite understand how to do this for `HDFHistoryIO`, could you have a look?

TotoGaz:
Hello @wrtobin 

I'm having hard time fixing the bug described by https://github.com/GEOSX/GEOSX/issues/1851
Basically my `TestingBuffer` class emphasizes that _the data provided are correct_, but it's the way we write them to disk which is surely faulty here.

In other words, I think that `HDFHistoryIO` most likely has a problem when it comes to dumping the data in the `hdf5` file.
But I really do not know how to compose with this `hdf5` library.

From what I had to do with `TestingBuffer`, I expect that for `hdf5` too, the `m_dims` are not in sync in with the provided data when the collection and the output are not in sync (see https://github.com/GEOSX/GEOSX/issues/1851).
In `TestingBuffer` I fixed this mismatch by storing `m_dims` every time new data are stored, even if `m_dims` do not change.

I do not quite understand how to do this for `HDFHistoryIO`, could you have a look?

src/coreComponents/fileIO/timeHistory/HDFFile.cpp
```diff
@@ -82,12 +82,12 @@ HDFFile::~HDFFile()
...
 }
 
-bool HDFFile::checkInTarget( const string & name ) const
+bool HDFFile::hasDataset( const string & name ) const
 {
   int exists = 0;
   H5E_BEGIN_TRY
   {
-    exists = H5Gget_objinfo( this->operator int64_t(), name.c_str(), 0, NULL );
+    exists = H5Gget_objinfo( *this, name.c_str(), 0, NULL );
```
jafranc, line 90: is `this` implicitly convertible to `int64_t` or some `hid_t` typedef or is the called function changed ?

TotoGaz: `this` is convertible to `int64_t ` thanks to its `operator int64_t()`.

TotoGaz:
Hello, I think this is ready for final review on my side. Who wants to enjoy!?

wrtobin:
@TotoGaz I'm going to take a look at #1851 in the context of this PR when I have the chance. Hopefully today or tomorrow.
I'll set aside some time leave a final review at the same time.

TotoGaz:
Great @wrtobin thank you for your help!

TotoGaz:
Hi @wrtobin! Did you have time to investigate the issue?
Or maybe it could be better to try to merge as is (this branch solves already quite some issues) and deal with this last `async` pb in another PR?

wrtobin:
@TotoGaz I wasn't able to fully investigate the outstanding bug before I left for a vacation unfortunately, though I eliminated a couple possibilities. I will continue to take a look at the async issue when I am able this week. In the meantime I'm all for merging this PR into develop.

TotoGaz:
OK @wrtobin thanks for the update. We're on the process of merging with @francoishamon .
We'll keep you posted. Hopefully the last patch will be very little. 🤞

wrtobin:
@TotoGaz  
I have a 5-line change that should solve #1851, and actually effects the validity of all non-scalar time history writes (we weren't forwarding the buffer head appropriately on each row write for non-scalar cases).
Should I wait for this to go in and do a quick follow-up bugfix PR or would you prefer I push the fix directly to this branch?

TotoGaz:
Great news @wrtobin, thx for working on this!
@francoishamon what do you think? If you did not already perform all your tests again then let's include @wrtobin's patch?

francoishamon:
> Great news @wrtobin, thx for working on this! @francoishamon what do you think? If you did not already perform all your tests again then let's include @wrtobin's patch?

@TotoGaz sounds great, let's include the new patch and then I will run some tests again before merging. Thanks!

TotoGaz:
OK @wrtobin you can push!

TotoGaz:
@wrtobin Can you do the final review? (I'll remove `TestingBuffer`).

src/coreComponents/fileIO/timeHistory/PackCollection.cpp
```diff
@@ -57,99 +56,140 @@ void PackCollection::initializePostSubGroups( )
...
+HistoryMetadata PackCollection::getMetaData( DomainPartition const & domain, localIndex collectionIdx ) const
 {
   Group const * targetObject = this->getTargetObject( domain, m_objectPath );
-  WrapperBase const & target = targetObject->getWrapperBase( m_fieldName );
-  if( m_setNames.size() != 0 )
+  WrapperBase const & targetField = targetObject->getWrapperBase( m_fieldName );
+
+  if( collectAll() )
+  {
+    localIndex const packCount = m_setsIndices[0].size() == 0 ? targetField.size() : m_setsIndices[0].size(); // TODO CHECK!
```
TotoGaz, line 66: This looks OK; I've been in a case where I could test it.
Add comments about what `m_setsIndices` means and why we take this first index.

src/coreComponents/fileIO/timeHistory/PackCollection.cpp
```diff
@@ -57,99 +56,140 @@ void PackCollection::initializePostSubGroups( )
...
     metadata.setName( metadata.getName() + " " + m_setNames[ collectionIdx ] );
     return metadata;
   }
-  else
+}
+
+/**
+ * @brief Extracts the wrapper names that exist in the @p sets group of @p group.
+ * @param setNames The candidate wrapper names.
+ * @param omb The @p group (@p ObjectManagerBase) that which @p sets sub group may contain the @p setNames.
```
TotoGaz, line 86: Poor wording

src/coreComponents/fileIO/timeHistory/PackCollection.cpp
```diff
@@ -57,99 +56,140 @@ void PackCollection::initializePostSubGroups( )
...
 {
-  Group const * targetObject = this->getTargetObject( domain, m_objectPath );
-  WrapperBase const & target = targetObject->getWrapperBase( m_fieldName );
-  GEOSX_ERROR_IF( !target.isPackable( false ), "The object targeted for collection must be packable!" );
-  localIndex numSets = m_setNames.size( );
-  std::vector< localIndex > oldSetSizes( numSets == 0 ? 1 : numSets );
-  bool collectAll = (numSets == 0);
-  if( !collectAll )
+  // In the current function, depending on the context, the target can be considered as a `Group` or as an `ObjectManagerBase`.
+  // This convenience lambda function is a writing shortcut to get lighter syntax while not changing the whole function.
```
TotoGaz, line 110:
```suggestion
  // This convenience lambda function is a shortcut to get lighter syntax while not changing the whole function.
```

src/coreComponents/fileIO/timeHistory/PackCollection.cpp
```diff
@@ -57,99 +56,140 @@ void PackCollection::initializePostSubGroups( )
...
-    for( localIndex ii = 0; ii < targetObject->size(); ++ii )
+    ObjectManagerBase const * targetOMB = asOMB( targetGrp );
+    for( std::size_t setIdx = 0; setIdx < numSets; ++setIdx )
     {
-      m_setsIndices[0][ii] = ii;
+      string const & setName = setNames[setIdx];
+      array1d< localIndex > & setIndices = m_setsIndices[setIdx];
+
+      SortedArrayView< localIndex const > const & set = targetOMB->getSet( setName );
+      localIndex setSize = set.size();
```
TotoGaz, line 160:
```suggestion
      localIndex const setSize = set.size();
```

src/coreComponents/fileIO/timeHistory/PackCollection.cpp
```diff
@@ -200,55 +240,54 @@ void PackCollection::buildMetaCollectors( )
...
       coordField = ElementSubRegionBase::viewKeyStruct::elementCenterString();
     }
-    else
-    {
-      GEOSX_ERROR( "Data collection for \"" << m_objectPath << "\" is unimplemented." );
-    }
+//    else
+//    {
+//      GEOSX_ERROR( "Data collection for \"" << m_objectPath << "\" is unimplemented." );
+//    }
```
TotoGaz, line 246: Let's remove this

src/coreComponents/fileIO/timeHistory/PackCollection.cpp
```diff
@@ -200,55 +240,54 @@ void PackCollection::buildMetaCollectors( )
...
-    target.pack< true >( buffer, false, true, events );
+    targetField.pack< true >( buffer, false, true, events );
   }
   m_setChanged = false;
   GEOSX_ASYNC_WAIT( 6000000000, 10, testAllDeviceEvents( events ) );
 }
 
+bool PackCollection::collectAll() const
+{
+  return m_setNames.empty() or std::find( m_setNames.begin(), m_setNames.end(), "all" ) != m_setNames.end();
```
TotoGaz, line 289: Comment

francoishamon (approved):
I tested the PR on some cases and it seems to work great, thanks @TotoGaz

jhuang2601 (approved):
Thanks for fixing the reported issues on timeHistory!

src/coreComponents/fileIO/timeHistory/PackCollection.cpp
wrtobin: when `nullptr` is assigned to an actual pointer variable, it decays to the null value for that pointer type, in pretty much all cases 0, the implicit conversion of that to bool will be `false`, and nonzero value (including an uninitialized pointer var) will be true. `coordField != nullptr` (which would resolve to `coordField != (char const *)0`) would avoid the uninitialized case, though I doubt we'll ever actually run into that here.

src/coreComponents/fileIO/timeHistory/HistoryCollectionBase.cpp
```diff
@@ -0,0 +1,124 @@
...
+{
+  for( localIndex collectionIdx = 0; collectionIdx < numCollectors(); ++collectionIdx )
+  {
+    // std::function defines the == and =! comparable against nullptr_t to check the
+    // function pointer is actually assigned (an error would be thrown on the call attempt even so)
+    GEOSX_ERROR_IF( m_bufferProviders[collectionIdx] == nullptr,
+                    "History collection buffer retrieval function is unassigned, did you declare a related TimeHistoryOutput event?" );
+    // using GEOSX_ERROR_IF_EQ caused type issues since the values are used in streams
+    // we don't explicitly update the index sets here as they are updated in the buffer callback (see
+    // TimeHistoryOutput.cpp::initCollectorParallel( ) )
```
wrtobin, line 68: The opposite of this comment is now true.
```suggestion
```

src/coreComponents/fileIO/timeHistory/HistoryCollectionBase.hpp
```diff
@@ -0,0 +1,123 @@
...
+
+#if defined(GEOSX_USE_PYGEOSX)
+#include "fileIO/python/PyHistoryCollectionType.hpp"
+#endif
+
+namespace geosx
+{
+
+/**
+ * @brief Technical code factorisation after refactoring. Feel free to challenge this class.
```
wrtobin, line 28: I'm fine with this class, but we need to document or remove this comment

src/coreComponents/fileIO/timeHistory/HistoryCollectionBase.hpp
```diff
@@ -0,0 +1,123 @@
...
+   * @return Return PyHistoryCollection type.
+   */
+  virtual PyTypeObject * getPythonType() const override
+  { return python::getPyHistoryCollectionType(); }
+#endif
+
+protected:
+
+  ///@cond DO_NOT_DOCUMENT
+  // Doxygen fails with error message `warning: documented empty return type of...`
```
wrtobin, line 73: Ooh doxygen...

wrtobin (changes requested):

wrtobin:
@TotoGaz All of my review points are related to small cleanup and documentation and shouldn't require any additional testing. 

I think the refactor is looking great from a functional and interface perspective.

src/coreComponents/fileIO/timeHistory/PackCollection.cpp
```diff
@@ -200,55 +240,54 @@ void PackCollection::buildMetaCollectors( )
...
-    string metaName( "coordinates" );
-    std::unique_ptr< PackCollection > coordCollector = std::make_unique< PackCollection >( metaName, this );
-    coordCollector->getWrapper< string >( PackCollection::viewKeysStruct::objectPathString() ).reference() = m_objectPath;
-    coordCollector->getWrapper< string >( PackCollection::viewKeysStruct::fieldNameString() ).reference() = string( coordField );
-    coordCollector->getWrapper< string_array >( PackCollection::viewKeysStruct::setNamesString() ).reference() = m_setNames;
-    coordCollector->getWrapper< localIndex >( PackCollection::viewKeysStruct::onlyOnSetChangeString() ).reference() = 1;
-    // don't recursively keep creating coordinate metacollectors
-    coordCollector->disableCoordCollection( );
-    m_metaCollectors.push_back( std::move( coordCollector ) );
+    string metaName( "coordinates" ); // FIXME Name?
```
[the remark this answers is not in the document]
TotoGaz: @francoishamon I'm not very qualified about this name.
Do you have any advice?
I'd go for `collect`, `collected` or `collector`?

src/coreComponents/fileIO/timeHistory/PackCollection.cpp
TotoGaz: In this precise case, `coordField` is either defined to an existing value or explicitly set to `nullptr`.
So I guess that checking its value is valid. Am I correct?

src/coreComponents/fileIO/timeHistory/TestingBuffer.hpp
```diff
@@ -0,0 +1,67 @@
...
+  std::vector< std::vector< double > > m_buffers;
+
+  localIndex m_numStorages;
+};
+
+}
+
+
+
+#endif // include guard
```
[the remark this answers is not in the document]
TotoGaz: Yes, we'll surely do something like this eventually.
This feature is too critical to lots of our users no to be protected by some tests (integrated and/or more unitary).
We plan to think about it with @francoishamon in the very next future.

src/coreComponents/fileIO/timeHistory/PackCollection.cpp
```diff
@@ -57,99 +56,140 @@ void PackCollection::initializePostSubGroups( )
...
-  else
+}
+
+/**
+ * @brief Extracts the wrapper names that exist in the @p sets group of @p group.
+ * @param setNames The candidate wrapper names.
+ * @param omb The @p group (@p ObjectManagerBase) that which @p sets sub group may contain the @p setNames.
+ * @return The set names, guarantied to be unique.
+ */
+std::vector< string > getExistingWrapperNames( arrayView1d< string const > setNames, ObjectManagerBase const * omb )
```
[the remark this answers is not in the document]
TotoGaz: Yes, I've done it. (Like I had done in the `ObjectManagerBase` `packing` functions).
It definitely bring clearer view.

src/coreComponents/fileIO/Outputs/TimeHistoryOutput.cpp
```diff
@@ -51,60 +54,56 @@ TimeHistoryOutput::TimeHistoryOutput( string const & name,
...
-      metaMetadata.setName( collector.getTargetName() + " " + metaMetadata.getName( ) );
-      m_io.emplace_back( std::make_unique< HDFHistIO >( outputFile, metaMetadata, m_recordCount ) );
-      metaCollector.registerBufferCall( ii, [this, ii, ioCount, &domain, &metaCollector] ()
+      if( !prefix.empty() )
+      {
+        metadata.setName( prefix + metadata.getName() );
+      }
+
+      m_io.emplace_back( std::make_unique< HDFHistoryIO >( outputFile, metadata, m_recordCount ) );
+//      m_io.emplace_back( std::make_unique< TestingBuffer >( metadata.getName(), metadata.getDims() ) );
```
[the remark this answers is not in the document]
TotoGaz: done

src/coreComponents/fileIO/timeHistory/PackCollection.cpp
wrtobin: @TotoGaz Yeah since we initialize the variable in all cases it should never erroneously return `true`

src/coreComponents/fileIO/timeHistory/HistoryCollectionBase.hpp
TotoGaz: Is `Intermediate class for code factorisation` OK for you? Because this is what I wanted to express.

wrtobin: I'd go slightly stronger with `Intermediate class solely for code factorisation`, but honestly either is fine.

src/coreComponents/fileIO/timeHistory/HistoryCollectionBase.cpp
TotoGaz: Could you help me rewrite what you think is the proper comment?
I must confess it was unclear to me, so I was not able to update the comment relevantly.

src/coreComponents/fileIO/timeHistory/PackCollection.cpp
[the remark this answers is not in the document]
TotoGaz: I'm not completely aware of all of this, but I've been trying the cases of the related issues, and I could not spot the `coordinates` keyword in the `hdf5` files. I do not have the whole flow in my mind...

src/coreComponents/fileIO/timeHistory/HistoryCollectionBase.hpp
TotoGaz: 🤷

src/coreComponents/fileIO/timeHistory/HistoryCollectionBase.hpp
TotoGaz: This is fine I'll go that way.

src/coreComponents/fileIO/Outputs/TimeHistoryOutput.cpp
```diff
@@ -51,60 +54,56 @@ TimeHistoryOutput::TimeHistoryOutput( string const & name,
...
   // rank == 0 does time output for the collector
-  if( MpiWrapper::commRank( ) == 0 )
+  if( MpiWrapper::commRank() == 0 )
   {
-    HistoryMetadata timeMetadata = collector.getTimeMetadata( );
-    m_io.emplace_back( std::make_unique< HDFHistIO >( outputFile, timeMetadata, m_recordCount, 1, 2, MPI_COMM_SELF ) );
-    collector.registerTimeBufferCall( [this]() { return m_io.back()->getBufferHead( ); } );
+    HistoryMetadata timeMetadata = collector.getTimeMetaData();
+    m_io.emplace_back( std::make_unique< HDFHistoryIO >( outputFile, timeMetadata, m_recordCount, 1, 2, MPI_COMM_SELF ) );
+//    m_io.emplace_back( std::make_unique< TestingBuffer >( timeMetadata.getName(), timeMetadata.getDims() ) );
```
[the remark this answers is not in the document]
TotoGaz: done

src/coreComponents/fileIO/timeHistory/HistoryCollectionBase.cpp
wrtobin: we can actually just remove it, since the call to `this->updateSetsIndices()` on line 69 now explicitly updates the set indices, since that was previously happening inside the buffer callback called on line 71.

src/coreComponents/fileIO/timeHistory/PackCollection.cpp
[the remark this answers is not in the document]
wrtobin: It looks like I was indeed misremembering, since e.g. the async case from #1851 produces the following datasets:
```
fluid_phaseDensity       Dataset {100/Inf, 1827/Inf, 2}
fluid_phaseDensity\ Time Dataset {100/Inf, 1/Inf}
fluid_phaseDensity\ elementCenter Dataset {100/Inf, 1827/Inf, 3}
```
Where the `... elementCenter ...` dataset is what is collected by the `coordinates` collector, so I guess really this name is fairly arbitrary.

src/coreComponents/fileIO/timeHistory/HistoryCollectionBase.cpp
TotoGaz: OK thx!

src/coreComponents/fileIO/timeHistory/PackCollection.cpp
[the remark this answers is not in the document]
TotoGaz: OK; most probably some `Group` constraint that must have a `name`.
I'll go for flat `metaCollector` then.

src/coreComponents/fileIO/timeHistory/TestingBuffer.hpp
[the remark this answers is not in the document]
francoishamon: Yes, I agree, it would be great to write some tests for time history, let me know if I can help

src/coreComponents/fileIO/timeHistory/PackCollection.cpp
TotoGaz: done

wrtobin (approved):

PR #1877, opened 2022-04-14
Experimental BiCGStab implementation 

castelletto1:


src/coreComponents/linearAlgebra/utilities/AsyncRequest.hpp
```diff
@@ -0,0 +1,70 @@
...
+    m_result = std::make_unique< T >( T{} );
+    m_request = std::make_unique< MPI_Request >();
+    m_function( *m_request, *m_result );
+  }
+
+  /**
+   * @brief Completes the managed MPI request.
+   * @return the result of the non blocking operation
+   */
+  T const & complete() const
```
TotoGaz, line 54: This does not sound like the correct wording.
`complete` usually means "_If not already completed, sets the value returned by `get`and related methods to the given value._" see java doc [here](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#complete-T-).

I would change this to `get` like https://en.cppreference.com/w/cpp/thread/future/get

src/coreComponents/linearAlgebra/utilities/AsyncRequest.hpp
```diff
@@ -0,0 +1,70 @@
...
+  {
+    m_result = std::make_unique< T >( T{} );
+    m_request = std::make_unique< MPI_Request >();
+    m_function( *m_request, *m_result );
+  }
+
+  /**
+   * @brief Completes the managed MPI request.
+   * @return the result of the non blocking operation
+   */
```
TotoGaz, line 53: Please comment that this call is blocking.

src/coreComponents/linearAlgebra/utilities/AsyncRequest.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+namespace geosx
+{
+
+/**
+ * @brief Helper class for managing non-blocking operation.
+ * @tparam T the type of the result of the non-blocking operation
+ */
+template< typename T >
+class AsyncRequest
```
TotoGaz, line 32: What about using https://en.cppreference.com/w/cpp/thread/async ?

src/coreComponents/linearAlgebra/unitTests/testLinearSolverParametersEnums.cpp
```diff
@@ -33,6 +33,7 @@ TEST( LinearSolverParametersEnums, SolverType )
   ASSERT_EQ( "gmres", toString( EnumType::gmres ) );
   ASSERT_EQ( "fgmres", toString( EnumType::fgmres ) );
   ASSERT_EQ( "bicgstab", toString( EnumType::bicgstab ) );
+  ASSERT_EQ( "expBicgstab", toString( EnumType::expBicgstab ) );
```
TotoGaz, line 36: For info, were you helped by this test?

TotoGaz (commented):
Small comments about the `async` part.
Sorry for chiming in a draft PR 😉

src/coreComponents/linearAlgebra/utilities/AsyncRequest.hpp
klevzoff: I was initially a bit skeptical about `std::async` because I was unsure about object lifetimes (we need persistent addresses and lifetimes for both send/recv buffers and the `MPI_Request` object until the very end), but upon some further thinking, since the objects have to be heap-allocated anyway, it makes no problem just moving them into the completion function as it is passed to `std::async`.

Check out new implementations of `iDot` and `iDotMultiple` in `VectorBase` and `BlockVectorView` in the latest commit. It also removes `AsyncRequest` which is no longer needed. Thanks for the suggestion!

PR #1878, opened 2022-04-15
Add region list to field synchronization 

CusiniM:
Resolves https://github.com/GEOSX/GEOSX/issues/1867
Resolves https://github.com/GEOSX/GEOSX/issues/1908

- [x] @TotoGaz Change the warning message into an error here https://github.com/GEOSX/GEOSX/blob/develop/src/coreComponents/mesh/ObjectManagerBase.cpp#L271-L276.


src/coreComponents/mesh/mpiCommunications/NeighborCommunicator.cpp
```diff
@@ -620,5 +620,159 @@ void NeighborCommunicator::unpackBufferForSync( std::map< string, string_array >
...
+
+  for( auto const & fields : fieldsTobeSync )
+  {
+    fieldSyncKey key = fields.first;
+    switch( key.location )
+    {
+      case Location::Node:
+      {
+        bufferSize += nodeManager.packSize( fields.second, nodeGhostsToSend, 0, onDevice, events );
+      }
```
TotoGaz, line 651:
```suggestion
      {
        bufferSize += nodeManager.packSize( fields.second, nodeGhostsToSend, 0, onDevice, events );
        break;
      }
```
Important for all your switches.

src/coreComponents/mesh/mpiCommunications/NeighborCommunicator.cpp
```diff
@@ -620,5 +620,159 @@ void NeighborCommunicator::unpackBufferForSync( std::map< string, string_array >
...
+                                               MeshLevel const & mesh,
+                                               int const commID,
+                                               bool onDevice,
+                                               parallelDeviceEvents & events )
+{
+  GEOSX_MARK_FUNCTION;
+
+  NodeManager const & nodeManager = mesh.getNodeManager();
+  EdgeManager const & edgeManager = mesh.getEdgeManager();
+  FaceManager const & faceManager = mesh.getFaceManager();
```
TotoGaz, line 634: Maybe something like
```c++
std::map< Location, ObjectManagerBase const & > locToMgr{
  { Location::Node, mesh.getNodeManager() },
  { Location::Edge, mesh.getEdgeManager() },
  { Location::Face, mesh.getFaceManager() }
}
```
(or maybe with a `ObjectManager const *` or with `auto locToMgr{...}`) followed by
```c++
    switch( key.location )
    {
      case Location::Node:
      case Location::Edge:
      case Location::Face:
        bufferSize += locToMgr.at( key.location ).packSize( fields.second, nodeGhostsToSend, 0, onDevice, events );
        break;
      case Location::Elem:
        ...
```
?

src/coreComponents/mesh/mpiCommunications/NeighborCommunicator.hpp
```diff
@@ -46,6 +46,37 @@ class NeighborCommunicator
...
+
+  enum class Location
+  {
+    Elem, //!< location is element (like pressure in finite volumes)
+    Face, //!< location is face (like flux in mixed finite elements)
+    Edge, //!< location is edge (like flux between fracture elements)
+    Node  //!< location is node (like displacements in finite elements)
+  };
+
+  struct fieldSyncKey
```
TotoGaz, line 58:
```suggestion
  struct FieldSyncKey
```
Not a huge fan of the `Key` suffix, but we'll see in the end.

src/coreComponents/mesh/mpiCommunications/NeighborCommunicator.cpp
```diff
@@ -620,5 +620,159 @@ void NeighborCommunicator::unpackBufferForSync( std::map< string, string_array >
...
+  ElementRegionManager const & elemManager = mesh.getElemManager();
+
+  arrayView1d< localIndex const > const & nodeGhostsToSend = nodeManager.getNeighborData( m_neighborRank ).ghostsToSend();
+  arrayView1d< localIndex const > const & edgeGhostsToSend = edgeManager.getNeighborData( m_neighborRank ).ghostsToSend();
+  arrayView1d< localIndex const > const & faceGhostsToSend = faceManager.getNeighborData( m_neighborRank ).ghostsToSend();
+
+  int bufferSize = 0;
+
+  for( auto const & fields : fieldsTobeSync )
```
TotoGaz, line 643: Eventually, I think that having `FieldSyncId` which contains `location`, `regionName` and the `wrapperNames` will be the most elegant (no need to use `first` and `second` and we are only looping on the values).

Let's see what we get when it comes to filling the `map` first! We can decide in the end!

src/coreComponents/mesh/mpiCommunications/NeighborCommunicator.hpp
```diff
@@ -46,6 +46,37 @@ class NeighborCommunicator
...
+  {
+    string regionName;
+    Location location;
```
TotoGaz, line 61: Let's play with `union`! 🤣

src/coreComponents/mesh/mpiCommunications/NeighborCommunicator.cpp
TotoGaz: BTW, it's a bit strange to extract some info `mgr->getNeighborData( m_neighborRank ).ghostsToSend()` from the `mgr` in order to give it back. Maybe we can do something better, but maybe out of scope too.

CusiniM: We could probably only pass the `neighborRank` to the the `mgr`. This change would be quite invasive though ( I think)

src/coreComponents/mesh/mpiCommunications/NeighborCommunicator.hpp
CusiniM: for now it's only used as key which is why I added the key suffix but let's see what we end up doing. There is probably a better name as well.

src/coreComponents/physicsSolvers/contact/SolidMechanicsEmbeddedFractures.cpp
```diff
@@ -664,18 +664,19 @@ void SolidMechanicsEmbeddedFractures::applySystemSolution( DofManager const & do
     updateJump( dofManager, domain );
   }
 
-
+  std::vector< NeighborCommunicator::SyncFieldsID > fieldsTobeSync;
```
TotoGaz, line 667: I think that you should change this to
```c++
  std::vector< NeighborCommunicator::SyncFieldsID > fieldsTobeSync;
  forMeshTargets( domain.getMeshBodies(), [&] ( string const &,
                                                MeshLevel & mesh,
                                                arrayView1d< string const > const & )
  {
    NeighborCommunicator::SyncFieldsID fieldsID( getFractureRegionName(), NeighborCommunicator::Location::Elem );
    fieldsID.addField( string( extrinsicMeshData::contact::dispJump::key() ) );
    fieldsID.addField( string( extrinsicMeshData::contact::deltaDispJump::key()) );
    fieldsTobeSync.emplace_back( fieldsID );
  } );
  CommunicationTools::getInstance().synchronizeFields2( fieldsTobeSync,
                                                        mesh,
                                                        domain.getNeighbors(),
                                                        true );
```
or at least (I think it's less good)
```c++
  forMeshTargets( domain.getMeshBodies(), [&] ( string const &,
                                                MeshLevel & mesh,
                                                arrayView1d< string const > const & )
  {
    std::vector< NeighborCommunicator::SyncFieldsID > fieldsTobeSync;
    NeighborCommunicator::SyncFieldsID fieldsID( getFractureRegionName(), NeighborCommunicator::Location::Elem );
    fieldsID.addField( string( extrinsicMeshData::contact::dispJump::key() ) );
    fieldsID.addField( string( extrinsicMeshData::contact::deltaDispJump::key()) );
    fieldsTobeSync.emplace_back( fieldsID );
    CommunicationTools::getInstance().synchronizeFields2( fieldsTobeSync,
                                                          mesh,
                                                          domain.getNeighbors(),
                                                          true );
  } );
```
Because your implementation feeds `fieldsTobeSync` at each step of the loop and `synchronizes` fields that have already been synchronized because `fieldsTobeSync` is never flushed.

On another subject (it's a bit of a detail), having a `SyncFieldsID( string const regionName, Location const location, std::vector< string > & fieldNames)` could let you `fieldsTobeSync.emplace_back( getFractureRegionName(), NeighborCommunicator::Location::Elem, { extrinsicMeshData::contact::dispJump::key(), extrinsicMeshData::contact::deltaDispJump::key() } )`.

Also, for `fieldsID.addField( string( extrinsicMeshData::contact::dispJump::key() ) );`, maybe can you implement the proper `fieldsID.addField` such that you do not need the explicit `string( ... )` in the call?

src/coreComponents/mesh/mpiCommunications/CommunicationTools.cpp
```diff
@@ -947,5 +947,70 @@ void CommunicationTools::synchronizeFields( const std::map< string, string_array
...
+}
+
+void CommunicationTools::synchronizePackSendRecvSizes2( std::vector< NeighborCommunicator::SyncFieldsID > const & fieldsTobeSync,
+                                                        MeshLevel & mesh,
+                                                        std::vector< NeighborCommunicator > & neighbors,
+                                                        MPI_iCommData & icomm,
+                                                        bool onDevice )
+{
+  GEOSX_MARK_FUNCTION;
+  icomm.resize( neighbors.size() );
```
TotoGaz, line 981: So you removed `  icomm.fieldNames().insert( fieldNames.begin(), fieldNames.end() );`?

src/coreComponents/mesh/mpiCommunications/NeighborCommunicator.hpp
```diff
@@ -46,6 +46,32 @@ class NeighborCommunicator
...
+  {
+    Elem, //!< location is element (like pressure in finite volumes)
+    Face, //!< location is face (like flux in mixed finite elements)
+    Edge, //!< location is edge (like flux between fracture elements)
+    Node  //!< location is node (like displacements in finite elements)
+  };
+
+  struct SyncFieldsID
+  {
+    string regionName;
```
klevzoff, line 60: Can this become a list of regions? Typical use case for any cell-centered (FV) solver will be to pass its whole list of `targetRegions` for a given mesh directly. We shouldn't force the solver to additionally iterate over regions and add repeated records with the same `fieldNames`.

src/coreComponents/mesh/mpiCommunications/CommunicationTools.cpp
CusiniM: For now...I was trying to understand if that field was used anywhere.

src/coreComponents/mesh/mpiCommunications/NeighborCommunicator.hpp
CusiniM: yes, sure. Some solvers have different dofs for each region type so they won't exactly pass the full list but I guess having a region list is indeed better. 

I am also open to suggestions about where to define this class and this `enum` so that they can also be used in the `DofManager` which currently calls that `syncrhonizeFields` as well and will need to provide a list of regions. However, the `DofManager` already has a struct that basically holds the same information of this one so it would make sense to use the same one instead of having to copy the information once again.

src/coreComponents/physicsSolvers/contact/SolidMechanicsEmbeddedFractures.cpp
CusiniM: Unfortunately, the first one won't work because we need to pass the `mesh`  and that's what we are looping on so the `synchronizeFields` call has to be inside the lambda.

src/coreComponents/mesh/mpiCommunications/CommunicationTools.cpp
CusiniM: for now, this is not done. I am adding a piece at a time.

CusiniM:
@rrsettgast, @klevzoff, @TotoGaz, this works and fixes https://github.com/GEOSX/GEOSX/issues/1867. Have a look and, if you are happy with it, I can get rid of the old methods and update all the solvers. I d like for us to agree on this before spending time doing this. 

In order to fully fix the issue I will also have to change these calls in the `DofManager`:

- https://github.com/GEOSX/GEOSX/blob/8445c1c2a70eb0bf42a5c25f9c079f71fd16617e/src/coreComponents/linearAlgebra/DofManager.cpp#L203
- https://github.com/GEOSX/GEOSX/blob/8445c1c2a70eb0bf42a5c25f9c079f71fd16617e/src/coreComponents/linearAlgebra/DofManager.cpp#L1307

it will require a few changes to the `DofManager` which should also avoid defining its own L`ocation enum class ` since it could be the same as the one used by this `SyncFieldID`. 

Note that I ve finally decided to move the definition of  this struct to `DataTypes` to avoid creating  a number of new dependencies. In particular, the `MPI_iCommData` does not include much apart from DataTypes and it does need to be aware of the `SyncFieldID` struct. Anyways, One we are all happy I can go ahead and wrap this up.

src/coreComponents/physicsSolvers/contact/SolidMechanicsEmbeddedFractures.cpp
```diff
@@ -664,18 +664,22 @@ void SolidMechanicsEmbeddedFractures::applySystemSolution( DofManager const & do
...
-                                                         mesh,
-                                                         domain.getNeighbors(),
-                                                         true );
+    std::vector< SyncFieldsID > fieldsTobeSync;
+    array1d< string > regionNames, fieldNames;
+    regionNames.emplace_back( getFractureRegionName() );
+    fieldNames.emplace_back( extrinsicMeshData::contact::dispJump::key() );
+    fieldNames.emplace_back( extrinsicMeshData::contact::deltaDispJump::key() );
+
+    fieldsTobeSync.emplace_back( SyncFieldsID{ FieldLocation::Elem, regionNames, fieldNames } );
```
TotoGaz, line 677:
```suggestion
    fieldsTobeSync.emplace_back( FieldLocation::Elem, regionNames, fieldNames );
```
or
```suggestion
    fieldsTobeSync.emplace_back( SyncFieldsID( FieldLocation::Elem, regionNames, fieldNames ) );
```
?

src/coreComponents/physicsSolvers/contact/SolidMechanicsEmbeddedFractures.cpp
```diff
@@ -664,18 +664,22 @@ void SolidMechanicsEmbeddedFractures::applySystemSolution( DofManager const & do
...
+
+    CommunicationTools::getInstance().synchronizeFields2( fieldsTobeSync,
```
TotoGaz, line 679: In this precise case, this means that we'll always provide a `std::vector< SyncFieldsID >` with one unique element.
Do you think this will this always be the case? If so, we could simplify `synchronizeFields2` such that it receives a unique instance of `SyncFieldsID` and not a `std::vector`?

src/coreComponents/common/DataTypes.hpp
```diff
@@ -724,6 +724,38 @@ struct TypeName
   }
 };
 
+
+enum class FieldLocation
```
TotoGaz, line 728: I'm really not in favor of having this here.
This file contains "_technical_" types.
I do not have time right now to further check, but at least I would put this in its own header.

src/coreComponents/physicsSolvers/contact/SolidMechanicsEmbeddedFractures.cpp
CusiniM: I think having a vector can be useful. If, for example, we had a solver with a mixed formulation (e.g., velocity unknowns at the faces &  pressures at the cell centres) one may want to pass two different fields. The same could occur if the same solver had specific variables for different regions. I am not sure we have a use case right now (maybe the hybridFVM?) but I do think it's useful.

src/coreComponents/physicsSolvers/contact/SolidMechanicsEmbeddedFractures.cpp
CusiniM: I think the first one is better.

src/coreComponents/common/DataTypes.hpp
CusiniM: well, moving it to its own header is not a big deal. I just don't think it's a good idea to define it in `NeighborCommunicator` as I had tried to do before.

src/coreComponents/mesh/mpiCommunications/CommunicationTools.cpp
```diff
@@ -947,5 +947,148 @@ void CommunicationTools::synchronizeFields( const std::map< string, string_array
   synchronizeUnpack( mesh, neighbors, icomm, onDevice );
 }
 
+void CommunicationTools::synchronizeFields2( std::vector< SyncFieldsID > const & fieldsTobeSync,
```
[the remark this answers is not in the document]
CusiniM: I just didn't want to delete the old functions for now and, for some reason, the compiler does not let me define 2 methods with the same name and different signatures... The `2` designation will disappear once I remove all the old methods.

src/coreComponents/mesh/mpiCommunications/MPI_iCommData.hpp
```diff
@@ -69,6 +69,16 @@ class MPI_iCommData
    */
   std::map< string, string_array > & fieldNames() { return m_fieldNames;}
 
+  /**
+   * @return Reference to the field names registered with the communication data.
+   */
+  std::vector< SyncFieldsID > const & getFieldsTobeSync() const { return m_fieldsTobeSync;}
```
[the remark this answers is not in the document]
CusiniM: To be as is : "fields that need to be sync". We could really say unpacked since that's what this variable is used for.

src/coreComponents/common/DataTypes.hpp
CusiniM: Well, I guess we should do different things for `FieldLocation` and `SyncFieldsID`. The `SyncFieldsID` could, indeed, just be defined in the communications module. The location of a dof is a wider concept though and used across the code. The `DofManager` was defining it but if we want to use it in other placess (e.g., the communication routines) we should define it elsewhere coz I don't think we want all those things to depend on the `DofManager`.

src/coreComponents/mesh/mpiCommunications/MPI_iCommData.hpp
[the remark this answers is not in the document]
rrsettgast: that was a joke. `ToBe`

src/coreComponents/common/DataTypes.hpp
```diff
@@ -724,6 +724,38 @@ struct TypeName
...
+  Face,   //!< location is face (like flux in mixed finite elements)
+  Edge,   //!< location is edge (like flux between fracture elements)
+  Node    //!< location is node (like displacements in finite elements)
+};
+
+struct SyncFieldsID
+{
+  FieldLocation location;
+  array1d< string > regionNames;
+  array1d< string > fieldNames;
```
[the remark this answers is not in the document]
CusiniM: I don't know. I don't really have a strong opinion. In terms of readability I find using `fieldsToBeSync.fields.first `and `fieldsToBeSync.fields.second` a little worse than than just `fieldsToBeSync.regionNames` and `fieldsToBeSync.fieldNames` but it's not a huge difference.

src/coreComponents/physicsSolvers/contact/ContactSolverBase.cpp
```diff
@@ -232,17 +232,21 @@ real64 ContactSolverBase::explicitStep( real64 const & GEOSX_UNUSED_PARAM( time_
...
                                                 arrayView1d< string const > const & )
   {
-    CommunicationTools::getInstance().synchronizeFields( fieldNames,
-                                                         mesh,
-                                                         domain.getNeighbors(),
-                                                         true );
+    std::vector< SyncFieldsID > fieldsTobeSync;
+    array1d< string > regionNames, fieldNames;
+    regionNames.emplace_back( getFractureRegionName() );
+    fieldNames.emplace_back( viewKeyStruct::fractureStateString() );
```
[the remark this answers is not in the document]
CusiniM: I did this using `std::initializer_list< string >`. It becomes something like:

```
fieldsToBeSync.emplace_back( SyncFieldsID( FieldLocation::Node, regionNames, { keys::IncrementalDisplacement, keys::TotalDisplacement} ) );
```

src/coreComponents/linearAlgebra/DofManager.cpp
```diff
@@ -1276,6 +1276,8 @@ void DofManager::reorderByRank()
   // synced, and a value that contans the name of the field to be synced.
   std::map< std::pair< string, string >, std::map< string, string_array > > fieldsToSync;
 
+  std::map< std::pair< string, string >, std::vector< SyncFieldsID > > fieldsToBeSync;
```
TotoGaz, line 1279: Why are you keeping `fieldsToSync` just above?

src/coreComponents/mesh/mpiCommunications/MPI_iCommData.hpp
```diff
@@ -62,12 +62,12 @@ class MPI_iCommData
...
    * @return Reference to the field names registered with the communication data.
    */
-  std::map< string, string_array > const & getFieldNames() const { return m_fieldNames;}
+  std::vector< SyncFieldsID > const & getFieldsToBeSync() const { return m_fieldsToBeSync;}
 
   /**
    * @return Reference to the field names registered with the communication data.
    */
-  std::map< string, string_array > & fieldNames() { return m_fieldNames;}
+  std::vector< SyncFieldsID > & fieldsToBeSync() { return m_fieldsToBeSync;}
```
TotoGaz, line 70: I think you should define a real setter instead.
This solution is roughly equivalent to having a `public` variable.
But it also depends on the impact of the change...

CusiniM:
@TotoGaz I think it fixes the issue with the wrappers. feel free to test it, then approve it, and then we can merge it. Unfortunately due to a small change in `LagrangianContactSolver` which was not storing the name of the fracture region, it will need a rebaseline of the integratedTests.

src/coreComponents/linearAlgebra/DofManager.cpp
```diff
@@ -183,11 +183,10 @@ void DofManager::createIndexArray( FieldDescription const & field )
 
     forMeshSupport( field.support, *m_domain, [&]( MeshBody &, MeshLevel & mesh, auto const & regions )
     {
-      Location constexpr LOC = decltype(loc)::value;
+      FieldLocation constexpr LOC = decltype(loc)::value;
       using helper = ArrayHelper< globalIndex, LOC >;
 
-      std::map< string, string_array > fieldNames;
-      fieldNames[MeshHelper< LOC >::syncObjName].emplace_back( field.key );
+      std::vector< SyncFieldsID > fieldsToBeSync = { SyncFieldsID( field.location, {field.key}, regions ) };
```
TotoGaz, line 189:
```suggestion
      std::vector< SyncFieldsID > const fieldsToBeSync{ SyncFieldsID( field.location, {field.key}, regions ) };
```

src/coreComponents/mesh/generators/PAMELAMeshGenerator.cpp
```diff
@@ -503,9 +505,13 @@ void PAMELAMeshGenerator::importFields( DomainPartition & domain ) const
     GEOSX_ERROR_IF( cellBlockPtr == nullptr, "Internal logic error (PAMELA subregion not found)" );
 
     importFieldsOnSubRegion( *regionPtr, *cellBlockPtr, region, subRegion );
+    regionNames.emplace_back( region.getName() );
   } );
 
-  CommunicationTools::getInstance().synchronizeFields( { { "elems", m_fieldNamesInGEOSX } },
+  std::vector< string > tmp( m_fieldNamesInGEOSX.begin(), m_fieldNamesInGEOSX.end() );
+  std::vector< SyncFieldsID > fieldsToBeSync = { SyncFieldsID( FieldLocation::Elem, tmp, regionNames ) };
```
TotoGaz, line 512:
```suggestion
  std::vector< string > const tmp( m_fieldNamesInGEOSX.begin(), m_fieldNamesInGEOSX.end() );
  std::vector< SyncFieldsID > const fieldsToBeSync{ SyncFieldsID( FieldLocation::Elem, tmp, regionNames ) };
```

src/coreComponents/mesh/mpiCommunications/mpiCommunicationUtils.hpp
```diff
@@ -0,0 +1,64 @@
...
+  {}
+
+  SyncFieldsID( FieldLocation const location, std::vector< string > const & fields, std::vector< string > const & regions ):
+    location( location ),
+    fieldNames( fields ),
+    regionNames( regions )
+  {}
+
+  SyncFieldsID( FieldLocation const location, std::vector< string > const & fields ):
+    SyncFieldsID( location, fields, std::vector< string >{""} )
```
TotoGaz, line 46: You are building a `SyncFieldsID` with one region with en empty name. Change to
```suggestion
    SyncFieldsID( location, fields, std::vector< string >{} )
```
or
```suggestion
    SyncFieldsID( location, fields, {} )
```
if it's working. This will make an empty `SyncFieldsID::regionNames` which is most likely what you want to do.

src/coreComponents/mesh/mpiCommunications/mpiCommunicationUtils.hpp
```diff
@@ -0,0 +1,64 @@
...
+    regionNames( regions.begin(), regions.end() )
+  {}
+
+  SyncFieldsID( FieldLocation const location, std::vector< string > const & fields, std::vector< string > const & regions ):
+    location( location ),
+    fieldNames( fields ),
+    regionNames( regions )
+  {}
+
+  SyncFieldsID( FieldLocation const location, std::vector< string > const & fields ):
```
TotoGaz, line 45: So what is the intended behavior when there is no `regions`?

src/coreComponents/mesh/generators/VTKMeshGenerator.cpp
```diff
@@ -868,7 +868,10 @@ void VTKMeshGenerator::importFields( DomainPartition & domain ) const
     }
   }
 
-  CommunicationTools::getInstance().synchronizeFields( { { "elems", m_fieldNamesInGEOSX } },
+  std::vector< string > tmp( m_fieldNamesInGEOSX.begin(), m_fieldNamesInGEOSX.end() ), regionNames;
+  std::vector< SyncFieldsID > fieldsToBeSync{ SyncFieldsID( FieldLocation::Elem, tmp, regionNames ) };
```
TotoGaz, line 872:
```suggestion
  std::vector< string > tmp( m_fieldNamesInGEOSX.begin(), m_fieldNamesInGEOSX.end() );
  std::vector< SyncFieldsID > fieldsToBeSync{ SyncFieldsID( FieldLocation::Elem, tmp ) };
```

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.cpp
```diff
@@ -586,18 +586,26 @@ void CompositionalMultiphaseFVM::applySystemSolution( DofManager const & dofMana
...
+                                                   extrinsicMeshData::flow::deltaTemperature::key() },
+                                                 regionNames ) );
+    }
+    else
+    {
+      fieldsToBeSync.emplace_back( SyncFieldsID( FieldLocation::Elem,
+                                                 { extrinsicMeshData::flow::deltaPressure::key(),
+                                                   extrinsicMeshData::flow::deltaGlobalCompDensity::key() },
+                                                 regionNames ) );
     }
```
TotoGaz, line 606:
```suggestion
    SyncFieldsID tmp( FieldLocation::Elem, { extrinsicMeshData::flow::deltaPressure::key(), extrinsicMeshData::flow::deltaGlobalCompDensity::key() }, regionNames );
    if( m_isThermal )
    {
        tmp.fieldNames.push_back( extrinsicMeshData::flow::deltaTemperature::key() );
    }
    std::vector< SyncFieldsID > const fieldsToBeSync{ tmp };
```

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseHybridFVM.cpp
```diff
@@ -898,13 +898,20 @@ void CompositionalMultiphaseHybridFVM::applySystemSolution( DofManager const & d
...
-    fieldNames["elems"].emplace_back( extrinsicMeshData::flow::deltaGlobalCompDensity::key() );
-    CommunicationTools::getInstance().synchronizeFields( fieldNames,
+    std::vector< SyncFieldsID > fieldsToBeSync;
+    fieldsToBeSync.emplace_back( SyncFieldsID( FieldLocation::Elem,
+                                               { extrinsicMeshData::flow::deltaPressure::key(),
+                                                 extrinsicMeshData::flow::deltaGlobalCompDensity::key() },
+                                               regionNames ) );
+
+    fieldsToBeSync.emplace_back( SyncFieldsID( FieldLocation::Face,
+                                               { extrinsicMeshData::flow::deltaFacePressure::key() } ) );
```
TotoGaz, line 910:
```c++
    std::vector< SyncFieldsID > const fieldsToBeSync{
        SyncFieldsID( FieldLocation::Elem, { extrinsicMeshData::flow::deltaPressure::key(), extrinsicMeshData::flow::deltaGlobalCompDensity::key() }, regionNames ),
        SyncFieldsID( FieldLocation::Face, { extrinsicMeshData::flow::deltaFacePressure::key() } )
    };
```
I'll fix those when I try to run the tests on my box.

TotoGaz (commented):
I'll fix syntax elements when I run the cases on my box.
Two question about `SyncFieldsID` nevertheless.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.cpp
CusiniM: WE could also just emplace back a  new `SyncToBeField` entry and get rid of the temporary.

src/coreComponents/mesh/mpiCommunications/mpiCommunicationUtils.hpp
CusiniM: I think we can just leave it empty. Regions are only meaningful for `Location::Elem`

src/coreComponents/mesh/mpiCommunications/mpiCommunicationUtils.hpp
CusiniM: I think you will get that "ambiguous constructor" error because of the `arrayView1d<string>` version of the constructor but go ahead and try.

TotoGaz:
Hello @CusiniM great work. I've run your PR on my box and I could spot the following "unmanaged" wrappers:
```./update/run/proppantTransport/FlowProppantBedTransport_2d_02/FlowProppantBedTransport_2d_02.data:Wrappers "externalForce" were requested from "nodeManager" but are not available.
./update/run/proppantTransport/FlowProppantBedTransport_2d_03/FlowProppantBedTransport_2d_03.data:Wrappers "externalForce" were requested from "nodeManager" but are not available.
./update/run/proppantTransport/FlowProppantTransport_2d_04/FlowProppantTransport_2d_04.data:Wrappers "externalForce" were requested from "nodeManager" but are not available.
./update/run/proppantTransport/ProppantSlotTest_03/ProppantSlotTest_03.data:Wrappers "externalForce" were requested from "nodeManager" but are not available.
./update/run/singlePhaseFlow/fractureFlow_2d_02/fractureFlow_2d_02.data:Wrappers "externalForce" were requested from "nodeManager" but are not available.
./update/run/singlePhaseFlow/fractureFlow_2d_04/fractureFlow_2d_04.data:Wrappers "externalForce" were requested from "nodeManager" but are not available.
./update/run/singlePhaseFlow/fractureMatrixFlow_2d_04/fractureMatrixFlow_2d_04.data:Wrappers "externalForce" were requested from "nodeManager" but are not available.
```
All `externalForce` on `nodeManager`.

Otherwise I could validate (except some `Group has a child 'fractureRegionName' in the file to compare but not the baseline file.`)

src/coreComponents/mesh/mpiCommunications/mpiCommunicationUtils.hpp
```diff
@@ -0,0 +1,133 @@
...
+
+#ifndef GEOSX_MESH_MPICOMMUNICATIONS_MPICOMMUNICATIONUTILS_HPP_
+#define GEOSX_MESH_MPICOMMUNICATIONS_MPICOMMUNICATIONUTILS_HPP_
+
+#include "common/DataTypes.hpp"
+
+namespace geosx
+{
+
+struct SyncFieldsID
```
rrsettgast, line 27:
```suggestion
class fieldIdentifiers
```

Is there anything about this that is specific to the communications or syncing of fields?

src/coreComponents/mesh/mpiCommunications/mpiCommunicationUtils.hpp
```diff
@@ -0,0 +1,133 @@
...
+{
+  void addFields(FieldLocation const location, std::vector<string> const fieldNames )
+  {
+    string const key = generateKey(location);
+    addFields(fieldNames, key);
+  }
+
+  void addFields(FieldLocation const location, std::vector<string> const fieldNames, std::vector<string> const regionNames )
```
rrsettgast, line 35: I think the location is known already to be `elems`? We wouldn't ever use this for anything other than elements?

src/coreComponents/mesh/mpiCommunications/mpiCommunicationUtils.hpp
```diff
@@ -0,0 +1,133 @@
...
+  {
+    string regionName(key);
+    regionName.erase(0, 5);
+    return regionName;
+  }
+
+  static FieldLocation const getLocation( string const & key )
+  {
+    FiedlLocation location;
+    if (key.contains("nodes"))
```
rrsettgast, line 59: the strings should be stored specified in a `static constexpr char const * ` like all the viewkeystructs...which coincidently is now a terrible name

src/coreComponents/mesh/mpiCommunications/mpiCommunicationUtils.hpp
```diff
@@ -0,0 +1,133 @@
...
+
+  std::map< string, array1d<string> > const & getFields() const
+  {
+    return m_fields;
+  }
+  
+  static string const getRegionName( string const & key )
+  {
+    string regionName(key);
+    regionName.erase(0, 5);
```
rrsettgast, line 52: I think you should be able to get the length of the key once you have defined them properly as proposed below

src/coreComponents/mesh/mpiCommunications/mpiCommunicationUtils.hpp
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  static string const generateKey( FieldLocation const location, string const regionName ) const
+  {
+    if ( location == FieldLocation::Elem ) 
+    {
+      return strcat("elems/", regionName);
+    }else
+    {
+      return generateKey(location);
```
rrsettgast, line 118: error? or recursive?

src/coreComponents/mesh/mpiCommunications/mpiCommunicationUtils.hpp
```diff
@@ -0,0 +1,133 @@
...
+    }
+  }
+
+  void addFields(std::vector<string> const fieldNames, string const key)
+  {
+    for ( auto const & field : filedNames )
```
rrsettgast, line 124: no need for `auto` here?

src/coreComponents/mesh/mpiCommunications/mpiCommunicationUtils.hpp
CusiniM: yes, unless we start having regions also on other managers. I guess I can remove that but then change the function name to something like `addElementFields` ?

src/coreComponents/mesh/mpiCommunications/mpiCommunicationUtils.hpp
CusiniM: just the way is used t.b.h.

src/coreComponents/mesh/mpiCommunications/mpiCommunicationUtils.hpp
rrsettgast: Yeah...I think it is strange to take an argument, but only allow one valid value for that argument.

src/coreComponents/mesh/mpiCommunications/mpiCommunicationUtils.hpp
```diff
@@ -0,0 +1,139 @@
...
+  void addElementFields( std::vector<string> const fieldNames, std::vector<string> const regionNames )
+  {
+    for (string const & regionName : regionNames)
+    {
+      string const key = generateKey( regionName );
+      addFields(fieldNames, key);
+    }
+  }
+
+  std::map< string, array1d<string> > const & getFields() const
```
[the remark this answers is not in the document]
rrsettgast: @TotoGaz The changes are motivated by a longer term goal to treat elements and other objects in a similar manner. I view the most straight forward way to specify a field, or a group of fields is to have a `map<string,vector<string>>` where the key is the path to the object that contains the fields, and the value are the names of the fields on that object. Rather than having a separate container that may or may not be empty  to hold region names, this simply embeds the information into the key.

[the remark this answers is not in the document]
TotoGaz: If we want to go that way, it appears useful to have a dedicated strong `Key` class that would hold the appropriate information in a convenient way, could return the `std::optional` when we're not sure we have something, could store the information in `union` for mutually exclusive fields, etc.

Plain `string` is flat, colorless, cannot be encapsulated.
For end users to define it in input files, it's surely OK, but when writing C++ code for devs, it's less obvious.

CusiniM:
I have tested this and it works. Just needs approvals and a rebaseline and then it can be merged.

src/coreComponents/physicsSolvers/surfaceGeneration/SurfaceGenerator.cpp
```diff
@@ -522,11 +522,15 @@ int SurfaceGenerator::separationDriver( DomainPartition & domain,
...
-  fieldNames["face"].emplace_back( string( extrinsicMeshData::RuptureState::key() ) );
-  fieldNames["node"].emplace_back( string( SolidMechanicsLagrangianFEM::viewKeyStruct::forceExternalString() ) );
+  FieldIdentifiers fieldsToBeSync;
 
-  CommunicationTools::getInstance().synchronizeFields( fieldNames, mesh, domain.getNeighbors(), false );
+  fieldsToBeSync.addFields( FieldLocation::Face, { extrinsicMeshData::RuptureState::key() } );
+  if( nodeManager.hasWrapper( SolidMechanicsLagrangianFEM::viewKeyStruct::forceExternalString() ) )
+  {
+    fieldsToBeSync.addFields( FieldLocation::Node, { SolidMechanicsLagrangianFEM::viewKeyStruct::forceExternalString() } );
+  }
```
TotoGaz, line 531: I think you should comment why you have this conditional registratrion.

src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp
```diff
@@ -208,14 +208,18 @@ real64 HydrofractureSolver::solverStep( real64 const & time_n,
...
-        fieldNames["elems"].emplace_back( extrinsicMeshData::flow::pressure::key() );
-        fieldNames["elems"].emplace_back( extrinsicMeshData::flow::pressure_n::key() ); // in case there is a time step cut
-        fieldNames["elems"].emplace_back( "elementAperture" );
-
-        CommunicationTools::getInstance().synchronizeFields( fieldNames,
+        FieldIdentifiers fieldsToBeSync;
+
+        fieldsToBeSync.addElementFields( { extrinsicMeshData::flow::pressure::key(),
+                                           extrinsicMeshData::flow::pressure_n::key(),
+                                           "elementAperture" },
```
TotoGaz, line 215:
```suggestion
                                           SurfaceElementSubRegion::viewKeyStruct::elementApertureString() },
```

TotoGaz (commented):
Good work.
So is it `fieldsToBeSynched` or `fieldsToSync`?
Last `FieldsIdentifiers` modifications (and `NeighborCommunicator.cpp`) decrease readability and make the entry ticket cost heavier.

rrsettgast (approved):

PR #1881, opened 2022-04-18
Implemented single-phase accumulation kernel class

francoishamon:
I resume the work to move the flow assembly kernels to the new style based on a kernel class. 

In this PR, I implemented the single-phase accumulation kernel class customized for the different types of subregions (`CellElementSubRegion` and `SurfaceElementSubRegion`). I also moved `densityOld` from the solver to the constitutive model which requires some small changes in `SinglePhaseBase` and `SinglePhaseWell`.

This PR does not introduce numerical diffs and is ready for review. The only remaining task is to figure out what to do with the unit test (now much harder to implement without bringing the xml as for multiphase tests).

Rebaseline done in https://github.com/GEOSX/integratedTests/pull/225


francoishamon:
Hello @CusiniM and @frankfeifan the PR is ready for your review. It does two things:
- Single-phase accumulation term computed in a kernel class
- Density saved in the `SinglePhaseBase` model

I was not thinking about thermal extension when I did this PR, so let me know if there is something in the PR that will be a blocker for your work.

Also, for now I disabled the unit test, but I will write a new one computing finite-difference derivatives just before merging.

src/coreComponents/constitutive/fluid/SingleFluidBase.cpp
```diff
@@ -43,23 +44,26 @@ SingleFluidBase::SingleFluidBase( string const & name, Group * const parent )
...
-    for( localIndex q = 0; q < numG; ++q )
-    {
-      initialDensity[k][q] = newDensity[k][q];
-    }
-  } );
+void SingleFluidBase::saveConvergedState() const
+{
+  arrayView2d< real64 const > density   = m_density;
+  arrayView2d< real64 >       density_n = m_density_n;
+  density_n.setValues< parallelDevicePolicy<> >( density );
```
CusiniM, line 66: `density_n.setValues< parallelDevicePolicy<> >( m_density.toViewConst() );` ?

src/coreComponents/constitutive/fluid/SingleFluidBase.hpp
```diff
@@ -178,15 +181,16 @@ class SingleFluidBase : public ConstitutiveBase
   arrayView2d< real64 const > dDensity_dPressure() const { return m_dDensity_dPressure; }
 
   arrayView2d< real64 const > initialDensity() const { return m_initialDensity; }
+  arrayView2d< real64 const > density_n() const { return m_density_n; }
 
   arrayView2d< real64 > viscosity() { return m_viscosity; }
   arrayView2d< real64 const > viscosity() const { return m_viscosity; }
 
   arrayView2d< real64 > dViscosity_dPressure() { return m_dViscosity_dPressure; }
   arrayView2d< real64 const > dViscosity_dPressure() const { return m_dViscosity_dPressure; }
```
CusiniM, line 190: Do we still need `non const` accessors?  I think these properties are now always updated inside the constitutive models. Aren't they? We could try to do this in a separate PR though.

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBaseKernels.hpp
```diff
@@ -129,128 +132,353 @@ struct MobilityKernel
...
     // Residual contribution is mass conservation in the cell
-    localAccum = poreVolNew * densNew - poreVol_n * dens_n;
+    stack.localResidual = stack.poreVolumeNew * m_density[ei][0] - stack.poreVolume_n * m_density_n[ei][0];
 
     // Derivative of residual wrt to pressure in the cell
-    localAccumJacobian = dPoreVol_dPres * densNew + dDens_dPres * poreVolNew;
+    stack.localJacobian = stack.dPoreVolume_dPres * m_density[ei][0] + m_dDensity_dPres[ei][0] * stack.poreVolumeNew;
+
+    // Customize the kernel with this lambda
+    kernelOp();
```
CusiniM, line 273: we should be able to use this to add the thermal derivatives so I would say that this design is perfect.

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBaseKernels.hpp
```diff
@@ -129,128 +132,353 @@ struct MobilityKernel
...
+   * @brief Kernel variables (dof numbers, jacobian and residual) located on the stack
+   */
+  struct StackVariables
+  {
+public:
+
+    // Pore volume information
+
+    /// Pore volume at time n+1
+    real64 poreVolumeNew = 0.0;
```
CusiniM, line 204: should we just call this `poreVolume` ?

CusiniM (approved):
This looks good and I think it's perfectly compatible with the idea of adding thermal.

src/coreComponents/constitutive/fluid/SingleFluidBase.cpp
francoishamon: Done

src/coreComponents/constitutive/fluid/SingleFluidBase.hpp
francoishamon: Yes you are right they are useless, I removed them the non `const` accessors

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBaseKernels.hpp
francoishamon: I modified the name to `poreVolume` here and in the compositional kernels

francoishamon:
The PR is ready, I will do the rebaseline as soon as it is my turn to merge

PR #1887, opened 2022-04-22
Computing cell volumes using numerical integration

castelletto1:
This PR introduces cell volume computation based on numerical integration (at the moment exploiting finite element capabilities). It replaces the current implementation based on dividing hexahedra, wedges, and pyramids into tetrahedra. It provides a more accurate evaluation of cell volumes whenever non planar quadrilateral faces are present.

New baselines (to get rid of diffs for stress fields) required for:

- `EmbFrac_Compression_CoulombFriction_01`
- `EmbFrac_Compression_CoulombFriction_04`
- `EmbFrac_Compression_Frictionless_01`
- `EmbFrac_Compression_Frictionless_04 `

Rebaseline done in https://github.com/GEOSX/integratedTests/pull/218


src/coreComponents/mesh/CMakeLists.txt
```diff
@@ -105,7 +105,7 @@ set( mesh_sources
      utilities/ComputationalGeometry.cpp
      )
 
-set( dependencyList schema dataRepository constitutive metis )
+set( dependencyList schema dataRepository constitutive finiteElement metis )
```
TotoGaz, line 108: I would better expect `finiteElement` to depend on `mesh` than the opposite?
Maybe this means that we need to extract some `computationalGeometry` cmake target?

src/coreComponents/mesh/utilities/ComputationalGeometry.hpp
```diff
@@ -446,47 +450,15 @@ void getBoundingBox( localIndex const elemIndex,
...
-
-  return 1.0/12.0 * ( LvArray::tensorOps::AiBi< 3 >( X7_X1plusX6_X0, X7_X2crossX3_X0 ) +
-                      LvArray::tensorOps::AiBi< 3 >( X6_X0, X7_X2plusX5_X0crossX7_X4 ) +
-                      LvArray::tensorOps::AiBi< 3 >( X7_X1, X5_X0crossX7_X4plusX3_X0 ) );
+  constexpr integer numQuadraturePoints = 8;
+  real64 result =  finiteElement::H1_Hexahedron_Lagrange1_GaussLegendre2::transformedQuadratureWeight( 0, X );
+  for( localIndex q=1; q<numQuadraturePoints; ++q )
+  {
+    result = result + finiteElement::H1_Hexahedron_Lagrange1_GaussLegendre2::transformedQuadratureWeight( q, X );
+  }
```
TotoGaz, line 460:
```suggestion
  real64 result =  0;
  for( localIndex q=0; q<numQuadraturePoints; ++q )
  {
    result = result + finiteElement::H1_Hexahedron_Lagrange1_GaussLegendre2::transformedQuadratureWeight( q, X );
  }
```
Won't the compiler be able to optimize the dummy `0` initialization of this loop?

src/coreComponents/mesh/utilities/ComputationalGeometry.hpp
```diff
@@ -520,24 +480,15 @@ real64 tetVolume( real64 const X[][3] )
...
-                              LVARRAY_TENSOROPS_INIT_LOCAL_3( X[5] ) };
-
-  return tetVolume( tet1 ) + tetVolume( tet2 ) + tetVolume( tet3 );
+  constexpr integer numQuadraturePoints = 6;
+  real64 result =  finiteElement::H1_Wedge_Lagrange1_Gauss6::transformedQuadratureWeight( 0, X );
+  for( localIndex q=1; q<numQuadraturePoints; ++q )
+  {
+    result = result + finiteElement::H1_Wedge_Lagrange1_Gauss6::transformedQuadratureWeight( q, X );
+  }
+  return result;
```
klevzoff, line 491: All these functions are identical now, the only difference being the type of finite element used, and parameters that are derived from it. Would you mind extracting a function, say
```c++
template< typename FE_TYPE >
GEOSX_HOST_DEVICE inline 
real64 elementVolume( real64 const (&X)[FE_TYPE::numNodes][3] )
{
  // quadrature loop
}
```
and then delegating to it?
```suggestion
  return elementVolume< H1_Wedge_Lagrange1_Gauss6 >( X );
```

src/coreComponents/finiteElement/elementFormulations/FiniteElementBase.hpp
```diff
@@ -32,7 +32,6 @@
 #include "mesh/NodeManager.hpp"
 #include "mesh/EdgeManager.hpp"
 #include "mesh/FaceManager.hpp"
-#include "mesh/CellElementSubRegion.hpp"
```
klevzoff, line 35: The three includes above it should also be removed

src/coreComponents/mesh/utilities/ComputationalGeometry.hpp
```diff
@@ -446,47 +450,15 @@ void getBoundingBox( localIndex const elemIndex,
...
-
-  real64 X7_X2plusX5_X0crossX7_X4[3];
-  LvArray::tensorOps::crossProduct( X7_X2plusX5_X0crossX7_X4, X7_X2plusX5_X0, X7_X4 );
-
-  real64 X5_X0crossX7_X4plusX3_X0[3];
-  LvArray::tensorOps::crossProduct( X5_X0crossX7_X4plusX3_X0, X5_X0, X7_X4plusX3_X0 );
-
-  return 1.0/12.0 * ( LvArray::tensorOps::AiBi< 3 >( X7_X1plusX6_X0, X7_X2crossX3_X0 ) +
-                      LvArray::tensorOps::AiBi< 3 >( X6_X0, X7_X2plusX5_X0crossX7_X4 ) +
-                      LvArray::tensorOps::AiBi< 3 >( X7_X1, X5_X0crossX7_X4plusX3_X0 ) );
```
rrsettgast, line 489: @castelletto1 You will not find a more accurate method of calculating the volume of a hex than this one. I have a Mathematica script somewhere that proves this.

src/coreComponents/mesh/CellElementSubRegion.cpp
```diff
@@ -320,6 +336,8 @@ void CellElementSubRegion::
                               m_elementType, getName() ) );
     }
   }
+
+  LvArray::tensorOps::scale< 3 >( m_elementCenter[ k ], 1.0 / m_numNodesPerElement );
```
klevzoff, line 340: As a general comment, I think the existence of this single-cell function `calculateElementCenterAndVolume` as a (non-static) member of `CellElementSubRegion` is a bit weird, and prevents us from moving this computation on device, if we wanted to. It's only ever called from inside the kernel launch of `calculateElementGeometricQuantities` but has to repeatedly switch on element type (which is constant).

Ideally I would like to:
1. Do the `switch` in `calculateElementGeometricQuantities`, outside of the kernel launch.
2. Inside each `case`, delegate to a helper free function, say `elementGeometricQuantitiesKernel`, templated on `ElemenType`, passing input views `X` and `m_toNodesRelation` along with output views `m_elementVolume` and `m_elementCenter`.
3. `elementGeometricQuantitiesKernel` does the kernel launch (maybe take launch policy as parameter too) and uses compile time knowledge of element shape to allocate `Xlocal` of the right size, unroll loops, etc. as well as delegate to the correct FE type for volume calculation.

(this might require establishing a compile-time mapping between `ElemenType` and the corresponding first-order `finiteElement` class, if one does not exist somewhere in code already).

Perhaps all of this can be addressed separately, to avoid scope creep of this PR.

src/coreComponents/mesh/utilities/ComputationalGeometry.hpp
castelletto1: I don't question that the current implementation based on the [tetrakis hexahedron](https://en.wikipedia.org/wiki/Tetrakis_hexahedron) is extremely efficient (see report [Efficient Computation of Volume of
Hexahedral Cells](https://www.osti.gov/servlets/purl/632793) by J. Grandy). However, the fundamental assumption is that hexahedra are treated using triangular boundaries. Conversely, my goal is to compute the volume of hexahedra, and also of wedges and pyramids, accounting for possible non planar (bilinear) faces. This is consistent with what is done in the finite element solvers. If a finite element based solver is coupled with a finite volume solver, and different kernels are used to assemble cell-wise contribution, it is important to ensure that cell volumes are always the same.

src/coreComponents/finiteElement/elementFormulations/FiniteElementBase.hpp
castelletto1: Aren't they required by the following abstract initialization method?

https://github.com/GEOSX/GEOSX/blob/8000c6567184c3502d22bec74069b3302a1cee29/src/coreComponents/finiteElement/elementFormulations/FiniteElementBase.hpp#L116-L136

klevzoff: You're right, I didn't spot it. Although we could in principle forward declare them - by the time the body of `fillMeshData` (where they are used) is instantiated, full definitions will be available in the calling scope. But it's not a great solution either. I just don't like that we still have an implicit circular dependency `finiteElement <-> mesh` now through the use of these mesh objects. But this probably requires some design rethinking.

src/coreComponents/mesh/CMakeLists.txt
castelletto1: The dependency on `finiteElement` for `computationalGeometry` is a temporary solution to have access to isoparametric Jacobian transformation matrix/mapping from the parent space to the physical space. Ultimately such capabilities should be provided through a numerical integration library. I am happy to volunteer for implementing it, but this should be an independent effort. @rrsettgast @klevzoff thoughts?

src/coreComponents/mesh/utilities/ComputationalGeometry.hpp
castelletto1: Done.

src/coreComponents/mesh/utilities/ComputationalGeometry.hpp
castelletto1: Yes, that's better.

castelletto1:
I'm getting this error with `gcc` (on quartz and [Travis](https://app.travis-ci.com/github/GEOSX/GEOSX/jobs/568047065)) for unit test `testMimeticInnerProducts`. I've never seen it before, any idea?

```
Building CXX object coreComponents/unitTests/finiteVolumeTests/CMakeFiles/testMimeticInnerProducts.dir/testMimeticInnerProducts.cpp.o
In function 'void makeTetra(geosx::array2d<double>&, geosx::FaceManager::NodeMapType&, geosx::array1d<int>&, geosx::real64 (&)[3], geosx::real64&, geosx::real64 (&)[3], geosx::real64&, geosx::integer, geosx::arraySlice2d<double>&)':
cc1plus: error: 'void* __builtin_memset(void*, int, long unsigned int)' specified size between 18446744065119617048 and 18446744073709551612 exceeds maximum object size 9223372036854775807 [-Werror=stringop-overflow=]
cc1plus: all warnings being treated as errors
make[3]: *** [coreComponents/unitTests/finiteVolumeTests/CMakeFiles/testMimeticInnerProducts.dir/testMimeticInnerProducts.cpp.o] Error 1
```

TotoGaz:
https://github.com/GEOSX/LvArray/pull/250 and https://github.com/GEOSX/GEOSX/pull/1859

castelletto1:
> [OSX/LvArray#250](https://github.com/GEOSX/LvArray/pull/250) and #1859

I'm not sure I understand what is going on here. The problem appears only with test `testMimeticInnerProducts.TPFA_tetra`. Commenting out that single test, everything works fine.

TotoGaz:
Do you have the patch from https://github.com/GEOSX/LvArray/pull/250/files in your code base?

castelletto1:
Yes, the `LvArray` submodule is pointing to the last commit (see [Travis](https://app.travis-ci.com/github/GEOSX/GEOSX/jobs/568190173)).

castelletto1:
After several failed attempts to disable the problematic warning using 

```c++
#pragma GCC diagnostic ignored "-Wstringop-overflow"
```

in `testMimeticInnerProducts.cpp`, I opted for @klevzoff's suggestion and modified directly `CMakeLists.txt`

https://github.com/GEOSX/GEOSX/blob/c6aa7518e02e8d0935f903d8ce137d2dfb933088/src/coreComponents/unitTests/finiteVolumeTests/CMakeLists.txt#L32-L39

This unit test is now passing also for `gcc 8.2.0` (Pecan CPU build on Travis).

src/coreComponents/mesh/CellElementSubRegion.cpp
castelletto1: I agree, let's address this in another PR.

klevzoff (approved):

PR #1890, opened 2022-04-26
Refactored well solvers before coupling with poromechanics

francoishamon:
As a preliminary step towards coupling the wells with the poromechanics solvers, I refactored the well coupled solvers following this [discussion](https://github.com/GEOSX/GEOSX/discussions/1631) about coupled solvers started by @rrsettgast. I introduced the generic coupled solver described in this [comment](https://github.com/GEOSX/GEOSX/discussions/1631#discussioncomment-1439476) by @klevzoff:
```
template< typename ... SOLVERS >
class CoupledSolver
{
protected:
  std::tuple< SOLVERS *... > m_solvers;
}
```
and wrote the well solvers as derived classes of `CoupledSolver`. 

I did not activate the coupled wells+poromechanics in this PR because I would like to first get some feedback on the new structure, but it should be easy after this PR using the `RESERVOIR_SOLVER` template (a sign inconsistency between solvers needs to be resolved first).

The PR does not introduce numerical diffs and is ready for review.


src/coreComponents/physicsSolvers/multiphysics/CoupledSolver.hpp
```diff
@@ -0,0 +1,241 @@
...
+  virtual void
+  setupDofs( DomainPartition const & domain,
+             DofManager & dofManager ) const override
+  {
+    forEachArgInTuple( std::tuple< SOLVERS *... >{}, [&]( auto t, auto idx )
+    {
+      GEOSX_UNUSED_VAR( t );
+      auto & solver = std::get< idx() >( m_solvers );
+      solver->setupDofs( domain, dofManager );
+    } );
```
CusiniM, line 75: This will set up all Dofs for all solverts but not the coupling blocks. For some solvers, like https://github.com/GEOSX/GEOSX/blob/bb16d72e8dc43a94f3f55e9b6aa6dc7de003f669/src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoromechanicsSolver.cpp#L121 the `setupDofs` call also adds coupling terms across variables of the two solvers. Would it make sense to make this function non-virtual and add another virtual function, e.g., `setupCoupling` and have solvers overload that one?

src/coreComponents/physicsSolvers/multiphysics/CoupledReservoirAndWellsBase.hpp
```diff
@@ -0,0 +1,166 @@
...
+
+namespace geosx
+{
+
+template< typename RESERVOIR_SOLVER, typename WELL_SOLVER >
+class CoupledReservoirAndWellsBase : public CoupledSolver< RESERVOIR_SOLVER, WELL_SOLVER >
+{
+public:
+
+  using AbstractBase = CoupledSolver< RESERVOIR_SOLVER, WELL_SOLVER >;
```
TotoGaz, line 33: It's not very `Abstract`, is it?

src/coreComponents/physicsSolvers/multiphysics/SinglePhaseReservoirAndWells.cpp
```diff
@@ -33,46 +32,102 @@ namespace geosx
...
 
-SinglePhaseReservoir::SinglePhaseReservoir( const string & name,
-                                            Group * const parent ):
-  ReservoirSolverBase( name, parent )
+namespace
 {
-  m_linearSolverParameters.get().mgr.strategy = LinearSolverParameters::MGR::StrategyType::singlePhaseReservoirFVM;
+
+template< typename SINGLEPHASE_RESERVOIR_SOLVER > class
+  SinglePhaseCatalogNames {};
```
TotoGaz, line 39: same

src/coreComponents/physicsSolvers/multiphysics/SinglePhaseReservoirAndWells.cpp
```diff
@@ -33,46 +32,102 @@ namespace geosx
...
+{}
+
+template< typename SINGLEPHASE_RESERVOIR_SOLVER >
+SinglePhaseReservoirAndWells< SINGLEPHASE_RESERVOIR_SOLVER >::
+~SinglePhaseReservoirAndWells()
 {}
 
-void SinglePhaseReservoir::initializePostInitialConditionsPreSubGroups()
+template< typename SINGLEPHASE_RESERVOIR_SOLVER >
+void
```
TotoGaz, line 76: I must confess I'm a bit surprise to see those templates in a `cpp` file (same for some other class `CoupledReservoirAndWellsBase.cpp` above.)

src/coreComponents/physicsSolvers/multiphysics/SinglePhaseReservoirAndWells.hpp
```diff
@@ -13,52 +13,50 @@
...
 
-#include "physicsSolvers/multiphysics/ReservoirSolverBase.hpp"
+#include "physicsSolvers/multiphysics/CoupledReservoirAndWellsBase.hpp"
+#include "physicsSolvers/fluidFlow/wells/SinglePhaseWell.hpp"
 
 namespace geosx
 {
 
-class SinglePhaseReservoir : public ReservoirSolverBase
+template< typename SINGLEPHASE_RESERVOIR_SOLVER >
```
TotoGaz, line 29: Maybe (or maybe not) add a trait to validate that `SINGLEPHASE_RESERVOIR_SOLVER` really is a single phase.

src/coreComponents/physicsSolvers/multiphysics/CoupledReservoirAndWellsBase.cpp
```diff
@@ -0,0 +1,291 @@
...
+    ElementRegionManager const & elemManager = meshLevel.getElemManager();
+
+    string const wellDofKey = dofManager.getKey( getWellSolver()->wellElementDofName() );
+    string const resDofKey = dofManager.getKey( getWellSolver()->resElementDofName() );
+
+    ElementRegionManager::ElementViewAccessor< arrayView1d< globalIndex const > > const & resElemDofNumber =
+      elemManager.constructArrayViewAccessor< globalIndex, 1 >( resDofKey );
+
+    ElementRegionManager::ElementViewAccessor< arrayView1d< integer const > > const & resElemGhostRank =
+      elemManager.constructArrayViewAccessor< integer, 1 >( ObjectManagerBase::viewKeyStruct::ghostRankString() );
```
TotoGaz, line 126: Could not you use some `Accessor`?

src/coreComponents/physicsSolvers/multiphysics/CompositionalMultiphaseReservoirAndWells.cpp
```diff
@@ -0,0 +1,404 @@
...
+      arrayView1d< localIndex const > const & perfWellElemIndex =
+        perforationData->getReference< array1d< localIndex > >( PerforationData::viewKeyStruct::wellElementIndexString() );
+
+      // get the element region, subregion, index
+      arrayView1d< localIndex const > const & resElementRegion =
+        perforationData->getReference< array1d< localIndex > >( PerforationData::viewKeyStruct::reservoirElementRegionString() );
+      arrayView1d< localIndex const > const & resElementSubRegion =
+        perforationData->getReference< array1d< localIndex > >( PerforationData::viewKeyStruct::reservoirElementSubregionString() );
+      arrayView1d< localIndex const > const & resElementIndex =
+        perforationData->getReference< array1d< localIndex > >( PerforationData::viewKeyStruct::reservoirElementIndexString() );
```
TotoGaz, line 298: Still some little `extrinsicMeshData` work!

src/coreComponents/physicsSolvers/multiphysics/CompositionalMultiphaseReservoirAndWells.cpp
```diff
@@ -0,0 +1,404 @@
...
+
+  Base::getWellSolver()->setFlowSolverName( m_reservoirSolverName );
+
+  integer const & useMassFlow = Base::getReservoirSolver()->template getReference< integer >( CompositionalMultiphaseBase::viewKeyStruct::useMassFlagString() );
+  integer const & useMassWell = Base::getWellSolver()->template getReference< integer >( CompositionalMultiphaseWell::viewKeyStruct::useMassFlagString() );
+  GEOSX_THROW_IF( useMassFlow != useMassWell,
+                  GEOSX_FMT( "CompositionalMultiphaseReservoir '{}': the input flag {} must be the same in the flow and well solvers, respectively '{}' and '{}'",
+                             this->getName(), CompositionalMultiphaseBase::viewKeyStruct::useMassFlagString(),
+                             Base::getReservoirSolver()->getName(), Base::getWellSolver()->getName() ),
+                  InputError );
```
TotoGaz, line 102: yummy 😋

TotoGaz (commented):
Nice move!

src/coreComponents/physicsSolvers/multiphysics/CoupledReservoirAndWellsBase.cpp
```diff
@@ -0,0 +1,291 @@
...
+
+  rhs.scale( -1.0 );
+  solution.zero();
+  SolverBase::solveLinearSystem( dofManager, matrix, rhs, solution );
+}
+
+template class CoupledReservoirAndWellsBase< SinglePhaseFVM< SinglePhaseBase >, SinglePhaseWell >;
+template class CoupledReservoirAndWellsBase< SinglePhaseHybridFVM, SinglePhaseWell >;
+template class CoupledReservoirAndWellsBase< CompositionalMultiphaseFVM, CompositionalMultiphaseWell >;
+template class CoupledReservoirAndWellsBase< CompositionalMultiphaseHybridFVM, CompositionalMultiphaseWell >;
```
klevzoff, line 288: Hmm, I don't particularly like neither having to explicitly instantiate a base class with specific sub-solver types (if someone forgets to add an instantiation, they get tons of linker errors), nor the fact that we're recompiling common code 4 times that doesn't even depend (for the most part) on sub-solver types. Moving everything to a header would only solve the first problem. The second one might require multiple virtual inheritance, which I'm not too keen on using. Need to think about this one. @TotoGaz thoughts?

klevzoff (commented):
Haven't reviewed everything yet, a few comments so far

src/coreComponents/physicsSolvers/multiphysics/CoupledReservoirAndWellsBase.cpp
TotoGaz: > (if someone forgets to add an instantiation, they get tons of linker errors)

Relying on linker errors to make things work is a bit surprising.
I remember we have this kind of tricks in some `Packing` functions...
I cannot find it but I still remember my emotions when I figured our what was going on... 😢 

> we're recompiling common code 4 times that doesn't even depend (for the most part) on sub-solver types.

Then maybe we can split the implementation so OK with this? (Some technical class inherited privately to reduce visibility can help?).
Could we trick into having each of these `template class...` lines in its own `.cpp` file?

> Moving everything to a header would only solve the first problem.

Is is possible to `static_assert` or `enable_if` something such that we keep the templates in the `.cpp` but error messages would be explicit?

> The second one might require multiple virtual inheritance, which I'm not too keen on using. 

This is new kind of `C++` expertise we surely do not want to put into `GEOSX`.
It's already challenging enough.

> @TotoGaz thoughts?

Never, only actions 💪 💥

klevzoff: > The second one might require multiple virtual inheritance, which I'm not too keen on using.

Actually a much simpler solution would be to put all code that does not depend on `SOLVERS` pack types into free functions that would get compiled once in the `.cpp` and called from solver methods. The free functions would take solver object as a parameter, along with domain and everything else needed. This leaves a trivial amount of code in the solver member functions, which can move into the header entirely.

francoishamon: Ok thanks, I will do that. I implemented all the other suggestions, so I will push now and work on this one. Thanks!

src/coreComponents/physicsSolvers/multiphysics/CoupledSolver.hpp
francoishamon: Yes, you are right, I implemented what you suggested

src/coreComponents/physicsSolvers/multiphysics/CoupledReservoirAndWellsBase.cpp
francoishamon: This brings us back to yesterday's discussion on Mark's PR about using accessors for the dofNumbers:
https://github.com/GEOSX/GEOSX/pull/1733#discussion_r860336851
So yes, maybe we should implement the accessors for dynamic keys (if possible?)

src/coreComponents/physicsSolvers/multiphysics/CompositionalMultiphaseReservoirAndWells.cpp
francoishamon: Yes, you are right, I added the `extrinsicMeshData` for these quantities

src/coreComponents/physicsSolvers/multiphysics/CompositionalMultiphaseReservoirAndWells.cpp
francoishamon: Now this is lighter, it looks like this:
https://github.com/GEOSX/GEOSX/blob/1dc1cf9ca37833654eefc891eaad228b6dd03dc1/src/coreComponents/physicsSolvers/multiphysics/CoupledReservoirAndWellsBase.cpp#L78-L83
for now

src/coreComponents/physicsSolvers/multiphysics/SinglePhaseReservoirAndWells.cpp
```diff
@@ -33,46 +32,102 @@ namespace geosx
...
+  Base::postProcessInput();
 
-  if( m_flowSolver->getLinearSolverParameters().mgr.strategy == LinearSolverParameters::MGR::StrategyType::singlePhaseHybridFVM )
+  SINGLEPHASE_RESERVOIR_SOLVER * reservoirSolver =
+    &this->getParent().template getGroup< SINGLEPHASE_RESERVOIR_SOLVER >( m_reservoirSolverName );
+  std::get< toUnderlying< typename Base::SolverType >( Base::SolverType::Reservoir ) >( m_solvers ) = reservoirSolver;
+
+  SinglePhaseWell * wellSolver =
+    &this->getParent().template getGroup< SinglePhaseWell >( m_wellSolverName );
+  std::get< toUnderlying< typename Base::SolverType >( Base::SolverType::Well ) >( m_solvers ) = wellSolver;
```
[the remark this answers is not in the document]
francoishamon: Thanks, I implemented the first part with the `setSubSolvers` functions in the base `CoupledSolver` class. For now it is simply used as 
https://github.com/GEOSX/GEOSX/blob/1dc1cf9ca37833654eefc891eaad228b6dd03dc1/src/coreComponents/physicsSolvers/multiphysics/CoupledReservoirAndWellsBase.cpp#L80

Maybe the first of to encapsulate the handling of the sub-solver names in the `CoupledSolver` is a little too error prone? I like the second one, I could do something like 
```
class WellSolverBase
{
  static string coupledSolverAttributePrefix() { return "well"; }
};
```
and then both the flow solvers and the poromechanics solvers would have:
```
static string coupledSolverAttributePrefix() { return "reservoir"; }
```
because my intent is to be able to use either pure flow or coupled poromechanics solvers as `RESERVOIR_SOLVER` here.

src/coreComponents/physicsSolvers/multiphysics/SinglePhaseReservoirAndWells.cpp
francoishamon: Yes, you are right, I am going to work on that now

src/coreComponents/physicsSolvers/multiphysics/CoupledReservoirAndWellsBase.cpp
```diff
@@ -0,0 +1,291 @@
...
+
+template< typename RESERVOIR_SOLVER, typename WELL_SOLVER >
+RESERVOIR_SOLVER *
+CoupledReservoirAndWellsBase< RESERVOIR_SOLVER, WELL_SOLVER >::
+getReservoirSolver() const { return std::get< toUnderlying< SolverType >( SolverType::Reservoir ) >( m_solvers ); }
+
+template< typename RESERVOIR_SOLVER, typename WELL_SOLVER >
+WELL_SOLVER *
+CoupledReservoirAndWellsBase< RESERVOIR_SOLVER, WELL_SOLVER >::
+getWellSolver() const { return std::get< toUnderlying< SolverType >( SolverType::Well ) >( m_solvers ); }
```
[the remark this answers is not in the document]
francoishamon: I removed the template parameter

src/coreComponents/physicsSolvers/multiphysics/SinglePhaseReservoirAndWells.cpp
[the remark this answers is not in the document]
francoishamon: Done, I implemented the second part of your suggestion as well, with the `SolverType::coupledSolverAttributePrefix()`. For now, I kept `static string coupledSolverAttributePrefix() { return "flow"; }` in the flow solvers to avoid a change in the xml.

src/coreComponents/physicsSolvers/multiphysics/CoupledReservoirAndWellsBase.cpp
francoishamon: Done, I removed 
```
template class CoupledReservoirAndWellsBase< SinglePhaseFVM< SinglePhaseBase >, SinglePhaseWell >;
template class CoupledReservoirAndWellsBase< SinglePhaseHybridFVM, SinglePhaseWell >;
template class CoupledReservoirAndWellsBase< CompositionalMultiphaseFVM, CompositionalMultiphaseWell >;
template class CoupledReservoirAndWellsBase< CompositionalMultiphaseHybridFVM, CompositionalMultiphaseWell >;
```
in the base class, moved the function implementations in the base class, and added free functions to move (almost all) the implementations details to the cpp. For one function (`setupSystem`), I left the body in the hpp, because the implementations depends on the derived types.

src/coreComponents/physicsSolvers/multiphysics/SinglePhaseReservoirAndWells.cpp
francoishamon: I removed the templates from the implementations in the `CoupledReservoirAndWells.cpp` file, but I left them in `SinglePhaseReservoirAndWells.cpp` and `CompositionalMultiphaseReservoirAndWells.cpp`. I left the templates in these files because the derived classes `SinglePhaseFVM.cpp` and `CO2BrineFluid.cpp` follow the same pattern (i.e., they have templates), but if you think they should be removed in `SinglePhaseReservoirAndWells.cpp` and `CompositionalMultiphaseReservoirAndWells.cpp` as well, let me know.

src/coreComponents/physicsSolvers/multiphysics/SinglePhaseReservoirAndWells.hpp
francoishamon: I have to find a good way to do that. Would it be something like adding:
```
static bool isSinglePhase() { return "true"; }
```
in `SinglePhaseBase.hpp` and adding 
```
static bool isSinglePhase() { return "false"; }
```
in `CompositionalMultiphaseBase.hpp` or do you mean something else?

francoishamon:
I addressed the comments made last week during the dev meeting and earlier:
- As suggested by @CusiniM, the base class `CoupledSolver` has a `setupCoupling` function that can be overridden by derived coupled solvers.
- As suggested by @rrsettgast, the base class `CouplerSolver` has a `assembleCouplingTerms` function that can be overridden by derived coupled solvers, since this is the pattern still used by some of them (wells, hydrofracture).
- I kept the getters `getReservoirSolver` and `getWellSolver` in the derived coupled solver because as mentioned by @klevzoff I cannot use a base class type (`CompositionalMultiphaseBase`, `SinglePhaseBase`) to access the flow solver.

I did a cleanup of the code in this PR, so if everyone likes the class `CoupledSolver`, the PR is ready

src/coreComponents/physicsSolvers/multiphysics/CompositionalMultiphaseReservoirAndWells.cpp
```diff
@@ -0,0 +1,406 @@
...
+  if( catalogName() == CompositionalCatalogNames< CompositionalMultiphaseBase >::name() )
+  {
+    if( dynamicCast< CompositionalMultiphaseFVM * >( this->getReservoirSolver() ) )
+    {
+      m_linearSolverParameters.get().mgr.strategy = LinearSolverParameters::MGR::StrategyType::compositionalMultiphaseReservoirFVM;
+    }
+    else if( dynamicCast< CompositionalMultiphaseHybridFVM * >( this->getReservoirSolver() ) )
+    {
+      m_linearSolverParameters.get().mgr.strategy = LinearSolverParameters::MGR::StrategyType::compositionalMultiphaseReservoirHybridFVM;
+    }
```
klevzoff, line 115: I wonder if instead of dynamic casting we could just switch on the child solver's MGR strategy type, i.e.
```c++
switch( this->getReservoirSolver()->getLinearSolverParameters().mgr.strategy )
{
  case LinearSolverParameters::MGR::StrategyType::compositionalMultiphaseFVM: { ...; break; }
  case LinearSolverParameters::MGR::StrategyType::compositionalMultiphaseHybridFVM { ...; break; }
}
```
This still contains a large degree of hardcoding to specific solver types, but enables us to not include `CompositionalMultiphaseFVM.hpp`, etc. at all in this file (in combination with my other suggestion about catalog names, which may or may not be implemented in this PR).

The only issue might be that for `SinglePhaseFVM` there is currently no MGR strategy, but a trivial single-level one could be added for consistency. Again, this can be addressed at a later point.

klevzoff (approved):
LGTM!

src/coreComponents/physicsSolvers/multiphysics/CompositionalMultiphaseReservoirAndWells.cpp
```diff
@@ -0,0 +1,406 @@
...
+/*
+   // Class specialization for a RESERVOIR_SOLVER set to MultiphasePoromechanics
+   template<> class CompositionalCatalogNames< MultiphasePoromechanics >
+   {
+   public:
+   static string name() { return "CompositionalMultiphasePoromechanicsReservoir"; }
+   };
+ */
+
+}
```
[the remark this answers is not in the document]
CusiniM: Yes, please, this would simplify a lot the grouping I do in the GEOSX GUI by solver type. Right now we don't have strict rules about the names of the solvers so coming up with regexp to group the solvers was pretty hard.

src/coreComponents/physicsSolvers/multiphysics/CompositionalMultiphaseReservoirAndWells.cpp
francoishamon: Yes, I agree with the suggestion, I will implement the change before merging

src/coreComponents/physicsSolvers/multiphysics/CompositionalMultiphaseReservoirAndWells.cpp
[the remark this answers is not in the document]
francoishamon: Yes, I am on board with the suggested changes (revamping the naming scheme), I propose to do the changes after discussion with a larger group possibly in a separate PR to make sure I do them right

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
```diff
@@ -849,12 +849,12 @@ void CompositionalMultiphaseWell::initializeWells( DomainPartition & domain )
         subRegion.getExtrinsicData< extrinsicMeshData::well::gravityCoefficient >();
 
       // get the element region, subregion, index
-      arrayView1d< localIndex const > const & resElementRegion =
-        perforationData.getReference< array1d< localIndex > >( PerforationData::viewKeyStruct::reservoirElementRegionString() );
-      arrayView1d< localIndex const > const & resElementSubRegion =
-        perforationData.getReference< array1d< localIndex > >( PerforationData::viewKeyStruct::reservoirElementSubregionString() );
-      arrayView1d< localIndex const > const & resElementIndex =
-        perforationData.getReference< array1d< localIndex > >( PerforationData::viewKeyStruct::reservoirElementIndexString() );
+      arrayView1d< localIndex const > const resElementRegion =
```
rrsettgast, line 852: was the use of extrinsic data structures required for this PR, or an add on?

src/coreComponents/physicsSolvers/multiphysics/CoupledSolver.hpp
```diff
@@ -0,0 +1,304 @@
...
+                           real64 const dt,
+                           DomainPartition & domain,
+                           DofManager const & dofManager,
+                           CRSMatrixView< real64, globalIndex const > const & localMatrix,
+                           arrayView1d< real64 > const & localRhs ) override
+  {
+    forEachArgInTuple( m_solvers, [&]( auto & solver, auto )
+    {
+      solver->applyBoundaryConditions( time_n, dt, domain, dofManager, localMatrix, localRhs );
+    } );
```
rrsettgast, line 252: I think the order doesn't matter....is there a scenario where it would?

rrsettgast (approved):
looks great

src/coreComponents/physicsSolvers/multiphysics/CoupledSolver.hpp
francoishamon: Maybe in corner cases (say, if a node is shared by boundary faces where we apply different boundary conditions), the order of application of the boundary conditions could accidentally matter (before and after the PR). To fix that, we could implement an xml-decided order for the application of the boundary conditions.

francoishamon:
I checked that this PR does not break the integrated tests, so I am merging it now since it is at the top of the merge queue

PR #1896, opened 2022-04-29
Bugfix/castelletto1/pamela wedge import

castelletto1:


herve-gross (approved):
Tested on 3D_10x10x10_compressible_pamela_wedge_gravity_smoke and 3D_10x10x10_compressible_pamela_pyramid_gravity_smoke with success. Thanks!

src/coreComponents/mesh/CellElementSubRegion.cpp
```diff
@@ -337,6 +337,9 @@ void CellElementSubRegion::
     }
   }
 
+  GEOSX_ERROR_IF( m_elementVolume[k] <= 0.0,
+                  GEOSX_FMT( "Negative volume for element {} type {} in subregion {}",
+                             k, m_elementType, getName() ) );
```
castelletto1, line 342: Based on @klevzoff's suggestion, I added a check on the volume sign to always stop the simulation when a negative volume is computed. This would have immediately revealed the nature of the error obtained with
- `3D_10x10x10_compressible_pamela_wedge_gravity_smoke`
- `3D_10x10x10_compressible_pamela_pyramid_gravity_smoke`

PR #1899, opened 2022-05-03
Fix misplaced initialization for boundary tags 

acitrain:
This little PR fixes a misplaced intialization for two arrays (freeSurfaceFaceIndicator and freeSurfaceNodeIndicator) used when we compute Free Surface boundary condition inside the acoustic solver.

Due to this bug, we were unable to use several free surface boundary. The arrays were re-initialized each time we switch between the differents zone defined inside the Box block inside the XML.


francoishamon (approved):

mndiaye24 (approved):

sframba (approved):

labesse40 (approved):

XL64 (approved):

francoishamon:
I checked that this PR does not break the integrated tests, so I am going to merge it since it is at the top of the merge queue

PR #1900, opened 2022-05-03
Seismic taces: bugfix and unit test

sframba:
This PR corrects a bug in the evaluation of seismic traces in the acoustic wave propagator. Currently, the interpolation is done after the variables are prepared for the next iteration, and thus one of the pressure values is not taken into account. Furthermore, some trace samples are not currently calculated due to numerical error in the evaluation of the seismogram time. This PR solves these issues, and adds a simple unit test for the seismogram evaluation feature.


sframba:
@labesse40 could you please check that this PR does not impact your python workflows?

src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEM.cpp
```diff
@@ -312,20 +314,22 @@ void AcousticWaveEquationSEM::computeSeismoTrace( real64 const time_n, real64 co
...
         for( localIndex inode = 0; inode < receiverConstants.size( 1 ); ++inode )
         {
-          ptmpNp1 += pressure_np1[receiverNodeIds[ircv][inode]] * receiverConstants[ircv][inode];
-          ptmpN += pressure_n[receiverNodeIds[ircv][inode]] * receiverConstants[ircv][inode];
+          vtmpNp1 += var_at_np1[receiverNodeIds[ircv][inode]] * receiverConstants[ircv][inode];
+          vtmpN += var_at_n[receiverNodeIds[ircv][inode]] * receiverConstants[ircv][inode];
         }
-        //Temporary linear interpolation.
-        p_rcvs[iSeismo][ircv] = a1*ptmpN + a2*ptmpNp1;
+        // linear interpolation between the pressure value at time_n and time_(n+1)
```
XL64, line 325: If I understand well you changed pressure by var as it can be another variable, maybe update the comment ?

src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEM.cpp
```diff
@@ -339,19 +343,21 @@ void AcousticWaveEquationSEM::computeSeismoTrace( real64 const time_n, real64 co
...
       }
     } );
   }
+
 }
 
 /// Use for now until we get the same functionality in TimeHistory
-void AcousticWaveEquationSEM::saveSeismo( localIndex iSeismo, real64 valPressure, string const & filename )
+/// TODO: move implementation into WaveSolverBase
+void AcousticWaveEquationSEM::saveSeismo( localIndex const iSeismo, real64 val, string const & filename )
```
XL64, line 357: val is also const, no ?

src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEM.cpp
```diff
@@ -582,15 +598,31 @@ real64 AcousticWaveEquationSEM::explicitStep( real64 const & time_n,
...
+    arrayView1d< real64 > const p_n = nodeManager.getExtrinsicData< extrinsicMeshData::Pressure_n >();
+    arrayView1d< real64 > const p_np1 = nodeManager.getExtrinsicData< extrinsicMeshData::Pressure_np1 >();
+    arrayView2d< real64 > const p_rcvs   = m_pressureNp1AtReceivers.toView();
+    for( real64 timeSeismo;
+         (timeSeismo = m_dtSeismoTrace*m_indexSeismoTrace) <= time_n && m_indexSeismoTrace < m_nsamplesSeismoTrace;
+         m_indexSeismoTrace++ )
     {
-      computeSeismoTrace( time_n, dt, m_indexSeismoTrace, p_np1, p_n );
-      m_indexSeismoTrace++;
+      computeSeismoTrace( time_n, 0, timeSeismo, m_indexSeismoTrace, p_np1, p_n, p_rcvs );
```
XL64, line 623: Can't we factorize that with the same loop in explicitStep ?

XL64 (approved):

src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEM.cpp
```diff
@@ -582,15 +598,31 @@ real64 AcousticWaveEquationSEM::explicitStep( real64 const & time_n,
...
+  } );
+  return dt;
+}
+void AcousticWaveEquationSEM::cleanup( real64 const time_n,
+                                       integer const cycleNumber,
+                                       integer const eventCounter,
+                                       real64 const eventProgress,
+                                       DomainPartition & domain )
+{
+  // compute the last seismic traces, if needed
```
jeannepellerin, line 610: If I understand well, this only depends on timings.
Would it make more sense to test before looping on the MeshBodies and getting views on data ?

src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEM.cpp
jeannepellerin: The values are copied anyway for simple types. 
I would omit them for all simple types, but that may not be the policy in Geosx.

XL64: Indeed, it's juste to be coherent between iSeismo and val.

src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEM.cpp
sframba: Exactly, the method is more general and could be later moved to WaveSolverBase, once the elastic solver is finished. I will update the comment.

src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEM.cpp
sframba: I agree, I'll add _const_ for clarity.

src/coreComponents/physicsSolvers/wavePropagation/WaveSolverBase.hpp
```diff
@@ -96,17 +96,32 @@ class WaveSolverBase : public SolverBase
...
-   * @param iseismo index number of the seismo trace
-   * @param val_np1 the array to save the value at the receiver position
+   * @brief Compute the seismic traces for a given variable at each receiver coordinate at a given time, using the pressure values at the
+   * last two timesteps.
+   * @param time_n the time corresponding to the pressure values pressure_n
+   * @param dt the simulation timestep
+   * @param timeSeismo the time at which the seismogram is computed
+   * @param iSeismo the index of the seismogram time in the seismogram array
+   * @param pressure_np1 the pressure values at time_n + dt
+   * @param pressure_n the pressure values at time_n
```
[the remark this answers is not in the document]
sframba: Sort of-- SRC=source, RCV=receiver, but I will expand the variable name (and update the method doc) to make it more clear

src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEM.cpp
sframba: you mean the for loop? if so, yes, will do

acitrain (commented):
[src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEM.hpp](url)

 127   * @brief Overridden from ExecutableGroup. Used to write las seismogram if needed.

Little typo (last)

src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEM.cpp
sframba: yes, the part of the interpolation process that depends only on time values could be done outside these loops, it would make more sense. However the performance hit is rally minimal I think

src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEM.cpp
```diff
@@ -292,17 +292,19 @@ void AcousticWaveEquationSEM::addSourceToRightHandSide( integer const & cycleNum
...
 
-
-void AcousticWaveEquationSEM::computeSeismoTrace( real64 const time_n, real64 const dt, localIndex const iSeismo, arrayView1d< real64 > const pressure_np1, arrayView1d< real64 > const pressure_n )
+void AcousticWaveEquationSEM::computeSeismoTrace( real64 const time_n,
+                                                  real64 const dt,
+                                                  real64 const timeSeismo,
+                                                  localIndex iSeismo,
+                                                  arrayView1d< real64 > const var_at_np1,
+                                                  arrayView1d< real64 > const var_at_n,
+                                                  arrayView2d< real64 > var_at_receivers )
```
francoishamon, line 301:
```suggestion
                                                  arrayView1d< real64 const > const var_np1,
                                                  arrayView1d< real64 const > const var_n,
                                                  arrayView2d< real64 > const varAtReceivers )
```
just to be consistent with the style convention used in the other parts of the code, where we use `time_n`, `density_n`, etc for the variables evaluated at time `n`. Also, whenever the variable name does not explicitly mention the time level (`_n`, etc), we try to use camelCase, hence `varAtReceivers`.

Also, I suggest to change the template from `real64` to `real64 const` for the two `arrayView1d` to clearly state that this function does not modify their values.

src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEM.cpp
```diff
@@ -582,17 +593,39 @@ real64 AcousticWaveEquationSEM::explicitStep( real64 const & time_n,
...
+{
+  // compute the remaining seismic traces, if needed
+  forMeshTargets( domain.getMeshBodies(), [&] ( string const &,
+                                                MeshLevel & mesh,
+                                                arrayView1d< string const > const & )
+  {
+    NodeManager & nodeManager = mesh.getNodeManager();
+    arrayView1d< real64 > const p_n = nodeManager.getExtrinsicData< extrinsicMeshData::Pressure_n >();
+    arrayView1d< real64 > const p_np1 = nodeManager.getExtrinsicData< extrinsicMeshData::Pressure_np1 >();
+    arrayView2d< real64 > const p_rcvs   = m_pressureNp1AtReceivers.toView();
```
francoishamon, line 610:
```suggestion
    arrayView1d< real64 const > const p_n = nodeManager.getExtrinsicData< extrinsicMeshData::Pressure_n >();
    arrayView1d< real64 const > const p_np1 = nodeManager.getExtrinsicData< extrinsicMeshData::Pressure_np1 >();
    arrayView2d< real64 > const pReceivers   = m_pressureNp1AtReceivers.toView();
```

src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEM.cpp
```diff
@@ -582,17 +593,39 @@ real64 AcousticWaveEquationSEM::explicitStep( real64 const & time_n,
...
+    computeAllSeismoTraces( time_n, 0, p_np1, p_n, p_rcvs );
+  } );
+}
+
+void AcousticWaveEquationSEM::computeAllSeismoTraces( real64 const time_n,
+                                                      real64 const dt,
+                                                      arrayView1d< real64 > const var_at_np1,
+                                                      arrayView1d< real64 > const var_at_n,
+                                                      arrayView2d< real64 > var_at_receivers )
```
francoishamon, line 619:
```suggestion
                                                      arrayView1d< real64 const > const var_np1,
                                                      arrayView1d< real64 const > const var_n,
                                                      arrayView2d< real64 > const varAtReceivers )
```

src/coreComponents/unitTests/CMakeLists.txt
```diff
@@ -10,3 +10,4 @@ add_subdirectory( finiteVolumeTests )
 add_subdirectory( fileIOTests )
 add_subdirectory( fluidFlowTests )
 add_subdirectory( wellsTests )
+add_subdirectory( physicsSolversTests )
```
francoishamon, line 13:
```suggestion
add_subdirectory( wavePropagationTests )
```
since we already have some unit tests for physics solvers split between multiple folders (`fluidFlowTests`, `wellTests`, etc)

src/coreComponents/unitTests/physicsSolversTests/testWavePropagation.cpp
```diff
@@ -0,0 +1,227 @@
...
+  for( int i=0; i<10; i++ )
+  {
+    propagator->solverStep( time_n, dt, i, domain );
+    time_n += dt;
+  }
+  // cleanup (triggers calculation of the remaining seismograms data points)
+  propagator->cleanup( 1.0, 10, 0, 0, domain );
+
+  // retrieve seismo
+  arrayView2d< real64 > const p_rcvs = propagator->getReference< array2d< real64 > >( AcousticWaveEquationSEM::viewKeyStruct::pressureNp1AtReceiversString() ).toView();
```
francoishamon, line 195:
```suggestion
  arrayView2d< real64 > const p_rcvs = propagator->getReference< array2d< real64 > >( AcousticWaveEquationSEM::viewKeyStruct::pressureNp1AtReceiversString() ).toView();
  p_rcvs.move( LvArray::MemorySpace::host, false );
```
to make sure that on the GPU platforms, the latest values of `p_rcvs` are available on the CPU when we do things like:
```
ASSERT_TRUE( std::abs( p_rcvs[i][8] ) > 0 );
```
I may be wrong, but I think this manual move is needed because `p_rcvs` is moved back to the CPU only if we go inside [this loop](https://github.com/GEOSX/GEOSX/blob/7664d3659f488d335fb37d94965623d881b90a9a/src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEM.cpp#L335), which may not always be the case as it depends on the value of `m_nsamplesSeismoTrace` and `iSeismo`?

francoishamon (approved):
Looks great, thanks for this bugfix and for adding a unit test @sframba

src/coreComponents/unitTests/physicsSolversTests/testWavePropagation.cpp
labesse40: I think we could consider moving the data from GPU to CPU only at the last iteration since we don't need it before. 
Also in our case using pygeosx, if we want to do some visualization during the simulation, we could imagine having a ".move()" if the variable lives on GPU, inside the routine of "get_wrapper('path_to_variable').value()" on C++ side. This could be useful not only for our variable but for any variable living on GPU.

labesse40 (approved):

src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEM.cpp
sframba: style changed with the last commits, I also added the const keywords as required (you are right, it is good practice)

src/coreComponents/unitTests/CMakeLists.txt
sframba: Done

src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEM.cpp
sframba: Done

src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEM.cpp
sframba: Done

labesse40:
> @labesse40 could you please check that this PR does not impact your python workflows?



> @labesse40 could you please check that this PR does not impact your python workflows?
It seems to be working
Screenshot from 2022-05-20 15-31-33
Screenshot from 2022-05-20 15-31-48

PR #1901, opened 2022-05-04
Adding pressure jump stabilization to coupled poromechanics

ryar9534:
This PR implements pressure jump stabilization for finite volume discretizations of mass equations in coupled multiphase poromechanics problems. This includes:

- Adding a stabilization weight term to the stencil to account for the geometric dependence of stabilization terms
- Registering some solid solid properties to the mesh for use in defining the stabilization coefficient
- Added code to group elements into macroelements for use in local stabilization
- Creating a new FVM kernel which adds the stabilizing flux to the mass conservation equations either to all faces (global stabilization) or to faces interior to a macroelement (local stabilization)

This PR is ready for review. 

Rebaseline done in https://github.com/GEOSX/integratedTests/pull/272


inputFiles/poromechanics/StabilizedPoroStaircase_initTemp.txt
```diff
@@ -0,0 +1,2 @@
+300
```
joshua-white, line 1: Are these table files unique, or can we point to existing files for an unstabilized version of the PoroStaircase problem?  Just want to avoid lots of near duplicate material.

src/coreComponents/constitutive/solid/SolidExtrinsicData.hpp
```diff
@@ -0,0 +1,59 @@
...
+namespace geosx
+{
+
+namespace extrinsicMeshData
+{
+
+namespace solid
+{
+
+EXTRINSIC_MESH_DATA_TRAIT( bulkModulus,
```
joshua-white, line 33: This is a subtle issue, but this approach implicitly assumes we have an elastic model with a constant bulk modulus and shear modulus as fixed parameters.  Many of our material models don't directly use these parameters (e.g. anisotropic elasticity,  pressure-dependent elasticity, and modified Cam-Clay) and their value actually depends on the loading state (e.g. for a plastic model the bulk modulus may soften with time from an initial value).  I would instead suggest adding a method in the material models themselves to return the current value, e.g. `solidModel.getBulkModulus(k,q)` which we can use in each cell to construct an appropriate scaling constant (probably lagged in time to avoid nonlinearities).  The thing to register on the mesh is probably the scaling constant itself, which can be updated at the end of each timestep in a loop. I recognize this may be overkill, but the current implementation may not work for a large class of our material models.  Perhaps we can discuss this more in person.

inputFiles/poromechanics/StabilizedPoroStaircase.xml
```diff
@@ -0,0 +1,299 @@
+<?xml version="1.0" ?>
```
TotoGaz, line 1: Please use `xml` extension and not `txt`.

src/coreComponents/physicsSolvers/fluidFlow/StabilizedCompositionalMultiphaseFVMKernels.hpp
```diff
@@ -0,0 +1,390 @@
...
+  /**
+   * @brief The type for element-based data. Consists entirely of ArrayView's.
+   *
+   * Can be converted from ElementRegionManager::ElementViewConstAccessor
+   * by calling .toView() or .toViewConst() on an accessor instance
+   */
+  template< typename VIEWTYPE >
+  using ElementViewConst = ElementRegionManager::ElementViewConst< VIEWTYPE >;
+
+  using AbstractBase = isothermalCompositionalMultiphaseFVMKernels::FaceBasedAssemblyKernelBase;
```
TotoGaz, line 60: Is this class `Abstract`?

TotoGaz (commented):
Are you basically enforcing the usage of the technique for everyone?
I only glanced at it, but can we still, _e.g._ use the `TwoPointFluxApproximation` without this patch?

src/coreComponents/physicsSolvers/fluidFlow/StabilizedCompositionalMultiphaseFVMKernels.hpp
francoishamon: Yes, the class is really abstract (if having a pure virtual function means abstract)

inputFiles/poromechanics/StabilizedPoroStaircase_initTemp.txt
francoishamon: Since the PR was open, a similar integrated test has been merged into `develop`. So, Ryan's PR is now lighter and only adds stabilization to an existing integrated test. These tables have been removed from the PR.

inputFiles/poromechanics/StabilizedPoroStaircase.xml
francoishamon: We removed this xml from the PR because this integrated test was merged some time ago in another PR.

src/coreComponents/physicsSolvers/fluidFlow/StabilizedCompositionalMultiphaseFVMKernels.hpp
francoishamon: Yes, the class `FaceBasedAssemblyKernelBase` is really abstract (if having a pure virtual function makes it abstract).

src/coreComponents/constitutive/solid/SolidExtrinsicData.hpp
ryar9534: This was a great point! I have updated things to reflect this suggested structure.

francoishamon (approved):
This PR looks good, I think it is ready to be merged after the integrated tests have been updated

joshua-white (approved):
Looks great.  One minor note is that the elastic models are a little unique in actually storing bulkModulus and shearModulus arrays that can be directly returned.  Other models can return a locally calculated value at the quadrature point, but we would need to allocate an array somewhere to store these results.  We can worry about adding other solid models in a later PR, though, once the stabilization strategies stabilizes (pun intended).

PR #1905, opened 2022-05-05
Add a pygeosx unit test

sframba:
This PR adds the first unit test for the pygeosx python interface. This test runs a very simple wave propagation problem, handling the time loop in python. The test command directly executes the pygeosx python venv, and thus tests the python/C++ wrappers as well as the python script interface. Tested methods include problem initialization, explicit timestepping and seismogram trace retrieval. The test is only executed if ENABLE_PYGEOSX=true during cmake configuration.


paloma-martinez (approved):

XL64:
I am currently adding a test in my backward PR which is in python, maybe we can discuss (I use pytest, i didn't create the subdirectory which is a good idea, I have to create a wrapper to source pygeosx activate file before running my python script)

src/pygeosx/unitTests/testWaveEquationAdjoint.py
```diff
@@ -0,0 +1,279 @@
...
+       print("===============================================================")
+       print(" Running pygeosx with solver:",solver)
+       print(" Source coordinatesi : ",srcPos.to_numpy())
+       print(" Receiver coordinates: ",recvPos.to_numpy())
+       print(" simulation duration : ",maxTime)
+       print(" Time Step           : ",dt)        
+       if vel_info:
+           print(" Min Vel             : ",minVal)
+           print(" Max Vel             : ",maxVal)
+       print("===============================================================")
```
cssherman, line 217: @TotoGaz and I have been trying to systematically replace any print calls with python logging calls.  This can greatly improve our ability to debug problems/outputs

src/pygeosx/unitTests/testPygeosxAcoustic.py
```diff
@@ -0,0 +1,32 @@
...
+
+comm = MPI.COMM_WORLD
+rank = comm.Get_rank()
+
+def test_forwardPropagationAcoustic():
+    argv = ["testPygeosxAcoustic.py", "-i", "acous_5x5x5.xml"]
+    solver = "acousticSolver"
+    forwardPropagation(argv, solver)
+
+def forwardPropagation(argv, solver):
```
cssherman, line 13: Could you add docstrings to any functions you are adding to python files?  The format is up to you, but I personally prefer the google style: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

cssherman:
As a general note, this feels more like a pygeosx integrated test than a unit test.

We definitely need to add proper unit tests as well, which should test each of the components in isolation (to the highest degree possible).  For these, I would recommend using the python unit test framework, similar to what we use here: https://github.com/GEOSX/GEOSX/blob/develop/src/coreComponents/python/modules/geosx_xml_tools_package/geosx_xml_tools/tests/test_manager.py

src/pygeosx/unitTests/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+    string(REGEX MATCHALL "^ *def *test_([A-Za-z_0-9])+ *" found ${line})
+    if (found)
+      string(REGEX REPLACE "^ *def *test_(([A-Za-z_0-9])+) *" "\\1" name "${found}")
+      configure_file(${CMAKE_CURRENT_SOURCE_DIR}/pytest_template.sh.in
+        ${CMAKE_CURRENT_BINARY_DIR}/${name}.sh
+        FILE_PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE GROUP_READ GROUP_EXECUTE WORLD_READ WORLD_EXECUTE)
+      blt_add_test(NAME ${name} COMMAND ${CMAKE_CURRENT_BINARY_DIR}/${name}.sh)
+    endif()
+  endforeach()
+endmacro(parse_test_file)
```
TotoGaz, line 45: IMO this is the key question since
- This does the test management/discovery by hand. But that's what `pytest` and `geosxats` already do.
- On top of this, we let `ctest` manage the python tests one by one and tells `pytest` to do it. And this does not prevent us from relying on the `pytest` dependency, since our tests explicitly use it (and it's nice).

Doing this, I'd be surprised that we benefit from an integrated test management of `pytest`.

My personnal long term goal would clearly to let `pytest` do all the discovery, reporting, ...
And then extract from `geosxats` the specific HPC parts (submissions (`python dask`?), comparisons, rebaselines). Or maybe make a `geosxats` module for `pytest` like many exist!
There is a question of tests in parallel.
That would mean less code on our side and we focus on more scientific code.

TotoGaz: In a nutshell, here are tracks you may want to explore.
- Do not parse `python` using cmake.
- Try to use the test discovery feature that `pytest` provides out of the box.
- If you think it makes sense and if the implementation makes sense, you can add a single `run python test` target in `ctest`.
- Be sure to properly split _integrated tests_ and _unit tests_. _Integrated tests_ should be developed using the `geosxats` fwk, like the others integrated tests.

@cssherman did I forget something?

cssherman: I'm going to work on adding a couple of simple unit tests using the method described by @TotoGaz that we can build on.  Once this is in place, we will try to get pygeosx working with the geosxats integrated testing system

rrsettgast:
@sframba Is this deprecated? If so, please close

sframba:
I close the PR, the codebase has changed too much in the meantime

PR #1910, opened 2022-05-09
A stable input file for end-users?

TotoGaz:
Hello there, this PR is not for merge but serves as a POC for discussion https://github.com/GEOSX/GEOSX/discussions/1911 where we should speak about it (this is just in case someone wants to have a look).

Some points I’d like to mention:

- My code is terrible; some elements are hard coded for the sake of the poc. It's only there for the poc.
- Error management is not done.
- The CLI is a hack.
- The `cmake`/`TPL` is wrong.
- The way I use `yaml-cpp` is weak (no `struct` deserialization; some clever stuff is surely possible (see swagger)). But there also is a good reason for this. Defining a proper input file is complicated and will require a lot of workshops and post-its.

### DIFF
--- a/inputFiles/simplePDE/Laplace_base.xml
+++ b/inputFiles/simplePDE/Laplace_base.xml
@@ -29,7 +29,7 @@
     <PeriodicEvent
       name="outputs"
       timeFrequency="1.0"
-      target="/Outputs/siloOutput"/>
+      target="/Outputs/vtkOutput"/>
 
     <PeriodicEvent
       name="restarts"
@@ -71,6 +71,8 @@
     <Silo
       name="siloOutput"
       parallelThreads="32"/>
+    <VTK
+      name="vtkOutput"/>
 
     <Restart
       name="restartOutput"/>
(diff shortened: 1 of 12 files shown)


cssherman:
If this is a path that we want go down, we should figure out how to adapt the xml schema for yaml in a scalable fashion.  There's probably some tools out there already....

TotoGaz:
Hello @cssherman and thx for the comment

> we should figure out how to adapt the xml schema for yaml in a scalable fashion

What's in the `yaml` is (in theory) quite different form what's in the `xml`.
- In the `yaml` you should find _what_ is needed. The pure business need from the "client"; like an API somehow (with limitations, of course).
- In the `xml`, we find a lot of _how_ (with some limitations too). E.g. `events`, `targets`...

So I would not be so surprised the discrepancy between `yaml` and `xml` to be quite significant.

Nevertheless, you may be interested in tools like swagger which links `yaml` to some "`struct`/model".
You may have a look at https://github.com/jbeder/yaml-cpp/wiki/Tutorial#converting-tofrom-native-data-types and the `Vec3` example in the very bottom.
This brings some nice abstraction and can be very convenient to develop nice API and nice code too.

TotoGaz:
Please use https://github.com/GEOSX/GEOSX/discussions/1911 for discussions 😉

Issue #1844, opened 2022-03-24
Better user control of turning on and off plotting of fields.


rrsettgast:
**What is the requested feature?**
We need to have user control of what fields are output to the plot files.  Specifically, we need:

1. Ability to turn off all field output. This means setting all `WrapperBase::m_plotLevel` to `NOPLOT`.
2. Add ability to change the `WrapperBase::m_plotLevel` on an individual basis. This should be applied after 1.

@cssherman Any additional suggestion?

PR #1912, opened 2022-05-09
Implemented VTK/Silo field selection from input file

francoishamon:
For large cases, our default output VTK/Silo strategy produces big output files that are no easy to manipulate/share. This PR implements a way to select in the XML the fields that we want to include in the VTK/Silo file. The `Outputs` block looks like that:
```
  <Outputs>                                                                                                                                                                              
    <VTK                                                                                                                                                                                 
      name="vtkOutputPoromecaHyst"                                                                                                                                                       
      fieldNames="{ TotalDisplacement, pressure, phaseVolumeFraction, rockReservoir_stress, rockBurdens_stress }"/>                                                                      
  </Outputs>   
```

Rebaseline done in https://github.com/GEOSX/integratedTests/pull/235


src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.hpp
```diff
@@ -117,6 +117,16 @@ class VTKPolyDataWriterInterface
     m_pvd.setFileName( joinPath( m_outputDir, m_outputName ) + ".pvd" );
   }
 
+  /**
+   * @brief Set the names of the fields to output
+   * @param[in] fieldNames the fields to output
+   */
+  void setFieldNames( array1d< string > const & fieldNames )
```
TotoGaz, line 124:
```suggestion
  void setFieldNames( array1d< string const > const & fieldNames )
```

src/coreComponents/fileIO/silo/SiloFile.hpp
```diff
@@ -662,8 +662,26 @@ class SiloFile
...
+    m_fieldNames.insert( fieldNames.begin(), fieldNames.end() );
+  }
+
 private:
 
+  /**
+   * @brief Check if plotting is enabled for this field
+   * @param[in] wrapperPlotLevel the plot level of the wrapper
+   * @param[in] wrapperName the name of the wrapper
+   */
```
TotoGaz, line 680: `@return`

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp
```diff
@@ -912,5 +913,21 @@ void VTKPolyDataWriterInterface::write( real64 const time,
...
+  // check if the logLevel is sufficient high for plotting
+  bool plotEnabled = ( wrapperPlotLevel <= m_plotLevel );
+
+  // override the logLevel if the fieldNames list was provided
+  if( !m_fieldNames.empty() )
+  {
+    auto search = m_fieldNames.find( wrapperName );
+    plotEnabled = ( search != m_fieldNames.end() );
+  }
+  return plotEnabled;
```
TotoGaz, line 928: It looks like I've seen this code before 🤣

src/coreComponents/fileIO/silo/SiloFile.cpp
```diff
@@ -3167,5 +3167,21 @@ int SiloFile::getMeshType( string const & meshName ) const
   return meshType;
 }
 
+bool SiloFile::isFieldPlotEnabled( PlotLevel const wrapperPlotLevel,
```
TotoGaz, line 3170: You could add a
```c++
bool SiloFile::isFieldPlotEnabled( string WrapperBase const & wrapper ) const
{
    return this->isFieldPlotEnabled( wrapper.getPlotLevel(), wrapper.getName() );
}
```
which can bring some smoother reading.

inputFiles/poromechanics/PoroElastic_deadoil_3ph_baker_2d.xml
```diff
@@ -279,10 +279,12 @@
 
   <Outputs>
     <Silo
-      name="siloOutput"/>
+      name="siloOutput"
+      fieldNames="{ TotalDisplacement, pressure, skeleton_stress }"/>
```
TotoGaz, line 283: If possible, I would definitely implement a `from` `to` (some kind of a mapping).
That could be very useful for our users.

cssherman:
Thanks for working on this!  I had been tinkering with the same idea, but ran into some issues with my approach (temporarily toggling plot level flags before handing it off to the individual readers).

What are your thoughts on allowing the user to enter a path-like string as well, so that we can discriminate between similarly named fields in different regions?  Or is the current region filter enough sufficient?

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.hpp
francoishamon: Done

src/coreComponents/fileIO/silo/SiloFile.hpp
francoishamon: Done, I fixed the Doxygen here

src/coreComponents/fileIO/silo/SiloFile.cpp
francoishamon: Done, I implemented what you suggested

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp
francoishamon: Yes, I agree, this code is duplicated. It is not easy to do something better because they live in different folders (`vtk` and `silo`), but I will find something

inputFiles/poromechanics/PoroElastic_deadoil_3ph_baker_2d.xml
francoishamon: @TotoGaz what kind of mapping do you suggest? Is there a case in which we want the names that appear in the vtk files to be different from the `fieldNames` in GEOSX?

francoishamon:
> Thanks for working on this! I had been tinkering with the same idea, but ran into some issues with my approach (temporarily toggling plot level flags before handing it off to the individual readers).
> 
> What are your thoughts on allowing the user to enter a path-like string as well, so that we can discriminate between similarly named fields in different regions? Or is the current region filter enough sufficient?

Yes, thanks for the suggestion, I agree that this could be an additional refinement. For my own usage, the current filter is sufficient because:
- For the material fields, I can use the name of the model as an additional filter if I don't want to output the quantity everywhere.
- For the regular fields, I typically need them on all the target regions of the solvers, so not having a region-filter is good enough.

We can discuss that at the next developer meeting.

inputFiles/poromechanics/PoroElastic_deadoil_3ph_baker_2d.xml
TotoGaz: Oh, it's the output, not the input, stupid me 😵‍💫 
Forget what I said.

CusiniM (approved):
I ll just throw the idea in there. Feel free to ignore it. 

Maybe, we should have a step to check whether the fields specified actually exist within GEOSX. It should be fairly easy to check if there is, at least, an `extrinsicMeshData` with that name. I guess it's a lot harder to know whether any of the solver specified has actually registered the required field. I am only saying this coz we kind of require the user to know that the variable pressure inside GEOSX is called `pressure` and not `Pressure` or `PRESSURE`.

francoishamon:
> I ll just throw the idea in there. Feel free to ignore it.
> 
> Maybe, we should have a step to check whether the fields specified actually exist within GEOSX. It should be fairly easy to check if there is, at least, an `extrinsicMeshData` with that name. I guess it's a lot harder to know whether any of the solver specified has actually registered the required field. I am only saying this coz we kind of require the user to know that the variable pressure inside GEOSX is called `pressure` and not `Pressure` or `PRESSURE`. Running a long test case and finding out at the end that the name of a field was wrong so the field one wanted to plot is wrong can be a bit frustrating.

Yes, good idea, I will give it a try, thanks!

jhuang2601 (approved):

cssherman:
This addresses #1844

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp
francoishamon: Done, this piece of code was indeed duplicated, but I removed the duplication in the latest commit

francoishamon:
Following the comment from @CusiniM I added a check to make sure that the `fieldNames` specified by the user are indeed registered either on the `nodeManager` or on an element subregion. The check takes place the first time the vtk ouput is called, and issues a warning if a field names is not registered.

Following a comment from @rrsettgast during a dev meeting I added a flag that can be used to specify whether we only plot the `fieldNames` listed by the user in the xml, or the `fieldNames` *and* all the fields with the required plot level.

PR #1913, opened 2022-05-11
Using lists in FieldSpecification `objectPath`

klevzoff:
This PR adds an ability to use a list of names instead of a single name at each level of `objectPath` in a `FieldSpecification`. Let me know if you think this is a good/awful idea.

The syntax is as follows:
```
objectPath="someGroup/{a,b,c}/someSubgroup/{x,y}"
```
It also updates the schema regex to accept the new syntax, essentially by allowing `string` inputs to be unconstrained (but not `string_array`, `path` or `path_array`).

A typical example would be setting a value on multiple (but not all) regions of the mesh. This avoids duplication of almost identical `FieldSpecification` blocks for multiple regions.
```xml
<FieldSpecification
  name="BurdenPermeability"
  initialCondition="1"
  objectPath="ElementRegions/{ Overburden, Underburden }"
  setNames="{ all }"
  scale="1e-20"/>
```

In addition, this PR fixes an inconsistency by changing the group name `FaceManager` -> `faceManager`. None of the XMLs need to be updated, because the key under which the group was registered was already camelCase, but the group's name itself was PascalCase. This does require a rebaseline of all integrated tests though.

Resolves https://github.com/GEOSX/GEOSX/issues/1771


rrsettgast:
This is great. I had proposed to have a general `FieldPath` object, but this is a worthy addition to that concept. What do you think about adding a `FieldPath` object that is used across the code, and encapsulates functionality to manage the iteration over that path/paths?

src/coreComponents/fieldSpecification/FieldSpecificationManager.hpp
```diff
@@ -233,80 +222,97 @@ class FieldSpecificationManager : public dataRepository::Group
...
+        array1d< string > subGroupNames;
+        LvArray::input::stringToArray( subGroupNames, targetPath[pathLevel] );
+        for( string const & subGroupName : subGroupNames )
+        {
+          applyToPathRecursive( &targetGroup->getGroup( subGroupName ),
+                                targetPath.subspan( pathLevel + 1 ),
+                                fs, std::forward< LAMBDA >( lambda ) );
+        }
+        return;
+      }
```
TotoGaz, line 282: You could surely do preprocessing and consider `targetPath` as a `Span< vector< string > >`, even with a lot of arrays of length 1.
That would allow you to have one single path for your algo (and keep string parsing out of this loop)?
Plus, (it's more a matter of style and what's done in the skip is unclear to me), but I would extract the `skip` part as a lambda function outside of the `pathLevel` loop to get something really concise.

src/coreComponents/fieldSpecification/FieldSpecificationManager.hpp
```diff
@@ -233,80 +222,97 @@ class FieldSpecificationManager : public dataRepository::Group
...
-          || target.getName() == "edgeManager" ) // TODO these 3 strings are harcoded because for the moment, there are
-                                                 // inconsistencies with the name of the Managers...
-        && target.getName() != ObjectManagerBase::groupKeyStruct::setsString()
-        && target.getName() != ObjectManagerBase::groupKeyStruct::neighborDataString() )
+    // Set-based application makes sense only for classes derived from ObjectManagerBase...
+    // except for ElementRegionManager and ElementRegionBase, which are intermediate layers.
+    ObjectManagerBase const * const manager = dynamicCast< ObjectManagerBase * >( &target );
+    if( manager != nullptr
+        && target.getName() != MeshLevel::groupStructKeys::elemManagerString
+        && target.getParent().getName() != ElementRegionManager::groupKeyStruct::elementRegionsGroup() )
```
TotoGaz, line 299: Maybe
```suggestion
    if( auto manager = dynamicCast< ObjectManagerBase const * >( &target )
        && target.getName() != MeshLevel::groupStructKeys::elemManagerString
        && target.getParent().getName() != ElementRegionManager::groupKeyStruct::elementRegionsGroup() )
```
So `null` ptr `manager` cannot be accessed in other scope in future devs.

src/coreComponents/common/Span.hpp
```diff
@@ -219,9 +219,14 @@ class Span
    * @brief @return a new span of @p count elements starting at @p offset
    * @param offset starting index
    * @param count number of elements
+   * @note if @p count == -1, the result contains all remaining elements
```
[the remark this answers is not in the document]
klevzoff: Yes, this is equivalent to `std::span`'s behavior, I just missed it before.

src/coreComponents/fieldSpecification/FieldSpecificationManager.hpp
```diff
@@ -233,80 +222,97 @@ class FieldSpecificationManager : public dataRepository::Group
...
+  {
+    constexpr std::array< char const *, 2 > groupsToSkip =
+    {
+      ElementRegionManager::groupKeyStruct::elementRegionsGroup(),
+      ElementRegionBase::viewKeyStruct::elementSubRegions()
+    };
+
+    for( std::size_t pathLevel = 0; pathLevel < targetPath.size(); ++pathLevel )
+    {
+      // Skip certain groups in the tree by jumping directly to them
```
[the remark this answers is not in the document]
klevzoff: Yeah. I think the new version is more or less equivalent to what the previous code did, but it took me a while to understand the intent there, and it's not generic skipping behavior for any list of groups in any order. Basically, it mostly just covers the `A/B` case, or `A/C/B` case where we *don't* want to skip `C`, and we also *know* that `A` always comes before `B` in a tree.

Perhaps the new code structure, comments and variable naming are a bit misleading. I just wanted to move two similar branches into a loop, but it's not much (any?) more clear than before. I would probably go with @rrsettgast's proposal to move all this logic into a separate class that handles arbitrary tree navigation (maybe borrowing some ideas from XPath syntax?).

src/coreComponents/fieldSpecification/FieldSpecificationManager.hpp
klevzoff: I've tried it and it's a compiler error, because `&&` takes priority over assignment, and then it tries to implicitly convert pointer to `bool` or something. Basically, we can't do it until C++17 where init-statement can be separated by `;` from the branch condition.

TotoGaz: Even with the parenthesis that solve the world? 🤣

src/coreComponents/fieldSpecification/FieldSpecificationManager.hpp
klevzoff: This would be a lot of extra allocations happening though - whereas strings will mostly be SBO-optimized, vectors are not :( Not saying we should necessarily care too much (even though we pass through this code multiple times on every BC application) but would be great if this parsing only ever happened once. Ideally I'd like to store pre-parsed path as a
```c++
vector< variant< string, array1d< string > > > m_path;
```
inside the `FieldSpecification` object.

src/coreComponents/fieldSpecification/FieldSpecificationManager.hpp
klevzoff: > Even with the parenthesis that solve the world?

Sadly, that would not be [valid syntax](https://en.cppreference.com/w/cpp/language/if) :(
```
condition - one of
 * expression which is contextually convertible to bool
 * declaration of a single non-array variable with a brace-or-equals initializer
```

klevzoff:
> This is great. I had proposed to have a general `FieldPath` object, but this is a worthy addition to that concept. What do you think about adding a `FieldPath` object that is used across the code, and encapsulates functionality to manage the iteration over that path/paths?

Yeah, that sounds like a great idea. It would probably be more like `GroupPath`, since we want to end up with a `Group` at the end. But yeah, having a single mechanism for handling data repository navigation would be awesome.

src/coreComponents/fieldSpecification/FieldSpecificationManager.hpp
TotoGaz: Yes there may be a _tiny_ overhead (how much?), but also we would separate the parsing of the string from the algorithm which operates on some kind of tree; having separate responsibilities is always good. 

There is no strong reason why we should deal with "_different types of nodes_" in the tree and have a big `if` in the middle of the loop.

klevzoff:
Closed in favor of https://github.com/GEOSX/GEOSX/pull/1963

PR #1921, opened 2022-05-18
Fixed single-phase well MGR recipes

francoishamon:
This PR fixes the single-phase MGR recipes that were using a direct solver to solve the coarse grid. Since the function `HYPRE_MGRDirectSolverCreate` is now disabled in hypre, I implemented another reduction approach in which wells are eliminated first:

For FVM
1- Eliminate the well block
2- Solve the cell-centered pressure system with BoomerAMG

For hybridFVM:
1- Eliminate the well block
2- Eliminate the cell-centered pressures
3- Solve the face-centered pressure system with BoomerAMG

This is also more consistent with the multiphase MGR recipes.


castelletto1 (approved):
Thank you for fixing this!

PR #1922, opened 2022-05-18
Use range based for loops instead of callback functions.

TotoGaz:
For discussion.

`DomainPartition::forMeshBodies` and `MeshBody::forMeshLevels` higher order functions are changed to member functions returning a `std::vector` of pointers to `MeshBody` and `MeshLevel` instances respectively.
Then range based for loops are used instead of the callback/lambda mechanism.

The main benefit is this should allow mocking by better decoupling the data provider from the action on these data.

Note that I had to "_duplicate_" those member functions because sometimes in the code we had to keep returning a `Group` and not a `std::vector`. Nothing that can't be fixed, but this would require more time; it's often just a matter of registration.

I've decided to return `std::vector` of pointers because creating `std::vector` of references to `abstract` classes can be a problem. To be discussed.
We may also return modern feature `ranges` but that would lead to other issues (c++20, complicated types to return/non standardized type erasure, compilation complicated across the compilers, differences between the `std` implementation and https://github.com/ericniebler/range-v3...). Or `std::generator` with c++23. I believe it's OK to adopt a less optimal solution since we'll be able to change it at minimal cost. To be discussed too.

So we go from
```c++
meshBodies.forSubGroups< MeshBody >( [&]( MeshBody & meshBody )
{
  meshBody.forMeshLevels( [&]( MeshLevel & mesh )
  {
```
to
```c++
for( MeshBody * meshBody: domain.getMeshBodies() )
{
  for( MeshLevel * meshLevel: meshBody->getMeshLevels() )
  {
```
And I've implemented a small dummy pipeline (`ranges`-style, see `MeshBody.cpp`) such that we can flat map a unique loop like
```c++
for( MeshLevel * mesh: domain.getMeshBodies() | flattenMeshLevels() )
{
```
Implementations are raw, we can surely do better.
The key refactoring, if we go that way, may be about `forMeshTargets`.


src/coreComponents/mesh/MeshBody.cpp
```diff
@@ -46,5 +46,24 @@ string MeshBody::intToMeshLevelString( localIndex const meshLevel )
...
+    }
+  }
+  return result;
+}
+
+std::function< std::vector< MeshLevel * >( MeshBody * ) > flattenMeshLevels()
+{
+  return []( MeshBody * mb )
+  { return mb->getMeshLevels(); };
+}
```
TotoGaz, line 67: This is a bit dummy impl.

src/coreComponents/mesh/MeshBody.hpp
```diff
@@ -54,11 +54,23 @@ class MeshBody : public dataRepository::Group
...
+    return result;
+  }
+
+  std::vector< MeshLevel * > getMeshLevels()
+  {
+    std::vector< MeshLevel * > result;
+    auto filler = [&]( MeshLevel & ml){ result.emplace_back( &ml ); };
+    m_meshLevels.forSubGroups< MeshLevel >( filler );
+    return result;
+  }
```
TotoGaz, line 73: This is a bit dummy impl too.

src/coreComponents/physicsSolvers/SolverBase.hpp
```diff
@@ -666,6 +666,46 @@ class SolverBase : public ExecutableGroup
...
+      for( const MeshBody * meshBody: meshBodies )
+      {
+        if( meshBodyName != meshBody->getName() )
+        { continue; }
+
+        for( MeshLevel const * meshLevel: meshBody->getMeshLevels() )
+        {
+          lambda( meshBodyName, *meshLevel, regionNames );
+        }
+      }
```
TotoGaz, line 685: This impl is not very good, but if we go that way, other accessors would be done.
Maybe we can keep it also since we'll never have huge number of `MeshBody`.

src/coreComponents/finiteVolume/FluxApproximationBase.cpp
```diff
@@ -63,22 +63,17 @@ void FluxApproximationBase::initializePreSubGroups()
 {
   DomainPartition & domain = this->getGroupByPath< DomainPartition >( "/Problem/domain" );
 
-  Group & meshBodies = domain.getMeshBodies();
-
-  meshBodies.forSubGroups< MeshBody >( [&]( MeshBody & meshBody )
+  for( MeshLevel * mesh: domain.getMeshBodies() | flattenMeshLevels() )
```
rrsettgast, line 66: is there a drawback to the nested looping? I don't know that I understand the motivation to flatten?

src/coreComponents/mesh/MeshBody.cpp
rrsettgast: same question as earlier about the desire to flatten `MeshBody` and `MeshLevel` loops. If they aren't callbacks, I don't understand the motivation.

src/coreComponents/mesh/MeshBody.hpp
```diff
@@ -54,11 +54,23 @@ class MeshBody : public dataRepository::Group
...
    * @return reference to the meshLevels group.
    */
-  Group & getMeshLevels() { return m_meshLevels; }
-  /**
-   * @copydoc getMeshLevels()
-   */
-  Group const & getMeshLevels() const { return m_meshLevels; }
+  Group & getMeshLevelsGrp() { return m_meshLevels; }
+
+  std::vector< const MeshLevel * > getMeshLevels() const
```
rrsettgast, line 59:
```suggestion
  std::vector< MeshLevel const *> getMeshLevels() const
```
🙄

src/coreComponents/physicsSolvers/SolverBase.hpp
rrsettgast: I think we can change `m_meshTargets` s.t. we don't need to pass `meshBodies`

rrsettgast (commented):
I think this looks pretty good....but if we have different types of `MeshLevel` or `MeshBodies` then it is a bit limiting?

src/coreComponents/finiteVolume/FluxApproximationBase.cpp
TotoGaz: Not really, it's just that eventually we only loop on the mesh levels, so having one loop is fine too _imho_.

Flattening reduces the "_depth_" of the function which can always be nice: sometimes we do have deep functions and it's complicated to know where we are. No big deal but it can be nice.

In the current case of `FluxApproximationBase`, this changes a double loop into a simple one (same real complexity of course). Not a critical improvement.

We could also rename into
```c++
  for( MeshLevel * mesh: domain.getMeshBodies() | getMeshLevels() ) // or GetMeshLevels()
```
for more obvious reading (in the end, this flattening is a reading concern).

_One point which I find of interest though_: the `meshBody` instance which was only used to retrieve the mesh levels is now invisible. I think it's interesting.

But let's be honest, my main idea was to mimic the `flatMap` classical pattern which
- perfectly matches here,
- can be really fluid to read (personal opinion).

TotoGaz:
> if we have different types of `MeshLevel` or `MeshBodies` then it is a bit limiting?

For the moment, our `for*(...)` member functions (`forMeshBodies`, `forSubGroups`, etc.) basically
- provide the data
- select the data such that they match our `class` predicate (_e.g._ `forSubGroups< MeshLevel >(...)` will only select instances of type `MeshLevel`)
- apply the functor

This PR's `std::vector` impl does less things since it considers one type (`MeshBody`, `MeshLevel`) and deliver the data to the client code.

Let's imagine a case with one `MeshBody` instance and multiple `MeshLevels`.
```mermaid
classDiagram
    MeshBody
    MeshLevelABC
    MeshLevelImpl0
    MeshLevelImpl1

    MeshLevelABC <|-- MeshLevelImpl0
    MeshLevelABC <|-- MeshLevelImpl1

    MeshLevelABC *-- MeshBody

    class MeshBody{
      + std::vector~MeshLevelABC*~ getMeshLevels()
    }
```
Let's assume that we do not want `MeshBody` to know about `MeshLevelImpl*`.

If some actions are generic at the `MeshLevelABC` level, nothing needs to be done.

If we need to select only a given type of `MeshLevelImpl`, it's now the duty of the client code to filter in what it wants to keep.
`c++20`'s `std::ranges` would do something like
```c++
for( MeshLevelImpl0 * mesh: domain.getMeshBodies() | filter( MeshLevelImpl0Predicate{} ) )
```
We surely can go the same way with (in the spirit of `flattenMeshLevels`)
We can achieve this by implementing it ourselves, but I'd prefer geosx not to implement too many of those utility functions.

A more old school and heavy solution would be
```c++
for( MeshLevelABC * tmp: domain.getMeshBodies() ){
  if ( MeshLevelImpl0 * mesh = cast<MeshLevelImpl0*>(tmp)) {
    ...
```
or 
```c++
for( MeshLevelABC * tmp: domain.getMeshBodies() ){
  if ( not MeshLevelImpl0Predicate( tmp ) ) continue;
  MeshLevelImpl0 * mesh = cast<MeshLevelImpl0*>(tmp);
```
if we do not want to nest in a `if`.

klevzoff:
I support the overall direction, however I'm opposed to an interface that forces dynamic allocations just to iterate over a collection (sometimes multiple times, as in the case of flattening). It's only going to get worse as we opt more and more looping patterns into the new approach (regions, sub-regions, etc.). At the very least I'd want to see some SBO'ed container returned, to cover the common use case of single-body, single-level simulation. How about [`std::basic_string< MeshBody * >`](https://godbolt.org/z/xcnej4n61) :smile: ?

Another issue with returning vectors is limited expressiveness. For example, we have a pattern
```c++
elemManager.forElementSubRegionsComplete( [&]( localIndex const er,
                                               localIndex const esr,
                                               ElementRegionBase & region,
{
  // do something that requires all 4 things
} );
```
Even split into 2-level nested range-based loops, we'd be iterating over vectors of something like `pair` or `tuple`, which, in absence of structured bindings, is less readable than lambda-based version:
```c++
for( auto && regionPair : elemManager.regions() | enumerate )
{
  ElementRegionBase * region = regionPair.first;
  localIndex const er = regionPair.second;
  for (auto && subRegionPair : region->subregions() | enumerate )
  {
    ElementSubRegionBase * subRegion = subRegionPair.first;
    localIndex const esr = subRegionPair.second;

    // do something that requires all 4 things
  }
}

```

TotoGaz:
> I'm opposed to an interface that forces dynamic allocations just to iterate over a collection

While admittedly not the ideal approach, could we compute figures to get a taste of how much runtime it will cost for a typical GEOSX simulation?
(I speak of runtime because I do not think that memory is really involved here, do you confirm?)

If we have 10 `MeshBody` instances and 10 `MeshLevel` instances, and that everywhere in the code we loop "the `std::vector` way", how much slower will the simulation be?

> (sometimes multiple times, as in the case of flattening)

The flattening (which is really optional) performs one `std::vector` allocation more than the standard double loop, since it gathers all the `MeshLevel` in one big `std::vector`.

> ... elemManager.forElementSubRegionsComplete ...

We could also do some 
```c++
for( localIndex er = 0; er <  elemManager.numRegions(); ++er)
{
  ElementRegionBase * region = elemManager.region( er );
  for( localIndex esr = 0; esr <  region.numSubRegions(); ++esr)
  {
    ElementSubRegionBase * subRegion = region.subRegion( esr );
```
which is not very cool to duplicate everywhere.
Also, structure bindings is really close I think since it's c++17 and CUDA 11 compatible.
So having one big accessor like `elemManager.getElementSubRegionsComplete()` is close (or some `flatten` stuff).

The main motivation of this PR is to be a step towards clearer and modular codebase (maybe using additional abstraction in the future).

PR #1924, opened 2022-05-20
Feature/castelletto1/suppress move logging

castelletto1:
This PR removes default move logging. The proposed change replaces flag `--suppress-move-logging` by `--trace-data-migration`. 

Requires updated integrated tests submodule (baselines are the same) https://github.com/GEOSX/integratedTests/pull/227

### DIFF
--- a/src/coreComponents/mainInterface/ProblemManager.cpp
+++ b/src/coreComponents/mainInterface/ProblemManager.cpp
@@ -199,7 +199,11 @@ void ProblemManager::parseCommandLineInput()
     Path::pathPrefix() = splitPath( inputFileName ).first;
   }
 
-  if( opts.suppressMoveLogging )
+  if( opts.displayMoveLogging )
+  {
+    chai::ArrayManager::getInstance()->enableCallbacks();
+  }
+  else
   {
     chai::ArrayManager::getInstance()->disableCallbacks();
   }
--- a/src/coreComponents/mainInterface/ProblemManager.cpp
+++ b/src/coreComponents/mainInterface/ProblemManager.cpp
@@ -199,7 +199,7 @@ void ProblemManager::parseCommandLineInput()
     Path::pathPrefix() = splitPath( inputFileName ).first;
   }
 
-  if( opts.displayMoveLogging )
+  if( opts.traceDataMigration )
   {
     chai::ArrayManager::getInstance()->enableCallbacks();
   }
--- a/src/coreComponents/mainInterface/initialization.cpp
+++ b/src/coreComponents/mainInterface/initialization.cpp
@@ -98,7 +98,7 @@ std::unique_ptr< CommandLineOptions > parseCommandLineOptions( int argc, char * 
     PROBLEMNAME,
     OUTPUTDIR,
     TIMERS,
-    SUPPRESS_MOVE_LOGGING,
+    DISPLAY_MOVE_LOGGING,
     PAUSE_FOR,
   };
 
@@ -117,7 +117,7 @@ std::unique_ptr< CommandLineOptions > parseCommandLineOptions( int argc, char * 
     { SUPPRESS_PINNED, 0, "s", "suppress-pinned", Arg::None, "\t-s, --suppress-pinned \t Suppress usage of pinned memory for MPI communication buffers" },
     { OUTPUTDIR, 0, "o", "output", Arg::nonEmpty, "\t-o, --output, \t Directory to put the output files" },
     { TIMERS, 0, "t", "timers", Arg::nonEmpty, "\t-t, --timers, \t String specifying the type of timer output." },
-    { SUPPRESS_MOVE_LOGGING, 0, "", "suppress-move-logging", Arg::None, "\t--suppress-move-logging \t Suppress logging of host-device data migration" },
+    { DISPLAY_MOVE_LOGGING, 0, "", "display-move-logging", Arg::None, "\t--display-move-logging \t Display logging of host-device data migration" },
     { PAUSE_FOR, 0, "", "pause-for", Arg::numeric, "\t--pause-for, \t Pause geosx for a given number of seconds before starting execution" },
     { 0, 0, nullptr, nullptr, nullptr, nullptr }
   };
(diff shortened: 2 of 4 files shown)


src/coreComponents/mainInterface/ProblemManager.cpp
```diff
@@ -199,7 +199,11 @@ void ProblemManager::parseCommandLineInput()
     Path::pathPrefix() = splitPath( inputFileName ).first;
   }
 
-  if( opts.suppressMoveLogging )
+  if( opts.displayMoveLogging )
+  {
+    chai::ArrayManager::getInstance()->enableCallbacks();
```
TotoGaz, line 204: I could only find track of the callbacks here https://github.com/GEOSX/LvArray/blob/develop/src/ChaiBuffer.hpp#L447

So I guess it's find to do it like that right now, but we must be aware of this since it's a bit dangerous 😉

TotoGaz (approved):
If you have time, maybe can you fix the doc in `src/docs/sphinx/QuickStart.rst` with the correct `help` from `geosx`:

```Options:
-?, --help
-i, --input,             Input xml filename (required)
-r, --restart,           Target restart filename
-x, --x-partitions,      Number of partitions in the x-direction
-y, --y-partitions,      Number of partitions in the y-direction
-z, --z-partitions,      Number of partitions in the z-direction
-s, --schema,            Name of the output schema
-b, --use-nonblocking,   Use non-blocking MPI communication
-n, --name,              Name of the problem, used for output
-s, --suppress-pinned    Suppress usage of pinned memory for MPI communication buffers
-o, --output,            Directory to put the output files
-t, --timers,            String specifying the type of timer output.
--suppress-move-logging  Suppress logging of host-device data migration
--pause-for,             Pause geosx for a given number of seconds before starting execution
```

src/coreComponents/mainInterface/ProblemManager.cpp
castelletto1: The proposed change is intended to explicitly turn on/off callbacks using chai's methods ([here](https://github.com/LLNL/CHAI/blob/f90689b2958eda7025681c6390c7505a06509f6e/src/chai/ArrayManager.hpp#L389-L397)) instead of relying on chai ArrayManager's default constructor ([here](https://github.com/LLNL/CHAI/blob/f90689b2958eda7025681c6390c7505a06509f6e/src/chai/ArrayManager.cpp#L32)), which turn them on by default. Could you clarify why you believe this change could be potentially dangerous?

src/coreComponents/mainInterface/initialization.cpp
```diff
@@ -114,10 +114,10 @@ std::unique_ptr< CommandLineOptions > parseCommandLineOptions( int argc, char *
...
     { NONBLOCKING_MPI, 0, "b", "use-nonblocking", Arg::None, "\t-b, --use-nonblocking, \t Use non-blocking MPI communication" },
     { PROBLEMNAME, 0, "n", "name", Arg::nonEmpty, "\t-n, --name, \t Name of the problem, used for output" },
-    { SUPPRESS_PINNED, 0, "s", "suppress-pinned", Arg::None, "\t-s, --suppress-pinned \t Suppress usage of pinned memory for MPI communication buffers" },
+    { SUPPRESS_PINNED, 0, "s", "suppress-pinned", Arg::None, "\t-s, --suppress-pinned, \t Suppress usage of pinned memory for MPI communication buffers" },
     { OUTPUTDIR, 0, "o", "output", Arg::nonEmpty, "\t-o, --output, \t Directory to put the output files" },
-    { TIMERS, 0, "t", "timers", Arg::nonEmpty, "\t-t, --timers, \t String specifying the type of timer output." },
-    { SUPPRESS_MOVE_LOGGING, 0, "", "suppress-move-logging", Arg::None, "\t--suppress-move-logging \t Suppress logging of host-device data migration" },
+    { TIMERS, 0, "t", "timers", Arg::nonEmpty, "\t-t, --timers, \t String specifying the type of timer output" },
+    { DISPLAY_MOVE_LOGGING, 0, "", "display-move-logging", Arg::None, "\t--display-move-logging, \t Display logging of host-device data migration" },
```
klevzoff, line 120: How about having a short option for this too? My vote is for `-M` (uppercase, I don't want to reserve lowercase `m` for something that's not a primary parameter):
```suggestion
    { DISPLAY_MOVE_LOGGING, 0, "M", "display-move-logging", Arg::None, "\t-M, --display-move-logging, \t Display logging of host-device data migration" },
```

klevzoff (approved):
Thanks, I've been wanting to reverse the default for a while.

On a very minor pedantic note, that phrasing "display memory logging" sounds a bit awkward to me, we typically either display something or log something, not both. I don't insist on any changes, but if I were to reword this option, I would probably go for something along "trace host-device migration". Naming is hard though, I'm open to better suggestions (or just keeping the minimal change).

src/coreComponents/mainInterface/ProblemManager.cpp
klevzoff: Cleaning this up requires a slightly bigger refactoring. In particular, we need to remove the `setName("")` call from `ChaiBuffer` constructors, so that there's no logging by default and we aren't forced to disable callbacks globally. The method should also be renamed to reflect the purpose, e.g. `enableLogging(string id)`.

Then, on GEOSX side, in `Wrapper<>`, we can check if logging option is enabled in `GeosxState` and optionally call `enableLogging` on the wrapped object. Finally, we then remove all ad-hoc `setName()` calls on various objects (e.g. linear algebra matrices/vectors) and instead register them all in data repository, to manage this all centrally.

But this is a topic for another PR :)

rrsettgast (approved):
Just make sure to have @corbett5 take a look before your merge.

src/coreComponents/mainInterface/initialization.cpp
castelletto1: I'll go for `trace-data-migration`.

corbett5 (commented):
Looks good to me! I certainly agree this is a nice change. It's a good sign that people are both using the GPU and that it works well enough that we don't need/want the logging on by default.

PR #1927, opened 2022-05-23
Implemented initialization elastic time step for coupled problems

francoishamon:
This PR implements the necessary components to perform the initialization time step in Discussion #1926.

That is, assembling and solving the following system:
```
delta x = - J^-1 R     
```
where the Jacobian looks like
```
     | J_uu J_up J_us |
J =  |        I       |
     |             I  |  
```
and the residual has been modified appropriately to fix pressure and compositions at their initial values.

To do that, I needed:
- A function in `CompositionalMultiphaseBase` that uses `FieldSpecifications` functions to modify the flow (J_p* and J_s*) blocks of the Jacobian matrix appropriately (similar to Dirichlet BC).
- A task to trigger the call of this function.

I implemented this for both the  `SinglePhasePoromechanicsSolver` and `MultiphasePoromechanicsSolver`.

I left the initialization MGR recipe outside of the PR for simplicity (the initialization works well without it) but we can add it later. 

Rebaseline done in https://github.com/GEOSX/integratedTests/pull/295


src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -1745,7 +1757,85 @@ void CompositionalMultiphaseBase::applyDirichletBC( real64 const time_n,
...
+  {
+    mesh.getElemManager().forElementSubRegions( regionNames,
+                                                [&]( localIndex const,
+                                                     ElementSubRegionBase const & subRegion )
+    {
+      globalIndex const rankOffset = dofManager.rankOffset();
+      string const dofKey = dofManager.getKey( viewKeyStruct::elemDofFieldString() );
 
+      arrayView1d< integer const > const ghostRank = subRegion.ghostRank();
+      arrayView1d< globalIndex const > const dofNumber = subRegion.getReference< array1d< globalIndex > >( dofKey );
```
TotoGaz, line 1786: Could we define `dofNumber` as a "field" and use the `getField` pattern?
Or even a std getter?

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -1745,7 +1757,85 @@ void CompositionalMultiphaseBase::applyDirichletBC( real64 const time_n,
...
+      integer const isThermal = m_isThermal;
+      forAll< parallelDevicePolicy<> >( subRegion.size(), [=] GEOSX_HOST_DEVICE ( localIndex const ei )
+      {
+        if( ghostRank[ei] >= 0 )
+        {
+          return;
+        }
+
+        globalIndex const dofIndex = dofNumber[ei];
+        localIndex const localRow = dofIndex - rankOffset;
```
TotoGaz, line 1802: Maybe it's not really for this PR, but could we get this pattern embedded/unified somewhere.
Maybe like `localIndex const localRow = dofManager.buildLocalRow( dofIndex )`?
So we could document what going on. But maybe there are some other constraints also...

src/coreComponents/physicsSolvers/multiphysics/PoromechanicsInitialization.cpp
```diff
@@ -0,0 +1,134 @@
...
+    m_poromechanicsSolver->performStressInitialization( true );
+  }
+  else
+  {
+    GEOSX_LOG_LEVEL_RANK_0( 1, GEOSX_FMT( "Task `{}`: at time {}s, physics solver `{}` has completed stress initialization",
+                                          getName(), time_n, m_poromechanicsSolverName ) );
+    m_poromechanicsSolver->performStressInitialization( false );
+  }
+
+  return false;
```
TotoGaz, line 123: So it always returns `false`?

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoromechanicsSolver.hpp
```diff
@@ -105,8 +105,19 @@ class MultiphasePoromechanicsSolver : public CoupledSolver< SolidMechanicsLagran
...
+   * @brief Utility function to update the stabilization parameters at each time step
+   * @param[in] domain the domain partition
+   */
   void updateStabilizationParameters( DomainPartition & domain ) const;
 
+  /*
+   * @brief Utility function to set the stress initialization flag
+   * @param[in] performStressInitialization true if the solver has to initialize stress, false otherwise
+   */
+  void performStressInitialization( integer const performStressInitialization )
```
TotoGaz, line 118: This is a bit ambiguous since this member function does not `perform` anything. `setStressInitialization`?

src/coreComponents/physicsSolvers/multiphysics/PoromechanicsInitialization.hpp
```diff
@@ -0,0 +1,106 @@
...
+#ifndef SRC_CORECOMPONENTS_PHYSICSSOLVERS_MULTIPHYSICS_POROMECHANICSINITIALIZATION_HPP_
+#define SRC_CORECOMPONENTS_PHYSICSSOLVERS_MULTIPHYSICS_POROMECHANICSINITIALIZATION_HPP_
+
+#include "events/tasks/TaskBase.hpp"
+
+namespace geosx
+{
+
+class MultiPhasePoromechanicsSolver;
+class SinglePhasePoromechanicsSolver;
```
TotoGaz, line 28: Do you actually need this?

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.hpp
```diff
@@ -324,6 +324,32 @@ class CompositionalMultiphaseBase : public FlowSolverBase
...
+   * @param[in] time current time
+   * @param[in] dt time step
+   * @param[in] dofManager degree-of-freedom manager associated with the linear system
+   * @param[in] domain the domain
+   * @param[in] localMatrix local system matrix
+   * @param[in] localRhs local system right-hand side vector
+   * @detail This function is meant to be called when the flag m_keepFlowVariablesConstantDuringStep is on
+   *         The main use case is the initialization step in coupled problems during which we solve an elastic problem for a fixed pressure
+   */
+  void keepFlowVariablesConstantDuringStep( real64 const time,
```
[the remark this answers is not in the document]
francoishamon: Here, unfortunately, we cannot make this function `private` or `protected` to compile on the GPU platforms because of the error:
```
The enclosing parent function ("keepFlowVariablesConstantDuringStep") for an extended __host__ __device__ lambda cannot have private or protected access within its class
```
We could move the body of `keepFlowVariablesConstantDuringStep` to a separate kernel file to overcome this issue, but this would be odd because `keepFlowVariablesConstantDuringStep` simply calls the `FieldSpecificationOps` utility functions...

src/coreComponents/physicsSolvers/multiphysics/PoromechanicsInitialization.cpp
francoishamon: I added a comment before the `return` to explain why this function always returns `false`. In fact, all the `execute` functions must return false, because `true` is the signal for early return and terminates the simulation with the following message:
```
***** WARNING
***** LOCATION: /g/g90/hamon1/geosx/GEOSX/src/main/main.cpp:46
***** Controlling expression (should be false): state.getState() != State::COMPLETED
Simulation exited early.
```
See here in the `EventManager`: https://github.com/GEOSX/GEOSX/blob/c5d775303be29bad5956bc303d32da4571a9c095/src/coreComponents/events/EventManager.cpp#L186-L201

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoromechanicsSolver.hpp
francoishamon: Done, I renamed this function `setStressInitialization` as you suggested

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
francoishamon: It may be difficult to use the `getField` pattern because `dofKey` is computed dynamically by the `DofManager`. 

However, we could add a standard getter in `ObjectManagerBase`:
```
arrayView1d< globalIndex const > getDofNumber( string const & dofKey ) const 
{
  return getReference< array1d< globalIndex > >( dofKey );
}
```
and then we would use it as:
```
arrayView1d< globalIndex const > const dofNumber = subRegion.getDofNumber( dofKey );
```
Let me know what you think, I can add it in a separate PR

src/coreComponents/physicsSolvers/multiphysics/PoromechanicsInitialization.hpp
francoishamon: Yes, you are right, I removed it and it still compiled so they seem unnecessary.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
francoishamon: Yes, we could do something like that, but we do not want to bring the `DofManager` in the kernel. 

So either we would compute a vector of `localRow`s in advance, or the `DofManager` would have some kind of kernel wrapper to do `localRow = dofIndex - rankOffset`. It may be an overkill, but maybe there is another way to unify this that I don't see.

TotoGaz: I guess it's OK like this.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
TotoGaz: That looks more engaging!

TotoGaz (approved):

Issue #1632, opened 2021-10-06
[Bug] real64 xml inputs do not throw an error when alpha characters are present


cssherman:
Using the sedov problem as an example, I found that the code does not throw an error when it detects alpha characters when parsing real64 values (at least) from an input xml.  For example, you could replace the forceDt in the sedov example, and it would run:

```xml
<PeriodicEvent
      name="solverApplications"
      forceDt="NotANumber"
      target="/Solvers/lagsolve"/>
```

klevzoff:
Fixed

PR #1928, opened 2022-05-24
Adding an extra string parsing check for attributes

cssherman:
This PR adds a quick check to the istringstream for attribute parsing errors (#1632).  It will catch previously silent errors that occur when unexpected alpha characters are included.  For example

- `floatVar="1.2e3"` is the correct value, which would be parsed as 1.2e3
- `floatVar="badstring1.2e3"` would be parsed as 0
- `floatVar="1badstring.2e3"` would be parsed as 1
- `floatVar="1.2e3badstring"` would be parsed as 1.2e3
- `floatVar="      1.2e3     "` would also be parsed as 1.2e3

To catch all of the above cases, we can check to see if the error flag is set (catches errors at the beginning) and that the eof flag is not set (catches errors in the middle / end of the string).  This won't catch spaces at the beginning of the string, which should be OK.


src/coreComponents/dataRepository/xmlWrapper.hpp
```diff
@@ -120,6 +120,9 @@ stringToInputVariable( T & target, string const & value )
 {
   std::istringstream ss( value );
   ss >> target;
+  GEOSX_THROW_IF( ss.fail() | !ss.eof(),
```
klevzoff, line 123: Is bitwise `or` intentional? It's not wrong, just a little unconventional.

klevzoff (approved):
Thanks!

If we want to be strict about leading whitespace, we could also throw `std::noskipws` at the stream. But I'm not sure it's worth it.

cssherman:
I'll work on adding the proposed unit tests here

cssherman:
@klevzoff - any thoughts on the new unit tests?

src/coreComponents/dataRepository/unitTests/testXmlWrapper.cpp
```diff
@@ -175,6 +175,175 @@ TEST( testXmlWrapper, TensorIllFormed )
...
+                     std::make_tuple( "1beta234", 0, true ),
+                     std::make_tuple( "1.234gamma", 0, true ),
+                     std::make_tuple( "1.2.3", 0, true ),
+                     std::make_tuple( "1e2.3 ", 0, true ),
+                     std::make_tuple( "1 ", 0, true ),
+                     std::make_tuple( "1e", 0, true ),
+                     std::make_tuple( "1e-", 0, true ),
+                     std::make_tuple( "1e+", 0, true )));
+
+class integerAttributeTestFixture : public ::testing::TestWithParam< std::tuple< string, integer, bool > >
```
klevzoff, line 299: The three tests (`real64`, `real32` and `integer`) are quite similar, any chance we can combine all the shared testing code (both `parseString` and as much as possible from `TEST_P` blocks) in one base template, e.g.
```c++
template< typename T >
class AttributeReadTestFixture : public ::testing::TestWithParam< std::tuple< string, T, bool > >
{
  void parseString() { ... }
  void test() { ... }
};
```
and then (repeated for all types)
```c++
class real64AttributeTestFixture : public AttributeReadTestFixture< real64 > {};
TEST_P( real64AttributeTestFixture, testParsing )
{
  this->test();
}
INSTANTIATE_TEST_CASE_P( real64AttributeTests, real64AttributeTestFixture, ... );
```

Essentially going as close as possible to a type-parametrized test without declaring one (since it's not compatible with value-parametrized).

The only thing that differs across tests is the actual comparison assertion, it can be moved into a set of top-level functions overloaded or specialized for the different types.

src/coreComponents/dataRepository/unitTests/testXmlWrapper.cpp
```diff
@@ -175,6 +175,175 @@ TEST( testXmlWrapper, TensorIllFormed )
...
+  string attributeString;
+  real64 expectedValue;
+  real64 parsedValue;
+  bool failureFlag;
+
+  void parseString()
+  {
+    GEOSX_LOG_RANK_0( "Parsing string: " << attributeString );
+    xmlWrapper::stringToInputVariable< real64 >( parsedValue, attributeString );
+    ASSERT_NEAR( expectedValue, parsedValue, 1e-10 );
```
klevzoff, line 192: Unless it causes any trouble, I would maybe suggest `EXPECT_DOUBLE_EQ` for this, to avoid having to choose absolute tolerances.

(As a side note, I prefer `EXPECT` macros over `ASSERT` ones, unless I really need to terminate the test early. Not that it matters here.)

src/coreComponents/dataRepository/unitTests/testXmlWrapper.cpp
```diff
@@ -175,6 +175,175 @@ TEST( testXmlWrapper, TensorIllFormed )
...
+  string attributeString;
+  real32 expectedValue;
+  real32 parsedValue;
+  bool failureFlag;
+
+  void parseString()
+  {
+    GEOSX_LOG_RANK_0( "Parsing string: " << attributeString );
+    xmlWrapper::stringToInputVariable< real32 >( parsedValue, attributeString );
+    ASSERT_NEAR( expectedValue, parsedValue, 1e-10 );
```
klevzoff, line 252: Similarly, `EXPECT_FLOAT_EQ` would probably do the trick?

src/coreComponents/dataRepository/unitTests/testXmlWrapper.cpp
```diff
@@ -175,6 +175,175 @@ TEST( testXmlWrapper, TensorIllFormed )
...
+  string attributeString;
+  integer expectedValue;
+  integer parsedValue;
+  bool failureFlag;
+
+  void parseString()
+  {
+    GEOSX_LOG_RANK_0( "Parsing string: " << attributeString );
+    xmlWrapper::stringToInputVariable< integer >( parsedValue, attributeString );
+    ASSERT_TRUE( expectedValue == parsedValue );
```
klevzoff, line 311: `EXPECT_EQ`?

src/coreComponents/dataRepository/unitTests/testXmlWrapper.cpp
```diff
@@ -175,6 +175,175 @@ TEST( testXmlWrapper, TensorIllFormed )
...
+{
+protected:
+  string attributeString;
+  integer expectedValue;
+  integer parsedValue;
+  bool failureFlag;
+
+  void parseString()
+  {
+    GEOSX_LOG_RANK_0( "Parsing string: " << attributeString );
```
klevzoff, line 309: Not a big deal, but it's better if a cleanly passing test does not print anything. If you really need to see the source string upon failure, maybe 
```c++
SCOPED_TRACE( "Parsing string: " << attributeString );
```
will do the job?

klevzoff (approved):
Nice! Some minor comments on the tests, take what you like.

PR #1931, opened 2022-05-25
Fixed misleading behavior in source flux boundary condition 

### LINKED ISSUE #1930: [Cleanup] Misleading behavior of SourceFlux boundary condition
**Describe the issue**
I have a case with a reservoir region made of multiple (imported) cell blocks:
```
    <CellElementRegion                                                                                                                                                                                          
      name="reservoir"                                                                                                                                                                                          
      cellBlocks="{ 2_hexahedra, 2_tetrahedra, 2_pyramids,                                                                                                                                                      
                    3_hexahedra, 3_tetrahedra, 3_pyramids,                                                                                                                                                      
                    4_hexahedra, 4_tetrahedra, 4_pyramids }"                                                                                                                                                    
      materialList="{ porousRockReservoir, fluid, relperm }"/
```
in which I want to inject at a rate of -47.5646879756 kg/s. My source flux looks like 
```
    <SourceFlux                                                                                                                                                                                                 
      name="sourceTerm"                                                                                                                                                                                         
      component="0"                                                                                                                                                                                             
      setNames="{ well }"                                                                                                                                                                                       
      objectPath="ElementRegions/reservoir"                                                                                                                                                                     
      functionName="totalMassTable"                                                                                                                                                                             
      scale="1"/>   
```
But, in this case, the `Box` named `well` touches 2 cell blocks. As a result, the `scale` normalization [here](https://github.com/GEOSX/GEOSX/blob/2b14bfa4cfc418a5b884bc5a8a7a01d4dfd84980/src/coreComponents/fieldSpecification/FieldSpecificationBase.hpp#L770) is done independently in each cell block, and we end up injecting - 2 * 47.5646879756 kg/s.

**Proposed cleanup**
I propose to remove the normalization from `FieldSpecificationBase`, and give the solvers the responsibility to normalize across multiple cell blocks, as done in PR #1931 

Another possible solution is to issue a warning/an error when the source flux touches multiple cell blocks

francoishamon:
This PR fixes #1930  I moved the computation of the size scaling factor to the solvers, which makes it possible to normalize consistently the `scale` across multiple cell blocks.

The PR passes the integrated tests and is ready for review.

### DIFF
--- a/src/coreComponents/fieldSpecification/FieldSpecificationBase.cpp
+++ b/src/coreComponents/fieldSpecification/FieldSpecificationBase.cpp
@@ -24,8 +24,7 @@ namespace geosx
 using namespace dataRepository;
 
 FieldSpecificationBase::FieldSpecificationBase( string const & name, Group * parent ):
-  Group( name, parent ),
-  m_normalizeBySetSize( false )
+  Group( name, parent )
 {
   setInputFlags( InputFlags::OPTIONAL_NONUNIQUE );
 
--- a/src/coreComponents/fieldSpecification/FieldSpecificationBase.hpp
+++ b/src/coreComponents/fieldSpecification/FieldSpecificationBase.hpp
@@ -502,9 +502,6 @@ public:
 protected:
   virtual void postProcessInput() override;
 
-  /// The flag used to decide if the BC value is normalized by the size of the set on which it is applied
-  bool m_normalizeBySetSize;
-
 private:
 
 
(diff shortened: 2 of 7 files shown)


klevzoff (approved):
Great! The new behavior is the correct one in presence of multiple regions/subregions.

I think potentially we might also want to distribute the source flux proportionally to cell volumes. With this PR it would be fairly easy to do (just replacing the `globalIndex` scaling factor with a `real64` one). But this is something to be discussed separately.

francoishamon:
I checked the integrated tests and they still pass with this PR so I will merge it now

Issue #1940, opened 2022-06-01
[Feature] Importing nodesets with VTKMesh


cssherman:
**What is the requested feature?**
Currently, there is no method to import mesh nodesets using the vtk format.  The two options I see for this include:

1) Use the existing infrastructure to import binary masks for each set as a DataArray
2) Convert nodesets into a VTK_POLY_VERTEX cells

Each of these would require the user to supply a list of nodeset names to import or would require special keywords to alert the parser to them automatically.   Option 1 would be quite easy to implement, but could increase file sizes without compression.  Option 2 would be more compact, but would require us to build a new mesh conversion tool.

cssherman:
Any thoughts on the approach @rrsettgast @joshua-white ?

As a side-note, I'm working to convert a large problem to use the VTK mesh due to some issues that I've encountered with the partitioning in PAMELA.  If we go with option 1, then I could put together a quick prototype.

TotoGaz:
@cssherman you may be interested in https://github.com/GEOSX/GEOSX/issues/1932 which, while not equivalent, may add some features of interest.
Getting things consistent would help 😉

PR #1942, opened 2022-06-02
Adding method to import nodesets in vtk meshes, convert abaqus meshes to vtu

cssherman:


cssherman:
I've tested the prototype methods implemented here on a couple of simple block meshes, which were converted from an abaqus format.  Qualitatively at least, it behaved as expected in serial and with 2x2x2 partitions.

cssherman:
Also, I can confirm that the default compression in VTK does a good job of minimizing the size of nodeset masks

cssherman:
@AntoineMazuyer @francoishamon - could you give this a look?  This would be useful to include in the new VTK documentation if you are on board with the approach

francoishamon (approved):
Sorry for the slow review, this looks good to me

cssherman:
#1940

PR #1943, opened 2022-06-02
Testing the black python code formatter

cssherman:
Applying to the scripts and python directories.  Note: black ignores and throws a warning for python files where it detects python 2.x patterns (such as print)


src/coreComponents/python/modules/hdf5_wrapper_package/hdf5_wrapper/wrapper.py
```diff
@@ -44,127 +43,127 @@ def __getitem__(self, k):
...
-    self.target[k] = h5py.ExternalLink(target, '/')
+        self.target[k] = h5py.ExternalLink(target, "/")
 
-  def keys(self):
-    """
+    def keys(self):
+        """
     @brief get a list of groups and arrays located at the current level
     @return a list of strings 
     """
```
cssherman, line 113: Black doesn't check the indentation of multi-line docstrings after the first line, as that could unexpectedly modify its contents

scripts/file_search.py
```diff
@@ -17,48 +17,62 @@ def html_tail():
...
+                    "-i",
+                    "-n",
+                    "-A{}".format(n_lines_after),
+                    "-B{}".format(n_lines_before),
+                    todo_token,
+                    file_name,
+                ],
+                stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE,
+            )
```
TotoGaz, line 65: Does this stay like this with longer lines? I'm unpleased to use 12 lines for a `grep`. But so be it.

scripts/SchemaToRSTDocumentation.py
```diff
@@ -1,229 +1,242 @@
...
+            tmp = comment_node.text.split(" => ")
+            att_name = tmp[0]
+            att_description = (
+                tmp[1].replace("\\\\", "\\").replace("\n", "\\n").replace("|", "\\|")
+            )
+            attribute_map[type_name][att_name]["Description"] = att_description
+            if len(tmp) > 2:
+                attribute_map[type_name][att_name]["Registered By"] = [
+                    x for x in tmp[2].split(", ")
+                ]
```
TotoGaz, line 135: Can these new lines be removed with longer line threshold?

scripts/makeSampleMeshFilesCuboctahedral.py
```diff
@@ -184,40 +206,88 @@ def print_to_vtu(fname, nodes, faces, index_map):
     cell_types = [42 for _ in range(ncells)]
     face_connectivity, face_offsets = translate_faces(faces, index_map)
     # Write to file
-    with open(fname, 'w') as outfile:
+    with open(fname, "w") as outfile:
```
TotoGaz, line 209: So many lines for a repetitive pattern. We loose the overview of the function that cannot be contained in a screen anymore. 🤷

scripts/postProcessing/hydraulicFractureVerification/HydrofractureSolutions.py
```diff
@@ -91,288 +85,306 @@ def Solutions( self, mu, E, q, KI, time, time_radplots, distanceFromInjection ):
...
+            * scipy.special.beta(-1.5, 8.0 / 3.0)
+            * (
+                2.0
+                / 3.0
+                * rho
+                * rho
+                * scipy.special.hyp2f1(-1.0 / 6.0, 2, 1.5, rho * rho)
+                - 0.5 * scipy.special.hyp2f1(-7.0 / 6.0, 1, 0.5, rho * rho)
+            )
+        )
```
TotoGaz, line 238: 🙈

src/coreComponents/python/modules/geosx_xml_tools_package/geosx_xml_tools/unit_manager.py
```diff
@@ -46,54 +56,94 @@ def buildUnits(self):
...
+                "value": 0.1589873,
+                "alt": ["bbl", "barrels"],
+                "usePrefix": True,
+            },
+            "horsepower": {
+                "value": 745.7,
+                "alt": ["hp", "horsepowers"],
+                "usePrefix": True,
+            },
+        }
```
TotoGaz, line 146: This is the kind of issues I do not like with autoformatter.
Sometimes you want to impose specific formatting because it's more readable...
Whatever 🤷

scripts/file_search.py
cssherman: Agreed, it *should* be responsive to line length

scripts/SchemaToRSTDocumentation.py
```diff
@@ -1,229 +1,242 @@
...
+                    "Default": node_use,
+                }
+
+        # Read the remaining values stored in comments (description, registered_by)
+        for comment_node in child_node.iter(etree.Comment):
+            tmp = comment_node.text.split(" => ")
+            att_name = tmp[0]
+            att_description = (
+                tmp[1].replace("\\\\", "\\").replace("\n", "\\n").replace("|", "\\|")
+            )
```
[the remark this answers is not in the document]
cssherman: I'm suspicious that it is driven by this part of the style guide:  https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#call-chains

scripts/postProcessing/hydraulicFractureVerification/HydrofractureSolutions.py
cssherman: That is a bit gross.  Apparently this was adopted into a recent version of pep8: https://peps.python.org/pep-0008/#should-a-line-break-before-or-after-a-binary-operator.  We could dial up the line length to eleven, and avoid these issues entirely.

cssherman: Although these lines in particular are a bit annoying, and would improve with a few additional parentheses (I don't like naked divide expressions, as you have to start thinking about order of operations).

scripts/SchemaToRSTDocumentation.py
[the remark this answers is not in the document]
TotoGaz: More https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#parentheses I think.
I'm uncomfortable with the formatter adding code, would it be parenthesis. But it looks like it's safe...

TotoGaz:
I think it's better with longer lines.
Maybe the result is not perfect but it's really better than nothing.
I think you should go!

TotoGaz:
Maybe consider adding a `black` check in the CI?

cssherman:
> Maybe consider adding a `black` check in the CI?
> 
> ```
>   --check                         Don't write the files back, just return the
>                                   status. Return code 0 means nothing would
>                                   change. Return code 1 means some files would
>                                   be reformatted. Return code 123 means there
>                                   was an internal error.
> ```


Absolutely

cssherman:
> I think it's better with longer lines. Maybe the result is not perfect but it's really better than nothing. I think you should go!

Sounds good!

cssherman:
@TotoGaz - After digging around in the cmake files, I found that blt already supports python formatting with yapf (a competitor to black, which is also quite opinionated).  I'll check to see how the code looks with that instead

PR #1947, opened 2022-06-06
Format python files with YAPF

cssherman:
It turns out that BLT supports YAPF python formatting, so it may be more useful than black.   It seems like it may handle formatting long lines a bit better as well.


host-configs/LLNL/quartz-gcc@8.1.0.cmake
```diff
@@ -24,4 +24,7 @@ set(MKL_LIBRARIES ${MKL_ROOT}/lib/libmkl_intel_lp64.so
 set(ENABLE_PYGEOSX ON CACHE BOOL "")
 set(Python3_EXECUTABLE /usr/gapps/GEOSX/thirdPartyLibs/python/linux-rhel7-x86_64-gcc@8.1.0/python@3.8.5/bin/python3 CACHE PATH "")
 
+# Enable yapf formatting
+set(YAPF_EXECUTABLE /g/g17/sherman/Python/virtual/smart/bin/yapf CACHE PATH "" FORCE)
```
cssherman, line 28: I'll figure out where yapf exits on LC systems and update this

src/yapf.cfg
```diff
@@ -0,0 +1,6 @@
+# Style settings example shown in
+# https://github.com/google/yapf/blob/master/README.rst
+[style]
+based_on_style = pep8
+spaces_before_comment = 4
+split_before_logical_operator = true
```
cssherman, line 6: @TotoGaz - these are settings used to format the files in the PR.  There are many more options compared to black if we want to tinker: https://github.com/google/yapf#knobs

cssherman: Similar to uncrustify, the user can format/check the python files by running `make yapf_style` or `make_yapf_check`

TotoGaz (commented):
Having it already in BLT and managed like we do the `doxygen` is nice.
Maybe increasing `COLUMN_LIMIT` like we did for `black`?

cssherman:
Another feature I like with yapf is the ability to turn off the formatter for a group of lines like so:

```python
# yapf: disable
FOO = {
    # ... some very large, complex data literal.
}

BAR = [
    # ... another large data literal.
]
# yapf: enable
```

Or a single line:
```python

BAZ = {
    (1, 2, 3, 4),
    (5, 6, 7, 8),
    (9, 10, 11, 12),
}  # yapf: disable
```

scripts/checkParameterCase.py
```diff
@@ -10,66 +9,65 @@ def parse_schema_element(root,
                          path,
                          xsd='{http://www.w3.org/2001/XMLSchema}',
                          recursive_types=['PeriodicEvent', 'SoloEvent', 'HaltEvent']):
-  """
+    """
```
klevzoff, line 12: Looks like docstrings are not properly aligned after changing indentation: https://github.com/google/yapf/issues/575

We should probably do it manually once

src/coreComponents/python/modules/geosx_xml_tools_package/geosx_xml_tools/unit_manager.py
```diff
@@ -46,54 +46,215 @@ def buildUnits(self):
...
-                    'kilo':  {'value': 1e3,  'alt': 'k'},
-                    'hecto': {'value': 1e2,  'alt': 'H'},
-                    'deca':  {'value': 1e1,  'alt': 'D'},
-                    '':      {'value': 1.0,  'alt': ''},
-                    'deci':  {'value': 1e-1, 'alt': 'd'},
-                    'centi': {'value': 1e-2, 'alt': 'c'},
-                    'milli': {'value': 1e-3, 'alt': 'm'},
-                    'micro': {'value': 1e-6, 'alt': 'mu'},
-                    'nano':  {'value': 1e-9, 'alt': 'n'}}
+        prefixes = {
```
klevzoff, line 49: Let's restore original formatting and wrap these into `# yapf: disable`?

klevzoff (approved):
I like what I see for the most part

scripts/checkParameterCase.py
cssherman: Agreed, and it seems that it avoids them intentionally.  I'll do a pass to fix these

src/coreComponents/python/modules/geosx_xml_tools_package/geosx_xml_tools/unit_manager.py
cssherman: Agreed

PR #1950, opened 2022-06-09
Fix VTK memory leak

TotoGaz:
Reference count of `vtk` smart pointers looks flawed when implicitly converted to/from raw pointers.
Resulting in memory leak.

And I now use pointers everywhere for `vtk` instances instead of references for more fluid notations.


AntoineMazuyer (approved):
Ok for me. 

Reading the [vtk doc](https://vtk.org/Wiki/VTK/Tutorials/SmartPointers#Creating_an_Object_with_a_Smart_Pointer), I find that the recommended use is : 

`vtkSmartPointer<vtkObject> MyObject = vtkSmartPointer<vtkObject>::New();`

It's basically what you are proposing all the way. If you want to keep `auto` it's ok me.

TotoGaz:
> If you want to keep `auto` it's ok me.

Yes, I think that `auto` is fine in that context.

PR #1956, opened 2022-06-22
Add VTK Outputs wrappers for PyGEOSX

paloma-martinez:
This PR adds a wrapper for the VTKOuputs for pygeosx.


src/coreComponents/fileIO/python/PyVTKOutput.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+  if( !PyArg_ParseTuple( args, "dd", &time, &dt ) )
+  {
+    return nullptr;
+  }
+
+  geosx::ProblemManager & problem = self->group->getGroupByPath< ProblemManager >( "/Problem/" );
+  geosx::DomainPartition & domain = problem.getDomainPartition();
+
+  int cycleNumber = int(time/dt);
```
labesse40, line 85: `int(round(time/dt))` otherwise we can have precision error.

src/coreComponents/fileIO/python/PyVTKOutput.cpp
```diff
@@ -0,0 +1,204 @@
...
+  char const * const path = PyBytes_AsString( asciiPath );
+  if( path == nullptr )
+  {
+    return nullptr;
+  }
+
+  try
+  {
+    self->group->setOutputDirectory( path );
+    self->group->postProcessInput();
```
labesse40, line 125: I am not sure that calling `postProcessInput()` here is a good idea. I think it could break if you call `setOutputDir()` before calling `apply_initial_conditions()` on python side. Can you test it ?

src/coreComponents/fileIO/python/PyVTKOutput.cpp
```diff
@@ -0,0 +1,204 @@
...
+  char const * const filename = PyBytes_AsString( asciiFileNameRoot );
+  if( filename == nullptr )
+  {
+    return nullptr;
+  }
+
+  try
+  {
+    self->group->setPlotFileRoot( filename );
+    self->group->postProcessInput();
```
labesse40, line 161: Same remark as `setOutputDir()`. Can you test it ?

src/coreComponents/fileIO/python/PyVTKOutput.cpp
```diff
@@ -0,0 +1,204 @@
...
+  }
+  Py_RETURN_NONE;
+}
+
+
+static PyMethodDef PyVTKOutput_methods[] = {
+  { "output", (PyCFunction) output, METH_VARARGS, "wrapper to routine VTKOutput::execute"},
+  { "setOutputDir", (PyCFunction) setOutputDir, METH_VARARGS, "wrapper to change directory output"},
+  { "setOutputFileRootName", (PyCFunction) setOutputFileRootName, METH_VARARGS, "wrapper to change the root name of the output files and subfolders"},
+  { nullptr, nullptr, 0, nullptr }      /* Sentinel */
```
labesse40, line 175: I would add a `reinit()` method that calls the GEOSX `reinit()` method which would be overwriten in the `VtkFiles` with a call to `postProcessInput()` probably.

labesse40 (changes requested):

src/coreComponents/fileIO/python/PyVTKOutput.cpp
paloma-martinez: It seems to be OK even if I call `apply_initial_conditions()` after the `setOutputDir()`. No breaking and the name is correctly changed.

src/coreComponents/fileIO/python/PyVTKOutput.cpp
paloma-martinez: No problem here either with the same kind of test.

src/coreComponents/fileIO/python/PyVTKOutput.cpp
labesse40: Nice !

src/coreComponents/fileIO/python/PyVTKOutput.cpp
labesse40: Edit : Since the `postProcessInput()` call in the `set()` routines works. There is no need of `reinit()` here.

labesse40 (approved):

src/coreComponents/fileIO/vtk/VTKPVDWriter.cpp
```diff
@@ -52,5 +52,13 @@ void VTKPVDWriter::addData( real64 time, string const & filePath ) const
   dataSetNode.append_attribute( "timestep" ) = time;
   dataSetNode.append_attribute( "file" ) = filePath.c_str();
 }
+
+void VTKPVDWriter::reinitData() const
```
labesse40, line 56: Is there a way we can access `m_pvdFile` from `VTKOutput` and put these 2 lines directly in `VTKOuput::reinit()` ? This would avoid to have 3 reinit() methods ( `reinit()` `reinitData()` `clearData()` ).

src/coreComponents/fileIO/vtk/VTKPVDWriter.cpp
```diff
@@ -52,5 +52,13 @@ void VTKPVDWriter::addData( real64 time, string const & filePath ) const
...
+{
+  auto collectionNode = m_pvdFile.child( "VTKFile" ).child( "Collection" );
+
+  while( collectionNode.remove_child( "DataSet" ) )
```
labesse40, line 60: Is there a way we can access `m_pvdFile` from `VTKOutput` and put these 2 lines directly in `VTKOuput::reinit()` ? This would avoid to have 3 reinit() methods ( `reinit()` `reinitData()` `clearData()` ).

sframba (approved):
OK for me, now that we have avoided adding a dependency to mainInterface in fileIO's CMakeLists.txt file

PR #1961, opened 2022-06-24
Formatting schema string type names to avoid compiler-related diffs

cssherman:


src/coreComponents/schema/docs/AcousticSEM_other.rst
```diff
@@ -1,20 +1,20 @@
...
+Name                      Type                                                                                                                                        Description                                                             
+========================= =========================================================================================================================================== ======================================================================= 
+indexSeismoTrace          integer                                                                                                                                     Count for output pressure at receivers                                  
+maxStableDt               real64                                                                                                                                      Value of the Maximum Stable Timestep for this solver.                   
+meshTargets               geosx_mapBase< string, LvArray_Array< string, 1, camp_int_seq< long, 0l >, int, LvArray_ChaiBuffer >, std_integral_constant< bool, true > > MeshBody/Region combinations that the solver will be applied to.
```
cssherman, line 8: The string type name changes tend to be buried in these maps.  I've added a simple regex to convert the compiler-dependent values to be string (tested on clang and gcc)

klevzoff (approved):
Nice!

PR #1969, opened 2022-07-08
Azure Pipelines

bmhan12:
- This PR adds azure pipelines for ci testing.
  - Job time limit is a max of 6 hours, which should help with longer cuda jobs.
- Moves the "geosx_linux_build" logic to a `ci_build_and_test.sh` script.
  -  Necessary change, as azure pipelines [does not support yaml anchors](https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/?view=azure-pipelines#see-also), cuts down on script duplication.
  - Renames ` travis_build_and_test.sh` -->  `scripts/build_test_helper.sh`

- [ ] TODO Post PR: Currently only the linux Docker jobs have been migrated to azure pipelines. Still have osx and total docker jobs to migrate over from Travis CI.


klevzoff (approved):
This is great! Looks like Azure has beefier machines, so the individual builds complete faster as well, in addition to higher time limit. I also like being able to see status of the individual stages/jobs in the preview popup.

TotoGaz:
I'm off until ~July 25th so I won't be able to have a deep check.
At first glance it looks OK.
Maybe could you remove the duplicated steps (`documentation`, `sub module check`...) to reduce what we do in travis.
I would replace `ubuntu-20.04` by their real names to control the upgrades.
What about mac? looks like azure provides such machines. I was not working?

rrsettgast (approved):
Looks good.

Issue #1417, opened 2021-05-04
ASCII table output


joshua-white:
**What is the requested feature?**
Write an `AsciiTableOutput` class derived from `OutputBase` that takes an `array2D` and writes it to output, using path checking and sanitation features from `OutputBase`.  This can replace the TriaxialDriver's direct output, and may be useful for writing solver statistics, etc.

Alternative would be use HDF5 output, but for some data a direct-to-text approach will make the user's life easier.

paveltomin:
can it be closed now since https://github.com/GEOS-DEV/GEOS/pull/2984 is merged?

PR #1979, opened 2022-07-19
Implemented solver statistics

francoishamon:
This PR adds some reporting for solver statistics: number of time steps and successful/failed nonlinear/linear iterations.

This is done by adding a sub-group `SolverStatistics` to `SolverBase` that is registered automatically in all simulations. The statistics are then registered on the sub-group `SolverStatistics` to be accessible by `TimeHistory` when it can output scalar quantities.

For now, the statistics are automatically output to terminal at the end of the simulation. This is done by overriding the `cleanup` method of `ExecutableGroup` in `SolverBase`. This terminal output takes place just after the `Cleaning up events` message and looks like this:

Fully coupled schemes:
```
reservoirAndWellsSolver, number of time steps: 52
reservoirAndWellsSolver, number of successful nonlinear iterations: 165
reservoirAndWellsSolver, number of successful linear iterations: 1466
reservoirAndWellsSolver, number of time step cuts: 0
reservoirAndWellsSolver, number of discarded nonlinear iterations: 0
reservoirAndWellsSolver, number of discarded linear iterations: 0
```

Sequential schemes (only outer iterations stats are reported, but sub-solvers stats available in `TimeHistory`):
```
FlowProppantTransport, number of time steps: 60
FlowProppantTransport, number of successful nonlinear iterations: 60
FlowProppantTransport, number of time step cuts: 0
FlowProppantTransport, number of discarded nonlinear iterations: 0
```

Explicit schemes (only the number of time steps is reported in the terminal):
```
lagsolve, number of time steps: 100
```

Ready for review, requires rebaseline of all integrated tests done in https://github.com/GEOSX/integratedTests/pull/237
Solves #1417


klevzoff (approved):
Nice!

CusiniM (approved):
This looks good. I would add to the stats also the number of configuration iterations if possible. It's going to be 0 for most cases but it is useful for contact solvers. It should not be hard since that loop is now directly done in `SolverBase`.

src/coreComponents/physicsSolvers/SolverStatistics.cpp
```diff
@@ -0,0 +1,117 @@
...
+    setApplyDefaultValue( 0.0 ).
+    setDescription( "Cumulative number of successful linear iterations" );
+
+  registerWrapper( viewKeyStruct::numFailedNonlinearIterationsString(), &m_numFailedNonlinearIterations ).
+    setApplyDefaultValue( 0.0 ).
+    setDescription( "Cumulative number of failed nonlinear iterations" );
+
+  registerWrapper( viewKeyStruct::numFailedLinearIterationsString(), &m_numFailedLinearIterations ).
+    setApplyDefaultValue( 0.0 ).
+    setDescription( "Cumulative number of failed linear iterations" );
```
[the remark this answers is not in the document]
francoishamon: Done, thanks for the catch

src/coreComponents/physicsSolvers/SolverStatistics.hpp
```diff
@@ -0,0 +1,159 @@
...
+
+  /**
+   * @brief Deleted assignment operator.
+   */
+  SolverStatistics & operator=( SolverStatistics const & ) = delete;
+
+  /**
+   * @brief Deleted move operator.
+   */
+  SolverStatistics & operator=( SolverStatistics && ) = delete;
```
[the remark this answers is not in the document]
francoishamon: That's good to know. I removed all of them in the latest commit.

src/coreComponents/physicsSolvers/SolverStatistics.cpp
```diff
@@ -0,0 +1,117 @@
...
+      GEOSX_LOG_RANK_0( getParent().getName() << ", number of successful linear iterations: " << m_numSuccessfulLinearIterations ); //
+    }
+
+    GEOSX_LOG_RANK_0( getParent().getName() << ", number of time step cuts: " << m_numTimeStepCuts );
+    GEOSX_LOG_RANK_0( getParent().getName() << ", number of failed nonlinear iterations: " << m_numFailedNonlinearIterations );
+    if( !isOuterLoopSolver ) // don't print for the outer iterations in sequential schemes
+    {
+      GEOSX_LOG_RANK_0( getParent().getName() << ", number of failed linear iterations: " << m_numFailedLinearIterations );
+    }
+  }
```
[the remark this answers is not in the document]
francoishamon: For now, I just implemented your first suggestion, since we are going to rely on parsing the log during the next weeks, until `TimeHistory` can be used to output these numbers. After that, I agree that we should a prettier log output for the stats.

francoishamon:
> This looks good. I would add to the stats also the number of configuration iterations if possible. It's going to be 0 for most cases but it is useful for contact solvers. It should not be hard since that loop is now directly done in `SolverBase`.

@CusiniM I added the number of successful/failed configuration iterations, but you should probably have a look to make sure it did it the way you wanted. To summarize:

Non-contact solvers:

They always report: `number of successful configurations iterations: 0`

Contact solvers:
- I start incrementing the number of configuration iterations [here](https://github.com/GEOSX/GEOSX/blob/070a73182037ac70826cffd663386475aa508bdb/src/coreComponents/physicsSolvers/SolverBase.cpp#L634). This means that the first called to `solveNonlinearSystem` does not count as a configuration iteration. Is it what you want?
- If the time step converges after multiple configuration iterations, all the Newton and linear iterations of the `nonlinearImplicitStep` are considered as `successful`. If the time step fails, all the Newton and linear iterations of the `nonlinearImplicitStep` are considered as `failed` (even though Newton's method converged in some of the configuration iterations).

It is a little bit difficult to explain, so I attached the log of the `ContactMechanics_simpleCubes_smoke` test that has time step cuts and multiple configuration iterations. It gives at the end of the log:
```
lagrangiancontact, number of time steps: 11                                                                                                                                              
lagrangiancontact, number of successful nonlinear iterations: 117                                                                                                                        
lagrangiancontact, number of successful configuration iterations: 15                                                                                                                     
lagrangiancontact, number of successful linear iterations: 117                                                                                                                           
lagrangiancontact, number of time step cuts: 1                                                                                                                                           
lagrangiancontact, number of failed nonlinear iterations: 17                                                                                                                             
lagrangiancontact, number of failed configuration iterations: 4                                                                                                                          
lagrangiancontact, number of failed linear iterations: 17  
```

[ContactMechanics_SimpleCubes_smoke.txt](https://github.com/GEOSX/GEOSX/files/9155874/ContactMechanics_SimpleCubes_smoke.txt)

src/coreComponents/physicsSolvers/SolverStatistics.hpp
```diff
@@ -0,0 +1,150 @@
...
+
+  /**
+   * @brief Struct to serve as a container for variable strings and keys.
+   * @struct viewKeyStruct
+   */
+  struct viewKeyStruct
+  {
+    /// String key for the number of time steps
+    static constexpr char const * numTimeStepsString() { return "numTimeSteps"; }
+    /// String key for the number of time steps
```
castelletto1, line 92: Update comment

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -844,7 +844,7 @@ void CompositionalMultiphaseBase::computeHydrostaticEquilibrium()
       localIndex const equilIndex = equilNameToEquilId.at( fs.getName() );
       real64 const minElevation = LvArray::math::min( globalMinElevation[equilIndex], datumElevation );
       real64 const maxElevation = LvArray::math::max( globalMaxElevation[equilIndex], datumElevation );
-      real64 const elevationIncrement = LvArray::math::min( fs.getElevationIncrement(), maxElevation - minElevation );
+      real64 const elevationIncrement = LvArray::math::min( fs.getElevationIncrement(), maxElevation - minElevation + 1e-12 );
```
castelletto1, line 847: Why `1e-12`?

castelletto1 (approved):
Thank you, this is very useful.

castelletto1:
> > This looks good. I would add to the stats also the number of configuration iterations if possible. It's going to be 0 for most cases but it is useful for contact solvers. It should not be hard since that loop is now directly done in `SolverBase`.
> 
> @CusiniM I added the number of successful/failed configuration iterations, but you should probably have a look to make sure it did it the way you wanted. To summarize:
> 
> Non-contact solvers:
> 
> They always report: `number of successful configurations iterations: 0`
> 
> ...

I have to confess that having statistics on `successful configurations` for non-contact solvers is not very nice. Could we be more general and formalize output for all solvers in terms of inner/outer nonlinear iterations?

CusiniM:
> > This looks good. I would add to the stats also the number of configuration iterations if possible. It's going to be 0 for most cases but it is useful for contact solvers. It should not be hard since that loop is now directly done in `SolverBase`.
> 
> @CusiniM I added the number of successful/failed configuration iterations, but you should probably have a look to make sure it did it the way you wanted. To summarize:
> 
> Non-contact solvers:
> 
> They always report: `number of successful configurations iterations: 0`
> 
> Contact solvers:
> 
>     * I start incrementing the number of configuration iterations [here](https://github.com/GEOSX/GEOSX/blob/070a73182037ac70826cffd663386475aa508bdb/src/coreComponents/physicsSolvers/SolverBase.cpp#L634). This means that the first called to `solveNonlinearSystem` does not count as a configuration iteration. Is it what you want?
> 
>     * If the time step converges after multiple configuration iterations, all the Newton and linear iterations of the `nonlinearImplicitStep` are considered as `successful`. If the time step fails, all the Newton and linear iterations of the `nonlinearImplicitStep` are considered as `failed` (even though Newton's method converged in some of the configuration iterations).
> 
> 
> It is a little bit difficult to explain, so I attached the log of the `ContactMechanics_simpleCubes_smoke` test that has time step cuts and multiple configuration iterations. It gives at the end of the log:
> 
> ```
> lagrangiancontact, number of time steps: 11                                                                                                                                              
> lagrangiancontact, number of successful nonlinear iterations: 117                                                                                                                        
> lagrangiancontact, number of successful configuration iterations: 15                                                                                                                     
> lagrangiancontact, number of successful linear iterations: 117                                                                                                                           
> lagrangiancontact, number of time step cuts: 1                                                                                                                                           
> lagrangiancontact, number of failed nonlinear iterations: 17                                                                                                                             
> lagrangiancontact, number of failed configuration iterations: 4                                                                                                                          
> lagrangiancontact, number of failed linear iterations: 17  
> ```
> 
> [ContactMechanics_SimpleCubes_smoke.txt](https://github.com/GEOSX/GEOSX/files/9155874/ContactMechanics_SimpleCubes_smoke.txt)



> > > This looks good. I would add to the stats also the number of configuration iterations if possible. It's going to be 0 for most cases but it is useful for contact solvers. It should not be hard since that loop is now directly done in `SolverBase`.
> > 
> > 
> > @CusiniM I added the number of successful/failed configuration iterations, but you should probably have a look to make sure it did it the way you wanted. To summarize:
> > Non-contact solvers:
> > They always report: `number of successful configurations iterations: 0`
> > ...
> 
> I have to confess that having statistics on `successful configurations` for non-contact solvers is not very nice. Could we be more general and formalize output for all solvers in terms of inner/outer nonlinear iterations?

So, 2 things:
- I think we can change the name to outer loop instead of configuration. Maybe, we can also just add an if statement that only outputs the configuration iterations info if the number of successful configuration iterations is different from 0 ? This way it will only show up for contact solvers.
- What about using `wasted` nonlinear and linear iterations instead of `failed`?

src/coreComponents/physicsSolvers/SolverStatistics.hpp
francoishamon: Done

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
francoishamon: I found a better way to avoid the division by zero at the next line by doing:
```
localIndex const numPointsInTable = ( elevationIncrement > 0 ) ? std::ceil( (maxElevation - minElevation) / elevationIncrement ) + 1 : 1;
```
so I removed the hard-coded constant.

francoishamon:
> So, 2 things:
> 
> * I think we can change the name to outer loop instead of configuration. Maybe, we can also just add an if statement that only outputs the configuration iterations info if the number of successful configuration iterations is different from 0 ? This way it will only show up for contact solvers.
> * What about using `wasted` nonlinear and linear iterations instead of `failed`?

I changed `failed` to `wasted` in the latest commit.

I also removed the output of configuration iterations (now called "outer-loop iterations") from the non-contact solvers using a simple `if` statement.

joshua-white:
Perhaps `discarded` sounds better than `wasted`?  They were necessary work, even if just to say the current timestep/configuration was incorrect.

castelletto1:
> Perhaps `discarded` sounds better than `wasted`? They were necessary work, even if just to say the current timestep/configuration was incorrect.

I agree that `discarded` may sound better, however `wasted` iterations is pretty standard terminology in reservoir simulation.

francoishamon:
> > Perhaps `discarded` sounds better than `wasted`? They were necessary work, even if just to say the current timestep/configuration was incorrect.
> 
> I agree that `discarded` may sound better, however `wasted` iterations is pretty standard terminology in reservoir simulation.

I replaced `wasted` with `discarded` to denote the nonlinear/linear iterations that took place before time step cuts

PR #1983, opened 2022-07-22
Improved parsing speed for table files

klevzoff:
This PR adds optimized functions for reading and parsing numerical values from files, which improves initialization time for `TableFunction` objects with large tables.

The specific case I have that motivated this contains a high-resolution heterogeneous permeability field with some 22M linefeed-separated values in the voxel file (~250 MB on disk). It used to take 13s to just read the file, adding to the already slow startup. With this PR reading the table takes 0.67s, for ~20x improvement (on an ext4 SSD; results may vary depending on hardware and filesystem).

The speedup comes from:
1. Replacing line-by-line reading with a fixed-size 16kb buffer (size chosen empirically based on best performance)
2. Replacing formatted stream extraction operators (`>>`) with a high-performance floating point parser from [fast_float](https://github.com/fastfloat/fast_float) library

I chose to integrate the library in-source (under `thirdparty`, similar to `optionparser`) since it is:
  * small, header-only and has a single-header version
  * not available though spack
  * not likely to need version updates
  * Apache 2.0 licensed, which I think is LGPL compatible (and we already use Apache-licensed libraries anyway, e.g. METIS)

At the moment only `TableFunction` is using the new parsing functions, but I've put them into a separate file in case they're needed elsewhere and added some unit testing (which helped catch a few bugs).


cssherman:
Interesting!

When I've thought about handling large tables in the past, I've also considered adding an option to read gzipped files or an separate hdf5 table type.  Both of these would be pretty straightforward to implement, and would be easy to build.  It's hard to predict, but the IO increase from the gzip method could be neutral to positive depending on the disk type.  Thoughts @klevzoff ?

klevzoff:
> When I've thought about handling large tables in the past, I've also considered adding an option to read gzipped files or an separate hdf5 table type. Both of these would be pretty straightforward to implement, and would be easy to build. It's hard to predict, but the IO increase from the gzip method could be neutral to positive depending on the disk type. Thoughts @klevzoff ?

My first thought was to add support for reading simple binary files which would be blazingly fast, even faster if we compressed with something like [fpzip](https://computing.llnl.gov/projects/fpzip). On the other hand, writing and maintaining binary files may a bit of a hassle, depending on data source/tool. I decided not to pursue this after realizing I can get acceptable speed with text files. `hdf5` is a nice alternative to plain binary (and lets you combine coordinate and voxel arrays in a single file, which is convenient).

`gzip` might work well, but I have no idea about how well textual FP representation compresses (if that's what you had in mind). We'd still have to parse the text upon decompression without it becoming a bottleneck, so hopefully parts of this PR would prove useful for that.

Luckily we don't have to decide one way or the other as these options are all additive. Personally I don't think I will have cases larger than ~100M cells in the near future, but I'm curious to see what people's experiences/requirements are.

rrsettgast (commented):
Add the `fast_float` version information in a separate file as you suggested when we discussed this last?

rrsettgast (approved):

PR #1987, opened 2022-07-25
Feature/ascuotto/vtk import

aure-lily:
Added .vts and .pvts files import. The .vts file import uses 3.4 less memory than .vtu file.


src/coreComponents/mesh/generators/VTKMeshGenerator.hpp
```diff
@@ -32,6 +32,8 @@
 #include <unordered_map>
 
 class vtkUnstructuredGrid;
+// class vtkImageData;
```
sframba, line 35: same (commented line to be removed)

src/coreComponents/unitTests/meshTests/testVTKImport.cpp
```diff
@@ -91,60 +91,99 @@ TEST( VTKImport, cube )
...
     // The "2" set are all the boundary nodes (64 - 8 inside nodes = 56),
     // minus an extra node that belongs to regions -1 and 9 only.
-    SortedArray< localIndex > const & nodesRegion2 = cellBlockManager.getNodeSets().at( "2" );
-    ASSERT_EQ( nodesRegion2.size(), expected( 55, { 39, 27 } ) );
-
-    // Region "9" has only one quad, on the greater `x` direction.
-    // This hex will belong to MPI rank 1.
-    SortedArray< localIndex > const & nodesRegion9 = cellBlockManager.getNodeSets().at( "9" );
-    ASSERT_EQ( nodesRegion9.size(), expected( 4, { 0, 4 } ) );
+    std::cout << "> NodeSets : \n";
```
sframba, line 103: the "`cout`" are just for debugging purposes? If so, please remove them, else they will pollute the unit test output

src/coreComponents/mesh/generators/VTKMeshGenerator.cpp
```diff
@@ -150,56 +176,100 @@ loadMesh( Path const & filePath )
...
+    for( int c = 0; c < numCell; c++ )
+    {
+      cells->InsertNextCell( mesh.GetCell( c ));
+    }
+  }
+  else
+  {
+    double bounds[6];
+    mesh.GetBounds( bounds );
+    double xmin=bounds[0], xmax=bounds[1];
```
sframba, line 194: `zmin` and `zmax` are `double`s or are they always integers? If they are always integers, please declare them as integers and round and cast the values of `bounds` (see comment below). This would make the loops at line 201 more readable.

src/coreComponents/mesh/generators/VTKMeshGenerator.cpp
```diff
@@ -150,56 +176,100 @@ loadMesh( Path const & filePath )
...
+    double ymin=bounds[2], ymax=bounds[3];
+    double zmin=bounds[4], zmax=bounds[5];
+    int dx = 1, dy = 1;
+    int nx = (xmax-xmin)/dx +1, ny = (ymax-ymin)/dy +1;
+    //int nz = (zmax-zmin)/dz +1; // not used
+
+    for( int k=zmin; k<zmax; k++ )
```
sframba, line 201: This loop looks weird--why does the indexing begin with `xmin`, `ymin `and `zmin`? Aren't they the "geographical" bounds of your system, and not integers? Are they given in meters? Or are they just integers?

src/coreComponents/mesh/generators/VTKMeshGenerator.cpp
```diff
@@ -592,6 +663,7 @@ std::vector< int > getGeosxToVtkNodeOrdering( ElementType const elemType )
     case ElementType::Prism5:        return { 0, 1, 2, 3, 4, 5, 6, 7, 8, 9 };
     case ElementType::Prism6:        return { 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11 };
     case ElementType::Polyhedron:    return { 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12 }; // TODO
+      // case ElementType::Voxel:         return { 0, 1, 2, 3, 4, 5, 6, 7}; // For ImageData
```
sframba, line 666: remove

src/coreComponents/mesh/generators/VTKMeshGenerator.cpp
```diff
@@ -548,17 +619,17 @@ void extendCellMapWithRemoteKeys( VTKMeshGenerator::CellMapType & cellMap )
   }
 }
 
+using CellMapType = std::map< ElementType, std::unordered_map< int, std::vector< vtkIdType > > >;
 /**
  * @brief Collect lists of VTK cell indices organized by type and attribute value.
- * @param[in] mesh the vtkUnstructuredGrid that is loaded
+ * @param[in] mesh the vtkUnstructuredGrid or vtkImageData that is loaded
```
sframba, line 625: vtkStructuredGrid

src/coreComponents/mesh/generators/VTKMeshGenerator.cpp
```diff
@@ -600,10 +672,10 @@ std::vector< int > getGeosxToVtkNodeOrdering( ElementType const elemType )
  * @brief Fill @p cellBlock with the appropriate nodes and local/global mappings.
  * @param[in] elemType the vtk cell type for cells of the CellBlock being written
  * @param[in] cellIds the cell indexes of cell type \p cellType within this region
- * @param[in] mesh the vtkUnstructuredGrid that is loaded
+ * @param[in] mesh the vtkUnstructuredGrid or vtkImageData that is loaded
```
sframba, line 675: same

src/coreComponents/mesh/generators/VTKMeshGenerator.cpp
```diff
@@ -646,6 +720,7 @@ string getElementTypeName( ElementType const type )
     case ElementType::Prism5:      return "pentagonalPrisms";
     case ElementType::Prism6:      return "hexagonalPrisms";
     case ElementType::Polyhedron:  return "polyhedra";
+    //case ElementType::Voxel:       return "voxels";
```
sframba, line 723: remove

src/coreComponents/mesh/generators/VTKMeshGenerator.cpp
```diff
@@ -771,7 +846,7 @@ void importRegularField( std::vector< vtkIdType > const & cellIds,
   } );
 }
 
-void printMeshStatistics( vtkUnstructuredGrid & mesh,
+void printMeshStatistics( vtkDataSet & mesh,                     // vtkUnstructuredGrid
```
sframba, line 849: remove comment

src/coreComponents/mesh/generators/VTKMeshGenerator.cpp
```diff
@@ -1041,7 +1116,7 @@ void VTKMeshGenerator::writeCells( CellBlockManager & cellBlockManager ) const
 
 /**
  * @brief Build the "surface" node sets from the surface information.
- * @param[in] mesh The vtkUnstructuredGrid that is loaded
+ * @param[in] mesh The vtkUnstructuredGrid or vtkImageData that is loaded
```
sframba, line 1119: same

src/coreComponents/mesh/generators/VTKMeshGenerator.cpp
```diff
@@ -150,56 +176,100 @@ loadMesh( Path const & filePath )
...
 template< typename INDEX_TYPE >
 ArrayOfArrays< INDEX_TYPE, INDEX_TYPE >
-buildElemToNodes( vtkUnstructuredGrid & mesh )
+buildElemToNodes( vtkDataSet & mesh )
 {
   // According to VTK docs, IsStorageShareable() indicates whether pointers extracted via
   // vtkCellArray::GetCellAtId() are pointers into internal storage rather than temp buffer
   // and thus results can be used in a thread-safe way.
-  return mesh.GetCells()->IsStorageShareable()
+  return GetCellArray( mesh )->IsStorageShareable()                        // changed mesh.GetCells() to GetCellArray
```
sframba, line 265: Careful, here you are building again the list of cells, this can be costly. I suggest creating the list here (`cells= GetCellArray( mesh )`), and then passing `cells` to `buildElemToNodesImpl` as an argument. This requires changing the signature of the method.

sframba (approved):

src/coreComponents/mesh/generators/VTKMeshGenerator.cpp
```diff
@@ -150,18 +181,32 @@ loadMesh( Path const & filePath )
...
+  if( mesh.IsA( "vtkStructuredGrid" ) || mesh.IsA( "vtkUnstructuredGrid" ))
+  {
+    int numCell = mesh.GetNumberOfCells();
+    for( int c = 0; c < numCell; c++ )
+    {
+      cells->InsertNextCell( mesh.GetCell( c ));
+    }
+  }
+  return cells;
+}
```
klevzoff, line 196: It would be nice to avoid constructing a new `vtkCellArray` when the grid is unstructured and already contains one. I suggest the following:
  * change return type to `vtkSmartPointer< vtkCellArray >`
  * have a branch that is executed when `mesh.IsA( "vtkUnstructuredGrid" )` is true and simply returns `vtkUnstructuredGrid::SafeDownCast( &mesh )->GetCells()`
  * do what you're already doing for all other types of input datasets.

src/coreComponents/mesh/generators/VTKMeshGenerator.cpp
```diff
@@ -265,7 +311,7 @@ redistributeByKdTree( vtkUnstructuredGrid & mesh )
   rdsf->SetInputDataObject( &mesh );
   rdsf->SetNumberOfPartitions( MpiWrapper::commSize() );
   rdsf->Update();
-  return vtkUnstructuredGrid::SafeDownCast( rdsf->GetOutputDataObject( 0 ) );
+  return vtkDataSet::SafeDownCast( rdsf->GetOutputDataObject( 0 ) );
```
klevzoff, line 314: FYI (though you are probably already aware of this) `vtkRedistributeDataSetFilter` always outputs a `vtkUnstructuredGrid`, regardless of the type of input dataset. Moreover, the following ParMETIS-based redistribution step is only suitable for unstructured grids. It'll still work here because of the aforementioned conversion, but the results will likely be suboptimal for structured inputs. At the very least I suggest setting `partitionRefinement="0"` in your input files to disable the ParMETIS step. 

Further down the road I would propose avoiding `vtkRedistributeDataSetFilter` altogether, implementing a different partitioning strategy (maybe based on `-x`/`-y`/`-z` flags similar to `InternalMeshGenerator`) and preserving the dataset as `vtkStructuredGrid` until the very end when we have to populate the GEOSX cell blocks. This would probably speed up the loading and reduce memory footprint even further.

src/coreComponents/mesh/generators/VTKMeshGenerator.cpp
```diff
@@ -150,18 +181,32 @@ loadMesh( Path const & filePath )
...
+    }
+  }
+  return cells;
+}
+
 template< typename INDEX_TYPE, typename POLICY >
 ArrayOfArrays< INDEX_TYPE, INDEX_TYPE >
-buildElemToNodesImpl( vtkUnstructuredGrid & mesh )
+buildElemToNodesImpl( vtkDataSet & mesh,
+                      const vtkNew< vtkCellArray > & cells )
```
klevzoff, line 201: We use east-const convention throughout the code, would you mind making sure the changes in this file are consistent with it?
```suggestion
                      vtkNew< vtkCellArray > const & cells )
```

klevzoff (approved):
Nice work! Just a couple additional comments and thoughts (sorry to drop in randomly on a PR like that)

src/coreComponents/mesh/generators/VTKMeshGenerator.cpp
aure-lily: I changed the function with your suggestions in commit 7b16966

acitrain (approved):

PR #1988, opened 2022-07-25
Move gcc 10 debug job back to Travis

bmhan12:
The gcc 10 debug job is failing on Azure Pipelines as its hitting the [10GB space limit ](https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml&viewFallbackFrom=vsts#capabilities-and-limitations) on Microsoft's hosted agents.

Moving the job back to Travis ([limit is 50 GB](https://docs.travis-ci.com/user/reference/overview/#virtualisation-environment-vs-operating-system)).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -176,5 +176,11 @@ jobs:
   - stage: builds
     name: Mac_OSX
     <<: *geosx_osx_build
+  - stage: builds
+    name: Ubuntu debug (20.04, gcc 10.3.0, open-mpi 4.0.3)
+    script: scripts/ci_build_and_test.sh
+    env:
+    - DOCKER_REPOSITORY=geosx/ubuntu20.04-gcc10
+    - CMAKE_BUILD_TYPE=Debug
   - stage: return_status
     <<: *return_script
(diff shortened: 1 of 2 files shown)


jhuang2601 (approved):
Thanks for looking into this issue!

TotoGaz:
Can't we just pay Microsoft for more disk space?

bmhan12:
> Can't we just pay Microsoft for more disk space?

I considered that, but from what they have noted here (https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml#hardware), they don't plan on accepting those requests.

jhuang2601:
> > Can't we just pay Microsoft for more disk space?
> 
> I considered that, but from what they have noted here (https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml#hardware), they don't plan on accepting those requests.

They are not doing the business wisely.

rrsettgast:
Can we look at lightening our images? @TotoGaz @bmhan12 ?

klevzoff:
> Can we look at lightening our images? @TotoGaz @bmhan12 ?

The point of having a debug build in CI is to run tests with assertions enabled. Perhaps we can replace a full-on debug build with one that has a low optimization level and not defining `NDEBUG` but no actual debug information (i.e. no `-g` flag).

TotoGaz:
> Can we look at lightening our images? @TotoGaz @bmhan12 ?

Most of our `tpl` images should already be as light as possible.
Specially the `ubuntu` ones.

We could try to remove the temporaries when building `geosx` though.
Some kind of `make clean` after `make install` if that's possible.

Issue #1970, opened 2022-07-08
[Docs] Python docsting format, sphinx autodocumentation


cssherman:
**What needs to be documented?**
The format of the docstrings throughout the code is inconsistent.  I would recommend using the [google style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html).  Once this is complete, we can auto-generate python API's to place in the sphinx documentation:

```

Data Managers API
==========================

.. automodule:: orion.managers.manager_base
    :members:

.. automodule:: orion.managers.orion_manager
    :members:
```

cssherman:
Any thoughts @klevzoff @TotoGaz ?

TotoGaz:
Having a clearly defined style is nice.
Do we have a way to impose the google style in IDE?

PR #1991, opened 2022-07-29
Add python tools documentation to sphinx

cssherman:


cssherman:
Addresses #1970

cssherman:
I've triggered a doc build for this branch here:

https://geosx-geosx.readthedocs-hosted.com/en/docs-sherman-updatepythondocstrings/docs/sphinx/pythonTools/pythonAPI.html

Python docstrings that don't conform to the expected google style will often show up as warnings/errors in the documentation build log.  To generate the autodocs, sphinx needs to load the targeted modules.  To avoid issues with missing packages in the RTD build, I'm using the `autodoc_mock_imports` flag in the conf.py file.

host-configs/LLNL/quartz-base.cmake
```diff
@@ -33,4 +33,7 @@ set(ENABLE_PETSC OFF CACHE BOOL "Enables PETSc." FORCE)
 # YAPF python formatting
 set(YAPF_EXECUTABLE /usr/gapps/GEOSX/thirdPartyLibs/python/linux-rhel7-x86_64-gcc@8.1.0/python@3.8.5/bin/yapf CACHE PATH "" FORCE)
 
+# Sphinx
+set(SPHINX_EXECUTABLE /usr/gapps/GEOSX/thirdPartyLibs/python/documentation/documentation/bin/sphinx-build CACHE PATH "" FORCE)
```
TotoGaz, line 37: The sphinx executable is really called `sphinx-build` and not `sphinx`?

src/conf.py
```diff
@@ -16,6 +16,16 @@
 import sys
 import shutil
 
+# Add python modules to be documented
+python_root = './coreComponents/python/modules'
+python_modules = ['geosx_mesh_tools_package',
```
TotoGaz, line 21: BTW, you can use `tuples` instead of list if you do not want the list to be extended.

src/coreComponents/python/modules/geosx_mesh_tools_package/geosx_mesh_tools/abaqus_converter.py
```diff
@@ -8,13 +8,16 @@
...
+    Convert an abaqus mesh to gmsh 2 format, preserving nodeset information.
+
+    If the code encounters any issues with region/element indices,
+    the conversion will attempt to continue, with errors
+    indicated by -1 values in the output file.
+
+    Args:
+        input_mesh (str): path of the input abaqus file
+        output_mesh (str): path of the output gmsh file
+        logger (logging.Logger): an instance of logging.Logger
```
TotoGaz, line 20: So could se star using type hints?

TotoGaz (approved):

src/coreComponents/python/modules/geosx_mesh_tools_package/geosx_mesh_tools/abaqus_converter.py
cssherman: Sounds good to me.  With this change, do we want to use the backwards-compatible syntax, or the in-place syntax?
https://peps.python.org/pep-0484/#suggested-syntax-for-python-2-7-and-straddling-code

host-configs/LLNL/quartz-base.cmake
cssherman: Yeah, I was surprised when I came across this too.  Sphinx builds four executables: sphinx-apidoc, sphinx-autogen, sphinx-build, sphinx-quickstart

src/coreComponents/python/modules/geosx_mesh_tools_package/geosx_mesh_tools/abaqus_converter.py
cssherman: I don't think that we do at this point

TotoGaz: Maybe it's the first move then 🤣 ?

src/coreComponents/python/modules/geosx_xml_tools_package/geosx_xml_tools/main.py
```diff
@@ -5,9 +5,22 @@
 import os
 import time
 from geosx_xml_tools import xml_processor
+from typing import Callable, Any, List, Union, Tuple
```
cssherman, line 8: @TotoGaz - I've tested implementing python type hints in this file.  Any thoughts?  I did catch a couple of small errors in the docstring types when I was doing this, so it is certainly a useful tool.

TotoGaz (commented):
Do you also have a tool that's able to validate that we are doing nothing wrong with those hints? (Some static compiler, in some sense!)

cssherman:
> Do you also have a tool that's able to validate that we are doing nothing wrong with those hints? (Some static compiler, in some sense!)

Yes, I'm using the SublimeLinter mypy plugin to validate the results on the fly.  It was easy to setup, and I can add some pointers on other tools

src/coreComponents/python/modules/geosx_xml_tools_package/geosx_xml_tools/main.py
```diff
@@ -132,8 +172,16 @@ def preprocess_parallel():
     return format_geosx_arguments(compiled_name, unknown_args)
 
 
-def format_geosx_arguments(compiled_name, unknown_args):
-    # Return GEOSX arguments
+def format_geosx_arguments(compiled_name: str, unknown_args: List[str]) -> List[str]:
```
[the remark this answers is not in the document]
cssherman: Agreed (I need to re-wire my default python typing for things like this!).  In this case though, we are passing through one of the results of argparse.parse_know_args, which is a list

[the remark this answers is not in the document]
TotoGaz: At least `unknown_args` could be defined as a `Iterable[str]`, couldn't it?
I usually tend to give minimal contracts (here it's a bit nitpicking 🤣!)

Issue #841, opened 2020-03-18
MPI issue using Surface Generator


andrea-franceschini:
**Bug description**
Running this [test case](https://github.com/GEOSX/GEOSX/files/4346294/test_crack.txt), that needs this [mesh file](https://github.com/GEOSX/GEOSX/files/4346298/crack01.txt), with more than 1 process, e.g. `mpirun -np 2 geosx -x 2`, the code waits forever. If I interrupt it with Ctrl+C, I have:

```
Frame 1: cxx_utilities::handler(int, int, int)
Frame 2: 
Frame 3: opal_progress
Frame 4: ompi_request_default_wait
Frame 5: ompi_coll_base_sendrecv_actual
Frame 6: ompi_coll_base_allgather_intra_two_procs
Frame 7: MPI_Allgather
Frame 8: void geosx::MpiWrapper::allGather<long>(long, LvArray::Array<long, 1, camp::int_seq<long, 0l>, long, LvArray::NewChaiBuffer>&, ompi_communicator_t*)
Frame 9: geosx::CommunicationTools::AssignNewGlobalIndices(geosx::ObjectManagerBase&, std::set<long, std::less<long>, std::allocator<long> > const&)
Frame 10: geosx::SurfaceGenerator::SeparationDriver(geosx::DomainPartition*, geosx::MeshLevel*, std::vector<geosx::NeighborCommunicator, std::allocator<geosx::NeighborCommunicator> >&, int, int, bool, double)
Frame 11: geosx::SurfaceGenerator::SolverStep(double const&, double const&, int, geosx::DomainPartition*)
Frame 12: geosx::SurfaceGenerator::Execute(double, double, int, int, double, geosx::dataRepository::Group*)
Frame 13: geosx::EventBase::Execute(double, double, int, int, double, geosx::dataRepository::Group*)
Frame 14: geosx::EventManager::Run(geosx::dataRepository::Group*)
Frame 15: geosx::ProblemManager::RunSimulation()
Frame 16: main
Frame 17: __libc_start_main
Frame 18: _start
```

As now, I'm using the branch #799, but the `surfaceGenerator` kernel is the same as in `develop`.

**Platform:**
- Machine: Ubuntu 18.04
- Compiler: gcc 7.4.0
- Cmake: 3.10.2

**Note**
The extensions are: `xml` for the main input and `msh` for the mesh. `github` forced me to use `txt`.

cssherman:
[...]

PR #1992, opened 2022-07-29
Replacing PAMELA mesh with VTK mesh

castelletto1:
This PR

- replaces all `msh` meshes by either `vtk` (using ascii format) or `vtu` (using binary format) meshes in [`inputFiles`](https://github.com/GEOSX/GEOSX/tree/develop/inputFiles) folder depending on the model size. Specifically, I kept legacy `vtk` and ASCII for small size problems (less than 1,000 cells) for readability purposes.
- replaces the Egg model `msh` mesh by `vtu` mesh (using binary format) in GEOSXDATA repo

Remaining task/issue:
- [x] updating [tutorial 2](https://geosx-geosx.readthedocs-hosted.com/en/latest/docs/sphinx/tutorials/step02/Tutorial.html) (see issue #2015)
- [x] updating [tutorial 3](https://geosx-geosx.readthedocs-hosted.com/en/latest/docs/sphinx/tutorials/step03/Tutorial.html) in Read the Docs, which will have effects on the [Meshes](https://geosx-geosx.readthedocs-hosted.com/en/latest/coreComponents/mesh/docs/Mesh.html#meshes) documentation in the user Guide (see issue #2015)
- [x] the only parallel integrated test using a mesh that is not internally generated, i.e. `ContactMechanics_SingleFracCompression_03`, produces a `FAIL RUN` because of ghosting problems for fracture elements (see issue #2014)

I don't think we want to keep the PAMELA tests located [here](https://github.com/GEOSX/GEOSX/tree/develop/inputFiles/singlePhaseFlow/pamela_test). @herve-gross please confirm. 

Numerical diffs are introduced due to the different ordering used by PAMELA and VTK mesh generators (visual inspection  confirms that solutions are the same).

GEOSXDATA updating  done in: https://github.com/GEOSX/GEOSXDATA/pull/11
Rebaseline done in https://github.com/GEOSX/integratedTests/pull/244


joshua-white (approved):
Looks really great to me.  Nice cleanup.

inputFiles/lagrangianContactMechanics/ContactMechanics_PassingCrack_base.xml
```diff
@@ -138,6 +138,13 @@
       scale="0.0"
       setNames="{ rear }"/>
   </FieldSpecifications>
+ 
+  <Functions>
+    <TableFunction
+      name="apertureTable"
+      coordinates="{ -1.0e-3, 0.0 }"
+      values="{ 0.0, 0.0 }"/>
+  </Functions>
```
TotoGaz, line 147: This is not related to your PAMELA/VTK fix, is it?

inputFiles/phaseField/PhaseFieldFracture_SpectralSplit.xml
```diff
@@ -48,21 +48,33 @@
...
+    <Box
+      name="left"
+      xMin="{ -0.01, -0.01, -0.01 }"
+      xMax="{  0.01,  1.01,  0.11 }"/>
+
+    <Box
+      name="right"
+      xMin="{  0.99, -0.01, -0.01 }"
+      xMax="{  1.01, 1.01,  0.11 }"/>
+  </Geometry>
```
TotoGaz, line 76: Why do you have to add this?

inputFiles/phaseField/PhaseFieldFracture_VolDevSplit.xml
```diff
@@ -48,11 +48,33 @@
...
+    <Box
+      name="left"
+      xMin="{ -0.01, -0.01, -0.01 }"
+      xMax="{  0.01,  1.01,  0.11 }"/>
+
+    <Box
+      name="right"
+      xMin="{  0.99, -0.01, -0.01 }"
+      xMax="{  1.01, 1.01,  0.11 }"/>
+  </Geometry>
```
TotoGaz, line 76: Same question

TotoGaz (approved):
So what about the PAMELA code?
When are we going to stop linking to it?

> the only parallel integrated test using a mesh that is not internally generated, i.e. ContactMechanics_SingleFracCompression_03, produces a FAIL RUN because of ghosting problems for fracture elements.

The tests are supposed to protect the features of `GEOSX`.
Unfortunately, this feature in parallel is notoriously buggy and very hard to fix 🤷 😢 (it would require a dedicated person at least and I'm afraid we do not have the resources right now).
If you cannot make it work at minimal cost, I'd say we can suspend this test as it's not the protector of any reliable feature.

> I don't think we want to keep the PAMELA tests located [here](https://github.com/GEOSX/GEOSX/tree/develop/inputFiles/singlePhaseFlow/pamela_test).

IMHO eventually, this should disappear too.

inputFiles/lagrangianContactMechanics/ContactMechanics_PassingCrack_base.xml
castelletto1: Yes, this is unrelated. `ContactMechanics_PassingCrack` was removed from the integrated tests. As a result, the `xml` file was not updated to ensure that attribute `apertureTableName` was not empty. I decided to modify to make sure it can run.

As a side comment, specifying an aperture makes sense when contact mechanics is coupled with fluid flow. For pure contact mechanics problem this field is irrelevant and should not be a required input.

inputFiles/phaseField/PhaseFieldFracture_SpectralSplit.xml
castelletto1: `top`, `bottom`, `left` and `right` were attributes associated to faces in the original `msh` file. In the `vtk ` mesh I got rid of cell attributes and simply kept three-dimensional cells with the above boxes introduced to identify boundary nodes.

inputFiles/phaseField/PhaseFieldFracture_VolDevSplit.xml
castelletto1: See answer above.

castelletto1:
> The tests are supposed to protect the features of `GEOSX`.
> Unfortunately, this feature in parallel is notoriously buggy and very hard to fix 🤷 😢 (it would require a dedicated person at least and I'm afraid we do not have the resources right now).
> If you cannot make it work at minimal cost, I'd say we can suspend this test as it's not the protector of any reliable feature.

OK, I propose we disable the parallel test and open an issue to remember to reactive it, most likely when issue #841 will be resolved.

TotoGaz:
> OK, I propose we disable the parallel test and open an issue to remember to reactive it, most likely when issue #841 will be resolved.

I'm fine with this. Still require additional mgmt validation though.

src/docs/sphinx/advancedExamples/validationStudies/faultMechanics/singleFracCompression/Example.rst
```diff
@@ -76,7 +76,7 @@ The following figure shows the mesh used in this problem.
 
 Here, we load the mesh with ``PAMELAMesh`` (see :ref:`ImportingExternalMesh`).
 The syntax to import external meshes is simple: in the XML file,
```
jhuang2601, line 78:
```suggestion
Here, we load the mesh with ``VTKMesh``.
The syntax to import external meshes is simple: in the XML file,
```

jhuang2601 (approved):

PR #1995, opened 2022-08-02
Bugfix/mbader/wave propagation source injection mpi

nmbader:
Fix of 2 bugs with the source injection in the case of second order acoustic solver.
- Bug1 description: when the source is located in a cell that is local to a given MPI rank (say rank_0) but is ghost to one or more other MPI ranks (say rank_i), it will be injected at all nodes belonging to that cell within rank_0 but will have zero weights for the other rank_i. As a result, if some of the cell nodes are actually owned by a rank_i, after field synchronization, those nodes will have zero weights (incorrectly). This can be manifested in dimming the pressure field amplitudes, thus the modeled seismogram amplitudes (see attached figure).
- Bug1 fix: allow the source weights to have valid values (non-zero) whenever it is accessible by an MPI rank (so when either it is local or is a ghost cell).
- Bug2 description: when adding the source term to the RHS vector, a parallel loop over simultaneous sources is run, but the incremental addition to the RHS is done through a simple "+=" operator. This has a potential race condition when different sources contribute to the same node location.
- Bug2 fix: replace "+=" with atomicAdd.

Rebaseline done in https://github.com/GEOSX/integratedTests/pull/242

MPI_bug


francoishamon (approved):
Thanks for fixing the bug, @nmbader. The PR looks good. It will require a rebaseline because the PR changes a field name, and because the bug actually appears in the integrated tests.

sframba (approved):

PR #1996, opened 2022-08-03
Add PT-Scotch as a partitioning method option in VTKMesh

klevzoff:
This PR adds [PT-Scotch](https://gitlab.inria.fr/scotch/scotch) as an alternative mesh partitioning method to ParMETIS when importing a `VTKMesh`. We already had the library in out TPLs (as a dependency of PETSc), with https://github.com/GEOSX/thirdPartyLibs/pull/192 separating it into a standalone TPL. This PR provides an interface to the library and an input option to activate it.

I've had some problems with ParMETIS sometimes generating non-contiguous partitions (with one or a few cells not connected to the rest through a face). Also it is interesting to compare the performance/quality tradeoff of both libraries. I'm a little unhappy lately about the time it takes to redistribute semi-large meshes (~50-100M cells), but I've yet to investigate the bottlenecks. At least in theory, PT-Scotch is able to use multiple cores in contrast to ParMETIS. That said, it may turn out that the bottleneck is in ParMETIS's `MeshToDual` function, which is currently used to prepare input for both libraries. In that case I have other ideas to try.

This PR is ready for review. It will require a rebaseline of all VTK-based tests due to added input (`partitionMethod`). I have not considered switching any of those tests to actually use PT-Scotch, but it's an option.

Rebaseline PR: https://github.com/GEOSX/integratedTests/pull/241 (merged)
LvArray PR: https://github.com/GEOSX/LvArray/pull/263 (merged)

TODO:
  - [x] generate/commit schema + rst
  - [x] update spack package config
  - [x] run/rebaseline integrated tests
  - [x] ask someone to rebuild TPLs on LC and update host-configs


src/coreComponents/mesh/generators/VTKMeshGenerator.cpp
```diff
@@ -246,15 +253,49 @@ splitMeshByPartition( vtkUnstructuredGrid & mesh,
...
+    {
+      case VTKMeshGenerator::PartitionMethod::parmetis:
+      {
+        return parmetis::partition( graph.toViewConst(), elemDist, numProcs, comm, numRefinements );
+      }
+      case VTKMeshGenerator::PartitionMethod::ptscotch:
+      {
+#ifdef GEOSX_USE_SCOTCH
+        return ptscotch::partition( graph.toViewConst(), numProcs, comm );
+#else
```
francoishamon, line 289: Is there an equivalent of `ParMETIS_V3_RefineKway` in PT-Scotch? If not, maybe we can add a warning if someone tries to use a non-zero `numRefinements` with PT-Scotch.

francoishamon (approved):
Looks great. Once you have reached some conclusions on the partitioning quality of ParMetis vs PT-Scotch, please let us know. I am very interested in your findings.

src/coreComponents/mesh/generators/VTKMeshGenerator.cpp
klevzoff: No, I didn't find anything comparable in the user manual. My plan was to ignore the refinement level in that case, but I agree a warning is a good idea. In future we might add more options that are specific to either method and I hope to refactor the inputs a bit.

rrsettgast (approved):

rrsettgast:
@klevzoff Don't forget to update the LvArray hash.

PR #1998, opened 2022-08-03
Unified residual norm calculations in flow/well solvers

### LINKED ISSUE #1936: Nonlinear residuals should not be mesh-dependent
**Problem**
In single-phase flow solvers (FVM, HybridFVM, and by extension, single-phase poromechanics), the nonlinear residual is dependent on mesh resolution. E.g. for a particular problem (fixed-size cube with left-to-right pressure drop prescribed) we observe the following initial residuals:
| mesh        | R        |
| ----------- | -------- |
| 35x35x35    | 5.32e+02 |
| 70x70x70    | 4.26e+03 |
| 140x140x140 | 3.40e+04 |
| 280x280x280 | 2.72e+05 |

As you can see, each one is exactly 8 times larger that the previous. This poses a significant challenge when designing mesh refinement studies, as nonlinear tolerance has to be adapted to mesh resolution.

**Proposed solution**
We should make flow residuals independent of mesh resolution (as is already the case for mechanics and multiphase flow). Dependence comes from dividing the normalizer by the total cell count:
https://github.com/GEOSX/GEOSX/blob/11cbe7d403087bd341f0986598269f6bb5d67bb4/src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseFVM.cpp#L161
which effectively ends up multiplying the residual. It't not clear to me what the purpose of that division is, at least in the context of normal porous media flow. We end up dividing a *mesh-total* quantity (L2-norm of real mass balance residual) by a *cell-average* quantity (average mass per cell), which doesn't make sense to me. Is it something that was added for hydrofracturing problems? What is the rationale? I recall us having a discussion about this residual evaluation a long time ago, but I can't find a written record of it.

My proposal is to remove `globalResidualNorm[2]` (cell count) altogether.

@rrsettgast @francoishamon @CusiniM @castelletto1 and anyone else, objections?

francoishamon:
This PR unifies the calculation of the residual norms in the flow and well solvers. 

It works using the following approach:
- In `SolverBaseKernels`, the class `ResidualNormKernelBase` defines two kernels: one for the $L^\infty$ norm and one for the $L^2$ norm. In these two kernels, the `compute` functions are pure virtual and have to be implemented in derived classes.
- In the kernel file of each flow solver, the class `ResidualNormKernel` implements the two `compute` functions to define how the $L^\infty$ norm and the $L^2$ norm should be computed for each physics.
- Each flow solver has a new xml attribute `normType` that the user can use to choose between the $L^\infty$ norm (default) and the $L^2$ norm.

In this PR, the $L^\infty$ norm is computed as:
$|| R ||_{\infty} = max_i \bigg( \displaystyle \frac{|R_i|}{(\rho_t)_i \phi_i V_i} \bigg)$
and the $L^2$ norm is computed as:
$|| R ||_2 = \displaystyle \frac{ \sqrt{ \sum_i R_i \times R_i }}{ \sqrt{\sum_i (\rho_t)_i \phi_i V_i}}$

- [X] SinglePhaseFVM
- [X] ThermalSinglePhaseFVM
- [X] SinglePhaseHybridFVM
- [X] SinglePhaseWell
- [X] CompositionalMultiphaseFVM
- [X] ThermalCompositionalMultiphaseFVM
- [X] CompositionalMultiphaseHybridFVM
- [X] CompositionalMultiphaseWell
- [X] ProppantTransport

The PR works for all the integrated tests but will probably introduce some small diffs (particularly, $L^2$ is now very different from what is was in `SinglePhaseFVM` for large meshes because I removed the multiplier).

The PR is ready for review. There are many small things that may make it difficult to review, like:
- Removed `m_inputFluxEstimate` from flow solvers
- Removed `m_currentTime` and `m_currentDt` from wells
- Passed `time_n` and `dt` as arguments of `calculateResidualNorm` (for normalization in wells and mimetic)

but if necessary, we can break the PR into smaller ones

Rebaseline done in https://github.com/GEOSX/integratedTests/pull/305

Resolves #1936


francoishamon:
If someone has a little bit of time in the next weeks, this (low-priority) PR is ready to be reviewed. It implements the option to choose between the $L^2$ and $L^\infty$ norms for all flow/well solvers. The main file to review in the PR is this [kernel file](https://github.com/GEOSX/GEOSX/blob/3940cf5741eaeb29da9a879a9c7c9970a9730ea1/src/coreComponents/physicsSolvers/SolverBaseKernels.hpp), used [here](https://github.com/GEOSX/GEOSX/blob/3940cf5741eaeb29da9a879a9c7c9970a9730ea1/src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBaseKernels.hpp#L468) for single-phase flow for instance.

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseHybridFVM.cpp
```diff
@@ -398,109 +398,124 @@ void SinglePhaseHybridFVM::saveAquiferConvergedState( real64 const & time,
...
+        if( subRegionResidualNorm[0] > localResidualNorm )
+        {
+          localResidualNorm = subRegionResidualNorm[0];
+        }
+      }
+      else
+      {
+        localResidualNorm += subRegionResidualNorm[0];
+        localResidualNormalizer += subRegionResidualNormalizer[0];
+      }
```
TotoGaz, line 470: This is not really object oriented.
Is there a smooth way not to get this `if`? Some dispatch obviously?

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseHybridFVM.cpp
```diff
@@ -398,109 +398,124 @@ void SinglePhaseHybridFVM::saveAquiferConvergedState( real64 const & time,
...
-    mesh.getElemManager().forElementSubRegions< ElementSubRegionBase >( regionNames, [&]( localIndex const,
-                                                                                          ElementSubRegionBase const & subRegion )
+    // step 1: compute the residual for the element-based mass conservation equations
+
+    elemManager.forElementSubRegions< ElementSubRegionBase >( regionNames,
+                                                              [&]( localIndex const,
+                                                                   ElementSubRegionBase const & subRegion )
     {
+      real64 subRegionResidualNorm[1]{};
+      real64 subRegionResidualNormalizer[1]{};
```
TotoGaz, line 433: Why do you allocate an array of length one and not a simple scalar?

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseHybridFVM.cpp
```diff
@@ -398,109 +398,124 @@ void SinglePhaseHybridFVM::saveAquiferConvergedState( real64 const & time,
...
 
-      arrayView1d< globalIndex const > const & elemDofNumber = subRegion.template getReference< array1d< globalIndex > >( elemDofKey );
-      arrayView1d< integer const > const & elemGhostRank = subRegion.ghostRank();
-      arrayView1d< real64 const > const & volume = subRegion.getElementVolume();
+      string const & fluidName = subRegion.getReference< string >( viewKeyStruct::fluidNamesString() );
+      SingleFluidBase const & fluid = getConstitutiveModel< SingleFluidBase >( subRegion, fluidName );
+      defaultViscosity += fluid.defaultViscosity();
+      subRegionCounter++;
```
TotoGaz, line 438: It was already there, but I do not quite understand why we divide by the number of sub regions.

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseHybridFVM.cpp
```diff
@@ -398,109 +398,124 @@ void SinglePhaseHybridFVM::saveAquiferConvergedState( real64 const & time,
...
-  real64 const elemResidualNorm = sqrt( globalResidualNorm[0] )
-                                  / ( ( globalResidualNorm[1] + m_fluxEstimate ) / (globalResidualNorm[2]+1) );
-  real64 const faceResidualNorm = sqrt( globalResidualNorm[3] );
-
-  real64 const residualNorm = ( elemResidualNorm > faceResidualNorm )
-                            ? elemResidualNorm
-                            : faceResidualNorm;
+  real64 const residualNorm = ( m_normType == solverBaseKernels::NormType::Linf )
+    ? MpiWrapper::max( localResidualNorm )
+    : sqrt( MpiWrapper::sum( localResidualNorm ) ) / MpiWrapper::sum( localResidualNormalizer );
```
TotoGaz, line 518: Same here about OOP.

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseFVM.cpp
```diff
@@ -106,128 +106,128 @@ void SinglePhaseFVM< BASE >::setupSystem( DomainPartition & domain,
...
-    real64 const flowResidual = residualFlowAllMeshes / numMeshTargets;
-    real64 const energyResidual = residualEnergyAllMeshes / numMeshTargets;
-
-    residual = std::sqrt( flowResidual*flowResidual + energyResidual*energyResidual );
+    real64 const flowResidualNorm = ( BASE::m_normType == solverBaseKernels::NormType::Linf )
+      ? MpiWrapper::max( localResidualNorm[0] )
+      : sqrt( MpiWrapper::sum( localResidualNorm[0] ) ) / MpiWrapper::sum( localResidualNormalizer[0] );
+    real64 const energyResidualNorm = ( BASE::m_normType == solverBaseKernels::NormType::Linf )
+      ? MpiWrapper::max( localResidualNorm[1] )
+      : sqrt( MpiWrapper::sum( localResidualNorm[1] ) ) / MpiWrapper::sum( localResidualNormalizer[1] );
```
TotoGaz, line 212: That could be nice to get OOP here too.

src/coreComponents/constitutive/solid/porosity/PorosityBase.cpp
```diff
@@ -36,26 +37,19 @@ PorosityBase::PorosityBase( string const & name, Group * const parent ):
...
-  registerWrapper( viewKeyStruct::newPorosityString(), &m_newPorosity ).
-    setPlotLevel( PlotLevel::LEVEL_0 ).
-    setApplyDefaultValue( 1.0 ); // will be overwritten but it's important for newly created faceElements.
+  registerWrapper( viewKeyStruct::defaultReferencePorosityString(), &m_defaultReferencePorosity ).
+    setInputFlag( InputFlags::REQUIRED ).
+    setDescription( "Default value of the reference porosity" );
 
-  registerWrapper( viewKeyStruct::oldPorosityString(), &m_porosity_n ).
-    setApplyDefaultValue( 1.0 ); // will be overwritten but it's important for newly created faceElements.
+  registerExtrinsicData( extrinsicMeshData::porosity::porosity{}, &m_newPorosity );
```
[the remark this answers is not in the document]
francoishamon: Yes this new `extrinsicData` for porosity was added here, but then I also added it in the net-to-gross PR (most likely merged first) in order to reduce the size of this PR.

src/coreComponents/physicsSolvers/SolverBase.hpp
```diff
@@ -395,16 +395,20 @@ class SolverBase : public ExecutableGroup
...
+   * @param dofManager degree-of-freedom manager associated with the linear system
+   * @param localRhs the system right-hand side vector
    * @return norm of the residual
    *
    * This function returns the norm of global residual vector, which is suitable for comparison with
    * a tolerance.
    */
   virtual real64
-  calculateResidualNorm( DomainPartition const & domain,
+  calculateResidualNorm( real64 const & time,
```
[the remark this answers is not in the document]
francoishamon: In fact here, it is quite convenient to have a dummy definition of `calculateResidualNorm` for the `SurfaceGenerator` as well as for the explicit solvers, since they don't care about residual norms.

src/coreComponents/physicsSolvers/SolverBaseKernels.hpp
```diff
@@ -0,0 +1,262 @@
...
+    m_ghostRank( ghostRank )
+  {}
+
+  /**
+   * @struct LinfStackVariables
+   * @brief Kernel variables located on the stack for Linf norm
+   */
+  struct LinfStackVariables
+  {
+public:
```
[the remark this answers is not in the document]
francoishamon: I removed the `public` keyword in this `struct`

src/coreComponents/physicsSolvers/SolverBaseKernels.hpp
```diff
@@ -0,0 +1,262 @@
...
+namespace geosx
+{
+
+namespace solverBaseKernels
+{
+
+/******************************** ResidualNormKernelBase ********************************/
+
+/**
+ * @class ResidualNormKernelBase
```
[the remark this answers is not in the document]
francoishamon: I removed this

src/coreComponents/physicsSolvers/SolverBaseKernels.hpp
```diff
@@ -0,0 +1,262 @@
...
+class ResidualNormKernelBase
+{
+public:
+
+
+  /// Compile time value for the number of norms to compute
+  static constexpr integer numNorm = NUM_NORM;
+
+  /// Const value used to make sure that normalizers are never zero
+  static constexpr real64 minNormalizer = 1e-12;
```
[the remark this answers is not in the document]
francoishamon: Yes I agree, but I would prefer not to do it in this PR. The PR is already going to change the results/convergence a little bit, so I just put the `constexpr` that we currently use in `develop` here to minimize potential differences.

src/coreComponents/physicsSolvers/SolverBaseKernels.hpp
```diff
@@ -0,0 +1,262 @@
...
+
+  };
+
+  /**
+   * @struct L2StackVariables
+   * @brief Kernel variables located on the stack for L2 norm
+   */
+  struct L2StackVariables : public LinfStackVariables
+  {
+public:
```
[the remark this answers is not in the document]
francoishamon: I removed the `public` keyword in this `struct`

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseHybridFVM.cpp
francoishamon: Yes, here I was trying to get a default value of viscosity for normalization purposes. But since the default viscosity lives in the constitutive model (attached to the subregion), I averaged by the number of subregions to get a representative value (instead of picking the default viscosity of a particular subregion).

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseHybridFVM.cpp
francoishamon: I see that you put a couple of comments on the way I did this `if/else`. I think I need help to find a better way to implement this, let's discuss it some time soon.

francoishamon: I put this code in helper functions, but it is still not exactly what you asked for. Now the code looks like:
https://github.com/GEOSX/GEOSX/blob/3eada543329ce0bd54db2f4d7adc0f42886f5cb9/src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.cpp#L224-L232
It is already slightly better than my first implementation, I think, but the `if/else` is still present.

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseHybridFVM.cpp
francoishamon: Here as well, I put this code in a helper function.

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseFVM.cpp
francoishamon: Some here, I used helpers:
https://github.com/GEOSX/GEOSX/blob/3eada543329ce0bd54db2f4d7adc0f42886f5cb9/src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseFVM.cpp#L203-L210
so not really OOP, but slightly better.

rrsettgast (approved):

src/coreComponents/physicsSolvers/contact/SolidMechanicsEmbeddedFractures.cpp
```diff
@@ -549,14 +549,16 @@ void SolidMechanicsEmbeddedFractures::applyTractionBC( real64 const time_n,
   } );
 }
 
-real64 SolidMechanicsEmbeddedFractures::calculateResidualNorm( DomainPartition const & domain,
+real64 SolidMechanicsEmbeddedFractures::calculateResidualNorm( real64 const & time,
```
CusiniM, line 552: when you merge, let us create an issue to remind us to change the contact solvers too.

CusiniM (approved):
I have some minor comments and I think that we should probably fix all other solvers as well but I m fine with merging this frist.

PR #1999, opened 2022-08-04
Restore Pangea2 deployment

TotoGaz:
It appears that
- The same name was used for all the installation directories. Resulting in installation clashes. Putting back the hash of the version solves the issue.
- The `.travis.yaml` / `scripts/ci_build_and_test.sh` split did erase some shell variables. I did duplicate `CONTAINER_NAME` and recompute `GEOSX_TPL_DIR` to deal with this.


bmhan12 (approved):
Good catch!

PR #2008, opened 2022-08-12
Use brew's tap to install precise version of open-mpi

TotoGaz:
This fixes the build failure on Mac OSX.
I've been modified the way we install `open-mpi` using `brew` and its `tap`s.
A strong difference is that from now `brew` recompiles `open-mpi`, which takes ~5 more minutes for GEOSX or TPL builds.

I also must confess that I do not know anything about `brew` and that I simply stack command lines there. 

Related to https://github.com/GEOSX/thirdPartyLibs/pull/194

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: minimal
 
 env:
   global:
-  - GEOSX_TPL_TAG=192-765
+  - GEOSX_TPL_TAG=194-777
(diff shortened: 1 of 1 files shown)


bmhan12 (approved):
LGTM! 
(To be honest, I don't have much familiarity with homebrew either as a Windows laptop user)

TotoGaz:
> LGTM! (To be honest, I don't have much familiarity with homebrew either as a Windows laptop user)

Thx for validation.
I've found some example here and there on the internet, but it's voodoo to me!

Anyway, this looks a little cleaner than what we used to be doing with the hash code.
Time will tell 🤞

PR #2011, opened 2022-08-13
Move fracture fields from EdgeManager to FaceElementSubRegion.

TotoGaz:
Part of the https://github.com/GEOSX/GEOSX/issues/2007 effort.
Needs to rebaseline.


src/coreComponents/mesh/EdgeManager.cpp
```diff
@@ -33,33 +33,14 @@ using namespace dataRepository;
...
-    setPlotLevel( PlotLevel::NOPLOT ).
-    setDescription( "A map of fracture connector local indices to edge local indices." ).
-    setSizedFromParent( 0 );
-
-  registerWrapper( viewKeyStruct::fractureConnectorsEdgesToFaceElementsIndexString(),
-                   &m_fractureConnectorEdgesToFaceElements ).
-    setPlotLevel( PlotLevel::NOPLOT ).
-    setDescription( "A map of fracture connector local indices face element local indices" ).
-    setSizedFromParent( 0 );
-
```
TotoGaz, line 62: @CusiniM do you think I need to register them again on `FaceElementSubRegion`?

CusiniM: hmm, I am not sure if they are ever accessed through the wrapper interface. I guess we can try not to. Although if not registered I don't think they ll be in restart files which is something we may want since it helps catching deviations in integratedTests.

TotoGaz: Yes, the current code compiles validates the integrated tests.
But I was thinking about the restart files indeed.
I bet I'll have to register and rebaseline then.

CusiniM: yeah, I think you'll have to. Well, I d say you can go ahead and then this PR looks ready.

CusiniM (approved):

PR #2012, opened 2022-08-13
Add getter for region type.

TotoGaz:
In TPFA, we only want to compute the cell region connexion.
So when iterating along the regions, we need to filter out the well and face regions.

This PR proposes a simple solution, but we could imagine to other filtering techniques.
Also note that we need to access the index of the region in the region sub group.

Part of the https://github.com/GEOSX/GEOSX/issues/2007 effort.

### DIFF
--- a/src/coreComponents/finiteVolume/TwoPointFluxApproximation.cpp
+++ b/src/coreComponents/finiteVolume/TwoPointFluxApproximation.cpp
@@ -105,11 +105,15 @@ void TwoPointFluxApproximation::computeCellStencil( MeshLevel & mesh ) const
 
   ArrayOfArraysView< localIndex const > const faceToNodes = faceManager.nodeList().toViewConst();
 
-  // make a list of region indices to be included
+  // We only include the cell regions, discarding face and well regions.
   SortedArray< localIndex > regionFilter;
-  for( string const & regionName : m_targetRegions.at( mesh.getParent().getParent().getName() ) )
+  for( string const & regionName: m_targetRegions.at( mesh.getParent().getParent().getName() ) )
   {
-    regionFilter.insert( elemManager.getRegions().getIndex( regionName ) );
+    ElementRegionBase const & region = elemManager.getRegion( regionName );
+    if( region.getType() == ElementRegionBase::Type::CELL )
+    {
+      regionFilter.insert( elemManager.getRegions().getIndex( regionName ) );
+    }
   }
 
   stencil.reserve( faceManager.size() );
--- a/src/coreComponents/finiteVolume/TwoPointFluxApproximation.cpp
+++ b/src/coreComponents/finiteVolume/TwoPointFluxApproximation.cpp
@@ -105,15 +105,11 @@ void TwoPointFluxApproximation::computeCellStencil( MeshLevel & mesh ) const
 
   ArrayOfArraysView< localIndex const > const faceToNodes = faceManager.nodeList().toViewConst();
 
-  // We only include the cell regions, discarding face and well regions.
+  // make a list of region indices to be included
   SortedArray< localIndex > regionFilter;
-  for( string const & regionName: m_targetRegions.at( mesh.getParent().getParent().getName() ) )
+  for( string const & regionName : m_targetRegions.at( mesh.getParent().getParent().getName() ) )
   {
-    ElementRegionBase const & region = elemManager.getRegion( regionName );
-    if( region.getType() == ElementRegionBase::Type::CELL )
-    {
-      regionFilter.insert( elemManager.getRegions().getIndex( regionName ) );
-    }
+    regionFilter.insert( elemManager.getRegions().getIndex( regionName ) );
   }
 
   stencil.reserve( faceManager.size() );
--- a/src/coreComponents/finiteVolume/TwoPointFluxApproximation.cpp
+++ b/src/coreComponents/finiteVolume/TwoPointFluxApproximation.cpp
@@ -107,15 +107,14 @@ void TwoPointFluxApproximation::computeCellStencil( MeshLevel & mesh ) const
 
   // make a list of region indices to be included
   SortedArray< localIndex > regionFilter;
-  for( string const & regionName : m_targetRegions.at( mesh.getParent().getParent().getName() ) )
+  arrayView1d< string const > const targetRegions = m_targetRegions.at( mesh.getParent().getParent().getName() );
+  elemManager.forElementRegionsComplete< CellElementRegion >( targetRegions,
+                                                              [&]( localIndex,
+                                                                   localIndex const ei,
+                                                                   CellElementRegion const & )
   {
-    ElementRegionBase const & region = elemManager.getRegion( regionName );
-    // Using the catalog name as a trick to select only cell element regions...
-    if( region.getCatalogName() == CellElementRegion::catalogName() )
-    {
-      regionFilter.insert( elemManager.getRegions().getIndex( regionName ) );
-    }
-  }
+    regionFilter.insert( ei );
+  } );
 
   stencil.reserve( faceManager.size() );
 
--- a/src/coreComponents/finiteVolume/TwoPointFluxApproximation.cpp
+++ b/src/coreComponents/finiteVolume/TwoPointFluxApproximation.cpp
@@ -109,7 +109,12 @@ void TwoPointFluxApproximation::computeCellStencil( MeshLevel & mesh ) const
   SortedArray< localIndex > regionFilter;
   for( string const & regionName : m_targetRegions.at( mesh.getParent().getParent().getName() ) )
   {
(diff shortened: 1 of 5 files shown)


src/coreComponents/finiteVolume/TwoPointFluxApproximation.cpp
```diff
@@ -105,11 +105,15 @@ void TwoPointFluxApproximation::computeCellStencil( MeshLevel & mesh ) const
...
 
-  // make a list of region indices to be included
+  // We only include the cell regions, discarding face and well regions.
   SortedArray< localIndex > regionFilter;
-  for( string const & regionName : m_targetRegions.at( mesh.getParent().getParent().getName() ) )
+  for( string const & regionName: m_targetRegions.at( mesh.getParent().getParent().getName() ) )
   {
-    regionFilter.insert( elemManager.getRegions().getIndex( regionName ) );
+    ElementRegionBase const & region = elemManager.getRegion( regionName );
+    if( region.getType() == ElementRegionBase::Type::CELL )
```
TotoGaz, line 113: @CusiniM what do we want to do for wells? We did not consider the case last time we met.

src/coreComponents/mesh/ElementRegionBase.hpp
```diff
@@ -303,6 +319,11 @@ class ElementRegionBase : public ObjectManagerBase
     static constexpr char const * elementSubRegions() { return "elementSubRegions"; }
   };
 
+protected:
+
+  /// The type of the region
+  Type m_type;
```
TotoGaz, line 325: Do I have to register this attribute?

src/coreComponents/mesh/ElementRegionBase.hpp
```diff
@@ -68,6 +68,22 @@ class ElementRegionBase : public ObjectManagerBase
 
   ///@}
 
+  /// The actual region type
+  enum Type
+  {
+    CELL,
+    FACE,
+    WELL
+  };
```
rrsettgast, line 77: Isn't this some sort of anti-pattern?

src/coreComponents/finiteVolume/TwoPointFluxApproximation.cpp
rrsettgast, line 113: Doesn't this do the job without having to add the derived info to the base class?
```suggestion
    if( region.getCatalgoName() == CellElementRegion::catalogName() )
```

rrsettgast (commented):
I don't think I understand why we would want to put an enum of the derived types in the base.

https://softwareengineering.stackexchange.com/questions/219543/should-a-class-know-about-its-subclasses

TotoGaz:
> I don't think I understand why we would want to put an enum of the derived types in the base.

While I understand your comment, I do not think we’re exactly in this kind of pattern there.

My use case is that I need to be sure that we are only computing the TPFA stencils between volumic cells. (@CusiniM correct me if I’m wrong).

Our current hierarchy is
```
ElementRegionBase
├── CellElementRegion
├── SurfaceElementRegion
└── WellElementRegion
```
(Which first poses a question already (not really the scope but…): what about the `WellElementRegion` is _you are not using wells_ but still want _line_ elements?)

However, I understand the
```c++
enum Type {
  CELL, FACE, WELL
}
```
which can be changed for something like
```c++
enum Dimension {
   ONE, TWO, THREE
}
```
or even something like
```c++
int ElementRegionBase::getDimension() const {
  return …; // can be 1, 2 or 3
}
```
as a print of the actual dimension of the `ElementRegionBase`.
And we could imagine to insert/split into the GEOSX class hierarchy some `(One|Two|Three)DimElementRegion` classes.

This being said, using the `enum Type`, `ElementRegionBase` does not know about it’s subclasses. I more interpret it as knowing it’s dimension, which looks acceptable. How these 1, 2, 3-dimensional element regions are being implemented is unknown to `ElementRegionBase`; it does not even know that they exist.

Anyway, we could also imagine some kind of
```c++
bool isA< T >() const {
  return dynamic_cast< const T * >( this );
}
```
Is it better? I don’t know…

I’m fine with the `catalog` solution, even if I feel this is not how `catalog` should be used. 😉

CusiniM:
> > I don't think I understand why we would want to put an enum of the derived types in the base.
> 
> While I understand your comment, I do not think we’re exactly in this kind of pattern there.
> 
> My use case is that I need to be sure that we are only computing the TPFA stencils between volumic cells. (@CusiniM correct me if I’m wrong).
> 
> Our current hierarchy is
> 
> ```
> ElementRegionBase
> ├── CellElementRegion
> ├── SurfaceElementRegion
> └── WellElementRegion
> ```
> 
> (Which first poses a question already (not really the scope but…): what about the `WellElementRegion` is _you are not using wells_ but still want _line_ elements? And also, a `Cell` is not the same kind as a `Surface` which is not the same kind as a `Well`)
> 
> However, I understand the
> 
> ```c++
> enum Type {
>   CELL, FACE, WELL
> }
> ```
> 
> which can be changed for something like
> 
> ```c++
> enum Dimension {
>    ONE, TWO, THREE
> }
> ```
> 
> or even something like
> 
> ```c++
> int ElementRegionBase::getDimension() const {
>   return …; // can be 1, 2 or 3
> }
> ```
> 
> as a print of the actual dimension of the `ElementRegionBase`. And we could imagine to insert/split into the GEOSX class hierarchy some `(One|Two|Three)DimElementRegion` or `(Line|Surface|Volume)ElementRegion` classes.
> 

> Anyway, we could also imagine some kind of
> 
> ```c++
> bool isA< T >() const {
>   return dynamic_cast< const T * >( this );
> }
> ```
> 
> Is it better? I don’t know…
> 
> I’m fine with the `catalog` solution, even if I feel this is not how `catalog` should be used. 😉

I think @rrsettgast idea of using the static catalog name for this is actually a good one. It does uniquely define the type of the region so I don't really see a drawback. I should have thought about it. 

About the wells... The truth is that wells do not have a stencil at the moment. Should they? Maybe... they certainly could but it's not the way they were initially implemented and I think it would be quite a lot of work to change that now. So, in practice, we never want to include wells in our stencils.

TotoGaz:
> I think @rrsettgast idea of using the static catalog name for this is actually a good one. It does uniquely define the type of the region so I don't really see a drawback. I should have thought about it.

I've pushed a version with the `catalog` trick.
Even if I do not think it's best to use in our stencil algorithm an info that relates to xml input files and C++ factory registration.

src/coreComponents/finiteVolume/TwoPointFluxApproximation.cpp
```diff
@@ -109,7 +109,12 @@ void TwoPointFluxApproximation::computeCellStencil( MeshLevel & mesh ) const
   SortedArray< localIndex > regionFilter;
   for( string const & regionName : m_targetRegions.at( mesh.getParent().getParent().getName() ) )
   {
-    regionFilter.insert( elemManager.getRegions().getIndex( regionName ) );
+    ElementRegionBase const & region = elemManager.getRegion( regionName );
+    // Using the catalog name as a trick to select only cell element regions...
```
rrsettgast, line 113: I don't know why this is referred to as a "trick". There isn't much tricky about it. If we prefer, we can create a virtual function to return `typeid(T)` and compare that instead. Perhaps that would be less of a "trick"?

TotoGaz: Because in this TPFA code _which decides what are the connections between the discretized domains_,
we end up using features that are designed to be used 1) _for the user input file_ and 2) _the technical registration of products in our factories_.

The computation domain of the TPFA does not belong to the xml/factory world.
And vice versa.

Using/calling functions that are not of the same "_world_" or "_domain_" or "_level_" should not "_really_" happen.
Imagine that at some moment we change something about the `catalog`, TPFA gets impacted.

This is why I consider this is a trick.

PS: I'm pretty sure there is an established wording for this, but I cannot figure it out right now, sorry.
It will hopefully get back to my mind again 🤷

klevzoff: In this case we might be able to rewrite this whole loop using our usual filtering technique (commentware, to be tested):
```c++
arrayView1d< string const > const targetRegions =
  m_targetRegions.at( mesh.getParent().getParent().getName() );

elemManager.forElementRegionsComplete< CellElementRegion >( targetRegions,
                                                            [&]( localIndex, 
                                                                 localIndex const ei, 
                                                                 CellElementRegion const & )
{
  regionFilter.insert( ei );
} );
```

In general, I agree that using strings as subtype discriminators is definitely a bit... unorthodox, even if there is no performance constraint. Luckily, due to our downcast-based filtering, we have largely avoided relying on catalog names for things like this. Not sure if this will be always possible though.

CusiniM (approved):
yeah, this seems a typical GEOSX way of doing things!

PR #2013, opened 2022-08-13
Fix TPL discrepancy between travis and azure

TotoGaz:
I forgot that we have duplicated the `GEOSX_TPL_TAG` on travis and azure.
Even if this is not a problem because there's only impact on the Mac OSX build, I prefer to keep it clean.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,5 +1,5 @@
 variables:
-  GEOSX_TPL_TAG: 192-765
+  GEOSX_TPL_TAG: 194-777
 
 # only build non-draft, merge target PR's to develop
 trigger: none


klevzoff:
If we plan to keep Travis jobs around, would be nice to have a way of automatically avoiding this de-sync of TPL tags. Maybe instead of maintaining separate variables in yaml scripts, we write the tag to a file stored in the repo, and have `ci_build_and_test.sh` retrieve the tag from that file?

TotoGaz:
> If we plan to keep Travis jobs around...

I do not really know what's planned for travis vs azure.
@rrsettgast @bmhan12 do you have any insights?

> would be nice to have a way of automatically avoiding this de-sync of TPL tags. Maybe instead of maintaining separate variables in yaml scripts, we write the tag to a plaintext file stored in the repo, and have `ci_build_and_test.sh` retrieve the tag from that file?

We can even parse `.travis.yaml` from `azure` with `yq` 🤣

bmhan12:
> I do not really know what's planned for travis vs azure.
@rrsettgast @bmhan12 do you have any insights?

@TotoGaz @rrsettgast  I think with the Out of Memory issues we saw with Azure, Travis may have been our best CI service after all.
The Azure jobs are within the memory limit currently, but I am worried we will hit it again in the future, and Microsoft[ does not plan to accept requests for more space](https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml#hardware). With Travis the concern is the time limit, but in the past we have been able to increase it to pass the [advertised paid limit ](https://www.travis-ci.com/pricing/) (120 minutes apparently). 

I'm leaning towards going fully back to Travis. We could run a small set of jobs on Azure to add [10 free parallel jobs to our overall parallel job limits](https://docs.microsoft.com/en-us/azure/devops/pipelines/licensing/concurrent-jobs?view=azure-devops&tabs=ms-hosted#how-much-do-parallel-jobs-cost), but that does not seem very clean.

TotoGaz:
About the limitation for Azure agent limitation, it looks like we can have our own agents.
These agents can obviously be in the azure cloud 🤣 https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/agents?view=azure-devops&tabs=browser#azure-virtual-machine-scale-set-agents

Do we have any info about that?

We would pay for what we use, maybe could we split the pools between the "free" ones and the big ones we would buy (if it make sense, which is really not obvious)!

bmhan12:
> Do we have any info about that?

@TotoGaz We talked to LC, and they don't have any self-hosted agents for Azure. They are currently paying for additional concurrent/parallel jobs that is shared across some projects,  while other projects are just using the free ones (e.g. Axom, conduit).

TotoGaz:
> We talked to LC, and they don't have any self-hosted agents for Azure. They are currently paying for additional concurrent/parallel jobs that is shared across some projects, while other projects are just using the free ones (e.g. Axom, conduit).

Hmm... And do we know if they could help us on that? and perform some investigations?
I mean, it's something a lot of people could benefit from!

I think that with the appropriate permissions, doing a proof of concept like https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/scale-set-agents?view=azure-devops looks within the reach of a few days of work? (and most probably even less)

bmhan12:
> Hmm... And do we know if they could help us on that? and perform some investigations?
I mean, it's something a lot of people could benefit from!

Maybe... I think the concern is security on their end.
For example, gitlab is how we run nightlyTests on LC machines. We _could_ also enable this so it also runs as CI for PR's, however only users with a LC account can actually login to view the results of runs. That is why for nightlyTests we pipe the output from gitlab (and also for bamboo previously) to a separate repo for viewability.
As far as I know, gitlab is the way to access LC's self-hosted agents, but it comes with access restrictions.

TotoGaz:
BTW, can you review this PR to reset consistency while we find a long term solution? 😉

bmhan12 (approved):

PR #2017, opened 2022-08-17
Use VTK Global IDs if available

sframba:
When loading a parallel VTK file, global IDs for cells and vertices are computed via the `vtkGenerateGlobalIds` class. This can be costly, since it involves sorting and identifying points globally, which entails a lot of MPI communications and floating point calculations. However, sometimes these IDs are already available, for example when the parallel mesh was obtained by cutting a known "master" mesh. Moreover, gluing points together can miss some nuance in the input geometry. For example, it is impossible to introduce a localized disconnection in the mesh without physically displacing the impacted points. For these reasons, global ID arrays represent useful information.

In order to save valuable initialization time, this PR adds a flag, called `useGlobalIds`, to the `VTKMesh` input XML field. When `useGlobalIds`, the mesh loader will try to find global ID arrays for both cells and vertices in the input file. These arrays must be flagged as `GLOBAL_IDS` in the input file. If both are found, the calculation of global IDs is skipped. If `useGlobalIds` is set to anything but `1`, instead, global IDs are always computed, as before. Deactivating this option can be useful if the global ID arrays are available in the current mesh, but should not be used.

A first test of this PR on an acoustic wave propagation problem on a mesh with about 94M hexahedral elements shows that we can indeed gain something in problem initialization, if `GlobalIds` are available:

| Without `GlobalIds` | With `GlobalIds` |
| :------------: | :---------: |
`time to execute initialize [92.30274181]` | `time to execute initialize [77.40696464]` |

and Caliper shows that this comes indeed from `geosx::vtk::redistributeMesh`:

| Avg time/rank without `GlobalIds` | Avg time/rank with `GlobalIds` |
| :------------: | :---------: |
`18.642143` | `0.510772` |

Rebaseline done in https://github.com/GEOSX/integratedTests/pull/249


XL64 (approved):

src/coreComponents/mesh/generators/VTKMeshGenerator.hpp
```diff
@@ -165,6 +166,9 @@ class VTKMeshGenerator : public ExternalMeshGeneratorBase
   /// Number of graph partitioning refinement iterations
   integer m_partitionRefinement = 0;
 
+  /// Whether global id arrays should be used, if available
+  integer m_useGlobalIds = 0;
```
TotoGaz, line 170: The default value of the field does not match the default value registered.
Is this intended?

src/coreComponents/mesh/generators/VTKMeshGenerator.cpp
```diff
@@ -399,21 +405,34 @@ findNeighborRanks( std::vector< vtkBoundingBox > boundingBoxes )
...
 
-  // Generate global IDs for vertices and cells
-  vtkNew< vtkGenerateGlobalIds > generator;
-  generator->SetInputDataObject( &loadedMesh );
-  generator->Update();
-  vtkSmartPointer< vtkDataSet > mesh =
-    vtkDataSet::SafeDownCast( generator->GetOutputDataObject( 0 ) );
+  // Generate global IDs for vertices and cells, if needed
+  vtkSmartPointer< vtkDataSet > mesh;
+  if( useGlobalIds != 1
```
TotoGaz, line 421: Please be more C++ friendly :)
A value of 2 is supposed to be `true` and this is not how you're dealing with it.
Maybe change the signature to `bool const useGlobalIds` instead of `int const useGlobalIds` and properly deal with conversion (or let C++ do it) at call?

src/coreComponents/mesh/generators/VTKMeshGenerator.cpp
```diff
@@ -399,21 +405,34 @@ findNeighborRanks( std::vector< vtkBoundingBox > boundingBoxes )
...
+  {
+    vtkNew< vtkGenerateGlobalIds > generator;
+    generator->SetInputDataObject( &loadedMesh );
+    generator->Update();
+    mesh = vtkDataSet::SafeDownCast( generator->GetOutputDataObject( 0 ) );
+  }
+  else
+  {
+    mesh.TakeReference( &loadedMesh );
+    std::cout << "using global Ids... wish me luck" << std::endl;
```
TotoGaz, line 433:
```suggestion
    GEOSX_LOG( "Using global Ids defined in array GLOBAL_IDS." );
```
Plus, can you add small checks to be sure that the fields have the proper dimensions, etc...
If you think this is tricky to use (`wish me luck`), we should be careful I think.

src/coreComponents/mesh/generators/VTKMeshGenerator.cpp
```diff
@@ -399,21 +405,34 @@ findNeighborRanks( std::vector< vtkBoundingBox > boundingBoxes )
...
-  vtkNew< vtkGenerateGlobalIds > generator;
-  generator->SetInputDataObject( &loadedMesh );
-  generator->Update();
-  vtkSmartPointer< vtkDataSet > mesh =
-    vtkDataSet::SafeDownCast( generator->GetOutputDataObject( 0 ) );
+  // Generate global IDs for vertices and cells, if needed
+  vtkSmartPointer< vtkDataSet > mesh;
+  if( useGlobalIds != 1
+      || loadedMesh.GetPointData()->GetGlobalIds() == nullptr
+      || loadedMesh.GetCellData()->GetGlobalIds() == nullptr )
```
TotoGaz, line 423: Which means that somebody can request to use provided global ids but something may be "wrong" (`loadedMesh.GetPointData()->GetGlobalIds() == nullptr`) and you end up generating them.

Please change the behavior: if the user asks to use the global ids and it's not possible, please enter an error message and kill the simulation because what's happening is not what is expected and something needs to be fixed (remove the `useGlobalIds` flags or better fill the `GLOBAL_IDS` field...)

src/coreComponents/schema/schema.xsd
```diff
@@ -1324,6 +1324,8 @@ stress - traction is applied to the faces as specified by the inner product of i
 		<xsd:attribute name="translate" type="R1Tensor" default="{0,0,0}" />
 		<!--name => A name is required for any non-unique nodes-->
 		<xsd:attribute name="name" type="string" use="required" />
+		<!--useGlobalIds => If set to 1, mesh loading will use the GlobalId attribute for vertices and meshes, if available. This can avod unnecessary calculations. GlobalId arrays must be flagged as GLOBAL_IDS in the input mesh.--> 
+		<xsd:attribute name="useGlobalIds" type="integer" defaault="1" />
```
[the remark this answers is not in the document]
TotoGaz: How could there be a mistake in `default` in this file which should be automatically generated?

src/coreComponents/schema/docs/VTKMesh.rst
```diff
@@ -14,6 +14,7 @@ partitionRefinement integer                                1         Number of p
...
 scale               R1Tensor                               {1,1,1}   Scale the coordinates of the vertices by given scale factors (after translation)                                                                                                                                                                                                
 translate           R1Tensor                               {0,0,0}   Translate the coordinates of the vertices by a given vector (prior to scaling)                                                                                                                                                                                                  
+useGlobalIds        integer                                1         If set to 1, mesh loading will use the GlobalId attribute for vertices and meshes, if available. This can avod unnecessary calculations. GlobalId arrays must be flagged as GLOBAL_IDS in the input mesh.
```
TotoGaz, line 17: Same, with the `avod` and generated file.
Maybe you did not regenerate the file after fixing the typo in your `registerWrapper`?

TotoGaz (commented):
I may be using this for fractures where I have duplicated nodes! 🎉

rrsettgast:
A general question. Why wouldn't we always use the global index that is read in from the external file?

src/coreComponents/schema/schema.xsd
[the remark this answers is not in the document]
sframba: @TotoGaz I forgot to commit the auto generated file. Will do in the next commit.

src/coreComponents/mesh/generators/VTKMeshGenerator.hpp
sframba: @TotoGaz No, this is the default value that should be used for backwards compatibility. I will change the default everywhere to `0`

sframba:
> I may be using this for fractures where I have duplicated nodes! 🎉

Good to know! I figured it might be useful in some cases, but I didn't know which ones exactly

src/coreComponents/schema/docs/VTKMesh.rst
sframba: Yep, I forgot to commit the auto-generated file, I will correct this in the next commit

src/coreComponents/mesh/generators/VTKMeshGenerator.cpp
sframba: Whoops, forgot to change it from debug to logging. Thanks. I will add the basic checks

src/coreComponents/schema/schema.xsd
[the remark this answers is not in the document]
sframba: After today's discussion, I will set the default to `1` instead: the global ID array should be used whenever available

src/coreComponents/mesh/generators/VTKMeshGenerator.cpp
sframba: I fundamentally agree, but in this case, we need a bit more nuance with the useGlobalIds option. If we set it to `true` by default, then the run would abort whenever a mesh doesn't contain ids (and I suspect that's the majority) -- cue a very confused user...
I suggest we allow three values:
 - `useGlobalIds<0`: it will never use global Ids from the mesh, even if they are present
 - `useGlobalIds=0`: the default value, it will use Ids if present and valid, and generate them if not available;
 - `useGlobalIds>0` it will always use mesh Ids, and never generate them. It would abort if they are not available.

what do you think?

sframba:
> A general question. Why wouldn't we always use the global index that is read in from the external file?
Good question. Based on today's discussion, I suggest allowing three values (see also comment above):
 - `useGlobalIds<0`: it will never use global Ids from the mesh, even if they are present
 - `useGlobalIds=0`: the default value, it will use Ids if present and valid, and generate them if not available;
 - `useGlobalIds>0` it will always use mesh Ids, and never generate them. It would abort if they are not available.
What do you think?

TotoGaz:
> * `useGlobalIds<0`: it will never use global Ids from the mesh, even if they are present
> * `useGlobalIds=0`: the default value, it will use Ids if present and valid, and generate them if not available;
> * `useGlobalIds>0` it will always use mesh Ids, and never generate them. It would abort if they are not available.

I'm fine having the 3 solutions.
But maybe solution 2 could be instead: _Use if present; but fail if invalid. If not present, generate_.
With the first solution proposed, you could be thinking the ids are getting used, which may not be the case.

src/coreComponents/mesh/generators/VTKMeshGenerator.cpp
sframba: With latest comit, we have 3 possibilities (<0, 0, >0) so I have kept the flag as an `int`, but it is now handled in a more C++ friendly way :-)

sframba:
> I'm fine having the 3 solutions. But maybe solution 2 could be instead: _Use if present; but fail if invalid. If not present, generate_. With the first solution proposed, you could be thinking the ids are getting used, which may not be the case.

Sounds good. I have added the checks on the arrays.

TotoGaz:
@klevzoff out of curiosity, why does this PE require rebaseline?
It changes something in the `VTKMeshGenerator` which is a temporary and (I think) gets ditched.
How could this influence any restart file for example?
Is there anything else I'm missing?

klevzoff:
@TotoGaz it does add a new input attribute `useGlobalIds` though, so all existing vtk-basd tests will fail check

TotoGaz:
> @TotoGaz it does add a new input attribute `useGlobalIds` though, so all existing vtk-basd tests will fail check

I assume that there is one option that matches exactly the previous behavior, so updating the `xml` input files should be enough?
And those files are in the `GEOSX` repository, we shouldn't have to update the `integratedTests` repository, should we?

PR #2018, opened 2022-08-17
Re-enable object libraries build

bmhan12:
This PR:

- For object libraries, fixes the link-order undefined reference error between hypre and umpire.
- Sets object libraries build as default.
- Changes Azure Pipelines CI build for CUDA debug to use one process to prevent out of memory error.

### DIFF
--- a/scripts/build_test_helper.sh
+++ b/scripts/build_test_helper.sh
@@ -67,8 +67,9 @@ if [[ "$*" == *--test-documentation* ]]; then
 fi
 
 # "Make" target check (builds geosx executable target only if true)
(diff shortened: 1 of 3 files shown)


klevzoff (approved):

rrsettgast (approved):

TotoGaz:
@bmhan12 Do you want to go like that or do you want to make some extra check with the new machines?

bmhan12:
> @bmhan12 Do you want to go like that or do you want to make some extra check with the new machines?

@TotoGaz  This can go in as-is.

PR #2025, opened 2022-08-24
Using `large` VMs travis

TotoGaz:
If this succeeds, let's see if we can find define one size per job (CUDA builds get more).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 language: minimal
 
+vm:
+  size: x-large
+
 env:
   global:
   - GEOSX_TPL_TAG=194-777
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: minimal
 
-#vm:
-#  size: x-large
+vm:
+  size: large
 
 env:
   global:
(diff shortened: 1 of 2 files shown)


rrsettgast:
@TotoGaz Let's just merge this. It is what we are paying for anyways.

.travis.yml
```diff
@@ -16,6 +19,8 @@ geosx_osx_build: &geosx_osx_build
   os: osx
   osx_image: xcode13.2
   install:
+  - echo "running nproc"; nproc
+  - top -l 1 -s 0 | grep PhysMem
```
TotoGaz, line 23: Do we need those lines merged?

TotoGaz:
I cannot approve my own PR 🤣

bmhan12 (approved):

.travis.yml
bmhan12: I'm fine, either way. We can keep it for debugging purposes or remove for cleanliness.

TotoGaz: I'll think we'll find the cleaner way to get this information after all our final moves.

You're next in line in the merge queue for both incoming PRs.
Can you merge? Consider the order you think most appropriate for merging 😉

bmhan12: > Can you merge? Consider the order...

I do not have merge authority.
Feel free to merge this PR first.

PR #2027, opened 2022-08-25
add multiBody/multiLevel support to VTK output system

andrembcosta:
This pull request allows the VTK output to define multiple blocks if the GEOSX problem contains multiple MeshBodies or MeshLevels.


klevzoff:
I pushed some changes to add proper nesting of body/level blocks in the VTM file also, such that they reflect the structure of vtu files.

image

There is actually an issue with running the multi-body example in parallel, which I think is related to how `InternalMeshGenerator` and `SpatialPartition` interact (the latter doesn't seem to handle multiple bodies properly). This PR is not the right place to fix it however.

klevzoff (approved):
This is good to be merged imo. The XML example is not yet included in the integrated test suite, but it cannot currently be run in parallel anyway, without fixing the mesh generator. This XML can serve as a debug/test case for that.

We should also inquire the status of https://github.com/GEOSX/GEOSX/pull/1742 which seems to address the same functionality, but hasn't been updated in a long time.

andrembcosta:
Thanks Sergey! I thought about the integrated test as well, but in addition to the issue on the parallel run, it seems like the VTK stuff doesn't get written to the restart files, so, it wouldn't actually get tested?

klevzoff:
> Thanks Sergey! I thought about the integrated test as well, but in addition to the issue on the parallel run, it seems like the VTK stuff doesn't get written to the restart files, so, it wouldn't actually get tested?

Yeah, I guess what I meant is to have just the input file itself be run regularly so that it doesn't fall behind. Files that aren't part of the test suite tend to rot over time. But you're right that we can't actually test the correctness of vtk output automatically.

PR #2028, opened 2022-08-25
Spack Update

bmhan12:
This PR:

- Updates spack build, supports uberenv or spack environments to build third party libraries (Currently I think uberenv is much cleaner, spack environments produces a lot of spack-specific files that are hard to clean).

TODO (either this PR or follow-up PRs) :
- [x] Generate two host-configs for each system/compiler : one for LvArray, one for GEOS
  - LvArray host-configs are prefixed with `lvarray-` (e.g. `lvarray-quartz-toss_4_x86_64_ib-clang@14.0.6.cmake`)
- [ ] Migrate Github Actions Docker containers over to using spack
- [ ] Cleanup package recipes (upstream novel changes to spack, try recipes with newer spack release, etc.)
- [x] On blueos, `blas/lapack` implementation is spack-built `netlib-lapack` instead of pre-built system `essl`. 
     - Issues with building `suite-sparse` and `superlu-dist`  with `essl`
- [x] Same note, different `suite-sparse` (5.7.2) and `superlu-dist` (6.4.0) versions from current TPLs
- [x]  Only one tested so far on each quartz and lassen system, build with other compilers.
  - [x] Build with clang and gcc compilers on quartz and lassen
- [x]  `pygeosx` /ATS dependencies build
- [ ] `petsc` build  (untested, not working)
~- [ ] Spack build uses CUDA 11 to get around [spack-built RAJA issues with cub](https://raja.readthedocs.io/en/develop/sphinx/user_guide/config_options.html#programming-model-back-end-support). Upgrade `thirdPartyLibs` to CUDA 11.~
- [x] Explore spack environments, see if there is a setting to redirect generated file input

Change List:
- ~~LvArray change GEOSX/LvArray#264 (bugfix for failing test with CUDA 11)~~
- LvArray Change: GEOS-DEV/LvArray#305
- thirdPartyLibs change: GEOS-DEV/thirdPartyLibs#252
- Removes `uberenv` files, adds `uberenv` as a submodule
-  Updated spack version to latest release (v0.20.0)
- Moved configuration files from `scripts/uberenv/spack_configs` --> `scripts/spack_configs`
- Moved package recipe files from `scripts/uberenv/spack_packages` --> `scripts/spack_packages/packages` 
  -  The redundant `packages` subdirectory is required by spack environments for custom package locations.
- Created a `scripts/pygeosx_configs/spack.yaml` and `pygeosx` recipe specifically for building GEOS's python dependencies.
  - Idea is the python dependencies take a long time to build, change less frequently, and have more restrictive compiler requirements than GEOS. 
 - Adds and updates package recipes and adds patch files from spack.
 - Update sphinx documentation.


bmhan12:
@wrtobin @rrsettgast 
This can be reviewed!
I have it tested and working with only clang@14 on quartz and clang@10.0.1 on lassen, but I figured there were already alot of changes before iterating further.

wrtobin:
I have some WIP lassen ea toss4 env files, do we want to try getting them into this branch or would a follow-up be preferable?

.uberenv_config.json
```diff
@@ -0,0 +1,11 @@
+{
+"package_name": "geosx",
+"package_version": "develop",
+"package_final_phase": "hostconfig",
+"package_source_dir": "../..",
```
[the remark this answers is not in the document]
bmhan12: > I assume this is relative to the invocation point which must be scripts/uberenv?

Yes, it's relative to the uberenv.py script: https://uberenv.readthedocs.io/en/latest/#build-configuration

scripts/spack_packages/packages/geosx/package.py
```diff
@@ -55,41 +50,40 @@ class Geosx(CMakePackage, CudaPackage):
...
-    version('develop', branch='develop', submodules='True')
+    # GEOSX needs submodules to build, but not necessary to build dependencies
+    version('develop', branch='develop')
 
     # SPHINX_BEGIN_VARIANTS
 
     variant('shared', default=True, description='Build Shared Libs.')
     variant('caliper', default=True, description='Build Caliper support.')
-    variant('mkl', default=False, description='Use the Intel MKL library.')
-    variant('essl', default=False, description='Use the IBM ESSL library.')
```
wrtobin, line 65: Do we not have the virtual blas / lapack dependencies? Or are those only in our current host-configs to be used in building our TPLs (e.g. hypre)?

bmhan12: > Do we not have the virtual blas / lapack dependencies? Or are those only in our current host-configs to be used in building our TPLs (e.g. hypre)?

They are specified in the environment `spack.yaml` files like so for blueos:

```
 packages:
    all:
      compiler: [gcc, clang]
      providers:
        mpi: [spectrum-mpi]
        blas: [essl]
        lapack: [essl]

...
    essl:
      buildable: False
      externals:
        - spec: essl@6.3.0.2
          prefix: /usr/tcetmp/packages/essl/essl-6.3.0.2/
```

wrtobin: Yeah but we don't directly depend on them? e.g. I don't see `depends("blas")` in our spack recipe. So presumably we only depend on them transitively through e.g. hypre.

bmhan12: > Yeah but we don't directly depend on them? I don't see depends("blas") in our spack recipe. So presumably we only depend on them transitively through e.g. hypre.

Ah, we still do (wasn't sure so went and double-checked, as you're right, it could have been pulled in from a difference dependency). There is a "virtual packages" section with mpi/blas/lapack in the original recipe that has remained unchanged: https://github.com/GEOS-DEV/GEOS/blob/00ae74bb80f2a50f536a95036051b5f20592b425/scripts/spack_packages/packages/geosx/package.py#L87-L92

wrtobin (approved):
Are you planning to do any work on getting CI/CD setup for our spack+uberenv builds on LC systems? 

I have a branch I was working on in the TPL repo for the tioga spack build (plus verifying the current cmake TPL builds on LC systems as well). 

I was thinking we do the TPL CI (even spack+uberenv) in that repo until we're ready to fully switch to spack, then obsolete that repo and merge the yaml into the main geos repo?

bmhan12:
> Are you planning to do any work on getting CI/CD setup for our spack+uberenv builds on LC systems?

I was planning on getting the rest of the compilers working on LC systems next.
After that would be getting a spack/uberenv build script working on LC systems,
then iterating on spack/uberenv for the Docker images.

>I have a branch I was working on in the TPL repo for the tioga spack build (plus verifying the current cmake TPL builds on LC >systems as well).

👍 

>I was thinking we do the TPL CI (even spack+uberenv) in that repo until we're ready to fully switch to spack, then obsolete >that repo and merge the yaml into the main geos repo?

Yes, I think doing the spack changes in the TPL CI repository is the way to go.

rrsettgast:
@bmhan12 It seems like we are adding a lot of customized `package.py` files that should we shouldn't have in our repo. Can't we get those changes into the spack repo?

I don't think we want to be on the hook for maintaining these package files??

src/coreComponents/unitTests/fieldSpecificationTests/CMakeLists.txt
```diff
@@ -13,6 +13,22 @@ else()
...
+  list( APPEND dependencyList pugixml )
+endif()
+
+if (TARGET fmt::fmt)
+  list( APPEND dependencyList fmt::fmt )
+endif()
+
+if (TARGET fmt)
+  list( APPEND dependencyList fmt )
+endif()
```
rrsettgast, line 30: @bmhan12 Why are there scoped and unscoped dependencies?

bmhan12: > Why are there scoped and unscoped dependencies?

I found that the spack and the tplMirror built versions have different scopes for these two packages. This was added to support both build types.

PR #2032, opened 2022-08-26
Use real32 floats in WaveSolver

XL64:
This PR change real64 to real32 in waveSolvers

Rebaseline done in https://github.com/GEOSX/integratedTests/pull/266


acitrain (approved):

src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEM.hpp
```diff
@@ -203,63 +203,63 @@ namespace extrinsicMeshData
 
 EXTRINSIC_MESH_DATA_TRAIT( Pressure_nm1,
                            "pressure_nm1",
-                           array1d< real64 >,
+                           real32_array,
```
sframba, line 206: Question: why `real32_array` and not `array1d< real32 >`?

XL64: There is no particular reason, they are aliases. I could have used either, do you want me to change it ?

TotoGaz: Yes please, we tend not to use those aliases too much.

sframba (approved):

rrsettgast:
Are you always going to want `real32` for these variables? i.e. Are there cases where you would want to keep `real64`? If so, perhaps it makes sense to template `AcousticWaveEquationSEM` on the floating point type so you can easily have both?

XL64:
> 

We should not need real64 precision.

sframba:
image
Comparison between real32 and real32. The results are identical within numerical precision.

PR #2033, opened 2022-08-27
Added support for net-to-gross

francoishamon:
This PR adds the support of the old-fashioned net-to-gross property found in many reservoir models (multiplier for the porosity and the horizontal permeability). Although this could also be done externally, I thought it would be useful for users who may not know how net-to-gross works to have this inside GEOSX.

In passing, the PR also adds the `PorosityExtrinsicData` (not used yet, but added by Ryan for stabilisation and by me for norm calculation in other open PRs).

Requires rebaseline. Rebaseline done in https://github.com/GEOSX/integratedTests/pull/259


src/coreComponents/constitutive/solid/porosity/PorosityBase.cpp
```diff
@@ -71,45 +65,36 @@ void PorosityBase::allocateConstitutiveData( dataRepository::Group & parent,
...
 {
-  localIndex const numE = numElem();
-  localIndex const numQ = numQuad();
-
-  arrayView2d< real64 const > newPorosity     = m_newPorosity;
-  arrayView2d< real64 >       porosity_n      = m_porosity_n;
-  arrayView2d< real64 >       initialPorosity = m_initialPorosity;
+  arrayView2d< real64 const > const newPorosity = m_newPorosity;
+  arrayView2d< real64 > const porosity_n  = m_porosity_n;
+  porosity_n.setValues< parallelDevicePolicy<> >( newPorosity );
```
TotoGaz, line 88: Do you need the `setValues` to be on `views`?
otherwise, maybe
```c++
  m_porosity_n.setValues< parallelDevicePolicy<> >( m_newPorosity.toViewConst() );
```
is enough?

src/coreComponents/constitutive/solid/porosity/PorosityBase.cpp
```diff
@@ -71,45 +65,36 @@ void PorosityBase::allocateConstitutiveData( dataRepository::Group & parent,
...
-      initialPorosity[k][q] = newPorosity[k][q];
-    }
-  } );
+void PorosityBase::initializeState() const
+{
+  arrayView2d< real64 const > const newPorosity = m_newPorosity;
+  arrayView2d< real64 > const porosity_n = m_porosity_n;
+  arrayView2d< real64 > const initialPorosity = m_initialPorosity;
+  porosity_n.setValues< parallelDevicePolicy<> >( newPorosity );
+  initialPorosity.setValues< parallelDevicePolicy<> >( newPorosity );
```
TotoGaz, line 97: Same

src/coreComponents/constitutive/permeability/PermeabilityBase.hpp
```diff
@@ -114,17 +114,36 @@ class PermeabilityBase : public ConstitutiveBase
...
 
+  /**
+   * @brief Const/non-mutable accessor for dPerm_dPressure.
+   * @return Accessor
+   */
   arrayView3d< real64 const > dPerm_dPressure() const { return m_dPerm_dPressure; }
 
+  /**
+   * @brief Utility function to scale the horizontal permeability (for instance, by net-to-gross)
+   * @param[in] scalingFactors the vector of scaling factors for the horizontal permeability
```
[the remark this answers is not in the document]
francoishamon: Done, I added a doxygen comment to say that `scalingFactor` has one value by cell

src/coreComponents/constitutive/solid/CoupledSolidBase.hpp
```diff
@@ -127,6 +127,13 @@ class CoupledSolidBase : public ConstitutiveBase
   arrayView2d< real64 const > const  getDinternalEnergy_dTemperature() const
   { return getSolidInternalEnergyModel().getDinternalEnergy_dTemperature(); }
 
+  /*
+   * @brief Utility function to scale the reference porosity (for instance, by net-to-gross)
+   * @param[in] scalingFactors the vector of scaling factors for the reference porosity
```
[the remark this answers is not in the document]
francoishamon: Same here, I added a Doxygen comment to say that `scalingFactors` has one value per cell

src/coreComponents/constitutive/solid/porosity/PorosityBase.cpp
```diff
@@ -71,45 +65,36 @@ void PorosityBase::allocateConstitutiveData( dataRepository::Group & parent,
...
+  getExtrinsicData< extrinsicMeshData::porosity::referencePorosity >().
+    setApplyDefaultValue( m_defaultReferencePorosity );
 }
 
-
-void PorosityBase::saveConvergedState() const
+void PorosityBase::scaleReferencePorosity( arrayView1d< real64 const > scalingFactors ) const
 {
   localIndex const numE = numElem();
-  localIndex const numQ = numQuad();
```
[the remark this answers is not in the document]
francoishamon: Here, the GitHub diff is misleading: it shows that I removed the quadrature, but in fact I just simplified `saveConvergedState` using `setValues`, and I added `scaleReferencePorosity` which does need the number of quadrature points

src/coreComponents/constitutive/solid/porosity/PorosityBase.cpp
francoishamon: You are right, I implemented what you suggested

src/coreComponents/constitutive/solid/porosity/PorosityBase.cpp
francoishamon: Same here, you are right, I implemented what you suggested

src/coreComponents/constitutive/permeability/PermeabilityBase.cpp
```diff
@@ -44,6 +44,20 @@ PermeabilityBase::deliverClone( string const & name,
   return ConstitutiveBase::deliverClone( name, parent );
 }
 
+void PermeabilityBase::scaleHorizontalPermeability( arrayView1d< real64 const > scalingFactors ) const
+{
+  localIndex const numElems = m_permeability.size( 0 );
+  for( localIndex ei = 0; ei < numElems; ++ei )
+  {
+    // NOTE: enforcing 1 quadrature point
```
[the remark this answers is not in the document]
francoishamon: The permeability models are all implemented with one quadrature point because there is really no need to use the default behavior (number of quadrature points decided by the finite-element scheme in coupled simulations). This is done in `allocateConstitutiveData` below:
https://github.com/GEOSX/GEOSX/blob/d90763008ffda699a0fdcd094e43ed69da9626c4/src/coreComponents/constitutive/permeability/PermeabilityBase.cpp#L64-L66

So in the new code, I just followed this. After your comment, I modified the new loop to have something cleaner in case someone decides to increase the number of quadrature points:
https://github.com/GEOSX/GEOSX/blob/e5c09cc8da5d3029e044be3c202e5cf65d3ce597/src/coreComponents/constitutive/permeability/PermeabilityBase.cpp#L51-L58 
where `numQuad` is the second dimension of `m_permeability`

TotoGaz (approved):

src/coreComponents/constitutive/solid/porosity/PorosityBase.cpp
```diff
@@ -71,45 +65,31 @@ void PorosityBase::allocateConstitutiveData( dataRepository::Group & parent,
...
-    for( localIndex q = 0; q < numQ; ++q )
-    {
-      porosity_n[k][q]      = newPorosity[k][q];
-      initialPorosity[k][q] = newPorosity[k][q];
-    }
-  } );
+void PorosityBase::initializeState() const
+{
+  m_porosity_n.setValues< parallelDevicePolicy<> >( m_newPorosity.toViewConst() );
+  m_initialPorosity.setValues< parallelDevicePolicy<> >( m_newPorosity.toViewConst() );
```
rrsettgast, line 92: Note you are launching two kernels here. Maybe we need a `setValues` that can perform a batch.

rrsettgast (approved):

src/coreComponents/constitutive/solid/porosity/PorosityBase.cpp
francoishamon: Yes, that would be very useful to have a batch `setValues`. I can go back to the `forAll` loop in a future PR.

klevzoff: Realistically though, what we need is an array copying interface that would dispatch to `memcpy`/`cudaMemcpy` when types/shapes match

PR #2036, opened 2022-08-30
Remove hardcoded name "faceElementSubRegion" and "embeddedSurfaceSubRegion".

TotoGaz:
Part of the https://github.com/GEOSX/GEOSX/issues/2007 effort

Rebaseline: https://github.com/GEOSX/integratedTests/pull/252


src/coreComponents/mesh/SurfaceElementRegion.cpp
```diff
@@ -34,15 +34,14 @@ SurfaceElementRegion::SurfaceElementRegion( string const & name, Group * const p
     setApplyDefaultValue( m_subRegionType ).
     setDescription( "Type of surface element subregion. Valid options:\n* " + EnumStrings< SurfaceSubRegionType >::concat( "\n* " ) );
 
+  registerWrapper( viewKeyStruct::faceBlockString(), &m_faceBlockName ).
+    setInputFlag( InputFlags::OPTIONAL ).
+    setDefaultValue( "FractureSubRegion" ).
```
TotoGaz, line 39: @CusiniM I've set the default to some value such that we do not have to define the field in all `xmls`.
Do you think that this is a good idea?
If you want to define some BC in the `xml`, it's better than you impose your own "fracture" name (instead of using the default value which would break the `xml` input if it changes)...
What do you think?

CusiniM: Well, in reality, BC can be applied just by knowing the `RegionName`. I usually do not use `subRegions` to specify BC. So, I can't of like not having to modify everything.

TotoGaz: Yes, I was imprecise when mentioning BC, I meant `FieldSpecification`.
For instance, in all the files I've modified, you'll find things like
```
    <FieldSpecification
      name="initialFracturePressure"
      initialCondition="1"
      setNames="{ all }"
      objectPath="ElementRegions/Fracture/embeddedSurfaceSubRegion"
      fieldName="pressure"
      scale="5e6"/>
```
which point to `embeddedSurfaceSubRegion`. (I know it should not, but...). Maybe use `MeshObjectPath`?

CusiniM: so, in theory, just `ElementRegions/Fracture/` should work.

TotoGaz: It does not on my `geosx`... Does it work on your version?

CusiniM (approved):
Looks good.

Issue #2029, opened 2022-08-25
Rollback azure pipeline, fine grained usage of travis VMs, travis pipeline cleaning


TotoGaz:
It appears that we were not using the VM sizes we were paying `travis-ci` for.

- [x] Use the `large` machines for all the `travis` pipeline. Done in https://github.com/GEOSX/GEOSX/pull/2025
- [x] Rollback all the azure-pipeline actions: `git` history will allow to get back to the files if we want to go back to it. Done in https://github.com/GEOSX/GEOSX/pull/2025
- [x] We could add the `check_pr_is_assigned` in its own initial step such that we do not initiate jobs for nothing (100% waste). Done in https://github.com/GEOSX/GEOSX/pull/2053
- [x] https://config.travis-ci.com/explore offers an interactive way to test the `.travis.yaml` file, so we should endeavor to remove the warnings or informative messages (it should be a tiny action). Done in https://github.com/GEOSX/GEOSX/pull/2053
- [x] Try to use specific sizes for each steps (tipically, `check` steps sould run on `medium` VMs, and maybe running the GPUs on `x-large`?). I'm not sure that it's indeed possible, but https://config.travis-ci.com/explore seemed to consider it as an option, so let's try 🤞

bmhan12:
> Try to use specific sizes for each steps...

Yes, a [github search](https://github.com/search?q=filename%3A.travis.yml+large) shows size can be adjusted for each job launch.

TotoGaz:
I think we should rush https://github.com/GEOSX/GEOSX/pull/2025 for immediate relief of our build queues (tick box one).
Maybe we can merge it as is and do a better cleaning after?

TotoGaz:
@rrsettgast I'm trying to finish the last tick box.
Whatever I may do in the config file, I stick to `large` VMs.

Do you think we could ask the travis support to stop our specific VM choice so we may be able to better configure them?
According to our current `.travis.yml`, it should be equivalent and seamless.

Actually, even if this does not allow a fine grain configuration, I also prefer relying on explicit documentation than on hidden backend configurations.

rrsettgast:
The use of larger VM's did not work.

TotoGaz:
> The use of larger VM's did not work.

I know:

> Whatever I may do in the config file, I stick to large VMs.

But is it because the `travis` support imposes a specific VM size like they used to do when we couldn't chose the VM sizes ? Or is it because it's impossible?

[...]

PR #2037, opened 2022-08-30
Rollback Azure Pipelines Jobs To Travis

### LINKED ISSUE #1981: [Feature] Azure Pipelines Follow-Up
Follow-up work for #1969 .

Currently only the linux Docker jobs have been migrated to azure pipelines. 

Still have osx and total docker jobs to migrate over from Travis CI.

bmhan12:
This PR:
- Moves Azure jobs back to Travis
- Deletes Azure configuration file
- Adds a `--reduce-install-logs` flag to do a `make install` instead of ` make install VERBOSE=1` to preventing hitting [log length error](https://docs.travis-ci.com/user/common-build-problems/#log-length-exceeded).
- Relates to #2029, Closes #1981


scripts/build_test_helper.sh
```diff
@@ -78,7 +78,13 @@ if [[ "$*" == *--build-exe-only* ]]; then
...
 else
   or_die make -j $(nproc) VERBOSE=1
-  or_die make install VERBOSE=1
+
+  # Verbosity check for installation to prevent hitting Travis log limit
+  if [[ "$*" == *--reduce-install-logs* ]]; then
+    or_die make install
+  else
+    or_die make install VERBOSE=1
+  fi
```
TotoGaz, line 87: I like how those `if [[ "$*" == *...* ]];` reflect all the issues we happen to face during the project 😉

TotoGaz (approved):
I think that we could remove the `ci_build_and_test.sh` because it induces duplications and is now not really required.
We can do it in other PRs down the line.

TotoGaz:
@bmhan12 Do you want I merge this right now?
I think it makes sense.

bmhan12:
> @bmhan12 Do you want I merge this right now?

@TotoGaz Yes, that would be great.

PR #2039, opened 2022-08-31
Added Newton global damping by default in compositional solvers

francoishamon:
Jacques and I have some difficult cases in which the residual grows uncontrollably when the time step size is too large, which ends up crashing the simulation (line search does not fix the problem). To overcome the problem, this PR does two things:
- [X] It turns on global damping on component density, pressure, and temperature by default (50% change allowed between iterations).
- [X] It introduces a `maxAllowedResidualNorm` attribute in `SolverBase` set by default to 1e9. If the residual norm is above this value, the time step is cut to stop Newton divergence.

In passing, the PR unifies the solution scaling and solution check kernels in the flow/well solvers and adds some logging.

Requires rebaseline. Rebaseline done in https://github.com/GEOSX/integratedTests/pull/257


src/coreComponents/physicsSolvers/fluidFlow/ThermalCompositionalMultiphaseBaseKernels.hpp
```diff
@@ -494,6 +494,269 @@ struct SolidInternalEnergyUpdateKernel
...
+                   string const dofKey,
+                   ElementSubRegionBase const & subRegion,
+                   arrayView1d< real64 const > const localSolution )
+  {
+    arrayView1d< real64 const > const pressure =
+      subRegion.getExtrinsicData< extrinsicMeshData::flow::pressure >();
+    arrayView1d< real64 const > const temperature =
+      subRegion.getExtrinsicData< extrinsicMeshData::flow::temperature >();
+    arrayView2d< real64 const, compflow::USD_COMP > const compDens =
+      subRegion.getExtrinsicData< extrinsicMeshData::flow::globalCompDensity >();
```
TotoGaz, line 628: There also, these statements can surely go on one line.

src/coreComponents/physicsSolvers/fluidFlow/ThermalCompositionalMultiphaseBaseKernels.hpp
```diff
@@ -494,6 +494,269 @@ struct SolidInternalEnergyUpdateKernel
...
+  void computeSolutionCheck( localIndex const ei,
+                             StackVariables & stack ) const
+  {
+    Base::computeSolutionCheck( ei, stack, [&] ()
+    {
+      // compute the change in temperature
+      real64 const newTemp = m_temperature[ei] + m_scalingFactor * m_localSolution[stack.localRow + m_numComp + 1];
+      if( newTemp < minTemperature )
+      {
+        stack.localMinVal = 0;
```
TotoGaz, line 704: Why not `stack.localMinVal = newTemp`?

src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseBaseKernels.hpp
```diff
@@ -914,98 +915,504 @@ class ElementBasedAssemblyKernelFactory
...
+  {}
+
+  /**
+   * @struct StackVariables
+   * @brief Kernel variables located on the stack
+   */
+  struct StackVariables
   {
-    RAJA::ReduceSum< ReducePolicy< POLICY >, real64 > localSum( 0.0 );
+public:
```
TotoGaz, line 961:
```suggestion
```

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseHybridFVM.cpp
```diff
@@ -556,53 +495,42 @@ bool CompositionalMultiphaseHybridFVM::checkSystemSolution( DomainPartition cons
 {
   GEOSX_MARK_FUNCTION;
 
-  // 1. check cell-centered variables for each region
   string const elemDofKey = dofManager.getKey( viewKeyStruct::elemDofFieldString() );
-  localIndex localCheck = 1;
+  integer localCheck = 1;
```
jafranc, line 499: IIUC in `XXX::SolutionCheckKernel` the value for new pressure is checked to be non-negative ? Could add that as a comment to know what is checked.

src/coreComponents/physicsSolvers/fluidFlow/ThermalCompositionalMultiphaseBaseKernels.hpp
```diff
@@ -494,6 +494,269 @@ struct SolidInternalEnergyUpdateKernel
...
+    real64 constexpr eps =
+      isothermalCompositionalMultiphaseBaseKernels::minDensForDivision;
+
+    Base::computeScalingFactor( ei, stack, [&] ()
+    {
+      // compute the change in temperature
+      real64 const temp = m_temperature[ei];
+      if( temp > eps )
+      {
+        real64 const absTempChange = LvArray::math::abs( m_localSolution[stack.localRow + m_numComp + 1] );
```
jafranc, line 569: As we repeat the scaling operation quite a few time over the classes, is there a solution in which it can be factorized and in which `m_maxRelativeXChange`, `local_offfset` and `m_X` are picked from argument ?

jafranc (approved):

src/coreComponents/physicsSolvers/fluidFlow/ThermalCompositionalMultiphaseBaseKernels.hpp
jafranc: is there a throwing strategy on the `0` value ? as negative temperature is cut-off at 0 as well as pressure is IIRC

src/coreComponents/physicsSolvers/NonlinearSolverParameters.cpp
```diff
@@ -74,6 +74,11 @@ NonlinearSolverParameters::NonlinearSolverParameters( string const & name,
     setApplyDefaultValue( 0 ).
     setDescription( "Number of Newton's iterations." );
 
+  registerWrapper( viewKeysStruct::maxAllowedResidualNormString, &m_maxAllowedResidualNorm ).
+    setApplyDefaultValue( 1e9 ).
```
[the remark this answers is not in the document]
francoishamon: In fact we don't want something as large as the max value of `real64` because we want the solver to cut the time step before the residual norm reaches these very large values. 

Here, I chose `1e9` because some solvers (mostly the `SurfaceGenerator`) commonly have a very large residual norm, but for typical field-scale reservoir simulations `1e9` is already too large, so I would probably use the `maxAllowedResidualNorm` to reduce this value by hand.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.cpp
```diff
@@ -256,19 +256,16 @@ real64 CompositionalMultiphaseFVM::scalingForSystemSolution( DomainPartition con
 {
   GEOSX_MARK_FUNCTION;
 
+
   // check if we want to rescale the Newton update
-  if( m_maxCompFracChange >= 1.0 )
+  if( m_maxCompFracChange >= 1.0 ||
+      m_maxRelativePresChange >= 1.0 ||
+      ( m_maxRelativeTempChange >= 1.0 || !m_isThermal ) )
```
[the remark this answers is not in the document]
francoishamon: Yes, done. I added some named boolean variables to clarify what I am trying to do:
https://github.com/GEOSX/GEOSX/blob/69e0c7e56e36e755218fc5e7d2adcfc24e504d85/src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.cpp#L259-L268
and thanks for your comment, because there was a bug in my initial check

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWellKernels.hpp
```diff
@@ -750,146 +751,90 @@ struct ResidualNormKernel
...
-            minVal.min( 0 );
-          }
-        }
-      }
-    } );
-    return minVal.get();
+    arrayView1d< real64 const > const pressure =
+      subRegion.getExtrinsicData< extrinsicMeshData::well::pressure >();
+    arrayView2d< real64 const, compflow::USD_COMP > const compDens =
+      subRegion.getExtrinsicData< extrinsicMeshData::well::globalCompDensity >();
```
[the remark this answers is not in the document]
francoishamon: Sorry, I work on a small screen, so I don't like long lines :) . But I put everything on the same line in the latest commit

src/coreComponents/physicsSolvers/fluidFlow/ThermalCompositionalMultiphaseBaseKernels.hpp
```diff
@@ -494,6 +494,269 @@ struct SolidInternalEnergyUpdateKernel
...
+    m_maxRelativeTempChange( maxRelativeTempChange ),
+    m_temperature( temperature )
+  {}
+
+  /**
+   * @brief Compute the local value of the scaling factor
+   * @tparam FUNC the type of the function that can be used to customize the kernel
+   * @param[in] ei the element index
+   * @param[inout] stack the stack variables
+   * @param[in] kernelOp the function used to customize the kernel
```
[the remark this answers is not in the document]
francoishamon: I had messed up the Doxygen here, this is fixed

src/coreComponents/physicsSolvers/fluidFlow/ThermalCompositionalMultiphaseBaseKernels.hpp
```diff
@@ -494,6 +494,269 @@ struct SolidInternalEnergyUpdateKernel
...
+   */
+  GEOSX_HOST_DEVICE
+  void computeScalingFactor( localIndex const ei,
+                             StackVariables & stack ) const
+  {
+    real64 constexpr eps =
+      isothermalCompositionalMultiphaseBaseKernels::minDensForDivision;
```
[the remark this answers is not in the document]
francoishamon: I put everything on the same line

src/coreComponents/physicsSolvers/fluidFlow/ThermalCompositionalMultiphaseBaseKernels.hpp
francoishamon: Same here

src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseBaseKernels.hpp
francoishamon: I removed the `public` keyword

src/coreComponents/physicsSolvers/fluidFlow/ThermalCompositionalMultiphaseBaseKernels.hpp
francoishamon: In fact here, `minVal` is an integer flag initialized to 1, and set to zero if the solution is invalid. It may not be very clear I agree, but I could not find a way to work with `bool` in the `RAJA::ReduceMin`.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.cpp
```diff
@@ -280,56 +277,36 @@ real64 CompositionalMultiphaseFVM::scalingForSystemSolution( DomainPartition con
...
+                                                     localSolution )
+  : isothermalCompositionalMultiphaseBaseKernels::
+          ScalingForSystemSolutionKernelFactory::
+          createAndLaunch< parallelDevicePolicy<> >( m_maxRelativePresChange,
+                                                     m_maxCompFracChange,
+                                                     dofManager.rankOffset(),
+                                                     m_numComponents,
+                                                     dofKey,
+                                                     subRegion,
+                                                     localSolution );
```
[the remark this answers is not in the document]
francoishamon: In fact, the dispatch is here to launch the appropriate kernel (isothermal vs thermal). The isothermal and the thermal kernel have the same name, but they live in different namespaces. So even if we pass a default value for `m_maxRelativeTempChange`, we will still need the dispatch.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.cpp
```diff
@@ -350,30 +327,32 @@ bool CompositionalMultiphaseFVM::checkSystemSolution( DomainPartition const & do
...
+          SolutionCheckKernelFactory::
+          createAndLaunch< parallelDevicePolicy<> >( m_allowCompDensChopping,
+                                                     scalingFactor,
+                                                     dofManager.rankOffset(),
+                                                     m_numComponents,
+                                                     dofKey,
+                                                     subRegion,
+                                                     localSolution );
 
       localCheck = std::min( localCheck, subRegionSolutionCheck );
```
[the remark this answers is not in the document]
francoishamon: Yes, you are right, in the latest commit I used `std::min` everywhere for consistency

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.cpp
```diff
@@ -280,56 +277,36 @@ real64 CompositionalMultiphaseFVM::scalingForSystemSolution( DomainPartition con
...
       {
-        scalingFactor = minVal.get();
+        scalingFactor = subRegionScalingFactor;
       }
     } );
   } );
 
-  return LvArray::math::max( MpiWrapper::min( scalingFactor, MPI_COMM_GEOSX ), m_minScalingFactor );
+  return LvArray::math::max( MpiWrapper::min( scalingFactor ), m_minScalingFactor );
 }
```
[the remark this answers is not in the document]
francoishamon: I am a little bit reluctant to putting the MPI reduction in the `subRegion` loop because in some cases we have many `subRegions` and this may cause a lot of communications. That is why I always put the MPI reduction at the end of the function, outside the loops, even if it is not very pretty...

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseHybridFVM.cpp
francoishamon: Yes, you are right, I put the comments that you suggested before all the kernel launches.

src/coreComponents/physicsSolvers/fluidFlow/ThermalCompositionalMultiphaseBaseKernels.hpp
francoishamon: Yes, we could have a utility function in the kernel class that does that for us to avoid some redundancy in the code. I have not implemented it yet, but I am going to think about a good way

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWellKernels.hpp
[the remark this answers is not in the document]
klevzoff: I also split these as they often break even generous column limits like 160.
IMHO this should be (in some distant and bright future)
```
    auto const compDens = subRegion.getField< fields::well::globalCompDensity >();
```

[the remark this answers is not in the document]
TotoGaz: > ```
>     auto const compDens = subRegion.getField< fields::well::globalCompDensity >().toViewConst();
> ```

I'm not really sure about the auto. It's always tedious to swim upstream to find what the type you're manipulating!
But `field` instead of `extrinsicData` is a happy 14 characters reduction. 🎉 (plus I understand).
We could have a `getFieldV/getFieldView` and `getFieldCV/getFieldConstView` for the `views` and `const views`?

> in some distant and bright future

You say this because we won't be there! 🤣 🤣 🤣

PR #2043, opened 2022-09-01
Changed treatment of gravity in poromechanics kernels

### LINKED ISSUE #2041: [Bug] Poromechanics problems with gravity do not produce expected stress field
**Describe the bug**
I noticed some odd behavior in a model with poromechanics + gravity, so I put together a small test problem.  In the example, I have a column of elements, with rollers on the +/-x, +/-y, and -z faces, and a constant pressure (0) on the z+ surface.  The pressure and stress fields are initialized with values from a table (the python script used to build these is included).

The initial stress, pressure, and displacement in the model are:

initial_stress_displacement

Applying either the solid mechanics or fluid solvers *alone*, yields correct looking results:

final_stress_displacement_mechanics_only

However, when the poroelastic solver is applied, the stress/displacement fields do not make sense:

final_stress_displacement_poroelastic

**To Reproduce**

See the attached example here.  In the event block, you can change which solvers are applied to the problem and see how the values change.

[poroelastic_gravity.zip](https://github.com/GEOSX/GEOSX/files/9465570/poroelastic_gravity.zip)

francoishamon:
This PR changes the formulation of the body force and pressure terms in the poromechanics kernels (single-phase and multiphase) to move away from the delta formulation (delta pressure, delta body force, stress interpreted as a delta from an initial stress). I am going to come up with a simple validation test that can also be added to the integrated tests, and then ask for a review.

The PR also removes some constitutive fields that are no longer needed after the change: 
- `initialDensity` in `SingleFluidBase`
- `initialTotalMassDensity` in `MultiFluidBase`

Resolves #2041 

Requires rebaseline


CusiniM:
I would use the case created by @cssherman in https://github.com/GEOSX/GEOSX/issues/2041 as an integratedTest.

francoishamon:
> I would use the case created by @cssherman in #2041 as an integratedTest.

Sorry for the delay, I just added the test created by Chris to this PR.

The change proposed in this PR will also affect no-gravity poromechanics simulations because it changes the formulation of the total stress in the code. Specifically, the pressure term is no longer a pressure increment (but the actual pressure value) and the body force term is no longer an increment either. 

So, in the poromechanics simulations with this PR, the `stress` variable of the xml file is an effective stress that does not include the pressure term, and we have to initialize it properly when the simulation has either gravity and/or a non-zero initial pressure to start from an equilibrium. In our integrated tests and validation tests, I added this effective stress initialization where it was missing to make sure that we don't have numerical diffs. The only thing that I was not able to do is [changing this script](https://geosx-geosx.readthedocs-hosted.com/en/latest/docs/sphinx/advancedExamples/validationStudies/wellboreProblems/deviatedPoroElasticWellbore/deviatedPoroElasticWellboreExample2Figure.py) and its input data because it was generated outside GEOSX. 

The main advantage is that the stress computed by GEOSX is not anymore a delta with respect an equilibrium configuration, but now has a meaningful value that users can use in their analyses. But I may be wrong. Is everyone on board with the change?

TotoGaz (approved):

francoishamon:
If everyone is on board with the changes, this PR is ready to be merged.

src/docs/sphinx/advancedExamples/validationStudies/faultMechanics/faultVerification/Example.rst
```diff
@@ -2,7 +2,7 @@
 
 
 ###########################################################
-Verification of Induced Stresses Along a Fault
+Verification of Induced Sresses Along a Fault
```
jhuang2601, line 5:
```suggestion
Verification of Induced Stresses Along a Fault
```

src/coreComponents/constitutive/solid/PorousSolid.hpp
```diff
@@ -378,61 +368,52 @@ class PorousSolidUpdates : public CoupledSolidUpdates< SOLID_TYPE, BiotPorosity,
...
                          real64 const & dPorosity_dVolStrain,
                          real64 const & dPorosity_dPressure,
                          real64 const ( &gravityVector )[3],
-                         real64 ( & bodyForceIncrement )[3],
+                         real64 ( & bodyForce )[3],
                          real64 ( & dBodyForce_dVolStrainIncrement )[3],
                          real64 ( & dBodyForce_dPressure )[3],
                          real64 ( & dBodyForce_dComponents )[3][NUM_MAX_COMPONENTS] ) const
   {
     computeBodyForce( solidDensity,
```
jhuang2601, line 409: `bodyForce` is induced by gravity only. How about tectonic stress/strain? Any plan for these?

jhuang2601 (approved):

CusiniM (approved):
I personally prefer it this way. Isn't it also more consistent with what we do for the `stress` in general?

src/docs/sphinx/advancedExamples/validationStudies/faultMechanics/faultVerification/Example.rst
francoishamon: My bad, this change was not intentional. It is fixed now.

castelletto1 (approved):
Looks good to me. Thanks!

src/coreComponents/constitutive/solid/PorousSolid.hpp
francoishamon: For stress/strain, the formulation that we have in place works well and does not require any change. For field-scale models, we still have to test the poromechanical initialization from a given regional stress, but my understanding is that we have everything we need now (I may be wrong :) ).

francoishamon:
> I personally prefer it this way. Isn't it also more consistent with what we do for the `stress` in general?

Yes, with the previous formulation, stress had to be interpreted as a delta stress from an initial equilibrium. Now, the initial stress has to be provided to GEOSX (we have used PR #1927 to compute the initial stress in the field-scale simulations) and the stress computed by GEOSX is no longer a delta wrt to the initial state. This is easier when users want to visualize / analyze / post-process the stress variable.

PR #2044, opened 2022-09-02
Implemented small changes in OBL

francoishamon:
This PR makes small changes in OBL based on the feedback from Stephan de Hoop (and indirectly, from @mkhait I think) from TU Delft:
- [X] Added an update of the last component fraction to make sure that component fractions sum to one in the output files (during the simulation, we don't care since the last component fraction is not a primary variables)
- [X] Changed the default value of `rockKineticRateFactor` to 1, since the current default value of 0 was not appropriate for most cases.

In passing the PR removes some unused fields from the OBL kernels.

Requires a rebaseline. Rebaseline done in https://github.com/GEOSX/integratedTests/pull/258


src/coreComponents/physicsSolvers/fluidFlow/ReactiveCompositionalMultiphaseOBL.cpp
```diff
@@ -161,6 +161,21 @@ void ReactiveCompositionalMultiphaseOBL::implicitStepComplete( real64 const & ti
...
                                                      ElementSubRegionBase & subRegion )
     {
+      // for output purposes (visualization, etc) we update the last component fraction
+      integer const numComp = m_numComponents;
+      arrayView2d< real64, compflow::USD_COMP > const compFrac =
+        subRegion.getExtrinsicData< extrinsicMeshData::flow::globalCompFraction >();
+
+      forAll< parallelDevicePolicy<> >( subRegion.size(), [=] GEOSX_HOST_DEVICE ( localIndex const ei )
+      {
+        compFrac[ei][numComp-1] = 1.0;
```
TotoGaz, line 171: This passes `uncrustify`? I thought we needed `numComp-1`. However...

src/coreComponents/physicsSolvers/fluidFlow/ReactiveCompositionalMultiphaseOBL.cpp
```diff
@@ -161,6 +161,21 @@ void ReactiveCompositionalMultiphaseOBL::implicitStepComplete( real64 const & ti
                                                 [&]( localIndex const,
                                                      ElementSubRegionBase & subRegion )
     {
+      // for output purposes (visualization, etc) we update the last component fraction
```
TotoGaz, line 164: Actually, can you be more precise in this comment about what you actually do?
It looks like you enforce the sum to 1?
You could also have divided every component by the sum? Why only modifying the last slot?

TotoGaz (approved):

src/coreComponents/physicsSolvers/fluidFlow/ReactiveCompositionalMultiphaseOBL.cpp
francoishamon: We have to update the last component fraction only because it is never updated by `ApplySystemSolution` here:
https://github.com/GEOSX/GEOSX/blob/51a10972167b388766c2b130bfd99560a6f12af9/src/coreComponents/physicsSolvers/fluidFlow/ReactiveCompositionalMultiphaseOBL.cpp#L487-L493
In fact, this last component fraction is never used during the simulation and the purpose of this change is just to make sure that the output component fractions (say, in VTK or Time History) have the correct values and sum to one. All the other component fractions have already been updated by the Newton update in `ApplySystemSolution` and should not be modified at this point.

src/coreComponents/physicsSolvers/fluidFlow/ReactiveCompositionalMultiphaseOBL.cpp
francoishamon: Yes, this passes `uncrustify`. What change would you like to see here?

src/coreComponents/physicsSolvers/fluidFlow/ReactiveCompositionalMultiphaseOBL.cpp
```diff
@@ -339,7 +354,7 @@ real64 ReactiveCompositionalMultiphaseOBL::calculateResidualNorm( DomainPartitio
 
   if( getLogLevel() >= 1 && logger::internal::rank == 0 )
   {
-    std::cout << GEOSX_FMT( "    ( Rfluid ) = ( {:4.2e} ) ;", residual );
+    std::cout << GEOSX_FMT( "    ( Rflow ) = ( {:4.2e} ) ;", residual );
   }
```
rrsettgast, line 358: Couldn't this be replaced with `GEOSX_LOG_LEVEL_RANK_0`?


definitions:
```
#define GEOSX_LOG_LEVEL_RANK_0( minLevel, msg ) GEOSX_LOG_RANK_0_IF( this->getLogLevel() >= minLevel, msg )

#define GEOSX_LOG_RANK_0_IF( EXP, msg ) \
  do { \
    if( ::geosx::logger::internal::rank == 0 && EXP ) \
    { \
      std::ostringstream oss; \
      oss << msg; \
      std::cout << oss.str() << std::endl; \
    } \
  } while( false )

```

rrsettgast (approved):
Just one suggestion about a log statement

src/coreComponents/physicsSolvers/fluidFlow/ReactiveCompositionalMultiphaseOBL.cpp
francoishamon: Yes, sounds good, I made the modification.

I think we have used `if( getLogLevel() >= 1 && logger::internal::rank == 0 )` instead of `GEOSX_LOG_LEVEL_RANK_0` in all single-physics solvers to have a concise output in our simulation logs (taking the example of a poromechanics simulation with wells):
```
Attempt:  0, ConfigurationIter:  0, NewtonIter:  3                                                                                                                          
( Rsolid ) = ( 1.41e-13 ) ;     ( Rflow ) = ( 5.45e-03 ) ;     ( Rwell ) = ( 3.27e-07 ) ;     ( R ) = ( 5.45e-03 ) ;  
```
instead of:
```
Attempt:  0, ConfigurationIter:  0, NewtonIter:  3                                                                                                                          
( Rsolid ) = ( 1.41e-13 ) ;
( Rflow ) = ( 5.45e-03 ) ;
( Rwell ) = ( 3.27e-07 ) ;
( R ) = ( 5.45e-03 ) ;  
```
where the line breaks come from the `std::endl` in GEOSX_LOG_LEVEL_RANK_0`. 

But maybe we should just let the coupled solver issue the residual norm values in a uniform way [here](https://github.com/GEOSX/GEOSX/blob/cd04f12c24a0004432673ce5c667512175c95ffd/src/coreComponents/physicsSolvers/multiphysics/CoupledSolver.hpp#L245) and remove all this norm logging from the single-physics solvers. This would make it easier to work on a more concise/organized formatting (maybe with a table or something like that).

klevzoff: I may or may not have said this before, but in my opinion the appropriate place to do residual norm logging is in `SolverBase::solveNonlinearSystem`, in the middle of Newton loop, not the calculation functions. If we need to modify the signature of `calculateResidualNorm` to return multiple values (a vector of sub-norms), let's do it. Otherwise we get into a nesting problem: what if a sub-solver of a coupled solver is itself a coupled solver? Only the solver that is executing the nonlinear loop has the correct information about what/when/how to print (are we in a middle of a line search? what level of detail has the user requested?). Also single responsibility principle suggests that a compute function should not also do logging.

PR #2048, opened 2022-09-06
split fracture stencil and trigger fracture stencil for imported faults

TotoGaz:


src/coreComponents/finiteVolume/TwoPointFluxApproximation.cpp
```diff
@@ -247,74 +252,10 @@ void TwoPointFluxApproximation::addToFractureStencil( MeshLevel & mesh,
...
-#if SET_CREATION_DISPLACEMENT==1
-    if( initFlag )
-    {
-      aperture[kfe] = 1.0e99;
-    }
-#endif
-  } );
-
-#endif
   SortedArray< localIndex > allNewElems;
```
CusiniM, line 255: is this useful?

CusiniM (approved):

src/coreComponents/finiteVolume/TwoPointFluxApproximation.cpp
TotoGaz: It's some code for initialization.
It should not be there anymore.
I've removed it!

PR #2050, opened 2022-09-07
Skipped the shallow mesh copies in the vtk output

francoishamon:
I am creating a PR for this fix proposed by @klevzoff that hopefully I haven't messed up. I am not sure it was meant to be merged, but it turned out to be very useful to make sure that the VTK output is still usable for our field-scale examples.

With the current `develop`, we have large VTK files in which the information is duplicated because we output the shallow copies of the mesh to the file:
Screen Shot 2022-09-07 at 4 38 44 PM

After this PR, we get back to:
Screen Shot 2022-09-07 at 9 45 28 PM

The integrated tests pass, and the `testMultipleBodies.xml` test seems to produce a valid VTK output as well. But I don't mind closing the PR if someone is already working on a proper fix (I missed the discussion on this issue).

### DIFF
--- a/src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp
+++ b/src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp
@@ -1000,6 +1000,12 @@ void VTKPolyDataWriterInterface::writeVtmFile( integer const cycle,
   {
     meshBody.forMeshLevels( [&]( MeshLevel const & meshLevel )
     {
+
+      if( meshLevel.isShallowCopy() )
+      {
+        return;
+      }
+
       ElementRegionManager const & elemManager = meshLevel.getElemManager();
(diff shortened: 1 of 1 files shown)


andrembcosta:
I did not notice this when adding the MeshLevel loops. Thanks for catching it and fixing it promptly Francois.

andrembcosta (approved):

klevzoff (approved):

klevzoff:
> I did not notice this when adding the MeshLevel loops. Thanks for catching it and fixing it promptly Francois.

You couldn't have noticed because shallow mesh levels weren't a thing yet, they were added in another PR that was developed in parallel and merged right before yours. If anything, I messed up by not re-testing your PR after I merged `develop` into it.

rrsettgast:
While I appreciate everyone trying to take responsibility, this is clearly my bad. The logic for shallow `MeshLevel` is the offending concept, and the output of shallow `MeshLevel` was not implemented properly. This does highlight a great team we have!

PR #2055, opened 2022-09-08
Fix submodules quick start doc.

TotoGaz:
The submodules paths do not match the doc anymore.
https://github.com/GEOSX/GEOSX/blob/develop/.gitmodules


jafranc (approved):
Looks great !

PR #2059, opened 2022-09-12
Feature/andrembcosta/pressurized phase field

andrembcosta:
Adding new kernels for phase-field fracture model in a pressurized environment. This is similar to Frank's work, but the pressure field are assumed to be constant (they will be solved for using EFEM in a global domain and this model will only be applied in the vicinity of crack tips).

This PR add kernels for the terms in red (green terms are the standard phase-field equations). Terms in the first equation are in SolidMechanicsSmallStrainQuasiStaticPressureKernel.hpp, whereas the terms in the second equation are in PhaseFieldDamagePressureFEMKernels.hpp.

image

The kernels are written using the PDEUtilities functions from https://github.com/GEOSX/GEOSX/pull/1600. New Linear and BilinearFormUtilities were added for H1 spaces.


src/coreComponents/constitutive/solid/Damage.hpp
```diff
@@ -182,11 +288,20 @@ class DamageUpdates : public UPDATE_BASE
...
+  }
+
+  arrayView2d< real64 > const m_newDamage;
+  arrayView3d< real64 > const m_damageGrad;
   arrayView2d< real64 > const m_strainEnergyDensity;
+  arrayView2d< real64 > const m_volStrain;
   real64 const m_lengthScale;
   real64 const m_criticalFractureEnergy;
   real64 const m_criticalStrainEnergy;
+  arrayView1d< real64 > const m_biotCoefficient;
```
CusiniM, line 304: I don't think the `biotCoefficient` should be stored in the Damage model. Are you using a `PorousSolid`? I would assume that you are if there is a poroelastic solver.

src/coreComponents/finiteElement/BilinearFormUtilities.hpp
```diff
@@ -263,6 +270,158 @@ struct Helper< PDEUtilities::FunctionSpace::H1vector,
...
+                       real64 const & A,
+                       real64 const (&dNudX)[numTrialDOF][3],
+                       real64 const weight )
+  {
+    for( int a = 0; a < numTestDOF; ++a )
+    {
+      for( int b = 0; b < numTrialDOF; ++b )
+      {
+        mat[a][b] = mat[a][b] + A * LvArray::tensorOps::AiBi< 3 >( dNvdX[a], dNudX[b] ) * weight;
+        //must verify voigt notation and order of indices in dNvdX and dNudX
```
CusiniM, line 324: clean up commented code.

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.hpp
```diff
@@ -291,6 +296,7 @@ class SolidMechanicsLagrangianFEM : public SolverBase
   real64 m_maxForce = 0.0;
   integer m_maxNumResolves;
   integer m_strainTheory;
+  integer m_pressureEffectsFlag;
```
CusiniM, line 299: I don't love this. It's hard to understand what it means for a general `solidMechanicsSolver`.

src/coreComponents/physicsSolvers/simplePDE/PhaseFieldDamageFEM.hpp
```diff
@@ -126,12 +126,14 @@ class PhaseFieldDamageFEM : public SolverBase
...
   {
+  #if 0
     static constexpr char const * coeffNameString() { return "coeffField"; }
+  #endif
     static constexpr char const * localDissipationOptionString() { return "localDissipation"; }
     static constexpr char const * solidModelNamesString() { return "solidMaterialNames"; }
 
     dataRepository::ViewKey timeIntegrationOption = { "timeIntegrationOption" };
-    dataRepository::ViewKey fieldVarName = { "fieldName" };
+    dataRepository::ViewKey fieldVarName = { "Damage" };
```
CusiniM, line 136: most keys are lowercase so I would use 'damage'

src/coreComponents/constitutive/solid/Damage.hpp
andrembcosta: I think it is a bit strange to have the biotCoefficient here as well. However, I don't need to use a PorousSolid for this problem, so, if I were to use one, I would need to turn off the flow solvers. I believe it lead to more modifications in the code, so, I did this approach instead.

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.hpp
andrembcosta: That is also true, but it reduces a lot of code compared to using a full porousSolid material, and having to switch off the flow solvers.

src/coreComponents/physicsSolvers/simplePDE/PhaseFieldDamageFEM.hpp
andrembcosta: ok

rrsettgast:
@frankfeifan Can you look at this and evaluate whether you want to clean it up, or create another branch to implement features?

PR #2060, opened 2022-09-13
Removed MPI_CXX_BOOL

corbett5:
`MPI_CXX_BOOL` isn't actually part of the MPI standard, and it has led to issues on Perlmutter.

### DIFF
--- a/src/coreComponents/common/MpiWrapper.hpp
+++ b/src/coreComponents/common/MpiWrapper.hpp
@@ -52,7 +52,6 @@ typedef int MPI_Datatype;
 #define MPI_LONG_LONG_INT  ((MPI_Datatype)0x4c000809)
 #define MPI_UNSIGNED_LONG_LONG ((MPI_Datatype)0x4c000819)
 #define MPI_LONG_LONG      MPI_LONG_LONG_INT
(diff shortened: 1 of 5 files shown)


klevzoff (approved):
It seems `MPI_CXX_BOOL` is defined in MPI-3.0, but we shouldn't require anything beyond MPI-1 unless we have a very good reason to.

Infinoid:
I have been seeing this crash on NERSC Perlmutter:

```
aborting job:
application called MPI_Abort(MPI_COMM_WORLD, 1) - process 0
MPICH ERROR [Rank 1] [job id 3148367.2] [Tue Sep 13 07:29:53 2022] [nid001916] - Abort(671684611) (rank 1 in comm 0): Fatal error in PMPI_Allreduce: Invalid datatype, error stack:
PMPI_Allreduce(472): MPI_Allreduce(sbuf=0x21abf998, rbuf=0x7ffe65593cdf, count=1, datatype=MPI_DATATYPE_NULL, op=MPI_LOR, comm=comm=0x84000003) failed
PMPI_Allreduce(415): Datatype for argument datatype is a null datatype
```

I tried this patch and no longer see the above crash.  (Still having other issues, but it seems to get past this one.)

PR #2062, opened 2022-09-14
Implemented time stepping based on state change

francoishamon:
This PR enables variable time stepping based on state change, and implements the standard time-stepping heuristics for the compositional solver.

In passing, the PR:
- Makes the naming in the xml fields defined in `NonlinearSolverParameters` more uniform: before we had `dt`, `timestep`, and `timeStep`, now everything is `timeStep`.
- Changes the signature of `setNextDt` in `SolverBase` (now takes a domain and returns the new time step size).

Time-stepping based on Newton convergence and state change can be combined or disabled independently:
- Setting very large target change values (like 1e9) will yield a time-stepping based on Newton convergence only.
- Setting`timeStepIncreaseIterLimit` close to 1 will yield a time-stepping based on state change.
- In between, GEOSX will choose the most conservative strategy between the two.

Ready for review, requires rebaseline. 
Rebaseline done in https://github.com/GEOSX/integratedTests/pull/261


src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -74,6 +73,27 @@ CompositionalMultiphaseBase::CompositionalMultiphaseBase( const string & name,
...
 
+  this->registerWrapper( viewKeyStruct::solutionChangeScalingFactorString(), &m_solutionChangeScalingFactor ).
+    setSizedFromParent( 0 ).
+    setInputFlag( InputFlags::OPTIONAL ).
+    setApplyDefaultValue( 0 ).
+    setDescription( "Damping factor for solution change targets" );
+  this->registerWrapper( viewKeyStruct::targetPressureChangeString(), &m_targetPressureChange ).
+    setSizedFromParent( 0 ).
+    setInputFlag( InputFlags::OPTIONAL ).
+    setApplyDefaultValue( 1.37895e7 ). // 2000 psi (taken from user guide)
```
CusiniM, line 84: We don't really specify units in GEOSX   so one could easily run the code considering pressures to be in `bar`, for example, and then adjust all other properties accordingly. This, would make this value kinda meaningless. It's also trued that it's an optional parameter though.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -1733,6 +1755,83 @@ void CompositionalMultiphaseBase::chopNegativeDensities( DomainPartition & domai
...
+  {
+    maxAbsoluteTemperatureChange = MpiWrapper::max( maxAbsoluteTemperatureChange );
+    GEOSX_LOG_LEVEL_RANK_0( 1, getName() << ": Max absolute temperature change: "<< maxAbsoluteTemperatureChange << " K" );
+  }
+
+  real64 const eps = LvArray::NumericLimits< real64 >::epsilon;
+
+  real64 const nextDtPressure = currentDt *  ( 1.0 + m_solutionChangeScalingFactor ) * m_targetPressureChange
+                                / std::max( eps, maxAbsolutePressureChange + m_solutionChangeScalingFactor * m_targetPressureChange );
+  real64 const nextDtPhaseVolFrac = currentDt *  ( 1.0 + m_solutionChangeScalingFactor ) * m_targetPhaseVolFracChange
```
CusiniM, line 1825: Is this time of timestepping very effective for compositional flow? I have kind of a hard time guessing what good values are for max changes and for the `m_solutionChangeScalingFactor`. I am guessing it's mostly based on previous experience.

CusiniM (approved):
Looks good to me.

TotoGaz (commented):
Would it make sense to delegate the `dt` computation to some ~`TimeStepComputer` instead of making every solver tightly linked to one imposed policy?
Any user could select his strategy at runtime?

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
francoishamon: Here I tried to follow the standard heuristics implemented in other reservoir simulators to avoid re-inventing the wheel. The implementation matches what is implemented in other reservoir simulators like MRST, AD-GPRS and the commercial ones (they all follow Hui Cao's thesis). 

The default values I picked are also inspired by their default values, with the most important ones being the `targetPhaseVolFractionChange` equal to 0.2 in most simulators. As for the `solutionChangeScalingFactor`, it has a minor impact and the default value of 0 is good enough, I think

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
francoishamon: Yes, you are right. I can just replace the `targetPressureChange` (absolute) with a `targetRelativePressureChange`, that will solve this issue.

francoishamon:
> Would it make sense to delegate the `dt` computation to some ~`TimeStepComputer` instead of making every solver tightly linked to one imposed policy? Any user could select his strategy at runtime?

Yes, that would sense to have the user select the strategy at runtime. I think for the solution strategy based on state change, it is good to have it hard-coded in the single-physics solver, because these state change heuristics are so standard. But, we could have a cleaner way to let the user choose between nonlinear convergence and state change, and we could also think of other time stepping schemes for specific cases or specific physics (MRST has some good examples).

src/coreComponents/physicsSolvers/NonlinearSolverParameters.hpp
```diff
@@ -92,23 +95,41 @@ class NonlinearSolverParameters : public dataRepository::Group
 
   /**
    * @brief Calculates the upper limit for the number of iterations to allow a
-   * cut to the next timestep.
-   * @return The scaled value of the limit (m_dtCutIterLimit * m_maxIterNewton)
+   * decrease to the next time step.
+   * @return The scaled value of the limit (m_timeStepDecreaseIterLimit * m_maxIterNewton)
    */
-  integer dtCutIterLimit() const
+  integer timeStepDecreaseIterLimit() const
```
jafranc, line 101: Maybe name it  `currentTimeStepDecreaseIterLimit` as it is not a direct getter but return the applied value ?

src/coreComponents/physicsSolvers/NonlinearSolverParameters.hpp
```diff
@@ -92,23 +95,41 @@ class NonlinearSolverParameters : public dataRepository::Group
...
 
   /**
    * @brief Calculates the lower limit for the number of iterations to force an
-   * increase to the next timestep.
-   * @return The scaled value of the limit (m_dtIncIterLimit * m_maxIterNewton)
+   * increase to the next time step.
+   * @return The scaled value of the limit (m_timeStepIncreaseIterLimit * m_maxIterNewton)
    */
-  integer dtIncIterLimit() const
+  integer timeStepIncreaseIterLimit() const
```
jafranc, line 112: same with `currentTimeStepIncreaseLimit` (maybe)

jafranc (approved):
Looks great ! Many  thanks for this

src/coreComponents/physicsSolvers/NonlinearSolverParameters.hpp
francoishamon: Hi @jafranc, yes I see your point, however here the `timeStepDecreaseIterLimit` function returns a value that is constant for all the time steps of the simulations. So, do you mind if I keep this name to make it clear that the returned value is not specific to the current time step?

jafranc: Not at all. We can work with this name :)

Issue #2023, opened 2022-08-22
modify build strategy to use dynamic linking again.


rrsettgast:
**Describe the issue**
At some point we were having problems running nsightcompute when dynamically linking our libraries, so we switched to static linking. This is undesirable for two reasons:

1. GEOSX is LGPL, so we should be dynamically linking, and
2. the size of the statically linked unit tests are very large. Each executable contains all the static objects. Not good

**Proposed cleanup**
Switch back to dynamic linking as default. I think that this requires us to clean up our circular dependencies once and for all. We should maintain the option to statically link in case there are any problems later.

bmhan12:
On the note of circular dependencies, I ran the tools we added last time we looked at circular dependencies in #1352:

##  Graphviz

Running `config-build.py` with the `--graphviz` option ([code link](https://github.com/GEOSX/GEOSX/blob/1446a80f6adbc174dab503ff1bbee03eab21163b/scripts/config-build.py#L132)) produces the graph of the dependencies according to CMake.

Note: In order to get this graph, I built GEOSX with shared libraries on and object libraries turned off. For whatever reason, I couldn't get CMake to detect the object libraries registered through blt to show up.

![dependency_graphviz](https://user-images.githubusercontent.com/53839862/188676225-8e8bae66-489e-41f8-bd1c-89855f87c1d3.png)

By CMake's view, we don't have any circular dependencies currently.

## circular_dependencies.sh

This script produces two files: 
[circular_dependencies.txt](https://github.com/GEOSX/GEOSX/files/9498540/circular_dependencies.txt) and [dependency_matrix.txt](https://github.com/GEOSX/GEOSX/files/9498536/dependency_matrix.txt).

`circular_dependencies.txt` shows what header includes exist between two components.

Currently, there are 9:

1. common & codingUtilities
2. common & mesh
3. codingUtilities & mainInterface
4. constitutive & mesh
5. constitutive & fileIO
6. constitutive & physicSolvers
7. mesh & finiteElement
8. finiteVolume & physicsSolvers
9. physicsSolvers & mainInterface

`dependency_matrix.txt` shows if a component contains headers to another component (yes/no)

## find_cycles.py

I did the following to run this script:
```
ml ninja (load the module)
 
./scripts/config-build.py -hc host-configs/LLNL/quartz-clang\@10.0.0.cmake -ecc -G Ninja
 
[...]

Issue #2016, opened 2022-08-15
[Bug] Unit tests not passing on Mac OSX


TotoGaz:
31 unit tests are not passing on mac osx 12.4

```
Total Test time (real) = 144.45 sec
The following tests FAILED:
	110 - testXML (Failed)
	111 - testXMLFile_basic_input (Failed)
	112 - testXMLFile_include_input (Failed)
	113 - testXMLFile_multiple_input (Failed)
	114 - testConformingVirtualElementOrder1 (Failed)
	115 - testDofManager (Failed)
	116 - testLAIHelperFunctions (Failed)
	126 - testDamage (Subprocess aborted)
	132 - testTriaxial_druckerPragerExtended (Failed)
	133 - testTriaxial_elasticIsotropic (Failed)
	134 - testTriaxial_elasticIsotropicPressureDependent (Failed)
	135 - testTriaxial_modifiedCamClay (Failed)
	136 - testTriaxial_delftEggUseLinear (Failed)
	137 - testTriaxial_modifiedCamClayVolumetric (Failed)
	138 - testTriaxial_delftEggLoadPathDryUseLinear (Failed)
	139 - testTriaxial_delftEggLoadPathWetUseLinear (Failed)
	140 - testTriaxial_delftEggCase1 (Failed)
	141 - testTriaxial_delftEggCase2 (Failed)
	142 - testPVT (Failed)
	144 - testMeshGeneration (Failed)
	146 - testPAMELAImport (Failed)
	147 - testVTKImport (Failed)
	149 - testVTKImport_mpi (Failed)
	156 - testThermalCompMultiphaseFlow (Failed)
	157 - testThermalSinglePhaseFlow (Failed)
	158 - testCompMultiphaseFlow (Failed)
	159 - testCompMultiphaseFlowHybrid (Failed)
	160 - testReactiveCompositionalMultiphaseOBL (Failed)
	161 - testReservoirSinglePhaseMSWells (Failed)
	163 - testReservoirCompositionalMultiphaseMSWells (Failed)
	164 - testWavePropagation (Failed)
```

Additional logs for one case (all logs are more or less identical)
```
***** Rank 0: Could not find keyword “InternalMesh” in this context. Please be sure that all your keywords are properly spelled or that input file parameters have not changed.
All available keys are: [
InternalWell
] (modifié) 
[11 h 52](https://fcmaelstrom.slack.com/archives/D03T8RUB96Z/p1660243972252339)
test 110
[...]

PR #2076, opened 2022-09-22
Shared Library

bmhan12:
This PR:

- Sets the default to build `geosx_core` as a shared library
  - For clang@upstream on lassen release, build directory size from static to shared: 13GB --> 4.4 GB
  - For clang@10.0.0 on quartz release, build directory size from static to shared: 4.2GB --> 1.5 GB
- Add CMake config output for what CMake targets are being built and linking behavior for each configuration of `GEOSX_BUILD_SHARED_LIBS` and `GEOSX_BUILD_OBJ_LIBS`
- Related to #2023

As part of this PR I also tried building `geosx_core` as a static library and linking executables to `geosx_core`, but ran into the errors detailed in issue #2016 . This is likely related to circular dependencies exposed by building GEOSX this way.


klevzoff (approved):

TotoGaz (approved):

bmhan12:
Moving this out of the merge queue... seeing the lassen integrated tests hang with shared libraries (or might be my terminal/ allocation). Need to debug further against develop.

rrsettgast:
> Moving this out of the merge queue... seeing the lassen integrated tests hang with shared libraries (or might be my terminal/ allocation). Need to debug further against develop.

Maybe ask LC for help

rrsettgast:
@bmhan12 what happens if you use cmake object libraries, statically link geosx, and dynamically link all unit tests?

rrsettgast (approved):

PR #2078, opened 2022-09-23
Increased well verbosity; made well shut down more robust

francoishamon:
This PR addresses some feature requests for the wells:
- [X] Added a `logLevel` in `WellControls` to output the well rates, well BHP, well surface conditions, well surface density, etc
- [X] Added an output of the computed well transmissibilities (for debugging and comparison with other simulators)
- [X] Made the well shut down more robust (new method leads to a 20% reduction in linear iterations)
- [X] Fixed the capitalization of `wellRegionUniqueSubRegion`, since it appears in the xml

The PR is ready to be reviewed and will require a rebaseline. Rebaseline done in https://github.com/GEOSX/integratedTests/pull/263


src/coreComponents/physicsSolvers/fluidFlow/wells/WellSolverBase.hpp
```diff
@@ -233,6 +233,22 @@ class WellSolverBase : public SolverBase
...
+  /**
+   * @brief apply a special treatment to the wells that are shut
+   * @param time_n the time at the previous converged time step
+   * @param dt the time step size
+   * @param domain the physical domain object
+   * @param dofManager degree-of-freedom manager associated with the linear system
+   * @param matrix the system matrix
+   * @param rhs the system right-hand side vector
+   */
+  virtual void applyWellShutDown( real64 const time_n,
```
TotoGaz, line 245: Is the `apply` in `applyWellShutDown` mandatory of could it be `shutDownWell`?

src/coreComponents/physicsSolvers/multiphysics/CoupledReservoirAndWellsBase.hpp
```diff
@@ -175,16 +180,54 @@ class CoupledReservoirAndWellsBase : public CoupledSolver< RESERVOIR_SOLVER, WEL
...
 
-protected:
-
   virtual void
   initializePostInitialConditionsPreSubGroups() override
   {
     Base::initializePostInitialConditionsPreSubGroups( );
 
-    coupledReservoirAndWellsInternal::initializePostInitialConditionsPreSubGroups( this );
+    DomainPartition & domain = this->template getGroupByPath< DomainPartition >( "/Problem/domain" );
```
TotoGaz, line 188: 🙈

TotoGaz (approved):

src/coreComponents/physicsSolvers/fluidFlow/wells/WellSolverBase.hpp
francoishamon: Done, I changed these function names to `shutDownWell`, I agree it is better

src/coreComponents/physicsSolvers/multiphysics/CoupledReservoirAndWellsBase.hpp
francoishamon: Yes, this is a problem I am having constantly in these initialization functions. I am not aware of a getter for the domain partition that could be used in the physics solvers, but maybe there is one? Or we could add one? That would definitely be cleaner that this.

PR #2079, opened 2022-09-23
Fix nvcc crash related to fast_float

klevzoff:
This PR fixes an `nvcc` crash that was happening on my machine (Ubuntu 20.04, cuda-10.1.243) but for some reason not in CI and apparently on none of the real systems either (lassen, etc.):
```
Error: Internal Compiler Error (codegen): "there was an error in verifying the lgenfe output!"
```

Figured out that it was crashing when compiling the call to `fast_float::from_chars()`. The proposed solution that seems to work for me:
  * Move all calls to `fast_float` into a separate translation unit (for consistency moved parsing of integral types there as well)
  * Force it to be compiled with host compiler

### DIFF
--- a/src/coreComponents/codingUtilities/CMakeLists.txt
+++ b/src/coreComponents/codingUtilities/CMakeLists.txt
@@ -15,6 +15,7 @@ set( codingUtilities_headers
 # Specify all sources
 #
 set( codingUtilities_sources
+     Parsing.cpp
      StringUtilities.cpp )
 
 set( dependencyList common fast_float )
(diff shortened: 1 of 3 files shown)


klevzoff:
I edited a file with web editor and now I can't seem to trigger Travis on this PR...?

TotoGaz:
> I changed a file with web editor and now I can't seem to trigger Travis on this PR...?

It should be OK, but yesterday I experienced issues on travis, so maybe this is what happened. 🤷

PR #2080, opened 2022-09-23
Fix nvcc crash related to fast_float

klevzoff:
This PR fixes an `nvcc` crash that was happening on my machine (Ubuntu 20.04, cuda-10.1.243) but for some reason not in CI and apparently on none of the real systems either (lassen, etc.):
```
Error: Internal Compiler Error (codegen): "there was an error in verifying the lgenfe output!"
```

Figured out that it was crashing when compiling the call to `fast_float::from_chars()`. The proposed solution that seems to work for me:
  * Move all calls to `fast_float` into a separate translation unit (for consistency moved parsing of integral types there as well)
  * Force it to be compiled with host compiler

### DIFF
--- a/src/coreComponents/codingUtilities/CMakeLists.txt
+++ b/src/coreComponents/codingUtilities/CMakeLists.txt
@@ -15,6 +15,7 @@ set( codingUtilities_headers
 # Specify all sources
 #
 set( codingUtilities_sources
+     Parsing.cpp
      StringUtilities.cpp )
 
 set( dependencyList common fast_float )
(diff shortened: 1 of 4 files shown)


klevzoff:
@TotoGaz @rrsettgast could you take a look please, this is a small PR but it fixes all remaining CUDA debug compilation issues for me

src/coreComponents/codingUtilities/CMakeLists.txt
```diff
@@ -30,6 +31,9 @@ blt_add_library( NAME       codingUtilities
                  OBJECT     ${GEOSX_BUILD_OBJ_LIBS}
                )
 
+# Avoid compiling with nvcc which sometimes crashes on fast_float
+set_source_files_properties( Parsing.cpp PROPERTIES LANGUAGE CXX )
```
rrsettgast, line 35: what is the standard `LANGUAGE` property value when running on a machine with GPU?

rrsettgast (approved):

src/coreComponents/codingUtilities/CMakeLists.txt
klevzoff: BLT automatically sets `LANGUAGE=CUDA` for all source files in any target that has a dependency on `cuda` target

PR #2081, opened 2022-09-23
Renable parallelDevicePolicy in SEM

XL64:


src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEM.hpp
```diff
@@ -31,7 +31,7 @@ class AcousticWaveEquationSEM : public WaveSolverBase
 {
 public:
 
-  using EXEC_POLICY = serialPolicy;//parallelDevicePolicy< 32 >;
+  using EXEC_POLICY = parallelDevicePolicy< 32 >;
```
XL64, line 34: @rrsettgast Hello, before we had parallelDevicePolicy without the 32, what is the best solution ?

sframba (approved):

PR #2091, opened 2022-09-28
First version of `meshDoctor`

TotoGaz:
This first version of a `checkMesh`
- finds collocated nodes for the input mesh
- is clearly over engineered for so little feature (but at least, it should be OK to extend it 🤣 ).

The idea is to get a first version because otherwise we'll never have anything...
And to be able to make it evolve easily. Will it be the case?

EDIT: And what about the name? I was considering extending it to be able to generate the vtk global indices, split nodes, etc. Then it's no more a _check_ mesh. Maybe a `meshDoctor` 🤣 ?


scripts/checkMesh/checks/collocated_nodes.py
```diff
@@ -0,0 +1,53 @@
...
+    locator.InitPointInsertion(output, points.GetBounds())
+
+    # original ids to/from filtered ids.
+    # original_to_filtered = numpy.ones(points.GetNumberOfPoints(), dtype=int) * -1
+    filtered_to_original = numpy.ones(points.GetNumberOfPoints(), dtype=int) * -1
+
+    rejected_points = defaultdict(list)
+    point_id = vtk.reference(0)
+    for i in range(points.GetNumberOfPoints()):
+        is_inserted = locator.InsertUniquePoint(points.GetPoint(i), point_id)
```
klevzoff, line 41: I'm curious to see how well this performs on large meshes. Intuitively I'd think crossing Python/C++ boundaries multiple times inside a hot loop is not a viable approach, but happy to be proven wrong.

scripts/checkMesh/checks/collocated_nodes.py
```diff
@@ -0,0 +1,53 @@
...
+    locator.SetTolerance(options.tolerance)
+    output = vtk.vtkPoints()
+    locator.InitPointInsertion(output, points.GetBounds())
+
+    # original ids to/from filtered ids.
+    # original_to_filtered = numpy.ones(points.GetNumberOfPoints(), dtype=int) * -1
+    filtered_to_original = numpy.ones(points.GetNumberOfPoints(), dtype=int) * -1
+
+    rejected_points = defaultdict(list)
+    point_id = vtk.reference(0)
```
XL64, line 39: I don't really get what is a `vtk.reference()`, does point_id change within the loop ? If not `filtered_to_original[point_id.get()] = i` always fill the same index of the array. I must have missed something...

scripts/checkMesh/parsing/cli_parsing.py
```diff
@@ -0,0 +1,100 @@
...
+    :return: The struct
+    """
+    parser = argparse.ArgumentParser(description='Inspects meshes for GEOSX.')
+    parser.add_argument('-i',
+                        '--vtk-input-file',
+                        metavar='VTK_MESH_FILE',
+                        type=str,
+                        required=True,
+                        dest="vtk_input_file")
+    parser.add_argument('-c',
```
XL64, line 72: Wouldn't it be simpler to have each checker providing a populate parser method which would add option to the parser than passing all through the check option and then parse it?

scripts/checkMesh/checks/collocated_nodes.py
TotoGaz: Hell @klevzoff thanks for your comment.

I've tested the function on a ~6.5 million nodes mesh and it lasts ~7.3 seconds on a 2013 CPU.
Difficult to tell what's the influence of the python/c++ bridge.

- The 6.5M _empty_ loop uses 0.1 sec in python (doing nothing, just `pass`).
- The same loop calling a simple `virtual` function from python (like `locator.GetNumberOfNodes()` which returns an `int`) takes 0.55 secondes.

So I assume that if we have two python/c++ calls in this loop, maybe 1 of the 7.3 seconds is used in python/c++ transfer? Not a very smart estimate though.

scripts/checkMesh/checks/collocated_nodes.py
TotoGaz: The doc says that `the index of the point (either new or the duplicate) is returned via point_id.`.
So if the point is inserted, then `point_id` contains the new id of the point (it's been merged, it has a new numbering) and we keep track of the mapping `new_index -> old_index`.

If it's not inserted, then `point_id` contains the node that was already at that location. But `point_id` is the new numbering, in the destination points array. I found most useful for the user to get the old index, so he can look for it in his data.

`point_id` is a `vtk.reference()` because it's written by the `locator.InsertUniquePoint` call.

See https://vtk.org/doc/nightly/html/classvtkIncrementalOctreePointLocator.html

scripts/checkMesh/parsing/cli_parsing.py
TotoGaz: It's possible and it makes sense, but it surely depends on the number of checks will eventually get.
I did not think that way to be honest, but it's a valuable idea.

scripts/checkMesh/checks/collocated_nodes.py
XL64: Thanks for the explanation!

scripts/checkMesh/parsing/cli_parsing.py
XL64: Ok good to me

francoishamon (approved):
I don't know enough vtk to comment on the code, but this `meshDoctor` feature looks really useful.

Maybe later we can add a function that checks whether a given cell has the expected number of unique points for his type (say, a hexahedra should have 8 unique points)? This is something that was making GEOSX crash for some ill-formed PEBI meshes that I tested recently.

scripts/checkMesh/parsing/cli_parsing.py
TotoGaz: I think I'll got for something like option 2, which looks like a quite efficient way to show the options.
But that pattern somehow induces some kind of introspection to know which options are activated, which are not.
I'll have a look.

TotoGaz:
> Maybe later we can add a function that checks whether a given cell has the expected number of unique points for his type (say, a hexahedra should have 8 unique points)?

This is this idea, and yes we could add this check.

But are your sure this is your root cause?
Maybe the vtk global id generation process (I think) did merge the collocated nodes, resulting in apparently same nodes in GEOSX, while different in the vtk mesh?
Actually, the current check should be able to find collocated nodes and maybe it's enough.

cssherman:
> Hello @cssherman I was unaware of `coreComponents/python/modules/geosx_mesh_tools_package/geosx_mesh_tools` but we could surely merge the tools into something more consistent? BTW, why do we put this into the `coreComponents` folder. That's looks a bit unexpected to me.

Sure, I think that the tools are quite compatible.  In terms of the location, this is just where Randy ended up placing them during the last major file reorganization effort about a year ago.  I don't think that the scripts subdirectory is the best place for them, tbh, but maybe a submodule would be appropriate?

TotoGaz:
> Sure, I think that the tools are quite compatible. In terms of the location, this is just where Randy ended up placing them during the last major file reorganization effort about a year ago. I don't think that the scripts subdirectory is the best place for them, tbh, but maybe a submodule would be appropriate?

I can move this `meshDoctor` in the same location for consistency.
But eventually I think we should move it from there.
Is it important that it's under `src` (for deployment needs or checks, _e.g._)?

cssherman:
Looks great!  Elsewhere, I'm using google-style docstings, as I think they are a bit easier to read in their raw form.  In `src/conf.py`, I've enabled `sphinx.ext.napoleon`, which should allow us to build with either format if that is your preference.  It may also be worth adding a documentation page similar to `src/docs/sphinx/pythonTools/geosx_mesh_tools.rst` to hold the automodule and argparse documentation.

TotoGaz:
> I'm using google-style docstings, as I think they are a bit easier to read in their raw form. In `src/conf.py`, I've enabled `sphinx.ext.napoleon`, which should allow us to build with either format if that is your preference.

I think I should engage into another PR to merge the current existing tools and those that I'm currently developing.
And also putting all our conventions altogether.

I'm finishing `vtk` tools to generate fracture information, generate global ids, etc. I'll push it very soon.
Soon I'll make another check which _e.g._ lists all the cell types and informs if there are unsupported types, which can remove cell that we don't want, etc.

> It may also be worth adding a documentation page similar to `src/docs/sphinx/pythonTools/geosx_mesh_tools.rst` to hold the automodule and argparse documentation.

My current approach is to have modules that load dynamically and bring their own arguments dynamically.
The idea is that I do not want an option to be proposed to the end user if it's not available (a warning message is displayed that the module did not load instead).
Because I don't want a module which does not load to prevent from using other modules (for ex, if I simply want to address `vtk` stuff, I do not need `meshio` used for meshes conversions.).
So I do not know how compatible this can be with the `automodule` process (obviously, it can be OK if all the modules are loaded properly, but we need to be precise about it.)

cssherman:
@TotoGaz - sounds good.  As far as module availability is concerned, we simply need to include them in this `conf.py` entry (as we don't want to build these on the RTD instance):

```python
autodoc_mock_imports = ["pygeosx", "pylvarray", "meshio", "lxml", "mpi4py", "h5py"]
```

cssherman (approved):

TotoGaz:
Hello @cssherman I've pushed a more recent version of the scripts. Could you give it a glance?

cssherman:
@TotoGaz - what do you think our policy on python type hint validation should be in the codebase?  Now that I've been using these for a few months, I've found that getting the last couple of type definitions to properly resolve takes quite a bit of effort.  I'm using a combination of `mypy` and it's `SublimeLinter` plugin with the `--ignore-missing-imports` flag (we don't want to be worrying about missing definitions from other projects), which produces around 30 (mostly nit-picky) warnings here.  Otherwise, what I've looked over seems good.

TotoGaz:
Actually I've never used the type hint for real.
I tried to start this "_project_" with it because it was actually already helping me with all there indices everywhere!

I've used what my IDE was giving me, and my development organization for python is not great for the moment.
For instance, I'd be surprised that my annotated code 100% validates.
Most probably because there are things I do not fully understand right now.
And it's not so easy to get all the good practices.

Eventually (and quite soon I think), we should propose a set of options for out linter (like for `uncrustify`?) and check it in on travis.
Can we select what we want to be hinted? E.g. function signatures should be annotated, but variables may not be?

cssherman:
That's good to hear.  From what I've read, type hints aren't 100% mature yet anyway (for example, they don't handle some nested structures well).  There are a _**ton**_ of options available to `mypy`, but most of them seem to be geared to making the checks more restrictive, not limiting their scope.

TotoGaz:
@cssherman Are you OK if I merge this as is?
I intend to move it into `coreComponents/python/modules` too.

Also, I think that https://docs.python.org/3/library/collections.abc.html#module-collections.abc is interesting to reach more precise type hints.

cssherman:
> @cssherman Are you OK if I merge this as is? I intend to move it into `coreComponents/python/modules` too.
> 
> Also, I think that https://docs.python.org/3/library/collections.abc.html#module-collections.abc is interesting to reach more precise type hints.

Yeah, sounds good to me!

PR #2098, opened 2022-09-30
Added extrinsic mesh data for solid mechanics

francoishamon:
This PR adds extrinsic mesh data registration in the solid mechanics solvers.

In passing, the PR enforces camel case for the fields registered by the solid mechanics solvers to be consistent with the rest of the code: 
- `TotalDisplacement` -> `totalDisplacement`
- `Velocity` -> `velocity`
etc

Resolves https://github.com/GEOSX/GEOSX/issues/1215

Ready for review, no rebaseline necessary

### DIFF
--- a/src/coreComponents/dataRepository/KeyNames.hpp
+++ b/src/coreComponents/dataRepository/KeyNames.hpp
@@ -30,41 +30,11 @@ namespace keys
 
 /// @cond DO_NOT_DOCUMENT
 
-//static constexpr auto ReferencePosition = "ReferencePosition";
-static constexpr auto referencePositionString = "ReferencePosition";
-
-static constexpr auto TotalDisplacement = "TotalDisplacement";
-static constexpr auto IncrementalDisplacement = "IncrementalDisplacement";
-static constexpr auto Velocity = "Velocity";
-static constexpr auto Acceleration = "Acceleration";
-static constexpr auto Mass = "Mass";
-static constexpr auto Force = "Force";
-static constexpr auto Strain = "Strain";
-static constexpr auto Name = "name";
-static constexpr auto Size = "size";
 static constexpr auto ProblemManager = "Problem";
 static constexpr auto ConstitutiveManager = "Constitutive";
-static constexpr auto ConstitutiveBase = "ConstitutiveBase";
-static constexpr auto solverNames = "solverNames";
-
 static constexpr auto schema = "schema";
-
-static constexpr auto time = "time";
-static constexpr auto cycle = "cycle";
-static constexpr auto beginTime = "beginTime";
-static constexpr auto endTime = "endTime";
-static constexpr auto dt = "dt";
-
 static constexpr auto domain  = "domain";
-static constexpr auto solvers = "solvers";
-static constexpr auto simulationParameterMap = "simulationParameterMap";
-static constexpr auto FE_Space    = "FE_Space";
-//static constexpr auto FEM_Nodes    = "FEM_Nodes";
-//static constexpr auto FEM_Edges    = "FEM_Edges";
-//static constexpr auto FEM_Faces    = "FEM_Faces";
-//static constexpr auto FEM_Elements = "FEM_Elements";
 static constexpr auto cellManager = "cellManager";
-static constexpr auto functionManager = "FunctionManager";
 
 /// @endcond
 
--- a/src/coreComponents/fileIO/silo/SiloFile.cpp
+++ b/src/coreComponents/fileIO/silo/SiloFile.cpp
@@ -35,6 +35,7 @@
 #include "fileIO/Outputs/OutputUtilities.hpp"
 #include "mesh/DomainPartition.hpp"
 #include "mesh/MeshBody.hpp"
+#include "physicsSolvers/solidMechanics/SolidMechanicsExtrinsicData.hpp"
 
 #include <iostream>
 
@@ -1625,7 +1626,7 @@ void SiloFile::writeMeshLevel( MeshLevel const & meshLevel,
     }
   }
 
-  if( nodeManager.hasWrapper( keys::TotalDisplacement ) )
+  if( nodeManager.hasWrapper( extrinsicMeshData::solidMechanics::totalDisplacement::key() ) )
   {
     arrayView2d< real64 const, nodes::TOTAL_DISPLACEMENT_USD > const & totalDisplacement = nodeManager.totalDisplacement();
     for( localIndex a = 0; a < numNodes; ++a )
--- a/src/coreComponents/mesh/NodeManager.hpp
+++ b/src/coreComponents/mesh/NodeManager.hpp
@@ -22,6 +22,7 @@
 #include "mesh/generators/CellBlockManagerABC.hpp"
 #include "mesh/ObjectManagerBase.hpp"
 #include "mesh/simpleGeometricObjects/GeometricObjectManager.hpp"
+#include "physicsSolvers/solidMechanics/SolidMechanicsExtrinsicData.hpp"
 #include "ToElementRelation.hpp"
 
 namespace geosx
@@ -247,12 +248,6 @@ public:
     /// @return String to access the location of the nodes
     static constexpr char const * EmbSurfNodesPositionString() { return "EmbSurfNodesPosition"; }
 
-    /// @return String to access the displacement
-    static constexpr char const * totalDisplacementString() { return "TotalDisplacement"; }
-
-    /// @return String to access the incremental displacement
-    static constexpr char const * incrementalDisplacementString() { return "IncrementalDisplacement"; }
-
     /// @return String to access the edge map
     static constexpr char const * edgeListString() { return "edgeList"; }
 
@@ -271,12 +266,6 @@ public:
     /// Accessor to reference position
     dataRepository::ViewKey referencePosition       = { referencePositionString() };
 
-    /// Accessor to displacement
-    dataRepository::ViewKey totalDisplacement       = { totalDisplacementString() };
-
-    /// Accessor to incremental displacement
-    dataRepository::ViewKey incrementalDisplacement = { incrementalDisplacementString() };
-
     /// Accessor to edge map
     dataRepository::ViewKey edgeList                = { edgeListString() };
 
@@ -292,11 +281,6 @@ public:
     /// Accessor to element map
     dataRepository::ViewKey elementList             = { elementListString() };
 
-    /// Accessor to velocity
-    dataRepository::ViewKey velocity                = { dataRepository::keys::Velocity };
-
-    /// Accessor to acceleration
-    dataRepository::ViewKey acceleration            = { dataRepository::keys::Acceleration };
   }
   /// viewKeys
   viewKeys;
@@ -424,7 +408,7 @@ public:
    * @note An error is thrown if the total displacement does not exist
    */
   array2d< real64, nodes::TOTAL_DISPLACEMENT_PERM > & totalDisplacement()
-  { return getReference< array2d< real64, nodes::TOTAL_DISPLACEMENT_PERM > >( viewKeys.totalDisplacement ); }
+  { return getExtrinsicData< extrinsicMeshData::solidMechanics::totalDisplacement >(); }
 
   /**
    * @brief Provide an immutable arrayView to the total displacement array.
@@ -432,7 +416,7 @@ public:
    * @note An error is thrown if the total displacement does not exist
    */
   arrayView2d< real64 const, nodes::TOTAL_DISPLACEMENT_USD > totalDisplacement() const
-  {return getReference< array2d< real64, nodes::TOTAL_DISPLACEMENT_PERM > >( viewKeys.totalDisplacement ); }
+  { return getExtrinsicData< extrinsicMeshData::solidMechanics::totalDisplacement >(); }
 
   /**
    * @brief Get a mutable incremental displacement array.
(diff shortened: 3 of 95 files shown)


src/coreComponents/mesh/NodeManager.hpp
```diff
@@ -424,63 +408,63 @@ class NodeManager : public ObjectManagerBase
    * @note An error is thrown if the total displacement does not exist
    */
   array2d< real64, nodes::TOTAL_DISPLACEMENT_PERM > & totalDisplacement()
-  { return getReference< array2d< real64, nodes::TOTAL_DISPLACEMENT_PERM > >( viewKeys.totalDisplacement ); }
+  { return getExtrinsicData< extrinsicMeshData::solidMechanics::totalDisplacement >(); }
```
francoishamon, line 411: I was unusure about what to do with these getters. They are handy and used everywhere in the code but they seem to introduce a dependency on the solvers (that register the total displacement) now that the string is no longer hard-coded in the `NodeManager`. I was going to remove them, but I don't know what other people think.

klevzoff: They should've never existed in the first place IMO

src/coreComponents/physicsSolvers/surfaceGeneration/SurfaceGenerator.cpp
```diff
@@ -522,9 +523,9 @@ int SurfaceGenerator::separationDriver( DomainPartition & domain,
   FieldIdentifiers fieldsToBeSync;
 
   fieldsToBeSync.addFields( FieldLocation::Face, { extrinsicMeshData::RuptureState::key() } );
-  if( nodeManager.hasWrapper( SolidMechanicsLagrangianFEM::viewKeyStruct::forceExternalString() ) )
+  if( nodeManager.hasWrapper( extrinsicMeshData::solidMechanics::externalForce::key() ) )
```
TotoGaz, line 526: I think we have a `hasExtrinsicData` for this kind of pattern?

src/coreComponents/dataRepository/KeyNames.hpp
```diff
@@ -30,41 +30,11 @@ namespace keys
...
-//static constexpr auto ReferencePosition = "ReferencePosition";
-static constexpr auto referencePositionString = "ReferencePosition";
-
-static constexpr auto TotalDisplacement = "TotalDisplacement";
-static constexpr auto IncrementalDisplacement = "IncrementalDisplacement";
-static constexpr auto Velocity = "Velocity";
-static constexpr auto Acceleration = "Acceleration";
-static constexpr auto Mass = "Mass";
-static constexpr auto Force = "Force";
-static constexpr auto Strain = "Strain";
```
TotoGaz, line 42: So great to see all these key vanish!
This is awesome!

src/coreComponents/fileIO/silo/SiloFile.cpp
```diff
@@ -1625,7 +1626,7 @@ void SiloFile::writeMeshLevel( MeshLevel const & meshLevel,
     }
   }
 
-  if( nodeManager.hasWrapper( keys::TotalDisplacement ) )
+  if( nodeManager.hasWrapper( extrinsicMeshData::solidMechanics::totalDisplacement::key() ) )
```
TotoGaz, line 1629: Please use `hasExtrinsicData`?

src/coreComponents/mesh/NodeManager.hpp
TotoGaz: The pattern we are aiming is to let the "solvers" add the fields they need.
`NodeManager` manages nodes; and as an `ObjectManager` it allows fields registrations on nodes.

We need to shrink the contracts.

For instance, why would we want all the code manipulating the `NodeManager` to be bothered with the `incrementalDisplacement`?

Now be careful if you want to remove them, because that means that everywhere these getters are accessed, it's assumed that the field is there (and therefore registered).
So the patch may not be trivial in the sense where you not only need to change the getter call into and getExtrinsicData call, but also think about the registration. It shouldn't be part of this PR.

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.cpp
```diff
@@ -1049,7 +1017,10 @@ SolidMechanicsLagrangianFEM::
                                          parallelDevicePolicy< 32 > >( targetSet,
                                                                        time_n + dt,
                                                                        targetGroup,
-                                                                       keys::TotalDisplacement, // TODO fix use of
+                                                                       extrinsicMeshData::solidMechanics::totalDisplacement::key(), // TODO
+                                                                                                                                    // fix
+                                                                                                                                    // use
+                                                                                                                                    // of
```
TotoGaz, line 1023: Hmm, can you do something about the indentation here?

TotoGaz (commented):
Awesome to see, little by little, this work being done.
I'm sorry that you're alone doing it 🤷

src/coreComponents/mesh/NodeManager.hpp
francoishamon: Done, I removed these accessors from the `NodeManager`, I think it is simpler to do it in this PR.

src/coreComponents/physicsSolvers/surfaceGeneration/SurfaceGenerator.cpp
francoishamon: Done, I replaced all the `hasWrapper` with `hasExtrinsicData`.

src/coreComponents/fileIO/silo/SiloFile.cpp
francoishamon: Same here

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.cpp
francoishamon: I fixed this indentation

src/coreComponents/mesh/NodeManager.hpp
TotoGaz: Yes, the issue didn't look as challenging as I first thought, since the registrations are already centralized in some solver function ! Good news!

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.cpp
```diff
@@ -1049,9 +1026,7 @@ SolidMechanicsLagrangianFEM::
                                          parallelDevicePolicy< 32 > >( targetSet,
                                                                        time_n + dt,
                                                                        targetGroup,
-                                                                       keys::TotalDisplacement, // TODO fix use of
-                                                                       // dummy
-                                                                       // name
+                                                                       extrinsicMeshData::solidMechanics::totalDisplacement::key(),
```
TotoGaz, line 1029: It's no more a "dummy name"?

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsStateReset.cpp
```diff
@@ -85,9 +85,17 @@ bool SolidMechanicsStateReset::execute( real64 const time_n,
...
+      arrayView2d< real64, nodes::VELOCITY_USD > const velocity =
+        nodeManager.getExtrinsicData< extrinsicMeshData::solidMechanics::velocity >();
+      arrayView2d< real64, nodes::TOTAL_DISPLACEMENT_USD > const totalDisplacement =
+        nodeManager.getExtrinsicData< extrinsicMeshData::solidMechanics::totalDisplacement >();
+      arrayView2d< real64, nodes::INCR_DISPLACEMENT_USD > const & incrementalDisplacement =
+        nodeManager.getExtrinsicData< extrinsicMeshData::solidMechanics::incrementalDisplacement >();
+
+      velocity.zero();
+      totalDisplacement.zero();
+      incrementalDisplacement.zero();
```
TotoGaz, line 97:
```suggestion
      nodeManager.getExtrinsicData< extrinsicMeshData::solidMechanics::velocity >().zero();
      nodeManager.getExtrinsicData< extrinsicMeshData::solidMechanics::totalDisplacement >().zero();
      nodeManager.getExtrinsicData< extrinsicMeshData::solidMechanics::incrementalDisplacement >().zero();
```
I don't know if it's any better.

TotoGaz (approved):

src/coreComponents/fileIO/coupling/ChomboCoupler.cpp
```diff
@@ -159,8 +160,10 @@ void ChomboCoupler::copyNodalData()
   NodeManager const & nodes = m_mesh.getNodeManager();
   localIndex const numNodes = nodes.size();
   arrayView2d< real64 const, nodes::REFERENCE_POSITION_USD > const & referencePos = nodes.referencePosition();
-  arrayView2d< real64 const, nodes::TOTAL_DISPLACEMENT_USD > const & displacement = nodes.totalDisplacement();
-  arrayView2d< real64 const, nodes::TOTAL_DISPLACEMENT_USD > const & velocity = nodes.velocity();
+  arrayView2d< real64 const, nodes::TOTAL_DISPLACEMENT_USD > const & displacement =
```
rrsettgast, line 163:
```suggestion
  solidMechanics::array2dLayoutTotalDisplacementConst const & displacement =
```
Not that this is any more compact, but could we do something like this to make the LHS easier to type? `auto` doesn't work because we have to specify the `const` inside the template template argument.

src/coreComponents/physicsSolvers/contact/SolidMechanicsEmbeddedFractures.cpp
```diff
@@ -77,7 +78,7 @@ void SolidMechanicsEmbeddedFractures::postProcessInput()
   {
     linParams.mgr.strategy = LinearSolverParameters::MGR::StrategyType::solidMechanicsEmbeddedFractures;
     linParams.mgr.separateComponents = true;
-    linParams.mgr.displacementFieldName = keys::TotalDisplacement;
+    linParams.mgr.displacementFieldName = extrinsicMeshData::solidMechanics::totalDisplacement::key();
```
rrsettgast, line 81: This is kind of long. Would anyone be opposed to using `using namespace extrinsicMeshData` in source  files?

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsStateReset.cpp
rrsettgast: I think it is more compact...and equally understandable.

rrsettgast (approved):
This is wonderful. Just a couple of suggestions/questions.

src/coreComponents/physicsSolvers/contact/SolidMechanicsEmbeddedFractures.cpp
TotoGaz: We could be considering renaming `extrinsicMeshData` to something like `fields`?
This would lead to things like
```c++
arrayView1d< real64 const > const & p = subregion.getField< fields::well::pressure >();
```
and
```
nodeMgr.registerField< fields::solidMechanics::totalDisplacement >();
```
which I find very clarifying (for new comers, but for old folks too!). (eventually, we should be declaring on quadratures, but it's another question 🤣 )
We could rename `EXTRINSIC_MESH_DATA_TRAIT` into `DECLARE_FIELD` or `DECLARE_FIELD_INFO` as well?

rrsettgast: I think it would be fine to do this.

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.cpp
francoishamon: I put back the comment before the function call

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsStateReset.cpp
francoishamon: Done

src/coreComponents/dataRepository/KeyNames.hpp
francoishamon: I struggled to remove this file entirely because of two remaining keys:
- `ProblemManager` -> actually used in the `dataRepository` (`xmlWrapper`, `WrapperBase`, `Group` class)
- `cellManager` -> used in two distant parts of the code (`problemManager` and the mesh classes)  

so for now, I kept the file `KeyNames.hpp` in the code

src/coreComponents/physicsSolvers/contact/SolidMechanicsEmbeddedFractures.cpp
francoishamon: Done, I added `using namespace extrinsicMeshData` at the top of the cpp files that I touched in this PR. In the next PR we can do what @TotoGaz suggests

src/coreComponents/fileIO/coupling/ChomboCoupler.cpp
francoishamon: Done, now in this PR, the accessor calls look like:
```
using namespace extrinsicMeshData;

solidMechanics::arrayViewConst2dLayoutTotalDisplacement const & u =
  nodeManager.getExtrinsicData< solidMechanics::totalDisplacement >();
```

PR #2099, opened 2022-09-30
Added component fraction local chopping in OBL

### LINKED ISSUE #2115: Error appear when run the 3D_10x10x10_compressible_smoke.xml case
There is the print log:

./build-your-platform-release/bin/geosx -i inputFiles/singlePhaseFlow/3D_10x10x10_compressible_smoke.xml
GEOSX version 0.2.0 (develop, sha1: c3d046a7b)
Adding Mesh: InternalMesh, mesh
Adding Event: PeriodicEvent, solverApplications
Adding Event: PeriodicEvent, outputs
Adding Solver of type SinglePhaseFVM, named SinglePhaseFlow
Adding Geometric Object: Box, source
Adding Geometric Object: Box, sink
Adding Output: Silo, siloOutput
Adding Object CellElementRegion named mainRegion from ObjectManager::Catalog.
mesh: total number of nodes = 1331
mesh: total number of elems = 1000
regionQuadrature: meshBodyName, meshLevelName, regionName, subRegionName = mesh, Level0, mainRegion, cellBlock
mesh/Level0/mainRegion/cellBlock/water allocated 1 quadrature points
mesh/Level0/mainRegion/cellBlock/rock allocated 1 quadrature points
***** ERROR
***** LOCATION: /home/chaojie/code/GEOSX/src/coreComponents/mesh/MeshObjectPath.cpp:215
***** Controlling expression (should be false): !foundMatch
***** Rank 0: Specified name (elementSubRegions) did not find a match with a object in group (elementSubRegions). Objects that are present in (elementSubRegions) are: cellBlock

francoishamon:
This PR implements the component fraction local chopping in OBL to facilitate Stephan de Hoop's work in GEOSX (otherwise component fraction go out of bounds and results are invalid).

The little chopping algorithm mimics what is implemented in DARTS.

Resolves #2115 

The code is working and does not require a rebaseline.


src/coreComponents/physicsSolvers/fluidFlow/ReactiveCompositionalMultiphaseOBL.cpp
```diff
@@ -372,7 +372,7 @@ real64 ReactiveCompositionalMultiphaseOBL::scalingForSystemSolution( DomainParti
     return 1.0;
   }
 
-  real64 constexpr eps = ReactiveCompositionalMultiphaseOBLKernels::minValueForDivision;
+  real64 constexpr eps = reactiveCompositionalMultiphaseOBLKernels::minValueForDivision;
```
TotoGaz, line 375: Since you have the `using` up there, do you need the namespace here too?

src/coreComponents/physicsSolvers/fluidFlow/ReactiveCompositionalMultiphaseOBL.cpp
```diff
@@ -1218,9 +1223,63 @@ void ReactiveCompositionalMultiphaseOBL::applyDirichletBC( real64 const time,
...
+                                                     ElementSubRegionBase & subRegion )
+    {
+      arrayView2d< real64, compflow::USD_COMP > const compFrac =
+        subRegion.template getExtrinsicData< extrinsicMeshData::flow::globalCompFraction >();
+
+      forAll< parallelDevicePolicy<> >( subRegion.size(), [=] GEOSX_HOST_DEVICE ( localIndex const ei )
+      {
+        bool isScalingRequired = false;
+
+        // Step 1: chop the component fractions between 0 and 1
```
TotoGaz, line 1244: Why do you need `eps` for each `compFrac`?
Can't it be 1 or 0?

src/coreComponents/physicsSolvers/fluidFlow/ReactiveCompositionalMultiphaseOBL.cpp
francoishamon: Yes, you are right, I removed the namespace here

src/coreComponents/physicsSolvers/fluidFlow/ReactiveCompositionalMultiphaseOBL.cpp
francoishamon: Here, what is implemented follows exactly what is in DARTS, which is why I would prefer to keep it like this. This type of chopping after the Newton update does not have any theoretical foundation and is often full of tolerances and buffers... but it works quite well. 

One thing to note is that the `eps` ensures that each component always exists in the system (even just as a trace), which may have an impact on the linear solver behavior... but we have not tested that yet with OBL.

TotoGaz: ok; maybe comment in the code that you're using the recipe from DARTS.
Does it make sense?

TotoGaz (approved):

src/coreComponents/physicsSolvers/fluidFlow/ReactiveCompositionalMultiphaseOBL.cpp
francoishamon: Yes, I added this in the comment before the chopping algorithm

PR #2107, opened 2022-10-05
Configure github's codespaces

TotoGaz:


.vscode/settings.json
```diff
@@ -0,0 +1,16 @@
+{
```
CusiniM, line 1: these two files in .vscode will conflict with individual settings won't they? I am not too sure what will happen coz they have been added to the repo but they are in `.gitignore`. May I ll just try to checkout this branch.

.travis.yml
```diff
@@ -7,7 +7,7 @@ vm:
 
 env:
   global:
-  - GEOSX_TPL_TAG=200-813
+  - GEOSX_TPL_TAG=203-819
```
CusiniM, line 10: I m lost. Why is this different?

CusiniM (commented):
Well, apart from the `.vscode` question this looks good. I ll test it and approve it once I am sure those files don t generate conflicts.

.vscode/settings.json
TotoGaz: They should not conflict because of the `.gitignore` and could be used as examples for new comers.
But maybe one needs to be careful about the update if it gets merged.

.travis.yml
TotoGaz: Because I've updated a docker image in the TPL to use the one provided by microsoft to use codespaces.
But the main issue is that it duplicates the tag. That sucks.

TotoGaz: Maybe we can put an explicit `UPDATE_ME` in the json ? Or make the travis file read the json file?
I do not really know what to do.
I don't know either if this feature will ever get used...

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM docker.io/geosx/ubuntu20.04-gcc10:203-819
```
jafranc, line 1: Is there a need for `python3.x` or explicit version of `git` ? I guess it is already bundled unto the import. Would it be needed to add few ines of comment to remind versions that are such bundled and that we rely on

.vscode/settings.json
```diff
@@ -0,0 +1,16 @@
...
+    "cmake": {
+        "sourceDirectory": "${workspaceFolder}/src",
+        "buildDirectory": "${userHome}/geosx-build",
+        "cacheInit": "/workspaces/GEOSX/host-configs/environment.cmake",
+        "configureSettings": {
+            "BLT_MPI_COMMAND_APPEND": "--allow-run-as-root;--use-hwthread-cpus;--oversubscribe",
+            "ENABLE_OPENMP": "OFF",
+            "ENABLE_HYPRE": "OFF",
+            "ENABLE_TRILINOS": "ON",
```
jafranc, line 10: Are we sure we want to dev/try with `Trilinos` ? IIUC most users are using `HYPRE` isn't it @CusiniM  ?

.travis.yml
```diff
@@ -7,7 +7,7 @@ vm:
...
   - secure: CGs2uH6efq1Me6xJWRr0BnwtwxoujzlowC4FHXHdWbNOkPsXf7nCgdaW5vthfD3bhnOeEUQSrfxdhTRtyU/NfcKLmKgGBnZOdUG4/JJK4gDSJ2Wp8LZ/mB0QEoODKVxbh+YtoAiHe3y4M9PGCs+wkNDw/3eEU00cK12DZ6gad0RbLjI3xkhEr/ZEZDZkcYg9yHAhl5bmpqoh/6QGnIg8mxIqdAtGDw+6tT0EgUqjeqc5bG5WwsamKzJItHSXD5zx8IJAlgDk4EzEGjZe0m56YnNfb9iwqqUsmL3Cuwgs7ByVDYw78JC5Kv42YqoxA5BxMT2mFsEe37TpYNXlzofU7ma2Duw9DGXWQd4IkTCcBxlyR0I0bfo0TmgO+y7PYG9lIyHPUkENemdozsZcWamqqkqegiEdRhDVYlSRo3mu7iCwTS6ZTALliVyEYjYxYb7oAnR3cNywXjblTCI8oKfgLSY+8WijM9SRl57JruIHLkLMCjmRI+cZBfv5tS2tYQTBPkygGrigrrN77ZiC7/TGyfggSN0+y0oYtOAgqEnBcKcreiibMW7tKcV2Z1RFD9ZvIkSc1EXLUPDP8FX1oyhmqBMqVo8LksrYLDJHQ05+F3YNgl2taSt7uMjQ4e8iZ3/IjFeMnbylDw+cj/RbS520HXsFPbWFm2Pb9pceA9n6GnY=
 
 # The integrated test repository contains large data (using git lfs) and we do not use them here.
```
jafranc, line 13: If `git-lfs` is only used for integratedTests (and we do not take them here) should we add it to `Dockerfile` ?

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+
+RUN apt-get update
+RUN apt-get install -y \
+    ca-certificates \
+    curl \
+    git-lfs \
```
jafranc, line 7: If `git-lfs` is only used for integratedTests (and we do not take them here) should we add it to `Dockerfile` ?

jafranc (commented):
Looks great !! How will you orchestrate `paraview` with all that (as you added deps in Dockerfile for it) ? 
Are you planning on some `quick-start` doc page on it in a separate PR or should it be in this one ?

rrsettgast:
@TotoGaz Should I be able to test at his point?

.devcontainer/Dockerfile
TotoGaz: I think we do, because the idea is to provide a full dev env as fast as possible.
Having `git-lfs` installed can be convenient in that regard.
The idea is not to have the tiniest container here imo.

TotoGaz:
> @TotoGaz Should I be able to test at his point?

Given that your _github_ organization allows _codespaces_ and that you can switch to this PR's source code, then yes, it should be up and running.

.devcontainer/Dockerfile
TotoGaz: I don't think we have any specific needs w.r.t. any version of git or python (we have something with brew, but it's another story.)

.vscode/settings.json
TotoGaz: I don't know, we can change.

TotoGaz:
> Looks great !! 

Thx 😉 

> How will you orchestrate `paraview` with all that (as you added deps in Dockerfile for it) ?

I don't know yet, it's still something I should be working on. But I did not really have the time already.

> Are you planning on some `quick-start` doc page on it in a separate PR or should it be in this one ?

Yes, but I'd like to coordinate with the `vs-code` doc as well.

rrsettgast:
@TotoGaz It appears that using code spaces costs money? Am I missing something?
https://docs.github.com/en/billing/managing-billing-for-github-codespaces/about-billing-for-github-codespaces

TotoGaz:
> @TotoGaz It appears that using code spaces costs money?

Yes it does since you are buying resources in the cloud.

In our case, we may want to involve people for a very limited amount of time (engineers to help us fixing bugs here and there). For this kind of use case, having a GEOSX env set up in minutes is very useful and will save us a significant amount of time (and money 🤷 ).
For people investing on the long term on GEOSX for HPC, it's surely not very useful (except that you can get used to a working GEOSX and it's build system first before setting up your own env, which can be nice).
For non HPC related interns, they can start working in 10 minutes, not having to wait for accounts, permissions, etc.

So it's not the absolute solution, but more another string to our bow.
Like the auto deployment for us: surely not the best (in the absolute), we're aware of it, but a great tool nevertheless.

And finally, nobody can never say it's so complicated to start developing in GEOSX: in 3 minutes you are building GEOSX in a validated env.

.vscode/settings.json
TotoGaz: @francoishamon What do you think is best?

francoishamon: I agree with @jafranc, the default configuration should be hypre, this is the setting of most users

TotoGaz: done

TotoGaz:
@rrsettgast @CusiniM I've used a link which is built on the fly from `.vscode-codespaces` to `.vscode`.
So I think that the situation is nice now (w.r.t. conflicts with others' settings).
The `TPL` tag in the `.devcontainer/Dockerfile` is set to `DEFINE_ME` (I've added a comment on what to do).

CusiniM:
> @rrsettgast @CusiniM I've used a link which is built on the fly from `.vscode-codespaces` to `.vscode`. So I think that the situation is nice now (w.r.t. conflicts with others' settings). The `TPL` tag in the `.devcontainer/Dockerfile` is set to `DEFINE_ME` (I've added a comment on what to do).

Nice!

TotoGaz:
@rrsettgast @CusiniM is it possible to merge this PR as is?
It think there's no more conflict, and it can be used as an example on how to use `vscode` as well.

rrsettgast (approved):

CusiniM (approved):

PR #2113, opened 2022-10-07
Small fix to python script to generate sample mesh files.

CusiniM:
This is just a small fix to the script I found in our repo. We could probably add something to generate simple meshes to the mesh_tools but this worked for what I needed.

### DIFF
--- a/scripts/makeSampleMeshFiles.py
+++ b/scripts/makeSampleMeshFiles.py
@@ -5,7 +5,7 @@ import numpy as np
 import vtk
 
 
-def perturb(point_array, index_set):
+def perturb(point_array, index_set, max_nudge):
     # Option for later: parameterize spacing and change perturbation size
     nrows = np.size(point_array, 0)
 
@@ -17,7 +17,6 @@ def perturb(point_array, index_set):
     z_nudge = np.zeros((nrows, 1))
 
     # Make each perturbation between +/-0.1, uniform distribution
-    max_nudge = 0.1
     min_nudge = -1 * max_nudge
 
     if 'x' in index_set:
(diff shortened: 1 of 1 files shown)


scripts/makeSampleMeshFiles.py
```diff
@@ -265,14 +264,26 @@ def generate_all_types(n):
...
+
+    parser.add_argument("-sd",
+                        "--seed",
+                        dest="seed",
                         type=int,
                         nargs='?',
                         const=5,
                         default=5,
-                        help='Integer random number generator seed, default=5')
+                        help='Integer random number generator seed, default=5')
```
cssherman, line 282: It sounds like the random seed isn't guaranteed to produce the same results across versions of numpy/platforms, FYI.  If you are trying to limit diffs in the repo, it may be worth noting these somewhere

scripts/makeSampleMeshFiles.py
```diff
@@ -265,14 +264,26 @@ def generate_all_types(n):
...
-                        help='Integer random number generator seed, default=5')
+                        help='Integer random number generator seed, default=5')                                        
+
+
     args = parser.parse_args()
 
     for ncubes in args.sideLengths:
-        np.random.seed(seed)
-        generate_all_types(ncubes)
+        np.random.seed(args.seed)
```
cssherman, line 288: Is setting the seed in the loop intentional?

cssherman (approved):

scripts/makeSampleMeshFiles.py
CusiniM: I didn't really crate the original script but I am guessing the idea was to be able to reproduce the same meshes.

PR #2114, opened 2022-10-07
Modify compositional FaceBasedAssemblyKernel to work with interesecting fractures.

CusiniM:
In this PR we modify the compositional `FaceBasedAssemblyKernels` so that they work with intersecting fractures. In fact, the current implementation does not consider all connections in a `SurfaceElementStencil` entry. However, these modifications rely onthe assumption that a tpfa scheme is employed and fluxes can always be assembled by pairs.

It does not require a rebaseline.


rrsettgast:
Wait..there was no support for intersecting fractures at all?

CusiniM:
I get some numerical diffs with these modifications but they are weird...

Screen Shot 2022-10-10 at 4 36 24 PM

These are the logs.

[testLogs.zip](https://github.com/GEOSX/GEOSX/files/9749856/testLogs.zip)

I have a hard time understanding whether these are real diffs or not. Can anyone guess what else could be causing them?

francoishamon:
> I get some numerical diffs with these modifications but they are weird...
> 
> Screen Shot 2022-10-10 at 4 36 24 PM
> 
> These are the logs.
> 
> [testLogs.zip](https://github.com/GEOSX/GEOSX/files/9749856/testLogs.zip)
> 
> I have a hard time understanding whether these are real diffs or not. Can anyone guess what else could be causing them?
> 
> 2 things that I find weird:
> 
> * only some tests fail and not all of them ( I guess that's what I would expect if I had introduced a bug)
> * the only diffs are nonlinear iterations for a couple of tests and `phaseMobility_n` for one of them and `deltaPressure` for another. Plus 1 thermal tests seg faults but the others work fine.

Maybe we should first understand where the thermal case segfaults. It is actually the only one that uses the function that you modified, because the others are OBL thermal, so they have their own function to compute the fluxes. We could run the case in debug to see what is going on.

For the others, maybe the PR has introduced a tiny change in the Jacobian matrix that makes it slightly harder for GEOSX to converge? But still it converges to the same solution as before. It is really hard to tell without looking into the Jacobian matrix.

CusiniM:
> > I get some numerical diffs with these modifications but they are weird...
> > Screen Shot 2022-10-10 at 4 36 24 PM
> > These are the logs.
> > [testLogs.zip](https://github.com/GEOSX/GEOSX/files/9749856/testLogs.zip)
> > I have a hard time understanding whether these are real diffs or not. Can anyone guess what else could be causing them?
> > 2 things that I find weird:
> > 
> > * only some tests fail and not all of them ( I guess that's what I would expect if I had introduced a bug)
> > * the only diffs are nonlinear iterations for a couple of tests and `phaseMobility_n` for one of them and `deltaPressure` for another. Plus 1 thermal tests seg faults but the others work fine.
> 
> Maybe we should first understand where the thermal case segfaults. It is actually the only one that uses the function that you modified, because the others are OBL thermal, so they have their own function to compute the fluxes. We could run the case in debug to see what is going on.
> 
> For the others, maybe the PR has introduced a tiny change in the Jacobian matrix that makes it slightly harder for GEOSX to converge? But still it converges to the same solution as before, hence the absence of numerical diffs on the primary fields. It is really hard to tell without looking into the Jacobian matrix.

Thanks, this is useful intel! I will compare some matrices.

CusiniM:
Made some progress.... Now, the test that was giving a seg fault runs and everything is the same for the first iteration. Then, there are some diffs.

TotoGaz:
> everything is the same for the first iteration. Then, there are some diffs.

How many iterations? 🤣

rrsettgast:
@CusiniM @klevzoff I do have a question as to whether or not we could have a generic kernel that will perform the stencil looping, and then call member functions of the specific kernel for single phase or multiphase. Or, we could just pass the body of the loop in a lambda. It doesn't seem hard to do...but perhaps we haven't done it for a reason?

src/coreComponents/physicsSolvers/fluidFlow/ThermalCompositionalMultiphaseFVMKernels.hpp
```diff
@@ -280,9 +280,11 @@ class FaceBasedAssemblyKernel : public isothermalCompositionalMultiphaseFVMKerne
       : Base::StackVariables( size, numElems ),
       dCompFlux_dT( size, numComp ),
       energyFlux( 0.0 ),
-      dEnergyFlux_dP( size ),
-      dEnergyFlux_dT( size ),
-      dEnergyFlux_dC( size, numComp )
+      dEnergyFlux_dP( 2 ),
+      dEnergyFlux_dT( 2 ),
+      dEnergyFlux_dC( 2, numComp ),
```
francoishamon, line 285:
```suggestion
      dConvectiveEnergyFlux_dP( 2 ),
      dConvectiveEnergyFlux_dT( 2 ),
      dConvectiveEnergyFlux_dC( 2, numComp ),
```
Maybe we can rename the `energyFlux` as `convectiveEnergyFlux` to make the distinction with the new `conductiveEnergyFlux`.

src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseFVMKernels.hpp
```diff
@@ -582,246 +582,278 @@ class FaceBasedAssemblyKernel : public FaceBasedAssemblyKernelBase
...
-          stack.localFluxJacobian[ic][localDofIndexComp]          =  m_dt * stack.dCompFlux_dC[ke][ic][jc];
-          stack.localFluxJacobian[numEqn + ic][localDofIndexComp] = -m_dt * stack.dCompFlux_dC[ke][ic][jc];
+          for( integer ke = 0; ke < 2; ++ke )
+          {
+            localIndex const localDofIndexPres = k[ke] * numDof;
+            stack.localFluxJacobian[eqIndex0][localDofIndexPres] += m_dt * stack.dCompFlux_dP[ke][ic];
+            stack.localFluxJacobian[eqIndex1][localDofIndexPres] -= m_dt * stack.dCompFlux_dP[ke][ic];
+
+            // std::cout << "dCompFlux_dP [" << eqIndex0 << "]" << m_dt * stack.dCompFlux_dP[ke][ic] << std::endl;
+            // std::cout << "dCompFlux_dP [" << eqIndex1 << "]" << -m_dt * stack.dCompFlux_dP[ke][ic] << std::endl;
```
francoishamon, line 843:
```suggestion
```

francoishamon (approved):
Looks great!

CusiniM:
> @CusiniM @klevzoff I do have a question as to whether or not we could have a generic kernel that will perform the stencil looping, and then call member functions of the specific kernel for single phase or multiphase. Or, we could just pass the body of the loop in a lambda. It doesn't seem hard to do...but perhaps we haven't done it for a reason?

I think there used to be more differences between the 2. I don't see why we cannot have a generic stencil loop. I would get this merged first though so that we can add some compositional fracture tests before modifying the looping. Somehow even with these small modifications I had introduced some bugs and ti took me a while to find them.

src/coreComponents/physicsSolvers/fluidFlow/ThermalCompositionalMultiphaseFVMKernels.hpp
CusiniM: ah, yes, done.

src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseFVMKernels.hpp
```diff
@@ -476,8 +476,8 @@ class FaceBasedAssemblyKernel : public FaceBasedAssemblyKernelBase
       : stencilSize( size ),
       numFluxElems( numElems ),
       compFlux( numComp ),
-      dCompFlux_dP( size, numComp ),
-      dCompFlux_dC( size, numComp, numComp ),
+      dCompFlux_dP( 2, numComp ),
+      dCompFlux_dC( 2, numComp, numComp ),
```
TotoGaz, line 480: So I can see `2` popping here and there, while previously we had `size` and `maxStencilSize`.
Could you at least create constants so we know which `2` is when we need to refactor.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.cpp
```diff
@@ -452,6 +452,7 @@ void CompositionalMultiphaseFVM::applyBoundaryConditions( real64 time_n,
 {
   GEOSX_MARK_FUNCTION;
 
+  // GEOSX_UNUSED_VAR( time_n, dt, domain, dofManager, localMatrix, localRhs );
```
TotoGaz, line 455: please remove.

src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseFVMKernels.hpp
CusiniM: I can but this assembly pattern only works with a `two-point flux approximation` so it is important to keep in mind that it becomes a constant that can't really be changed. But I guess I will just define it directly in this file.

CusiniM: Actually, what we should do (but in a separate PR) is a review of the meaning of some of the constants defined in `StencilBase` coz with all the fracture work some of them have kind of changed meaning. And we could probably remove the MPFA  stencil.

TotoGaz: If `size` and `maxStencilSize` have become identical, maybe add a `STENCIL_SIZE = 2`.
Eventually having a review of the constants looks nice indeed.

src/coreComponents/physicsSolvers/fluidFlow/ThermalCompositionalMultiphaseFVMKernels.hpp
```diff
@@ -482,53 +514,46 @@ class FaceBasedAssemblyKernel : public isothermalCompositionalMultiphaseFVMKerne
...
+    localIndex k[2];
+    localIndex connectionIndex = 0;
 
-    // Step 1: add dCompFlux_dTemp to localFluxJacobian
-    for( integer ic = 0; ic < numComp; ++ic )
+    for( k[0] = 0; k[0] < stack.numConnectedElems; ++k[0] )
     {
-      for( integer ke = 0; ke < stack.stencilSize; ++ke )
+      for( k[1] = k[0] + 1; k[1] < stack.numConnectedElems; ++k[1] )
       {
```
TotoGaz, line 526: Since you do not loop over `k`, maybe a simpler solution like 
```c++
    for( int k0 = 0; k0 < stack.numConnectedElems; ++k0 )
    {
      for( int k1 = k0 + 1; k1 < stack.numConnectedElems; ++k1 )
      {
```
This reduces the visibility of `k`.
And also, is it the same `2` as in `numFluxSupportPoints `?
Here, it's a `2` because it's the size of a pair, isn't it? Would it change with other schemes?
We could also even rename `k0, k1` with other names?

src/coreComponents/physicsSolvers/fluidFlow/ThermalCompositionalMultiphaseFVMKernels.hpp
```diff
@@ -329,32 +313,46 @@ class FaceBasedAssemblyKernel : public isothermalCompositionalMultiphaseFVMKerne
...
     //
     // Computing dCompFlux_dT and the enthalpy flux requires quantities already computed in the base computeFlux,
     // such as potGrad, phaseFlux, and the indices of the upwind cell
     // We use the lambda below (called **inside** the phase loop of the base computeFlux) to access these variables
     Base::computeFlux( iconn, stack, [&] ( integer const ip,
+                                           localIndex const (&k)[2],
+                                           localIndex const (&seri)[2],
+                                           localIndex const (&sesri)[2],
+                                           localIndex const (&sei)[2],
+                                           localIndex const connectionIndex,
```
TotoGaz, line 326: Do we want to replace the `2` here as well?

src/coreComponents/physicsSolvers/fluidFlow/ThermalCompositionalMultiphaseFVMKernels.hpp
CusiniM: This assembly pattern will only work for a two point scheme. Honestly, I would not spend too much time refining this. I feel this looping pattern should be generalized and moved somewhere else so that it can be used both here and in the singlePhase kernels. Right now they look identical so I would prefer keeping this just as it is and then change both at once.

TotoGaz: OK

PR #2109, opened 2022-10-06
Add RESQML input

untereiner:
created branch for adding RESQML input via [fesapi](https://github.com/F2I-Consulting/fesapi/).

PR #2117, opened 2022-10-10
add flags to enable fesapi

untereiner:
created branch to add flags to enable fesapi from the third parties (related to #2109)

related to https://github.com/GEOSX/thirdPartyLibs/pull/202


rrsettgast:
@bmhan12 @wrtobin There needs to be some spack/uberenv work in conjunction with this?

untereiner:
@rrsettgast I think Fesapi would need some spack work. Fesapi has zlib, minizip and boost (only headers) as dependencies.
I checked that boost is already in spack and available as a header only dependency. It is exactly what fesapi needs. 
It would even be easier than with CMake's ExternalProject_Add function.

TotoGaz:
> @rrsettgast I think Fesapi would need some spack work. Fesapi has zlib, minizip and boost (only headers) as dependencies. I checked that boost is already in spack and available as a header only dependency. It is exactly what fesapi needs. It would even be easier than with CMake's ExternalProject_Add function.

But on the long term, what would be your exact requirements? Do you have a clear picture on this?

klevzoff:
We'll need to implement a spack package recipe for fesapi, but it can be done after the initial integration into GEOSX is complete IMO. For now, to avoid breaking spack build, I propose we modify it to explicitly disable fesapi (spit out `ENABLE_FESAPI=OFF` into the generated host-config).

untereiner:
> have



> But on the long term, what would be your exact requirements? Do you have a clear picture on this?

In a first step for file based access:
- hdf5 (default numerical values storage for RESQML) 
- minizip (in order to deal with zip files)
- boost (only headers for having uuid support).

In a long-term view with connection to OSDU:
- [FETPAPI](https://github.com/F2I-Consulting/fetpapi)
- AVRO (for serialization and deserialization of message)
- Boost (libraries, not only headers, mainly Beast sublibrary for websocket support) 
- OpenSSL (TLS support).

untereiner:
I think it is good now

src/cmake/GeosxOptions.cmake
```diff
@@ -123,6 +123,15 @@ blt_append_custom_compiler_flag( FLAGS_VAR GEOSX_NINJA_FLAGS
...
                                )
 
+# clang-13 and gcc complains about unused-but-set variable.
+include(CheckCXXCompilerFlag)
+CHECK_CXX_COMPILER_FLAG("-Wunused-but-set-variable" CXX_UNUSED_BUT_SET_VAR)
+if (ENABLE_GBENCHMARK)
+    blt_add_target_compile_flags(TO benchmark
+                                FLAGS $<$<AND:${CXX_UNUSED_BUT_SET_VAR},$<COMPILE_LANGUAGE:CXX>>:-Wno-unused-but-set-variable>
+                                )
+endif()
```
TotoGaz, line 133: Can you remind me what was the problem that lead to this patch

untereiner: with clang 13.4 (image xcode13.4) this flag detects an unused variable in gbenchmark (you can find the error in the logs). It will also happen with a recent gcc (mine is gcc 12.2). I just disable it for the  benchmark target.

TotoGaz (approved):

bmhan12 (approved):

untereiner:
@TotoGaz  Is it possible to merge both PRs in the next hours ?

TotoGaz:
You are second in line https://github.com/GEOSX/GEOSX/projects/18

TotoGaz:
> ok, do you have any idea when it will happen ?

@rrsettgast is going to give https://github.com/GEOSX/GEOSX/pull/2149 a last view.

PR #2123, opened 2022-10-14
Acoustic wave solver optimization

sframba:
This PR provides a significant boost to the performance of the Acoustic wave propagator. In particular:
 * The assembly of the (approximate) stiffness matrix exploits the properties of the SEM method to reduce the number of matrix elements needed. The matrix is now constructed as in [1]. This leads to a complexity of $p^5$ instead of $p^9$, with $p$ being the polynomial order of the numerical scheme;
 * An unneeded loop was removed from the mass matrix computation;
 * The damping matrix is computed via a slightly more efficient loop on faces.

Moreover, the C++ structure of Gauss-Lobatto discretizations in the code is simplified by using a single template class `Qk_Hexahedron_Lagrange_GaussLobatto`. This PR uses it to add the Q2 and Q4 formulations that were missing so far. 

In order to specialize the SEM implementation and distinguish it from the FE case, the optional tag `formulation` of the `FiniteElementSpace` section in the `xml` input has been used. One now has to specify `SEM` instead of `default` when running the Acoustic SEM solver to activate the SEM method. No change is needed in the input file for the other solvers.

Another modification needed to enable high-order SEM solvers, originally from PR [2045](https://github.com/GEOSX/GEOSX/pull/2045), is included: the removal of the auxiliary Face Manager `faceManager0`, originally added by @acitrain. This resolves issue [2122](https://github.com/GEOSX/GEOSX/issues/2122).

The results have been checked and the gain in performance seems to be very significant, with a factor of 2 in running time for Q1 and a factor of 50 (!) for Q3, cf. PR [2084](https://github.com/GEOSX/GEOSX/pull/2084).
After this PR is merged, PR [2045](https://github.com/GEOSX/GEOSX/pull/2045) will no longer be needed.

[1] M. Duruflé, [_Intégration numérique et éléments finis d'ordre élevé appliqués aux équations de Maxwell en régime harmonique_](https://pastel.archives-ouvertes.fr/tel-00068590/document) (in French)


acitrain (approved):

francoishamon (approved):
Looks good to me

sframba:
@YohannDudouit, @rrsettgast is it ok to merge ?

PR #2124, opened 2022-10-14
Output hypre version

victorapm:
This PR adds an output for the hypre version being used. This helps the reproducibility of numerical experiments.

Output example (first three lines):
```
Max threads: 20
Using hypre version: v2.26.0-0-g3c04212d7
GEOSX version 0.2.0 (feature/victorapm/hypre-version, sha1: 3fdf93dbb)
```


rrsettgast:
@victorapm Why is `HYPRE_DEVELOP_STRING` not defined on some platforms?

victorapm:
@rrsettgast, the variable `HYPRE_DEVELOP_STRING` is defined only when configuring hypre from a folder that contains git data (`.git` exists on the top level). Thus, it is not defined when configuring hypre from a tarball/zip file. I suppose that's what happened for the platforms you mentioned. I've added a print for `HYPRE_RELEASE_VERSION`, which is always defined during hypre's configure, in d55c14b for the case when `HYPRE_DEVELOP_STRING` is not defined.

rrsettgast:
@victorapm The hype version should be after the GEOSX version, and indented a couple of spaces.

klevzoff:
> The hype version should be after the GEOSX version

This might not be super straightforward to do... GEOSX version is printed using Logger macros which aren't usable until the call to `basicSetup()`, but that call is also the one that initializes hypre. We might need to refactor the code a bit to split "environment initialization" and "print config" phases.

victorapm:
@rrsettgast @klevzoff How about moving the call at https://github.com/GEOSX/GEOSX/blob/b854b7c8b740542883afe02190d0c3cf9f1f5b59/src/main/main.cpp#L36 to Line 34 (before the `basicSetup` call)?

victorapm:
How about moving the line https://github.com/GEOSX/GEOSX/blob/b854b7c8b740542883afe02190d0c3cf9f1f5b59/src/main/main.cpp#L36 right before https://github.com/GEOSX/GEOSX/blob/b854b7c8b740542883afe02190d0c3cf9f1f5b59/src/coreComponents/mainInterface/initialization.cpp#L260

klevzoff:
> How about moving the line

Yes, looks like this should work

rrsettgast:
Maybe we should be doing this with all our submodules and TPL's. Submodules are less important because the hash is related to the GEOSX hash....unless we are looking at output from a branch where the submodule was modified and not updated at the time of the execution.

victorapm:
@rrsettgast @klevzoff let's pause the review of this PR. I realized the hypre's branch information is not correct here and will need to make changes in the hypre side to fix that. I'll let you know when this PR is ready again. Thanks!

src/coreComponents/linearAlgebra/interfaces/hypre/HypreInterface.cpp
```diff
@@ -42,6 +42,14 @@ void HypreInterface::initialize()
...
 #endif
   HYPRE_SetMemoryLocation( hypre::memoryLocation );
+
+  // Hypre version info
+#if defined(HYPRE_DEVELOP_STRING) && defined(HYPRE_BRANCH_NAME)
+  GEOSX_LOG_RANK_0( "  Using hypre development version: " << HYPRE_DEVELOP_STRING <<
+                    " (" << HYPRE_BRANCH_NAME << ")" );
+#elif defined(HYPRE_RELEASE_VERSION)
+  GEOSX_LOG_RANK_0( "  Using hypre release version: " << HYPRE_RELEASE_VERSION );
+#endif
```
rrsettgast, line 52: if either `HYPRE_DEVELOP_STRING` , and `HYPRE_BRANCH_NAME` is not defined, this will pass through without any output.

rrsettgast (approved):

src/coreComponents/linearAlgebra/interfaces/hypre/HypreInterface.cpp
victorapm: We always have `HYPRE_RELEASE_VERSION` defined from hypre, so there will be an output from line 51. I could have replaced line 50 with just an `#else` but decided to do `#elif defined(HYPRE_RELEASE_VERSION)`

PR #2126, opened 2022-10-17
Import face block from vtk

TotoGaz:
First WIP version


TotoGaz:
Hello @CusiniM , as a first step, can you have a look at `VTKFaceBlockUtilities.cpp` mostly?
The rest needs to be cleaned afterwards and is more straightforward.

src/coreComponents/mesh/generators/VTKFaceBlockUtilities.cpp
```diff
@@ -0,0 +1,633 @@
...
+  GEOSX_ERROR_IF( fieldData == nullptr, "Fault and fracture information is supposed to be defined in the VTK field data. None was found." );
+
+  FaceData const faceData( fieldData, faceBlockName );
+
+  vtkIdType const num2dElements = faceData.num2dElements();
+  localIndex const num2dFaces = computeNum2dFaces( vtkMesh, faceData );
+
+  internal::ElementToFace const elemToFaces( cellBlockManager.getCellBlocks() );
+
+  ToCellRelation< array2d< localIndex > > elem2dToElems = compute2dElemToElems( faceData, elemToFaces );
```
CusiniM, line 606: maybe, since in the code they will eventually be called `faceElements` it would be better to refer to them as such even here. `elem2dToElems` -> `faceElementToElems`

TotoGaz: While I get your point, currently the [`FaceBlockAPI`](https://github.com/GEOSX/GEOSX/blob/develop/src/coreComponents/mesh/generators/FaceBlockABC.hpp) uses the `2dElems` everywhere. So I wanted to be consistent in that regard?
Don't you think it would be better to make all the renaming at once?
Because we don't know when we'll be able to do it... not if it's gonna be the selected wording: if `faceElement` looks nice, what about `edgeFace`? Is it better than `2dFace`?

CusiniM: Yes, it would be better to do it at once. I am also not sure what the best names are for the other things like `2dFace`, etc... I was just trying to think of something easy coz I personally find it a bit hard while reviewing this PR. I keep forgetting what a 2dFace actually is. But, TBH, this 2d elements concept is confusing in itself. 

In general, this function `importFracture` (and in general this entire file) seems good to me.

src/coreComponents/mesh/generators/VTKFaceBlockUtilities.cpp
```diff
@@ -0,0 +1,633 @@
...
+    localIndex const & edge = face2dToEdges[fi];
+    std::vector< int > const & elems2d = edgesToElems2d[edge];
+    face2dToElems2d.resizeArray( fi, elems2d.size() );
+    std::copy( elems2d.begin(), elems2d.end(), face2dToElems2d[fi].begin() );
+  }
+
+  return face2dToElems2d;
+}
+
+void importFracture( string const & faceBlockName,
```
CusiniM, line 592: Maybe it should be called `importFractureNetwork` coz it can actually import multiple fractures at once can't it?

TotoGaz: Yes that's right.

TotoGaz: To be precise, it can import one network of multiple fractures, yes!

src/coreComponents/mesh/ElementRegionManager.cpp
```diff
@@ -126,6 +127,26 @@ void ElementRegionManager::generateMesh( CellBlockManagerABC & cellBlockManager
...
+
+  // Some mappings of the surfaces subregions point to elements in other subregions and regions.
+  // For the moment, those mappings only point to cell block indices.
+  // The following makes use of cell block to subregions mappings to finalize the surfaces information.
+  // It also creates the mappings concerning the regions.
+  array2d< localIndex > const blockToSubRegion = this->getCellBlockToSubRegionMap( cellBlockManager );
+  this->forElementRegions< SurfaceElementRegion >( [&]( SurfaceElementRegion & elemRegion )
+  {
+    FaceElementSubRegion & subRegion = elemRegion.getUniqueSubRegion< FaceElementSubRegion >();
+    // While indicated as a `ToElementRelation`, the relation contains cell block information.
```
CusiniM, line 139:
```suggestion
    // While indicated as a `ToCellRelation`, the relation contains cell block information
```
?

src/coreComponents/mesh/FaceElementSubRegion.cpp
```diff
@@ -79,17 +79,64 @@ FaceElementSubRegion::FaceElementSubRegion( string const & name,
...
+    };
+
+    std::vector< integer > sizes( num2dElements );
+    for( int i = 0; i < num2dElements; ++i )
+    {
+      sizes[i] = m_toNodesRelation[i].size();
+    }
+    std::set< integer > const s( sizes.cbegin(), sizes.cend() );
+
+    if( s.size() > 1 )
```
CusiniM, line 118: Let's maybe add a short comment of what's happening here.

src/coreComponents/mesh/generators/VTKFaceBlockUtilities.cpp
```diff
@@ -0,0 +1,636 @@
...
+  {
+    // First, formatting the duplicated nodes' information.
+    {
+      string const DUPLICATED_POINTS_INFO = "duplicated_points_info";
+      vtkDataArray * duplicatedNodes = fieldData->GetArray( DUPLICATED_POINTS_INFO.c_str() );
+      GEOSX_ERROR_IF( duplicatedNodes == nullptr, "Mandatory VTK field data array \"" + DUPLICATED_POINTS_INFO + "\" not found." );
+
+      // vtk will provide the field as a (numComponents, numTuples) table.
+      // numTuples is actually the number of duplicated nodes too.
+      // But numComponents is just the maximum number of nodes multiplied nodes at one location.
```
CusiniM, line 74:
```suggestion
      // But numComponents is just the maximum number of multiplied nodes at one location.
```

src/coreComponents/mesh/generators/VTKFaceBlockUtilities.cpp
```diff
@@ -0,0 +1,636 @@
...
+      // For a T shaped fracture, the nodes at the T intersection will be tripled.
+      // But all the other nodes will only get doubled.
+      // Nevertheless, numComponents will equal 3 for all the nodes.
```
CusiniM, line 77:
```suggestion
      // Nevertheless, numComponents will be equal 3 for all the nodes.
```

src/coreComponents/mesh/generators/VTKFaceBlockUtilities.cpp
```diff
@@ -0,0 +1,636 @@
...
+  /**
+   * @brief Given the global (vtk) id, returns the cell block index @p ei belongs to.
+   * @param[in] ei global cell index.
+   * @return The cell block index.
+   */
+  int getCellBlockIndex( int const & ei ) const
+  {
+    auto const it = std::find_if( m_ranges.cbegin(), m_ranges.cend(), [&]( auto const & p ) -> bool
+    {
+      return p.first <= ei and ei <= p.second;
```
CusiniM, line 253:
```suggestion
      return p.first <= ei && ei <= p.second;
```

TotoGaz:
@CusiniM I've done all your comments.

PR #2138, opened 2022-10-21
Adding version info to the terminal output

rrsettgast:
Adds version info for:

- [x]  compiler version
- [x]  MPI version
- [x]  HDF5 version
- [x]  VTK version
- [x]  RAJA version
- [x]  UMPIRE version
- [x]  Conduit Version
- [x] caliper version
- [x] adiak version
- [x] metis version
- [x] parametis version

### DIFF
--- a/src/coreComponents/common/initializeEnvironment.cpp
+++ b/src/coreComponents/common/initializeEnvironment.cpp
@@ -202,12 +202,6 @@ void setupCaliper( cali::ConfigManager & caliperManager,
   adiak::value( "CUDA runtime version", cudaRuntimeVersion );
   adiak::value( "CUDA driver version", cudaDriverVersion );
 
-  GEOSX_LOG_RANK_0( "  - caliper version: "<<cali_caliper_version() );
-
-  GEOSX_LOG_RANK_0( "  - adiak version: "<<
-                    ADIAK_VERSION<<"."<<
-                    ADIAK_MINOR_VERSION<<"."<<
-                    ADIAK_POINT_VERSION );
 }
 #endif // defined( GEOSX_USE_CALIPER )
 
--- a/src/coreComponents/common/initializeEnvironment.cpp
+++ b/src/coreComponents/common/initializeEnvironment.cpp
@@ -201,6 +201,11 @@ void setupCaliper( cali::ConfigManager & caliperManager,
 #endif
   adiak::value( "CUDA runtime version", cudaRuntimeVersion );
   adiak::value( "CUDA driver version", cudaDriverVersion );
+
+  GEOSX_LOG_RANK_0( "  - Caliper Version: "<<cali_caliper_version() );
+
+cali_caliper_version()
+
 }
 #endif // defined( GEOSX_USE_CALIPER )
 
--- a/src/coreComponents/common/initializeEnvironment.cpp
+++ b/src/coreComponents/common/initializeEnvironment.cpp
@@ -202,10 +202,12 @@ void setupCaliper( cali::ConfigManager & caliperManager,
   adiak::value( "CUDA runtime version", cudaRuntimeVersion );
   adiak::value( "CUDA driver version", cudaDriverVersion );
 
-  GEOSX_LOG_RANK_0( "  - Caliper Version: "<<cali_caliper_version() );
+  GEOSX_LOG_RANK_0( "  - caliper version: "<<cali_caliper_version() );
 
-cali_caliper_version()
-
+  GEOSX_LOG_RANK_0( "  - adiak version: "<<
+                    ADIAK_VERSION<<"."<<
+                    ADIAK_MINOR_VERSION<<"."<<
+                    ADIAK_POINT_VERSION );
 }
 #endif // defined( GEOSX_USE_CALIPER )
 
--- a/src/coreComponents/linearAlgebra/interfaces/hypre/HypreInterface.cpp
+++ b/src/coreComponents/linearAlgebra/interfaces/hypre/HypreInterface.cpp
@@ -44,11 +44,11 @@ void HypreInterface::initialize()
   HYPRE_SetMemoryLocation( hypre::memoryLocation );
 
   // Hypre version info
-#if defined(HYPRE_DEVELOP_STRING) && defined(HYPRE_BRANCH_NAME)
-  GEOSX_LOG_RANK_0( "  Using hypre development version: " << HYPRE_DEVELOP_STRING <<
+#if defined(HYPRE_DEVELOP_STRING) 
+  GEOSX_LOG_RANK_0( "  - hypre development version: " << HYPRE_DEVELOP_STRING <<
                     " (" << HYPRE_BRANCH_NAME << ")" );
 #elif defined(HYPRE_RELEASE_VERSION)
-  GEOSX_LOG_RANK_0( "  Using hypre release version: " << HYPRE_RELEASE_VERSION );
+  GEOSX_LOG_RANK_0( "  - hypre release version: " << HYPRE_RELEASE_VERSION );
 #endif
 }
 
--- a/src/coreComponents/linearAlgebra/interfaces/hypre/HypreInterface.cpp
+++ b/src/coreComponents/linearAlgebra/interfaces/hypre/HypreInterface.cpp
@@ -45,8 +45,12 @@ void HypreInterface::initialize()
 
   // Hypre version info
 #if defined(HYPRE_DEVELOP_STRING)
+
+#if defined(HYPRE_BRANCH_NAME)
   GEOSX_LOG_RANK_0( "  - hypre development version: " << HYPRE_DEVELOP_STRING <<
                     " (" << HYPRE_BRANCH_NAME << ")" );
+#else
+  GEOSX_LOG_RANK_0( "  - hypre development version: " << HYPRE_DEVELOP_STRING );
 #elif defined(HYPRE_RELEASE_VERSION)
   GEOSX_LOG_RANK_0( "  - hypre release version: " << HYPRE_RELEASE_VERSION );
 #endif
--- a/src/coreComponents/linearAlgebra/interfaces/hypre/HypreInterface.cpp
+++ b/src/coreComponents/linearAlgebra/interfaces/hypre/HypreInterface.cpp
@@ -45,12 +45,12 @@ void HypreInterface::initialize()
 
   // Hypre version info
 #if defined(HYPRE_DEVELOP_STRING)
-
 #if defined(HYPRE_BRANCH_NAME)
   GEOSX_LOG_RANK_0( "  - hypre development version: " << HYPRE_DEVELOP_STRING <<
                     " (" << HYPRE_BRANCH_NAME << ")" );
 #else
   GEOSX_LOG_RANK_0( "  - hypre development version: " << HYPRE_DEVELOP_STRING );
+#endif
 #elif defined(HYPRE_RELEASE_VERSION)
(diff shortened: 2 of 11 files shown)


rrsettgast:
on lassen, this looks like 
```
GEOSX version: 0.2.0 (feature/settgast/outputVersionInfo, sha1: c1308b34f)
  - clang version: 9.0.0
  - cuda version: 10.1
  - openmp version: 201107
  - MPI version: Open MPI v10.3.1.03rtm0, package: IBM Spectrum MPI, ident: IBM Spectrum MPI, repo rev: IBM_SPECTRUM_MPI_10.03.01.02_2020.08.03_XAA0, Unreleased developer copy
  - HDF5 version: 1.12.1
  - VTK version: 9.1.0
  - RAJA version: 2022.3.0
  - UMPIRE version: 2022.3.0
  - Conduit Version: 0.8.2
  - metis version: 5.1.0
  - parmetis version: 4.0.3
  - superlu_dist version: 6.3.0 (February 23, 2020)
  - umfpack version: UMFPACK V5.7.9 (O
  ```

File: src/coreComponents/linearAlgebra/interfaces/hypre/HypreInterface.cpp [hunk 1]
```diff
@@ -44,11 +44,11 @@ void HypreInterface::initialize()
   HYPRE_SetMemoryLocation( hypre::memoryLocation );
 
   // Hypre version info
-#if defined(HYPRE_DEVELOP_STRING) && defined(HYPRE_BRANCH_NAME)
-  GEOSX_LOG_RANK_0( "  Using hypre development version: " << HYPRE_DEVELOP_STRING <<
+#if defined(HYPRE_DEVELOP_STRING)
```
victorapm, line 47: Should this be moved to the newly added `outputVersionInfo` such that all the version stuff can be found in a single place?

src/coreComponents/common/initializeEnvironment.cpp
```diff
@@ -201,6 +201,11 @@ void setupCaliper( cali::ConfigManager & caliperManager,
 #endif
   adiak::value( "CUDA runtime version", cudaRuntimeVersion );
   adiak::value( "CUDA driver version", cudaDriverVersion );
+
+  GEOSX_LOG_RANK_0( "  - Caliper Version: "<<cali_caliper_version() );
```
victorapm, line 205: Same comment as for hypre.

victorapm (approved):
Thanks!!

src/coreComponents/mainInterface/initialization.cpp
```diff
@@ -253,12 +256,92 @@ std::unique_ptr< CommandLineOptions > parseCommandLineOptions( int argc, char *
...
+                      RAJA_VERSION_PATCHLEVEL );
+
+    GEOSX_LOG_RANK_0( "  - UMPIRE version: " << 
+                      umpire::get_major_version()<<"."<<
+                      umpire::get_minor_version()<<"."<<
+                      umpire::get_patch_version() );
+  }
+
+  {
+    GEOSX_LOG_RANK_0( "  - Conduit Version: "<<CONDUIT_VERSION );
```
victorapm, line 331: Very minor... "Version" -> "version"

src/coreComponents/linearAlgebra/interfaces/hypre/HypreInterface.cpp
victorapm: Also, I think it's better to leave the check `defined(HYPRE_BRANCH_NAME)` in there because this variable has been added this week while `HYPRE_DEVELOP_STRING` exists for nearly a year. If someone tries to use a commit in hypre in that window (last year - this week), this would break. This is very unlikely, but just to be in the same side...

rrsettgast: > Also, I think it's better to leave the check `defined(HYPRE_BRANCH_NAME)` in there because this variable has been added this week while `HYPRE_DEVELOP_STRING` exists for nearly a year. If someone tries to use a commit in hypre in that window (last year - this week), this would break. This is very unlikely, but just to be in the safe side...

Yeah...I had something more elaborate here....I don't know what happened....I'll fix it again.

src/coreComponents/mainInterface/initialization.cpp
```diff
@@ -253,12 +263,112 @@ std::unique_ptr< CommandLineOptions > parseCommandLineOptions( int argc, char *
 }
 
 
+void outputVersionInfo()
+{
+#if defined(__clang__)
```
victorapm, line 268:
```suggestion
#if defined(__INTEL_COMPILER)
  GEOSX_LOG_RANK_0( "  - Intel compiler version: " << __INTEL_COMPILER << " (" << __VERSION << ")" );
#endif
```

src/coreComponents/mainInterface/initialization.cpp
rrsettgast: thanks

src/coreComponents/mainInterface/initialization.cpp
rrsettgast: I stopped trying to support intel compilers a few years back. Are they better these days? Is there a compelling reason to support them?

src/coreComponents/mainInterface/initialization.cpp
```diff
@@ -21,7 +21,17 @@
...
 
 // TPL includes
+#include <conduit/conduit.hpp>
+#include <hdf5.h>
+#include <metis.h>
 #include <optionparser.h>
+#include <parmetis.h>
+#include <superlu_defs.h>
+#include <umfpack.h>
+#include <vtkVersion.h>
```
klevzoff, line 31: I'm not sure how I feel about a soup of all TPL includes in one place. For some of them, we've gone through some effort to restrict their inclusion to a single translation unit. Of course, this adds just one more unit so no big deal, and it's not like there are any actual conflicts between them presently. 

Some of them could also be handled by adding a `getVersionString()` function in their respective GEOSX "interfaces" (e.g. in `SuperLUDist.[hpp|cpp]`) that hide away the TPL includes and macros (still have to include the GEOSX-side headers here though). Others I guess don't have a single entry point in the code.

klevzoff (approved):
Looks good, just have minor/optional suggestions

src/coreComponents/mainInterface/initialization.cpp
victorapm: > We stopped trying to support intel compilers

Didn't know about that...

src/coreComponents/mainInterface/initialization.cpp
TotoGaz: I was about to comment the same thing (or so).
If this PR is merged as is, we'll end up with the main initialization depending on TPLs that have nothing to do here.
And this, regardless of protecting with `#ifdef/#endif` guards which is just a technical pirouette when we don't use it: when we do use the TPL, the issue is there.

`vtk` being a straightforward illustration since it should be invisible except where we're dealing with input and outputs.
If we adopt this pattern, we'll end up having the main interface depending on everything. And I expect that a lot of things depend on the main interface too...

> I'm not sure how I feel about a soup of all TPL includes in one place.

Personally I don't feel that's long term solution. For the dependency/architecture considerations more than for the translation units though.
Having neat and carefully trimmed dependencies is key to long term success.

To move forward, do we have other open-source software to get inspiration?
I expect some would generate some bootstrap tool to generate env info?

klevzoff: I actually don't have too much of a problem with link-level dependencies, since `mainInterface` _by design_ is already dependent on everything else, and `initialization.cpp`'s role is specifically to initialize all dependencies. It seems if we want to have organized output of **runtime** dependency versions in one place, there's little chance to avoid this pattern. An alternative is to generate and inject a dependency version string at config time via CMake?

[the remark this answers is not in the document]
rrsettgast: Yeah. I had written a guard using `GEOSX_USE_VTK` until i realized it didn't exist.

src/coreComponents/mainInterface/initialization.cpp
```diff
@@ -253,12 +263,112 @@ std::unique_ptr< CommandLineOptions > parseCommandLineOptions( int argc, char *
...
+                    __ibmxl_release__<<"."<<
+                    __ibmxl_modification__<<"."<<
+                    __ibmxl_ptf_fix_level__ );
+#endif
+#elif defined(__GNUC__)
+  GEOSX_LOG_RANK_0( "  - gcc version: " <<
+                    __GNUC__<<"."<<
+                    __GNUC_MINOR__<<"."<<
+                    __GNUC_PATCHLEVEL__ );
+#endif
```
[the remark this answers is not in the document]
rrsettgast: Yeah...that is probably a better idea. Correct me if I am wrong, but `getCompilerIdString()` would just have a bunch of guarded statements like it is implemented, but instead of the `GEOSX_LOG_RANK_0()`, they would just fill the return string. Is that what you had in mind?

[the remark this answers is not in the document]
klevzoff: Yes, exactly what I was thinking of

src/coreComponents/mainInterface/initialization.cpp
rrsettgast: I think it cmake of what is in this PR. Creating a getVersionString() for each dependency requires an interface wrapper for each dependency, and we still need to have the interface included in the `mainInterface`, so it really doesn't resolve anything I think?

@TotoGaz If you have a solution that you like, please feel free to suggest it. I added this PR because I thought we had a need, and I didn't think of anything better within reason and time constraints for implementing it. Perhaps the cmake solution would be the cleanest way I can think of.

src/coreComponents/mainInterface/initialization.cpp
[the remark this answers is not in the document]
rrsettgast: I shouldn't be so lazy...not that I have time for this sort of "work".

src/coreComponents/mainInterface/initialization.cpp
rrsettgast: OK. We can use the cmake variables `PACKAGE_VERSION` created from the `find_package` call for: `Conduit, pugixml, RAJA, CHAI, Umpire, Adiak, Caliper, VTK, fmt, Python3, CUDAToolkit`. The rest are going to either involve parsing headers for their version info, or including like is done currently in this PR.

TotoGaz: > since mainInterface by design is already dependent on everything else, and initialization.cpp's role is specifically to initialize all dependencies.

When I read the code, I understand that the role is more to initialize the environment in which geosx is embedded.
Like `MPI`, the `logger`, etc. What's the "lingua franca" of GEOSX.
And in a quite agnostic way for the LAI (even though the implementation and the registration could be done in a more separate way: despite some details, `initialization` could be unaware of the `LAI` it's actually triggering).

In a similar way, couldn't the `LAI` provide it's own `getLAIVersion() -> string or map<string, string>` which can be displayed on screen? At least the current state of dependencies is not shifted.
I think we still need a solution for `metis`, `parmetis`, `MPI`, `HDF5`?
For `MPI`, I think a solution equivalent to the compilers' solutions is acceptable.

For `HDF5`, calling any of its executable outputs the version. e.g. 
```
# ./bin/h5ls -V
h5ls: Version 1.12.1
```
For `(par)metis` I don't know yet because their tools do not provide this kind of information.

PR #2140, opened 2022-10-23
Update Multi-resolution publication data.

CusiniM:


jhuang2601:
@CusiniM The papers needs to be ordered by the family name of the first author.

jhuang2601:
> > @CusiniM The papers needs to be ordered by the family name of the first author.
> 
> ah, I thought the order was based on publication date.

Nicola told me this when I added a new publication couple weeks ago.

jhuang2601 (approved):

PR #2142, opened 2022-10-28
Add python unit tests

cssherman:


src/docs/doxygen/GeosxConfig.hpp
```diff
@@ -141,7 +141,7 @@
 #define fmt_VERSION 8.0.1
 
 /// Version information for python
-/* #undef Python3_VERSION */
+#define Python3_VERSION 3.8.5
```
TotoGaz, line 144: I'm surprised that this file is committed...
Where do you get this `3.8.5` from? Is it generated through `GeosxConfig.in`?

src/pygeosx/unitTests/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+
+
+add_custom_target( pygeosx_unit_tests
+                   COMMAND ${CMAKE_BINARY_DIR}/lib/PYGEOSX/bin/python3 -m pip install pytest --quiet --disable-pip-version-check
```
TotoGaz, line 4: I do not think that the installation should be done there, shouldn't it?
Don't you end up trying to trigger the installation everytime?

src/docs/doxygen/GeosxConfig.hpp
cssherman: Yeah, I have seen this happen before on occasion...  It is definitely generated as part of the config/build process, and we should try to squash it.

src/pygeosx/unitTests/CMakeLists.txt
cssherman: Long term, no, I don't think we should be doing this here.  For systems that already have pytest installed, pip just silently checks that the required packages are installed and moves on.

Right now, we are installing pygeosx by:

1) Creating a virtualenv off of the base python distribution (bin/lib/PYGEOSX)
2) Manually linking the pygeosx/pylvarray libraries in the PYGEOSX env

Since we aren't using the typical setup process for a package, we aren't enforcing required packages anywhere (aside from build errors if required python/package headers aren't present).  Any thoughts?  Also, are we targeting any systems where pip install is not viable?

cssherman:
@TotoGaz - here are some other unit tests that I have in mind:

1) problem.run and pygeosx return methods
2) wrapper access with the modifiable flags turned off

Any other obvious ones stand out?  Do we need to test the behavior in parallel runs, or would it be fine to add them to pygeosx_tools (which has some methods for mpi calls, etc.)?

src/pygeosx/unitTests/CMakeLists.txt
TotoGaz: I would say that it's not GEOSX's responsibility to provide a way to install `pytest`.
Furthermore, there are so many ways to build envs in python that even imposing `pip` may not be right.

TotoGaz:
I actually don't use `pygeosx` enough to be helpful here.
I would say that testing the parallel behavior for tiny runs makes sense. Where does it make most sense?
Is the proposed approach compatible with what you mention? It should!

One concern was to be sure on how to trigger those tests.
So it's by running `pytest` in the proper folder and letting it discover the tests.
How will this behave if we have lots of folders here and there? Maybe this won't happen, or we may not let this happen?

I'm good and I think despite the other comments, we could start by merging the PR like this?

I think @sframba could bring great insights of what's really important too.

src/docs/doxygen/GeosxConfig.hpp
cssherman: @TotoGaz - I've chatted with @rrsettgast about this, and he is planning to look into it

src/pygeosx/unitTests/CMakeLists.txt
cssherman: @TotoGaz - I've removed the pip command here, so this is now up to the user

cssherman:
@TotoGaz @sframba - FYI, many of the changes that are in this PR are due to the yapf code formatter being applied.  In another PR, I'm working to add this to the CI tests so that these sorts of changes don't pile up in the future.

sframba:
> @TotoGaz - here are some other unit tests that I have in mind:
> 
> 1. problem.run and pygeosx return methods
> 2. wrapper access with the modifiable flags turned off
> 
> Any other obvious ones stand out? Do we need to test the behavior in parallel runs, or would it be fine to add them to pygeosx_tools (which has some methods for mpi calls, etc.)?



> I actually don't use `pygeosx` enough to be helpful here. I would say that testing the parallel behavior for tiny runs makes sense. Where does it make most sense? Is the proposed approach compatible with what you mention? It should!
> 
> One concern was to be sure on how to trigger those tests. So it's by running `pytest` in the proper folder and letting it discover the tests. How will this behave if we have lots of folders here and there? Maybe this won't happen, or we may not let this happen?
> 
> I'm good and I think despite the other comments, we could start by merging the PR like this?
> 
> I think @sframba could bring great insights of what's really important too.

Thanks @cssherman, this is really useful for us! And thanks @TotoGaz for the help and feedback. 
LGTM, aside from run/initialization and wrapper access, we also explicitly call `cleanup` (to finalize writes), `execute`(to run an explicit solver step) and `reinit` to re-run a simulation without having to reinitialize GEOSX. They are all in `physicsSolvers/python/PySolver.cpp`. Also, note that we wrappers to set some fields, (e.g. to update the velocity model during an inversion process), so it could be useful to test also wrappers in write mode.

Would you like us to add a test file in the `pygeosx/unitTests` folder? It could be called `solverTests`, next to your `dataRepositoryTests`.

Also, I am looping in @paloma-martinez , she might have some additional insight.

src/pygeosx/unitTests/dataRepositoryTests/test_pygeosxDataRepository.py
```diff
@@ -0,0 +1,181 @@
...
+        Note: this error does not get called on certain scalar wrappers (time, cycle)
+
+        Args:
+            target_key (str): Key value for the target wrapper
+            geosx_proplem (pygeosx.problem): The problem handle
+        """
+        # yapf: enable
+        x = geosx_problem.get_wrapper(target_key).value()
+        if hasattr(x, "to_numpy"):
+            print('Test_a')
```
sframba, line 140: Is this print a debug leftover?

src/pygeosx/unitTests/dataRepositoryTests/test_pygeosxDataRepository.py
```diff
@@ -0,0 +1,181 @@
...
+    def test_problem_complete(self, geosx_problem):
+        """
+        Test running the problem to completion
+        """
+        max_iter = 100
+        ii = 0
+        while (pygeosx.run() != pygeosx.COMPLETED) and (ii <= max_iter):
+            ii += 1
+
+        assert ii != max_iter
```
sframba, line 172: would it be possible to check that it actually takes exactly `10` iterations (deduced from the xml file, since there should be no other events happening)?

src/pygeosx/unitTests/dataRepositoryTests/test_pygeosxDataRepository.py
```diff
@@ -0,0 +1,181 @@
...
+        x += 1
+        del x
+
+        # Get another copy of the wrapper
+        x = geosx_problem.get_wrapper(target_key).value()
+        if hasattr(x, "to_numpy"):
+            x = x.to_numpy()
+
+        # Check to see if the value changed appropriately
+        dx = x_init + 1 - x
```
sframba, line 116: I think it would be safer to use the absolute value of the difference, in case the discrepancy goes in the other direction

src/pygeosx/unitTests/dataRepositoryTests/test_pygeosxDataRepository.py
```diff
@@ -0,0 +1,181 @@
...
+        # yapf: enable
+        x = geosx_problem.get_wrapper(target_key).value()
+        if hasattr(x, "to_numpy"):
+            print('Test_a')
+            x = x.to_numpy()
+
+        with pytest.raises(ValueError):
+            x += 1
+
+    def test_problem_run(self, geosx_problem):
```
sframba, line 146: This can be triggered also by the getting access to the solver and calling `solver.execute`, this other approach could be tested in a new pytest file (`test_pygeosxSolver.py`?), cf. comment in the PR.

sframba (approved):

src/pygeosx/unitTests/dataRepositoryTests/test_pygeosxDataRepository.py
cssherman: Missed this one.   Yes, at a minimum it should be abs(sum(values))

PR #2149, opened 2022-11-01
Allow choice of finite element types in Finite Element Dispatch

sframba:
This PR is a partial answer to Issue [2145](https://github.com/GEOSX/GEOSX/issues/2145).

As of now, solvers calling the `finiteElement::dispatch3D` method have no way of limiting the number of template instantiations of the lambda function given as a parameter to the function. This can create useless specializations, causing large compilation units and a bloated compilation process, as well as increasing the size of the `geosx` executable. With the advent of higher order elements, this risks becoming an issue, as the number of dispatched types increases and differences between formulations start to appear (e.g., Gauss hexahedra vs. Gauss-Lobatto hexahedra).

This PR replaces the finite element dispatcher (defined in `FiniteElementDispatch.hpp`) with a recursive version based on a variadic template, in the vein of `ConstitutivePassThruHandler`. This allows each solver that calls the `finiteElement::dispatch3D` method to specify the types of finite elements on which the dispatch must happen. 

Notice that in this PR, very little actual optimization is done: all the current dispatch calls are still done with all the available order 1 finite element types (using an alias `ALL_FE_TYPES` that contains the whole list). Only the SEM wave propagators are specialized to Gauss-Lobatto hexahedra.

What is left to be done:
 - Further specialize the types of finite elements in the dispatch to the actual types that are handled by the solver, as needed;
 - Use a mechanism similar to that of PR [2086](https://github.com/GEOSX/GEOSX/pull/2086) to create separate compilation units for each dispatch. This might require writing the lambda function passed to the dispatch in a separate compilation unit, and thus is best left for another time.

### DIFF
--- a/src/coreComponents/finiteElement/FiniteElementDispatch.hpp
+++ b/src/coreComponents/finiteElement/FiniteElementDispatch.hpp
@@ -30,178 +30,127 @@
 #include "elementFormulations/Q3_Hexahedron_Lagrange_GaussLobatto.hpp"
 #include "LvArray/src/system.hpp"
 
-
+#define BASE_FE_TYPES finiteElement::H1_Hexahedron_Lagrange1_GaussLegendre2, \
+  finiteElement::H1_Wedge_Lagrange1_Gauss6, \
+  finiteElement::H1_Tetrahedron_Lagrange1_Gauss1, \
+  finiteElement::H1_Pyramid_Lagrange1_Gauss5, \
+  finiteElement::Q3_Hexahedron_Lagrange_GaussLobatto
+#ifdef GEOSX_DISPATCH_VEM
+#define VEM_TYPES finiteElement::H1_Tetrahedron_VEM_Gauss1, \
+  finiteElement::H1_Wedge_VEM_Gauss1, \
+  finiteElement::H1_Hexahedron_VEM_Gauss1, \
+  finiteElement::H1_Prism5_VEM_Gauss1, \
+  finiteElement::H1_Prism6_VEM_Gauss1, \
+  finiteElement::H1_Prism7_VEM_Gauss1, \
+  finiteElement::H1_Prism8_VEM_Gauss1, \
+  finiteElement::H1_Prism9_VEM_Gauss1, \
+  finiteElement::H1_Prism10_VEM_Gauss1, \
+  finiteElement::H1_Prism11_VEM_Gauss1
+#define ALL_FE_TYPES BASE_FE_TYPES, VEM_TYPES
+#else
+#define ALL_FE_TYPES BASE_FE_TYPES
+#endif
 
 namespace geosx
 {
 namespace finiteElement
 {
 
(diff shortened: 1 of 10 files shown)


klevzoff (approved):
I think this PR is good to be merged as is, given the immediate problem it is designed to solve.

At some point in future, I would like to change the dispatch mechanism to use tables (similar to `common/TypeDispatch.hpp`) instead of recursive variadic templates, and maybe do the same for constitutive models and unify everything under one implementation. Recursive handling is only useful when downcasting to some intermediate class in a tree-like hierarchy is required instead of an exact match. In other cases it is needlessly slowing down both runtime and compile-time (although probably by an epsilon for both in our case).

What that means is that we will be specifying dispatch lists via some
```c++
using BaseFETypes = TypeList<
  finiteElement::H1_Hexahedron_Lagrange1_GaussLegendre2,
  finiteElement::Q3_Hexahedron_Lagrange_GaussLobatto,
  finiteElement::H1_Wedge_Lagrange1_Gauss6,
  finiteElement::H1_Tetrahedron_Lagrange1_Gauss1,
  finiteElement::H1_Pyramid_Lagrange1_Gauss5
>
```
instead of a macro. How exactly it ties into CMake-based generation of explicit instantiations remains to be seen :) Again, this has no bearing on current PR.

sframba:
Thanks for the quick review @klevzoff. I agree that an approach based on table and type-to-index map would be more straightforward. In the case of this PR, I didn't want to change the behavior of the function, so I left the `dynamic_cast` for now.

As for the CMake-generated instantiations, it could simply be the "responsibility" of the calling solvers. For example one could replace the lambda passed to dispatch with a proper function, defined in a separate compilation unit (maybe the corresponding solver kernel?). Then one generates type-specific template instantiations of the kernel via the usual CMake configure mechanism. This would need to be done only for kernels which are already "big" and use a large number of `FE_TYPE`s.

src/coreComponents/finiteElement/FiniteElementDispatch.hpp
```diff
@@ -23,185 +23,140 @@
...
-    lambda( *ptr13 );
-  }
-  else if( auto * const ptr14 = dynamic_cast< H1_Prism10_VEM_Gauss1 * >(&input) ) // VEM on Prism10
-  {
-    lambda( *ptr14 );
-  }
-  else if( auto * const ptr15 = dynamic_cast< H1_Prism11_VEM_Gauss1 * >(&input) ) // VEM on Prism11
-  {
-    lambda( *ptr15 );
+    GEOSX_UNUSED_VAR( lambda );
```
XL64, line 92: You can remove this line and use LAMBDA && GEOSX_UNUSED_PARAM( lambda )

XL64 (approved):

sframba:
All integrated tests passing on Quartz.

rrsettgast (approved):

PR #2155, opened 2022-11-10
Build travis-ci using ninja

TotoGaz:
Build GEOSX with [ninja](https://ninja-build.org/) instead of make.

Related to https://github.com/GEOSX/thirdPartyLibs/pull/208 which installs `ninja` in the docker containers.


TotoGaz:
This build fails for `CUDA` builds with (I *think*) `clang` builds with errors like
```
CMake Error at cmake/blt/cmake/BLTPrivateMacros.cmake:435 (target_link_libraries):
  Error evaluating generator expression:
    $<AND:,$<COMPILE_LANGUAGE:CXX>>
  Parameters to $<AND> must resolve to either '0' or '1'.
  cmake/blt/cmake/BLTMacros.cmake:788 (blt_setup_target)
  coreComponents/LvArray/benchmarks/CMakeLists.txt:33 (blt_add_executable)
```

bmhan12 (approved):
Looks good to me!

PR #2156, opened 2022-11-10
Feature/enable hypre gpu on pangea3

XL64:
Enable HYPRE GPU on P3 and also remove warning on non reachable statement whitin MPI wrapper.

### DIFF
--- a/src/coreComponents/common/MpiWrapper.cpp
+++ b/src/coreComponents/common/MpiWrapper.cpp
@@ -156,45 +156,50 @@ int MpiWrapper::test( MPI_Request * request, int * flag, MPI_Status * status )
 {
 #ifdef GEOSX_USE_MPI
   return MPI_Test( request, flag, status );
-#endif
+#else
   *flag = 0;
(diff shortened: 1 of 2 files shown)


src/coreComponents/common/MpiWrapper.cpp
```diff
@@ -156,45 +156,50 @@ int MpiWrapper::test( MPI_Request * request, int * flag, MPI_Status * status )
 {
 #ifdef GEOSX_USE_MPI
   return MPI_Test( request, flag, status );
-#endif
+#else
```
TotoGaz, line 159: Is it mandatory to get this alongside the Pangea3 option update?

TotoGaz (commented):
I would split this PR into 2.

src/coreComponents/common/MpiWrapper.cpp
XL64: I'll separate.

TotoGaz (approved):

PR #2157, opened 2022-11-11
Build should not stop if FESAPI is not defined.

TotoGaz:


castelletto1 (approved):

rrsettgast:
Nobody rebuilt the TPL's on LC before merging the PR. Is there a case where FESAPI wouldn't be defined?

TotoGaz:
> Nobody rebuilt the TPL's on LC before merging the PR. Is there a case where FESAPI wouldn't be defined?

Yes, @YohannDudouit did try to build yesterday and we noticed the script mistake.
This prevents anyone from building GEOSX in the LLNL env.
Actually the nightly tests broke.

I plan to bypass the merge line to get this patch integrated. It looks critical to me. Are you guys OK?

rrsettgast:
What is the role of FESAPI now? Is it actually required?

TotoGaz:
FESAPI should be completely optional. It's clearly a mistake.
@YohannDudouit do you confirm that this patch leads you to full successful compilation in LLNL?

untereiner:
My mistake. It must be optional. Would it be possible to have a ci build with all tpl disabled to test these cases ?

TotoGaz:
> My mistake. It must be optional. Would it be possible to have a ci build with all tpl disabled to test these cases ?

Everything is always possible, but it requires resources (human + machines) to test all those combinations 🤷

rrsettgast:
@untereiner Sorry. Can you explain what we are using FESAPI to do exactly?

TotoGaz:
It's to plug RESQML into GEOSX

rrsettgast:
@TotoGaz Yes, I understand it is to plug RESQML into GEOSX. That wasn't my question.

TotoGaz:
Then I don't understand the question. Populating meshes from RESQML may be a more precise answer?

untereiner:
> Then I don't understand the question. Populating meshes from RESQML may be a more precise answer?

Exactly. It gives access to RESQML data stored in epc/h5 files.

PR #2161, opened 2022-11-14
Added face boundary condition to single-phase hybrid FVM

francoishamon:
This PR adds the face boundary condition to single-phase hybrid FVM, which we wanted to try to verify the Mandel problem with VEM on a PEBI mesh.

Ready for review.
Requires rebaseline, because an integrated test now has face boundary condition.
Rebaseline done in https://github.com/GEOSX/integratedTests/pull/277


jafranc (approved):
Great ! Thanks

PR #2170, opened 2022-11-22
Feature: SinglePhasePoromechanics with conforming fractures using lagrange multiplier based contact.

CusiniM:
In this PR we implement a singlephase poromechanics solver with conforming faults and contact enforced through Lagrange multipliers.

I have done some simple validations and it seems to be correct. I think it is ready to be reviewed, at least. I think we should get this into develop quickly and then start validating the solver thoroughly with analytical solutions and comparisons against the embedded one. 

I will also create and issue with a list of refactoring needed to improve some aspects of this solver (e.g., do contact constraints on device).

@andrea-franceschini I started from your branch but decided to start with a fresh one and port the functions that I needed.

rebaseline: https://github.com/GEOSX/integratedTests/pull/288


src/coreComponents/physicsSolvers/contact/ContactFields.hpp
```diff
@@ -108,15 +108,15 @@ DECLARE_FIELD( dTraction_dPressure,
 DECLARE_FIELD( fractureState,
                "fractureState",
                array1d< integer >,
-               FractureState::Stick,
+               FractureState::Open,
```
andrea-franceschini, line 111: Why do you need to change the default initial status?

andrea-franceschini:
> @andrea-franceschini I started from your branch but decided to start with a fresh one and port the functions that I needed.

Good choice! The original branch was very old and not that easy to use.

src/coreComponents/physicsSolvers/contact/ContactFields.hpp
CusiniM: This is actually something I wanted to discuss with you.

For the following problem (see image), the `LagrangianContact` solver (no poromechanics), starting with a stick condition as an initial guess the solution at the first iteration has some `NaN` which makes me think that the system is singular. Any idea why that may be the case? Maybe, one thing to do, is to not make everything crash if a `NaN` shows up but rather try a different configuration but I would like to understand if it's expected first.

simpleSetup

src/coreComponents/physicsSolvers/contact/LagrangianContactSolver.cpp
```diff
@@ -217,7 +217,10 @@ void LagrangianContactSolver::implicitStepComplete( real64 const & time_n,
                                                     real64 const & dt,
                                                     DomainPartition & domain )
 {
-  m_solidSolver->implicitStepComplete( time_n, dt, domain );
+  if( m_setupSolidSolverDofs )
+  {
+    m_solidSolver->implicitStepComplete( time_n, dt, domain );
+  }
```
francoishamon, line 223: After this PR is merged, we could explore ways to avoid the need for this flag. The need for this flag may indicate the current `LagrangianContactSolver` does too much work and should be broken down into two solvers (one coupled solver and one "single-physics" solver). Disclaimer: I don't know the existing `LagrangianContactSolver` very well, so I may be missing something.

**LagrangianContactSolver as a coupled solver**
The coupled solver would be something like that:
```
class LagrangianContactSolver : public CoupledSolver< SolidMechanicsLagrangianFEM, ContactSolver >
```
and would just be a driver for `setupDofs`, `implicitStepSetup`, `implicitStepComplete`, `applySystemSolution`, `applyBoundaryConditions` (functions where the current solver needs access to `m_solidSolver`). It would not contain any assembly functions.

**New single-physics solver**
The single-physics solver `ContactSolver` (or a better name) would have the assembly functions `assembleForceResidualDerivativeWrtTraction`, `assembleTractionResidualDerivativeWrtDisplacementAndTraction`, and `assembleStabilization` because these functions do not require access to the pointer to `m_solidSolver`. This single-physics solver would not make sense on its own, but to me this is ok.

**SinglePhasePoromechanicsConformingFractures**
Finally, in the new coupled solver added in this PR, we would have:
```
class SinglePhasePoromechanicsConformingFractures : public CoupledSolver< SinglePhasePoromechanicsSolver, ContactSolver >
```
and we would have the guarantee that `SolidMechanicsLagrangianFEM` can only be accessed by doing `poromechanicsSolver()->solidMechanicsSolver()`. So, with this approach, 1) we would not need `m_setupSolidSolverDofs` any more, and 2) we could still easily reuse the contact solver assembly functions.

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoromechanicsEmbeddedFractures.cpp
```diff
@@ -33,8 +33,8 @@ namespace geosx
 using namespace dataRepository;
 using namespace constitutive;
 
-SinglePhasePoromechanicsSolverEmbeddedFractures::SinglePhasePoromechanicsSolverEmbeddedFractures( const std::string & name,
-                                                                                                  Group * const parent ):
+SinglePhasePoromechanicsEmbeddedFractures::SinglePhasePoromechanicsEmbeddedFractures( const std::string & name,
+                                                                                      Group * const parent ):
```
francoishamon, line 37: Great that you removed `Solver` in the class name, we should keep propagating that to other coupled solvers.

src/coreComponents/finiteVolume/FluxApproximationBase.cpp
```diff
@@ -71,19 +71,26 @@ void FluxApproximationBase::initializePreSubGroups()
...
-
-        Group & stencilParentGroup = mesh.registerGroup( groupKeyStruct::stencilMeshGroupString() );
-        Group & stencilGroup = stencilParentGroup.registerGroup( getName() );
+        Group * stencilParentGroup = mesh.getGroupPointer( groupKeyStruct::stencilMeshGroupString() );
+        if ( stencilParentGroup == nullptr )
+        {
+          stencilParentGroup = &(mesh.registerGroup( groupKeyStruct::stencilMeshGroupString() ));
+        }
+        
+        Group & stencilGroup = stencilParentGroup->registerGroup( getName() );
```
francoishamon, line 80: If I understand correctly based on what we discussed, this is needed because the simulation will have two stencils objects (stabilization and matrix), is that right? Maybe add a comment here to remind us that this situation can happen, because I had completely forgotten that it was possible.

src/coreComponents/physicsSolvers/SolverBase.cpp
```diff
@@ -785,7 +785,7 @@ bool SolverBase::solveNonlinearSystem( real64 const & time_n,
                       m_localMatrix.toViewConstSizes(),
                       localRhs );
 
-//      LvArray::print<serialPolicy>( m_localMatrix.toViewConst() );
+    //  LvArray::print<serialPolicy>( m_localMatrix.toViewConst() );
```
francoishamon, line 788:
```suggestion
```

francoishamon (approved):
Looks good to me!

src/coreComponents/physicsSolvers/contact/LagrangianContactSolver.cpp
CusiniM: yes, I think this is doable. The only thing I am not too sure how to do is how to have a contact solver that is meaningful by itself. Otherwise we will just need a way to make sure a user cannot try to run it without a `SolidSolver`.

src/coreComponents/physicsSolvers/contact/ContactFields.hpp
andrea-franceschini: In my experience, `NaN` values occur when a part of the body becomes unconstrained during the simulation. For example, if the right side was constrained only by the fracture and this opens, then it is free to move (at least in some directions) and `NaN` appears. However, when the contact surface is in stick mode there is no way to justify the `NaN` other than as an error.

Would you mind sharing the input files used to run the problematic simulation?

CusiniM: @andrea-franceschini, have you had a chance to test this?

andrea-franceschini: Yes! Sorry @CusiniM for the delay.

The problem is the test case: it is too small. All nodes on the fracture are constrained and there is only one fracture element. This means that:
1) the fracture cannot _communicate_ with the surrounding 3D body
2) the stabilization cannot be built (it is a jump penalization, but since there is only one fracture element, there are no jumps)

If you refine the mesh a bit, increasing the number of elements from 2x1x1 to 4x2x2, the test case runs fine.

This is the new input, working with `fractureState` initialized as `Stick`. (It is a txt just because github complains about xml).
[conformingFracture_2d_contactOnly.txt](https://github.com/GEOSX/GEOSX/files/10225414/conformingFracture_2d_contactOnly.txt)

CusiniM: Hmm, yeah I had noticed that but I was trying to set up the simplest possible case. 
- Maybe the issue comes from the fact that the fracture nodes are not allowed to move in the vertical direction so the "slip" is completely blocked?
- I don't see why the stabilization would matter. For a single element it should not be necessary IMO.

andrea-franceschini: You are right! The lack of stabilization is not the problem.
The key point is to choose a different pattern for the constraints. The fracture has to be connected to the 3D body.
If only the two faces normal to the x-direction are fully constrained, the example runs fine.

[conformingFracture_2d_contactOnly_constrained.txt](https://github.com/GEOSX/GEOSX/files/10235993/conformingFracture_2d_contactOnly_constrained.txt)

CusiniM: If that's the case I think this PR can be wrapped up and merged and then we can start some testing.

andrea-franceschini: I agree.
Thanks @CusiniM for tacking care of this PR!

PR #2179, opened 2022-11-23
Implemented a fully coupled single-phase THM solver

francoishamon:
This PR implements the kernels needed to have a fully coupled single-phase thermo-hydro-mechanics solver in GEOSX.

After this morning's discussion, I removed all the "callbacks" (i.e., no lambda). In this PR, there is a moderate effort to avoid code duplication, but I really don't mind duplicating everything if we decide this is the way to go. 

The constitutive kernels are not included in the PR yet because they will conflict with @frankfeifan's work.

Will require rebaseline.


francoishamon:
Closing in favor of #2195

PR #2175, opened 2022-11-22
Refactor/wave solver improvment

acitrain:
This PR is a refactor + cleanup for the second order wave equation solvers (acoustic and elastic)
It contains mainly:
- Removal of some commented code 
- Creation of WaveSolverUtils header for sources and seismo utilities

PR #2184, opened 2022-11-25
Elastic  wave solver optimization

acitrain:
In this PR  we apply the optimization done in the acoustic kernel to get a significant boost of performance in our second order elastic wave kernel.

 This implies the optimization of mass and damping matrix computation and a rewrite of the stiffness computation following the same process as in the acoustic  with the same gain a complexity coming from $p^9$ to $p^5$, with p the order of the polynomial basis.
 
 Rebaseline done in https://github.com/GEOSX/integratedTests/pull/285


sframba (approved):

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEM.cpp
```diff
@@ -93,13 +93,69 @@ ElasticWaveEquationSEM::ElasticWaveEquationSEM( const std::string & name,
...
+    setDescription( "Moment of the source: 6 real values describing a symmetric tensor in Voigt notation."
+                    "The default value is { 1, 1, 1, 0, 0, 0 } (diagonal moment, corresponding to a pure explosion)." );
 }
 
 ElasticWaveEquationSEM::~ElasticWaveEquationSEM()
 {
   // TODO Auto-generated destructor stub
 }
 
+localIndex ElasticWaveEquationSEM::getNumNodesPerElem()
```
francoishamon, line 116: Can you just move this one and `AcousticWaveEquationSEM::getNumNodesPerElem` to `WaveSolverBase` to avoid code duplication?

Also, I think the goal is to call this function in `postProcessInput` (as you do in `AcousticWaveEquationSEM`), but it seems the `numNodesPerElem` is still hard-coded to 8 there, unless I am missing something.

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEM.cpp
```diff
@@ -530,68 +601,74 @@ void ElasticWaveEquationSEM::initializePostInitialConditionsPreSubGroups()
     NodeManager & nodeManager = mesh.getNodeManager();
     FaceManager & faceManager = mesh.getFaceManager();
 
+    /// get the array of indicators: 1 if the face is on the boundary; 0 otherwise
+    arrayView1d< integer > const & facesDomainBoundaryIndicator = faceManager.getDomainBoundaryIndicator();
```
francoishamon, line 605:
```suggestion
    arrayView1d< integer  const > const & facesDomainBoundaryIndicator = faceManager.getDomainBoundaryIndicator();
```
(I think)

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEMKernel.hpp
```diff
@@ -419,70 +387,96 @@ struct MassAndDampingMatrixKernel
...
+   * @param[in] X coordinates of the nodes
+   * @param[in] facesToElems map from faces to elements
+   * @param[in] facesToNodes map from face to nodes
+   * @param[in] facesDomainBoundaryIndicator flag equal to 1 if the face is on the boundary, and to 0 otherwise
+   * @param[in] freeSurfaceFaceIndicator flag equal to 1 if the face is on the free surface, and to 0 otherwise
+   * @param[in] velocity cell-wise velocity
+   * @param[out] damping diagonal of the damping matrix
+   */
+  template< typename EXEC_POLICY, typename ATOMIC_POLICY >
+  //std::enable_if_t< geosx::is_sem_formulation< std::remove_cv_t< FE_TYPE_ > >::value, void >
```
francoishamon, line 425:
```suggestion
```

src/coreComponents/finiteElement/elementFormulations/Qk_Hexahedron_Lagrange_GaussLobatto.hpp
```diff
@@ -377,6 +403,39 @@ class Qk_Hexahedron_Lagrange_GaussLobatto final : public FiniteElementBase
...
+   * @param coords the coordinate of the point where the gradient will be applied
+   * @param func Callback function accepting parameters: the gradient vector and the Jacobian
+   */
+  template< typename FUNC >
+  GEOSX_HOST_DEVICE
+  static void
+  evaluateGradient( int q,
+                    real64 const (&X)[numNodes][3],
+                    real64 const (&coords)[3],
+                    FUNC && func );
```
francoishamon, line 435: Is `evaluateGradient` used / going to be used anywhere? I did not find it in the solver code (not asking you to remove it, just trying to understand)

src/coreComponents/finiteElement/elementFormulations/Qk_Hexahedron_Lagrange_GaussLobatto.hpp
```diff
@@ -377,6 +403,39 @@ class Qk_Hexahedron_Lagrange_GaussLobatto final : public FiniteElementBase
...
+   * @param q The quadrature point index
+   * @param X Array containing the coordinates of the support points.
+   * @param func Callback function accepting three parameters: i, j and R_ij
+   */
+  template< typename FUNC >
+  GEOSX_HOST_DEVICE
+  static void
+  computeFirstOrderStiffnessTerm( int q,
+                                  real64 const (&X)[numNodes][3],
+                                  FUNC && func );
```
francoishamon, line 419: The callback functions are always a little bit cryptic. Maybe you could a comment to explain what it can be used for, and you could try a more meaningful name than `func` (that is not always easy)

francoishamon (approved):
Looks good!

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEM.cpp
acitrain: You are right for both things:
- Indeed I forgot an hardcoded number, i'll fix it.
- The movind into WaveSilverBase of this function is the last part of PR #2175 which refactor several things in particular this function. I prefer to do it inside the refactor as here we only want to touch files related tio the elastic stuff.

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEM.cpp
acitrain: Agreed

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEMKernel.hpp
acitrain: Miss that one thanks!

src/coreComponents/finiteElement/elementFormulations/Qk_Hexahedron_Lagrange_GaussLobatto.hpp
acitrain: Rha I forgot that we change the way to compute the sources and so we do not use this function anymore. Thanks I'll remove it.

src/coreComponents/finiteElement/elementFormulations/Qk_Hexahedron_Lagrange_GaussLobatto.hpp
acitrain: I'll try to update the comment. As you said it is often not easy to describe a callback...

PR #2185, opened 2022-11-28
Added MGR recipe for fully coupled single-phase THM solver

francoishamon:
This PR adds an MGR recipe for the new fully coupled single-phase THM solver implemented in https://github.com/GEOSX/GEOSX/pull/2179.

In this solver, we have 5 dofs (3 displacements dofs per node, 2 dof per elem: pressure and temperature).
In the new MGR recipe:
- We eliminate the displacements (F points)
- We solve the C-point system using Boomer AMG with `numFunctions = 2`


francoishamon:
Closing in favor of #2197

PR #2190, opened 2022-12-01
Change default linear solver parameters

victorapm:
This PR changes the default linear solver parameters:

- Number of relaxation sweeps changed from 2 to 1. New default generally leads to faster execution times.
- On the host, relaxation scheme changed from `fgs` to `l1sgs`.
- On the device, relaxation scheme changed from `fgs` to `l1jacobi`.
- On the device, coarsening algorithm changed from `HMIS` to `PMIS`.

@TotoGaz @XL64, this solves the performance issues on GPUs that we discussed recently. Another solution is to change the relaxation method via XML in case you want to try before this PR gets merged.

integratedTests PR: https://github.com/GEOS-DEV/integratedTests/pull/328


src/coreComponents/linearAlgebra/utilities/LinearSolverParameters.hpp
```diff
@@ -186,21 +186,27 @@ struct LinearSolverParameters
...
-    integer aggresiveNumLevels = 0;                 ///< Number of levels for aggressive coarsening.
-    PreOrPost preOrPostSmoothing = PreOrPost::both; ///< Pre and/or post smoothing
-    real64 threshold = 0.0;                         ///< Threshold for "strong connections" (for classical and smoothed-aggregation AMG)
-    integer separateComponents = false;             ///< Apply a separate component filter before AMG construction
+#if defined(GEOSX_USE_HYPRE_CUDA) || defined(GEOSX_USE_HYPRE_HIP)
+    string coarseningType = "PMIS";                             ///< Coarsening algorithm
+    SmootherType smootherType = SmootherType::l1jacobi;         ///< Smoother type
+#else
+    string coarseningType = "HMIS";                             ///< Coarsening algorithm
+    SmootherType smootherType = SmootherType::l1sgs;            ///< Smoother type
```
castelletto1, line 194: This options make sense if GEOS uses hypre as linear algebra interface. What happens if Trilinos or PETSc are chosen? I assume such options would have no effect with Trilinos and PETSc relying on the respective default parameters.

castelletto1 (approved):
Looks good to me, thanks!

src/coreComponents/linearAlgebra/utilities/LinearSolverParameters.hpp
victorapm: Good point, I didn't check that. Maybe @klevzoff can confirm?

rrsettgast (approved):

src/coreComponents/linearAlgebra/utilities/LinearSolverParameters.hpp
klevzoff: PETSc currently does not use any AMG parameters from input. Trilinos will use regular SGS/Jacobi instead of l1.

PR #2192, opened 2022-12-02
Small improvements for WellControls

### LINKED ISSUE #2189: [Bug] Error when producer and injector assigned the same rate table
**Description**

When two wells, one a producer and the other an injector are assigned the same table function as a rate function, GEOSX issues an error about a negative rate.

[voidage.txt](https://github.com/GEOSX/GEOSX/files/10127087/voidage.txt)

Attached is a simple reproducer `voidage.xml` (renamed to `voidage.txt`). This is a single phase flow only model in which there are two wells. The use case is to do "voidage replacement" - in this case, having the injector inject at the same rate as the producer. The two wells are therefore assigned the same `targetTotalRateTableName` in their `WellControls` elements.

**To Reproduce**

1. Run attached reproducer `voidage.xml`. 
2. GEOSX will issue the error
```
***** LOCATION: /tmp/GEOSX/src/coreComponents/physicsSolvers/fluidFlow/wells/SinglePhaseWell.cpp:141
***** Controlling expression (should be false): ( ( wellControls.isInjector() && targetTotalRate < 0.0 ) || ( wellControls.isProducer() && targetTotalRate > 0.0) )
***** Rank 0: WellControls named INJE: Target total rate cannot be negative

** StackTrace of 8 frames **
Frame 0: geosx::SinglePhaseWell::initializePostSubGroups()
[voidage.txt](https://github.com/GEOSX/GEOSX/files/10127078/voidage.txt)

Frame 1: geosx::dataRepository::Group::initialize()
Frame 2: geosx::dataRepository::Group::initialize()
Frame 3: geosx::dataRepository::Group::initialize()
Frame 4: geosx::ProblemManager::problemSetup()
Frame 5: geosx::GeosxState::initializeDataRepository()
Frame 6: main
Frame 7: __libc_start_main
Frame 8: /xxx/GEOSX-42f7895/bin/geosx
=====
```
This is despite the function being positive.

**Expected behavior**

Should run (?)

**Platform:**
- Machine: RHEL 7.6
- Compiler: gcc 8.3.0
- GEOSX Version 0.2.0 (HEAD, sha1: 42f7895)

**Additional context**

The workaround is trivial - redifine the function with a different name and use the different functions for producers and injectors. This makes the issue trivial but it would be nice to make it work either way.

dkachuma:
This PR makes small adjustments to ease the use of `WellControls`
- [x] Add a multiplier for well rate functions
  * Adds a multiplier `m_rateSign` used to determine the direction of flow: +1 for injectors and -1 for producers.
  * Ensures that `WellControls` does not change the assigned rate function.
  * Resolves #2189
- [x] Allows unspecified `targetBHP` for rate controlled wells
  * For a rate controlled well, default values of BHP will be assigned: 1.01325e5 Pa (1 atm/1.01325 bar/14.6959 psi) for producers and 1.01325e8 Pa (1000atm/1013.25 bar/14695.9 psi) for injectors.
- [x] Introduces a well status function for wells
  * Allows the well to be opened and shut independently of the BHP and/or rate functions.
  * The well will still be shut if a zero rate is specified regardless of the control mode
  
Rebaseline done in https://github.com/GEOSX/integratedTests/pull/284


src/coreComponents/physicsSolvers/fluidFlow/wells/WellControls.hpp
```diff
@@ -366,14 +364,23 @@ class WellControls : public dataRepository::Group
...
 
   /// BHP table
-  TableFunction * m_targetBHPTable;
+  TableFunction const * m_targetBHPTable;
+
+  /// Default BHP for rate controlled injectors
+  static constexpr real64 maximumBHP = 1.01325e8;
+
+  /// Default BHP for rate controlled producers
+  static constexpr real64 minimumBHP = 1.01325e5;
```
dkachuma, line 383: @francoishamon These values are hardcoded here. Is there a more appropriate place/policy for constants especially since they're dimensional.

francoishamon: Currently we don't have a good place or policy to put the constants in the code, so what you have done is good. But maybe this is a good occasion to start something cleaner in the `wells` folder and start extracting the constants in a separate file. 

So, we could create a file `WellConstants.hpp` in `src/coreComponents/physicsSolvers/fluidFlow/wells` and put the `maximumBHP` and `minimumBHP` there with inline code documentation as you already did. What do you think?

Also, we can bring up this topic either in an issue or at the developer meeting, because there are currently quite a few hard-coded values buried inside the code.

francoishamon (approved):
Thanks @dkachuma, everything looks good! Ready to merged, we just have to do the rebaseline because of the new `statusTableName` keyword

PR #2194, opened 2022-12-05
Fully coupled THM 1: Refactored poromechanics kernels

francoishamon:
This PR refactors the `SinglePhasePoromechanicsKernel` and `MultiphasePoromechanicsKernel` to make it easier to extend them to the thermal-hydro-mechanical case. 

The main changes are:
- The `quadraturePointKernel` function broken down into utility functions that will be called by derived classes
- Accumulation term et body force now computed directly in the physics solver kernels and no longer in `PorousSolid` constitutive model

The PR passes all the integrated tests except `PoroElastic_staircase_co2` because it adds a pressure derivative that was missing [here](https://github.com/GEOSX/GEOSX/blob/5f81bd32e2eb6043d13e6005bf47d258402441a4/src/coreComponents/constitutive/solid/PorousSolid.hpp#L286), so a rebaseline will be necessary. Rebaseline done in https://github.com/GEOSX/integratedTests/pull/287


src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoromechanicsKernel.hpp
```diff
@@ -378,55 +376,147 @@ class SinglePhase :
...
       1.0,
       detJxW );
   }
 
-  /**
-   * @copydoc geosx::finiteElement::ImplicitKernelBase::complete
-   */
+  GEOSX_HOST_DEVICE
+  GEOSX_FORCE_INLINE
+  void quadraturePointKernel( localIndex const k,
```
CusiniM, line 445: I think this functions is the same in `src/coreComponents/physicsSolvers/multiphysics/MultiphasePoromechanicsKernel.hpp` so maybe it can be moved to the base class ?

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoromechanicsKernel.hpp
```diff
@@ -135,189 +128,194 @@ class SinglePhase :
...
+   * @param[in] q the quadrature point index
+   * @param[in] strainIncrement the strain increment used in total stress and porosity computation
+   * @param[inout] stack the stack variables
    */
   GEOSX_HOST_DEVICE
   GEOSX_FORCE_INLINE
-  void setup( localIndex const k,
-              StackVariables & stack ) const
+  void
+  smallStrainUpdate( localIndex const k,
```
CusiniM, line 168: same thing for this function. I think it can be moved to the base class.

CusiniM (approved):
Looks great! I only have those comments about the possibility of moving a few more things to the base class. 

Also, I assume the refactoring of the porosity computation will be done elsewhere.

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoromechanicsKernel.hpp
francoishamon: Yes, for both, we can move them to the base class. There is one hurdle that I discovered while working on the PR: if we want the base class function `PoromechanicsBaseKernel::smallStrainUpdate` to call the derived class function `SinglePoromechanicsKernel::computeFluidIncrement`, the signature of the base class function has to look like:
```
  template< typename LEAF >
  GEOSX_HOST_DEVICE
  GEOSX_FORCE_INLINE
  void
  smallStrainUpdate( localIndex const k,
                     localIndex const q,
                     real64 const ( &strainIncrement )[6],
                     typename LEAF::StackVariables & stack ) const
  { ... }
```
where `LEAF` is `SinglePhasePoromechanicsKernel< SUBREGION_TYPE, CONSTITUTIVE_TYPE, FE_TYPE >` or  `MultiphasePoromechanicsKernel< SUBREGION_TYPE, CONSTITUTIVE_TYPE, FE_TYPE >`, so it scared me a little bit :). I don't know exactly what the implications of all these templates in such a simple function are on compilation time for instance (maybe none?). I will do some tests on Travis.

src/coreComponents/physicsSolvers/multiphysics/PoromechanicsKernelBase.hpp
```diff
@@ -0,0 +1,266 @@
...
+                           localIndex const targetRegionIndex,
+                           SUBREGION_TYPE const & elementSubRegion,
+                           FE_TYPE const & finiteElementSpace,
+                           CONSTITUTIVE_TYPE & inputConstitutiveType,
+                           arrayView1d< globalIndex const > const inputDispDofNumber,
+                           string const inputFlowDofKey,
+                           globalIndex const rankOffset,
+                           CRSMatrixView< real64, globalIndex const > const inputMatrix,
+                           arrayView1d< real64 > const inputRhs,
+                           real64 const (&inputGravityVector)[3],
```
TotoGaz, line 93:
```suggestion
                           real64 const (&gravityVector)[3],
```
Why the `input`?

src/coreComponents/physicsSolvers/multiphysics/PoromechanicsKernelBase.hpp
```diff
@@ -0,0 +1,266 @@
...
+
+    /// Linear momentum balance residual
+    real64 ( &localResidualMomentum )[numDispDofPerElem];
+    /// Derivative of linear momentum balance residual wrt displacement
+    real64 ( &dLocalResidualMomentum_dDisplacement )[numDispDofPerElem][numDispDofPerElem];
+    /// Derivative of linear momentum balance residual wrt pressure
+    real64 dLocalResidualMomentum_dPressure[numDispDofPerElem][1]{};
+
+    /// C-array storage for the element local row degrees of freedom.
+    globalIndex localPressureDofIndex[1]{};
```
TotoGaz, line 195: Same here for the `1`.

src/coreComponents/physicsSolvers/multiphysics/PoromechanicsKernelBase.hpp
```diff
@@ -0,0 +1,266 @@
...
+      {
+#if defined(CALC_FEM_SHAPE_IN_KERNEL)
+        stack.xLocal[a][i] = m_X[localNodeIndex][i];
+#endif
+        stack.u_local[a][i] = m_disp[localNodeIndex][i];
+        stack.uhat_local[a][i] = m_uhat[localNodeIndex][i];
+        stack.localRowDofIndex[a*3+i] = m_dofNumber[localNodeIndex]+i;
+        stack.localColDofIndex[a*3+i] = m_dofNumber[localNodeIndex]+i;
+      }
+    }
```
TotoGaz, line 227:
```suggestion
    constexpr integer something = 3;
    for( localIndex a = 0; a < numNodesPerElem; ++a )
    {
      localIndex const localNodeIndex = m_elemsToNodes( k, a );

      for( integer i = 0; i < something; ++i )
      {
#if defined(CALC_FEM_SHAPE_IN_KERNEL)
        stack.xLocal[a][i] = m_X[localNodeIndex][i];
#endif
        stack.u_local[a][i] = m_disp[localNodeIndex][i];
        stack.uhat_local[a][i] = m_uhat[localNodeIndex][i];
        stack.localRowDofIndex[a*something+i] = m_dofNumber[localNodeIndex]+i;
        stack.localColDofIndex[a*something+i] = m_dofNumber[localNodeIndex]+i;
      }
    }
```

src/coreComponents/physicsSolvers/multiphysics/PoromechanicsKernelBase.hpp
```diff
@@ -0,0 +1,266 @@
...
+
+  /// The rank-global displacement array.
+  arrayView2d< real64 const, nodes::TOTAL_DISPLACEMENT_USD > const m_disp;
+
+  /// The rank-global incremental displacement array.
+  arrayView2d< real64 const, nodes::INCR_DISPLACEMENT_USD > const m_uhat;
+
+  /// The gravity vector.
+  real64 const m_gravityVector[3]{};
+  real64 const m_gravityAcceleration;
```
TotoGaz, line 245: Can you explain the difference those two vectors?

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoromechanicsKernel.hpp
```diff
@@ -246,116 +235,363 @@ class Multiphase :
...
+   * @param[in] dPorosity_dTemperature the derivative of porosity wrt temperature
+   * @param[in] dSolidDensity_dPressure the derivative of solid density wrt pressure
+   * @param[inout] stack the stack variables
+   * @param[inout] bodyForceKernelOp the lambda function to customize the kernel
+   */
+  template< typename FUNC = NoOpFunc >
+  GEOSX_HOST_DEVICE
+  GEOSX_FORCE_INLINE
+  void
+  computeBodyForce( localIndex const k,
```
TotoGaz, line 338: What's exactly a `bodyForce`?

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoromechanicsKernel.hpp
TotoGaz: I guess it's the duplication question we spoke about last time?

TotoGaz (approved):

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoromechanicsKernel.hpp
francoishamon: Yes, I think we can get rid of this duplication as well, see [this comment](https://github.com/GEOSX/GEOSX/pull/2194#discussion_r1040016288). I will give it a try and we can see if we like it.

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoromechanicsKernel.hpp
```diff
@@ -246,116 +235,363 @@ class Multiphase :
...
+
+    real64 totalMassDensity = 0.0;
+    real64 dTotalMassDensity_dPressure = 0.0;
+    real64 dTotalMassDensity_dComponents[maxNumComponents]{};
+    real64 dPhaseMassDensity_dComponents[maxNumComponents]{};
+
+    for( integer ip = 0; ip < m_numPhases; ++ip )
+    {
+      totalMassDensity = totalMassDensity + phaseVolFrac( ip ) * phaseMassDensity( ip );
+      dTotalMassDensity_dPressure = dTotalMassDensity_dPressure
```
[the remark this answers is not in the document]
francoishamon: Yes, I had the same question. I think it matters to help the compiler use "fused multiply-add" on GPU platforms, but I may be very wrong.

@corbett5 @klevzoff @rrsettgast @YohannDudouit should we do:
```
x = x + y * z
```
or
```
x += y * z
```
? Does it matter?

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoromechanicsKernel.hpp
```diff
@@ -431,100 +667,158 @@ class Multiphase :
...
       1.0,
-      dPoreVolumeConstraint_dComponents,
+      stack.dPoreVolConstraint_dComponents,
       1.0,
       detJxW );
+  }
+
+  GEOSX_HOST_DEVICE
+  GEOSX_FORCE_INLINE
+  void quadraturePointKernel( localIndex const k,
```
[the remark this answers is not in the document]
francoishamon: This name has to be understood in the context of the base class `finiteElement::KernelBase` that has the generic `forAll` loop [here](https://github.com/GEOSX/GEOSX/blob/e9469bf176dbe673e0edd0cb6fbce37d5a4e512c/src/coreComponents/finiteElement/kernelInterface/KernelBase.hpp#L245). In this context, the name `quadraturePointKernel` makes sense since the kernel is called inside a loop over quadrature points. I think it is deliberately generic because it is called by many solvers for different physics (simple PDEs, solid mechanics, wave propagation) across the code.

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoromechanicsKernel.hpp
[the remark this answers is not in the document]
klevzoff: There probably were cases where some older version of NVCC were not issuing FMA instructions, but at least in CUDA 10.1.243 and newer it does not seem to differentiate between the two expressions: https://godbolt.org/z/qcq3c1sj9

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoromechanicsKernel.hpp
```diff
@@ -13,25 +13,36 @@
...
+ *   of lambda functions into kernel component functions.
+ * @struct NoOpFunc
+ */
+struct NoOpFunc
+{
+  template< typename ... Ts >
+  GEOSX_HOST_DEVICE
+  constexpr void
+  operator()( Ts && ... ) const {}
+};
```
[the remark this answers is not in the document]
francoishamon: Done, I put this in `codingUtilities/Utilities.hpp`.

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoromechanicsKernel.hpp
```diff
@@ -50,57 +61,60 @@ namespace poromechanicsKernels
...
+  using Base = PoromechanicsKernelBase< SUBREGION_TYPE,
+                                        CONSTITUTIVE_TYPE,
+                                        FE_TYPE >;
 
   /// Maximum number of nodes per element, which is equal to the maxNumTestSupportPointPerElem and
   /// maxNumTrialSupportPointPerElem by definition. When the FE_TYPE is not a Virtual Element, this
   /// will be the actual number of nodes per element.
   static constexpr int numNodesPerElem = Base::maxNumTestSupportPointsPerElem;
-  static constexpr int numMaxComponents = 3;
+  static constexpr int maxNumComponents = 3;
```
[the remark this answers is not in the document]
francoishamon: Here, the `3` of `maxNumComponents = 3` is not the same as the `3`s in 
```
public finiteElement::ImplicitKernelBase< SUBREGION_TYPE,
                                          CONSTITUTIVE_TYPE,
                                          FE_TYPE,
                                          3,
                                          3 >
```
which refer to the number of test/trial support points in the scheme. 

In the future, we will have to find a way to get rid of the `maxNumComponents`, but this is a little bit out of the scope of the PR.

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoromechanicsKernel.hpp
```diff
@@ -246,116 +235,363 @@ class Multiphase :
   void setup( localIndex const k,
               StackVariables & stack ) const
   {
-    for( localIndex a=0; a<numNodesPerElem; ++a )
+    // initialize displacement dof and pressure dofs
+    Base::setup( k, stack );
+
+    // setup component dofs
+    // Warning! This populates the array beyond numComponents (until maxNumComponents)
```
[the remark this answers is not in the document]
francoishamon: Done, I assigned the `-1` as you suggested

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoromechanicsKernel.hpp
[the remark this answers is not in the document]
francoishamon: Thanks for the info! I changed the code in the PR to use `+=`.

src/coreComponents/physicsSolvers/multiphysics/PoromechanicsKernelBase.hpp
francoishamon: Done, I removed `input`

src/coreComponents/physicsSolvers/multiphysics/PoromechanicsKernelBase.hpp
francoishamon: Done, this is now: 
```
/// C-array storage for the element local row degrees of freedom.
    globalIndex localPressureDofIndex{};
```

src/coreComponents/physicsSolvers/multiphysics/PoromechanicsKernelBase.hpp
francoishamon: Done

src/coreComponents/physicsSolvers/multiphysics/PoromechanicsKernelBase.hpp
francoishamon: Done, I added a comment to explain the difference between these two member variables.

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoromechanicsKernel.hpp
francoishamon: Done, I added a comment to explain what the body force is (standard name of `\rho g` in the linear momentum balance equation)

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoromechanicsKernel.hpp
```diff
@@ -246,116 +235,363 @@ class Multiphase :
...
+                                          + ( 1.0 - porosity ) * dSolidDensity_dPressure
+                                          + porosity * dTotalMassDensity_dPressure;
+    LvArray::tensorOps::scale< maxNumComponents >( dTotalMassDensity_dComponents, porosity );
+
+    // Step 3: finally, get the body force
+
+    LvArray::tensorOps::scaledCopy< 3 >( stack.bodyForce, m_gravityVector, mixtureDensity );
+    LvArray::tensorOps::scaledCopy< 3 >( stack.dBodyForce_dVolStrainIncrement, m_gravityVector, dMixtureDens_dVolStrainIncrement );
+    LvArray::tensorOps::scaledCopy< 3 >( stack.dBodyForce_dPressure, m_gravityVector, dMixtureDens_dPressure );
+    LvArray::tensorOps::Rij_eq_AiBj< 3, maxNumComponents >( stack.dBodyForce_dComponents, m_gravityVector, dTotalMassDensity_dComponents );
```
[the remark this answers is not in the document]
francoishamon: Yes, you are right. Sorry, this is confusing

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoromechanicsKernel.hpp
```diff
@@ -174,61 +170,54 @@ class Multiphase :
...
+    real64 dCompMassIncrement_dPressure[maxNumComponents]{};
+    /// Derivative of mass accumulation wrt comp density
+    real64 dCompMassIncrement_dComponents[maxNumComponents][maxNumComponents]{};
+
+    /// Pore volume constraint
+    real64 poreVolConstraint;
+    /// Derivative of pore volume constraint wrt pressure
+    real64 dPoreVolConstraint_dPressure;
+    /// Derivative of pore volume constraint wrt comp density
+    real64 dPoreVolConstraint_dComponents[1][maxNumComponents]{};
```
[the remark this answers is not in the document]
francoishamon: Here, I left the 2-dimensional array, otherwise it becomes very difficult to use the generic functions in `BilinearFormUtilities`. Maybe this is something that can be improved later for cell-centered P0 quantities coming from the flow equations.

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoromechanicsKernel.hpp
```diff
@@ -174,61 +170,54 @@ class Multiphase :
...
+
+    /// Derivative of momemtum balance residual wrt comp density
+    real64 dLocalResidualMomentum_dComponents[Base::StackVariables::numDispDofPerElem][maxNumComponents]{};
+
+    /// Mass balance residual
+    real64 localResidualMass[maxNumComponents]{};
+    /// Derivative of mass balance residual wrt volumetric strain increment
+    real64 dLocalResidualMass_dDisplacement[maxNumComponents][Base::StackVariables::numDispDofPerElem]{};
+    /// Derivative of mass balance residual wrt pressure
+    real64 dLocalResidualMass_dPressure[maxNumComponents][1]{};
```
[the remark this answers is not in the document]
francoishamon: For the same reasons as before (using `BilinearFormUtilities`), I left the 2-dimensional array here

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoromechanicsKernel.hpp
francoishamon: It is more difficult than I anticipated. If I move this function to the base kernel class, in the addition to the `LEAF` template mentioned above, I have to define dummy implementations of `computeFluidIncrement` and `computeBodyForce`, and it it really tricky which function is actually called in the code. So for now, `quadraturePointKernel` and `smallStrainUpdate` are duplicated. I think it is ok because these functions are very simple.

src/coreComponents/constitutive/solid/PorousSolid.hpp
```diff
@@ -461,16 +187,29 @@ class PorousSolidUpdates : public CoupledSolidUpdates< SOLID_TYPE, BiotPorosity,
...
     dTotalStress_dPressure[0] = -biotCoefficient;
     dTotalStress_dPressure[1] = -biotCoefficient;
     dTotalStress_dPressure[2] = -biotCoefficient;
     dTotalStress_dPressure[3] = 0;
     dTotalStress_dPressure[4] = 0;
     dTotalStress_dPressure[5] = 0;
+
+    dTotalStress_dTemperature[0] = -thermalExpansionCoefficientTimesBulkModulus;
+    dTotalStress_dTemperature[1] = -thermalExpansionCoefficientTimesBulkModulus;
+    dTotalStress_dTemperature[2] = -thermalExpansionCoefficientTimesBulkModulus;
```
sytuannguyen, line 208: factor 3 is missing

src/coreComponents/constitutive/solid/porosity/BiotPorosity.hpp
```diff
@@ -69,25 +69,32 @@ class BiotPorosityUpdates : public PorosityBaseUpdates
...
+                                               real64 const & deltaPressure,
+                                               real64 const & deltaTemperature,
+                                               real64 const (&strainIncrement)[6],
+                                               real64 const & thermalExpansionCoefficient,
+                                               real64 & dPorosity_dVolStrain,
+                                               real64 & dPorosity_dPressure,
+                                               real64 & dPorosity_dTemperature ) const
   {
     real64 const biotSkeletonModulusInverse = (m_biotCoefficient[k] - m_referencePorosity[k]) / m_grainBulkModulus;
+    real64 const porosityThermalExpansion = 3 * thermalExpansionCoefficient * m_biotCoefficient[k];
```
sytuannguyen, line 83: do you want to use directly m_thermalExpansionCoefficient here as you are using m_biotCoefficient, m_referencePorosity, etc.

src/coreComponents/constitutive/solid/porosity/BiotPorosity.hpp
```diff
@@ -69,25 +69,32 @@ class BiotPorosityUpdates : public PorosityBaseUpdates
...
 
-    real64 const porosity = m_porosity_n[k][q] +
-                            +m_biotCoefficient[k] * LvArray::tensorOps::symTrace< 3 >( strainIncrement ) + biotSkeletonModulusInverse * deltaPressure;
-
-    dPorosity_dPressure = biotSkeletonModulusInverse;
+    real64 const porosity = m_porosity_n[k][q]
+                            + m_biotCoefficient[k] * LvArray::tensorOps::symTrace< 3 >( strainIncrement )
+                            + biotSkeletonModulusInverse * deltaPressure
+                            + porosityThermalExpansion * deltaTemperature;
```
sytuannguyen, line 88: The sign of this term might be negative because in an isochore drained condition (zero strain and zero pressure change), an increase in temperature would cause a decrease in porosity?

src/coreComponents/constitutive/solid/porosity/BiotPorosity.hpp
```diff
@@ -69,25 +69,32 @@ class BiotPorosityUpdates : public PorosityBaseUpdates
...
 
     dPorosity_dVolStrain = m_biotCoefficient[k];
+    dPorosity_dPressure = biotSkeletonModulusInverse;
+    dPorosity_dTemperature = porosityThermalExpansion;
```
sytuannguyen, line 92: Same comment as above about the sign of this term

src/coreComponents/physicsSolvers/multiphysics/PoromechanicsKernelBase.hpp
```diff
@@ -0,0 +1,266 @@
...
+  /// The rank-global fluid pressure
+  arrayView1d< real64 const > const m_pressure;
+
+};
+
+} // namespace poromechanicsKernels
+
+} // namespace geosx
+
+#include "finiteElement/kernelInterface/SparsityKernelBase.hpp"
```
sytuannguyen, line 266: This reference is always included in the bottom, I did a try to move it up and everything still work. What is the special reason of putting this include in the bottom? (This comment is not particulaly related to this PR)

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoromechanicsKernel.hpp
CusiniM: I wish we could just use virtual methods....

src/coreComponents/constitutive/solid/porosity/BiotPorosity.hpp
jhuang2601: The sign convention, definition of `porosityThermalExpansion` and `thermalExpansionCoefficient` needs to be revisited. I would suggest to update the [rst file for Biot porosity](https://github.com/GEOSX/GEOSX/blob/develop/src/coreComponents/constitutive/docs/BiotPorosity.rst) and add adequate refence for the thermal part, which can help to verify the implementation.

francoishamon: Yes, thanks for catching the sign issue, it is fixed now. For now it does not matter because the thermal solver will be added in the next PR (here `deltaTemperature` is still always zero), but still. Also thanks for reminding me about the documentation, I will update it in #2195.

src/coreComponents/constitutive/solid/porosity/BiotPorosity.hpp
francoishamon: Yes, it is fixed now

src/coreComponents/physicsSolvers/multiphysics/PoromechanicsKernelBase.hpp
francoishamon: Yes, it is not necessary to include this everywhere, this was just a copy-paste, I removed it

jhuang2601 (approved):

src/coreComponents/constitutive/solid/porosity/BiotPorosity.hpp
francoishamon: Ideally, we would want the `thermalExpansionCoefficient` to live in one place (i.e., not duplicated), and I think it will make sense to have it in the solid model (see what I have done in #2195). That is why it is treated differently from the Biot coefficient and the reference porosity.

So in #2195, in the `PorousSolid` constitutive model, I get the `thermalExpansionCoefficient` from the solid model, and I pass it to the porosity model as an argument of the `updateFromPressureTemperatureAndStrain` function. I think it is cleaner like this.

sytuannguyen (approved):

PR #2202, opened 2022-12-06
Sequential coupling in CoupledSolver

CusiniM:
In this PR we:
- [x] add the sequential coupling to `CoupledSolver`. Branching between fully coupled (FIM) and sequentially coupled now occurs in `CoupledSolver`; 
- [x] turn the `PhaseFieldFracture` into a `CoupledSolver` that uses a sequential coupling;
- [x] properly modify `FlowProppantTransport`
- [x] removed redundant implementations of `solverStep` in coupled solvers.

addresses one of the points listed in https://github.com/GEOSX/GEOSX/issues/2176 

Rebaseline: https://github.com/GEOSX/integratedTests/pull/289


src/coreComponents/physicsSolvers/multiphysics/PhaseFieldFractureSolver.cpp
```diff
@@ -107,33 +86,6 @@ void PhaseFieldFractureSolver::implicitStepSetup( real64 const & GEOSX_UNUSED_PA
   } );
 }
```
frankfeifan, line 87: @andrembcosta just to confirm, do we still need these `TotalMeanStress` stuff in the phase-field solver? I think they are useless and we can remove them.

CusiniM: It does not seem to be used anywhere so I have just removed it.

src/coreComponents/physicsSolvers/multiphysics/CoupledSolver.hpp
```diff
@@ -42,6 +57,17 @@ class CoupledSolver : public SolverBase
...
   {
+    /// GEOS mainly uses FIM coupling so let's define FIM as the default.
+    registerWrapper( viewKeyStruct::couplingTypeString(), &m_couplingType ).
+      setInputFlag( dataRepository::InputFlags::OPTIONAL ).
+      setApplyDefaultValue( CouplingType::FIM ).
+      setDescription( "Type of coupling. Options are: Sequential and FIM" );
+
+    registerWrapper( viewKeyStruct::subcyclingOptionString(), &m_subcyclingOption ).
+      setInputFlag( dataRepository::InputFlags::OPTIONAL ).
+      setApplyDefaultValue( 0 ).
```
francoishamon, line 68: Maybe we should put the default option to `subcyclingOption = 1` because without subcycling, the results may be not converged and potentially inaccurate?

src/coreComponents/physicsSolvers/multiphysics/CoupledSolver.hpp
```diff
@@ -307,20 +332,194 @@ class CoupledSolver : public SolverBase
...
+                                                           cycleNumber,
+                                                           domain );
+
+        mapSolutionBetweenSolvers( domain, idx() );
+
+        if( dtReturnTemporary < dtReturn )
+        {
+          iter = 0;
+          dtReturn = dtReturnTemporary;
+        }
```
francoishamon, line 432: Later, we can add an option that will force the coupled solver to cut the time step if one of the solvers has a time step failure. Just to make sure that we have full control on the behavior of the outer iteration.

src/docs/doxygen/GeosxConfig.hpp
```diff
@@ -141,7 +141,7 @@
 #define fmt_VERSION 8.0.1
 
 /// Version information for python
-#define Python3_VERSION 3.8.5
+/* #undef Python3_VERSION */
```
francoishamon, line 144: I have the same modification in my PRs. I never know if we should commit this file or not?

src/coreComponents/physicsSolvers/multiphysics/PhaseFieldFractureSolver.cpp
```diff
@@ -36,333 +34,88 @@ using namespace constitutive;
...
-void PhaseFieldFractureSolver::initializePostInitialConditionsPreSubGroups()
-{}
-
 PhaseFieldFractureSolver::~PhaseFieldFractureSolver()
 {
   // TODO Auto-generated destructor stub
 }
 
-void PhaseFieldFractureSolver::resetStateToBeginningOfStep( DomainPartition & domain )
+void PhaseFieldFractureSolver::mapSolutionBetweenSolvers( DomainPartition & domain, integer const idx )
```
francoishamon, line 47:
```suggestion
void PhaseFieldFractureSolver::mapSolutionBetweenSolvers( DomainPartition & domain, integer const solverType )
```
to make it clear that the index that is passed is the solver type, even if we don't have the `enum` defined in the base `CoupledSolver` class

src/coreComponents/physicsSolvers/multiphysics/CoupledSolver.hpp
```diff
@@ -27,6 +27,21 @@
...
+enum class CouplingType : integer
+{
+  FIM,        ///< Fully-implicit coupling
+  Sequential  ///< Sequential coupling
+};
+
+ENUM_STRINGS( CouplingType,
+              "FIM",
+              "Sequential" );
+
```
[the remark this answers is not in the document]
francoishamon: I would vote for putting this and possibly the subcycling option in `NonlinearSolverParameters`, I think it is a good fit for anything that controls the nonlinear solver behavior and that is not physics dependent.

francoishamon (approved):
Looks great, thanks!

src/docs/doxygen/GeosxConfig.hpp
CusiniM: probably not but I always do `git add -u` so it added it automatically.

src/coreComponents/physicsSolvers/multiphysics/CoupledSolver.hpp
CusiniM: yeah, it's actually the only reason why I had set the default `subcycling` to 0 for now.

src/docs/doxygen/GeosxConfig.hpp
klevzoff: For any config macros that are platform or build type dependent, we typically "fix" them to particular values such that this header is never modified by the build:
https://github.com/GEOSX/GEOSX/blob/develop/src/cmake/GeosxConfig.cmake#L56-L64

We forgot to do this for all the newly added version variables. Most of our dependency version are fixed for most builds, but some, like Python, are system dependent.
 
You could either add `set(Python3_VERSION "3.8.5")` or `unset(Python3_VERSION)` as a workaround. In the first case, you should also commit this file (once and for all). Or there should be a separate PR to fix this for all variables.

src/coreComponents/physicsSolvers/multiphysics/PhaseFieldFractureSolver.cpp
andrembcosta: I have never used them.

src/coreComponents/physicsSolvers/multiphysics/CoupledSolver.hpp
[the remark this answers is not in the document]
CusiniM: done

src/coreComponents/physicsSolvers/multiphysics/CoupledSolver.hpp
CusiniM: The issue is that the `subcycling` in not very robust for now because if one of the solvers cuts the timestep things may go wrong. For now, the main sequential solver is phaseField for which we don't usually use subcycling. I think we can change it when we merge the sequential thermoporomechanics.

src/coreComponents/physicsSolvers/NonlinearSolverParameters.hpp
```diff
@@ -151,6 +154,15 @@ class NonlinearSolverParameters : public dataRepository::Group
     Parabolic, ///< use parabolic interpolation to define line search scaling factor.
   };
 
+  /**
+   * @brief Coupling type.
+   */
+  enum class CouplingType : integer
+  {
+    FIM,      ///< Fully-implicit coupling
```
joshua-white, line 162: Just a stylistic suggestion, but maybe write `FullyImplicit` for readability?

src/coreComponents/physicsSolvers/multiphysics/CoupledSolver.hpp
```diff
@@ -307,18 +306,181 @@ class CoupledSolver : public SolverBase
...
+                           solver->getDofManager(),
+                           solver->getLocalMatrix(),
+                           solver->getSystemRhs(),
+                           solver->getSystemSolution() );
+
+      solver->implicitStepSetup( time_n, dt, domain );
+
+    } );
+
+    // need to check what to do with this. this->implicitStepSetup( time_n, dt, domain );
```
joshua-white, line 374: is there a todo here, or can this be deleted?

joshua-white (approved):
Looks great to me.

src/coreComponents/physicsSolvers/NonlinearSolverParameters.hpp
CusiniM: sure, I don't see why not.

frankfeifan (approved):
It looks good to me.

Issue #2209, opened 2022-12-09
Improve separation between business and geometry in mesh generators


TotoGaz:
**Describe the issue**

Currently, the mesh generation tools depend on business domain, for example by looping over the constitutive relations of the sub regions when importing the fields. Fixing the blurred lines will help future developments and validations.

**Proposed cleanup**

In `MeshGeneratorBase` (and especially in its implementations):

For a first PR
- [ ] Split `importFields` into business and a geometry part.
- [ ] Challenge the `CellBlocABC::getExternalProperties` feature. Is it relevant w.r.t. `MeshGeneratorBase::importFields`?

As a second step, split `generateMesh` into business and geometry part.
  - [ ] Keeping the `CellBlockManager*` in the mesh/generator folder, and the `DomainPartition`, `MeshBody` in the business part.   The refactoring for `VTKMeshGenerator` and `InternalMeshGenerator` looks quite identical and within close reach.  
  - [ ] The work for `InternalWellGenerator` may be a tad more challenging because `WellElementRegion` are directly created in the `Generator`, without the help of any `CellBlock` or `FaceBlock` like we do for the other types of (sub) regions. To be studied with care.

On a more C++ technical point of view for a last PR (and during the continuous process).

- [ ] Be sure that all the remaining business includes (core + tests) are removed.
- [ ] Split the `generator` folder into its own `cmake` target, with the appropriate dependencies.
Can we only expose the `*ABC` classes and a few utility classes?
- [ ] Remove the `cmake` dependencies (core + tests) .
- [ ] See what we can do with the `public`/`private`/`interface` feature proposed by cmake.
- [ ] Search for the `Region` word in the `generators` folder to check if its the appropriate wording.
- [ ] Update the unit tests accordingly.

TotoGaz:
Particle manager (from PR https://github.com/GEOS-DEV/GEOS/pull/2767) puts back the dependency to `SpatialPartition`, `MeshBody`, `MeshLevel`, `ParticleRegion`...

@XL64 and myself had spent some time to remove this connection to the business domain, in order to let the module become a purely geometrical “module”.

My comment in the review https://github.com/GEOS-DEV/GEOS/pull/2767#discussion_r1367568662 asked to remove the link to business domain...

The PR was merged anyhow.

I have no more time to invest into any more cleaning.
Therefore I unassign myself from this issue; I will not be working on it anymore.

PR #2212, opened 2022-12-13
move non mesh parts of importFields to meshManager

XL64:
This change adresses some points of issue #2209


TotoGaz (commented):
First batch of comments.

src/coreComponents/mainInterface/ProblemManager.cpp
```diff
@@ -622,16 +625,76 @@ void ProblemManager::generateMesh()
...
 {
   GEOSX_MARK_FUNCTION;
   DomainPartition & domain = getDomainPartition();
   MeshManager & meshManager = this->getGroup< MeshManager >( groupKeys.meshManager );
+  ElementRegionManager & elemManager = domain.getMeshBody( this->getName() ).getBaseDiscretization().getElemManager();
 
   meshManager.forSubGroups< MeshGeneratorBase >( [&]( MeshGeneratorBase & generator )
   {
-    generator.importFields( domain );
+    FieldIdentifiers fieldsToBeSync;
```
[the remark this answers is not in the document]
XL64: ok

TotoGaz (commented):
Despite the place to attach the fields names, I think we're quite close from merging this PR!

src/coreComponents/mesh/MeshManager.cpp
```diff
@@ -75,13 +79,86 @@ void MeshManager::generateMeshLevels( DomainPartition & domain )
...
+  {
+    try
+    {
+      VTKMeshGenerator & g = dynamic_cast< VTKMeshGenerator & >( generator );
+      GEOSX_UNUSED_VAR( g );
+    }
+    catch( const std::bad_cast & e )
+    {
+      return;
+    }
```
TotoGaz, line 117: Why do you only work on `VTKMeshGenerator` and not on all `MeshGeneratorBase`s ?
I have a few comments about it.

- If we really wanted to do this, we would use some `forSubGroups< VTKMeshGenerator >` instead.
- Please use our `dynamicCast` (which may lead to a tiny differente behavior)
- But more importantly, by doing so, you expose the implementation `VTKMeshGenerator` which eventually should not be visible from outside of the `mesh` module.

I think that you could loop on all `MeshGeneratorBase` instances. Instances with no particular data to import will return empty dummy values (like `fieldNamesMapping` will be an empty `map`), and all the import loops will vanish. We'll remove a `if` in the code which is always nice. And the `resqml` guys won't have to modify anything from this code if they implement their own `MeshGeneratorBase`.

src/coreComponents/mesh/MeshManager.cpp
```diff
@@ -15,10 +15,14 @@
 
 #include "MeshManager.hpp"
 
+#include "mesh/mpiCommunications/CommunicationTools.hpp"
 #include "mesh/mpiCommunications/SpatialPartition.hpp"
 #include "generators/MeshGeneratorBase.hpp"
+#include "generators/VTKMeshGenerator.hpp"
```
TotoGaz, line 21: W.r.t. the other comment below, this can be removed.

TotoGaz (commented):
This `VTKMeshGenerator` stuff is important.

src/coreComponents/mesh/MeshManager.cpp
TotoGaz: Maybe implement an equivalent to `domain.hasMeshBody(...)`?

XL64:
@TotoGaz if you want to do a final review i think it's ready.

src/coreComponents/mesh/DomainPartition.hpp
```diff
@@ -174,6 +174,19 @@ class DomainPartition : public dataRepository::Group
...
+  /**
+   * @brief Check if a MeshBody is present given a name.
+   * @tparam KEY_TYPE The type of the key used to look up the MeshBody.
+   * @param key The key to the MeshBody.
+   * @return True is the MeshBody exists in the domain.
+   */
+  template< typename KEY_TYPE >
+// TODO uncomment these to disallow using a non string or char const * key
+//            std::enable_if_t< std::is_same< T, string >::value ||
+//                              std::is_same< T, const char * >::value, bool > = false >
```
TotoGaz, line 186: Please remove

src/coreComponents/mesh/MeshManager.cpp
```diff
@@ -15,10 +15,13 @@
 
 #include "MeshManager.hpp"
 
+#include "mesh/mpiCommunications/CommunicationTools.hpp"
 #include "mesh/mpiCommunications/SpatialPartition.hpp"
 #include "generators/MeshGeneratorBase.hpp"
 #include "common/TimingMacros.hpp"
 
+#include <unordered_set>
+#include <typeinfo>
```
TotoGaz, line 24: Can you double check that you need `#include <typeinfo>`?

src/coreComponents/mesh/MeshManager.cpp
```diff
@@ -75,13 +78,80 @@ void MeshManager::generateMeshLevels( DomainPartition & domain )
...
+                                              meshFieldName, geosxFieldName,
+                                              region.getName(), subRegion.getName() ) );
+
+        bool const isMaterialField = materialWrapperNames.count( geosxFieldName ) > 0 && wrapper.numArrayDims() > 1;
+        generator.importFieldsOnArray( region.getName(), meshFieldName, isMaterialField, wrapper );
+      }
+    } );
+
+    CommunicationTools::getInstance().synchronizeFields( fieldsToBeSync,
+                                                         domain.getMeshBody( generator.getName() ).getBaseDiscretization(),
```
TotoGaz, line 150: Maybe extract these calls as a separate variable, since it's already done line 114?

src/coreComponents/mesh/generators/InternalMeshGenerator.hpp
```diff
@@ -52,6 +52,21 @@ class InternalMeshGenerator : public MeshGeneratorBase
 
   virtual void generateMesh( DomainPartition & domain ) override;
 
+  /**
+   * @brief import fields from the mesh  on the array accessible via the given wrapper.
+   * @param cellBlockName name of the cell block to copy data from.
+   * @param meshFieldName name of the field in the meshd
+   * @param isMaterialField Indicate if we want to import material or regular fields
+   * @param wrapper Wrapper to access the array
+   */
```
TotoGaz, line 61:
```suggestion
```
I do not think you need to doxygen a derived function if it's the exact same doc.

src/coreComponents/mesh/generators/MeshGeneratorBase.hpp
```diff
@@ -78,13 +79,13 @@ class MeshGeneratorBase : public dataRepository::Group
...
+   * @param cellBlockName name of the cell block to copy data from.
+   * @param meshFieldName name of the field in the meshd
+   * @param isMaterialField Indicate if we want to import material or regular fields
+   * @param wrapper Wrapper to access the array
    */
-  virtual void importFields( DomainPartition & domain ) const
-  {
-    GEOSX_UNUSED_VAR( domain );
-  }
+  virtual void importFieldsOnArray( string const cellBlockName, string const meshFieldName, bool isMaterialField, dataRepository::WrapperBase & wrapper ) const = 0;
```
TotoGaz, line 88:
```suggestion
  virtual void importFieldsOnArray( string const & cellBlockName, string const & meshFieldName, bool isMaterialField, dataRepository::WrapperBase & wrapper ) const = 0;
```
It's not necessary to copy as const a string, since it'll be copied.
Even if at some place, we do define copied input as `const`, so we can't modify the input tmp param...

src/coreComponents/mesh/generators/VTKMeshGenerator.hpp
```diff
@@ -23,6 +23,7 @@
 #include "mesh/generators/VTKUtilities.hpp"
 
 #include <vtkDataSet.h>
+#include <string>
```
TotoGaz, line 26: We don't use `std::string`, we use `geosx::string` (as a matter of fact a `typedef` to `std::string`).
It's defined in `src/coreComponents/common/DataTypes.hpp`. (even if we still use some `std::string` or `std::to_string` here and there.)

src/coreComponents/mesh/generators/VTKMeshGenerator.hpp
```diff
@@ -87,7 +88,7 @@ class VTKMeshGenerator : public ExternalMeshGeneratorBase
    */
   virtual void generateMesh( DomainPartition & domain ) override;
 
-  virtual void importFields( DomainPartition & domain ) const override;
+  virtual void importFieldsOnArray( string const regionName, string const meshFieldNam, bool isMaterialField, WrapperBase & wrapper ) const override;
```
TotoGaz, line 91:
```suggestion
  void importFieldsOnArray( string const regionName, string const meshFieldNam, bool isMaterialField, WrapperBase & wrapper ) const override;
```

TotoGaz (commented):
I think we're really close from merging this.

TotoGaz (approved):

TotoGaz:
I've tested that this does not imply any rebaseline.
I'm merging this PR when it's my turn @XL64. Is it OK?

XL64:
> I've tested that this does not imply any rebaseline. I'm merging this PR when it's my turn @XL64. Is it OK?

Perfect

PR #2229, opened 2022-12-22
Changed abs to LvArray::math::abs

francoishamon:
This PR removes the remaining abs that are causing some compilation issues in the wave propagation solvers.

@cj215 @untereiner can you please try it and let me know if it compiles for you? I cannot check on my side


untereiner (approved):
ok for me [clang version 14.0.6]

acitrain (approved):

PR #2230, opened 2022-12-28
Fix for async wait bug #2173

### LINKED ISSUE #2173: Partitioning- and platform-dependent failures in multiple physics solvers
Users on several platforms have noticed the certain solvers will fail when a certain number of MPI ranks are requested, but not before.  The critical partitioning scenario seems to be platform dependent, and the problem has not year been reproduced on LLNL systems.  We have a consistent reproducer (`co2_thermal_3d.xml`) that fails on at least five different non-LLNL systems and that can be shared as needed.  Errors are slightly different on different systems, but often seem to related to the ghost maps as they are being set up.

On Sherlock (at Stanford), it has been observed that the number of MPI ranks on a node matters.  It will run with 64 ranks / 1 node, but fail for 64 ranks / 4 nodes.

On LLNL systems, the problem will fail if enough ranks are requested to create a single layer of elements on a given rank, but this may be an unrelated issue associated with neighbors-of-neighbors maps.  We mention it here just in case it proves relevant.

kjdugan:
Related to issue #2173, where there are system dependent nondeterministic mpi errors. This fix isolates multiple calls to mpi-wait with barriers.

This is a short-term hack which will resolve the issues seen with large processor counts on certain machines. This can open a discussion on the overall async mpi strategy used in the neighbor communication scheme.

resolves #2173

### DIFF
--- a/src/coreComponents/mesh/mpiCommunications/CommunicationTools.cpp
+++ b/src/coreComponents/mesh/mpiCommunications/CommunicationTools.cpp
@@ -689,11 +689,23 @@ void CommunicationTools::setupGhosts( MeshLevel & meshLevel,
 
   waitOrderedOrWaitAll( neighbors.size(), { sendSyncLists, postRecv, rebuildSyncLists }, unorderedComms );
 
+  // There are cases where the multiple waitOrderedOrWaitAll methods will clash
+  // with each other. Mostly at high processor counts (>256) and large
+  // meshes (~14M elements). Adding in these barriers isolates the wait calls
+  // from each other, and resolves some non-deterministic errors.
+  // IMO, this is a hack and we should revisit the async communication strategy
+  // here.
+  MpiWrapper::barrier();
+
   fixReceiveLists( nodeManager, neighbors );
   fixReceiveLists( edgeManager, neighbors );
   fixReceiveLists( faceManager, neighbors );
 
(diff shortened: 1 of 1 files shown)


kjdugan:
Looks like all the checks passed except for the PR-assigned test but I don't have permission to modify the assignees or change labels.

joshua-white (approved):

rrsettgast:
@kjdugan I suspect that this fix may not actually provide a rock solid "fix". In the discussion #2173 I propose that a rewrite of the wait functionalities ignores  all `MPI_Status` and `MPI_Request` associated with the send operations. I might be incorrect, but if I am correct then this is a fundamentally unsafe set of operations that needs to be addressed directly. While the `MPI_Barrier` may "fix" the problem, perhaps it does not fix the problem directly.

rrsettgast:
@kjdugan I have proposed a different fix in your branch. Can you check to see that it works for you? If not, I will just drop the commit.

kjdugan:
@rrsettgast, Your changes worked great. I did a little cleanup and documentation locally for the reasoning behind the waitAlls.

rrsettgast:
@kjdugan Great. Thank you for finding the issue. We will just merge this into develop, and create a follow up issue to implement a modification to the communication routines in order to avoid this reoccurring.

PR #2231, opened 2022-12-29
Updating xml inputs for optional features

cssherman:


wrtobin (approved):

rrsettgast (approved):
@cssherman Is this still going into develop?

cssherman:
@rrsettgast - I think that we ended up moving a different direction here, so we can cancel this.

PR #2233, opened 2023-01-04
DRAFT: Simplification of async communication

wrtobin:
We should be able to remove the explicit barriers by adding a phase that will act as an implicit barrier only on the neighborhood group rather than the entire comm.

### DIFF
--- a/src/coreComponents/mesh/mpiCommunications/CommunicationTools.cpp
+++ b/src/coreComponents/mesh/mpiCommunications/CommunicationTools.cpp
@@ -649,17 +649,18 @@ void CommunicationTools::setupGhosts( MeshLevel & meshLevel,
     neighbors[idx].unpackGhosts( meshLevel, commData.commID() );
     return MPI_REQUEST_NULL;
   };
+  // we implement this as a separate phase in order not to impact the unpackGhosts performance by waiting on a request that
+  //  doesn't need to be complete for 'later' ghosts to be unpacked.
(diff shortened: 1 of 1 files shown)


TotoGaz:
In a lot of languages or packages, when it comes to async, there is a callback pattern that fires the downstream "functor" when the conditions are fulfilled.
Could we benefit from such a thing here?
Having a thicker `request` in which we could stack `mpi` related calls?

wrtobin:
Ultimately that is what the activeWait functions do, they launch the first phase (callback) and then poll the async status via the returned MPI_Request, and when the correct conditions are met the next phase is invoked via callback.

The difference between the different activeWait functions is which set of MPI_Requests is done before a phase callback happens, whether its just the previous MPI_Request associated with the same neighbor index vs all MPI_Requests from the previous phase (including the request associated with the same neighbor, etc).

Allowing a phase to return a set of requests instead of a single request would be a way to simplify since the phase could return both the send and recv requests, but that is ultimately less performant / more synchronous since the recv request is the only thing that really needs to be done before the next phase, the send doesn't become important until we go to reuse the same buffers and tags in subsequent phases.

wrtobin:
We could probably actually rewrite the entire function (which contains 3 rounds of communication using activeWaits) to just be a single activeWait call, though I don't think most developers would want to / be capable of modifying / debugging that if a bug crops up. As it is this took a few people a month or more to notice.

Its a hard balance between developer accessibility and developing an interface that can get us the optimal async execution flow.

PR #2237, opened 2023-01-07
Added dedicated MGR recipe for OBL

francoishamon:
This PR adds an MGR recipe for `ReactiveCompositionalMultiphaseOBL`. 

In the OBL formulation, the dofs are 1 pressure and NC-1 component fractions, so it is slightly different from our standard formulation and does not require the elimination of a saturation equation. As a result we can use a one-level recipe in which we eliminate all the component fractions in one pass, and then solve the coarse (pressure) system with BoomerAMG.

It seems to work well on thermal and reactive OBL tests. 

Ready for review.


castelletto1 (approved):
Looks good to me!

PR #2238, opened 2023-01-10
Added Gaussian elimination F-relaxation for wells in MGR

francoishamon:
This PR adds Gaussian elimination for the F-relaxation for the well block in MGR (instead of jacobi that we were using before this PR). This change leads to good reduction in the number of linear iterations and in run time, particularly when the wells have a lot of perforations.

Because using Gaussian elimination is an overkill when all the wells are shut (which happens for instance in the post-injection period for CO2 injection), this PR adjusts the MGR recipe dynamically (i.e., during the simulation) to revert to simple Jacobi F-relaxation when all wells are shut. Maybe later we can modify the well code to entirely the wells from the Jacobian if they are shut until the end of the simulation.

Ready for review.
Requires a hypre update

LvArray PR: https://github.com/GEOSX/LvArray/pull/276


src/coreComponents/linearAlgebra/interfaces/hypre/mgrStrategies/CompositionalMultiphaseReservoirHybridFVM.hpp
```diff
@@ -133,10 +137,17 @@ class CompositionalMultiphaseReservoirHybridFVM : public MGRStrategyBase< 4 >
...
+    if( mgrParams.areWellsShut )
+    {
+      m_levelFRelaxType[0] = MGRFRelaxationType::jacobi;
+    }
+
     // Note: uncomment HYPRE_MGRSetLevelFRelaxMethod and comment HYPRE_MGRSetRelaxType breaks the recipe, this requires further
     // investigation
     //GEOSX_LAI_CHECK_ERROR( HYPRE_MGRSetLevelFRelaxMethod( precond.ptr, toUnderlyingPtr( m_levelFRelaxMethod ) ) );
-    GEOSX_LAI_CHECK_ERROR( HYPRE_MGRSetRelaxType( precond.ptr, 0 ));
+    GEOSX_LAI_CHECK_ERROR( HYPRE_MGRSetLevelFRelaxType( precond.ptr, toUnderlyingPtr( m_levelFRelaxType ) ));
```
castelletto1, line 150: I think this is the intended behavior. In my understanding the goal is to define the f-relaxation exclusively through `HYPRE_MGRSetLevelFRelaxType` (or `HYPRE_MGRSetRelaxType` for a two-level reduction) without specifying whether it is a single- or a multilevel method. Hence, we could completely get rid of `m_levelFRelaxMethod`:
https://github.com/GEOSX/GEOSX/blob/8fb694ec4bda02d1e8691b37310d94622c8be9b0/src/coreComponents/linearAlgebra/interfaces/hypre/HypreMGR.hpp#L85

@oseikuffuor1 do you agree?

@francoishamon since we have discussed several cleanups for MGR I would address this in another PR.

oseikuffuor1: @castelletto1 Yes you are correct. There should be no need to call MGRSetFrelaxMethod anymore.

castelletto1 (approved):

oseikuffuor1 (approved):
Looks good to me. Thanks @francoishamon

src/coreComponents/linearAlgebra/interfaces/hypre/HypreMGR.hpp
```diff
@@ -124,6 +128,39 @@ class MGRStrategyBase
       m_ptrLabels[i] = m_labels[i].data();
     }
   }
+
+  /**
+   * @brief Set up BoomerAMG to perform the mechanics F-solve for the first F-relaxation
+   * @param precond the preconditioner wrapper
+   * @param mgrData auxiliary MGR data
+   */
+  void setMechanicsFSolver( HyprePrecWrapper & precond,
```
TotoGaz, line 137: Why did you have to actually move this to the header file?

TotoGaz (commented):
LGTM for the travis part

src/coreComponents/linearAlgebra/interfaces/hypre/HypreMGR.hpp
francoishamon: Because now this function in the templated class `MGRStrategyBase`. I think this part will require a refactoring in which `setMechanicsFSolver` could become a free function. `setMechanicsFSolver` will also be renamed and will have counterparts to set the BoomerAMG options for the coarse system in MGR. So, many things are going to change soon and everything will be better organized.

CusiniM:
if you merge the related tpls PR I can then recompile the tpls on LC.

francoishamon:
> if you merge the related tpls PR I can then recompile the tpls on LC.

Ok thanks very much, I am going to do it now

francoishamon:
> This is ready. Just needs the LvArray PR to be approved and merged.

Great, thanks very much @CusiniM

PR #2239, opened 2023-01-11
Fix bug when we get the number of degrees of freedom which implies resize with 0 as dimension

acitrain:
In this PR we fix a bug found on the elastic solver, where the number of the degrees of freedom was badly get inside postProcessInput function, which implies that several call of the resize function was useless as we resize with 0 dimension.

This was hidden by duplicated code inside postProcessInput and initializePreSubGroups. 
Here the bug is fixed  fand we also remove the duplicated part for the second order acoustic solver.


sframba (approved):

sframba:
@CusiniM @francoishamon all the integrated tests on quartz, they all pas with this PR. I am putting it at the top of the merge queue so it can be merged quickly.

mndiaye24 (approved):

PR #2245, opened 2023-01-16
Fault slip test case

CusiniM:


mediev:
Hi Matteo,

1. I fixed the reservoir geometry (width=225, offset=75) and applied the Neumann condition for the top boundary according to the benchmark. 
2. In the case of plain strain conditions (zconstraint1 is activated) without fault (no fault segments), and no depletion, I made a comparison against DARTS (collocated FV). A snapshot with a comparison of pressures, total stresses, and displacements (linearly interpolated to cell centers) is here: https://drive.google.com/file/d/1_bwvoTC4F7RN8ungIZrfsbSaLXOJQTPE/view?usp=sharing
Pressures and displacements fit each other quite well, there is a high mismatch in stresses close to the Dirichlet boundaries, it seems that cell-centered FV is less accurate there.
displ
pres
sxx

3. The problem is that I have not managed to achieve a convergent solution in the presence of a fault in both cases (before and after my commits). Now the residual cant decrease more than 200 times for me. Before my commits, it decreased by 10 orders of magnitude, but it still has not achieved the target. Both logs are attached.
[log_before_commits.txt](https://github.com/GEOSX/GEOSX/files/10446429/log_before_commits.txt)
[log_current_state.txt](https://github.com/GEOSX/GEOSX/files/10446430/log_current_state.txt)

If it is easier to start debugging from the initial state, please, revert my commits.
I will think about how to assign pressure depletion in the reservoir.

Best,
Aleks

av-novikov:
It may be that LagrangianContactSolver::assembleStabilization creates problems with normal gap. But i struggle to understand what is going on there.

First I checked the tolerances for normal gap and normal traction in LagrangianContactSolver::computeTolerances. The one for the normal gap is much lower compared to the values we obtain (see below). 
image

If I run verticalFault_puremech.xml it violates the impenetrability condition. Negative normal traction indicates the problem in the assembly of contact equations.
image

If I run verticalFault_puremech.xml with commented assembleStabilization in LagrangianContactSolver::assembleSystem it slowly converges to all zero normal gaps, even though normal traction is oscillating
image
It does not converge if i change friction coefficient. My point is that the contribution from assembleStabilization leads to much higher normal gap values than without it. 

Generally, i do not understand why normal gap exhibits significant (compared to tolerance) changes while normal traction remains negative.

av-novikov:
Now fluid pressure inside a fault is aligned with reservoir in verticalFault_poromechanics.xml. Unfortunately, it has not improved results.

PR #2246, opened 2023-01-17
Add solver for first order elastic wave equation 

acitrain:
This PR add a solver for the first order elastic wave equation.

The equation are solved using Spectral Element Method in space and 2nd order Leap-Frog in time
The main difference with the second order equation are the following:

Two unknowns (valocity (a vector)  and stress tensor) instead of one
The velocity is considered continuous and the stress tensor discontinuous
The stiffness matrix is the product of a gradient and a basis function instead of two gradient. The computation of the stiffness values is optimize as done for the second order equation in PR #2184 

Integrated test PR: https://github.com/GEOS-DEV/integratedTests/pull/321


sframba (approved):

src/coreComponents/physicsSolvers/wavePropagation/ElasticFirstOrderWaveEquationSEMKernel.hpp
```diff
@@ -0,0 +1,636 @@
...
+          arrayView2d< real64 > const sourceConstants,
+          arrayView2d< real64 const > const receiverCoordinates,
+          arrayView1d< localIndex > const receiverIsLocal,
+          arrayView1d< localIndex > const rcvElem,
+          arrayView2d< localIndex > const receiverNodeIds,
+          arrayView2d< real64 > const receiverConstants,
+          arrayView2d< real32 > const sourceValue,
+          real64 const dt,
+          real32 const timeSourceFrequency,
+          localIndex const rickerOrder )
```
francoishamon, line 89: Maybe it is worth merging this function with `PrecomputeSourceAndReceiverKernel` from `AcousticFirstOrderWaveEquationSEMKernel.hpp` (and the other kernel files) because it is essentially the same kernel duplicated fourth times, unless I missed something?

francoishamon (approved):
Looks good to me. If you can add an integrated test before merging, that would be great, otherwise we will break the code without knowing

src/coreComponents/physicsSolvers/wavePropagation/ElasticFirstOrderWaveEquationSEMKernel.hpp
acitrain: Yes François I agree. 
In fact 3 of the 4 propagators (including the one in this PR) use the exact same function. 
However, as it will modify calls inside nearly all solvers, we think that we can keep it that way for the framework of this PR, in order to merge it (it started to be here for a long time...) and then we will do a refactoring PR which will adress this (and maybe other somilar points).

PR #2247, opened 2023-01-17
DofManager small refactoring to avoid registering solid model on fracture subRegions: 

CusiniM:
Currently we are required to add `SurfaceElementRegions` as targets to the `SollidMechanics` solver because it is in charge of registering the displacement dofs. Additionally, because of [this piece of code](https://github.com/GEOSX/GEOSX/blob/adc36adc83d926c91cec4d8c0f0edfd0d59e9ac3/src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.cpp#L205) all targets of a SollidMechanics` solver are required to have a solid model. This is problematic for `FaceElementSubRegions` that sit at the interface of two cells with different solid materials. 

In this PR we fix this issue by making fracture solver add displacement dofs to the dofManager and not requiring any more `SurfaceElementRegions`to be target of the solid solver. 

Rebaseline PR: https://github.com/GEOSX/integratedTests/pull/291


src/coreComponents/linearAlgebra/DofManager.cpp
```diff
@@ -311,25 +311,42 @@ void DofManager::addField( string const & fieldName,
...
+    // advanced processing
+    field.support = processedRegions;
+  }
+  else
+  {
+    FieldDescription & field = m_fields[getFieldIndex( fieldName )];
+    for( auto const & newRegion : processedRegions )
+    {
+      auto it = field.support.begin();
+      auto end1 = field.support.end();
```
CusiniM, line 337:
```suggestion
      auto const end1 = field.support.end();
```

src/coreComponents/linearAlgebra/DofManager.cpp
```diff
@@ -456,31 +474,60 @@ void DofManager::addCoupling( string const & rowFieldName,
...
+  }
+  else
   {
-    m_coupling.insert( { { colFieldIndex, rowFieldIndex }, coupling } );
+    CouplingDescription & coupling = m_coupling[ { rowFieldIndex, colFieldIndex } ];
+
+    for( auto const & newRegion : processedRegionList )
+    {
+      auto it    = coupling.support.begin();
+      auto end1  = coupling.support.end();
```
CusiniM, line 513:
```suggestion
      auto const end1  = coupling.support.end();
```

src/coreComponents/linearAlgebra/DofManager.hpp
```diff
@@ -71,6 +71,25 @@ class DofManager
...
+      bool added = false;
+      if( meshBodyName  == input.meshBodyName && meshLevelName == input.meshLevelName )
+      {
+        for( auto const & inputRegionName : input.regionNames )
+        {
+          if( std::find( regionNames.begin(), regionNames.end(), inputRegionName ) == regionNames.end() )
+          {
+            regionNames.push_back( inputRegionName );
+            std::sort( regionNames.begin(), regionNames.end() );
+          }
```
rrsettgast, line 86: maybe `regionNames` should be a set?

src/coreComponents/linearAlgebra/DofManager.cpp
```diff
@@ -456,31 +474,60 @@ void DofManager::addCoupling( string const & rowFieldName,
...
+    {
+      m_coupling.insert( { { colFieldIndex, rowFieldIndex }, coupling } );
+    }
+  }
+  else
   {
-    m_coupling.insert( { { colFieldIndex, rowFieldIndex }, coupling } );
+    CouplingDescription & coupling = m_coupling[ { rowFieldIndex, colFieldIndex } ];
+
+    for( auto const & newRegion : processedRegionList )
```
rrsettgast, line 510: looks like this is replicated. should we pull this code into a function?

rrsettgast (approved):

src/coreComponents/linearAlgebra/DofManager.hpp
```diff
@@ -71,6 +71,25 @@ class DofManager
...
+    bool add( Regions const & input )
+    {
+      bool added = false;
+      if( meshBodyName  == input.meshBodyName && meshLevelName == input.meshLevelName )
+      {
+        for( auto const & inputRegionName : input.regionNames )
+        {
+          if( std::find( regionNames.begin(), regionNames.end(), inputRegionName ) == regionNames.end() )
+          {
+            regionNames.push_back( inputRegionName );
```
TotoGaz, line 84: If you want to make the insertion into `regionNames` specific, maybe it would make sense to declare `regionNames` `private` to impose the usage of `add`?
Plus maybe renaming into `addRegionName` would be a tad clearer? (plus doxygen of course)

src/coreComponents/linearAlgebra/DofManager.cpp
```diff
@@ -456,31 +474,60 @@ void DofManager::addCoupling( string const & rowFieldName,
...
+    {
+      auto it    = coupling.support.begin();
+      auto end1  = coupling.support.end();
+      bool added = false;
+      while( it != end1 && !added )
+      {
+        added = it->add( newRegion );
+        it++;
+      }
```
TotoGaz, line 519:
```suggestion
      bool added = false;
      for( Regions & regions: coupling.support )
      {
        added = regions.add( newRegion );
        if( added ) break;
      }
```
Would this be equivalent?

src/coreComponents/linearAlgebra/DofManager.cpp
TotoGaz: I think the conster is the better. You may even try to use `cend` 😉

TotoGaz (commented):
First comments

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.cpp
```diff
@@ -36,6 +36,7 @@
 #include "mainInterface/ProblemManager.hpp"
 #include "mesh/DomainPartition.hpp"
 #include "mesh/FaceElementSubRegion.hpp"
+#include "mesh/CellElementSubRegion.hpp"
```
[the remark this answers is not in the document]
CusiniM: I guess not coz the compiler was complaining. I was kind of surprised though.

src/coreComponents/linearAlgebra/DofManager.hpp
CusiniM: hmm would have to check how else it's used though. I mean I can do it but it may need a lot more changes.

src/coreComponents/linearAlgebra/DofManager.cpp
CusiniM: looks equivalent to me.

src/coreComponents/linearAlgebra/DofManager.hpp
CusiniM: yeah...I was thinking that too. It's just requires more changes to some functions in `DofManager`. If we do that it would be nice to also find a better name to this struct `Regions` since it now contains also `meshBody` and `meshLevel`... Maybe it should be called `FieldSupport`?

TotoGaz: Furthermore, we have `FieldIdentifiers`, `MeshObjectPath` and `DofManager::Regions`, which while slightly different, all look alike. Could we rethink those?

CusiniM: I feel the `MeshObjectPath` has kind of its own functionality linked to how we also specify b.c. etc. `FieldIdentifiers` and `DofManager::FieldSupport` (I am changing the name) are indeed very similar and could, probably be consolidated into a single thing.

CusiniM: Although maybe `FieldDescription` is even similar.

TotoGaz: Indeed. We may try to at least unite them in some kind of file/namespace?

CusiniM: Let me close this PR with the functionality I need and then I will open a small to try out merging 1, 2 or even all these things into a single object

src/coreComponents/linearAlgebra/DofManager.cpp
CusiniM: done

src/coreComponents/physicsSolvers/contact/LagrangianContactSolver.cpp
```diff
@@ -539,14 +539,26 @@ void LagrangianContactSolver::setupDofs( DomainPartition const & domain,
     meshTargets[std::make_pair( meshBodyName, meshLevel.getName())] = std::move( regions );
   } );
 
+  dofManager.addField( solidMechanics::totalDisplacement::key(),
+                       FieldLocation::Node,
+                       3,
```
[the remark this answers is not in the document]
klevzoff: We decoupled DoF definitions from physical fields on purpose to allow more flexibility in assigning/blocking dofs (e.g. node-wise vs component-wise numbering). But it should be possible to add a method that takes a Field trait and adds a DoF according to it, since this is the default use case.

klevzoff (approved):

src/coreComponents/physicsSolvers/contact/ContactSolverBase.cpp
```diff
@@ -244,7 +244,7 @@ void ContactSolverBase::synchronizeFractureState( DomainPartition & domain ) con
   {
     FieldIdentifiers fieldsToBeSync;
 
-    fieldsToBeSync.addElementFields( { fields::contact::fractureState::key() }, { getFractureRegionName() } );
+    fieldsToBeSync.addElementFields( { fields::contact::fractureState::key() }, std::vector< string >( { getFractureRegionName() } ) );
```
CusiniM, line 247:
```suggestion
    fieldsToBeSync.addElementFields( { fields::contact::fractureState::key() }, { getFractureRegionName() } );
```

src/coreComponents/physicsSolvers/contact/LagrangianContactSolver.cpp
```diff
@@ -248,7 +248,7 @@ void LagrangianContactSolver::implicitStepComplete( real64 const & time_n,
     // Need a synchronization of deltaTraction as will be used in AssembleStabilization
     FieldIdentifiers fieldsToBeSync;
     fieldsToBeSync.addElementFields( { contact::deltaTraction::key() },
-                                     { getFractureRegionName() } );
+                                     std::vector< string >( { getFractureRegionName() } ) );
```
CusiniM, line 251:
```suggestion
                                      { getFractureRegionName() } );
```

src/coreComponents/physicsSolvers/contact/SolidMechanicsEmbeddedFractures.cpp
```diff
@@ -673,7 +673,7 @@ void SolidMechanicsEmbeddedFractures::applySystemSolution( DofManager const & do
 
     fieldsToBeSync.addElementFields( { contact::dispJump::key(),
                                        contact::deltaDispJump::key() },
-                                     { getFractureRegionName() } );
+                                     std::vector< string >( { getFractureRegionName() } ) );
```
CusiniM, line 676:
```suggestion
                                      { getFractureRegionName() } );
```

src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp
```diff
@@ -211,7 +211,7 @@ real64 HydrofractureSolver::fullyCoupledSolverStep( real64 const & time_n,
       fieldsToBeSync.addElementFields( { flow::pressure::key(),
                                          flow::pressure_n::key(),
                                          SurfaceElementSubRegion::viewKeyStruct::elementApertureString() },
-                                       { m_surfaceGenerator->getFractureRegionName() } );
+                                       std::vector< string >( { m_surfaceGenerator->getFractureRegionName() } ) );
```
CusiniM, line 214:
```suggestion
                                        { m_surfaceGenerator->getFractureRegionName() } );
```

src/coreComponents/mesh/generators/VTKUtilities.cpp
```diff
@@ -1735,7 +1735,7 @@ void importFieldOnCellElementSubRegion( integer const logLevel,
 
       // Now that we know that the subRegion has this wrapper, we can add the wrapperName to the list of fields to
       // synchronize
-      fieldsToBeSync.addElementFields( {wrapperName}, {region.getName()} );
+      fieldsToBeSync.addElementFields( {wrapperName}, std::vector< string >( {region.getName()} ) );
```
CusiniM, line 1738:
```suggestion
      fieldsToBeSync.addElementFields( {wrapperName},  {region.getName()}  );
```

src/coreComponents/linearAlgebra/DofManagerHelpers.hpp
```diff
@@ -263,7 +263,7 @@ struct MeshVisitor< LOC, LOC, VISIT_GHOSTS >
 {
   template< typename POLICY, typename ... SUBREGIONTYPES, typename LAMBDA >
   static void visit( MeshLevel const & meshLevel,
-                     std::vector< string > const & regions,
+                     std::set< string > const & regions,
```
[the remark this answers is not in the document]
CusiniM: done

CusiniM:
@av-novikov this PR is almost ready and fixes one of the issues you mentioned about having a fault between two different materials.

PR #2254, opened 2023-01-30
Skipped system setup to reduce data movement to GPU

francoishamon:
This PR adds a flag to skip the recomputation of the sparsity pattern after the first time step for flow solvers (some already had the flag) and for coupled solvers. 

The goal of this PR is to reduce the data movement to the GPU taking place after each sparsity pattern recomputation. 

The PR is ready for review and does not need a rebaseline.


francoishamon:
We still have a relatively large data movement from CPU to GPU at every Newton iteration for all implicit solvers when we create a dummy vector for the setup of the hypre solver (see the logs posted by @victorapm on the ecp-strech channel):
https://github.com/GEOSX/GEOSX/blob/4468c19d4db341073ff6de48df63ffeb74f50b1c/src/coreComponents/linearAlgebra/interfaces/hypre/HypreSolver.cpp#L210-L216
@victorapm is there a way to set up the solver without passing a dummy vector? What are the expected arguments (rhs and solution I assume)?

francoishamon:
We still have relatively large data movement from CPU to GPU at every Newton iteration when we assemble certain types of boundary conditions, like `SourceFlux` for the flow or imposed displacements for the mechanics. The moved vectors are allocated here:
https://github.com/GEOSX/GEOSX/blob/4468c19d4db341073ff6de48df63ffeb74f50b1c/src/coreComponents/fieldSpecification/FieldSpecificationBase.hpp#L755-L759
and moved in `computeRhsContribution`. For displacement boundary conditions the vectors can be quite large (almost all the boundary faces of the domain).

@rrsettgast is it something that we should take care of (for example by registering `dofArray` and `rhsContributionArray` in the data repository at the beginning of the simulation to move them only once, or by making them data members of the class)?

victorapm:
> is there a way to set up the solver without passing a dummy vector?

@francoishamon, unfortunately not since hypre's Krylov solvers require knowledge of the solution and right hand side vectors during setup. However, hypre could work by using information from the matrix only. That would require some changes in our infra-structure, but it is possible. Would this be something useful to GEOSX?

src/coreComponents/physicsSolvers/fluidFlow/FlowSolverBase.cpp
```diff
@@ -149,6 +149,32 @@ void FlowSolverBase::registerDataOnMesh( Group & meshBodies )
   }
 }
 
+real64 FlowSolverBase::solverStep( real64 const & time_n,
```
CusiniM, line 152: is this `solverStep` function even necessary? To me it looks like it's doing the same as `SolverBase`.

src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp
```diff
@@ -180,6 +180,7 @@ real64 HydrofractureSolver::fullyCoupledSolverStep( real64 const & time_n,
     int locallyFractured = 0;
     int globallyFractured = 0;
 
+    // recompute the sparsity pattern at every time step
```
CusiniM, line 183: we could change the value of that flag each time that there is a topology change. I can go in and implement it  in this PR if you like.

CusiniM (approved):
Looks good to me. We could just add the option for the `HydroFractureSolver` to only setUp the system if there is a topology change. It should be an easy thing to do.

src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp
francoishamon: Yes, that would be great, feel free to push your modifications to the PR.

src/coreComponents/physicsSolvers/fluidFlow/FlowSolverBase.cpp
francoishamon: Yes, for now it is necessary because the `solverStep` function is not implemented in `SolverBase`, but you are right that this function is a complete duplication of `solverStep` in `ContactSolverBase.cpp`, and also a duplication of `fullyCoupledSolverStep` in `CoupledSolver.hpp`. 

If we are ok to move the implementation of `solverStep` to `SolverBase.cpp`, then we get rid of these functions. And potentially, in `CoupledSolver.hpp`, we could just have:
```
virtual real64 fullyCoupledSolverStep( real64 const & time_n,                                                                                                                          
                                         real64 const & dt,                                                                                                                              
                                         int const cycleNumber,                                                                                                                          
                                         DomainPartition & domain )                                                                                                                      
 {       
      SolverBase::solverStep( time_n, dt, cycleNumber, domain );
 }
```
What do you think?

src/coreComponents/physicsSolvers/multiphysics/CoupledSolver.hpp
```diff
@@ -334,7 +334,11 @@ class CoupledSolver : public SolverBase
     real64 dtReturn = dt;
 
     // setup the coupled linear system
-    setupSystem( domain, m_dofManager, m_localMatrix, m_rhs, m_solution );
+    if( !m_systemSetupDone )
```
CusiniM, line 337:
```suggestion
    if( !systemSetupDone() )
```

src/coreComponents/physicsSolvers/simplePDE/LaplaceBaseH1.cpp
```diff
@@ -111,7 +111,11 @@ void LaplaceBaseH1::implicitStepSetup( real64 const & GEOSX_UNUSED_PARAM( time_n
                                        DomainPartition & domain )
 {
   // Computation of the sparsity pattern
-  setupSystem( domain, m_dofManager, m_localMatrix, m_rhs, m_solution );
+  if( !m_systemSetupDone )
```
CusiniM, line 114:
```suggestion
  if( !systemSetupDone() )
```

src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp
klevzoff: I thought about this (a long time ago) and came to the conclusion that, given the "decentralized" nature of GEOSX (different components driven by `EventManager` can interact with the mesh and modify it independently at arbitrary times), a more robust system for detecting mesh modifications should be based on some kind of timestamps (whether it be measured in event manager cycles or something else). I.e. any component, such as `SurfaceGenerator`, that modifies the mesh, must also save the timestamp of last modification in the mesh data structure, and every solver must check that timestamp against its own saved timestamp of when the linear system was last rebuilt (and do that for all mesh targets it operates on). Perhaps this warrants an infrastructure discussion?

francoishamon: Following up on this morning discussion, I started an implementation of the time stamp (mesh modification time stamp and system setup time stamp). I am not too sure about what to do for the multi-`MeshLevel` case, but the PR works and is ready to be looked at again.

src/coreComponents/physicsSolvers/multiphysics/CoupledSolver.hpp
```diff
@@ -368,13 +353,21 @@ class CoupledSolver : public SolverBase
...
+      // Only build the sparsity pattern if the mesh has changed
+      if( meshModificationTimestamp > solver->getSystemSetupTimestamp() )
+      {
+        solver->setupSystem( domain,
+                             solver->getDofManager(),
+                             solver->getLocalMatrix(),
+                             solver->getSystemRhs(),
+                             solver->getSystemSolution() );
+        solver->setSystemSetupTimestamp( meshModificationTimestamp );
+      }
```
klevzoff, line 370: Not sure if it would be better to roll the timestamp check/update into `setupSystem()`. Right now this piece is repeated in every solver that implements a custom solver step function. If moved, it will repeat in every solver that reimplements `setupSystem()` (not sure if there are more or less of these).

The latter can be improved by inverting the flow - always calling `SolverBase::setupSystem()` (i.e. not overriding it), and having it call some `virtual void fillSparsity()` that will be overridden by derived classes to invoke custom sparsity fill kernels. In that case, the timestamp check/bypass logic will only appear in code once. But this refactoring is beyond the scope of this PR.

klevzoff (approved):
Looks great!

src/coreComponents/physicsSolvers/multiphysics/CoupledSolver.hpp
francoishamon: Ok sounds good, I like the idea of having just one `setupSystem` and one place where we check the timestamp. I agree with the plan of doing that in a separate PR.

Issue #2243, opened 2023-01-13
Maximum Acceptable Face Nodes


philippechehade:
When trying to import a specific grid, the below error message was received:

***** Controlling expression (should be false): numFaceNodes > MAX_FACE_NODES
***** Rank 80: Node per face limit exceeded
Expected numFaceNodes <= MAX_FACE_NODES
  numFaceNodes = 10
  MAX_FACE_NODES = 9

What version of GeosX can be used to import grids with higher maximum face nodes? 

Version used : 490f777

@jeannepellerin

PR #2261, opened 2023-02-01
Changed FaceManager::MAX_FACE_NODES from 9 to 11.

MelReyCG:
In order to be consistent with the documentation (https://geosx-geosx.readthedocs-hosted.com/en/latest/coreComponents/mesh/docs/Mesh.html?highlight=prism#supported-formats) and the ElementType enum (which go up to Prism11), I'm changing FaceManager::MAX_FACE_NODES from 9 to 11 to support more nodes per face.
The issue #2243 should be solved.

@TotoGaz

### DIFF
--- a/src/coreComponents/mesh/FaceManager.hpp
+++ b/src/coreComponents/mesh/FaceManager.hpp
@@ -470,7 +470,7 @@ private:
   array2d< real64 > m_faceNormal;
 
   /// constant expression of the maximum number of nodes per faces
-  constexpr static int MAX_FACE_NODES = 9;
+  constexpr static int MAX_FACE_NODES = 11;
 
 };
 


TotoGaz (approved):

TotoGaz:
> The issue https://github.com/GEOSX/GEOSX/issues/2243 should be solved.

Does it actually solve it?

herve-gross:
@MelReyCG can you confirm if this PR fixes bug #2243 ?

MelReyCG:
I don't have access to a full developpement environement yet, but I will test that as soon as I can and get back to you.

jeannepellerin:
@PhilippeChehade checked, and we confirm that it solves the problem and that the flow initialization runs on the grid as expected.

PR #2265, opened 2023-02-01
Updating ATS

cssherman:
integratedTests PR: https://github.com/GEOS-DEV/integratedTests/pull/294
LvArray PR: https://github.com/GEOS-DEV/LvArray/pull/285
TPLs PR: https://github.com/GEOS-DEV/thirdPartyLibs/pull/220


cssherman:
@TotoGaz @CusiniM - with this PR, you should just need to run `make geosx_xml_tools` (this names needs updating) to install the required testing packages.  The tests _should_ then run like normal

cssherman:
Note: I've changed the old `make geosx_xml_tools` step to `make_geosx_python_tools`, as that is more descriptive.  I've also tried to update the cmake setup for python3, so that it will work better on other systems and produce better warnings.

src/coreComponents/python/CMakeLists.txt
```diff
@@ -1,61 +1,68 @@
...
+#    )
 
-# Set python style targets
-# set( python_sources modules/geosx_mesh_tools_package/geosx_mesh_tools/abaqus_converter.py )
+# blt_import_library( NAME python_interp
+#                     INCLUDES    ${PYTHON_INTERP_INCLUDE_DIRS}
+#                     HEADERS     ${PYTHON_INTERP_INCLUDE_DIRS}
+#                     HEADERS_OUTPUT_SUBDIR python_interp
+#                     LIBRARIES python2.7 
+#                   )
```
cssherman, line 13: I think this is very old leftover code when we were first testing python integration in the code

cssherman:
@TotoGaz @CusiniM @rrsettgast - A couple of questions:

1) Do we want to require all users building the code to supply the required python (>3.7, preferably with virtualenv installed) to build the code?  Currently, this is only mandatory for pygeosx builds.
2) Do we want to add the `make geosx_python_tools` step to the default make?

TotoGaz:
> 1. Do we want to require all users building the code to supply the required python (>3.7, preferably with virtualenv installed) to build the code?  Currently, this is only mandatory for pygeosx builds.

Currently, I'm able to run the integrated tests using the `python` environment provided by the package of my distribution (`python`, `numy`, `h5py`, `mpi4py`, ...). It's very convenient if the requirements are clear (`requirements.txt`?).

I'm not sure I answer your exact question, but I think that as long as we do have a compatible version of python and the appropriate packages, we should be able to run the integrated tests with no `cmake` action: pointing to the `geosx` executable and `integratedTests` folder should be enough.

> 2. Do we want to add the `make geosx_python_tools` step to the default make?

Is it a `installation` step or there's really some build?

cssherman:
@TotoGaz - this step calls `scripts/setupPythonEnvironment.bash` with the version of Python that cmake detects.  This script does the following:

1. Checks the python environment (alternatively, it can setup a miniconda environment)
2. Sets up a virtual environment if necessary
3. Installs the python modules in `src/coreComponents/python/modules/` and `integratedTests/scripts/`
4. Links key console scripts to the build/bin directory for the user's convenience

cssherman:
@CusiniM - it looks like the error message `No machine specifications for toss_3_x86_64_ib found, using generic.` is generated in the ats package here:  https://github.com/LLNL/ATS/blob/305737530f24a4663644ae9dfc83de727b9dda65/ats/configuration.py#L550

I'm suspicious that when geosxats runs in a new allocation that it cannot find the machine spec file...  It does work in an interactive session, so I'm not sure what is being called differently yet.

cssherman:
@CusiniM @TotoGaz - to help with the testing of this PR, I've added a quick method that allows users to override restartcheck parameters from the command line.  I've also added an option to restartcheck that will allow users to ignore values that are missing from target/baseline files (a common issue, which isn't handled by exclude for some reason).  These are specified via the new `-r/--restartCheckOverrides`.  For example, this would ignore the field "some_field" and ignore missing values:

`./geosxats -r exclude some_field -r skip_missing True`

cssherman:
@wrtobin - I've been testing the updated ats implementation on Lassen, and ran into some Cuda out of memory issues.  My first thought was this is due to over-allocation, but this seems to occur even when the tests are run sequentially (e.g.: `geosxats --ats sequential`).  The flag that you mentioned `--suppress-pinned` also doesn't seem to work.  Any thoughts?

One of the tests that reliably causes this issue is in the Hydrofracturing folder.  You can either give this folder to geosxats as an argument, or modify the main.ats file.

cssherman:
@TotoGaz - have you had a chance to test the current version of the tests on your machine?

cssherman:
@wrtobin - I've ran some of the failed tests on Lassen manually, and end up seeing the same errors (for the `lassen-clang@upstream` build).  These are the args that ATS are building for one of the hydraulic fracturing tests with jsrun:

```bash
BUILD_PATH=/g/g17/sherman/GEOS/build-lassen-clang@upstream-release
cd $BUILD_PATH/integratedTests/HydrofracturingSolver/
jsrun --smpiargs="-gpu" -n 1 -r 1 -a 9 -c 9 -g 1 $BUILD_PATH/bin/geosx -i kgdEdgeBased_C3D6_smoke.xml -x 3 -y 3 -z 1 -n kgdEdgeBased_C3D6 -o $BUILD_PATH/../integratedTests/tests/allTests/HydrofracturingSolver/kgdEdgeBased_C3D6_09 --suppress-pinned
```

And here is the error:
```bash
Time: 0s, dt:1s, Cycle: 0
CUDAassert: out of memory /usr/gapps/GEOSX/thirdPartyLibs/2023-01-23/install-lassen-clang@upstream-release/raja/include/RAJA/policy/cuda/MemUtils_CUDA.hpp:229
```

scripts/setupPythonEnvironment.bash
```diff
@@ -4,13 +4,17 @@
...
+                            "$PACKAGE_DIR/hdf5_wrapper_package"
+                            "$PACKAGE_DIR/pygeosx_tools_package"
+                            "$SCRIPT_DIR/../integratedTests/scripts/geosxats_package")
 declare -a LINK_SCRIPTS=("preprocess_xml"
                          "format_xml"
-                         "convert_abaqus")
+                         "convert_abaqus"
+                         "run_geosxats"
+                         "activate"
+                         "python")
```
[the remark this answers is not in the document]
cssherman: By default, this script is taking a cmake variable as an input.  If pygeosx is enabled, the version supplied will be `${CMAKE_BINARY_DIR}/lib/PYGEOSX/bin/python`, and if not it will be `${Python3_EXECUTABLE}`

TotoGaz:
Hello @cssherman I managed to make it work, thank you for your help.
The `--ats openmpi_install=/usr` to forward the specific arguments were fruitful.
Obviously there are lots of other things to do with this `ATS` update, but on my side I think the feature is OK.

host-configs/LLNL/lassen-clang@upstream.cmake
```diff
@@ -13,3 +13,5 @@ set(MPI_Fortran_COMPILER /usr/tce/packages/spectrum-mpi/spectrum-mpi-rolling-rel
 
 include(${CMAKE_CURRENT_LIST_DIR}/lassen-base.cmake)
 
+# ATS
+set(ATS_ARGUMENTS "--ats jsrun_omp --ats jsrun_bind=packed"  CACHE PATH "")
```
cssherman, line 17: These seem to be the magic arguments for getting ats to run correctly on lassen gpu builds

src/CMakeLists.txt
```diff
@@ -192,4 +192,14 @@ if (NOT GEOSX_INSTALL_SCHEMA)
...
+
+
+################################
+# Add integratedTests
+################################
+if( EXISTS "${CMAKE_SOURCE_DIR}/../integratedTests/CMakeLists.txt")
+  add_subdirectory( ${CMAKE_SOURCE_DIR}/../integratedTests integratedTests )
+else()
+  message( "Could not find the integratedTests submodule" )
+endif()
```
cssherman, line 205: Here is the new cmake hook for the integratedTests submodule

src/coreComponents/python/CMakeLists.txt
```diff
@@ -1,61 +1,55 @@
...
+          "${CMAKE_BINARY_DIR}/bin/preprocess_xml"
+          "${CMAKE_BINARY_DIR}/bin/format_xml"
+        )
+
+     add_custom_command( OUTPUT ${GEOSX_PYTHON_TOOLS_BINS}
+                         COMMAND bash ${CMAKE_SOURCE_DIR}/../scripts/setupPythonEnvironment.bash -p ${PYTHON_POST_EXECUTABLE} -b ${CMAKE_BINARY_DIR}/bin
+                         WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
+                       )
+
+     add_custom_target( geosx_python_tools
```
CusiniM, line 33: since we are adding a new target we could call it directly `geos_python_tools`. but it's also fine if we want to change all targets names at once

CusiniM (approved):

host-configs/LLNL/quartz-base.cmake
```diff
@@ -50,4 +50,9 @@ set(MKL_LIBRARIES ${MKL_ROOT}/lib/intel64/libmkl_intel_lp64.so
                   CACHE STRING "")
 
 
+# PYGEOSX
+set(ENABLE_PYGEOSX ON CACHE BOOL "")
+set(Python3_ROOT_DIR /usr/gapps/GEOSX/thirdPartyLibs/python/linux-rhel7-x86_64-gcc@8.1.0/python@3.8.5 CACHE PATH "")
+set(Python3_EXECUTABLE /usr/gapps/GEOSX/thirdPartyLibs/python/linux-rhel7-x86_64-gcc@8.1.0/python@3.8.5/bin/python3 CACHE PATH "")
+
```
cssherman, line 57: This needs to be updated for the current quartz install

CusiniM: you actually already have this above in the file. I think we can just delete these lines.

src/coreComponents/linearAlgebra/interfaces/hypre/HypreMGR.cpp
```diff
@@ -56,7 +56,7 @@ void hypre::mgr::createMGR( LinearSolverParameters const & params,
   // Hypre's parameters to use MGR as a preconditioner
   GEOS_LAI_CHECK_ERROR( HYPRE_MGRSetTol( precond.ptr, 0.0 ) );
   GEOS_LAI_CHECK_ERROR( HYPRE_MGRSetMaxIter( precond.ptr, 1 ) );
-  HYPRE_Int logLevel = LvArray::math::min( LvArray::integerConversion< HYPRE_Int >( params.logLevel - 1 ), 0 );
+  HYPRE_Int logLevel = LvArray::math::min( LvArray::integerConversion< HYPRE_Int >( params.logLevel - 1 ), LvArray::integerConversion< HYPRE_Int > ( 0 ) );
```
rrsettgast, line 59: Is this an error?

rrsettgast (approved):

src/coreComponents/linearAlgebra/interfaces/hypre/HypreMGR.cpp
wrtobin: Really gotta be sure about the type of that 0... 

Is there a compiler complaining about this somehow?

CusiniM: the mac compiler was complaining about this. It was interpreting 0 as an `int` and `HYPRE_Int` is a `long int` so it says the template parameter type` T` in `LvArray::math::min` is ambiguous. I am guessing other compilers are fine with it though.

wrtobin: Interesting failure mode though

wrtobin: Hmm I guess most compilers just don't care about the c-type coercion:

image

CusiniM: yes, it is `Apple clang version 14.0.0 (clang-1400.0.29.102)` to be precise. I was kind of surprised too but could not think of  a better fix

wrtobin: I mean per the standard the conversion is strictly the correct way to handle this, but very surprising in that almost any compiler will make such a literal up-conversion not a warning/error.

victorapm: Sorry for this!

CusiniM: blz, nao tem problema nao!

PR #2267, opened 2023-02-02
Use cell block name instead of region name in VTK field import.

TotoGaz:
- Uses the `subRegion` name instead of the `regionName` when importing the fields.
- Fixes the discrepancy between an argument name member function `MeshGeneratorBase::importFieldsOnArray` (was `cellBlockName`/`regioName`). Now always `cellBlockName`.
- Killing the simulation if no field could be imported: it's not normal I think.

I could run a simulation with fields imported from a `vtk` mesh.
@sframba @XL64 can you validate the fix?

Fixes https://github.com/GEOSX/GEOSX/issues/2266

### DIFF
--- a/src/coreComponents/mesh/MeshManager.cpp
+++ b/src/coreComponents/mesh/MeshManager.cpp
@@ -143,7 +143,7 @@ void MeshManager::importFields( DomainPartition & domain )
                                               region.getName(), subRegion.getName() ) );
 
         bool const isMaterialField = materialWrapperNames.count( geosxFieldName ) > 0 && wrapper.numArrayDims() > 1;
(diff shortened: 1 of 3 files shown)


sframba (approved):
Thanks Thomas, it is working very well for order 1, but fails at higher orders. May I suggest the following change that makes it work also for higher orders? In `MeshManager.cpp`, at line 114, instead of retrieving the `MeshLevel` associated to the base discretization, loop on all the mesh levels:
```
    dataRepository::Group & meshLevels = domain.getMeshBody( generator.getName() ).getMeshLevels();
    meshLevels.forSubGroups< MeshLevel >( [&] ( MeshLevel & meshLevel ) 
    {
      ...
    }
```
I have tested this at higher orders and it seems to fix the problem even in this case.

PR #2268, opened 2023-02-02
Modified integrated tests to test property import

francoishamon:
This PR modifies three existing integrated tests to add property import from the vtu file:
- `incompressible_pebi3d.xml`: import of porosity, permeability, net-to-gross
- `staircase_co2_wells_hybrid_3d.xml`: import of porosity and permeability
- `PoroElastic_hybridHexPrism_co2_3d.xml`: import of porosity, rock density and bulk modulus

The PR is ready for review and will require a rebaseline.
Rebaseline done in https://github.com/GEOSX/integratedTests/pull/297


inputFiles/compositionalMultiphaseWell/staircase_co2_wells_hybrid_3d.xml
```diff
@@ -59,9 +59,9 @@
 
     <VTKMesh
       name="mesh1"
-      file="staircase3d_tet.vtu"
-      fieldsToImport="{ FluidComposition }"
-      fieldNamesInGEOSX="{ globalCompFraction }"
+      file="staircase3d_tet_with_properties.vtu"
+      fieldsToImport="{ FluidComposition, PERM, PORO }"
```
TotoGaz, line 63: So the `PERM` in the vtk file is `(2.5e-16, 2.5e-16, 2.5e-16)` while in the original file it's only `(2.0e-16, 2.0e-16, 2.0e-16)`.

I assume it's to be sure that the `vtk` values are considered?

Similar question for `PORO` where your input is `0.11` and I think in the original case it was `0.1` (but I'm not sure).

I imagine there's the same kind of discrepancy for the other mesh?

francoishamon: Yes, exactly, I slightly changed the values of permeability and porosity compared to the xml default values to make sure that we notice it if the fields are not imported anymore.

Also, in the new `pebi3d_with_properties.vtu` file, the values are not homogeneous. It is a simple two-layer model. It contains flow properties like porosity and permeability, but also mechanical properties like rock density and bulk modulus because we are often asked how to import the latter, and up to now we did not have any example in the repository.

TotoGaz (approved):

Issue #2263, opened 2023-02-01
Segmentation fault in computeWellTransmissibility


philippechehade:
I have a model successfully initialized without wells.

When I tried adding an injection well to the model, the below error was obtained: 
```
MainGrid/Level0/wellregion/wellregionUniqueSubRegion/RELPERM allocated 1 quadrature points
Received signal 11: Segmentation fault ** StackTrace of 11 frames **
Frame 0: /lib64/libc.so.6 
Frame 1: geosx::PerforationData::computeWellTransmissibility(geosx::MeshLevel const&, geosx::WellElementSubRegion const&, LvArray::Array<LvArray::Array<LvArray::ArrayView<double const, 3, 2, int, LvArray::ChaiBuffer>, 1, camp::int_seq<long, 0l>, int, LvArray::ChaiBuffer>, 1, camp::int_seq<long, 0l>, int, LvArray::ChaiBuffer> const&) 
Frame 2: geosx::CoupledReservoirAndWellsBase<geosx::CompositionalMultiphaseBase, geosx::CompositionalMultiphaseWell>::initializePostInitialConditionsPreSubGroups()::{lambda(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, geosx::MeshLevel&, LvArray::ArrayView<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, 1, 0, int, LvArray::ChaiBuffer> const&)#1}::operator()(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, geosx::MeshLevel&, LvArray::ArrayView<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, 1, 0, int, LvArray::ChaiBuffer> const&) const::{lambda(int, geosx::WellElementSubRegion&)#1}::operator()(int, geosx::WellElementSubRegion) const 
[...]

PR #2269, opened 2023-02-02
Validate well to reservoir connections

dkachuma:
Implements a check on well to reservoir connections to ensure that all reservoir cells connected to wells are also targeted by a flow solver.

Should improve the error message related to issue #2263

### DIFF
--- a/src/coreComponents/physicsSolvers/multiphysics/CoupledReservoirAndWellsBase.cpp
+++ b/src/coreComponents/physicsSolvers/multiphysics/CoupledReservoirAndWellsBase.cpp
@@ -142,10 +142,12 @@ bool validateWellPerforations( SolverBase const * const solver,
       }
     } );
   } );
+
+  localIndex const badPerforationCount = MpiWrapper::max( badPerforation.first.empty() ? 0 : 1 );
+
   GEOSX_THROW_IF( !badPerforation.first.empty(),
                   GEOSX_FMT( "The well {} has a connection to the region {} which is not targeted by the solver", badPerforation.first, badPerforation.second ),
                   std::runtime_error );
-  localIndex const badPerforationCount = MpiWrapper::max( badPerforation.first.empty() ? 0 : 1 );
   return badPerforationCount == 0;
 }
 
--- a/src/coreComponents/physicsSolvers/multiphysics/CoupledReservoirAndWellsBase.cpp
+++ b/src/coreComponents/physicsSolvers/multiphysics/CoupledReservoirAndWellsBase.cpp
@@ -110,6 +110,45 @@ addCouplingNumNonzeros( SolverBase const * const solver,
   } );
 }
 
+bool validateWellPerforations( SolverBase const * const solver,
+                               WellSolverBase const * const wellSolver,
+                               DomainPartition const & domain )
+{
+  std::pair< string, string > badPerforation;
+  solver->forDiscretizationOnMeshTargets( domain.getMeshBodies(), [&] ( string const &,
+                                                                        MeshLevel const & meshLevel,
+                                                                        arrayView1d< string const > const & regionNames )
+  {
+    ElementRegionManager const & elemManager = meshLevel.getElemManager();
+    elemManager.forElementSubRegions< WellElementSubRegion >( regionNames, [&]( localIndex const, WellElementSubRegion const & subRegion )
+    {
+      PerforationData const * const perforationData = subRegion.getPerforationData();
+      WellControls const & wellControls = wellSolver->getWellControls( subRegion );
+
+      arrayView1d< localIndex const > const & resElementRegion =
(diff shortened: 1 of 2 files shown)


src/coreComponents/physicsSolvers/multiphysics/CoupledReservoirAndWellsBase.cpp
```diff
@@ -110,6 +110,47 @@ addCouplingNumNonzeros( SolverBase const * const solver,
...
+                               WellSolverBase const * const wellSolver,
+                               DomainPartition const & domain )
+{
+  std::pair< string, string > badPerforation;
+  solver->forDiscretizationOnMeshTargets( domain.getMeshBodies(), [&] ( string const &,
+                                                                        MeshLevel const & meshLevel,
+                                                                        arrayView1d< string const > const & regionNames )
+  {
+    ElementRegionManager const & elemManager = meshLevel.getElemManager();
+    elemManager.forElementSubRegions< WellElementSubRegion >( regionNames, [&]( localIndex const, WellElementSubRegion const & subRegion )
```
francoishamon, line 123: I have a suggestion to slightly change the way you loop over wells. In the current way, you loop over all the target regions of the coupled solver, then filter them out to get the well region. 

Instead, you could just loop over the well regions by doing: 
```suggestion
  wellSolver->forDiscretizationOnMeshTargets( domain.getMeshBodies(), [&] ( string const &,
                                                                        MeshLevel const & meshLevel,
                                                                        arrayView1d< string const > const & regionNames )
  {
    ElementRegionManager const & elemManager = meshLevel.getElemManager();
    elemManager.forElementSubRegions< WellElementSubRegion >( regionNames, [&]( localIndex const, WellElementSubRegion const & subRegion )
```

src/coreComponents/physicsSolvers/multiphysics/CoupledReservoirAndWellsBase.cpp
```diff
@@ -110,6 +110,47 @@ addCouplingNumNonzeros( SolverBase const * const solver,
...
+      // If the name of the region is not in the list of targetted regions, then we have a "bad" connection.
+      for( localIndex iperf = 0; iperf < perforationData->size(); ++iperf )
+      {
+        localIndex const er = resElementRegion[iperf];
+        string const regionName = elemManager.getRegion( er ).getName();
+        if( std::find( regionNames.begin(), regionNames.end(), regionName ) == regionNames.end())
+        {
+          badPerforation = {wellControls.getName(), regionName};
+          break;
+        }
```
francoishamon, line 141: and then, if you decide to follow the first suggestion, you can just search for the `regionName` in the target regions of the flow solver:
```suggestion
        if( std::find( flowTargetRegionNames.begin(), flowTargetRegionNames.end(), regionName ) == flowTargetRegionNames.end())
        {
          badPerforation = {wellControls.getName(), regionName};
          break;
        }
```
where `flowTargetRegionNames` would have been previously defined as:
```
arrayView1d< string const > const flowTargetRegionNames = solver->getReference< string >( viewKeyStruct::targetRegionsString );
```
With this approach, you would save a little bit by searching for `regionName` in a smaller array that does not contain the well regions. This would require to slightly change the arguments you pass to the function (`solver` is no longer the coupled solver).

But I have not tried, so there may be something that I am missing.

src/coreComponents/physicsSolvers/multiphysics/CoupledReservoirAndWellsBase.hpp
```diff
@@ -187,6 +191,10 @@ class CoupledReservoirAndWellsBase : public CoupledSolver< RESERVOIR_SOLVER, WEL
 
     DomainPartition & domain = this->template getGroupByPath< DomainPartition >( "/Problem/domain" );
 
+    // Validate well perforations: Ensure that each perforation is in a region targetted by the solver
+    if( !validateWellPerforations( domain ))
+      return;
```
francoishamon, line 196:
```suggestion
    if( !validateWellPerforations( domain ))
    {
      return;
    }
```
This is a nitpicking comment sorry, but this is the convention that we use in GEOS (at least, that is my understanding)

francoishamon (approved):
Great, thanks for solving this issue!

src/coreComponents/physicsSolvers/multiphysics/CoupledReservoirAndWellsBase.hpp
dkachuma: I had assumed this would be handled by uncrustify.

src/coreComponents/physicsSolvers/multiphysics/CoupledReservoirAndWellsBase.cpp
dkachuma: Make sense - and works. I have reordered the loop using the reservoir solver regions as the lookup list.

PR #2272, opened 2023-02-03
Split generateMesh into business and geometry parts

XL64:
Also defines interfaces for wells.

Relates to issue https://github.com/GEOS-DEV/GEOS/issues/2209

TPL update: https://github.com/GEOS-DEV/integratedTests/pull/15


src/coreComponents/mesh/generators/MeshGeneratorBase.hpp
```diff
@@ -19,19 +19,42 @@
...
+class MeshBody;
+
 class DomainPartition;
 
+class MeshGeneratorHelper
+{
+public:
+  bool hasMetisNeighborList() const { return m_hasMetisNeighborList; }
+  void setHasMetisNeighborList( bool value ) { m_hasMetisNeighborList = value; }
+  std::set<int> & getMetisNeighborList() {return m_metisNeighborList; }
```
TotoGaz, line 46: Returning a reference to non-const is basically not very different from having public attributes.
Once you get the reference, you cannot intercept any usage of the attribute which escaped your watch anymore.

src/coreComponents/mesh/generators/InternalMeshGenerator.cpp
```diff
@@ -543,18 +543,32 @@ static void getElemToNodesRelationInBox( ElementType const elementType,
  * @param partition
  * @param domain
  */
-void InternalMeshGenerator::generateMesh( DomainPartition & domain )
+MeshGeneratorHelper InternalMeshGenerator::generateCellBlockManager( CellBlockManager & cellBlockManager )
 {
   GEOSX_MARK_FUNCTION;
+  MeshGeneratorHelper meshGeneratorHelper;
+  meshGeneratorHelper.setHasMetisNeighborList( false );
+  SpatialPartition partition = meshGeneratorHelper.getSpatialPartition();
```
TotoGaz, line 551: Instead of using `SpatialPartition` directly here, what about simplifying `MeshGeneratorHelper` by simply providing `min` and `max`? It looks like this is the only information used.

TotoGaz (commented):
Good step forwards, we were already able to remove business at some significant places!

src/coreComponents/mesh/generators/MeshGeneratorBase.cpp
```diff
@@ -37,4 +39,27 @@ MeshGeneratorBase::CatalogInterface::CatalogType & MeshGeneratorBase::getCatalog
...
+  meshBody.createMeshLevel( 0 );
+  MeshLevel & meshLevel = meshBody.getBaseDiscretization();
+  CellBlockManager & cellBlockManager = meshBody.registerGroup< CellBlockManager >( keys::cellManager );
+
+  MeshGeneratorHelper meshGeneratorHelper = generateCellBlockManager( cellBlockManager );
+
+  this->generateWells( meshLevel ); // Accroche les informations des puits à l'instance du CellBlockManager crée à la ligne du dessus.
+  meshBody.setGlobalLengthScale( meshGeneratorHelper.getGlobalLength() );
+  return meshGeneratorHelper;
+}
```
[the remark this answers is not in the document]
XL64: Still have to work with CellBlockManagerABC instead of CellBlockManager everywhere I can.

src/coreComponents/mesh/generators/MeshGeneratorBase.cpp
```diff
@@ -37,4 +39,27 @@ MeshGeneratorBase::CatalogInterface::CatalogType & MeshGeneratorBase::getCatalog
...
+void MeshGeneratorBase::generateWells( MeshLevel & meshLevel ) {
+  forSubGroups<InternalWellGenerator>( [&]( InternalWellGenerator & wellGen ) {
+      wellGen.generateWellGeometry();
+      ElementRegionManager & elemManager = meshLevel.getElemManager();
+      WellElementRegion &
+        wellRegion = elemManager.getGroup( ElementRegionManager::groupKeyStruct::elementRegionsGroup() ).
+        getGroup< WellElementRegion >( wellGen.getWellRegionName() );
+      wellRegion.setWellGeneratorName( wellGen.getName() );
+      wellRegion.setWellControlsName( wellGen.getWellControlsName() );
+    } );
```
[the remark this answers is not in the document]
XL64: I did something like that

src/coreComponents/mesh/generators/MeshGeneratorBase.cpp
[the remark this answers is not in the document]
TotoGaz: Outside of the "module where we build the mesh", you should only see the abstractions (`*ABC`) classes.
Inside of this module, you can work with the implementations (and most probably you have to).

src/coreComponents/mesh/MeshManager.cpp
```diff
@@ -61,21 +62,32 @@ void MeshManager::generateMeshes( DomainPartition & domain )
   forSubGroups< MeshGeneratorBase >( [&]( MeshGeneratorBase & meshGen )
   {
     MeshBody & meshBody = domain.getMeshBodies().registerGroup< MeshBody >( meshGen.getName() );
-    meshGen.generateMesh(meshBody); meshBody.createMeshLevel( 0 );
+    meshBody.createMeshLevel( 0 );
+    MeshLevel & meshLevel = meshBody.getBaseDiscretization();
     CellBlockManager & cellBlockManager = meshBody.registerGroup< CellBlockManager >( keys::cellManager );
 
+    meshGen.generateMesh( cellBlockManager );
```
TotoGaz, line 69: I think that here we need to get something like
```suggestion
    CellBlockManagerABC const * cellBlockManager = meshGen.generateMesh();
```
such that you do no see the implementation `CellBlockManager`.

src/coreComponents/mesh/MeshManager.cpp
```diff
@@ -61,21 +62,32 @@ void MeshManager::generateMeshes( DomainPartition & domain )
...
+    meshBody.setGlobalLengthScale( cellBlockManager.getGlobalLength() );
+    ElementRegionManager & elemManager = meshLevel.getElemManager();
+    for ( WellBlock wellBlock : cellBlockManager.getWellBlocks() ) {
+      WellElementRegion &
+        wellRegion = elemManager.getGroup( ElementRegionManager::groupKeyStruct::elementRegionsGroup() ).
+        getGroup< WellElementRegion >( wellBlock.getWellRegionName() );
+      wellRegion.setWellGeneratorName( wellBlock.getName() );
+      wellRegion.setWellControlsName( wellBlock.getWellControlsName() );
+    }
```
TotoGaz, line 78: It's really cool to have this on this side of the mirror.
But I'm not sure this is the final place where this should be.
We can figure this out later.

src/coreComponents/mesh/MeshManager.cpp
```diff
@@ -61,21 +62,32 @@ void MeshManager::generateMeshes( DomainPartition & domain )
...
 
-    MeshGeneratorHelper meshGeneratorHelper = meshGen.generateMesh( meshBody );
-    if ( meshGeneratorHelper.hasMetisNeighborList() )
+    PartitionDescriptor & partitionDescriptor = cellBlockManager.getPartitionDescriptor();
```
TotoGaz, line 80: Do you need a non-`const` version of `PartitionDescriptor`?

src/coreComponents/mesh/MeshManager.cpp
```diff
@@ -61,21 +62,32 @@ void MeshManager::generateMeshes( DomainPartition & domain )
...
+      domain.getMetisNeighborList() = partitionDescriptor.getMetisNeighborList();
     }
     else
     {
       SpatialPartition & partition = dynamic_cast< SpatialPartition & >(domain.getReference< PartitionBase >( keys::partitionManager ) );
-      partition = meshGeneratorHelper.getSpatialPartition();
+      partition = partitionDescriptor.getSpatialPartition();
     }
-    meshBody.setGlobalLengthScale( meshGeneratorHelper.getGlobalLength() );
+    domain.getMeshBodies().registerGroup< MeshBody >( meshGen.getName() ).setGlobalLengthScale( cellBlockManager.getGlobalLength() );
```
TotoGaz, line 90: Isn't `meshBody` already available from line 64?

src/coreComponents/mesh/generators/MeshGeneratorBase.cpp
```diff
@@ -39,27 +39,17 @@ MeshGeneratorBase::CatalogInterface::CatalogType & MeshGeneratorBase::getCatalog
...
-  CellBlockManager & cellBlockManager = meshBody.registerGroup< CellBlockManager >( keys::cellManager );
+void MeshGeneratorBase::generateMesh( CellBlockManager & cellBlockManager ) {
 
-  MeshGeneratorHelper meshGeneratorHelper = generateCellBlockManager( cellBlockManager );
+  generateCellBlockManager( cellBlockManager ); 
 
-  this->generateWells( meshLevel ); // Accroche les informations des puits à l'instance du CellBlockManager crée à la ligne du dessus.
-  meshBody.setGlobalLengthScale( meshGeneratorHelper.getGlobalLength() );
-  return meshGeneratorHelper;
+  this->generateWells( cellBlockManager ); // Accroche les informations des puits à l'instance du CellBlockManager crée à la ligne du dessus.
```
TotoGaz, line 46: Warning, French.

src/coreComponents/mesh/generators/MeshGeneratorBase.hpp
```diff
@@ -100,9 +79,15 @@ class MeshGeneratorBase : public dataRepository::Group
    * @param[in] cellBlockManager to fill with the mesh informations
    * @return The global length scale
    */
-  virtual MeshGeneratorHelper generateCellBlockManager( CellBlockManager & cellBlockManager ) = 0;;
+  virtual void generateCellBlockManager( CellBlockManager & cellBlockManager ) = 0;
```
TotoGaz, line 82: I think this member function can be declared `private`.
And be careful of the `doxygen` comments.

src/coreComponents/mesh/generators/CellBlockManager.hpp
```diff
@@ -22,10 +22,88 @@
...
+   */
+  const string & getName() { return m_name; }
+  /**
+   * @brief Getter for the well region name
+   */
+  const string & getWellRegionName() { return m_wellRegionName; }
+  /**
+   * @brief Getter for the well controls name
+   */
+  const string & getWellControlsName() { return m_wellControlsName; }
```
TotoGaz, line 85: I'm surprised that this is the only information that's being created. I'd expect data like
```c++
  /// Physical location of the nodes
  array2d< real64 > m_nodeCoords;

  /// Global number of perforations
  globalIndex m_numPerforations;

  /// Absolute physical location of the perforation
  array2d< real64 > m_perfCoords;

  /// Well Peaceman index at the perforation
  array1d< real64 > m_perfTransmissibility;

  /// Global index of the well element
  array1d< globalIndex > m_perfElemId;
```
to appear there? Maybe it's WIP?

TotoGaz (commented):
Moving forward! 😉

src/coreComponents/mesh/generators/CellBlockManager.hpp
```diff
@@ -172,6 +250,33 @@ class CellBlockManager : public CellBlockManagerABC
...
+  /**
+   * @brief Setter for the global length
+   */
+  void setGlobalLength( real64 globalLength) { m_globalLength = globalLength; }
+
+  /**
+   * @brief Register a well block
+   * 
+   */
+  void registerWellBlock( const string & name, const string & wellRegionName, const string & wellControlsName )
```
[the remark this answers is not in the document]
TotoGaz: You can also have a look at the `importFractureNetwork` function [here](https://github.com/GEOSX/GEOSX/blob/develop/src/coreComponents/mesh/generators/VTKFaceBlockUtilities.cpp#L595-L634) where we build the `FaceBlock` data and then populate the `FaceBlock` with these computed data.

src/coreComponents/mesh/generators/InternalWellGenerator.hpp
```diff
@@ -30,7 +30,7 @@ namespace geosx
  *
  * This class processes the data of a single well from the XML and generates the well geometry
  */
-class InternalWellGenerator : public MeshGeneratorBase
+class InternalWellGenerator : public dataRepository::Group
```
TotoGaz, line 33: Try to hide `InternalWellGenerator` in the other parts of the code: we should be using `WellBlockABC` instead. That way we can separate the data from the producer.

src/coreComponents/mesh/generators/InternalWellGenerator.hpp
```diff
@@ -59,7 +59,7 @@ class InternalWellGenerator : public MeshGeneratorBase
   /**
    * @brief Default destructor.
    */
-  virtual ~InternalWellGenerator() override;
+  virtual ~InternalWellGenerator();
```
TotoGaz, line 62: I think that
```suggestion
  ~InternalWellGenerator() override;
```
would be the preferred combination.

src/coreComponents/mesh/generators/WellBlock.hpp
```diff
@@ -0,0 +1,297 @@
...
+   */
+  ///@{
+
+  // getters for element data
+
+  /**
+   * @brief Get the global number of well elements.
+   * @return the global number of elements
+   */
+  globalIndex getNumElements() const { return m_numElems; }
```
TotoGaz, line 49: Please add the `override` keyword when you override, it's very useful to prevent spelling or refactoring mistakes.
Plus we usually use the `numElements` naming convention. (see the `CellBlockManagerABC`, `FaceBlockABC` and `CellBlockABC` classes.)

src/coreComponents/mesh/generators/WellBlock.cpp
```diff
@@ -0,0 +1,498 @@
...
+  findPolylineHeadNodeIndex( internalWellGenerator );
+
+  // compute the location and distance from top of the well elements
+  discretizePolyline( internalWellGenerator );
+
+  // map the perforations to the well elements
+  connectPerforationsToWellElements( internalWellGenerator );
+
+  // make sure that the perforation locations are valid
+  checkPerforationLocationsValidity( internalWellGenerator );
```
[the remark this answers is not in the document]
XL64: Ok

src/coreComponents/mesh/generators/WellBlockABC.hpp
```diff
@@ -0,0 +1,133 @@
...
+namespace geosx
+{
+
+/**
+ * Abstract base class defining the information provided by any well block implementation.
+ *
+ * It's noteworthy that the WellblockABC is immutable oriented.
+ * The derived implementations need to have the modification/creation capabilities.
+ */
+class WellBlockABC
```
TotoGaz, line 34: @francoishamon what do you think of the interface of `WellBlockABC`? Does it make sense to you?

src/coreComponents/mesh/generators/WellBlock.hpp
XL64: I'll add it, I copied getters from InternalWellGenerator, they don't follow tje convention, I'll change that in WellBlock at least.

TotoGaz: > I'll add it, I copied getters from InternalWellGenerator, they don't follow tje convention, I'll change that in WellBlock at least.

Oh, I did not know that `InternalWellGenerator` did not respect the conventions... We need to fix this as well 🤣

src/coreComponents/mesh/ElementRegionManager.cpp
```diff
@@ -25,6 +25,7 @@
 #include "mesh/MeshManager.hpp"
```
TotoGaz, line 25:
```suggestion
```

src/coreComponents/mesh/ElementRegionManager.cpp
```diff
@@ -151,7 +152,7 @@ void ElementRegionManager::generateMesh( CellBlockManagerABC & cellBlockManager
 
 }
 
-void ElementRegionManager::generateWells( MeshManager & meshManager,
+void ElementRegionManager::generateWells( CellBlockManagerABC & cellBlockManager,
```
TotoGaz, line 155: Can you put `CellBlockManagerABC` `const`?
Conceptually, it should.

src/coreComponents/mesh/generators/WellBlock.hpp
```diff
@@ -208,15 +159,15 @@ class WellBlock : public WellBlockABC
   /// Radius area of the well (assumed to be valid for the entire well)
   real64 m_radius;
 
+  /// Name of the corresponding well region
+  string m_name;
```
TotoGaz, line 163: I think that this conflicts with the `m_name` from `Group`.

src/coreComponents/mesh/generators/WellBlock.cpp
```diff
@@ -25,474 +25,38 @@
...
-  m_numPerforations = internalWellGenerator.getNumPerforations();
-  m_perfCoords.resize( m_numPerforations, 3 );
-  m_perfDistFromHead.resize( m_numPerforations );
-  m_perfTransmissibility.resize( m_numPerforations );
-  m_perfElemId.resize( m_numPerforations );
+WellBlock::WellBlock( string const & name, Group * const parent ):
+  WellBlockABC( name, parent )
+{}
 
+void WellBlock::importFieldsFromInternalWellGenerator( InternalWellGenerator const & internalWellGenerator ) {
```
TotoGaz, line 33: Eventually, while heavy, sorry, we'll have to replace this with setters: we do not want `WellBlock` to depend on `InternalWellGenerator`, there will be other sources of information soon (or not so soon, but there will be 🤣 )!

Your IDE will surely let you define those setters automatically.

src/coreComponents/mesh/generators/WellBlockABC.hpp
```diff
@@ -126,6 +146,11 @@ class WellBlockABC
    */
   virtual arrayView1d< globalIndex const > getPerfElemIndex() const = 0;
 
+  virtual const string getName() const = 0;
```
TotoGaz, line 149: Maybe can you rely on `Group::getName`?

TotoGaz (commented):
Great!

src/coreComponents/mesh/generators/WellBlock.cpp
XL64: Emacs macros can also help :p; Done

src/coreComponents/mesh/generators/WellBlock.hpp
XL64: Removed, thanks

src/coreComponents/mesh/ElementRegionManager.cpp
XL64: We need nodeManager instead.

XL64: and some others on other files.

src/coreComponents/mesh/generators/MeshGeneratorBase.cpp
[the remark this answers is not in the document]
XL64: @TotoGaz There is no default constructor for CellBlockManager so I cannot construct it withoutthe mesh body to register it, or I don't know how to do it.

src/coreComponents/mesh/generators/InternalMeshGenerator.cpp
XL64: Later on we use other part of the spatialpartition :
```
grep --color -nH --null -e "\bpartition\b" InternalMeshGenerator.cpp
InternalMeshGenerator.cpp:543: * @param partition
InternalMeshGenerator.cpp:551:  SpatialPartition partition = partitionDescriptor.getSpatialPartition();
InternalMeshGenerator.cpp:563:    partition.setSizes( m_min, m_max );
InternalMeshGenerator.cpp:615:  // Get the first and last indices in this partition each direction
InternalMeshGenerator.cpp:625:      if( partition.isCoordInPartition( elemCenterCoords[i][k], i ) )
InternalMeshGenerator.cpp:636:        if( partition.isCoordInPartition( elemCenterCoords[i][k], i ) )
InternalMeshGenerator.cpp:644:  // Calculate number of elements in this partition from each region, and the
InternalMeshGenerator.cpp:730:  reduceNumNodesForPeriodicBoundary( partition, numNodesInDir );
InternalMeshGenerator.cpp:757:          setNodeGlobalIndicesOnPeriodicBoundary( partition, globalIJK );
InternalMeshGenerator.cpp:821:    if( std::any_of( partition.m_Periodic.begin(), partition.m_Periodic.end(), []( int & dimPeriodic ) { return dimPeriodic == 1; } ) )
InternalMeshGenerator.cpp:904:                // Fix local connectivity for single theta (y) partition (radial meshes only)
InternalMeshGenerator.cpp:985:  // 2) first local element in component partition is zero
```

XL64: Later on we use the SpatialPartition more:
```
grep --color -nH --null -e "\bpartition\b" InternalMeshGenerator.cpp
InternalMeshGenerator.cpp:543: * @param partition
InternalMeshGenerator.cpp:551:  SpatialPartition partition = partitionDescriptor.getSpatialPartition();
InternalMeshGenerator.cpp:563:    partition.setSizes( m_min, m_max );
InternalMeshGenerator.cpp:615:  // Get the first and last indices in this partition each direction
InternalMeshGenerator.cpp:625:      if( partition.isCoordInPartition( elemCenterCoords[i][k], i ) )
InternalMeshGenerator.cpp:636:        if( partition.isCoordInPartition( elemCenterCoords[i][k], i ) )
InternalMeshGenerator.cpp:644:  // Calculate number of elements in this partition from each region, and the
InternalMeshGenerator.cpp:730:  reduceNumNodesForPeriodicBoundary( partition, numNodesInDir );
InternalMeshGenerator.cpp:757:          setNodeGlobalIndicesOnPeriodicBoundary( partition, globalIJK );
InternalMeshGenerator.cpp:821:    if( std::any_of( partition.m_Periodic.begin(), partition.m_Periodic.end(), []( int & dimPeriodic ) { return dimPeriodic == 1; } ) )
InternalMeshGenerator.cpp:904:                // Fix local connectivity for single theta (y) partition (radial meshes only)
InternalMeshGenerator.cpp:985:  // 2) first local element in component partition is zero
```

src/coreComponents/mesh/generators/VTKMeshGenerator.hpp
```diff
@@ -85,7 +85,9 @@ class VTKMeshGenerator : public ExternalMeshGeneratorBase
    * surfaces of interest, with triangles and/or quads holding an attribute value
    * of 1, 2 or 3, three node sets named "1", "2" and "3" will be instantiated by this method
    */
-  virtual void generateMesh( DomainPartition & domain ) override;
+  virtual MeshGeneratorHelper generateCellBlockManager( CellBlockManager & cellBlockManager ) override;
+
+  void setNeighborList( std::set<int> neighborList );
```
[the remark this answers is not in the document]
XL64: Done

src/coreComponents/mesh/generators/MeshGeneratorBase.hpp
XL64: Done

src/coreComponents/mesh/generators/CellBlockManager.hpp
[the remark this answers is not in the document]
XL64: Done

src/coreComponents/mesh/MeshManager.cpp
XL64: Can't do so because Group(name, parent) needs parent for the constructor which is here meshBody, either we have CellBlockManager or meshBody as parameter of generateMesh() or I missed a thing.

src/coreComponents/mesh/MeshManager.cpp
XL64: It was because I was lazy ;).

src/coreComponents/mesh/MeshManager.cpp
XL64: Indeed

src/coreComponents/mesh/generators/MeshGeneratorBase.cpp
```diff
@@ -37,4 +39,17 @@ MeshGeneratorBase::CatalogInterface::CatalogType & MeshGeneratorBase::getCatalog
...
+  generateCellBlockManager( cellBlockManager ); 
+
+  this->generateWells( cellBlockManager ); // Accroche les informations des puits à l'instance du CellBlockManager crée à la ligne du dessus.
+}
+
+void MeshGeneratorBase::generateWells( CellBlockManager & cellBlockManager ) {
+  forSubGroups<InternalWellGenerator>( [&]( InternalWellGenerator & wellGen ) { 
+      wellGen.generateWellGeometry();
+      cellBlockManager.registerWellBlock( wellGen.getName(), wellGen.getWellRegionName(), wellGen.getWellControlsName() );
+    } );
```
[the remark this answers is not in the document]
XL64: generateWells()

src/coreComponents/mesh/generators/InternalWellGenerator.hpp
XL64: I think it's done.

src/coreComponents/mesh/generators/InternalMeshGenerator.cpp
```diff
@@ -543,12 +543,12 @@ static void getElemToNodesRelationInBox( ElementType const elementType,
...
-MeshGeneratorHelper InternalMeshGenerator::generateCellBlockManager( CellBlockManager & cellBlockManager )
+void InternalMeshGenerator::generateCellBlockManager( CellBlockManager & cellBlockManager )
 {
   GEOSX_MARK_FUNCTION;
-  MeshGeneratorHelper meshGeneratorHelper;
-  meshGeneratorHelper.setHasMetisNeighborList( false );
-  SpatialPartition partition = meshGeneratorHelper.getSpatialPartition();
+  PartitionDescriptor &partitionDescriptor = cellBlockManager.getPartitionDescriptor();
+  partitionDescriptor.setHasMetisNeighborList( false );
+  SpatialPartition partition = partitionDescriptor.getSpatialPartition();
```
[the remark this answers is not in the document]
TotoGaz: If you do
```c++
  SpatialPartition partition = partitionDescriptor.getSpatialPartition();
```
then `partition` will be a copy from the one provided by `partitionDescriptor.getSpatialPartition()`.
So in the rest of the function, when you update `partition` (`partition.setSizes(...)`), the update will never be sent back to `partitionDescriptor` (if this is the objective).
So you should use a reference (more or less `public`) or a `setPartition` to redefine the partition.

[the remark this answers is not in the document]
XL64: Indeed you are right.

src/coreComponents/mesh/generators/MeshGeneratorBase.cpp
```diff
@@ -37,4 +40,34 @@ MeshGeneratorBase::CatalogInterface::CatalogType & MeshGeneratorBase::getCatalog
   return catalog;
 }
 
+CellBlockManagerABC & MeshGeneratorBase::generateMesh( Group & parent )
+{
+  CellBlockManager & cellBlockManager = parent.registerGroup< CellBlockManager >( keys::cellManager );
+
+  generateCellBlockManager( cellBlockManager );
+
+  this->generateWells( cellBlockManager );
```
TotoGaz, line 49: What about calling this function `attachWellInfo` or `fillWellInfo`?

TotoGaz (commented):
We're nearly done!

src/coreComponents/mesh/ElementRegionManager.cpp
```diff
@@ -171,17 +173,17 @@ void ElementRegionManager::generateWells( MeshManager & meshManager,
 
     // get the global well geometry from the well generator
     string const generatorName = wellRegion.getWellGeneratorName();
-    InternalWellGenerator const & wellGeometry =
-      meshManager.getGroup< InternalWellGenerator >( generatorName );
+    WellBlockABC const & wellGeometry =
+      cellBlockManager.getGroup< WellBlockABC >( generatorName );
```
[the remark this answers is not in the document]
XL64: Done

src/coreComponents/mesh/MeshManager.cpp
```diff
@@ -59,7 +62,27 @@ void MeshManager::generateMeshes( DomainPartition & domain )
 {
   forSubGroups< MeshGeneratorBase >( [&]( MeshGeneratorBase & meshGen )
   {
-    meshGen.generateMesh( domain );
+    MeshBody & meshBody = domain.getMeshBodies().registerGroup< MeshBody >( meshGen.getName() );
+    meshBody.createMeshLevel( 0 );
+    MeshLevel & meshLevel = meshBody.getBaseDiscretization();
+
+    CellBlockManagerABC & cellBlockManager =  meshGen.generateMesh( dynamic_cast< Group & >( meshBody ) );
```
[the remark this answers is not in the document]
XL64: Done

src/coreComponents/mesh/MeshManager.cpp
```diff
@@ -59,7 +62,27 @@ void MeshManager::generateMeshes( DomainPartition & domain )
...
+    if( partitionDescriptor.hasMetisNeighborList() )
+    {
+      domain.getMetisNeighborList() = partitionDescriptor.getMetisNeighborList();
+    }
+    else
+    {
+      SpatialPartition & partition = dynamic_cast< SpatialPartition & >(domain.getReference< PartitionBase >( keys::partitionManager ) );
+      partition = partitionDescriptor.getSpatialPartition();
+    }
+    meshBody.setGlobalLengthScale( globalLength );
```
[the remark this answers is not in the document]
XL64: Done

src/coreComponents/mesh/generators/InternalMeshGenerator.cpp
```diff
@@ -543,18 +543,32 @@ static void getElemToNodesRelationInBox( ElementType const elementType,
  * @param partition
  * @param domain
  */
-void InternalMeshGenerator::generateMesh( DomainPartition & domain )
+void InternalMeshGenerator::generateCellBlockManager( CellBlockManager & cellBlockManager )
 {
   GEOSX_MARK_FUNCTION;
+  PartitionDescriptor & partitionDescriptor = cellBlockManager.getPartitionDescriptorInternal();
```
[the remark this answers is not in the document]
XL64: Done

src/coreComponents/mesh/MeshManager.cpp
```diff
@@ -59,7 +62,27 @@ void MeshManager::generateMeshes( DomainPartition & domain )
...
-    meshGen.generateMesh( domain );
+    MeshBody & meshBody = domain.getMeshBodies().registerGroup< MeshBody >( meshGen.getName() );
+    meshBody.createMeshLevel( 0 );
+    MeshLevel & meshLevel = meshBody.getBaseDiscretization();
+
+    CellBlockManagerABC & cellBlockManager =  meshGen.generateMesh( dynamic_cast< Group & >( meshBody ) );
+
+    PartitionDescriptor const & partitionDescriptor = cellBlockManager.getPartitionDescriptor();
+    real64 globalLength = cellBlockManager.getGlobalLength();
+    meshBody.setGlobalLengthScale( globalLength );
```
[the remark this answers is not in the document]
XL64: Done

src/coreComponents/mesh/generators/CellBlockManager.hpp
```diff
@@ -151,7 +151,9 @@ class CellBlockManager : public CellBlockManagerABC
 
   Group & getFaceBlocks() override;
 
-  Group & getWellBlocks() override;
+  Group & getLineBlocks() override;
```
TotoGaz, line 154: So we want to remove this?

XL64: In fact I was wrong, it was needed.

src/coreComponents/mesh/generators/CellBlockManager.cpp
```diff
@@ -681,6 +688,11 @@ FaceBlock & CellBlockManager::registerFaceBlock( string const & name )
   return this->getFaceBlocks().registerGroup< FaceBlock >( name );
 }
 
+WellBlock & CellBlockManager::registerWellBlock( string const & name )
+{
+  return this->getWellBlocks().registerGroup< WellBlock >( name );
```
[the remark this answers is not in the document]
XL64: I remove it from the ABC.

src/coreComponents/mesh/generators/CellBlockManagerABC.hpp
```diff
@@ -99,6 +100,14 @@ class CellBlockManagerABC : public dataRepository::Group
    */
   virtual Group & getFaceBlocks() = 0;
 
+  /**
+   * @brief Returns a group containing the well blocks as @p WellBlockABC instances.
+   * @return Mutable reference to the well blocks group.
+   *
+   * @note It should probably be better not to expose a non-const accessor here.
+   */
+  virtual Group & getWellBlocks() = 0;
```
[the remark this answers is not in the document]
XL64: I remove it from ABC and make it private.

src/coreComponents/mesh/MeshBody.hpp
```diff
@@ -166,6 +166,10 @@ class MeshBody : public dataRepository::Group
     return m_globalLengthScale;
   }
 
+  CellBlockManagerABC const & getCellBlockManager() const
```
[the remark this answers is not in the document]
XL64: ok

src/coreComponents/mesh/MeshManager.cpp
```diff
@@ -59,7 +62,27 @@ void MeshManager::generateMeshes( DomainPartition & domain )
...
   forSubGroups< MeshGeneratorBase >( [&]( MeshGeneratorBase & meshGen )
   {
-    meshGen.generateMesh( domain );
+    MeshBody & meshBody = domain.getMeshBodies().registerGroup< MeshBody >( meshGen.getName() );
+    meshBody.createMeshLevel( 0 );
+    MeshLevel & meshLevel = meshBody.getBaseDiscretization();
+
+    CellBlockManagerABC & cellBlockManager =  meshGen.generateMesh( dynamic_cast< Group & >( meshBody ) );
+
+    PartitionDescriptor const & partitionDescriptor = cellBlockManager.getPartitionDescriptor();
```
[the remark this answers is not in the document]
XL64: I'll move it not sure it is what you wanted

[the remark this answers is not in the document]
XL64: I'll move it noot sure it is how you wanted

src/coreComponents/mesh/generators/WellBlock.hpp
```diff
@@ -0,0 +1,290 @@
...
+  /**
+   * @brief Set the well transmissibility at the perforations.
+   * @param perfTransmissibility list of well transmissibility at all the perforations on the well
+   */
+  void setPerfTransmissibility( arrayView1d< real64 const > perfTransmissibility ) { m_perfTransmissibility = perfTransmissibility; }
+
+  /**
+   * @brief Get the global indices of the well elements connected to each perforation.
+   * @return list providing the global index of the connected well element for each perforation
+   */
```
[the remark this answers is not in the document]
XL64: removed all comment from overrided methods

src/coreComponents/mesh/generators/InternalWellGenerator.hpp
```diff
@@ -127,6 +130,45 @@ class InternalWellGenerator : public MeshGeneratorBase
    */
   globalIndex getNumElements() const { return m_numElems; }
 
+  /**
+   * @brief Get the number of well elements per segment.
+   * @return the global number of elements
+   */
+  globalIndex getNumElementsPerSegment() const { return m_numElemsPerSegment; }
```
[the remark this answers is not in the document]
XL64: ok

src/coreComponents/mesh/generators/PartitionDescriptor.hpp
```diff
@@ -0,0 +1,99 @@
...
+   * @brief Gets a reference to the list of metis neighbor list.
+   * @return A reference to the Metis neighbor list.
+   */
+  std::set< int > const & getMetisNeighborList() const { return m_metisNeighborList; }
+
+  /**
+   * @brief Sets the list of metis neighbor list.
+   * @param metisNeighborList A reference to the Metis neighbor list.
+   */
+  void setMetisNeighborList( std::set< int > const & metisNeighborList ) { m_metisNeighborList = metisNeighborList; }
```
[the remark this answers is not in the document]
XL64: En fait ici c'est n vecteur et plus tard ça devient un set, du coup obligé de faire la copie.

src/coreComponents/mesh/generators/PartitionDescriptor.hpp
```diff
@@ -54,15 +54,11 @@ class PartitionDescriptor
...
-   * @brief Sets the list of metis neighbor list.
-   * @param begin first iterator of the metis neighbor list.
-   * @parma end end iterator indicating the end of the metis neighbor list.
-   */
-  template< class InputIt >
-  void setMetisNeighborList( InputIt begin, InputIt end ) { m_metisNeighborList.insert( begin, end ); }
+  void setMetisNeighborList( std::vector< int > const & metisNeighborList ) {
+    for ( int i : metisNeighborList ) {
+      m_metisNeighborList.insert( i );
+    }
```
TotoGaz, line 60:
```suggestion
    m_metisNeighborList.insert( metisNeighborList.cbegin(), metisNeighborList.cend() );
```

src/coreComponents/mesh/generators/CellBlockManager.hpp
TotoGaz: So it may be `private`?

src/coreComponents/mesh/generators/CellBlockManagerABC.hpp
```diff
@@ -101,12 +102,20 @@ class CellBlockManagerABC : public dataRepository::Group
...
 
   /**
-   * @brief Returns a group containing the well blocks as @p WellBlockABC instances.
+   * @brief Returns a group containing the well blocks as @p LineBlockABC instances.
    * @return Mutable reference to the well blocks group.
    *
    * @note It should probably be better not to expose a non-const accessor here.
    */
-  virtual Group & getWellBlocks() = 0;
+  virtual Group & getLineBlocks() = 0;
```
TotoGaz: If we need `getLineBlocks` for the `CellBlockManager`, do we need it for the `ABC` version?

src/coreComponents/mesh/MeshManager.cpp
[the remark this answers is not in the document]
TotoGaz: Something like
```c++
    meshGen.generateMesh( meshBody );
    CellBlockManagerABC const & cellBlockManager = meshBody.getCellBlockManager();
    meshBody.setGlobalLengthScale( cellBlockManager.getGlobalLength() );

    PartitionDescriptor const & partitionDescriptor = cellBlockManager.getPartitionDescriptor();
    if( partitionDescriptor.hasMetisNeighborList() )	
```

src/coreComponents/mesh/generators/CellBlockManager.hpp
XL64: Yes

src/coreComponents/mesh/generators/CellBlockManagerABC.hpp
XL64: Indeed it has been removed, dont remember when, too much hollidays ^^...

XL64:
Don't know why I can"t answer your points but they seems all done.

src/coreComponents/mesh/generators/MeshGeneratorBase.hpp
```diff
@@ -122,12 +129,30 @@ class MeshGeneratorBase : public dataRepository::Group
...
+  SpatialPartition const & getSpatialPartition() const { return m_spatialPartition; }
+private:
+  /**
+   * @brief Fill the cellBlockManager object .
+   * @param[in] cellBlockManager to fill with the mesh informations
+   * @param[in] partition The number of domains in each dimesion (X,Y,Z)
+   */
+  virtual void fillCellBlockManager( CellBlockManager & cellBlockManager, array1d< int > const & partition ) = 0;
+
+  void attachWellInfo( CellBlockManager & cellBlockManager );
```
TotoGaz, line 145: Can you put the `private` below the `protected`?

src/coreComponents/mesh/generators/CellBlockManager.hpp
```diff
@@ -208,6 +241,7 @@ class CellBlockManager : public CellBlockManagerABC
   /**
    * @brief Trigger the face to nodes, edges and elements mappings.
    */
+
```
TotoGaz, line 244:
```suggestion
```

src/coreComponents/mesh/generators/CellBlockManagerABC.hpp
```diff
@@ -99,12 +101,26 @@ class CellBlockManagerABC : public dataRepository::Group
...
    * @brief Returns a group containing the cell blocks as CellBlockABC instances
    * @return Const reference to the Group instance.
    */
   virtual const Group & getCellBlocks() const = 0;
 
+  /**
+   * @brief Returns a group containing the face blocks as FaceBlockABC instances
+   * @return Const reference to the Group instance.
+   */
+  virtual const Group & getFaceBlocks() const = 0;
```
TotoGaz, line 122: Do you now if we still need the non const versions of the `getCellBlocks`, `getFaceBlocks`?

src/coreComponents/mesh/generators/InternalMeshGenerator.cpp
```diff
@@ -17,14 +17,6 @@
...
 
 #include "InternalMeshGenerator.hpp"
-
-#include "common/DataTypes.hpp"
-#include "common/TimingMacros.hpp"
-#include "mesh/DomainPartition.hpp"
-#include "mesh/MeshBody.hpp"
-#include "mesh/mpiCommunications/PartitionBase.hpp"
-#include "mesh/mpiCommunications/SpatialPartition.hpp"
-#include "mesh/MeshBody.hpp"
```
TotoGaz, line 27: 👍

src/coreComponents/mesh/generators/VTKMeshGenerator.hpp
```diff
@@ -51,7 +51,8 @@ class VTKMeshGenerator : public ExternalMeshGeneratorBase
 
   /**
    * @brief Generate the mesh using the VTK library.
-   * @param[in] domain the DomainPartition to be written
+   * @param[in] cellBlockManager the CellBlockManager to be written
```
TotoGaz, line 54:
```suggestion
   * @param[inout] cellBlockManager the CellBlockManager that will receive the meshing information
```

src/coreComponents/mesh/generators/MeshGeneratorBase.hpp
```diff
@@ -122,12 +129,30 @@ class MeshGeneratorBase : public dataRepository::Group
...
 
+  /**
+   * @brief Get the associatied SpatialPartition generated.
+   * @return The generated SpatialPartition
+   */
+  SpatialPartition const & getSpatialPartition() const { return m_spatialPartition; }
+private:
+  /**
+   * @brief Fill the cellBlockManager object .
+   * @param[in] cellBlockManager to fill with the mesh informations
```
TotoGaz, line 140:
```suggestion
   * @param[inout] cellBlockManager the CellBlockManager that will receive the meshing information
```

src/coreComponents/mesh/mpiCommunications/SpatialPartition.hpp
```diff
@@ -45,8 +45,33 @@ class SpatialPartition : public PartitionBase
...
+    return m_metisNeighborList;
+  }
+
+  /**
+   * @brief Sets the list of metis neighbor list.
+   * @param metisNeighborList A reference to the Metis neighbor list.
+   */
+  void setMetisNeighborList( std::vector< int > const & metisNeighborList )
+  {
+    m_metisNeighborList.insert( metisNeighborList.cbegin(), metisNeighborList.cend() );
```
TotoGaz, line 63:
```suggestion
    m_metisNeighborList.clear();
    m_metisNeighborList.insert( metisNeighborList.cbegin(), metisNeighborList.cend() );
```
Otherwise it's not a `set`, it's an insert. Maybe you want it to be an insert?

TotoGaz (commented):
I think we're close to the end. Heavy testing still needed 😉

TotoGaz (commented):
I think we're close to the end. Heavy testing still needed 😉

src/coreComponents/mesh/generators/CellBlockManagerABC.hpp
XL64: /appli_RD/LACOSTE/GEOSX/GEOSX/src/coreComponents/mesh/ElementRegionManager.cpp:130:    elemRegion.generateMesh( cellBlockManager.getFaceBlocks() );
/appli_RD/LACOSTE/GEOSX/GEOSX/src/coreComponents/mesh/ElementRegionManager.cpp:126:    elemRegion.generateMesh( cellBlockManager.getCellBlocks() );                                  |

src/coreComponents/mesh/generators/MeshGeneratorBase.hpp
XL64: 👍

CusiniM:
Can this PR be switched  back to draft since it has not been reviewed yet?

XL64:
> Can this PR be switched back to draft since it has not been reviewed yet?

It is ready for review (I though Thomas reviewing was enough but if anyone else wants to review), you can pass it to DRAFT if you want.

TotoGaz:
> It is ready for review (I though Thomas reviewing was enough but if anyone else wants to review), you can pass it to DRAFT if you want.

It's OK, I spoke with @CusiniM to tell him 😉

src/coreComponents/schema/schemaUtilities.cpp
```diff
@@ -176,7 +176,8 @@ void SchemaConstruction( Group & group,
             // Enforce uniqueness of element names
             // Note: this must be done at the parent element level
             xmlWrapper::xmlNode uniqueNameNode = targetIncludeNode.append_child( "xsd:unique" );
-            string uniqueNameNodeStr = targetName + subName + "UniqueName";
+            string path = std::regex_replace( group.getPath(), std::regex( "/" ), "" );
+            string uniqueNameNodeStr =  path + subName + "UniqueName";
```
TotoGaz, line 180: @cssherman What's your take on this snippet?

TotoGaz:
I've run the integrated tests and they are OK.
Of course, some keywords are changed, but it was expected.

TotoGaz (approved):
I'll work a bit on the integrated tests to be sure nothing's wrong, and then I guess this can be merged!

src/coreComponents/mesh/generators/PartitionDescriptor.hpp
```diff
@@ -0,0 +1,96 @@
...
+   */
+  std::set< int > const & getMetisNeighborList() const { return m_metisNeighborList; }
+
+  /**
+   * @brief Sets the list of metis neighbor list.
+   * @param metisNeighborList A reference to the Metis neighbor list.
+   */
+  void setMetisNeighborList( std::vector< int > const & metisNeighborList )
+  {
+    m_metisNeighborList.insert( metisNeighborList.cbegin(), metisNeighborList.cend() );
```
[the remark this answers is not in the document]
XL64: Added the clear()

untereiner:
Would a message to end-user saying that from now an InternalWell must be a child of a Mesh useful ? (For me it would have been)

TotoGaz:
>

TotoGaz:
> * Would a message to end-user saying that from now an InternalWell must be a child of a Mesh useful ? (For me it would have been)

Yes, if the fwk currently allows it conveniently, that's good.

> * Why keeping `meshName` in InternalWell if a well must be a child of a Mesh

It's surely a legacy move that was not challenged during the refactoring.
@XL64 Can you challenge the `WellGeneratorBase::viewKeyStruct` to see if
- they are all still used?
- they belong in `WellGeneratorBase` or would better fit `InternalWellGenerator`?

@untereiner thanks for your comments. Can you use the review features of GH so we better know what you're referring to?

BTW, I've run the integrated tests on quartz and despite new keyword (as expected), the results are OK.
I guess we can get this merged very soon! 🎉

src/coreComponents/mesh/WellElementRegion.hpp
```diff
@@ -121,17 +121,17 @@ class WellElementRegion : public ElementRegionBase
   /**
    * @brief Not implemented, this task is performed in GenerateWell.
    */
-  virtual void generateMesh( Group & ) override {}
```
untereiner, line 124: Another thing if I am not wrong , the rule is:  Use override (only) for a derived class' override to help maintenance. So virtual isn't necessary here.

untereiner (approved):

TotoGaz (approved):

src/coreComponents/mesh/generators/LineBlock.hpp
```diff
@@ -0,0 +1,259 @@
...
+  array2d< globalIndex > m_elemToNodesMap;
+
+  /// Volume of well elements
+  array1d< real64 > m_elemVolume;
+
+
+  // well node data
+
+  /// Number of nodes per well element
+  globalIndex m_numNodesPerElem;
```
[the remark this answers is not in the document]
untereiner: they are not used

src/coreComponents/mesh/generators/LineBlock.hpp
```diff
@@ -0,0 +1,259 @@
...
+
+  /// Global index of the well element
+  array1d< globalIndex > m_perfElemId;
+
+
+
+  // Auxiliary data
+
+  // Number of physical dimensions
+  int m_nDims;
```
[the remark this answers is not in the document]
untereiner: It is not used

XL64:
> * Would a message to end-user saying that from now an InternalWell must be a child of a Mesh useful ? (For me it would have been)
> 
>     * Why keeping `meshName` in InternalWell if a well must be a child of a Mesh

Hello, removed the meshName. For the error message I understand it is not clear but I think it's outside the scope of this PR. It's a more general issue with objectCatalog error message understandability.

src/coreComponents/mesh/WellElementRegion.hpp
XL64: Changed it

src/coreComponents/mesh/generators/LineBlock.hpp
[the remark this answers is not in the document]
XL64: Done

paveltomin:
any way to decouple mesh and well definitions after this PR?
if one want to put mesh definition in one file and wells definition in another, what should be the way?

untereiner:
Include an xml file with wells in the vtkmesh node ?

paveltomin:
i meant whole well definition solver+geometry:
```
<Problem>
	<Solvers gravityVector="{0.0, 0.0, -9.81}">
		<CompositionalMultiphaseWell name="wellSolver"
		                             useMass="1"
		                             logLevel="1"
		                             writeCSV="1"
		                             targetRegions="{ WellRegion }">
			<NonlinearSolverParameters logLevel="1"/>
			<WellControls name="wellControl"
			              logLevel="1"
			              type="injector"
			              useSurfaceConditions="1"
			              surfacePressure="1.013250e+05"
			              surfaceTemperature="288.71"
			              control="totalVolRate"
			              targetPhaseName="gas"
			              injectionTemperature="366.483"
			              injectionStream="{ 1.0, 0.0 }"
			              enableCrossflow="0"
			              targetBHP="4e7"
			              targetTotalRate="1000"
			              referenceElevation="-2200.0"/>
		</CompositionalMultiphaseWell>
	</Solvers>
	<Mesh>
		<VTKMesh name="mesh"
		         logLevel="5"
		         file="base.vtu"
		         regionAttribute="CellEntityIds"
		         partitionRefinement="3"
		         fieldsToImport="{ PORO, PERM, DENSITY, bulkM, shearM }"
		         fieldNamesInGEOS="{ rockPorosity_referencePorosity, rockPermeability_permeability, rock_density, rock_bulkModulus, rock_shearModulus }">
			<InternalWell name="well"
			              wellRegionName="WellRegion"
			              wellControlsName="wellControl"
			              radius="0.1"
			              polylineNodeCoords="{
							{ -4105, -32745, -1730 },
							{ -4105, -32745, -2600 }
			              }"
			              polylineSegmentConn="{ {0,1} }"
			              numElementsPerSegment="1"
			              logLevel="2">
				<Perforation name="well_perf"
				             distanceFromHead="500"/>
			</InternalWell>
		</VTKMesh>
	</Mesh>
</Problem>
```

untereiner:
You can’t 
As side note, I wasn’t fan of this change. I didn’t understand the motivations

paveltomin:
Ok, thanks for confirmed
The change looked logical at that time but brings this input inconvenience because it's quite common practice to have all well-related info into a separate file

untereiner:
In terms of conceptual modeling, well under mesh means « has a hierarchical relationship » which is not true. A well and a mesh are « related » to each other it’s a loose coupling. I would be in favor to undo this change.

PR #2278, opened 2023-02-06
add standard install locations 

untereiner:
- add GNUInstallDirs
- install examples / scripts / schema to CMAKE_INSTALL_DATAROOTDIR (i.e. read-only architecture-independent data root (share))


src/CMakeLists.txt
```diff
@@ -169,8 +170,10 @@ endif()
...
 install( DIRECTORY ../examples/
-         DESTINATION examples )
+         DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/${CMAKE_PROJECT_NAME}/examples )
 
 install( DIRECTORY ../scripts/
-         DESTINATION scripts )
+         DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/${CMAKE_PROJECT_NAME}/scripts )
 
+install( FILES ${SCHEMA_DIR}/schema.xsd
+         DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/${CMAKE_PROJECT_NAME}/schema )
```
TotoGaz, line 179: The creation of `schema.xsd` is done iff `ENABLE_XML_UPDATES AND ENABLE_MPI AND UNIX AND NOT CMAKE_HOST_APPLE AND NOT ENABLE_CUDA` evaluates to true. Otherwise we're using the last existing version of `schema.xsd` (yes, we commit generated code and I do agree that we shouldn't...).

Can you manage to reunite the creation of `schema.xsd` and its installation. I do not think we want a mismatch between `geosx` and its scheme.

TotoGaz (commented):
`CMAKE_INSTALL_DATAROOTDIR` is definitely better!

src/CMakeLists.txt
TotoGaz: The root of the issue is that, yes, we commit the _generated_ file `schema.xsd`, and I claim that it's a bad pattern (I think mostly driven by some `readthedocs` deployment considerations).

As such, there will always be a moment where the `schema.xsd` will be out of sync (Because someone will have forgotten to commit, for example. Or because the dev won't have `ENABLE_XML_UPDATES` set to `true` to save time. Or because some `CLion` remote working dev pattern won't update `schema.xsd` in your work tree, but only on the remote and you may not notice).

To enforce sync, we should only deploy the `schema.xsd` that has been generated by the `geosx` binary itself.
That way, our end-users will always have something reliable.

klevzoff: In that case any deployment of GEOSX on a GPU system will not have a schema installed... which is probably not what we want? Generation is only disabled with CUDA to avoid producing diffs on LvArray type names (something developers care about), but users need the schema regardless (whether freshly generated or pre-committed) for input validation. I agree that neither solution is perfect and we should ideally improve the schema generation to not produce diffs between systems/configurations.

TotoGaz: > In that case any deployment of GEOSX on a GPU system will not have a schema installed... which is probably not what we want?

What I do not want is that end-users get provided _wrong_ information. It's worse than everything, because they may be building their workflow on a brittle tool, and this always gets back to us as a broken feature and we have to fix it quick. And we do not have the resources for this.
Also, having brittle side-tools can ruin the confidence users would place in `geosx`. I do not think we need this.

> Generation is only disabled with CUDA to avoid producing diffs on LvArray type names (something developers care about), but users need the schema regardless (whether freshly generated or pre-committed) for input validation.

Then if the `schema.xsd` gets generated nevertheless in the installation folder but not into the `git` working tree, this should be fine? End-users would not care about the `LvArray` discrepancy? And this would generate no diff in `git`.

I'm simply saying that the committed `schema.xsd` cannot be considered as a trustworthy source of information.
I'd like `geosx` to produce it for consistency.
As such I'm fine having an installation step which basically does `geosx -s > /path/to/install/schema.xsd`.

> I agree that neither solution is perfect and we should ideally improve the schema generation to not produce diffs between systems/configurations.

I think that the whole generated-doc-in-the-git pattern should be revised.
If the only reason why we do this is `readthedocs`, then we should bundle the generated doc and get `readthedocs` download and deploy it.

untereiner: What I agree with you is that's a bad pattern to commit files that can be generated. 
For me it is to a specific CI step to generate the documentation files and push them into a `documentation` git repository for `readthedocs` to find its files. I have each time a bunch of rst files to commit without really understanding the reason. It gets the commits messy IHMO.

For the time being I will follow your idea: add a specific cmake target that generates always the schema to install it

src/CMakeLists.txt
```diff
@@ -129,9 +130,10 @@ set_target_properties( geosx PROPERTIES INSTALL_RPATH_USE_LINK_PATH TRUE )
 
 install(TARGETS geosx RUNTIME DESTINATION ${CMAKE_INSTALL_PREFIX}/bin)
 
+set(SCHEMA_DIR ${CMAKE_SOURCE_DIR}/coreComponents/schema)
```
TotoGaz, line 133: Why do you extract `SCHEMA_DIR` from the next `if` scope?
It looks to me that it's not used outside of this scope?

src/CMakeLists.txt
```diff
@@ -169,8 +171,18 @@ endif()
...
-
+         DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/${CMAKE_PROJECT_NAME}/scripts )
+
+add_custom_target( geosx_generate_install_schema
+         ALL
+         COMMAND geosx -s ${CMAKE_BINARY_DIR}/schema.xsd >generate_schema.log 2>&1 || (cat generate_schema.log && exit 1)
+         WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
+         DEPENDS geosx
+         COMMENT "Generating XML schema to install"
+         )
```
TotoGaz, line 185: I confess I was kind of fearing this: we do not have any GPU on travis so the CUDA drivers are not installed and `GEOSX` cannot be run:
```
[1002/1083] Generating XML schema to install
FAILED: CMakeFiles/geosx_generate_install_schema /tmp/build/CMakeFiles/geosx_generate_install_schema 
cd /tmp/build && /tmp/build/bin/geosx -s /tmp/build/schema.xsd >generate_schema.log 2>&1 || ( cat generate_schema.log && exit 1 )
GEOSX version: 0.2.0
  - c++ compiler: clang 10.0.0
  - cuda compiler version: 11.2
  - MPI version: Open MPI v4.0.3, package: Debian OpenMPI, ident: 4.0.3, repo rev: v4.0.3, Mar 03, 2020
  - HDF5 version: 1.12.1
...
CUDA ERROR (code = 35, CUDA driver version is insufficient for CUDA runtime version) at general.c:253
```
As options, we could
- Add a `cmake` option to control this `cmake` step. That implies to update `travis` as well, and I'm not sure this will be seamless to change the option in the `host-config`s only on `travis`.
- Ignore the error code (not my favorite option, obviously).
- Check for the `TRAVIS` environment variable to be present and set to `true` (see [here](https://docs.travis-ci.com/user/environment-variables/#default-environment-variables)), and deploy the schema accordingly.
- Define a `cmake` option to control this step, set to `true` by default for everyone, but overwritten to `false` if `TRAVIS` exists and is set to `true`.

If we chose the `cmake` option way, we'll surely have to update the doc as well.
(Not a big deal, but still needs to be done).

src/CMakeLists.txt
untereiner: you are true, previous attempt I didn't rollback

src/CMakeLists.txt
untereiner: This problem is CI related. We should not put a cmake ghost argument, CI specific, to resolve this CI issue IHMO.
What do you think of passing the option `--build-exe-only` to the `travis_build_and_test.sh` script in case of GPU related builds ? (as the green target `Ubuntu CUDA debug (20.04, clang 10.0.0 + gcc 9.4.0, open-mpi 4.0.3, cuda-11.2.152)`)

TotoGaz: It's obviously better to have the options properly defined in the build.
Here the option would be some `--deploy-scheme`. If you manage to make it happen, that's great.

I do not think we can rely on `--build-exe-only`: knowing that the tests can be built is important, even if we can't run them.

untereiner: Do you mean that the tests are build with option `--build-exe-only` ? If I understand the CMakeLists correctly they are build with the `all` target. 
What you want is to be able to (compile only the executable or executable and tests) and choose to install or not the schema as a separate step. Default behavior would install the schema. That's not so easy to implement only passing arguments to cmake to configure step.

scripts/travis_build_and_test.sh
```diff
@@ -58,6 +64,7 @@ or_die python3 scripts/config-build.py \
                -ip ${GEOSX_DIR} \
                --ninja \
                -DBLT_MPI_COMMAND_APPEND='"--allow-run-as-root;--oversubscribe"'
+               -DGEOSX_INSTALL_SCHEMA=${GEOSX_INSTALL_SCHEMA}
```
TotoGaz, line 67:
```suggestion
               -DBLT_MPI_COMMAND_APPEND='"--allow-run-as-root;--oversubscribe"' \
               -DGEOSX_INSTALL_SCHEMA=${GEOSX_INSTALL_SCHEMA}
```

scripts/travis_build_and_test.sh
```diff
@@ -92,4 +99,6 @@ if [[ "$*" != *--disable-unit-tests* ]]; then
   or_die ctest --output-on-failure -E "testUncrustifyCheck|testDoxygenCheck"
 fi
 
+
+
```
TotoGaz, line 103: ?

src/CMakeLists.txt
```diff
@@ -169,8 +170,22 @@ endif()
...
+
+if (NOT GEOSX_INSTALL_SCHEMA)
```
TotoGaz, line 186: Can you update the doc appropriately [here](https://geosx-geosx.readthedocs-hosted.com/en/latest/docs/sphinx/buildGuide/BuildProcess.html#configuration-options)?
Plus (I do not really know), but is there a place where we define the `default` values of our variables?

.travis.yml
```diff
@@ -187,14 +187,14 @@ jobs:
     env:
     - DOCKER_REPOSITORY=geosx/centos7.6.1810-gcc8.3.1-cuda10.1.243
     - CMAKE_BUILD_TYPE=Release
-    - BUILD_AND_TEST_ARGS="--disable-unit-tests --reduce-install-logs"
+    - BUILD_AND_TEST_ARGS="--disable-unit-tests --reduce-install-logs --disable-schema-deployment"
```
TotoGaz, line 190: Passing by, can you remove `--reduce-install-logs` which looks like it's become pretty useless...

src/CMakeLists.txt
untereiner: `GeosxOptions.cmake` ?

.travis.yml
```diff
@@ -187,7 +187,7 @@ jobs:
     env:
     - DOCKER_REPOSITORY=geosx/centos7.6.1810-gcc8.3.1-cuda10.1.243
     - CMAKE_BUILD_TYPE=Release
-    - BUILD_AND_TEST_ARGS="--disable-unit-tests --reduce-install-logs --disable-schema-deployment"
+    - BUILD_AND_TEST_ARGS="--disable-unit-tests --disable-schema-deployment"
```
TotoGaz, line 190: I think you forgot to remove https://github.com/GEOSX/GEOSX/blob/b6b0e875d1b55bca361497bb572ef83a25c31f65/scripts/travis_build_and_test.sh#L83-L87 which lead me to `--reduce-install-logs` being dummy.

scripts/travis_build_and_test.sh
```diff
@@ -99,6 +97,4 @@ if [[ "$*" != *--disable-unit-tests* ]]; then
   or_die ctest --output-on-failure -E "testUncrustifyCheck|testDoxygenCheck"
 fi
 
-
-
-exit 0
+exit 0
```
TotoGaz, line 100:
```suggestion
exit 0

```

scripts/travis_build_and_test.sh
```diff
@@ -37,12 +37,10 @@ if [[ -z "${GEOSX_DIR}" ]]; then
   exit 1
 fi
 
-GEOSX_INSTALL_SCHEMA=1
 if [[ "$*" == *--disable-schema-deployment* ]]; then
-  GEOSX_INSTALL_SCHEMA=0
+  GEOSX_INSTALL_SCHEMA=0
```
TotoGaz, line 41:
```suggestion
  GEOSX_INSTALL_SCHEMA=0
```

src/docs/doxygen/GeosxConfig.hpp
```diff
@@ -114,10 +114,10 @@
 /* #undef chai_VERSION */
 
 /// Version information for adiak
-#define adiak_VERSION ..
+/* #undef adiak_VERSION */
```
TotoGaz, line 117: Please rollback this file. I know it's generated automatically... Sorry for this.

.travis.yml
untereiner: yes

src/docs/doxygen/GeosxConfig.hpp
untereiner: Have'nt see it. should avoit git commit -a

scripts/travis_build_and_test.sh
```diff
@@ -57,7 +61,8 @@ or_die python3 scripts/config-build.py \
                -bp ${GEOSX_BUILD_DIR} \
                -ip ${GEOSX_DIR} \
                --ninja \
-               -DBLT_MPI_COMMAND_APPEND='"--allow-run-as-root;--oversubscribe"'
+               -DBLT_MPI_COMMAND_APPEND='"--allow-run-as-root;--oversubscribe"' \
+               -DGEOSX_INSTALL_SCHEMA=${GEOSX_INSTALL_SCHEMA}
```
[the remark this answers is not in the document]
untereiner: I initialized the variable with 1 so there are `-DGEOSX_INSTALL_SCHEMA=0` and `-DGEOSX_INSTALL_SCHEMA=1`

TotoGaz (approved):

Issue #2227, opened 2022-12-21
Well perforation issue : Invalid mapping


margauxraguenel:
**Describe the bug**
I'm trying to launch GEOSX on a real application case with one well. 
As preprocessing, I have checked all the cells concerned by the well and compute the cell centers: 
![image](https://user-images.githubusercontent.com/32702400/208930097-8061b391-3eeb-4c5f-a5c9-5a081d4a74e7.png)
These coordinates are used in the xml definition and well interpreted by GEOSX for the perforation coordinates:
![image](https://user-images.githubusercontent.com/32702400/208929869-b9b875f1-9a69-43d0-8813-7a39d7a0f475.png)

However, I got the following error:
![image](https://user-images.githubusercontent.com/32702400/208930511-103ce6b8-7d3e-4ec6-a62f-8721e1874552.png)

Additional information: 
No issue in sequential. 

**Platform (please complete the following information):**
- Machine P2
- Compiler: gcc 8.3.0
- GEOSX Version 88f9773

**Additional context**
Solution proposed in Issue #2205 does not work, a more recent version of GEOSX has been used.

@jeannepellerin 

duplicate issue #2256

dkachuma:
I have recreated the situation in the attached model 
[issue2227.tar.gz](https://github.com/GEOSX/GEOSX/files/10421877/issue2227.tar.gz). The mesh in this model is a single element vtu in which the single element is a pyramid with all 5 points having $z>0$. There is a well with a single segment whose element centre is well below 0 at $z=-105$. As such I would expect GEOSX to fail to find the intersection of this well with the mesh and issue an error. However, GEOSX runs fine and attaches the well to the single element.

This is because the pyramid has a base whose points are not coplanar. As such, in `computationalGeometry::isPointInsidePolyhedron`, the 5 face centres are calculated as
```
(-66.7,   0.0, 60.0)
(  0.0,  66.7, 60.0)
(  0.0, -66.7, 60.0)
( 66.7,   0.0, 60.0)
(  0.0,   0.0, 53.3)
```
while the element centre is `(0.0, 0.0, 52.0)`. That is, all face centres are above the element centre. As such, the computation fails to determine that the point `(0.0, 0.0, -105.0)` is actually outside the element.

untereiner:
I think the normal computed in `centroid_3DPolygon` of `ComputationalGeometry.hpp` is wrong.
[...]

PR #2280, opened 2023-02-07
Compute normal with newell's method

untereiner:


dkachuma:
I submitted a similar [issue against vtk (18787)](https://gitlab.kitware.com/vtk/vtk/-/issues/18787). They resolved it with the same algorithm. However, they had a concern about numerical stability when the points are too far from the origin. Should we translate the polygon closer to the origin before performing the calculation?

TotoGaz:
@dkachuma @francoishamon Do you think this could require a rebaseline?

francoishamon:
> @dkachuma @francoishamon Do you think this could require a rebaseline?

Let's me check, I will let you know

herve-gross:
@margauxraguenel let us know if this PR fixes bug #2227

src/coreComponents/mesh/utilities/ComputationalGeometry.hpp
```diff
@@ -207,29 +207,37 @@ real64 centroid_3DPolygon( arraySlice1d< localIndex const > const pointsIndices,
...
-    real64 const triangleArea = LvArray::tensorOps::l2Norm< 3 >( triangleNormal );
+  // close the cycle
+  LvArray::tensorOps::copy< 3 >( current, next );
+  LvArray::tensorOps::copy< 3 >( next, points[ pointsIndices[ 0 ] ] );
 
-    LvArray::tensorOps::add< 3 >( normal, triangleNormal );
+  real64 vc[3];
+  vc[0] = (current[1] - next[1]) * (current[2] + next[2]);
+  vc[1] = (current[2] - next[2]) * (current[0] + next[0]);
+  vc[2] = (current[0] - next[0]) * (current[1] + next[1]);
```
TotoGaz, line 233: Could the cycle be closed in the loop by using `( a + 1 ) % ( pointsIndices.size() )` instead of `a + 1`?
And then changing the loop boundaries into
```
for( localIndex a=0; a<pointsIndices.size(); ++a )
```
That could remove some duplicated code.
`current` and `next` could be enclosed in their scope too.

Surely having some `int const numPoints = pointsIndices.size()` could help.

TotoGaz (commented):
Do you have a reference for the algorithm?
@jeannepellerin if you know the algo, maybe you can review it?

francoishamon:
> @dkachuma @francoishamon Do you think this could require a rebaseline?

Yes, there is a rebaseline of all the integrated tests needed here. `elementCenter` and `faceCenter` for instance have large diffs, so we will have to be careful during the rebaseline.

dkachuma:
> `faceCenter`

The change to `faceCenter` is what really fixes the issue in this case.

untereiner:
@TotoGaz https://www.sciencedirect.com/science/article/pii/B978008050755250052X?via%3Dihub

jeannepellerin:
@dkachuma Do you think that the test that are failing on some platform are the result of numerical errors ?

src/coreComponents/mesh/utilities/ComputationalGeometry.hpp
```diff
@@ -207,29 +207,37 @@ real64 centroid_3DPolygon( arraySlice1d< localIndex const > const pointsIndices,
...
+
+  area = LvArray::tensorOps::l2Norm< 3 >( normal );
+  LvArray::tensorOps::scale< 3 >( center, 1.0 / real64( pointsIndices.size()) );
 
-    area += triangleArea;
-    LvArray::tensorOps::scaledAdd< 3 >( center, vc, triangleArea );
-  }
   if( area > areaTolerance )
   {
     LvArray::tensorOps::scale< 3 >( center, 1.0 / ( area * 3.0 ) );
```
dkachuma, line 243: This line is now incorrect because the face centre is simply calculated as an average of the face points not as a weighted sum of triangles.

margauxraguenel:
Hello,
I tested the PR on the case mentionned in bug #2227 and it fixes the issue.
Thank you all for the work.

untereiner:
@TotoGaz The tests doesn't have large diffs anymore... Does it require a rebaseline nevertheless ? What can I do to move this work forward ?

TotoGaz:
> @TotoGaz The tests doesn't have large diffs anymore... Does it require a rebaseline nevertheless ? What can I do to move this work forward ?

Yes it will.
@dkachuma can you review this PR?

dkachuma (commented):
Can we add a unit test?

dkachuma:
> 

One of the issues is that there might be numerical stability issues when the polygon is far from the origin. I just wanted to check that this still works if the values are very large. What I had was a symmetrical 6-pointed star which has the potential for terms to cancel out and give a zero normal (see [vtk issue 18787](https://gitlab.kitware.com/vtk/vtk/-/issues/18787)). Make sure the calculated normal is still meaningful if this is polygon is located far away.

untereiner:
@dkachuma What do think of this test ?

dkachuma (approved):

francoishamon:
Hello @untereiner,

I ran the integrated tests for your PR. The numerical diffs are attached. Could you please have a look and let us know if they are what you expect? Just to make sure before merging the PR.

Thanks

[integratedTestsLionel.txt](https://github.com/GEOSX/GEOS/files/11143164/integratedTestsLionel.txt)

untereiner:
I would say yes but I'm not sure of my answer. @dkachuma do you think the diffs are in an acceptable range ?

dkachuma:
> I would say yes but I'm not sure of my answer. @dkachuma do you think the diffs are in an acceptable range ?

Most of the diffs appear to be related to the faceCenter itself as expected.

However there's quite a large diff in the stress and pressure on `DeviatedPoroElasticWellbore_Injection/DeviatedPoroElasticWellbore_Drilling`. It might be worth exploring why a change in the face centre in these models would cause such a large change in the pressure. Let me run these models and look at them a bit more.

untereiner:
thank you @dkachuma

Issue #2287, opened 2023-02-08
Build issue related to missing braces under if statements


sytuannguyen:
Describe the bug
The build is blocked by an error message as shown below:

![image](https://user-images.githubusercontent.com/45685596/217412936-b0996c46-fef0-42db-8ad7-08213ed6ff10.png)

It is verified that this bug is purely related to the missing braces under two `if` statements in `lifoStorage.hpp`.

To Reproduce
Build actual version of GEOSX in TotalEnergies XRAI machine with default gcc version 10.2.0.

Expected behavior
Success build.

Screenshots

Platform (please complete the following information):

Machine [e.g. TotalEnergies/XRAI]
Compiler: [gcc 10.2.0]
GEOSX Version [#e5b513c]
Additional context

PR #2288, opened 2023-02-08
Fix build issue #2287

sytuannguyen:
This tiny PR aims to fix the build issue #2287 related to missing braces under two `if` statements


klevzoff (commented):
Note that both old and the new versions of the `LIFO_MARK_` do not function as intended. `GEOSX_MARK_` macros create a local object that uses constructor/destructor to enter/exit the timed region. If the object's lifetime is put under a conditional, it will only measure the duration of the branch (which is contains no other statements and thus will measure to ~0) and not the outer scope (which I assume this is intended for).

If we really need to conditionally enable timers at runtime, we'd have to either use procedural-style `GEOSX_MARK_BEGIN`/`GEOSX_MARK_END` macros, or (better) wrap those in a struct with constructor/destructor calling them conditionally.

sframba:
This is a valid remark? @XL64 do you think this needs changing, or was the `LIFO_MARK_` only intended to be a temporary feature to profile the LIFO performance?

TotoGaz:
I glanced at the `lifoStorage.hpp` and I'm not sure it makes sense to call `env` every time you emplace or pop something.
I bet the `LIFO_*` should simply be removed.

Plus I could spot a lot of `#ifdef GEOSX_USE_CUDA` adding new attributes or even this pattern 
```c++
#ifdef GEOSX_USE_CUDA
if ( condition ) {
    ...
} else
#endif
{
    ...
}
```
which is at least questionable.
I did not look into the details but I would vote for
```mermaid
classDiagram
LifoStorageBase <|-- LifoStorageCPU
LifoStorageBase <|-- LifoStorageGPU
```
with the appropriate amount of `protected` and maybe CRTP (I don't know if this is mandatory).
Each class in its own file, the `LifoStorageGPU` being only considered for GPU by `cmake`.
Then some last 
```c++
#ifdef GEOSX_USE_CUDA
template< typename T >
class LifoStorage = LifoStorageGPU< T >;
#else
template< typename T >
class LifoStorage = LifoStorageCPU< T >;
#endif
```
Maybe there's more than one class involved but still.

By the way, the classes in the `lifoStorage.hpp` don't even respect [the naming conventions](https://geosx-geosx.readthedocs-hosted.com/en/latest/docs/sphinx/developerGuide/Contributing/CodeStyle.html#class-struct-names).

So this file requires some work, yes.

W.r.t. the current PR, surely can we accept the fix to unlock @sytuannguyen .

sframba:
@TotoGaz makes sense to me, let's wait for @XL64's take on this. I In the meantime, there is no issue on our side about integrating @sytuannguyen 's solution and fix the build issue.

sytuannguyen:
> Hello @sytuannguyen , when the `check_pr_is_assigned` check is red image It's because the PR is not assigned. So you have to assign yourself like @francoishamon did.
> 
> Plus you have to link your PR to the issue on the right pane here image

@TotoGaz Thank you very much!

TotoGaz (approved):

XL64:
Sorry didn't see that all, I'll remove the if indeed. I'll look at the remarks to, I have some corrections to do on it anyway.

Issue #2290, opened 2023-02-13
Improve boundary condition error checking


rrsettgast:
**Describe the issue**
1. It is currently possible for a user to specify a boundary condition on a field that does not exist (i.e. "TotalDisplacement"), and the code will proceed without notifying the user that their boundary condition is invalid.
2. It is possible to specify a boundary condition on an empty set without warning. This should be allowed, but the user should see a warning.
3. Certain physics require that specific boundary conditions are applied to ensure a well posed problem. For example, quasi-static mechanics problems require kinematic boundary conditions that remove the rigid body modes.

**Proposed cleanup**

- [X] add error for boundary condition on a field that does not exist.
- [X] add warning for boundary condition applied to empty set.
- [ ] add sanity check for physics implementations that allow for xml level check....maybe hard to ensure that B.C.'s are sufficient to lead to a non-singular system matrix.

PR #2291, opened 2023-02-13
Added a validation function for boundary conditions

francoishamon:
Following up last week's discussion, I had started something. I submit it for discussion in this PR but it is just a proposal.

This PR adds a function to validate the boundary conditions (at initialization) in the `FieldSpecificationManager`. 
It works as follows:

1) If a `FieldSpecification` is requested on a field that does not exist, we have the following error message:
```
***** Rank 0: xconstraint: there is no fieldName named `TotalDisplacement` under this objectPath.                                                                                        
`TotalDisplacement` cannot be used to set up a FieldSpecification, check the XML input                                                                                                   
```
and the simulation stops

2) If a `FieldSpecification` is requested on a set that does not exist, we have the error message:
```
bulkModulus: there is no set named `alll` under this ElementRegions/underburden at the end of initialization.                                                                            
```
and the simulation stops

3) If a `FieldSpecification` is requested on a set that exists and is empty, we have the warning message:
```
sourceTerm: this FieldSpecification targets (an) empty set(s), check the content of the set `well` in `ElementRegions/reservoir`.                                                        
  -If `ElementRegions/reservoir` is in ElementRegions, the set(s) must contain at least an element                                                                                       
  -If `ElementRegions/reservoir` is faceManager, the set(s) must contain at least a face                                                                                                 
  -If `ElementRegions/reservoir` is nodeManager, the set(s) must contain at least a node 
```
and the simulation proceeds

I encountered the following difficulties:
- Some face boundary conditions target cell-based fields (in the flow solvers), which makes it difficult to validate the field names
- The `FieldSpecification::apply` function relies on the fact that `SurfaceGenerator`-related `FieldSpecification` are silently ignored at initialization, which is orthogonal to checking the boundary conditions
- The proposed method ignores the `startTime` and `endTime` of the BCs and does not have any notion of the solver's target regions

Related points 1 and 2 in #2290 

Rebaseline done in https://github.com/GEOSX/integratedTests/pull/302


src/coreComponents/fieldSpecification/FieldSpecificationManager.cpp
```diff
@@ -64,6 +64,142 @@ void FieldSpecificationManager::expandObjectCatalogs()
...
+
+      // 2.a) If we enter this loop, we know that the set has been created
+      //      Fracture/fault sets are created later and the "apply" call silently ignores them
+      isTargetSetCreated.at( setName ) = 1;
+
+      // 2.b) If the fieldName is registered on this target, we record it
+      //      Unfortunately, we need two exceptions:
+      //       - FieldSpecification that do not target a field, like Aquifer, Traction, Equilibrium, etc. For these, the check is not
+      // necessary (the user cannot mess up)
+      //       - Face boundary conditions that target cell-based quantities, like the face BC of the flow solvers
```
rrsettgast, line 106: should we exclude blank `fieldName` from the check? Maybe we should have all the special cases as a derived `FieldSpecification` instead of using the `fieldName` to specify a different type of BC?

see comment in [CompositionalMultiphaseBase.cpp](https://github.com/GEOSX/GEOSX/pull/2291/files#diff-a3f511da5cf9cbca1b89b3f09ed486c4928bf6e7a4fe736bb6de427a6a6b98ba)

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -1333,7 +1333,7 @@ void CompositionalMultiphaseBase::applySourceFluxBC( real64 const time,
   {
     fsManager.apply< ElementSubRegionBase >( time + dt,
                                              mesh,
-                                             FieldSpecificationBase::viewKeyStruct::fluxBoundaryConditionString(),
+                                             SourceFluxBoundaryCondition::catalogName(),
```
rrsettgast, line 1336: so instead of this, we would template the `FieldSpecification` type to `fsManager.apply<SourceFluxBoundaryCondition>()`. Would that be cleaner?

rrsettgast (commented):
some thoughts I had reading through this...

inputFiles/compositionalMultiphaseFlow/benchmarks/isothermalLeakyWell/isothermalLeakyWell_base_direct.xml
```diff
@@ -195,20 +195,6 @@
       fieldName="pressure"
       functionName="pressureFunction"
       scale="1"/>
-    <FieldSpecification
```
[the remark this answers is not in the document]
francoishamon: I found out that they were the sets `east`, `west`, `south`, `north` were empty while working on the PR, so I just removed them

src/coreComponents/fieldSpecification/FieldSpecificationManager.cpp
francoishamon: I think it would be a little bit cleaner, but we would still need to pass a `fieldName` like for the `Traction` [here](https://github.com/GEOSX/GEOSX/blob/3301d288bd10fb6f551c7e1766e12bb25bfe0eff/src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.cpp#L696-L699) or the `Aquifer` [here](https://github.com/GEOSX/GEOSX/blob/3301d288bd10fb6f551c7e1766e12bb25bfe0eff/src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.cpp#L778-L781) for instance.

A refactoring to facilitate checks and clarify the behavior could be as follows:

1. We don't rely on `FieldSpecificationBase` anymore to apply the field boundary conditions, but we create a new type (maybe just `FieldSpecification` inherited from `FieldSpecificationBase` or something better) to allow us to distinguish the "real" field applications (say, pressure, temperature, displacement) from the "complex" boundary conditions (`Traction`, `Aquifer`, etc). Potentially, we can do that without changing the xml.

2. We would have a) a `FieldSpecificationManager::apply` for the "real" field applications that would take the `fieldName` or even the extrinsic data struct and b) FieldSpecificationManager::apply` for the "complex" field applications that would not require a field name.

Using Points 1 and 2, we could also strictly check the field name in the validation step because we would have a mechanism to know when the `fieldName` is actually a field name and not the catalog name of the boundary condition class.

Then if we want to go further:

3. We add a mechanism to identify the face boundary conditions that do not target a field registered on the `FaceManager`. For instance, we would have a class `FlowFaceDirichletBoundaryCondition` inherited from `FieldSpecificationBase`, and we would know that for this type, we should check the field name on the element regions. This requires changing the xml. This is just a nice to have, because we already have a lot of custom checks and logs for the face BCs in the flow solvers.

4. I would be nice to enforce that all the `FieldSpecification`s that have `initialCondition="1"` can actually be applied at initialization, which is not the case when we have the `SurfaceGenerator` (they either have an empty set, or a not existent set). A typical example is [here](https://github.com/GEOSX/GEOSX/blob/3301d288bd10fb6f551c7e1766e12bb25bfe0eff/inputFiles/singlePhaseFlowFractures/fractureMatrixFlow_edfm_base.xml#L97). This is the main source of difficulties and it also what complicates the validation of the initial/boundary conditions. 

Using all this points, an added bonus is that we can simplify the content of the `apply` functions. For instance, we could throw an error if the set does not exist [here](https://github.com/GEOSX/GEOSX/blob/3301d288bd10fb6f551c7e1766e12bb25bfe0eff/src/coreComponents/fieldSpecification/FieldSpecificationBase.hpp#L138), we would not have to check for the empty field name [here](https://github.com/GEOSX/GEOSX/blob/3301d288bd10fb6f551c7e1766e12bb25bfe0eff/src/coreComponents/fieldSpecification/FieldSpecificationManager.hpp#L219), and we could remove the check on the initial condition [here](https://github.com/GEOSX/GEOSX/blob/3301d288bd10fb6f551c7e1766e12bb25bfe0eff/src/coreComponents/fieldSpecification/FieldSpecificationManager.hpp#L220). So, less extra validation would be needed and the behavior of the functions would be more transparent.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
francoishamon: Done, now it is `fsManager.apply< ElementSubRegionBase, SourceFluxBoundaryCondition >`

inputFiles/singlePhaseFlowFractures/fractureMatrixFlow_edfm_base.xml
```diff
@@ -94,13 +94,6 @@
   </Constitutive>
 
   <FieldSpecifications>
-    <FieldSpecification
-      name="frac"
-      initialCondition="1"
-      setNames="{ fracture }"
-      objectPath="faceManager"
-      fieldName="ruptureState"
-      scale="1"/>
```
francoishamon, line 103: Following our discussion on Tuesday, I removed some `ruptureState` `FieldSpecification` because they are indeed unused

inputFiles/hydraulicFracturing/walshQuarterNoChombo_base.xml
```diff
@@ -105,12 +105,6 @@
   </NumericalMethods>
 
   <FieldSpecifications>
-    <FieldSpecification name="waterDensity"
-               initialCondition="1"
-               setNames="{fracture}"
-               objectPath="ElementRegions"
-               fieldName="water_density"
-               scale="1000"/>
```
francoishamon, line 113: I removed this because it is unused

src/coreComponents/fieldSpecification/FieldSpecificationManager.cpp
```diff
@@ -64,6 +64,133 @@ void FieldSpecificationManager::expandObjectCatalogs()
...
+    }
+    bool areAllSetsMissing = true;
+    for( std::pair< string const, localIndex > & mapEntry : isTargetSetCreated )
+    {
+      mapEntry.second = MpiWrapper::max( mapEntry.second );
+      if( mapEntry.second == 1 )
+      {
+        areAllSetsMissing = false;
+      }
+    }
```
TotoGaz, line 146: Like above, maybe could you go the `mpi` "gathering" in a first loop, then analyze the result to set the value of `areAllSetsMissing` in a second loop, to better separate the two steps? It's a tad longer, but maybe it's a bit clearer, and with the `mpi` involved, one never knows. Up to you.

Would this be equivalent?
```suggestion
    for( std::pair< string const, localIndex > & mapEntry : isTargetSetCreated )
    {
      mapEntry.second = MpiWrapper::max( mapEntry.second );
    }
    bool areAllSetsMissing = true;
    for( std::pair< string const, localIndex > const & mapEntry : isTargetSetCreated )
    {
      if( mapEntry.second == 1 )
      {
        areAllSetsMissing = false;
        break;
      }
    }
```

src/coreComponents/fieldSpecification/FieldSpecificationManager.cpp
```diff
@@ -64,6 +64,133 @@ void FieldSpecificationManager::expandObjectCatalogs()
...
+      // necessary (the user cannot mess up)
+      //       - Face boundary conditions that target cell-based quantities, like the face BC of the flow solvers
+      if( targetGroup.hasWrapper( fieldName ) ||
+          flag == InputFlags::FALSE || // no need to check input if the input flag is false (Aquifer, Traction, Equilibrium do not target a
+                                       // field)
+          targetGroup.getName() == MeshLevel::groupStructKeys::faceManagerString() ) // the field names of the face BCs are not always
+                                                                                     // registered on
+                                                                                     // the faceManager...
+      {
+        isFieldNameFound = 1;
```
TotoGaz, line 114: It's a personal preference, but even if you're dealing with `integers` instead of `bool`, I'd stick to `true/false` when it comes to assigning.
```suggestion
        isFieldNameFound = true;
```
Still you to decide.

src/coreComponents/fieldSpecification/FieldSpecificationManager.cpp
```diff
@@ -64,6 +64,133 @@ void FieldSpecificationManager::expandObjectCatalogs()
...
+void FieldSpecificationManager::validateBoundaryConditions( MeshLevel & mesh ) const
+{
+  // loop over all the FieldSpecification of the XML file
+  this->forSubGroups< FieldSpecificationBase >( [&] ( FieldSpecificationBase const & fs )
+  {
+    localIndex isFieldNameFound = 0;
+    // map from set name to a flag (1 if targetSet not empty, 0 otherwise)
+    map< string, localIndex > isTargetSetNonEmpty;
+    // map from set name to a flag (1 if targetSet has been created, 0 otherwise)
+    map< string, localIndex > isTargetSetCreated;
```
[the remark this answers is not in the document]
francoishamon: Done, I now have `isTargetSetEmpty` to avoid the double negation

src/coreComponents/fieldSpecification/FieldSpecificationManager.cpp
francoishamon: Done

src/coreComponents/fieldSpecification/FieldSpecificationManager.cpp
```diff
@@ -64,6 +64,133 @@ void FieldSpecificationManager::expandObjectCatalogs()
...
+    } );
+
+    // Step 3: MPI synchronization
+
+    isFieldNameFound = MpiWrapper::max( isFieldNameFound );
+
+    bool areAllSetsEmpty = true;
+    for( std::pair< string const, localIndex > & mapEntry : isTargetSetNonEmpty )
+    {
+      mapEntry.second = MpiWrapper::max( mapEntry.second );
```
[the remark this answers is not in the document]
francoishamon: Currently, we cannot do MPI reductions for `bool`. 

The `bool` MPI type has been intentionally removed from our MPI wrapper function by @corbett5 in this commit: https://github.com/GEOSX/GEOSX/commit/4abbbda516bdd9f947a46b32e0b8e691fdb34f55#diff-c83e65f152397cfc121129cb04131b210e10bfd3da206c43e7f1e1407f9b8649

I asked @corbett5 and the reason for removing the `bool` type is that is not supported by all the MPI versions that we use. 

So, in the code added in this PR, I left the `integer` and did not use `bool`. I agree with you that it is less readable, but it is still good enough, I think

src/coreComponents/fieldSpecification/FieldSpecificationManager.cpp
```diff
@@ -64,6 +64,133 @@ void FieldSpecificationManager::expandObjectCatalogs()
...
+    GEOSX_LOG_RANK_0_IF( isFieldNameFound == 0 && areAllSetsEmpty,
+                         GEOSX_FMT( fieldNameNotFoundMessage,
+                                    fs.getName(), FieldSpecificationBase::viewKeyStruct::fieldNameString(),
+                                    fs.getFieldName(), FieldSpecificationBase::viewKeyStruct::objectPathString(), fs.getObjectPath(), fs.getFieldName() ) );
+    // if the field name was not found and some sets are not empty, the user misspelled the field name
+    GEOSX_THROW_IF( isFieldNameFound == 0 && !areAllSetsEmpty,
+                    GEOSX_FMT( fieldNameNotFoundMessage,
+                               fs.getName(), FieldSpecificationBase::viewKeyStruct::fieldNameString(),
+                               fs.getFieldName(), FieldSpecificationBase::viewKeyStruct::objectPathString(), fs.getObjectPath(), fs.getFieldName() ),
+                    InputError );
```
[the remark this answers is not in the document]
francoishamon: Yes, thanks, it is done

src/coreComponents/fieldSpecification/FieldSpecificationManager.cpp
francoishamon: Here, I left the code as is because of the limitations of our `MpiWrapper`s with `bool`.

src/coreComponents/unitTests/wavePropagationTests/testWavePropagationAcousticFirstOrder.cpp
```diff
@@ -116,14 +116,14 @@ char const * xmlInput =
   "    <FieldSpecification\n"
   "      name=\"cellVelocity\"\n"
   "      initialCondition=\"1\"\n"
-  "      objectPath=\"mesh/FE1/ElementRegions/Region/cb\"\n"
+  "      objectPath=\"ElementRegions/Region/cb\"\n"
```
francoishamon, line 119: @acitrain I changed the `objectPath` of your unit test because the `FieldSpecification` was silently ignored

acitrain: > @acitrain I changed the `objectPath` of your unit test because the `FieldSpecification` was silently ignored

Ok no problem @francoishamon. By this may be the reason why this test does not pass in @labesse40 last PR (that's weird but why not....) 
Thank you for noticing this !

src/coreComponents/fieldSpecification/FieldSpecificationManager.cpp
```diff
@@ -64,6 +64,133 @@ void FieldSpecificationManager::expandObjectCatalogs()
...
+    // if all sets are missing, we stop the simulation.
+    if( areAllSetsMissing )
+    {
+      // still loop over the map to get the set name
+      for( auto const & mapEntry : isTargetSetCreated )
+      {
+        GEOSX_THROW( GEOSX_FMT( "\n{}: there is no set named `{}` under the {} `{}`, check the XML input\n",
+                                fs.getName(), mapEntry.first, FieldSpecificationBase::viewKeyStruct::objectPathString(), fs.getObjectPath() ),
+                     InputError );
+      }
```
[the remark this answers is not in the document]
francoishamon: Now I collect all the set names, and I output an error message with all of them:
```
bottomInitialPressure: there is no set named `botom, bottomm` under the objectPath `ElementRegions/Channel`, check the XML input
```

francoishamon:
This PR is in good shape now, it is ready to be re-reviewed/merged if everyone likes the approach

acitrain (approved):

src/coreComponents/fieldSpecification/FieldSpecificationManager.cpp
```diff
@@ -64,6 +64,141 @@ void FieldSpecificationManager::expandObjectCatalogs()
...
+    for( std::pair< string const, localIndex > & mapEntry : isTargetSetEmpty )
+    {
+      mapEntry.second = MpiWrapper::min( mapEntry.second );
+    }
+    bool areAllSetsEmpty = true;
+    for( std::pair< string const, localIndex > & mapEntry : isTargetSetEmpty )
+    {
+      if( mapEntry.second == 0 ) // target set is not empty
+      {
+        areAllSetsEmpty = false;
```
TotoGaz, line 138: I think you can
```suggestion
        areAllSetsEmpty = false;
        break;
```
but not a big deal.

TotoGaz (approved):

src/coreComponents/fieldSpecification/FieldSpecificationManager.cpp
francoishamon: Sounds good, will do. Thanks!

francoishamon: Done

PR #2292, opened 2023-02-13
Negative time simulation for AcousticSEM 

labesse40:
This PR allows pygeosx users to start from negative time simulations without affecting GEOSX users.

Integrated test PR: https://github.com/GEOS-DEV/integratedTests/pull/322


src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEM.cpp
```diff
@@ -375,6 +384,7 @@ void AcousticWaveEquationSEM::initializePostInitialConditionsPreSubGroups()
 
     NodeManager & nodeManager = mesh.getNodeManager();
     FaceManager & faceManager = mesh.getFaceManager();
+    FaceManager & faceManager0 = this->getGroupByPath< FaceManager >( "/Problem/domain/MeshBodies/mesh/meshLevels/Level0/faceManager" );
```
acitrain, line 388: faceManager0 is not use anymore (it was an issue with high order which is fix now)

src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEM.cpp
```diff
@@ -226,12 +228,19 @@ void AcousticWaveEquationSEM::postProcessInput()
...
 
   localIndex const numSourcesGlobal = m_sourceCoordinates.size( 0 );
+  m_sourceNodeIds.resize( numSourcesGlobal, numNodesPerElem );
+  m_sourceConstants.resize( numSourcesGlobal, numNodesPerElem );
   m_sourceIsAccessible.resize( numSourcesGlobal );
 
   localIndex const numReceiversGlobal = m_receiverCoordinates.size( 0 );
+  m_receiverNodeIds.resize( numReceiversGlobal, numNodesPerElem );
+  m_receiverConstants.resize( numReceiversGlobal, numNodesPerElem );
   m_receiverIsLocal.resize( numReceiversGlobal );
```
acitrain, line 245: Same as before you can remove the arrays resized inside  initializePreSubGroups, norw it should work even with pygeosx as initializePreSubGroups is called in reinit so you could remove the duplicated resize (duplicated between initializePreSubGroups and postProcessInput) 
You can also remove "getNumNodesPerElem" call, it gives 0 here (found it last time...), it is anly useful in initializePreSubGroups

acitrain (approved):

PR #2294, opened 2023-02-13
Add checks to ensure that MPI_wait calls have completed prior to destruction of container.

rrsettgast:
This PR inserts some checks that `MPI_Request` contained in `MPI_iCommData` objects are completed before destruction of the `MPI_iCommData` object. For true asynchronous mpi, `MPI_iCommData` objects may have to either be ignored, or have a longer duration.


src/coreComponents/common/MpiWrapper.hpp
```diff
@@ -267,7 +258,7 @@ struct MpiWrapper
    * @return MPI_SUCCESS or and MPI_ERROR from internal calls to MPI_WaitAny.
    */
   static int activeWaitSomeCompletePhase( const int participants,
-                                          std::vector< std::function< MPI_Request ( int ) > > const & phases );
+                                          std::vector< std::tuple< MPI_Request *, MPI_Status *, std::function< void ( int ) > > > const & phases );
```
klevzoff, line 261: Maybe worth introducing a simple struct with named fields to replace tuple?

src/coreComponents/common/MpiWrapper.hpp
```diff
@@ -863,15 +854,16 @@ int MpiWrapper::iRecv( T * const buf,
                        MPI_Request * MPI_PARAM( request ) )
 {
 #ifdef GEOSX_USE_MPI
+  GEOSX_ERROR_IF( (*request)!=MPI_REQUEST_NULL && (*request)!=0,
```
klevzoff, line 857: What does the comparison of an `MPI_Request` handle with `0` do? I think that happens to work in both OpenMPI and MPICH due to chosen representations (pointer and int), but not guaranteed to be valid in general, since handle types are not standardized?

klevzoff (approved):

src/coreComponents/common/MpiWrapper.hpp
wrtobin: Yeah per the standard I think all we can guarantee is an `MPI_Request` handle is assignable the `MPI_REQUEST_NULL` value and comparable against that value, not that `0` is coercible to a comparable type / value.

rrsettgast: The intent is to check whether or not the `MPI_Request` is in use. i.e. the `MPI_Request` has been used in a non-blocking operation, but has not been "cleared" by an `MPI_Wait` etc. The completion of `MPI_Wait` means  `MPI_Request==MPI_REQUEST_NULL`. The `(*request)!=0` covers the case where the value has never been used.

src/coreComponents/common/MpiWrapper.hpp
rrsettgast: Yeah...I think we need something a little more structured in general. I want to re-organize the entirety async communication utilities.

src/coreComponents/common/MpiWrapper.hpp
rrsettgast: oh...is `MPI_Request` an `int` in the standard?

wrtobin: No per the standard it is an opaque handle.

wrtobin: So you can compare it to other `MPI_Request`s and `MPI_REQUEST_NULL`, but comparing it against anything else isn't necessarily well-defined. MPICH has it as an `int`, openmpi has it as a pointer, in general it could be anything.

rrsettgast: > So you can compare it to other `MPI_Request`s and `MPI_REQUEST_NULL`, but comparing it against anything else isn't necessarily well-defined. MPICH has it as an `int`, openmpi has it as a pointer, in general it could be anything.

im lazy

rrsettgast: OK. I removed that `0` check, and initialized all `MPI_Request` to `MPI_REQUEST_NULL`. It is a little heavy handed, but it is less error prone....but more incorrect usage prone.

src/coreComponents/common/MpiWrapper.hpp
rrsettgast: ....or to be precise, I want @wrtobin to re-organize it :)

PR #2305, opened 2023-02-21
Improve documentation for FaceBlockABC.

TotoGaz:
Bringing more precision about the mappings for 3d edges/2d faces.
Implicit index consistency also exposed.


CusiniM (approved):
I still need a few seconds to process the meaning of `2dFaceToEdge` but this improved documentation really helps.

TotoGaz:
> I still need a few seconds to process the meaning of `2dFaceToEdge` but this improved documentation really helps.

Thx for the review. I'll read all of this again, hopefully I'll be able to make this even clearer.
But it's not that easy 😉

Issue #2214, opened 2022-12-14
PVT File Reading Error


philippechehade:
**Describe the bug**
Simulating CO2 Injection in Aquifer using Ezrokhi's Correlations for Brine.
The below error is obtained : 
***** LOCATION: /tmp/GEOSX/src/coreComponents/constitutive/fluid/CO2BrineFluid.cpp:223
***** Controlling expression (should be false): true
'**** Rank 0: CO2BrineEzrokhiFluid fluid: invalid PVT function type '

The input files of the simulation are attached to this message 
The pvt data used for the liquid is the same as : inputFiles/compositionalMultiphaseWell/benchmarks/Class09Pb3/tables/pvtliquid_ez.txt
Version : 9d7cd0d
@jeannepellerin 
[Debug.zip](https://github.com/GEOSX/GEOSX/files/10226431/Debug.zip)

francoishamon:
Hello @philippechehade 
Thanks for reporting this.

To make your test work, just remove the blank lines at the end of the `pvtgas.txt`. The piece of code that reads this file is not very robust and does not work if there is a blank line at the end of the file.

That being said, this way of specifying the PVT functions using external files in the CO2-brine model is very old-fashioned (actually comes from another code) and we have discussed many times the fact that we should change it. The information should come from the xml file using proper tags and not from an external file. So now is the time to do it! I will fix this.

philippechehade:
Thank you @francoishamon 
The issue is solved !

MelReyCG:
@herve-gross I'm fixing this issue by adding verifications to ignore empty lines. I'm also slightly changing the way to acquire space separated parameters (from stringutilities::tokenize( str, " " ) to a new stringutilities::tokenizeBySpaces( str ), which ignore correctly all types of whitespace, including the one surrounding the input string).
The commit I pushed only affect the CO2BrineFluid object but I'm going to change ReactiveBrineFluid and testCO2BrineModels too.

MelReyCG:
PVT files reading robustness should be improved. The problems identified are :
1.	Empty lines in a PVT file results in a crash,
2.	The only whitespaces taken into account is the space (' '),
3.	Empty entries processing is inconsistent: 
- tokenize("1|2|3" , "|")  returns 3 parameters,
- tokenize("|1|2|3|" , "|") returns 4 parameters (with the first one being empty)
 
[...]

PR #2307, opened 2023-02-22
PVT file reading error

MelReyCG:
This PR aims to fix the problems we detected with the parsing of CO2BrineFluid::phasePVTParaFiles and ReactiveBrineFluid::flashModelParaFile attributes. With the proposed solution : 
- Empty lines no longer result in a crash,
- Whitespaces can no longer interfere with the table function parameters reading.

The issue #2214 is now solved.


MelReyCG:
Should we adapt the unit test "testCO2BrinePVTModels" to test these safeguards ? (by adding random whitespaces in one of the "TableContents" as an example).
@TotoGaz

untereiner:
Any reason std::isspace wasn't used before to check whitespaces delimiters ?

TotoGaz:
> Should we adapt the unit test "testCO2BrinePVTModels" to test these safeguards ? (by adding random whitespaces in one of the "TableContents" as an example). @TotoGaz

Yes, most likely! Comment what you do in the code so it does not get removed.

src/coreComponents/codingUtilities/StringUtilities.cpp
```diff
@@ -63,6 +63,44 @@ RETURN_TYPE tokenize( string const & str,
...
+          ++pos;
+        } while ( pos<strEnd && !std::isspace( str[pos] ) );
+
+        tokens.emplace_back( str.substr( lastPos, pos - lastPos ) );
+      }
+    }
+    
+    return tokens;
+  }
+}
```
TotoGaz, line 103: Would
```c++
std::vector< string > tokenizeBySpaces( string const & str )
{
    return tokenize( str, " \f\n\r\t\v", true );
}
```
be equivalent? I know you're also dealing with the last value being void, but you can surely manage this as a post processing of the `tokenize` call with something like `if ( std::isspace( str.back() ) and not result.empty() ) result.pop_back();`?

src/coreComponents/codingUtilities/StringUtilities.hpp
```diff
@@ -102,6 +102,10 @@ RETURN_TYPE tokenize( string const & str,
                       string const & delimiters,
                       bool const treatConsecutiveDelimAsOne = true );
 
+/// Subdivide string by whitespaces (see std::isspace()) (do not create any empty token).
+template< typename RETURN_TYPE = string_array >
+RETURN_TYPE tokenizeBySpaces( string const & str );
```
TotoGaz, line 107: I think that passing by we could update the `tokenize` function with something like
```c++
template< template< class... > class C = std::vector>
C< string > tokenize( string const & str,
                      string const & delimiters,
                      bool const treatConsecutiveDelimAsOne )
```
and use it like
```c++
std::list< string > res = tokenize< std::list >( a, b, c );
```
if you want to change the container. But the `string` will be implied.

Also, we should default to `std` containers when manipulating `string`.
They are more powerful and flexible than our HPC `LvArrays` in that precise context.
Could you make the update in the code (and use `std` containers)? There are like max 10 calls to this function.

src/coreComponents/codingUtilities/StringUtilities.cpp
klevzoff: In Python the idiomatic way to split a string with leading/trailing whitespace would be something like `s.strip().split()`, with the `split()` part already handled by our `tokenize()`, so maybe we could also just implement `strip()`?

TotoGaz: Yes, you're most probably correct, trimming the input is surely a better way.

MelReyCG: Currently, `tokenize("1|2|3", "|", true)` returns 3 results and `tokenize("|1|2|3|", "|", true)` returns 4 results with the first one being empty (where we could intuitively expect 3 or 5 results with empty entries at the sides).

I'm wondering, would it be be better to :
- Assume this behaviour is expected (and maybe document this on `tokenize()`) and just pre-trim the input of `tokenizeBySpaces()`, as @klevzoff suggested ?
- Adapt the `tokenize()` algorithm so it never create empty entries when treatConsecutiveDelimAsOne is on (and maybe rename this parameter to "dontCreateEmptyEntries", so its behaviour will be clearer) ?

src/coreComponents/codingUtilities/StringUtilities.hpp
MelReyCG: No problem, I will make the update right away.

src/coreComponents/codingUtilities/StringUtilities.cpp
klevzoff: > Currently, `tokenize("1|2|3", "|", true)` returns 3 results and `tokenize("|1|2|3|", "|", true)` returns 4 results with the first one being empty (where we could intuitively expect 3 or 5 results with empty entries at the sides).

Yep, that's clearly a bug if so. I would expect 5 results here, for parity with `str.split()` functions in Python, JS and numerous other languages. Sometimes empty substrings *are* significant, so I don't think we would want 3 results by default. Although we could add another boolean parameter to control this?

MelReyCG: As these options behavior would be exclusive, would an enum with three values be acceptable ? (treatEveryDelims, treatConsecutiveDelimAsOne, dontCreateEmptyEntries)

TotoGaz: > Currently, tokenize("1|2|3", "|", true) returns 3 results and tokenize("|1|2|3|", "|", true) returns 4 results with the first one being empty (where we could intuitively expect 3 or 5 results with empty entries at the sides).

We surely do not want this 4 entry result... 3 seems legit, while 5 can be acceptable. But 4 makes little sense.
I was thinking about the `Group::getGroupByPath` member function (which tokenizes on `/` with a precise usage of an initial `/` to define an absolute path), but it does not use the `tokenize` function, so it's OK.

I would personally go for the solutions that returns 5 tokens (that way, there's a difference between `1|2|3` and `|1|2|3|`), with maybe the possibility to trim (as an option?) the input string and remove the potentially first and last dummy values.

src/coreComponents/codingUtilities/StringUtilities.hpp
TotoGaz: Great! First be sure to fix your initial issue. Otherwise you may end up with conflicts because the footprint of your PR is going to grow.

src/coreComponents/codingUtilities/StringUtilities.cpp
TotoGaz: > As these options behavior would be exclusive, would an enum with three values be acceptable ? (treatEveryDelims, treatConsecutiveDelimAsOne, dontCreateEmptyEntries)

Are they? The input string `|a||b|` could be converted into
- `a`, `b`
- `a`, `-`, `b`
- `-`, `a`, `b`, `-`
- `-`, `a`, `-`, `b`, `-`

which are 4 choices?

MelReyCG: To sum up, the solution will be to adapt tokenize to have `tokenize( string const & str, string const & delimiters, bool const treatConsecutiveDelimAsOne=true, bool const preTrimStr=false )`.

Following this logic, `||a||b||` would give : 
- tokenize( [...], 1, 1 ) = `a`,`b`
- tokenize( [...], 0, 1 ) = `a`,`-`,`b`
- tokenize( [...], 1, 0 ) = `-`,`a`,`b`,`-`
- tokenize( [...], 1, 1 ) = `-`,`-`,`a`,`-`,`b`,`-`,`-`

TotoGaz (commented):
Great work, I especially appreciate your efforts on the unit tests.

src/coreComponents/codingUtilities/StringUtilities.cpp
```diff
@@ -38,40 +38,76 @@ string toLower( string const & input )
...
  **/
-template< typename RETURN_TYPE >
-RETURN_TYPE tokenize( string const & str,
-                      string const & delimiters,
-                      bool const treatConsecutiveDelimAsOne )
+template< template< class ... > class LIST_T = std::vector >
+LIST_T< string > tokenize( string const & str,
+                           string const & delimiters,
+                           bool const treatConsecutiveDelimAsOne,
+                           bool const preTrimStr )
```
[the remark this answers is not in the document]
MelReyCG: No problem, I made the multiple instantiations to conform with the original code, but it totally makes sense to move their definition to the hpp.

src/coreComponents/codingUtilities/StringUtilities.cpp
```diff
@@ -38,40 +38,76 @@ string toLower( string const & input )
...
   return tokens;
 }
+/**
+ * String tokenizing by whitespace function
+ **/
+template< template< class ... > class LIST_T = std::vector >
+LIST_T< string > tokenizeBySpaces( string const & str )
+{
+  return tokenize< LIST_T >( str, " \f\n\r\t\v", true, true );
+}
```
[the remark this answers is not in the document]
MelReyCG: I prefer to keep these functions flexible.

src/coreComponents/codingUtilities/StringUtilities.cpp
```diff
@@ -38,40 +38,76 @@ string toLower( string const & input )
 /**
  * String tokenizing function
  **/
-template< typename RETURN_TYPE >
-RETURN_TYPE tokenize( string const & str,
-                      string const & delimiters,
-                      bool const treatConsecutiveDelimAsOne )
+template< template< class ... > class LIST_T = std::vector >
```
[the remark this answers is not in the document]
MelReyCG: I choosed the second one.

src/coreComponents/codingUtilities/tests/testStringUtilities.cpp
```diff
@@ -22,26 +22,173 @@ using namespace geosx;
...
+      TokenizeTest( string( "ab" ), true, false, { "ab" } ),
+      TokenizeTest( string( "ab" ), false, true, { "ab" } ),
+      TokenizeTest( string( "ab" ), true, true, { "ab" } ),
+    };
+
+    for( TokenizeTest test : tokenizeTests )
+    {
+      std::vector< string > r = stringutilities::tokenize( test.m_strToTest, "|",
+                                                           bool(test.m_treatConsecutiveDelimAsOne),
+                                                           bool(test.m_preTrimStr));
```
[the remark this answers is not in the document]
MelReyCG: A mistake, I forgotten to erase these conversions from a test.

src/coreComponents/codingUtilities/StringUtilities.cpp
```diff
@@ -38,40 +38,76 @@ string toLower( string const & input )
...
 
   return tokens;
 }
+/**
+ * String tokenizing by whitespace function
+ **/
+template< template< class ... > class LIST_T = std::vector >
+LIST_T< string > tokenizeBySpaces( string const & str )
+{
+  return tokenize< LIST_T >( str, " \f\n\r\t\v", true, true );
```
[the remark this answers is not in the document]
MelReyCG: I don't think this function should ignore `\n` and make two substrings stick together. 
The input string shouldn't have `\n` in it, but if this was the case, that would mean the function user wants to separate his strings by line returns too.

src/coreComponents/codingUtilities/StringUtilities.cpp
```diff
@@ -38,40 +38,76 @@ string toLower( string const & input )
...
   {
     return {};
   }
 
-  RETURN_TYPE tokens;
-  size_t lastPos = 0;
-  size_t newPos;
-  while( ( newPos = str.find_first_of( delimiters, lastPos ) ) != string::npos )
+  LIST_T< string > tokens;
+  size_t tokenBegin, tokenEnd, strEnd;
```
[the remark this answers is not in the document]
MelReyCG: As they are string characters iterators, shouldn't it be string::size_type ?

src/coreComponents/constitutive/fluid/CO2BrineFluid.cpp
```diff
@@ -192,45 +192,48 @@ void CO2BrineFluid< PHASE1, PHASE2, FLASH >::createPVTModels()
...
         {
-          phase2InputParams[PHASE2::InputParamOrder::ENTHALPY] = strs;
+          GEOSX_THROW( GEOSX_FMT( "{}: invalid PVT function type '{}'", getFullName(), strs[0] ), InputError );
         }
       }
-      else
-      {
-        GEOSX_THROW( GEOSX_FMT( "{}: invalid PVT function type '{}'", getFullName(), strs[0] ), InputError );
-      }
     }
```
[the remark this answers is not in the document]
MelReyCG: If the PVT file isn't correctly formed, I want GEOSX to stop and throw an error.

src/coreComponents/codingUtilities/tests/testStringUtilities.cpp
```diff
@@ -22,26 +22,173 @@ using namespace geosx;
...
+      {
+        FAIL() << "tokenizeTests( '" << test.m_strToTest << "', \"|\""
+               << ", delimsAsOne=" << bool(test.m_treatConsecutiveDelimAsOne)
+               << ", preTrim=" << bool(test.m_preTrimStr)
+               << " ) failed : " << r.size()
+               << " results instead of " << test.m_results.size();
+      }
+      else
+      {
+        for( int i = 0, rEnd = r.size(); i < rEnd; ++i )
```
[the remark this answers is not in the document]
MelReyCG: It's an habit I have to precompute the iteration ending. Indeed, at high optimisation level, it didn't change anything in the assembly code (as the string is const).
My desire wasn't to obfuscate the code, I will apply your suggested change.

src/coreComponents/mesh/MeshObjectPath.cpp
```diff
@@ -63,7 +63,7 @@ std::vector< string >
 MeshObjectPath::fillPathTokens( string const & path,
                                 dataRepository::Group const & meshBodies ) const
 {
-  std::vector< string > pathTokens = stringutilities::tokenize< std::vector< string > >( path, "/" );
+  std::vector< string > pathTokens = stringutilities::tokenize( path, "/" );
```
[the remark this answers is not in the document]
MelReyCG: Of course!
I tested it and it is still returning an initial empty token.

src/coreComponents/codingUtilities/tests/testStringUtilities.cpp
```diff
@@ -20,28 +20,198 @@
 
 using namespace geosx;
 
+struct TokenizeTest
+{
+  string m_strToTest;
+  string m_delims;
+  bool m_treatConsecutiveDelimAsOne;
+  bool m_preTrimStr;
+  std::vector< string > m_results;
```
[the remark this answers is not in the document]
TotoGaz: > for now I ended constructor parameters with an underscore

Yes, I'm fine with this.

src/coreComponents/codingUtilities/tests/testStringUtilities.cpp
```diff
@@ -153,34 +168,24 @@ TEST( testStringUtilities, tokenize )
...
+        FAIL() << test << " failed: " << result <<" results instead of " << expected << ".";
       }
     }
 
   }
 
   // Spaces tokenizing test
   {
-    std::vector< TokenizeBySpacesTest > tokenizeBSTests = {
+    const std::vector< TokenizeBySpacesTest > tokenizeBSTests = {
```
TotoGaz, line 188:
```suggestion
   std::vector< TokenizeBySpacesTest > const tokenizeBSTests = {
```

src/coreComponents/codingUtilities/tests/testStringUtilities.cpp
```diff
@@ -193,23 +198,13 @@ TEST( testStringUtilities, tokenize )
 
     for( TokenizeBySpacesTest test : tokenizeBSTests )
     {
-      std::vector< string > r = stringutilities::tokenizeBySpaces( test.m_strToTest );
+      const std::vector< string > r = stringutilities::tokenizeBySpaces( test.strToTest );
```
TotoGaz, line 201:
```suggestion
     std::vector< string > const r = stringutilities::tokenizeBySpaces( test.strToTest );
```

TotoGaz (approved):
Just a last `const` swap and it's OK!

TotoGaz (approved):

PR #2312, opened 2023-02-23
Feature/change lifo default parameters and fix bugs

XL64:
This change fix bugs in LIFO + C/S; change default parameters; takes into account some remarks raised in #2288

Rebaseline PR: https://github.com/GEOS-DEV/integratedTests/pull/22


XL64:
@corbett5 , @rrsettgast I have an issue with LVArray. The LvArray I use to store data in FixedSizeDeque is using localIndex as it needs to receive arrays from GEOSX. I cannot allocate more than 8GB because of that (NbArray*ArraySize must be less than the maximum value of an int). Do you hava a solution so that I can copy an LvArray indexed with 32bits integers into an 64bits one. Or a solution to allocate more than 8GB in a int32 indexed 2D LvArray ? Both dimensions fit in int32 but the product doesn't.

corbett5:
You shouldn't have any problem copying from an LvArray::Array to another with a different index type using LvArray::memcpy.

XL64:
Indeed i just had to add some interfaces to theFixedSizeDeque and could use ptrdiff as storage. Thanks @corbett5

src/coreComponents/common/LifoStorage.hpp
```diff
@@ -0,0 +1,198 @@
...
+#else
+    if( numberOfBuffersToStoreOnDevice == -1 )
+    {
+      size_t free, total;
+      GEOSX_ERROR_IF( cudaSuccess != cudaMemGetInfo( &free, &total ), "Error getting CUDA device available memory" );
+      double freeGB = ( ( double ) free ) / ( 1024.0 * 1024.0 * 1024.0 );
+      LIFO_LOG_RANK( " LIFO : available memory on device " << freeGB << " GB" );
+      numberOfBuffersToStoreOnDevice = std::min( (int)( 0.8 * free / m_bufferSize ), m_maxNumberOfBuffers );
+    }
+#endif
```
TotoGaz, line 86: I think we want to remove this kind of
```c++
#ifndef GEOSX_USE_CUDA
    ...
#else
...
#endif
```
patterns in our code.
Is there a strong limitation that prevents you from putting
```
    numberOfBuffersToStoreOnDevice = 0;
```
in the `cuda` implementation and
```c++
    if( numberOfBuffersToStoreOnDevice == -1 )
    {
      size_t free, total;
      GEOSX_ERROR_IF( cudaSuccess != cudaMemGetInfo( &free, &total ), "Error getting CUDA device available memory" );
      double freeGB = ( ( double ) free ) / ( 1024.0 * 1024.0 * 1024.0 );
      LIFO_LOG_RANK( " LIFO : available memory on device " << freeGB << " GB" );
      numberOfBuffersToStoreOnDevice = std::min( (int)( 0.8 * free / m_bufferSize ), m_maxNumberOfBuffers );
    }
```
in the `host` one, and so on?
Eventually this file should only contain something like
```c++
#include "common/LifoStorageHost.hpp"
#ifdef GEOSX_USE_CUDA
#include "common/LifoStorageCuda.hpp"
#endif

#ifndef GEOSX_USE_CUDA
template< typename T, typename INDEX_TYPE >
class LifoStorage = LifoStorageCuda< T, INDEX_TYPE >
#else
template< typename T, typename INDEX_TYPE >
class LifoStorage = LifoStorageHost< T, INDEX_TYPE >
#endif
```

src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEM.cpp
```diff
@@ -887,16 +882,23 @@ real64 AcousticWaveEquationSEM::explicitStepForward( real64 const & time_n,
...
       arrayView1d< real32 > const p_dt2 = nodeManager.getField< fields::PressureDoubleDerivative >();
 
-      if( NULL == std::getenv( "DISABLE_LIFO" ) )
+      if( m_enableLifo )
       {
+        if( !m_lifo )
+        {
+          int const rank = MpiWrapper::commRank( MPI_COMM_GEOSX );
+          std::string lifoPrefix = GEOSX_FMT( "lifo/rank_{:05}/pdt2_shot{:06}", rank, m_shotIndex );
+          m_lifo = std::unique_ptr< LifoStorage< real32, localIndex > >( new LifoStorage< real32, localIndex >( lifoPrefix, p_dt2, m_lifoOnDevice, m_lifoOnHost, m_lifoSize ) );
```
TotoGaz, line 891: Maybe `make_unique`?

src/coreComponents/common/LifoStorageCommon.hpp
```diff
@@ -0,0 +1,295 @@
...
+   * Asynchroneously push a copy of the given LvArray into the LIFO
+   *
+   * @param array The LvArray to store in the LIFO, should match the size of the data used in constructor.
+   */
+  virtual void pushAsync( arrayView1d< T > array ) = 0;
+
+  /**
+   * Waits for last push to be terminated
+   */
+  virtual void pushWait() = 0;
```
TotoGaz, line 126: Do you have to define `virtual` members in this `class`?

src/coreComponents/common/LifoStorageCommon.hpp
```diff
@@ -0,0 +1,295 @@
...
+   * Waits for last pop to be terminated
+   */
+  virtual void popWait() = 0;
+
+  /**
+   * Check if the LIFO is empty
+   *
+   * @return true if the LIFO does not contain a buffer.
+   */
+  bool isEmpty()
```
TotoGaz, line 173: While I do prefer having the `isEmpty` wording since it's more explicit, I think the `STL` convention for containers is `empty`.

src/coreComponents/common/LifoStorageCommon.hpp
```diff
@@ -0,0 +1,295 @@
...
+
+namespace geosx
+{
+/**
+ * This class is used to store in a LIFO way buffers, first on device, then on host, then on disk.
+ */
+template< typename T, typename INDEX_TYPE >
+class LifoStorageCommon
+{
+protected:
```
TotoGaz, line 51: Please put `protected` and `private` components at the end of the class.

src/coreComponents/common/FourMutexesLock.hpp
```diff
@@ -0,0 +1,84 @@
...
+
+#include <mutex>
+
+namespace geosx
+{
+
+/**
+ * @brief Class to handle locks using 4 mutexes
+ */
+class FourMutexesLock
```
klevzoff, line 25: This and `TwoMutexesLock` could be unified in a variadic class for any number of mutexes (essentially a c++14 replacement for [scoped_lock](https://en.cppreference.com/w/cpp/thread/scoped_lock)). Without deduction guides the usage will be slightly less elegant, but still better than having duplicate classes.

src/coreComponents/common/LifoStorage.hpp
```diff
@@ -0,0 +1,198 @@
...
+    if( numberOfBuffersToStoreOnHost == -1 )
+    {
+      size_t free = sysconf( _SC_AVPHYS_PAGES ) * sysconf( _SC_PAGESIZE );
+      numberOfBuffersToStoreOnHost = std::max( 1, std::min( (int)( 0.8 * free / m_bufferSize ), m_maxNumberOfBuffers - numberOfBuffersToStoreOnDevice ) );
+      double freeGB = ( ( double ) free ) / ( 1024.0 * 1024.0 * 1024.0 );
```
XL64, line 91: Formula does not take into account the multiple MPI per node that share the memory, need to update that.

src/coreComponents/common/LifoStorage.hpp
XL64: In fact it's ifndef, I could remove the "numberOfBuffersToStoreOnDevice = 0;" which is not used. I didn't know how to put the numberOfBuffersToStoreOnDevice computation in the device part because it is used to initialize the FixedSizeDeque before the constructor content.
Maybe I can have a static method in each that compute it.

TotoGaz: _Without looking in the details, so I may be off topic:_
When you're in the `host` context, you know that `numberOfBuffersToStoreOnDevice` will be `-1` (IIUC).
Then you could directly provide the `-1` value for the initialization, without using any variable? (or something in the same mood).

XL64: I cannot do exactly the using thing because in the case of 0 buffer on device, even if CUDA is available, I can't use the CUDA Lifo.
A solution to limit the number of ifdef would be to define LIfoStorageCuda pointing to a LifoCudaHost in case CUDA is not available.
Like that :
```
#include "common/LifoStorageHost.hpp"
#ifdef GEOSX_USE_CUDA
#include "common/LifoStorageCuda.hpp"
#else
template< typename T, typename INDEX_TYPE >
using LifoStorageCuda = LifoStorageHost< T, INDEX_TYPE >;
#endif
```
Not sure it helps readability...

src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEM.cpp
XL64: Ok

src/coreComponents/common/LifoStorageCommon.hpp
XL64: ok

src/coreComponents/common/FourMutexesLock.hpp
XL64: Indeed I supposed I could do something on this, just did not figured out how to do it easily.

src/coreComponents/common/LifoStorageCommon.hpp
XL64: ok

src/coreComponents/common/LifoStorage.hpp
XL64: I'll improve that by dividing by the size of the node communicator but it does not take care of the case where several jobs are on the same node. Is there a geosx parameter to set an arbitrary limit on the available memory ? In that case I could get this capped memory per rank, remove process RSS and get the percentage.

XL64: I'll have the same issue if I oversubscribe on the GPU.

src/coreComponents/common/FourMutexesLock.hpp
XL64: I guess something like that should do the trick: https://blazingcpp.wordpress.com/2020/11/21/stdtuple-and-parameter-packs/

src/coreComponents/common/LifoStorage.hpp
XL64: I proposed something tell me if that suit you.

src/coreComponents/common/FourMutexesLock.hpp
klevzoff: It's a good article, but it uses C++17. I made a little prototype of what I mean in C++14, feel free to reuse as much as you'd like: https://godbolt.org/z/n4KdnWWrv

XL64: Thanks, I ended with a similar code (thanks to chatGPT also ^^) but I understood that std::apply was C++17 also. It seems not so that perfect.

XL64: I replaces one of my TwoMutexLock with this MultiMutexesLock() and get a deadlock for now:
```
/*
 * ------------------------------------------------------------------------------------------------------------
 * SPDX-License-Identifier: LGPL-2.1-only
 *
 * Copyright (c) 2018-2020 Lawrence Livermore National Security LLC
 * Copyright (c) 2018-2020 The Board of Trustees of the Leland Stanford Junior University
 * Copyright (c) 2018-2020 TotalEnergies
 * Copyright (c) 2019-     GEOSX Contributors
 * All rights reserved
 *
 * See top level LICENSE, COPYRIGHT, CONTRIBUTORS, NOTICE, and ACKNOWLEDGEMENTS files for details.
 * ------------------------------------------------------------------------------------------------------------
 */
#ifndef MULTIMUTEXESLOCK_HPP
#define MULTIMUTEXESLOCK_HPP

#include <mutex>
#include <tuple>

#include "codingUtilities/Utilities.hpp"

namespace geosx
{
// apply implementation taken from https://en.cppreference.com/w/cpp/utility/apply
namespace internal
{

template < class F, class Tuple, std::size_t... I >
constexpr decltype( auto ) apply( F&& f, Tuple&& t, std::index_sequence<I...> )
{
    return std::forward<F>( f )( std::get<I>(std::forward<Tuple>(t))...);
}

}

template < class F, class Tuple >
constexpr decltype(auto) apply( F&& f, Tuple&& t )
{
    return internal::apply( std::forward< F >( f ), std::forward< Tuple >( t ),
                            std::make_index_sequence< std::tuple_size< std::remove_reference_t<Tuple> >::value >{} );
}

/**
 * @brief Class to handle locks using 2 mutexes
 */
template < typename... Mutexes >
class MultiMutexesLock
{
public:
  /**
   * @brief Construct a multi mutexes lock and lock the mutexes.
   * @param mutexes The mutexes associated with the lock.
   */
    MultiMutexesLock( Mutexes&... mutexes )
      : m_islocked(false), m_mutexes(std::make_tuple(std::ref(mutexes)...))
    {
      lock();
    }

    void lock()
    {
      if( m_islocked ) return;
      apply( []( auto && ... mutexes, auto ){ std::lock( mutexes... ); }, m_mutexes );
      m_islocked = true;
    }

    void unlock()
    {
      if( !m_islocked ) return;
      forEachArgInTuple( m_mutexes, []( auto & mutex ){ mutex.unlock(); } );
      m_islocked = false;
    }

private:
  /// Indicate if the mutexes are owned by the lock
  bool m_islocked;
  /// Array of references to the mutexes
  std::tuple<Mutexes&...> m_mutexes;
};
}
#endif // MULTIMUTEXESLOCK_HPP
```

klevzoff: The code you posted is missing a destructor `~MultiMutexesLock()` which could explain the deadlock. Can you check if that is the case? And while we're at it, I would also add a deleted copy constructor to prevent accidental misuse. 

Another question I had was whether you actually need the publicly accessible `lock()` and `unlock()` functions. In my example code I made them private on purpose to emulate the purely scope-based locking control that `std::scoped_lock` implements (it also makes the code simpler, since you no longer need the `m_islocked` flag). It's up to you though.

XL64: Indeed I never use them directly. I'll make them private, and yes I was never calling the unlock() this is why I got the deadlock...

XL64: I need the publicity for waiting on condition variables.

src/coreComponents/common/LifoStorageCommon.hpp
XL64: @TotoGaz any advice on that ?

src/coreComponents/common/FourMutexesLock.hpp
XL64: Seems I get a race condition, need investigate that.

src/coreComponents/common/LifoStorage.hpp
XL64: I didn't copied your code correctly it seems (and running on my home directory didn't help...) No it seems OK, I removed the std::make_tuple(std::ref(mutexes)) and used your make_multilock(), tests passed fine.

src/coreComponents/common/FourMutexesLock.hpp
XL64: I didn't copied your code correctly it seems (and running on my home directory didn't help...) No it seems OK, I removed the std::make_tuple(std::ref(mutexes)) and used your make_multilock(), tests passed fine.

src/coreComponents/common/LifoStorage.hpp
XL64: It seems the first solution does not work without CUDA, I changed it with a new one. Any comment @TotoGaz ?

src/coreComponents/common/LifoStorage.hpp
```diff
@@ -0,0 +1,198 @@
...
+    if( numberOfBuffersToStoreOnHost < 0 )
+    {
+      GEOSX_ERROR_IF( numberOfBuffersToStoreOnHost < -100, "Error, numberOfBuffersToStoreOnHost should be greater than -100" );
+      size_t free = sysconf( _SC_AVPHYS_PAGES ) * sysconf( _SC_PAGESIZE );
+      numberOfBuffersToStoreOnHost = std::max( 1, std::min( ( int )( -0.01 * numberOfBuffersToStoreOnHost * free / m_bufferSize ), m_maxNumberOfBuffers - numberOfBuffersToStoreOnDevice ) );
+      double freeGB = ( ( double ) free ) / ( 1024.0 * 1024.0 * 1024.0 ) / MpiWrapper::nodeCommSize();
+      LIFO_LOG_RANK( " LIFO : available memory on host " << freeGB << " GB" );
+    }
+    LIFO_LOG_RANK( " LIFO : allocating "<< numberOfBuffersToStoreOnHost <<" buffers on host" );
+    LIFO_LOG_RANK( " LIFO : allocating "<< numberOfBuffersToStoreOnDevice <<" buffers on device" );
```
TotoGaz, line 84: So please try to move this to `LifoStorageCommon` and delegate the code between the `ifdef/else/endif` to the appropriate constructors in `LifoStorageHost` and `LifoStorageCuda`.
That way you should be able to remove the `LifoStorage` class and use the base class `LifoStorageCommon` as base class (that you'll be able to rename `LifoStorage` I think.)

src/coreComponents/common/LifoStorageCuda.hpp
```diff
@@ -0,0 +1,215 @@
...
+    m_pushToDeviceEvents( maxNumberOfBuffers ),
+    m_popFromDeviceEvents( maxNumberOfBuffers )
+  {}
+
+  /**
+   * Asynchroneously push a copy of the given LvArray into the LIFO
+   *
+   * @param array The LvArray to store in the LIFO, should match the size of the data used in constructor.
+   */
+  void pushAsync( arrayView1d< T > array )
```
TotoGaz, line 69: And w.r.t. to this pattern, do not forget to add the `override` keyword, and most probably the `final` as well (maybe define the whole class as `final` will be easier.)
That way, if you directly use the derived class and not the base class, the compiler will be able to make some optimisations.

src/coreComponents/common/LifoStorage.hpp
XL64: I tried it this morning but I'm still stick with the choose to tak a Host or Cuda Lifo depending the values of the constructor parameters and the remaining memory at the given time. This choice could be moved to the user code maybe but still with ifdef etc...

src/coreComponents/common/LifoStorageCuda.hpp
XL64: Done

wrtobin:
Just a heads up that we now support both CUDA and HIP/ROCM in geos and lvarray. 

Please rename cuda-specific terminology and macros to generic `device` versions where possible, and providing hip/rocm specific alternatives where the same code can't be used on multiple platforms (though reviewing this implementation I think everything can be made `device`-generic).

XL64:
> Just a heads up that we now support both CUDA and HIP/ROCM in geos and lvarray.
> 
> Please rename cuda-specific terminology and macros to generic `device` versions where possible, and providing hip/rocm specific alternatives where the same code can't be used on multiple platforms (though reviewing this implementation I think everything can be made `device`-generic).

I'm using directly camp::resource::CudaEvent or cudaMemGetInfo, so I definitely need ifdef for HIP but I'm not able to test it on HIP. I could implement it without testing but I'm not very confident with that... What do you think ?

XL64:
@TotoGaz , @klevzoff, @CusiniM Ok with that PR ? @wrtobin Did you see my question, do you have advices?

wrtobin:
@XL64 We definitely need to avoid cuda-specific types and functions in order to maintain cross-platform compatibility. That said none of our CI currently enforces this, and AMDs support for HIP/ROCM on consumer cards is... lacking, so access to reasonable native dev platforms is understandably limited.

I am working on getting our local MI250x platform added to our nightly regression testing, and eventually we'll add PR CI stages that are HIP/ROCM specific, but until we can ensure all our devs can access tools to perform that build I'm uncertain whether we should enforce this as a hard constraint, despite how much extra work that means for me.

I *want* to say we can have someone devote internal dev time to finalize this PR to support HIP/ROCM as well, but I'm the sole dev for our AMD support (at this moment) and I only have 20 hours each week to give to geos. Based on my current personal dev queue I can probably take a look at this in ~2 weeks.

TotoGaz:
This makes sense @wrtobin 
Also, this PR is meant to reorganize the code, leading to a better separation of a code which is already `cuda` specific. Maybe it's better to work on a cleaner version then?

wrtobin:
Certainly any changes which can be made to simplify accommodating HIP/Rocm and Cuda would be good. My primary HIP/ROCM generalized a lot of functions/macros/etc as `device` instead of specifically `cuda`. Both short- and long-term moves into that sort of generalization (if possible) would be good. 

(though without the ability to actually test on the HIP/Rocm build it may to some degree simply be down to generalizing to `device` and trusting in the short-term until someone at LLNL can test on Tioga)

XL64:
Can we validate this PR without implementing the HIP version as we have no machine. The CUDA specific code is already present before this PR. The CUDA code is protected with ifdefs and won't be compiled if GEOSX is not built with CUDA.
I opened an issue to fix it ater on (when we have a HIP machine) https://github.com/GEOS-DEV/GEOS/issues/2582
I'm leaving the team within 2 weeks, it would be better to integrate this before.

wrtobin (approved):

wrtobin:
This branch builds on Tioga.

sframba (approved):

herve-gross:
FYI, this PR causes 4 "unused variable" warning/errors on my compiler (Apple clang version 14.0.3 (clang-1403.0.22.14.1)). Probably nothing of importance, but I just wanted to mention it. Here is the log.


`In file included from /Users/hervegross/Documents/code/GEOS/src/coreComponents/physicsSolvers/wavePropagation/WaveSolverBase.cpp:20:
In file included from /Users/hervegross/Documents/code/GEOS/src/coreComponents/physicsSolvers/wavePropagation/WaveSolverBase.hpp:26:
In file included from /Users/hervegross/Documents/code/GEOS/src/coreComponents/common/LifoStorage.hpp:29:
/Users/hervegross/Documents/code/GEOS/src/coreComponents/common/LifoStorageCommon.hpp:158:28: error: use of undeclared identifier '_SC_AVPHYS_PAGES'
    size_t free = sysconf( _SC_AVPHYS_PAGES ) * sysconf( _SC_PAGESIZE );
                           ^
In file included from /Users/hervegross/Documents/code/GEOS/src/coreComponents/physicsSolvers/wavePropagation/WaveSolverBase.cpp:20:
In file included from /Users/hervegross/Documents/code/GEOS/src/coreComponents/physicsSolvers/wavePropagation/WaveSolverBase.hpp:26:
In file included from /Users/hervegross/Documents/code/GEOS/src/coreComponents/common/LifoStorage.hpp:30:
/Users/hervegross/Documents/code/GEOS/src/coreComponents/common/LifoStorageHost.hpp:156:49: error: unused parameter 'percent' [-Werror,-Wunused-parameter]
  static int computeNumberOfBufferOnDevice( int percent, size_t bufferSize, int maxNumberOfBuffers ) { return 0; }
                                                ^
/Users/hervegross/Documents/code/GEOS/src/coreComponents/common/LifoStorageHost.hpp:156:65: error: unused parameter 'bufferSize' [-Werror,-Wunused-parameter]
  static int computeNumberOfBufferOnDevice( int percent, size_t bufferSize, int maxNumberOfBuffers ) { return 0; }
                                                                ^
/Users/hervegross/Documents/code/GEOS/src/coreComponents/common/LifoStorageHost.hpp:156:81: error: unused parameter 'maxNumberOfBuffers' [-Werror,-Wunused-parameter]
  static int computeNumberOfBufferOnDevice( int percent, size_t bufferSize, int maxNumberOfBuffers ) { return 0; }
                                                                                ^
4 errors generated.
make[2]: *** [coreComponents/physicsSolvers/CMakeFiles/physicsSolvers.dir/wavePropagation/WaveSolverBase.cpp.o] Error 1
make[1]: *** [coreComponents/physicsSolvers/CMakeFiles/physicsSolvers.dir/all] Error 2`

wrtobin:
Often these are variables which are only used in logging/error macros, so they wind up being unused depending on compiler parameters / whether we suppress the macro, etc. I always try to add `GEOS_UNUSED_VAR/ARG` or (now that we have c++17) `[[maybe_unused]]` when I see these popup for a specific system / compiler / build type.

herve-gross:
Three unused vars were fixed in PR #2569

PR #2316, opened 2023-02-28
Docs about how to fill ElementRegions

MelReyCG:
Here is a proposal for modifications of the docs about how to fill `ElementRegions`:
- `ElementRegion` wasn't replaced by `CellElementRegion` everywhere in the exemples.
- `cellBlocks` & `materialList` attributes need brace symbols, but there weren't any in some exemples

### DIFF
--- a/src/coreComponents/mesh/docs/Mesh.rst
+++ b/src/coreComponents/mesh/docs/Mesh.rst
@@ -191,7 +191,7 @@ We advise users to use absolute path to the mesh file.
 
 GEOSX uses ``ElementRegions`` to support different physics
 or to define different constitutive properties.
-An ``ElementRegion`` is defined as a set of ``CellBlocks``.
+The ``ElementRegions`` block can contain several ``CellElementRegion`` blocks. A ``CellElementRegion`` is defined as a set of ``CellBlocks``.
 A ``CellBlock`` is an ensemble of elements with the same element geometry.
 
 .. figure:: mesh_multi.png
@@ -206,14 +206,14 @@ The ``ElementRegions`` are defined as below :
 .. code-block:: xml
 
   <ElementRegions>
-    <ElementRegion
+    <CellElementRegion
       name="Top"
-      cellBlocks="Top_hexahedra Top_wedges Top_tetrahedra"
-      materialList="water rock"/>
-    <ElementRegion
+      cellBlocks="{ Top_hexahedra, Top_wedges, Top_tetrahedra }"
+      materialList="{ water, rock }"/>
+    <CellElementRegion
       name="Bot"
-      cellBlocks="Bot_hexahedra Bot_wedges Bot_tetrahedra"
-      materialList="water rock"/>
+      cellBlocks="{ Bot_hexahedra, Bot_wedges, Bot_tetrahedra }"
+      materialList="{ water, rock }"/>
   </ElementRegions>
 
 You have to use the following syntax to declare your ``CellBlocks`` :
--- a/src/coreComponents/constitutive/docs/Constitutive.rst
+++ b/src/coreComponents/constitutive/docs/Constitutive.rst
@@ -22,7 +22,7 @@ These models are grouped together based on their input/output interface.
 
 In an input XML file, constitutive models are listed in the ``<Constitutive>`` block.
 Each parameterized model has its own XML tag, and each must be assigned a unique name via the ``name`` attribute.
-Names are used to assign models to regions of the physical domain via the ``materialList`` attribute of the ``<ElementRegion>`` node (see :ref:`XML_ElementRegions`).
+Names are used to assign models to regions of the physical domain via the ``materialList`` attribute of the ``<CellElementRegion>`` node (see :ref:`XML_ElementRegions`).
 In some cases, physics solvers must also be assigned specific constitutive models to use (see :ref:`Solvers`).
 
 A typical ``<Constitutive>`` and ``<ElementRegions>`` block will look like:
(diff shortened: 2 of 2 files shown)


src/coreComponents/mesh/docs/Mesh.rst
```diff
@@ -206,14 +206,14 @@ The ``ElementRegions`` are defined as below :
 .. code-block:: xml
 
   <ElementRegions>
-    <ElementRegion
+    <CellElementRegion
       name="Top"
-      cellBlocks="Top_hexahedra Top_wedges Top_tetrahedra"
-      materialList="water rock"/>
-    <ElementRegion
+      cellBlocks="{ Top_hexahedra, Top_wedges, Top_tetrahedra }"
```
francoishamon, line 211:
```suggestion
      cellBlocks="{ hexahedra, wedges, tetrahedra }"
```

src/coreComponents/mesh/docs/Mesh.rst
```diff
@@ -206,14 +206,14 @@ The ``ElementRegions`` are defined as below :
...
+      materialList="{ water, rock }"/>
+    <CellElementRegion
       name="Bot"
-      cellBlocks="Bot_hexahedra Bot_wedges Bot_tetrahedra"
-      materialList="water rock"/>
+      cellBlocks="{ Bot_hexahedra, Bot_wedges, Bot_tetrahedra }"
```
francoishamon, line 215:
```suggestion
      cellBlocks="{ hexahedra, wedges, tetrahedra, pyramids }"
```

When we removed the old mesh importer, the naming convention of the importing cell block changed. It is no longer possible to define a name (like `Bot` here) in the vtu file and have it prepended to the cell block name as done here. This page is really outdated, so maybe in the documentation, you can write something based on the following text, if you like it:

==== possible text ====
The naming of the imported `cellBlocks` depends on whether the data array called `regionAttribute` is present in the vtu file or not. This data array is used to define regions in the vtu file and assign the cells to a given region. The `regionAttribute` is a integer and not a string (unfortunately).

- If the vtu file does not contain `regionAttribute`, then all the cells are grouped in a single region, and the cell block names are just constructed from the cell types (`hexahedra`, `wedges`, `tetrahedra`, etc). We have then:
```
cellBlocks="{ hexahedra, wedges, tetrahedra, pyramids }"
```

- If the vtu file contains `regionAttribute`, then the cells are grouped by regions based on their individual (numeric) `regionAttribute`. In that case, the naming convention for the `cellBlocks` is `regionAttribute_elementType`. Assuming for instance that we want the `CellElementRegion` to contain all the cells of type `hexahedra`, `wedges`, and `tetrahedra` from the regions `1` and `3`, we write:
```
cellBlocks="{ 1_hexahedra, 1_wedges, 1_tetrahedra, 1_pyramids, 3_hexahedra, 3_wedges, 3_tetrahedra, 3_pyramids }"
```

Now, coming back to the example illustrated above, let's assume that the top region is identified in the vtu mesh by the `regionAttribute` 1, and that the bottom region is identified with 2. Then, we write
```
<ElementRegions>
  <CellElementRegion
    name="Top"
    cellBlocks="{ 1_hexahedra, 1_wedges, 1_tetrahedra, 1_pyramids }"
    materialList="{ water, rock }"/>
  <CellElementRegion
    name="Bot"
    cellBlocks="{ 2_hexahedra, 2_wedges, 2_tetrahedra, 2_pyramids }"
    materialList="{ water, rock }"/>
</ElementRegions>
```

We remind the user that all the imported `cellBlocks` must be included in one of the `CellElementRegion`s. Specifically, if some cells are meant to be inactive during the simulation, they still have to be included in a `CellElementRegion`. This `CellElementRegion` should simply not be included as a `targetRegion` of any of the solvers involved in the simulation.
==== end of text ====

Also, you can remove the lines just after that:
```
You have to use the following syntax to declare your CellBlocks :

nameOfTheRegionWithinTheMesh_typeOfTheElement
```

Last point, if you want: just above, there is this
```
<Mesh>
  <VTKMesh
    name="MyMeshName"
    file="/path/to/the/mesh/file.vtk"/>
</Mesh>
```
To provide more information, we could have something like the following text:

==== possible text ====
```
<Mesh>
  <VTKMesh
    name="MyMeshName"
    logLevel="1"
    file="/path/to/the/mesh/file.vtk"/>
</Mesh>
```
We advise users to use an absolute path to the mesh file and strongly recommend the use of a `logLevel` of 1 or more to obtain some information about the mesh import. This information contains for example the list of regions that are imported with their names, which is particularly useful to fill the `cellBlocks` field of the `ElementRegions` block (see below). Some information about the imported surfaces is also provided.
==== end of text =====

francoishamon (approved):
Great, thanks for doing that

src/coreComponents/mesh/docs/Mesh.rst
MelReyCG: Thanks a lot for these precisions!
I integrated all your suggestions. I just slightly reformulated the cellBlocks part because I prefer to show how the whole ``ElementRegions`` should be defined instead of presenting the ``cellBlocks`` without contextualizing it in an exemple.

src/coreComponents/mesh/docs/Mesh.rst
```diff
@@ -184,58 +184,97 @@ The mesh block has the following syntax:
...
    :width: 500
 
 In the example presented above, the mesh is is composed of two regions (*Top* and *Bot*).
-Each region contains 3 ``CellBlocks``.
+Each region contains 4 ``cellBlocks``.
 
-The ``ElementRegions`` are defined as below :
+- If the vtu file does not contain ``regionAttribute``, then all the cells are grouped in a single
+region, and the cell block names are just constructed from the cell types (hexahedra, wedges,
+tetrahedra, etc). Then in the exemple above, the ``ElementRegions`` can be defined as bellow:
```
francoishamon, line 216: You can visualize the documentation generated by ReadTheDocs here: https://geosx-geosx--2316.com.readthedocs.build/en/2316/coreComponents/mesh/docs/Mesh.html

To find this link, you just have to click on the "green check" on your PR (the thing that becomes a "red cross" if the test don't pass) and click on "details" and then "View docs" on the ReadTheDocs website.

You will see that the indentation is slightly messed up in a few places, if you don't mind fixing it. Let me know if you don't find the link

francoishamon (approved):
Great, thanks, just the indentation to fix in a few PRs, and we can merge the PR

PR #2323, opened 2023-03-07
Add interface nodesets to the wellbore mesh generator

sytuannguyen:
This tiny PR add two internal nodesets to the `InternalWellboreGenerator` for working with the casing-cement and cement-rock interfaces. These interfaces can be called in a xml input file for defining a field specification as following:

```
<FieldSpecification
      name="itConstraint_x"
      objectPath="nodeManager"
      fieldName="totalDisplacement"
      component="0"
      scale="0.0"
      setNames="{ rIT2 }"/>
```

This feature is important for working with the problem of a cased wellbore with imperfect thermo-hydro-mechanical interfaces between the layers.

### DIFF
--- a/src/coreComponents/mesh/generators/InternalWellboreGenerator.cpp
+++ b/src/coreComponents/mesh/generators/InternalWellboreGenerator.cpp
@@ -346,6 +346,10 @@ void InternalWellboreGenerator::coordinateTransformation( arrayView2d< real64, n
   SortedArray< localIndex > & tnegNodes = nodeSets["tneg"];
   SortedArray< localIndex > & tposNodes = nodeSets["tpos"];
 
+  // Nodesets for the casing-cement and cement-rock interfaces of a cased wellbore
+  SortedArray< localIndex > & rinterface1Nodes = nodeSets["rIT1"];
+  SortedArray< localIndex > & rinterface2Nodes = nodeSets["rIT2"];
+
   // Map to radial mesh
   for( localIndex a = 0; a < numNodes; ++a )
   {
@@ -383,6 +387,18 @@ void InternalWellboreGenerator::coordinateTransformation( arrayView2d< real64, n
     if( isEqual( X[a][1], m_max[1], m_coordinatePrecision ) )
     {
       tposNodes.insert( a );
+    }
+
+    // Nodeset of the first interface, next to rneg
+    if( isEqual( X[a][0], m_vertices[0][1], m_coordinatePrecision ) )
+    {
+      rinterface1Nodes.insert( a );
+    }
+
+    // Nodeset of the second interface
+    if( isEqual( X[a][0], m_vertices[0][2], m_coordinatePrecision ) )
+    {
+      rinterface2Nodes.insert( a );
     }
 
     X[a][0] = meshRact * cos( meshTheta );


herve-gross (approved):
looks good to me!

src/coreComponents/mesh/generators/InternalWellboreGenerator.cpp
```diff
@@ -346,6 +346,10 @@ void InternalWellboreGenerator::coordinateTransformation( arrayView2d< real64, n
   SortedArray< localIndex > & tnegNodes = nodeSets["tneg"];
   SortedArray< localIndex > & tposNodes = nodeSets["tpos"];
 
+  // Nodesets for the casing-cement and cement-rock interfaces of a cased wellbore
+  SortedArray< localIndex > & rinterface1Nodes = nodeSets["rIT1"];
+  SortedArray< localIndex > & rinterface2Nodes = nodeSets["rIT2"];
```
TotoGaz, line 351: Doing so you're adding a component to the public interface of GEOSX.
I think it should be document somewhere, otherwise you'll be the only one to be able to use the feature.

Changing the names in the future will result in a modification of all the input files.
It's something we'll be in trouble to fix right now.
So maybe could we use a more explicit name (`rIT1` and `rIT2` are complicated to  understand) so we hopefully do not have to change too often 🤣 .

sytuannguyen: @TotoGaz Thank you very much! Following your suggestion, a doc is added for the nodesets generated by the `InternalMesh` generator and the `InternalWellbore` generator. Also, more explicit `setNames` are now used.

TotoGaz (approved):

src/coreComponents/mesh/generators/InternalWellboreGenerator.cpp
```diff
@@ -346,6 +346,10 @@ void InternalWellboreGenerator::coordinateTransformation( arrayView2d< real64, n
   SortedArray< localIndex > & tnegNodes = nodeSets["tneg"];
   SortedArray< localIndex > & tposNodes = nodeSets["tpos"];
 
+  // Nodesets for the casing-cement and cement-rock interfaces of a cased wellbore
+  SortedArray< localIndex > & rCasingCementInterfaceNodes = nodeSets["rCasingCementInterface"];
```
CusiniM, line 350: to me it's a little unclear why these nodesets need to be predefined. They kind of mix geometry and physics. what if one had a wellbore like problem geometry but not regions called casing, cement and rock? I feel these sets should be defined using geometric objects or directly in the mesh through a `vtk` mesh.

TotoGaz: Yes, you're absolutely right.
Actually I saw it and I was like "_Let it go_", but I should not have.
Thanks for double checking it.

What do you think a proper name would be? `InternalInterface`, `ExternalInterface`?

CusiniM: those names certainly sound a lot better to me. However, there is no reason for always having 3 `cellblocks` so I would limit myself to a set defining the internal surface if doesn't already exist. interfaces between regions (or cellblocks) should not have predefined sets IMO. In a case with a single region those sets would become meaningless. 

It should be easy enough to create the set using a cylinder.

CusiniM: If a hollow cylinder does not exist yet, then we should create it and use it to create these nodesets

sytuannguyen: > It should be easy enough to create the set using a cylinder.

@CusiniM I like the idea of create nodeset within `Geometry` using a cylinder. It will be more generic. I will try with this idea.

sytuannguyen: @CusiniM @TotoGaz @herve-gross Following the suggestion of Mateo, I tested the `Cylinder` geometry. It works perfectly for this purpose. So I will close this PR. However I will propose some minor changes in another PR to clean the `Cylinder` class.

sytuannguyen:
@CusiniM @TotoGaz @herve-gross Following the suggestion of Mateo, I tested the Cylinder geometry. It works perfectly for this purpose. So I will close this PR. However I will propose some minor changes in another PR to clean the Cylinder class.

PR #2327, opened 2023-03-07
Chombo Coupling Fixes

wrtobin:
Requires hdf5_interface submodule update: https://github.com/GEOSX/hdf5_interface/pull/10

Probably going to impact some restart baselines since it reactivates several fields when using Quasistatic solver.

### DIFF
--- a/src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.cpp
+++ b/src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.cpp
@@ -146,20 +146,17 @@ void SolidMechanicsLagrangianFEM::registerDataOnMesh( Group & meshBodies )
     nodes.registerField< solidMechanics::incrementalDisplacement >( getName() ).
       reference().resizeDimension< 1 >( 3 );
 
-    if( m_timeIntegrationOption != TimeIntegrationOption::QuasiStatic )
-    {
-      nodes.registerField< solidMechanics::velocity >( getName() ).
-        reference().resizeDimension< 1 >( 3 );
+    nodes.registerField< solidMechanics::velocity >( getName() ).
+      reference().resizeDimension< 1 >( 3 );
 
-      nodes.registerField< solidMechanics::acceleration >( getName() ).
-        reference().resizeDimension< 1 >( 3 );
+    nodes.registerField< solidMechanics::acceleration >( getName() ).
+      reference().resizeDimension< 1 >( 3 );
 
-      nodes.registerField< solidMechanics::velocityTilde >( getName() ).
-        reference().resizeDimension< 1 >( 3 );
+    nodes.registerField< solidMechanics::velocityTilde >( getName() ).
+      reference().resizeDimension< 1 >( 3 );
 
-      nodes.registerField< solidMechanics::uhatTilde >( getName() ).
-        reference().resizeDimension< 1 >( 3 );
-    }
+    nodes.registerField< solidMechanics::uhatTilde >( getName() ).
+      reference().resizeDimension< 1 >( 3 );
 
     nodes.registerField< solidMechanics::mass >( getName() );
 
--- a/src/coreComponents/fileIO/Outputs/ChomboIO.cpp
+++ b/src/coreComponents/fileIO/Outputs/ChomboIO.cpp
@@ -75,17 +75,21 @@ bool ChomboIO::execute( real64 const GEOSX_UNUSED_PARAM( time_n ),
                         real64 const GEOSX_UNUSED_PARAM( eventProgress ),
                         DomainPartition & domain )
 {
-  if( m_coupler == nullptr )
-  {
-    GEOSX_ERROR_IF( m_waitForInput && m_inputPath == "/INVALID_INPUT_PATH", "Waiting for input but no input path was specified." );
-
-    m_coupler = new ChomboCoupler( MPI_COMM_GEOSX, m_outputPath, m_inputPath, domain.getMeshBody( 0 ).getBaseDiscretization() );
-  }
-
   if( cycleNumber < m_beginCycle )
   {
     return false;
   }
+  GEOSX_LOG_RANK_0( "Executing chombo coupling." );
+
+  if( m_coupler == nullptr )
+  {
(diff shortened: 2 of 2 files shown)


src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.cpp
```diff
@@ -146,20 +146,17 @@ void SolidMechanicsLagrangianFEM::registerDataOnMesh( Group & meshBodies )
     nodes.registerField< solidMechanics::incrementalDisplacement >( getName() ).
       reference().resizeDimension< 1 >( 3 );
 
-    if( m_timeIntegrationOption != TimeIntegrationOption::QuasiStatic )
-    {
-      nodes.registerField< solidMechanics::velocity >( getName() ).
-        reference().resizeDimension< 1 >( 3 );
+    nodes.registerField< solidMechanics::velocity >( getName() ).
```
CusiniM, line 149: can we maybe add a flag to only register these fields when the chombo coupling is on? coz registering them when they are not needing goes kind of in the opposite direction of reducing our memory footprint as much as possible.

wrtobin: Yeah I like that as a solution better.

src/coreComponents/dataRepository/Group.cpp
```diff
@@ -28,6 +28,9 @@ namespace geosx
 namespace dataRepository
 {
 
+/// @brief An empty orphan group node for null-value checks by address
+static Group NullGroup("null",nullptr);
```
wrtobin, line 32: I'm not super fond of this mechanism but its the least invasive way to support a `hasGroupByPath` implementation.

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.cpp
```diff
@@ -146,7 +146,7 @@ void SolidMechanicsLagrangianFEM::registerDataOnMesh( Group & meshBodies )
     nodes.registerField< solidMechanics::incrementalDisplacement >( getName() ).
       reference().resizeDimension< 1 >( 3 );
 
-    if( m_timeIntegrationOption != TimeIntegrationOption::QuasiStatic )
+    if( m_timeIntegrationOption != TimeIntegrationOption::QuasiStatic || Group::hasGroupByPath("/Outputs/ChombIO") )
```
wrtobin, line 149: We want to register these fields if there is a ChombIO output in the problem spec.
This is called after all the input has been parsed and the mesh created, so there should never be an issue with creation order, etc

src/coreComponents/dataRepository/Group.cpp
```diff
@@ -631,16 +638,32 @@ Group const & Group::getBaseGroupByPath( string const & path ) const
...
     }
   }
   while( currentPosition != string::npos );
 
   return *currentGroup;
 }
 
+bool Group::hasGroupByPath( string const & path ) const
+{
+  return &getBaseGroupByPath( path, false ) != &NullGroup;
```
wrtobin, line 664: `try except`ing the original `getBaseGroupByPath` and just catching/continuing if it threw also occurred to me, but we throw `std::abort` down the `GEOSX_ERROR` macro hierarchy, which per the spec isn't recoverable.

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.cpp
```diff
@@ -146,7 +147,8 @@ void SolidMechanicsLagrangianFEM::registerDataOnMesh( Group & meshBodies )
     nodes.registerField< solidMechanics::incrementalDisplacement >( getName() ).
       reference().resizeDimension< 1 >( 3 );
 
-    if( m_timeIntegrationOption != TimeIntegrationOption::QuasiStatic )
+    Group const & outputs = Group::getGroupByPath( GEOSX_FMT( "/{}", ProblemManager::groupKeysStruct().outputManager.key() ) );
```
wrtobin, line 150: Is there a cleaner/quicker way to get the `OutputManager`?

src/coreComponents/dataRepository/Group.cpp
```diff
@@ -589,7 +589,7 @@ void Group::enableLogLevelInput()
     setDescription( "Log level" );
 }
 
-Group const & Group::getBaseGroupByPath( string const & path ) const
+Group const & Group::getBaseGroupByPath( string const & path, bool hardQuery ) const
```
rrsettgast, line 592: wouldn't `getGroupPointer()` be an option here?

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.cpp
rrsettgast: I don't think so

src/coreComponents/dataRepository/Group.cpp
```diff
@@ -632,14 +632,42 @@ Group const & Group::getBaseGroupByPath( string const & path ) const
...
+        break;
+      }
     }
   }
   while( currentPosition != string::npos );
 
   return *currentGroup;
 }
 
+bool Group::hasGroupByPath( string const & path ) const
```
rrsettgast, line 650: Is this called anywhere? My search doesn't return anything???

wrtobin: No I can remove it. It was used in a previous version.

src/coreComponents/dataRepository/Group.cpp
wrtobin: The behavior could probably just be replicated with `getGroupByPath` for the containing group which is assumed to exist and then check that the subgroup exists in the parent group.

Even with that workaround (assuming a group exists -- though there are many groups it is safe to assume do exist), we don't have a concise way of saying "does this arbitrary path exist in the data repository" without erroring out if it doesn't.

rrsettgast (approved):

CusiniM:
we sure this does not require a rebaseline right?

wrtobin:
> we are sure this does not require a rebaseline right?

The only case it would modify would be one with chombo coupling active, which isn't currently covered by our testing.

PR #2333, opened 2023-03-10
Update kernelInterface.rst

IsaacJu-debug:
Found two typos while reading the doc


francoishamon (approved):
Thanks @IsaacJu-debug

PR #2337, opened 2023-03-13
Update issue templates to include EPIC issues

rrsettgast:
Add `EPIC` issue template

All the other changes are automatically generated from GitHub.


.github/ISSUE_TEMPLATE/bug-report.md
```diff
@@ -1,8 +1,8 @@
 ---
 name: Bug report
 about: Create a report to help us improve
-title: 'Your Title'
-labels: ["type: new", "type: bug"]
+title: Your Title
+labels: 'type: bug, type: new'
```
TotoGaz, line 5: Passing by, maybe we could remove the `type: new` flag.
It was an attempt and we do not use it very much, actually.

.github/ISSUE_TEMPLATE/epic.md
```diff
@@ -0,0 +1,18 @@
+---
+name: EPIC
+about: Create a new EPIC issue
+title: "[EPIC]: "
```
TotoGaz, line 4: We do not rely on keywords in the titles.
We rather filter on labels instead.

TotoGaz (approved):
If it's about testing some features with other tools, do we have to add the new template or could we simply create the issues by hand with the appropriate label?

.github/ISSUE_TEMPLATE/epic.md
TotoGaz, line 4:
```suggestion
title: "[EPIC] Your Title"
```
If we want to keep the label in the title, removing the `:` makes it lighter I think.

Issue #2320, opened 2023-03-03
[EPIC] Improve feedback to user on XML errors


MelReyCG:
Typical XML user syntax, path and naming errors  (additional commas or spaces, misplaced slash, missing fields) produce messages in the standard output which consistency and readibility by a user can be improved.

[error improvement exemples.pdf](https://github.com/GEOS-DEV/GEOS/files/11545728/error.improvement.exemples.pdf)
[general presentation.pdf](https://github.com/GEOS-DEV/GEOS/files/11545579/pres.xml.lines.dev.2.pdf)

----------------------------

- [x] **General need**: To indicate from where an error comes from, we could output it to the user in the message
GEOS PRs:
- #2404
- #2447
- #2448 
- #2449
- #2450 

----------------------------

- [x] **Case 1** - When adding a comma at start, end or between the entries of a {materialList, cellBlocks, cellBlockNames, phaseName, targetRegions}.
- Typical current output: `Controlling expression (should be false): inputStream.peek() != '}'`
- Suggested output: `"wellPressureCollection": Empty entry in line "{ water, ,rock }". Please remove a comma or add an entry. `
- GEOS PR #2357 
- LvArray PR [#280](https://github.com/GEOSX/LvArray/pull/280)

----------------------------

- [x] **Case 2.1** - Enter a wrong objectPath in a PackCollection.
- Typical current output: `std::bad_alloc` / `Segmentation fault`
- Suggested output: 
```
wellPressureCollection: Specified name (wellRegion1) did not find a match with a object in group (ElementRegions). 
Objects that are present in (ElementRegions) are:
reservoir, wellRegion
```
- GEOS PR #2341

----------------------------

- [x] **Case 2.2** - Enter a wrong objectPath in general
- Typical current output: `Group m/ElementRegion does not have a child wellRegion`
- Suggested output: ```Error in initialPressure objectPath : Group ElementRegions doesn't have a child named wellRegion123.
ElementRegions have the following children: wellRegion1,wellRegion2,meshRegion1```
- GEOS PR #2341

----------------------------

- [x] **Case 3** - Enter a wrong fieldName
- Typical current output: `Group m/domain/MeshBodies/cartesianMesh/meshLevels/Level0/ElementRegions/elementRegionsGroup/reservoir/elementSubRegions/cellBlock doesn't have a child pressures`
- Suggested output: 
```
"initialPressure": couldn'd find a field "pressures" in "ElementRegions/reservoir1"
[...]

PR #2341, opened 2023-03-16
ObjectPath XML errors feedback

MelReyCG:
This PR aims:

**1.** To prevents crashing when PathCollection::objectPath is wrong (thus, **solving the case 2.1 of the issue #2320**)

**2.** Add a context in case of a wrong PackCollection or FieldSpecification objectPath so we know from which object the error comes from (thus, **solving the case 2.2 of the issue #2320**).
In order to output the name of the Group with a wrong objectPath, this PR propose to use the GEOSX_THROW macros and try-catch to add a context to the originally thrown exception (inserting a line starting with `***** ERROR:`):
```
***** ERROR: initialPressure has a wrong objectPath. The following error have been thrown:
***** LOCATION: /appli_PITSI/MelvinRey/GEOSX_repos/GEOSX/src/coreComponents/mesh/MeshObjectPath.cpp:225
***** Controlling expression (should be false): !foundMatch
***** Rank 0: elementRegionsGroup doesn't have a child named badRegion.
elementRegionsGroup have the following children: reservoir, wellRegion
```

**3.** To fix Group::getPath() so it no longer starts with `m/` but with `/` (that `m` was the end of `Problem` which seemed wrongly cut). This fix will help a bit to output clearer path errors (this was the only use of getPath()). The returned path would be of the following form: `/Tasks/myPackCollection` (the leading `/` symbolizing it is an absolute path)
I added `testGroupPath` to test:
- the getPath() output,
- the getGroupByPath() input,
- if the Group hierarchy keeps being as we expect.


src/coreComponents/dataRepository/Group.hpp
```diff
@@ -1244,6 +1266,7 @@ class Group
 
   /**
    * @brief Return the path of this Group in the data repository.
+   * Starts with '/' followed up by the "Problem" children in which the Group is.
```
TotoGaz, line 1269: Are you sure?

src/coreComponents/dataRepository/Group.hpp
```diff
@@ -165,6 +165,16 @@ class Group
...
 
+  /**
+   * @brief @return a comma separated string containing all children name.
+   */
+  string dumpChildrenName() const;
+
+  /**
+   * @brief @return a comma separated string containing all children name.
+   */
+  string dumpWrappersName() const;
```
TotoGaz, line 176:
```suggestion
  string dumpChildrenNames() const;

  /**
   * @brief @return a comma separated string containing all children name.
   */
  string dumpWrappersNames() const;
```

src/coreComponents/dataRepository/Group.hpp
```diff
@@ -323,7 +333,10 @@ class Group
   T & getGroup( KEY const & key )
   {
     Group * const child = m_subGroups[ key ];
-    GEOSX_THROW_IF( child == nullptr, "Group " << getPath() << " doesn't have a child " << key, std::domain_error );
+    GEOSX_THROW_IF( child == nullptr,
+                    "Group " << getPath() << " has no child named " << key << std::endl <<
+                    getName() << " children are: " << dumpChildrenName(),
```
TotoGaz, line 338: Do you mean `"The children of " << getName() << " are: " << dumpChildrenName()`?

src/coreComponents/dataRepository/Group.cpp
```diff
@@ -267,6 +267,42 @@ string Group::dumpInputOptions() const
...
+    forSubGroupsIndex( [&] ( localIndex i, const Group & subGroup )
+    {
+      str += (i==0 ? subGroup.getName() : ", "+subGroup.getName());
+    } );
+    return str;
+  }
+  else
+  {
+    return "No children";
+  }
```
TotoGaz, line 284: Why don't you simply do
```c++
std::vector< string > children;
forSubGroups( [&]( Group const & subGroup ){ children.push_back( subGroup.getName() ); } );
return "{" + stringutilities::join( children, ", " ) + "}";
```
And I would make no difference for empty lists: it's empty, it's empty.

src/coreComponents/mesh/MeshObjectPath.cpp
```diff
@@ -80,14 +80,15 @@ MeshObjectPath::fillPathTokens( string const & path,
 
   int objectIndex = findObjectIndex();
 
-  GEOSX_ERROR_IF( objectIndex==-1,
-                  GEOSX_FMT( "Path ({}) does not contain a valid object type. "
-                             "Must contain one of ({},{},{},{})",
+  GEOSX_THROW_IF( objectIndex==-1,
+                  GEOSX_FMT( "Path {} does not contain a valid object type. "
```
TotoGaz, line 84: Are you sure it's more readable not to have the paths surrounded by `"` or anything?

src/coreComponents/dataRepository/Group.cpp
```diff
@@ -267,6 +267,42 @@ string Group::dumpInputOptions() const
...
+    {
+      str += (i==0 ? wrapper.getName() : ", "+wrapper.getName());
+      ++i;
+    } );
+    return str;
+  }
+  else
+  {
+    return "No children";
+  }
```
TotoGaz, line 303: same

src/coreComponents/fileIO/timeHistory/HistoryCollectionBase.cpp
```diff
@@ -84,108 +84,132 @@ bool HistoryCollectionBase::execute( real64 const time_n,
...
+      return targetGroup;
     }
-    return targetGroup;
+  }
+  catch( std::domain_error const & e )
+  {
+    throw std::domain_error( "\n***** ERROR: " + getName() +
+                             " has a wrong objectPath. The following error have been thrown:" +
+                             e.what() );
   }
```
TotoGaz, line 213: Why do you catch to resend the same (or nearly the same) exception?

src/coreComponents/unitTests/dataRepositoryTests/testGroupPath.cpp
```diff
@@ -0,0 +1,111 @@
...
+    "</Problem>";
+
+  std::vector< string > const groupPaths={
+    "/Mesh/mesh1",
+    "/domain/MeshBodies/mesh1/meshLevels/Level0/ElementRegions/elementRegionsGroup/Region2",
+    "/domain/Constitutive/shale",
+    "/Events/solverApplications",
+    "/NumericalMethods/FiniteElements/FE1",
+    "/Solvers/lagsolve",
+  };
```
TotoGaz, line 88:
```suggestion
  std::vector< string > const groupPaths{
    "/Mesh/mesh1",
    "/domain/MeshBodies/mesh1/meshLevels/Level0/ElementRegions/elementRegionsGroup/Region2",
    "/domain/Constitutive/shale",
    "/Events/solverApplications",
    "/NumericalMethods/FiniteElements/FE1",
    "/Solvers/lagsolve",
  };
```

src/coreComponents/dataRepository/Group.hpp
MelReyCG: I verified and I changed the comment to `Starts with '/' followed up by the hierarchy from the "Problem" children in which the Group is.`

src/coreComponents/fileIO/timeHistory/HistoryCollectionBase.cpp
MelReyCG: In order to add context to this exception, here is an example of this exception message:
```
***** ERROR: initialPressure has a wrong objectPath. The following error have been thrown:
***** LOCATION: /appli_PITSI/MelvinRey/GEOSX_repos/GEOSX/src/coreComponents/mesh/MeshObjectPath.cpp:225
***** Controlling expression (should be false): !foundMatch
***** Rank 0: elementRegionsGroup doesn't have a child named badRegion.
elementRegionsGroup have the following children: reservoir, wellRegion
```
This catch adds the first line, it helps to know from which XML attribute of which group the exception comes from. If the name isn't mentioned, it can be hard for a user retrieve from which Group this exception comes from.

src/coreComponents/dataRepository/Group.cpp
MelReyCG: You're right, the methods `dumpChildrenNames()` and `dumpChildrenNames()` should return nothing when no children are found.
But the user feedback is that they prefer to have a specific message when no children is found. I will output a specific message in that case.

src/coreComponents/mesh/MeshObjectPath.cpp
MelReyCG: As there are no spaces in this kind of path, the users feedback is that it doesn't imply any readability issue.
More over I think that's consistent with the other messages using getPath().
Anyway, these paths are quite hard to understand for users. What do you think of outputting the XML hierarchy instead to have a shorter and clearer one? (when applicable)

src/coreComponents/dataRepository/Group.hpp
TotoGaz: `Starts with '/' followed by the hierarchy of the children of the "Problem" in which the Group is.`?

src/coreComponents/mesh/MeshObjectPath.cpp
TotoGaz: What's most convenient w.r.t. users is really up to you, you can exchange with them better than I can.

src/coreComponents/fileIO/timeHistory/HistoryCollectionBase.cpp
TotoGaz: When you throw the initial exception, `getName` is already available, isn't it?
Can't you furbish the message once for all?

MelReyCG: The original exception is thrown from a `MeshObjectPath`, as it doesn't know which `Group` is using it to find an object, I don't think I can call `getName()` from there.

Do you think this would be cleaner ?
```
***** ERROR
***** LOCATION: /appli_PITSI/MelvinRey/GEOSX_repos/GEOSX/src/coreComponents/mesh/MeshObjectPath.cpp:225
***** Controlling expression (should be false): !foundMatch
***** initialPressure has a wrong objectPath.
elementRegionsGroup doesn't have a child named badRegion.
elementRegionsGroup have the following children: reservoir, wellRegion
```
Or are you talking about the GEOSX_THROW message ?

TotoGaz: OK; most likely there are multiple throws that I did not detect. (that's the problem of exceptions...).
Forget about this thread then!

src/coreComponents/dataRepository/Group.cpp
MelReyCG: You're right, the `dumpChildrenNames()` method should return nothing when there are no children. However, the user feedback is that they prefer to have a more verbose message, so I'm making a distinct message for that case (and doing that for dumpWrappersNames() too).

src/coreComponents/common/Logger.cpp
```diff
@@ -23,6 +23,32 @@
...
 
+std::string InputError::InsertExMsg( std::string const & originalMsg, std::string const & msgToInsert )
+{
+  std::string newMsg( originalMsg );
+
+  // we try to insert after the "***** Rank N: ", then wh try after "***** ", otherwise, we insert at the top.
+  size_t insertPos;
+  if( ( insertPos = newMsg.find( "***** Rank " ) ) != std::string::npos )
+  {
+    insertPos += 14;
```
TotoGaz, line 34: What does this 14 relate to?
Is it supposed to contain the rank id as well?

MelReyCG: Indeed, a `insertPos = newMsg.find(':', insertPos + 11 ) + 2;` will be much better here.

src/coreComponents/dataRepository/Group.hpp
```diff
@@ -335,7 +335,7 @@ class Group
     Group * const child = m_subGroups[ key ];
     GEOSX_THROW_IF( child == nullptr,
                     "Group " << getPath() << " has no child named " << key << std::endl <<
-                    getName() << " children are: " << dumpChildrenName(),
+                    ( numSubGroups()>0 ? "The children of " + getName() + " are: " + dumpChildrenNames() : getName() + " has no children." ),
```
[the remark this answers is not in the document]
MelReyCG: What do you think of putting this code into a dedicated method / function to avoid code repetition ? There are those two version (const / not const) which are duplicating this code, moreover we may re-use this output later, and that's why I put this in `dumpChildrenNames()` and `dumpWrappersNames()`.

src/coreComponents/common/Logger.cpp
TotoGaz: Then where do these 11 and 2 come from?
Are your properly managing the case where the rank id is 1, 2, 3, 4... characters long?

MelReyCG: Yes, it is properly managed with `insertPos = newMsg.find(':', insertPos + 11 ) + 2` (I tested it even with a length of 0).
I'll use `strlen()` to improve readability: 
```c++
  size_t insertPos = 0;
  // for readability purposes, we try to insert the message after the "***** Rank N: " or after "***** " instead of at the top.
  static auto constexpr rankLogStartStr =  "***** Rank ";
  static auto constexpr rankLogEndStr =  ": ";
  static auto constexpr simpleLogStartStr =  "***** ";
  if( ( insertPos = newMsg.find( rankLogStartStr ) ) != std::string::npos )
  {
    insertPos = newMsg.find( rankLogEndStr, insertPos + strlen( rankLogStartStr ) ) + strlen( rankLogEndStr );
  }
  else if( ( insertPos = newMsg.find_last_of( simpleLogStr ) ) != std::string::npos )
  {
    insertPos += strlen( simpleLogStartStr );
  }
  newMsg.insert( insertPos, msgToInsert );
```

src/coreComponents/common/Logger.cpp
untereiner, line 1: It's a shame that c++14 doesn't allow constexpr string (to use size() instead).  Since the length is know at compile time would it be better to use a static string to not use strlen ?

src/coreComponents/common/Logger.hpp
```diff
@@ -476,6 +476,21 @@ struct InputError : public std::runtime_error
...
+   */
+  InputError( std::exception const & subException, std::string const & msgToInsert ):
+    std::runtime_error( InsertExMsg( subException.what(), msgToInsert ) )
+  {}
+
+private:
+  /**
+   * @brief Returns an exception message in which we insert another message.
+   */
+  static std::string InsertExMsg( std::string const & originalMsg, std::string const & msgToInsert );
```
untereiner, line 493: why static ? could it be a free function ?

src/coreComponents/dataRepository/Group.hpp
```diff
@@ -323,7 +333,12 @@ class Group
   T & getGroup( KEY const & key )
   {
     Group * const child = m_subGroups[ key ];
-    GEOSX_THROW_IF( child == nullptr, "Group " << getPath() << " doesn't have a child " << key, std::domain_error );
+    if( child == nullptr )
+    {
+      GEOSX_THROW( "Group " << getPath() << " has no child named " << key << std::endl <<
```
untereiner, line 338: use GEOSX_FMT instead of streams ?

src/coreComponents/dataRepository/Group.hpp
```diff
@@ -334,7 +349,12 @@ class Group
   T const & getGroup( KEY const & key ) const
   {
     Group const * const child = m_subGroups[ key ];
-    GEOSX_THROW_IF( child == nullptr, "Group " << getPath() << " doesn't have a child " << key, std::domain_error );
+    if( child == nullptr )
+    {
+      GEOSX_THROW( "Group " << getPath() << " has no child named " << key << std::endl <<
```
untereiner, line 354: use GEOSX_FMT instead of streams ?

src/coreComponents/dataRepository/Group.hpp
```diff
@@ -1047,7 +1072,12 @@ class Group
   WrapperBase const & getWrapperBase( KEY const & key ) const
   {
     WrapperBase const * const wrapper = m_wrappers[ key ];
-    GEOSX_THROW_IF( wrapper == nullptr, "Group " << getPath() << " doesn't have a child " << key, std::domain_error );
+    if( wrapper == nullptr )
+    {
+      GEOSX_THROW( "Group " << getPath() << " has no wrapper named " << key << std::endl <<
```
untereiner, line 1077: use GEOSX_FMT instead of streams ?

src/coreComponents/dataRepository/Group.hpp
```diff
@@ -1058,7 +1088,12 @@ class Group
   WrapperBase & getWrapperBase( KEY const & key )
   {
     WrapperBase * const wrapper = m_wrappers[ key ];
-    GEOSX_THROW_IF( wrapper == nullptr, "Group " << getPath() << " doesn't have a child " << key, std::domain_error );
+    if( wrapper == nullptr )
+    {
+      GEOSX_THROW( "Group " << getPath() << " has no wrapper named " << key << std::endl <<
```
untereiner, line 1093: use GEOSX_FMT instead of streams ?

untereiner (changes requested):

src/coreComponents/common/Logger.cpp
TotoGaz: Maybe we can do something for c++17. If so, let's put a comment with `c++17` so we can find it when we move to c++17

src/coreComponents/dataRepository/Group.hpp
TotoGaz: I personally think that this would nest a `GEOSX_FMT` into a `GEOSX_THROW`, and that `GEOSX_FMT` forces to move back and forth the lines. I guess the added value of `fmt` is better formatting, possibility to reorder/reuse arguments, share error message templates under the form of shared strings. If any of this does not apply, I'm not sure it adds a lot to use `GEOSX_FMT`.

src/coreComponents/dataRepository/Group.hpp
```diff
@@ -165,6 +165,16 @@ class Group
...
 
+  /**
+   * @brief @return a comma separated string containing all sub groups name.
+   */
+  string dumpSubGroupsNames() const;
+
+  /**
+   * @brief @return a comma separated string containing all wrappers name.
+   */
+  string dumpWrappersNames() const;
```
TotoGaz, line 176: I can let it go like this because the pattern was already in place (`printDataHierarchy`, `dumpInputOptions`) but attaching the display format to the extraction of the data is not what we should be doing.

src/coreComponents/mesh/MeshObjectPath.cpp
```diff
@@ -123,11 +124,12 @@ MeshObjectPath::fillPathTokens( string const & path,
         existingMeshBodyAndLevel += "/n";
       } );
 
-      GEOSX_ERROR_IF( !levelNameFound,
-                      GEOSX_FMT( "Path ({}) specifies an invalid MeshBody or MeshLevel. ",
-                                 "existing MeshBodies: MeshLevels /n",
+      GEOSX_THROW_IF( !levelNameFound,
+                      GEOSX_FMT( "Path {} specifies an invalid MeshBody or MeshLevel. ",
+                                 "existing MeshBodies: {} /n",
```
TotoGaz, line 129:
```suggestion
                                 "existing MeshBodies: {} \n",
```
Is it a legit '/'?

src/coreComponents/unitTests/dataRepositoryTests/testGroupPath.cpp
```diff
@@ -0,0 +1,111 @@
...
+#include "mainInterface/ProblemManager.hpp"
+#include "mainInterface/initialization.hpp"
+#include "mainInterface/GeosxState.hpp"
+
+// TPL includes
+#include <gtest/gtest.h>
+#include <conduit.hpp>
+
+// Tests the Group::getGroup() and getPath() methods
+TEST( testGroupPath, testGlobalPaths )
```
TotoGaz, line 25: Can you add a test where you would test the inserted error message of the exception?

MelReyCG: Sure, I was wondering if I had to add tests for error messages.

src/coreComponents/mesh/MeshObjectPath.cpp
```diff
@@ -111,25 +111,36 @@ MeshObjectPath::fillPathTokens( string const & path,
...
+          existingMeshBodiesAndLevels += stringutilities::join( meshLevelsNames, ", " ) + " }\n";
+        } );
+
+        GEOSX_THROW( GEOSX_FMT( "Path {0} specifies an invalid MeshBody or MeshLevel. ",
+                                "existing MeshBodies: \n{1}\n",
+                                path,
+                                existingMeshBodiesAndLevels ),
+                     InputError );
+      }
       pathTokens.insert( pathTokens.begin()+1, unidentifiedName );
```
MelReyCG: @TotoGaz Indeed, the "/n" weren't legit here, I updated that part of the code.

src/coreComponents/mesh/MeshObjectPath.cpp
MelReyCG: Indeed, they weren't legit, I updated that part of the code in the next commit: [updating "invalid MeshBody" error message.](https://github.com/GEOSX/GEOS/pull/2341/commits/930dd64f45bcdcbf8ec1ae3340e4ed78c36d85c8)

src/coreComponents/dataRepository/Group.hpp
MelReyCG: I was thinking the same thing about this. I can stop there with for this PR, but I'm just curious about what you would prefer about where / how we could regroup these 4 methods ? A GroupView class which could reference a Group and expose displaying methods ?

src/coreComponents/common/Logger.cpp
MelReyCG: What do you think of We could introduce a constexpr strlen method which

MelReyCG: Do we have an idea of when we move to c++17?

src/coreComponents/common/Logger.hpp
MelReyCG: I changed it to a non-static method, thanks for this point.

src/coreComponents/common/Logger.hpp
```diff
@@ -490,7 +490,7 @@ struct InputError : public std::runtime_error
   /**
    * @brief Returns an exception message in which we insert another message.
    */
-  static std::string InsertExMsg( std::string const & originalMsg, std::string const & msgToInsert );
+  std::string InsertExMsg( std::string const & originalMsg, std::string const & msgToInsert );
```
TotoGaz, line 493: Where do you use `this` in this member function?
You can make it a free function only in the `cpp` file, or keep it `private static` if you want.
But I'm not sure to understand making it a member function.

TotoGaz:
> the whole approach of parsing error messages and attempting to insert text in the middle logically "backwards" and extremely brittle.

You're absolutely correct. I think we would better keep the initial information of the exception in a more meaningful way (not in a raw `string`). But I'd expect this to be a significantly larger amount of work... And maybe we would even have to deal with the `python` bindings and I'm totally unaware on how it works.

> If someone changes the message format in LvArray logging macro and replaces 5 stars with 4 or something, this functionality might be silently broken.

Most likely not, since this would result in broken unit tests.

> Not to stand in the way of merging this as is (it is an improvement for end users), but we should consider an overall change to how throwing macros work: instead of adding all the extra formatting (rank number, stars, etc.) at the source of throw, we should throw just the succinct error message and add all the other stuff at the end (when the message gets printed). This will allow us to easily "stack" multiple contexts (messages) line-by-line as the exception propagates up the call stack, without having to parse the previous text. We could even preserve the original exception object and wrap it in the new one that gets re-thrown (e.g. [nested_exception](https://en.cppreference.com/w/cpp/error/nested_exception)). 

This surely makes sense, but...

> ... if anyone has free cycles to do it.

... this is the eternal issue.

src/coreComponents/common/Logger.hpp
MelReyCG: I misunderstood the previous input of @untereiner. I'll change it to a free function in the `cpp`.

src/coreComponents/common/Logger.cpp
MelReyCG: In the meantime, I implemented the `constexpr strlen()`.

untereiner (approved):

TotoGaz (approved):

src/coreComponents/dataRepository/Group.hpp
```diff
@@ -334,7 +349,12 @@ class Group
   T const & getGroup( KEY const & key ) const
   {
     Group const * const child = m_subGroups[ key ];
-    GEOSX_THROW_IF( child == nullptr, "Group " << getPath() << " doesn't have a child " << key, std::domain_error );
+    if( child == nullptr )
```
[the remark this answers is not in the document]
MelReyCG: I forgot to change them back, I previously used `std::vector` to store the names.

PR #2345, opened 2023-03-18
Import DFN from `vtk` meshes.

TotoGaz:
This PR adds the possibility to
- Define and import a DFN from a `vtk` unstructured grid.
- Starts the support for `multi-block` `vtk` pattern for inputs.
- Imports fields onto the fractures.
- Updates the integrated tests for DFN import.

The code still does not support parallel processing.
Another PR will follow.

Documentation will be updated at the end of the process, when we're sure about the proper final choice.

Relates to https://github.com/GEOSX/GEOS/issues/2308

Related to https://github.com/GEOSX/integratedTests/pull/314


TotoGaz:
> @TotoGaz - I'm curious if you have tested this on a GPU enabled machine. I'm interested in trying this out on a large-ish model with ~1 million tets

Hello @cssherman sorry for late answer.
I did not test it on GPU, but this "import" code of this PR is not meant to be run on GPU.
All the rest remains unchanged.

Also, it's not working in parallel yet: this is my next action.

src/coreComponents/mesh/generators/MeshGeneratorBase.hpp
```diff
@@ -79,13 +79,28 @@ class MeshGeneratorBase : public dataRepository::Group
...
   /**
-   * @brief import fields from the mesh  on the array accessible via the given wrapper.
-   * @param cellBlockName name of the cell block to copy data from.
+   * @brief Describe which kind of block must be considered.
+   */
+  enum struct Block
+  {
+    VOLUMIC,
+    SURFACIC,
+    LINEIC
```
CusiniM, line 88: I am guessing this is for the wells?

src/coreComponents/mesh/generators/VTKFaceBlockUtilities.cpp
```diff
@@ -14,623 +14,621 @@
...
+    {
+      arrayView1d< globalIndex const > const l2g = m_localToGlobalMaps[i];
+      int const size = l2g.size();
+      for( int j = 0, start = 0; j < size - 1; ++j )
       {
-        std::vector< double > rawTuple( numComponents );
-        duplicatedNodes->GetTuple( i, rawTuple.data() );
-        std::set< int > nodes;
-        for( double const & v: rawTuple )
+        if( l2g[j] + 1 != l2g[j + 1] )
```
CusiniM, line 74: can you write a brief comment with the meaning of this if conditions. I am a bit lost.

src/coreComponents/mesh/generators/VTKFaceBlockUtilities.cpp
```diff
@@ -14,623 +14,621 @@
...
         {
-          // This comparison filters our dummy values.
-          if( v >= 0 )
-          { nodes.insert( int( v ) ); }
+          m_ranges[{ l2g[start], l2g[j] + 1 }] = { i, -1 };
```
CusiniM, line 76: what's the meaning of -1?

CusiniM (approved):
That `VTKFaceBlockUtilities` is quite "heavy" and I can't say that I understand all of it but, in general, it seems to make sense to me. Maybe a few more comments within the code could help. However, what matters is that it works! 

Thanks for this work, it is not easy and so important!

PR #2352, opened 2023-03-24
Changed volume constraint scaling for poromechanics

francoishamon:
This small PR changes the pore volume scaling for poromechanics. Previously we were using the latest value of porosity and neglecting the dependence on displacement in the Jacobian matrix. After this PR I propose to just use the value of porosity at the latest converged time step, which improves nonlinear convergence.

This PR requires the rebaseline of `ThermoPoroElastic_staircase_co2_smoke` because the number of nonlinear iterations is now reduced.

Rebaseline done in https://github.com/GEOSX/integratedTests/pull/309


joshua-white (approved):
Looks great!

PR #2354, opened 2023-03-24
Improve wording for travis jobs

TotoGaz:
- Being clearer about the fail-at-end process.
- Changing `pr` into `pull request`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -124,7 +124,7 @@ __geosx_return_script: &__geosx_return_script
            jq '[ .jobs[] | select( (.stage.name == "checks") and (.allow_failure == true) and (.state != "passed")) ] | length')
 
 stages:
-- check_pull_request_is_not_a_draft
+- check_that_pull_request_is_not_a_draft
 - checks
 - builds
 - fail_at_end
@@ -134,9 +134,9 @@ jobs:
   - name: code_style
   - name: documentation
   - name: check_submodules
-  - name: check_pull_request_is_assigned
+  - name: check_that_pull_request_is_assigned
   include:
-  - stage: check_pull_request_is_not_a_draft
+  - stage: check_that_pull_request_is_not_a_draft
     name: Check that the PR is not a draft (cancel job otherwise).
     <<: *__geosx_draft_script
   - stage: checks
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -124,19 +124,19 @@ __geosx_return_script: &__geosx_return_script
            jq '[ .jobs[] | select( (.stage.name == "checks") and (.allow_failure == true) and (.state != "passed")) ] | length')
 
 stages:
-- check_pr_is_not_a_draft
+- check_that_pull_request_is_not_a_draft
 - checks
 - builds
-- return_status
+- fail_at_end
```
rrsettgast, line 130: is this the final check for all jobs passing? Maybe "pass_all_tests" is a better name?

rrsettgast (approved):

.travis.yml
TotoGaz: It's the final test for all the `fail at end` tests.
If some "_normal_" tests fail, `travis` may never reach this step.

There is also a description `Checking if all the jobs have succeeded...` for the unique job the `fail_at_end` stage contains.

Maybe `check_that_fail_at_end_jobs_succeeded` is long but precise?

TotoGaz: @rrsettgast What naming would you prefer? I'm on top of the merge queue.

rrsettgast: It is still unclear to me what the final test is from the name. If any of the checks/builds fail, this fails. If they all pass, then this passes?

TotoGaz: @Bertbk As a new comer, what's your take on this?

rrsettgast: I think the problem i have is with the `fail_at_end` classification. While technically correct, it seems to indicate something of significance, but `fail_at_end` isn't the significant part. Would it be fair to say that these jobs are all "non-blocking" jobs? Meaning that their failure doesn't block the progression to the next stage of the tests. Can we call them `non_blocking_failure` jobs, and the final test would then be `non_blocking_jobs_succeeded`??

rrsettgast: I think the problem i have is with the `fail_at_end` classification. While technically correct, it seems to indicate something of significance, but `fail_at_end` isn't the significant part. Would it be fair to say that these jobs are all "non-blocking" jobs? Meaning that their failure doesn't block the progression to the next stage of the tests. Can we call them `non_blocking_failure` jobs, and the final test would then be `non_blocking_jobs_succeeded`??

TotoGaz: We can rename the `checks` stage to `non_blocking_failures`? For info, it contains
  - `code_style`
  - `documentation`
  - `check_submodules`
  - `check_that_pull_request_is_assigned`

And the very last one can be `check_that_non_blocking_jobs_succeeded` (for symmetry with other `check_*` names)?

PR #2364, opened 2023-03-29
Provide meaningful string names to Cylinder inputs

sytuannguyen:
Cylinder geometry will be extensively considered in the near future for wellbore problems. This tiny PR provides meaningfull strings to Cylinder inputs as follow:

`point1` is changed to `lowerFaceCenter`
`point2` is changed to `upperFaceCenter`

This is only changed on the user side for instant. This will help the users to better understand the input data in xml files.
Rebaseline is not needed for this PR.

Rebaseline done in https://github.com/GEOSX/integratedTests/pull/312

### DIFF
--- a/src/coreComponents/mesh/simpleGeometricObjects/Cylinder.cpp
+++ b/src/coreComponents/mesh/simpleGeometricObjects/Cylinder.cpp
@@ -43,7 +43,7 @@ Cylinder::Cylinder( const string & name, Group * const parent ):
 
   registerWrapper( viewKeyStruct::radiusString(), &m_radius ).
     setInputFlag( InputFlags::REQUIRED ).
-    setDescription( "Radius of the cylinder, outer radius of the annulus" );
+    setDescription( "Outer radius of the cylinder" );
 
   registerWrapper( viewKeyStruct::innerRadiusString(), &m_innerRadius ).
     setApplyDefaultValue( -1 ).
--- a/src/coreComponents/mesh/simpleGeometricObjects/Cylinder.cpp
+++ b/src/coreComponents/mesh/simpleGeometricObjects/Cylinder.cpp
@@ -35,7 +35,7 @@ Cylinder::Cylinder( const string & name, Group * const parent ):
 {
   registerWrapper( viewKeyStruct::point1String(), &m_point1 ).
     setInputFlag( InputFlags::REQUIRED ).
-    setDescription( "Center point the lower face of the cylinder" );
+    setDescription( "Center point of the lower face of the cylinder" );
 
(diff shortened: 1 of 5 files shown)


jhuang2601 (approved):

src/coreComponents/mesh/simpleGeometricObjects/Cylinder.hpp
```diff
@@ -72,8 +72,8 @@ class Cylinder : public SimpleGeometricObjectBase
 
   struct viewKeyStruct
   {
-    static constexpr char const * point1String() { return "point1"; }
-    static constexpr char const * point2String() { return "point2"; }
+    static constexpr char const * point1String() { return "lowerFaceCenter"; }
+    static constexpr char const * point2String() { return "upperFaceCenter"; }
```
francoishamon, line 76: @sytuannguyen Are you sure that there is the concept of "lower" and "upper" face in the `Cylinder` class? Upper and lower with respect to what?

sytuannguyen: @francoishamon You are right. How about firstFaceCenter and secondFaceCenter?

francoishamon: Yes, I think this is better

sytuannguyen: done

src/coreComponents/mesh/simpleGeometricObjects/Cylinder.cpp
```diff
@@ -35,15 +35,15 @@ Cylinder::Cylinder( const string & name, Group * const parent ):
...
 
   registerWrapper( viewKeyStruct::point2String(), &m_point2 ).
     setInputFlag( InputFlags::REQUIRED ).
-    setDescription( "Center point of the other face of the cylinder" );
+    setDescription( "Center point of the second face of the cylinder" );
 
   registerWrapper( viewKeyStruct::radiusString(), &m_radius ).
     setInputFlag( InputFlags::REQUIRED ).
-    setDescription( "Radius of the cylinder" );
+    setDescription( "Radius of the cylinder, outer radius of the annulus" );
```
francoishamon, line 46:
```suggestion
    setDescription( "Outer radius of the cylinder" );
```

francoishamon (approved):

src/coreComponents/mesh/simpleGeometricObjects/Cylinder.cpp
sytuannguyen: done

PR #2366, opened 2023-03-29
Sequential single-phase/multiphase flow-mechanics follow-up

francoishamon:
This PR implements some small changes in the flow-mechanics solvers/kernels targeting the sequential solution scheme.

In `CoupledSolver`:
- Added of a new residual-based convergence check.
- Added of some logging of nonlinear iterations (for the summary at the end).

In the solvers `SinglePhasePoromechanics` and `MultiphasePoromechanics`:
- Changed the order of the templates to force the flow solve to happen first, before the mechanics solve (otherwise, for standard field-scale injection problems initially at equilibrium the first mechanics solve is useless).

In the flow solvers:
- Added of `saveConvergedState` in a few places to make sure that (p - p_n) = 0 and (T - T_n) = 0 at the beginning of the simulation (before p_n and T_n were not initialize before the first porosity update).

In the `BiotPorosity` constitutive model:
- Changed the sign of the temperature term in the porosity update to make it consistent with the fully coupled update
- Made `thermalExpansionCoefficient` an input of the porosity model to reflect the fact that the thermal expansion coefficient appearing in the stress and in the porosity are different in the book by Coussy.

In the kernels:
- Changed the name of `FixedStressThermoPoroElasticKernel` to `FixedStressThermoPoromechanics` for consistency with the other kernels.
- Changed the density term to be a bulk density in the body force to make sure that sequential and fully coupled schemes produced to same results. This requires the computation of the bulk density outside the kernel. This should probably be refactored/improved at some point to pass a full "RHS" to the sequential mechanics kernel.

Now in the general case with gravity the sequential and fully coupled schemes produce the same results. I added a integrated test for multiphase poromechanics with gravity.

Rebaseline done in https://github.com/GEOSX/integratedTests/pull/317


CusiniM (approved):
Thanks, this is great!

inputFiles/poromechanics/PoroElastic_hybridHexPrism_co2_fim_3d.xml
```diff
@@ -0,0 +1,43 @@
+<?xml version="1.0" ?>
```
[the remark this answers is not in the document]
francoishamon: Yes, I replicated for this test what you did in Frank's PR for single-phase flow. I broke an existing test called `PoroElastic_hybridHexPrism_co2_3d` into three files: a base file, a fully coupled version and a sequential version. Both cases will be run as part of the integrated tests suite when this PR is merged.

src/coreComponents/physicsSolvers/SolverBase.cpp
```diff
@@ -604,6 +586,51 @@ bool SolverBase::lineSearchWithParabolicInterpolation( real64 const & time_n,
...
+  localMatrix.zero();
+  rhs.zero();
+
+  arrayView1d< real64 > const localRhs = rhs.open();
+
+  // The current implementation of this function is incomplete
+  // Ultimately, we want to modify the two calls below to only recompute the residual, and not the Jacobian matrix
+  // At that point, the localMatrix argument of this function will be removed
+
+  assembleSystem( time_n,
```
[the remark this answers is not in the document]
francoishamon: Yes, that would be ideal. I left it for a later PR because it seemed overwhelming, but I agree that it would be a good way to have a unique version of each kernel that could do either 1) Jacobian + residual assembly or just 2) residual assembly. That would save a lot of code and be quite useful in multiple places (line search, convergence check in the outer-loop of sequential schemes, etc)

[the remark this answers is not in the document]
klevzoff: While it's easy to just disable the final step (adding entries into the matrix), the goal IMO should be to also avoid all intermediate evaluations, stack allocations and memory accesses associated with derivatives, which I think is far from trivial. At that point, we should consider if it's easier to just have two copies of each assembly kernel (the residual-only one would be a pretty straightforward stripping down of the full one).

src/coreComponents/physicsSolvers/SolverBase.cpp
```diff
@@ -604,6 +586,51 @@ bool SolverBase::lineSearchWithParabolicInterpolation( real64 const & time_n,
   return lineSearchSuccess;
 }
 
+real64 SolverBase::assembleResidualOnlyAndCalculateNorm( real64 const & time_n,
```
klevzoff, line 589: Is the goal of combining two functions to allow solvers to fuse the two kernels together?

src/coreComponents/physicsSolvers/SolverBase.cpp
```diff
@@ -604,6 +586,51 @@ bool SolverBase::lineSearchWithParabolicInterpolation( real64 const & time_n,
...
+                           dofManager,
+                           localMatrix,
+                           localRhs );
+
+  rhs.close();
+
+  if( getLogLevel() >= 1 && logger::internal::rank==0 && !logMessage.empty() )
+  {
+    std::cout << logMessage;
+  }
```
klevzoff, line 625: This looks out of place, though I think I understand why it was done... is this just to get Line Search and residual printout closer together in the output (uninterrupted by logs from assembly functions)? Would it be too bad if we just printed this message in the Line search code before calling `assembleResidualOnlyAndCalculateNorm()`? 

On a related issue, IMO `calculateResidualNorm()` should not be printing anything, all the output in the nonlinear loop should be done by said loop directly. This would require us to return multiple values from the call (all the sub-norms to make them available for printing) so it would take a separate refactoring PR. Just noting.

inputFiles/poromechanics/PoroElastic_hybridHexPrism_co2_sequential_3d.xml
```diff
@@ -0,0 +1,55 @@
...
+      name="multiphasePoroelasticity"
+      solidSolverName="linearElasticity"
+      flowSolverName="twoPhaseFlow"
+      logLevel="1"
+      initialDt="1e11"
+      targetRegions="{ reservoir }">
+      <NonlinearSolverParameters
+        newtonTol="1.0e-8"
+	couplingType="Sequential"  
+        lineSearchAction="None"
```
joshua-white, line 21: Does the current implementation handle lineSearch cuts properly?

src/coreComponents/constitutive/solid/porosity/BiotPorosity.hpp
frankfeifan, line 1: You may also need to change `thermalExpansionCoefficient` used in `updateFromPressureTemperatureAndStrain` to `m_thermalExpansionCoefficient` defined in this porosity model, as `thermalExpansionCoefficient` therein is from the solid model.

frankfeifan (approved):
It looks good to me. Thanks!

inputFiles/poromechanics/PoroElastic_hybridHexPrism_co2_sequential_3d.xml
francoishamon: No, I don't think the current implementation can combine line search and sequential. So, after your comment I added a check to let the user know that it will not work:
https://github.com/GEOSX/GEOS/blob/471733282b52f510018a7520912c77bd0860c3e6/src/coreComponents/physicsSolvers/multiphysics/CoupledSolver.hpp#L536-L545

src/coreComponents/constitutive/solid/porosity/BiotPorosity.hpp
francoishamon: Done, thanks for the catch, I had missed it

src/coreComponents/physicsSolvers/SolverBase.cpp
francoishamon: The goal was to eliminate some redundant code in three places where we recompute the residual and calculate the norm (2 line search functions and the sequential convergence check). 

But, in the last commit I reverted my change to `SolverBase` and removed this function. I think it was odd to have this function not fully implemented (by not fully implemented, I mean still assembling the Jacobian). We can bring it back later when we have functions computing the residual only.

src/coreComponents/physicsSolvers/SolverBase.cpp
francoishamon: Yes, this logging was there to preserve the current output of the line search and not change our logging format (now reverted back to the original version).

I agree with you about the logging. @MelReyCG expressed interest in working on refactoring our logging strategy, maybe that will be a good opportunity to implement what you are suggesting

PR #2367, opened 2023-03-31
Fixed bug in the parallel computation of transmissibilities

### LINKED ISSUE #2365: Inconsistent behavior with varying number of ranks- floating point exception
Hello Everyone,
We’re trying to run a CO2 injection model in a tetrahedral grid with one well.

The model is run with:
- PANGEA-2 machine
- gcc 8.3.0 compiler
- the Geos version 250c547.

For the MPI ranks number n= 8, 14, 32 and 64, it works very well, but when we change to n=16,
we got this following error message:
image
 
Here you’ll find the simulation folder :

[SimuA.zip](https://github.com/GEOSX/GEOS/files/11101421/SimuA.zip)

francoishamon:
Users reported that the results (pressure, saturation, etc) produced by GEOS can vary quite a bit between a serial run and many-ranks runs. I used the test case posted in #2365 to identify an issue that was causing the issue.

In the computation of transmissibilities, [here](https://github.com/GEOSX/GEOS/blob/f529cfdb680b95eeabef2b0ff1494234d5e9a299/src/coreComponents/finiteVolume/TwoPointFluxApproximation.cpp#L200) we swap the indices of the element region/subregion/indices depending on the global indices, but we don't swap quantities like `stencilWeights` and `cellToFaceVec`, which results in an inconsistency [here](https://github.com/GEOSX/GEOS/blob/f529cfdb680b95eeabef2b0ff1494234d5e9a299/src/coreComponents/finiteVolume/CellElementStencilTPFA.hpp#L234). 

Swapping everything as done in this PR fixes the problem and produces consistent serial and parallel runs.

Resolves #2365 

Requires rebaseline of two integrated tests: `isothermalLeakyWell_smoke_3d_04` and `thermalLeakyWell_smoke_3d_04`

Rebaseline done in https://github.com/GEOSX/integratedTests/pull/313

### DIFF
--- a/src/coreComponents/finiteVolume/TwoPointFluxApproximation.cpp
+++ b/src/coreComponents/finiteVolume/TwoPointFluxApproximation.cpp
@@ -202,6 +202,11 @@ void TwoPointFluxApproximation::computeCellStencil( MeshLevel & mesh ) const
       std::swap( regionIndex[0], regionIndex[1] );
       std::swap( subRegionIndex[0], subRegionIndex[1] );
       std::swap( elementIndex[0], elementIndex[1] );
+      std::swap( stencilWeights[0], stencilWeights[1] );
+      std::swap( stencilStabilizationWeights[0], stencilStabilizationWeights[1] );
+      std::swap( cellToFaceVec[0][0], cellToFaceVec[1][0] );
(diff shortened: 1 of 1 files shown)


klevzoff (approved):
A very long-lived bug, courtesy of yours truly :)

TotoGaz:
So why do we do swap?

PR #2369, opened 2023-04-03
Draft: Memory Space Aware Packing

wrtobin:
Needs testing.

### DIFF
--- a/src/coreComponents/dataRepository/Wrapper.hpp
+++ b/src/coreComponents/dataRepository/Wrapper.hpp
@@ -276,24 +276,39 @@ public:
 
   ///////////////////////////////////////////////////////////////////////////////////////////////////
   /// @copydoc WrapperBase::isPackable
-  virtual
-  bool isPackable( bool onDevice ) const override
+  template < typename U = T >
+  typename std::enable_if_t< traits::HasMemorySpaceFunctions< U >, bool >
+  isPackableImpl( ) const
   {
-    if( onDevice )
+    if ( reference().getPreviousSpace() == LvArray::MemorySpace::host )
     {
-      // this isn't accurate if array/arraview return false for this, which I think they do
-      return bufferOps::can_memcpy< T >;
+      return bufferOps::is_packable< U >;
     }
     else
     {
-      return bufferOps::is_packable< T >;
+      return bufferOps::can_memcpy< U >;
     }
+  }
+
+  /// @copydoc WrapperBase::isPackable
+  template < typename U = T >
+  typename std::enable_if_t< ! traits::HasMemorySpaceFunctions< U >, bool >
+  isPackableImpl( ) const
+  {
+    return bufferOps::is_packable< U >;
+  }
+
+  /// @copydoc WrapperBase::isPackable
+  virtual
+  bool
+  isPackable( ) const override
+  {
+    return isPackableImpl( );
   }
 
   ///////////////////////////////////////////////////////////////////////////////////////////////////
   /// @copydoc WrapperBase::unpack
-  virtual
-  localIndex unpack( buffer_unit_type const * & buffer, bool withMetadata, bool onDevice, parallelDeviceEvents & events ) override final
+  localIndex unpackDeviceImpl( buffer_unit_type const * & buffer, bool withMetadata, parallelDeviceEvents & events )
   {
     localIndex unpackedSize = 0;
     if( withMetadata )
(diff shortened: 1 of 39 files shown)


src/coreComponents/dataRepository/Wrapper.hpp
```diff
@@ -276,80 +276,163 @@ class Wrapper final : public WrapperBase
 
   ///////////////////////////////////////////////////////////////////////////////////////////////////
   /// @copydoc WrapperBase::isPackable
-  virtual
-  bool isPackable( bool onDevice ) const override
+  template < typename U = T >
```
TotoGaz, line 279: Do you only want `U == T`? Is there a way to assert it?

src/coreComponents/dataRepository/Wrapper.hpp
```diff
@@ -276,80 +276,163 @@ class Wrapper final : public WrapperBase
...
+    if ( reference().getPreviousSpace() == LvArray::MemorySpace::host )
     {
-      // this isn't accurate if array/arraview return false for this, which I think they do
-      return bufferOps::can_memcpy< T >;
+      return bufferOps::is_packable< U >;
     }
     else
     {
-      return bufferOps::is_packable< T >;
+      return bufferOps::can_memcpy< U >;
```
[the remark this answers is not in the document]
wrtobin: the legacy `is_packable` was used to determine if the values were host-packable, which includes IIRC strings and other types that we cannot pack on device (ie we cannot instantiate device-packing functions for those types).

Changing these names to something that makes more sense is definitely a good idea though.

src/coreComponents/dataRepository/Wrapper.hpp
wrtobin: I'll make all the pack `impl` functions `private`, since the intent is only ever to access them through the packing interfaces, and those instantiations will only ever be `T`. 

We need to have this `U=T` template since SFINAE only works when we're using template type deduction, which if we directly use `T` which is known inside the class instantiation, we're not.

TotoGaz: Could we `static_assert` + `is_same` something or that would not be working?

wrtobin: I think as an additional layer of security to prevent internal misuse I can add a static assertion, yeah.

untereiner:
I have an issue to merge and test locally this branch because of the integratedTests

wrtobin:
Yeah I need to make some time to come back around on this and debug the runtime issues. Likely needs some traits cleanup of some sort. I have some local work put into that, but haven't had time to set aside since last week.

untereiner:
Thanks for the work! I was thinking of trying it and was hoping for more relevant feedback...

untereiner:
Sorry poor english... I was hoping to provide to **you** more relevant feedback on the work done.

untereiner:
There is the trait `is_noncontainer_type_packable` for `int` or `double` types that targets the host I guess.
and pack has a specialization for trivial types. 
```
template< bool DO_PACKING, typename T >
typename std::enable_if< std::is_trivial< T >::value, localIndex >::type
Pack( buffer_unit_type * & buffer,
      T const & var );
```
isn't this enough ?

untereiner:
@wrtobin I locally resolved the conflicts vs develop branch.. can I push the commits to this branch ?

untereiner:
> Based on the stack trace my assumption is that a piece of data to be packed wasn't able to correctly serialize (possibly calling an incorrect version of the packing functions due to SFINAE or possibly erroneously thinking it was in the wrong memory space (depending on the platform)), and probably reported either an erroneous (on at least 1 process) or 0-size (on all processes).
> 
> When going to the hdf file-writer we create subcomms for the MPIO file operations to collectively take place across. It is possible that based on the incorrect sizing information this communicator was ill-formed in some way, or simply a segmentation violation occurred during the file write due to the flawed packing and the ultimate failure happened at the next free() operation, being the comm free process.
> 
> Without attaching a debugger I can't say with any precision, the above is all just best estimate based on the changes in the branch.

I think your comment is right. The error is already present at commit #c71fb78af. But besides the bool argument and the new traits I don't see where an incorrect version of the packing could be called.

untereiner:
Hi @wrtobin, do you think you could take a look at it ?

wrtobin:
@untereiner You should be able to start trying to use this again. I fixed the bug and things seem to be working.

The primary issue was checking the device packability of the wrapper-contained type itself instead of the `decltype( referenceAsView() )` type, since device-packing operates on the type returned by that function in order to translate e.g. `arrays` to `arrayViews` automatically.

There is one bug I'm still trying to iron out that would allow me to remove the LvArray PR, since that currently only exists to allow us to bypass a host-packing which collects uninitialized memory for some reason. I'm going to try to fix that and drop the LvArray PR, but this branch should be testable now.

untereiner:
Thank you very much @wrtobin ! I'll try it!

wrtobin:
@untereiner There is still an existing bug with gathering host-side data, since we refactored the device-packing functions to allow us to pack either just underlying data (for timehistory) or both metadata and underlying data (for mpi buffer packing).

The host-side packing functions weren't similarly refactored, so packing on host will currently result in the buffer being corrupted/misaligned and the timehistory file will be garbage. 

I'm working on the minimal set of host-packing refactoring needed to get things working, I'm mostly done, but still debugging a couple things.

untereiner:
Hi @wrtobin I tested this branch with mine #2490 and its seems to work. I can output {max,min,average}pressure from regionStatistics in an hdf5 file.

src/coreComponents/dataRepository/BufferOps_inline.hpp
```diff
@@ -290,18 +404,33 @@ localIndex PackByIndex( buffer_unit_type * & buffer,
...
-             T_INDICES const & packIndices )
+             T_INDICES const & indices )
 {
   const typename MAP_TYPE::size_type length = var.size();
   localIndex sizeOfPackedChars = Pack< DO_PACKING >( buffer, length );
   for( typename MAP_TYPE::const_iterator i = var.begin(); i != var.end(); ++i )
   {
     sizeOfPackedChars += Pack< DO_PACKING >( buffer, i->first );
-    sizeOfPackedChars += PackByIndex< DO_PACKING >( buffer, i->second, packIndices );
+    sizeOfPackedChars += PackByIndex< DO_PACKING >( buffer, i->second, indices );
```
untereiner, line 433: @wrtobin it seems like you do not reuse the PackDataByIndex function you created just above by extracting the for loop

src/coreComponents/dataRepository/BufferOps_inline.hpp
```diff
@@ -274,6 +374,20 @@ PackByIndex( buffer_unit_type * & buffer,
...
+  for( localIndex a = 0; a < indices.size(); ++a )
+  {
+    T const * const data = var[indices[a]];
+    sizeOfPackedChars += PackPointerData< DO_PACKING >( buffer, data, var.sizeOfArray( indices[a] ) );
+  }
+  return sizeOfPackedChars;
+}
+
 template< bool DO_PACKING, typename T, typename T_indices >
 localIndex PackByIndex( buffer_unit_type * & buffer,
```
untereiner, line 392: same here as below

src/coreComponents/dataRepository/BufferOps_inline.hpp
wrtobin: The refactoring to internally reuse the `data` versions of these functions (packing without any metadata) from within the standard `pack` version isn't straightforward. 

The `data` versions are used in the `Wrapper` when we're packing without embedded metadata, which is the case when packing for `TimeHistory`. 

Really all that matters is that each pair of `pack/unpack` functions with the same variations work together. 

Although we don't implement `unpackDataByIndex`, since it isn't possible to unpack those structures without the embedded metadata in practice.

untereiner: it seems that something is going wrong with test testRestartBasic, when conduit is packing the nodes

untereiner: @wrtobin  is the refactoring necessary ? the tests are passing locally on my machine without it

wrtobin: Not having a `packDataByIndex` which doesn't include metadata in the serialization of the underlying data will cause the `TimeHistory` output to be corrupted when packing the data on host. 

The `TimeHistory` output requires the serialization only contain data since it buffers / writes out the entire serialized buffer of data, without inspecting / operating on it (other than a size check IIRC).

I have to look through the tests. Possibly a call is being made with the old signature, or when I modified the internal calls I mixed up a `true/false` since the old signature had two adjacent bool arguments, one for device-packing, and one to include metadata.

src/coreComponents/dataRepository/Wrapper.hpp
```diff
@@ -818,100 +777,308 @@ class Wrapper final : public WrapperBase
...
+  localIndex unpackHostImpl( buffer_unit_type const * & buffer, bool withMetadata, parallelDeviceEvents & )
+  {
+    localIndex unpackedSize = 0;
+    if( withMetadata )
+    {
+      string name;
+      unpackedSize += bufferOps::Unpack( buffer, name );
+      GEOS_ERROR_IF( name != getName(), "buffer unpack leads to wrapper names that don't match" );
+    }
+    unpackedSize += bufferOps::Unpack( buffer, *m_data );
```
untereiner, line 856: maybe `unpackedSize += bufferOps:UnpackData(buffer, *m_data);`

wrtobin:
Just pushed a fix for the ctest failure. I'm going to run regression tests on quartz and lassen and assuming they all pass I'll promote this for review / merge.

untereiner:
Great! I wasn’t able to find this one

untereiner:
Hi @wrtobin ! Are the regression tests passing ?

wrtobin:
Quartz was passing.
I haven't been able to run on Lassen yet, and I'm at a hackathon working on kernel performance on AMD until Thursday. 
I should be able to focus on Lassen testing on Friday.

untereiner:
Is Lassen passing @wrtobin ?

wrtobin:
No, but there are failing tests in develop on Lassen right now and the set of tests looks "similar", so I think this is probably fine, but I'm unwilling to merge in a change to the core infrastructure like this until our LC test platforms are all passing.

I'm hoping to take a look at and try to resolve the issues in develop early next week as we also have an OS upgrade coming for Lassen at the end of the month and we want the tests passing before the upgrade.

src/coreComponents/dataRepository/Wrapper.hpp
```diff
@@ -818,100 +777,308 @@ class Wrapper final : public WrapperBase
...
+    }
+  }
+
+  template< typename U = T >
+  typename std::enable_if_t< !traits::HasMemorySpaceFunctions< U >, bool >
+  isPackableImpl( ) const
+  {
+    static_assert( std::is_same< T, U >::value, "should only be instantiated for the wrapped type!" );
+    return bufferOps::is_host_packable< U >;
+  }
```
TotoGaz, line 805: I'm not 100% sure but maybe can you use a `if constexpr ( traits::HasMemorySpaceFunctions< U > ) { ...` instead of using `enable_if_t`?
Maybe we could even remove the need for the extra `U = T` template parameter.
And maybe having an `Impl` version for the `isPackable` function will become useless?

src/coreComponents/dataRepository/BufferOps.hpp
```diff
@@ -133,16 +166,34 @@ Pack( buffer_unit_type * & buffer,
 
 //------------------------------------------------------------------------------
 template< bool DO_PACKING, typename T, int NDIM, int USD >
-typename std::enable_if< is_packable< T >, localIndex >::type
+typename std::enable_if< is_host_packable< T >, localIndex >::type
+PackData( buffer_unit_type * & buffer,
+          ArrayView< T, NDIM, USD > const & var );
```
TotoGaz, line 171: So what is the difference between `PackData` and Pack?

src/coreComponents/dataRepository/Wrapper.hpp
```diff
@@ -818,100 +777,308 @@ class Wrapper final : public WrapperBase
...
+
+  template< typename U = T >
+  typename std::enable_if_t< traits::HasMemorySpaceFunctions< U >, bool >
+  isPackableByIndexImpl( ) const
```
TotoGaz, line 809: Same question here, and for the `unpack` flavors.

TotoGaz:
Hello @wrtobin do you think this could go in?
It's bringing an important feature that could help us moving on about `TimeHistory` and some scalar data packing which is critical to our end-users.

paveltomin:
knock knock

untereiner:
> knock knock

... knockin' on Heaven's door (?)

Hi @wrtobin ! It's almost one year old! How does the regression tests look ?

wrtobin:
There is a failure in 2 mpm tests. I anticipate it being relatively straightforward to track down tomorrow. Once that is done I will promote this from draft and place it in the queue. 

Note there are still issues with our integrated tests on GPU platforms (unrelated to this PR) that prevent exhaustive vetting, but in the interest of utility we've chosen to merge this and resolve any potential (but unexpected) correctness bugs post-merge.

PR #2380, opened 2023-04-06
Disabled time frequency for solver events

francoishamon:
I recently came across some input files from users in which `timeFrequency` was used in solver applications `PeriodicEvents`. This is not working properly because the solvers are selected the time step size, which conflicts with the time frequency specified in the XML file, resulting in skipped solver applications:
```
Time: 3000s, dt:4000s, Cycle: 3                                                                                                                                                          
Time: 7000s, dt:94000s, Cycle: 4                                                                                                                                                         
Time: 101000s, dt:4000s, Cycle: 5                                                                                                                                                        
    ( Rsolid ) = ( 9.01e-15 ) ;     ( Rflow ) = ( 1.11e-03 ) ;     ( Rwell ) = ( 3.19e-07 ) ;     ( R ) = ( 1.11e-03 ) ;                                                                 
    Attempt:  0, ConfigurationIter:  0, NewtonIter:  1  
```

This PR modifies some `Events` files to disable the `timeFrequency` and `cycleFrequency` keywords for solver events.

Requires rebaseline

### DIFF
--- a/inputFiles/solidMechanics/ExtendedDruckerPragerWellbore_smoke.xml
+++ b/inputFiles/solidMechanics/ExtendedDruckerPragerWellbore_smoke.xml
@@ -30,7 +30,7 @@
     maxTime="1.0">
     <PeriodicEvent
       name="solverApplications"
-      timeFrequency="1.0"
+      forceDt="1.0"
       targetExactTimestep="1"
       target="/Solvers/mechanicsSolver"/>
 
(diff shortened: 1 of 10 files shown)


cssherman (approved):
Looks good to me!  I'm wondering if it would be useful to add an event type, which doesn't include the timeFrequency and cycleFrequency options.  Would that be more or less confusing for users?

francoishamon:
> Looks good to me! I'm wondering if it would be useful to add an event type, which doesn't include the timeFrequency and cycleFrequency options. Would that be more or less confusing for users?

Yes, it could definitely be useful (as long as we don't require users to change their existing XML files)

PR #2385, opened 2023-04-10
Implemented a check on minimum pore volume

francoishamon:
This PR implements a check on the cell pore volume in the matrix. The user will have a warning at the beginning of the simulation if the cell pore volume is below a threshold (number of cells with a pore volume below the threshold, minimum pore volume).

Ready for review

### DIFF
--- a/src/coreComponents/physicsSolvers/fluidFlow/FlowSolverBase.cpp
+++ b/src/coreComponents/physicsSolvers/fluidFlow/FlowSolverBase.cpp
@@ -31,6 +31,7 @@
 #include "mesh/DomainPartition.hpp"
 #include "physicsSolvers/fluidFlow/FluxKernelsHelper.hpp"
 #include "physicsSolvers/fluidFlow/FlowSolverBaseFields.hpp"
+#include "physicsSolvers/fluidFlow/FlowSolverBaseKernels.hpp"
 #include "physicsSolvers/NonlinearSolverParameters.hpp"
 
 namespace geosx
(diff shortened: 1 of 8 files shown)


TotoGaz:
Is this something that could be done in `mesh_doctor`?

francoishamon:
> Is this something that could be done in `mesh_doctor`?

Yes, definitely, it would be very useful to have it in `mesh_doctor`. We also need it in GEOS because it can turn up with the internal meshes, but there are never enough checks

src/coreComponents/physicsSolvers/fluidFlow/FlowSolverBase.cpp
```diff
@@ -249,6 +250,53 @@ void FlowSolverBase::initializePreSubGroups()
   }
 }
 
+void FlowSolverBase::validatePoreVolumes( DomainPartition const & domain ) const
+{
+  real64 minPoreVolume = 0.0;
```
jhuang2601, line 255: `minPoreVolume` needs to be initialized with a non-zero value as `poreVolumeThreshold = 1e-4`; if no negative pore volume in the mesh, it will always report that the minimum pore volume is 0.

jhuang2601 (approved):
Thanks for implementing this check.

CusiniM (commented):
I wonder if we would like to also know whenever fracture elements have volumes below the threshold. They can also cause convergence issues.

src/coreComponents/physicsSolvers/fluidFlow/FlowSolverBase.cpp
francoishamon: Thanks for the catch! This is fixed now

PR #2386, opened 2023-04-11
Renaming geosx into geos

TotoGaz:
This currently is a placeholder


TotoGaz:
Replace by https://github.com/GEOS-DEV/GEOS/pull/2392

PR #2396, opened 2023-04-16
Bugfix/jmeng/validatebchighorder2

### LINKED ISSUE #2355: Field specifications not working for high-order wave propagation
**Describe the bug**
When running a wave propagation simulation at order higher than 1, no FieldSpecification can be given. GEOS exits with the following error:
```
initialPressure: there is/are no set(s) named 'all' under the objectPath 'mesh/FE1/nodeManager
```
This blocks all use of higher order wave propagation.

**To Reproduce**
1. Take the standard wave propagation smoke test,`inputFiles/wavePropagation/acous3D_small_base.xml`
2. Increase the order of the method by changing `order="1"` to `order="2"` in the `FiniteElementSpace` section of the `xml`
3. Also change the object paths of the FIeldSpecification for the variable `initialPressure` as follows: `objectPath="mesh/FE1/nodeManager"`
4. Run the smoke test `inputFiles/wavePropagation/acous3D_abc_fs_smoke.xml` that includes the edited file. `GEOSX` fails with the above error.

**Expected behavior**
The field specifications works and correctly updates the initial pressure values. If all the object paths are correctly updated in the input file, the simulation runs.

- All platforms
- `develop` version
 
**Additional context**
Probably a consequence of [PR 2291](https://github.com/GEOSX/GEOS/pull/2291)

jiemeng-total:
#### This pull request is to address the issue: #2355

To be specific: when running a wave propagation simulation at order higher than 1, no FieldSpecification can be given. GEOS exits with the following error, which blocks all use of higher order wave propagation.
```
initialPressure: there is/are no set(s) named 'all' under the objectPath 'mesh/FE1/nodeManager
```

#### The solution proposed in this PR:

In order to avoid the above problem happening, we add a new function in MeshObjectPath class namely _containsMeshLevel_ to check if the current mesh level being processed is contained in the object path. In this way, we avoid the unnecessary checking of the boundary conditions to base mesh-level in the high-order cases.

Fixes #2355


src/coreComponents/mesh/MeshLevel.cpp
```diff
@@ -202,6 +202,10 @@ MeshLevel::MeshLevel( string const & name,
       allNodes.insert( a );
     }
 
+    SortedArray< localIndex > & zposNodes  = nodeSets.registerWrapper< SortedArray< localIndex > >( string( "zpos" ) ).reference();
+    zposNodes = source.m_nodeManager->getSet("zpos");
+    m_faceManager->buildSets( *m_nodeManager );
```
francoishamon, line 207: In the simulations that use the internal mesh generator, `zpos` is always defined by default. However, in the simulations that use an imported mesh, `zpos` is not defined by default. So maybe here, we should create a set `zpos` in the high-order mesh level only it is already defined in the base mesh?

francoishamon (approved):
Looks great to me, thanks @jiemeng-total for fixing this :)

sframba (approved):

PR #2400, opened 2023-04-18
Github Actions

bmhan12:
This PR:

- Testing Travis CI workflow on Github Actions


wrtobin (approved):

.github/workflows/ci_tests.yml
```diff
@@ -0,0 +1,227 @@
+name: GEOS CI 
+on: pull_request
```
TotoGaz, line 2: I think we'll be able to trigger specific actions based on PR labels, that will be helpful.

.github/workflows/ci_tests.yml
```diff
@@ -0,0 +1,227 @@
...
+  # PR status must be "Open" to run CI.
+  check_pull_request_is_not_a_draft:
+    # We use the most recent ubuntu distribution available in Github Actions to ensure maximum support of google cloud's sdk.
+    runs-on: ubuntu-latest
+    steps:
+    - name: Check that the PR is not a draft (cancel rest of jobs otherwise)
+      run: |
+        echo "Is PR a draft?"
+        echo ${{ toJSON(github.event.pull_request.draft) }}
+        if [[ ${{ toJSON(github.event.pull_request.draft) }} == true ]]; then "false" ; else "true"; fi
```
TotoGaz, line 23: Could we use `if: github.event.pull_request.draft == false` instead?

.github/workflows/ci_tests.yml
```diff
@@ -0,0 +1,227 @@
...
+
+  # PR must be assigned to be merged.
+  # This job will fail if this is not the case.
+  check_pull_request_is_assigned:
+    needs: [check_pull_request_is_not_a_draft]
+    runs-on: ubuntu-latest
```
TotoGaz, line 29:
```suggestion
    runs-on: ubuntu-22.04
```
Please change the `ubuntu-latest`: It's a not that it matters a lot, but I think it's better if we keep the exact same system everytime.

.github/workflows/ci_tests.yml
```diff
@@ -0,0 +1,227 @@
...
+    steps:
+    # The integrated test sumbmodule repository contains large data (using git lfs).
+    # To save time (and money) we do not let Github Actions automatically clone all our (lfs) subrepositories and do it by hand.
+    - name: Checkout Repository
+      uses: actions/checkout@v3
+      with:
+        submodules: false
+        token: ${{ secrets.GITHUB_TOKEN }}
```
TotoGaz, line 51:
```suggestion
```
Do we need the token to checkout the public repository?
(same for other occurrences of this `actions`)

.github/workflows/ci_tests.yml
```diff
@@ -0,0 +1,227 @@
...
+        echo "Is PR Assigned? (Check for Assignee ID)"
+        echo $id
+        if [[ $id != null ]]; then "true" ; else "false"; fi
+
+  check_submodules:
+    needs: [check_pull_request_is_not_a_draft]
+    runs-on: ubuntu-latest
+    steps:
+    # The integrated test sumbmodule repository contains large data (using git lfs).
+    # To save time (and money) we do not let Github Actions automatically clone all our (lfs) subrepositories and do it by hand.
```
TotoGaz, line 46: It looks like there's a `lfs` option that we can set to `false` to prevent us from doing the submodule clone by hand. (in `ci_build_and_test.sh` and/or `test_submodule_updated.sh`)

scripts/ci_build_and_test.sh
```diff
@@ -0,0 +1,41 @@
...
+# Since this information is repeated twice, we use a variable.
+BUILD_DIR_MOUNT_POINT=/tmp/GEOSX
+
+
+# We need to keep track of the building container (hence the `CONTAINER_NAME`)
+# so we can extract the data from it later (if needed). Another solution would have been to use a mount point,
+# but that would not have solved the problem for the TPLs (we would require extra action to copy them to the mount point).
+CONTAINER_NAME=geosx_build
+# Now we can build GEOSX.
+while sleep 5m; do echo "... still building ..."; done &
```
TotoGaz, line 28:
```suggestion
```
This is not required anymore (it's actually been removed https://github.com/GEOS-DEV/GEOS/pull/2319)

scripts/ci_build_and_test.sh
```diff
@@ -0,0 +1,41 @@
...
+--volume=${BUILD_DIR}:${BUILD_DIR_MOUNT_POINT} \
+--cap-add=ALL \
+-e HOST_CONFIG=${HOST_CONFIG:-host-configs/environment.cmake} \
+-e CMAKE_BUILD_TYPE \
+-e GEOSX_DIR=${GEOSX_DIR} \
+-e ENABLE_HYPRE=${ENABLE_HYPRE:-OFF} \
+-e ENABLE_HYPRE_DEVICE=${ENABLE_HYPRE_DEVICE:-CPU} \
+-e ENABLE_TRILINOS=${ENABLE_TRILINOS:-ON} \
+${DOCKER_REPOSITORY}:${GEOSX_TPL_TAG} \
+${BUILD_DIR_MOUNT_POINT}/scripts/travis_build_and_test.sh ${BUILD_AND_TEST_ARGS};
```
TotoGaz, line 40: Can you indent the options please?

.github/workflows/ci_tests.yml
```diff
@@ -0,0 +1,227 @@
...
+        token: ${{ secrets.GITHUB_TOKEN }}
+
+    - id: 'auth'
+      uses: 'google-github-actions/auth@v1'
+      with:
+        credentials_json: '${{ secrets.GOOGLE_CLOUD_GCP }}'
+    - name: 'Set up Cloud SDK'
+      uses: 'google-github-actions/setup-gcloud@v1'
+      with:
+        version: '>= 363.0.0'
```
TotoGaz, line 205: Instead of having an extra `matrix`, could you rely on `GCP_BUCKET` being defined (or not) to run the extra steps for the deployment using an `if` statement?

scripts/ci_build_and_test.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+env
+
+git submodule update --init --recursive src/cmake/blt
+git submodule update --init --recursive src/coreComponents/LvArray
+git submodule update --init --recursive src/coreComponents/constitutive/PVTPackage
+git submodule update --init --recursive src/coreComponents/fileIO/coupling/hdf5_interface
```
TotoGaz, line 7: I'm pretty sure this can be removed using the `actions` without the `lfs`.

.github/workflows/ci_tests.yml
bmhan12: > Instead of having an extra `matrix`, could you rely on `GCP_BUCKET` being defined (or not) to run the extra steps for the deployment using an `if` statement?

Yes, though I wasn't sure if that was desirable. It reduces some code duplication, but then the auto deploy jobs would all be grouped together with jobs that do not upload to Google Cloud in the UI.

.github/workflows/ci_tests.yml
bmhan12: > Could we use if: github.event.pull_request.draft == false instead?

I gave that a try initially, but Github Actions takes a snapshot of the environment in each workflow.
So, let's say the draft job fails, so you change the PR from draft to Open.
When you re-run the failed job, it checks the saved draft state and fails again.

.github/workflows/ci_tests.yml
bmhan12: > I think we'll be able to trigger specific actions based on PR labels, that will be helpful.

Yes, Github Actions has alot of control over triggering workflow runs:
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows

In a future PR we can really customize/limit what triggers a CI run.

.github/workflows/ci_tests.yml
TotoGaz: We'll surely refactor all our builds very soon (playing only limited builds first, then doing all the builds when we're close to merging...).
I would say that having a condensed code will help us in this journey.

.github/workflows/ci_tests.yml
TotoGaz: Oh, this is interesting...

.github/workflows/ci_tests.yml
```diff
@@ -0,0 +1,218 @@
...
+    strategy:
+
+      # In-progress jobs will not be cancelled if there is a failure
+      fail-fast : false
+      matrix:
+        include:
+          - name: Ubuntu CUDA debug (20.04, clang 10.0.0 + gcc 9.4.0, open-mpi 4.0.3, cuda-11.2.152)
+            DOCKER_REPOSITORY: geosx/ubuntu20.04-clang10.0.0-cuda11.2.152
+            CMAKE_BUILD_TYPE: Debug
+            BUILD_AND_TEST_ARGS: "--disable-unit-tests --build-exe-only --disable-schema-deployment"
```
CusiniM, line 102: if we don't have any time limit anymore then, I think we should not need `--build-exe-only` and we could also remove `--disable-unit-tests` I think.

.github/workflows/ci_tests.yml
```diff
@@ -0,0 +1,218 @@
...
+            ENABLE_HYPRE: ON
+            ENABLE_HYPRE_DEVICE: CUDA
+            ENABLE_TRILINOS: OFF
+
+          - name: Ubuntu CUDA (20.04, clang 10.0.0 + gcc 9.4.0, open-mpi 4.0.3, cuda-11.2.152)
+            DOCKER_REPOSITORY: geosx/ubuntu20.04-clang10.0.0-cuda11.2.152
+            CMAKE_BUILD_TYPE: Release
+            BUILD_AND_TEST_ARGS: "--disable-unit-tests --disable-schema-deployment"
```
CusiniM, line 110: same comment as before.

.github/workflows/ci_tests.yml
```diff
@@ -0,0 +1,218 @@
...
+
+          - name: Ubuntu (20.04, gcc 10.3.0, open-mpi 4.0.3)
+            DOCKER_REPOSITORY: geosx/ubuntu20.04-gcc10
+            CMAKE_BUILD_TYPE: Release
+
+          # Matrix jobs that deploy to Google Cloud
+          - name: Pecan GPU (centos 7.7, gcc 8.2.0, open-mpi 4.0.1, mkl 2019.5, cuda 10.2.89p2)
+            DOCKER_REPOSITORY: geosx/pecan-gpu-gcc8.2.0-openmpi4.0.1-mkl2019.5-cuda10.2.89p2
+            CMAKE_BUILD_TYPE: Release
+            BUILD_AND_TEST_ARGS: "--disable-unit-tests --disable-schema-deployment"
```
CusiniM, line 136: same.

.github/workflows/ci_tests.yml
```diff
@@ -0,0 +1,218 @@
...
+            ENABLE_HYPRE: ON
+            ENABLE_HYPRE_DEVICE: CUDA
+            ENABLE_TRILINOS: OFF
+
+          - name: Centos (7.6, gcc 8.3.1, open-mpi 1.10.7, cuda 10.1.243)
+            DOCKER_REPOSITORY: geosx/centos7.6.1810-gcc8.3.1-cuda10.1.243
+            CMAKE_BUILD_TYPE: Release
+            BUILD_AND_TEST_ARGS: "--disable-unit-tests --disable-schema-deployment"
```
CusiniM, line 118: same

CusiniM (approved):

.github/workflows/ci_tests.yml
```diff
@@ -0,0 +1,218 @@
...
+        ENABLE_TRILINOS: ${{ matrix.ENABLE_TRILINOS }}
+        GCP_BUCKET: ${{ matrix.GCP_BUCKET }}
+      run: |
+        source ./scripts/ci_build_and_test.sh
+        TMP_DIR=/tmp
+        GEOSX_EXPORT_DIR=GEOSX-and-TPL-${GITHUB_SHA:0:7}
+        docker cp -a ${CONTAINER_NAME}:${GEOSX_TPL_DIR}/.. ${TMP_DIR}/${GEOSX_EXPORT_DIR}
+        GEOSX_BUNDLE=${TMP_DIR}/${GEOSX_EXPORT_DIR}.tar.gz
+        tar czf ${GEOSX_BUNDLE} --directory=${TMP_DIR} ${GEOSX_EXPORT_DIR}
+        CLOUDSDK_PYTHON=python3 gsutil cp -a public-read ${GEOSX_BUNDLE} gs://${GCP_BUCKET}/
```
TotoGaz, line 218: I think that eventually we should improve the `ci_build_and_test.sh` script such that it conveniently receives CLI arguments, removing the usage of implicit `env` variables.

TotoGaz (approved):
Great job @bmhan12 , moving to Github actions!

TotoGaz:
@bmhan12 I let you push your PR in the merge queue when you're confortable with it!

bmhan12:
> @bmhan12 I let you push your PR in the merge queue when you're comfortable with it!

👍 . Added to the merge queue!

rrsettgast (approved):

PR #2404, opened 2023-04-20
Add Group & Wrapper source XML line in errors

MelReyCG:
This PR aims to:
- Add the ability to know the line of xml nodes & attributes.
- Add metadata in `Group` & `Wrapper` in order to know from where they have been declared. When possible, this will allow to know in which source file and where exactly they have been defined (file line, offset), if not it will fallback to the Group hierarchy.
This metadata will allow us to formulate error messages that can be clearer and faster to solve for users ( issue #2320 ).
This PR propose to put this metadata in a `SourceContext` abstract object, so if the `Groups` & `Wrappers` are loaded differently, we will be able to change the attached information.
- Add a test to validate the Groups & Wrappers source file metadata, and also to verify if each read xml nodes and attributes correctly result in Group & Wrapper loading.

To impact the modifications on most GEOS files, 4 PR have been created to process multiple batches of files separately :
- #2447
- #2448 
- #2449
- #2450 

I planned to add another PR where I'll propose to properly wrap the following `xmlWrapper` types: `xmlNode`, `xmlAttribute`, xmlType


src/coreComponents/dataRepository/CMakeLists.txt
```diff
@@ -22,20 +22,22 @@ set( dataRepository_headers
...
-# Specify all sources
-#
-set( dataRepository_sources
+     SourceContext.hpp
+     )
+     
+     #
+     # Specify all sources
+     #
+     set( dataRepository_sources
```
untereiner, line 31: strange indentation

src/coreComponents/dataRepository/Group.cpp
```diff
@@ -158,9 +161,14 @@ void Group::processInputFileRecursive( xmlWrapper::xmlNode & targetNode )
     else
     {
       // Make sure child names are not duplicated
-      GEOS_ERROR_IF( std::find( childNames.begin(), childNames.end(), childName ) != childNames.end(),
-                     GEOS_FMT( "Error: An XML block cannot contain children with duplicated names ({}/{}). ",
-                               getPath(), childName ) );
+      if( std::find( childNames.begin(), childNames.end(), childName ) != childNames.end() )
```
untereiner, line 164: why replacing GEOS_ERROR_IF ?

src/coreComponents/unitTests/xmlTests/testXMLFile.cpp
```diff
@@ -35,6 +41,197 @@ TEST( testXML, testXMLFile )
...
+/**
+ * @brief Returns the xml nodes and attribute in hierarchy that GEOS is supposed to read in
+ * the Group hierarchy.
+ * @param document root xml document (which potentially contains includes).
+ * @param targetNode the function will be called on this node and its children.
+ * @param elementNames the list of xml elements names with the form "GroupName" or
+ * "GroupName/WrapperName". The node name is supposed to be the name attribute value, it it exists,
+ * or the tag name.
+ */
+void getGEOSElementsRecursive( xmlDocument const & document, xmlNode const & targetNode,
```
untereiner, line 53: I would suggest getElementsRecursive just to minimize search&replace in case of a new name change

untereiner (approved):

MelReyCG:
-> Merging with PR #2341 and #2357

src/coreComponents/dataRepository/Group.cpp
MelReyCG: Thanks for noticing it, it was for declaring a local variable but I forgot to change it back.

src/coreComponents/dataRepository/DataContext.hpp
```diff
@@ -0,0 +1,205 @@
...
+ *
+ * See top level LICENSE, COPYRIGHT, CONTRIBUTORS, NOTICE, and ACKNOWLEDGEMENTS files for details.
+ * ------------------------------------------------------------------------------------------------------------
+ */
+
+/**
+ * @file DataContext.hpp
+ */
+
+#ifndef GEOS_DATAREPOSITORY_SOURCECONTEXT_HPP_
```
jeannepellerin, line 19: To update to GEOS_DATAREPOSITORY_DATACONTEXT_HPP_
 as well as the endif comment

src/coreComponents/dataRepository/DataContext.hpp
```diff
@@ -0,0 +1,205 @@
...
+namespace dataRepository
+{
+
+class Group;
+class WrapperBase;
+
+/// This abstract class stores data that helps to retrieve an object:
+/// - from which position in a file (if applicable), see DataFileContext,
+/// - where it is located in the data hierarchy, see GroupContext and WrapperContext.
+/// Typically, the target object contain an unique_ptr< DataContext > instance of this class.
```
jeannepellerin, line 38: /// DataContext is an abstract class to storing contextual information on an object:
/// - its line position in a file if applicable, implementation in DataFileContext,
/// - its location in the data hierarchy, implementation in GroupContext and WrapperContext.
/// The target object contains a unique_ptr< DataContext > instance.

src/coreComponents/dataRepository/DataContext.hpp
```diff
@@ -0,0 +1,205 @@
...
+  virtual string toString() const = 0;
+
+  string getObjectName() const
+  { return m_objectName; }
+
+  /**
+   * @brief In some cases, we need to know if a DataContext is from a file. It means that it
+   * is a more user-friendly information compared to other DataContext classes.
+   * @return true if the context is from a file.
+   */
```
jeannepellerin, line 63: /**
   * @brief Flag on availability of a file. Used to provide more user-friendly information 
   * @return true if the context is built from a file.
   */

src/coreComponents/dataRepository/DataContext.hpp
```diff
@@ -0,0 +1,205 @@
...
+};
+
+/// Helps to know where a Group is in the hierarchy.
+/// See DataContext class for more infos.
+class GroupContext : public DataContext
+{
+public:
+
+  /**
+   * @brief Construct a new GroupContext object
```
jeannepellerin, line 89: All  Construct probably take an S

src/coreComponents/dataRepository/DataContext.cpp
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * See top level LICENSE, COPYRIGHT, CONTRIBUTORS, NOTICE, and ACKNOWLEDGEMENTS files for details.
+ * ------------------------------------------------------------------------------------------------------------
+ */
+
+/**
+ * @file DataContext.cpp
+ */
+
+//#include "codingUtilities/StringUtilities.hpp"
```
jeannepellerin, line 19: in or out ?

src/coreComponents/dataRepository/DataContext.hpp
```diff
@@ -0,0 +1,205 @@
...
+  WrapperContext( WrapperBase & wrapper );
+
+  /**
+   * @copydoc DataContext::toString()
+   */
+  virtual string toString() const;
+
+};
+
+/// Helps to know from where a Group or a Wrapper has been declared in its source file (a xml typically).
```
jeannepellerin, line 136: Stores information to retrieve where a Group or Wrapper has been declared in the input source file (e.g. XML)

src/coreComponents/dataRepository/DataContext.hpp
```diff
@@ -0,0 +1,205 @@
...
+   */
+  GroupContext( Group & group );
+
+  /**
+   * @brief Get the target Group object (wrapper's parent in case of a WrapperContext).
+   */
+  Group & getGroup() const;
+
+  /**
+   * @copydoc DataContext::toString()
```
jeannepellerin, line 100: Return the Group path ?

src/coreComponents/dataRepository/DataContext.hpp
```diff
@@ -0,0 +1,205 @@
...
+protected:
+
+  /**
+   * @brief Construct a new GroupContext object
+   * @param group see getGroup()
+   * @param objectName Target object name.
+   */
+  GroupContext( Group & group, string const & objectName );
+
+  /// see getGroup()
```
jeannepellerin, line 113: A reference to the Group related to this GroupContext

src/coreComponents/dataRepository/DataContext.hpp
```diff
@@ -0,0 +1,205 @@
...
+  /**
+   * @brief Construct the file context of a Group from an xml node.
+   */
+  DataFileContext( Group & group, xmlWrapper::xmlNodePos const & nodePos, string const & nodeTagName );
+  /**
+   * @brief Construct the file context of a Group from an xml attribute.
+   */
+  DataFileContext( WrapperBase & wrapper, xmlWrapper::xmlAttributePos const & attPos );
+
+  virtual string toString() const;
```
jeannepellerin, line 151: Comment:

Returns where in the source DataFile the Group or Wrapper was declared, file name along with the line or offset.

src/coreComponents/dataRepository/Group.cpp
```diff
@@ -127,22 +130,22 @@ string Group::getPath() const
 {
   // In the Conduit node heirarchy everything begins with 'Problem', we should change it so that
```
jeannepellerin, line 131: hierarchy

src/coreComponents/unitTests/xmlTests/testXMLFile.cpp
```diff
@@ -35,6 +41,197 @@ TEST( testXML, testXMLFile )
...
+  }
+
+  std::set< string > expectedElements;
+  getElementsRecursive( xmlDocument, xmlDocument.root().child( "Problem" ), expectedElements );
+
+  std::set< string > verifiedElements;
+  verifyGroupDataFileContextRecursive( xmlDocument, problemManager, verifiedElements );
+
+  std::set< string > notFound = getDifference( expectedElements, verifiedElements );
+  EXPECT_TRUE( notFound.empty() ) << "Infos : There should not exists xml element that were not in "
```
jeannepellerin, line 223: exist

src/coreComponents/unitTests/xmlTests/testXMLFile.cpp
```diff
@@ -35,6 +41,197 @@ TEST( testXML, testXMLFile )
...
+                                     "the Group hierarchy.\nNot in Group hierarchy : {"
+                                  << stringutilities::join( notFound, "," ) << "}";
+
+  std::set< string > notExpected = getDifference( verifiedElements, expectedElements );
```
jeannepellerin, line 227: exist

src/coreComponents/unitTests/xmlTests/testXMLFile.cpp
jeannepellerin, line 223: No S  to Information / info

src/coreComponents/unitTests/xmlTests/testXMLFile.cpp
```diff
@@ -35,6 +41,197 @@ TEST( testXML, testXMLFile )
...
+  auto it = std::set_difference( setA.begin(), setA.end(),
+                                 setB.begin(), setB.end(),
+                                 result.begin() );
+  result.resize( it - result.begin() );
+
+  return std::set< string >( result.begin(), result.end() );
+}
+
+// Tests
+// - if the line information of each nodes and attributes can be retrieved,
```
jeannepellerin, line 195: all

src/coreComponents/mesh/MeshObjectPath.cpp
```diff
@@ -212,12 +231,13 @@ void processTokenRecursive( dataRepository::Group const & parentGroup,
 
       }
     }
-    GEOS_ERROR_IF( !foundMatch,
-                   GEOS_FMT( "Specified name ({0}) did not find a match with a object in group ({1}). "
-                             "Objects that are present in ({1}) are:\n{2}",
-                             inputEntry,
+    GEOS_THROW_IF( !foundMatch,
+                   GEOS_FMT( "{0} doesn't have a child named {1}.\n"
+                             "{0} have the following children: {{ {2} }}",
```
jeannepellerin, line 236: "{0} has no child named {1}.\n"
"{0} has the following children: {{ {2} }}",

src/coreComponents/dataRepository/xmlWrapper.hpp
```diff
@@ -61,6 +59,179 @@ using xmlAttribute = pugi::xml_attribute;
...
+  /**
+   * @brief Constructor of this struct.
+   */
+  xmlAttributePos( string const & filePath, size_t line, size_t offsetInLine, size_t offset );
+  /**
+   * @return false if the position is undetermined.
+   */
+  bool isFound() const;
+  /**
+   * @return returns a string to locate the node (typically the source file and line).
```
jeannepellerin, line 88: Returns a string containing the file name and position in the file.

src/coreComponents/dataRepository/xmlWrapper.hpp
```diff
@@ -61,6 +59,179 @@ using xmlAttribute = pugi::xml_attribute;
...
+   */
+  string toString() const;
+};
+
+/// Stores the source document, file path, and position in the file of a xml node.
+/// It can help to retrieve the position (xmlAttributePos) of a xml attribute.
```
jeannepellerin, line 94: Used to retrieve the position of dataRepository::Wrapper in XML.

src/coreComponents/dataRepository/xmlWrapper.hpp
```diff
@@ -61,6 +59,179 @@ using xmlAttribute = pugi::xml_attribute;
...
+{
+  /// Reference of the main xmlDocument that contains all original file buffers.
+  xmlDocument const & document;
+
+  /**
+   * @brief Constructor of this struct.
+   */
+  xmlNodePos( xmlDocument const & document, string const & filePath, size_t line, size_t offsetInLine, size_t offset );
+  /**
+   * @return false if the position is undetermined.
```
jeannepellerin, line 105: undefined

src/coreComponents/dataRepository/xmlWrapper.hpp
```diff
@@ -61,6 +59,179 @@ using xmlAttribute = pugi::xml_attribute;
...
+  map< string, string > const & getOriginalBuffers() const;
+  /**
+   * @brief If load_file() has been loaded, returns the path of the source file.
+   * If another load method has been called, it returns a generated unique value.
+   */
+  string const & getFilePath() const;
+  /**
+   * @brief Compute the position of the given node if the node file information are loaded.
+   * @throws an InputError if the node position couldn't be determined despite the node source file
+   * informations being loaded.
```
jeannepellerin, line 150: @brief Compute the position of the given node.
 @throws an InputError if the node position is not found and source file is loaded.

src/coreComponents/dataRepository/DataContext.hpp
```diff
@@ -0,0 +1,270 @@
...
+{
+
+class Group;
+class WrapperBase;
+
+/// DataContext is an abstract class storing contextual information on an object:
+/// - its line position in a file, if applicable, implementation in DataFileContext,
+/// - its location in the data hierarchy, implementation in GroupContext and WrapperContext.
+/// Typically, the target object contains an unique_ptr< DataContext > instance of this class.
+class DataContext
```
TotoGaz, line 40: So it's a `HierarchyContext`?

src/coreComponents/dataRepository/DataContext.hpp
```diff
@@ -0,0 +1,270 @@
...
+
+  /**
+   * @brief Destroy the WrapperContext object
+   */
+  virtual ~WrapperContext() {}
+
+  /**
+   * @return the parent group DataContext followed by the wrapper name.
+   */
+  virtual string toString() const;
```
TotoGaz, line 151: Would something like https://en.cppreference.com/w/cpp/language/cast_operator be relevant?

src/coreComponents/dataRepository/DataContext.hpp
```diff
@@ -0,0 +1,270 @@
...
+   * @brief Get the target object name
+   */
+  string getObjectName() const
+  { return m_objectName; }
+
+  /**
+   * @brief Flag on availability of file information. Used to provide more user-friendly information.
+   * @return true if the context is from a file.
+   */
+  bool isDataFileContext() const
```
TotoGaz, line 72: I think we'll have to carefully have a look at this.

src/coreComponents/dataRepository/DataContext.hpp
```diff
@@ -0,0 +1,270 @@
...
+
+} /* namespace geos */
+
+
+
+/// @cond DO_NOT_DOCUMENT
+#ifdef GEOSX_USE_FMT
+#define GEOS_FMT_NS_PREFIX fmt
+#else
+#define GEOS_FMT_NS_PREFIX std
```
TotoGaz, line 242: Aren't those already defined somewhere else?

TotoGaz (commented):
First batch of comments.

src/coreComponents/unitTests/xmlTests/testXMLFile.cpp
```diff
@@ -35,6 +41,197 @@ TEST( testXML, testXMLFile )
...
+    string const & inputFileName = commandLine.getReference< string >( problemManager.viewKeys.inputFileName );
+    xmlDocument.load_file( inputFileName.c_str(), true );
+    problemManager.parseXMLDocument( xmlDocument );
+  }
+
+  GEOS_LOG( "Loaded files : " );
+  for( auto const & buffer: xmlDocument.getOriginalBuffers() )
+  {
+    GEOS_LOG( "    " << buffer.first << " (" << buffer.second.size() << " chars)" );
+  }
```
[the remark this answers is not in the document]
MelReyCG: I thought these logs could be useful for test debugging purposes, in order to manually validate the loaded files in case of a problem (especially nested inclusion issues).

src/coreComponents/dataRepository/DataContext.hpp
MelReyCG: The `DataContext` have the following children:
- `GroupContext` and `WrapperContext`, which could be qualified of `HierarchyContext`.
- `DataFileContext`, which aims to give an object its source file contextual information (file / line / offset). 

I'm changing the comment here as it can be confusing:
```
/// DataContext is an abstract class storing contextual information on an object:
/// - either its line position in a file, if applicable, implementation in DataFileContext,
/// - or its location in the data hierarchy, implementation in GroupContext and WrapperContext.
/// Typically, the target object contains an unique_ptr< DataContext > instance of this class.
class DataContext
```

Do you think I should rename `DataContext` to `HierarchyContext`?

src/coreComponents/dataRepository/DataContext.hpp
MelReyCG: I hesitated to use it but it seemed like a bad practice. Implicit casts can sometimes lead to unintentional implicit conversions, and explicit casts would not really be shorter here.

src/coreComponents/dataRepository/DataContext.hpp
MelReyCG: Oh ! `GEOS_FMT_NS` changed from `::fmt` to `fmt` (and `std` define too).

The old form forced me to add these defines, but they are no longer relevant.

src/coreComponents/dataRepository/DataContext.hpp
```diff
@@ -0,0 +1,270 @@
...
+#include "common/Format.hpp"
+
+namespace geos
+{
+
+namespace dataRepository
+{
+
+class Group;
+class WrapperBase;
```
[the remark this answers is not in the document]
MelReyCG: Why preferring includes over forward declaration? I had this habit from MSVC to lighten compilation and to solve the problem of circular references.

src/coreComponents/unitTests/xmlTests/testXMLFile.cpp
```diff
@@ -35,6 +41,197 @@ TEST( testXML, testXMLFile )
...
+  std::set< string > verifiedElements;
+  verifyGroupDataFileContextRecursive( xmlDocument, problemManager, verifiedElements );
+
+  std::set< string > notFound = getDifference( expectedElements, verifiedElements );
+  EXPECT_TRUE( notFound.empty() ) << "Info : There should not exist xml element that were not in "
+                                     "the Group hierarchy.\nNot in Group hierarchy : {"
+                                  << stringutilities::join( notFound, "," ) << "}";
+
+  std::set< string > notExpected = getDifference( verifiedElements, expectedElements );
+  EXPECT_TRUE( notExpected.empty() ) << "Info : There should not exist an object in the Group "
```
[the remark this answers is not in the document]
MelReyCG: `notExpected` is the list of unexpected elements (which is a set filled by the elements which are found in the data hierarchy but not in the xml hierarchy), it is expected to be empty.

src/coreComponents/dataRepository/DataContext.hpp
MelReyCG: What problem do you see here?

src/coreComponents/dataRepository/DataContext.cpp
```diff
@@ -0,0 +1,136 @@
...
+  m_offset( attPos.offset )
+{}
+
+string DataFileContext::toString() const
+{
+  std::ostringstream oss;
+  oss << m_objectName;
+  if( m_line != xmlWrapper::xmlDocument::npos )
+  {
+    oss << " (" << splitPath( m_filePath ).second << ", l." << m_line << ")";
```
wrtobin, line 121: Instead of 

`<filename>, l.<#>`, 

might we consider

`<filename>:<#>` 

since many editors / IDEs have smart-open-and-jump-to-line capabilities on file + line strings of that format.

MelReyCG: I agree that `<filename>:<#>` is better for us developers.
I've discussed it with users of different levels of experience, and for them, it's clear that the `<filename>, l.<#>` format helps them to pinpoint the error much more easily and quickly. Also, they are not accustomed to using IDEs.

TotoGaz (commented):
A first batch of comments

src/coreComponents/unitTests/xmlTests/testXMLFile.cpp
```diff
@@ -35,6 +41,197 @@ TEST( testXML, testXMLFile )
...
+  static const std::set< string > xmlOnlyNodes {
+    "ElementRegions"
+  };
+
+  // The Group name will be the name attribute value, or the node tag name if the name attribute
+  // doesn't exist.
+  string groupName = [&]() {
+    xmlAttribute nameAtt = targetNode.attribute( "name" );
+    return nameAtt ? string( nameAtt.value() ) : string( targetNode.name() );
+  }();
```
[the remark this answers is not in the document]
MelReyCG: Indeed, I'll keep the lamda to be able to const declare `groupName`.

src/coreComponents/dataRepository/xmlWrapper.cpp
```diff
@@ -71,12 +73,36 @@ template void stringToInputVariable( Tensor< real32, 3 > & target, string const
...
 
-void addIncludedXML( xmlNode & targetNode, int const level )
+/**
+ * @brief Adds the filePath and character offset info on the node in filePathString
+ * and charOffsetString attributes. This function allow to keep track of the source
+ * filename & offset of each node.
+ * @param node the target node to add the informations on.
+ * @param filePath the absolute path of the xml file containing the node.
+ */
+void addNodeFileInfo( xmlNode node, string const & filePath )
```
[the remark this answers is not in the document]
MelReyCG: Indeed, the `pugi::xmlNode` objects are kind of "pointers" to an internal node object (i.e. unique_ptr). Passing It by value in a function will modify the original node.
I'll document that behaviour on the `xmlWrapper::xmlNode` type.

src/coreComponents/dataRepository/DataContext.hpp
```diff
@@ -0,0 +1,261 @@
...
+  /**
+   * @brief Destroy the DataContext object
+   */
+  virtual ~DataContext() {}
+
+  /**
+   * @return A string that mention all the known informations to retrieve from where the target
+   * object comes from.
+   */
+  virtual string toString() const = 0;
```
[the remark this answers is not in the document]
MelReyCG: I will need it in some places that directly need a `string` as parameter, like exception constructors. It can make the code shorter at these places.

src/coreComponents/dataRepository/DataContext.hpp
```diff
@@ -0,0 +1,261 @@
...
+
+#ifndef GEOS_DATAREPOSITORY_DATACONTEXT_HPP_
+#define GEOS_DATAREPOSITORY_DATACONTEXT_HPP_
+
+#include "common/DataTypes.hpp"
+#include "common/Logger.hpp"
+#include "xmlWrapper.hpp"
+#include "common/Format.hpp"
+#include "Group.hpp"
+#include "WrapperBase.hpp"
```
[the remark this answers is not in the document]
MelReyCG: I think we are in the same situation that `Wrapper` and `Group` are: we need to access the `DataContext` from a `Group`, and we need a `Group` to initialize the `DataContext`. That's why I used the same technique to solve it (the forward declaration).

Another way to solve it would be to move the declaration of GroupContext & WrapperContext in a dedicated `GroupContext.hpp` file which would include `Group.hpp`. I would also construct `DataFileContext` directly from strings produced from outside of the `DataFileContext` class.

src/coreComponents/dataRepository/xmlWrapper.cpp
```diff
@@ -169,6 +195,207 @@ string buildMultipleInputXML( string_array const & inputFileList,
...
+xmlResult xmlDocument::load_string( const pugi::char_t * contents, bool loadNodeFileInfo,
+                                    unsigned int options )
+{
+  xmlResult result = pugi::xml_document::load_string( contents, options );
+
+  // keeping a copy of original buffer to allow line retrieval
+  if( loadNodeFileInfo )
+  {
+    new (&m_originalBuffers) map< string, string >();
+    m_originalBuffers[m_rootFilePath] = string( contents );
```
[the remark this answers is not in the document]
MelReyCG: Because the `xmlDocument::load_*` methods destroy the current xml tree. I need to free the previously loaded buffers to keep being synchronized with the loaded xml tree.

I'll add *"Free the existing tree then load ..."* to the comments of `xmlDocument::load_*` methods.

src/coreComponents/dataRepository/GroupContext.cpp
```diff
@@ -0,0 +1,84 @@
...
+      {
+        path.insert( 0, '/' + parentGroup->getName() + '(' +
+                     splitPath( parentContext.getFilePath() ).second +
+                     ",l." + std::to_string( parentContext.getLine() ) + ')' );
+        foundNearestLine=true;
+      }
+      else
+      {
+        path.insert( 0, '/' + parentGroup->getName() );
+      }
```
TotoGaz, line 55: Is it possible to change the architecture such that you can ask `parentContext.getDescription()` (or anything) that returns what it needs, depending on its real type, without having to down cast to the real type? Something like
```c++
for( Group const * parentGroup = &m_group; parentGroup->hasParent(); parentGroup = &parentGroup->getParent() )
{
    path.insert( 0, '/' + parentGroup->getName()  + parentContext.getDescription() );
}
```
whatever the real type of `DataContext` is?

That way, we could get rid of the `isDataFileContext` and the downcasting technique.
I could spot usages in the `tests` which we surely can adapt.
And another one in the `WrapperContext`, which we can surely adapt in the same way.

Can you investigate this?

src/coreComponents/dataRepository/GroupContext.cpp
```diff
@@ -64,19 +52,14 @@ string GroupContext::toString() const
...
-  if( m_group.getDataContext().isDataFileContext() )
-  {
-    return m_group.getDataContext().toString() + ", attribute " + m_targetName;
-  }
-  else
-  {
-    return m_group.getDataContext().toString() + "/" + m_targetName;
-  }
+  DataContext const & parentDC = m_group.getDataContext();
+  return parentDC.toString() + parentDC.getWrapperSeparator() + m_typeName;
```
TotoGaz, line 62:
```suggestion
  return m_group.getDataContext().getWrapperContext( m_typeName );
```
with something like
```c++
string DataContext::getWrapperContext( string typeName ) const {
  return this->toString() + this->getWrapperSeparator() + typeName;
}
```
which can hide `getWrapperSeparator`?

Does having a `DataContext::getWrapperSeparator` make sense?

src/coreComponents/dataRepository/GroupContext.cpp
```diff
@@ -38,21 +38,9 @@ string GroupContext::toString() const
   bool foundNearestLine = false;
   for( Group const * parentGroup = &m_group; parentGroup->hasParent(); parentGroup = &parentGroup->getParent() )
   {
-    if( !foundNearestLine && parentGroup->getDataContext().isDataFileContext() )
+    if( !foundNearestLine )
```
TotoGaz, line 41: Would we have something like
```c++
string const path = foundNearestLine ? parentGroup->getName() : parentGroup->getDataContext().getTargetNameInPath( foundNearestLine );
path.insert( 0, '/' + path );
```
make sense?
Would it make sense, based on the value of `foundNearestLine` to make `getTargetNameInPath` return either `parentGroup->getName()` or its actual content?
I do not have the big picture so I do not know... Tell me what you think.

I'm surely mistaken, but it looks to me that `foundNearestLine` is intended for optimization mostly?

src/coreComponents/dataRepository/DataContext.hpp
```diff
@@ -62,12 +61,11 @@ class DataContext
   string getTargetName() const
   { return m_targetName; }
 
-  /**
-   * @brief Flag on availability of file information. Used to provide more user-friendly information.
-   * @return true if the context is from a file.
-   */
-  bool isDataFileContext() const
-  { return m_isDataFileContext; }
+  virtual string getTargetNameInPath( bool & fileLineFound ) const
```
TotoGaz, line 64: Can you explain the `fileLineFound`? I do not really understand why we need it.

src/coreComponents/dataRepository/xmlWrapper.cpp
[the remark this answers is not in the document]
MelReyCG: I mistakenly thought that `std::map::clear()` does not clear all used memory, since it worked that way for `std::unordered_map::clear()`, and I forgot that the `new` syntax doesn't call the destructor.
I'll use `clear()` here.

src/coreComponents/dataRepository/GroupContext.cpp
MelReyCG: I submit another proposal where I added this version of `toString()` which takes in parameter the name of an inner object to append at the end of the simple `toString()` result:
```cpp
  /**
   * @return The result of toString() properly suffixed by the name of a contained object.
   */
  virtual string toString( string const & innerObjectName ) const
  { return toString() + '/' + innerObjectName; }
```

DataFileContext overrides this method:
```cpp
  string toString(string const & innerObjectName) const override
  { return toString() + ", attribute " + innerObjectName; }
```

This way, we no longer need `getWrapperSeparator()`.

src/coreComponents/dataRepository/GroupContext.cpp
MelReyCG: `foundNearestLine` is used here to only insert the line of the nearest `DataFileContext`. The goal is to not output the line of the further Groups because, according to our observations, they does not help the user.
I previously used `isDataFileContext()` for the same purpose, because I need a way to determine if a file line has already been found during the path construction.

src/coreComponents/dataRepository/DataContext.hpp
MelReyCG: The `fileLineFound` helps to know if the target name contains file line infos. It is used by `GroupContext::toString()`.
More info in this reply: https://github.com/GEOS-DEV/GEOS/pull/2404/#discussion_r1250949982

src/coreComponents/dataRepository/xmlWrapper.hpp
```diff
@@ -46,21 +46,220 @@ namespace geos
...
+};
+
+/// Wrapper class for the type of xml document.
+/// This class exists to intercept file / string loading methods, and to keep the loaded buffers,
+/// in order to retrieve the source file and line of nodes and attributes.
+class xmlDocument : public pugi::xml_document
+{
+public:
+  /// Error value for when an offset / line position is undefined.
+  static const size_t npos;
```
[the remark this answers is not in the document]
MelReyCG: So we have:
```cpp
  static constexpr size_t npos = string::npos;
```
... in the `.hpp`, and: 
```cpp
constexpr size_t xmlDocument::npos;
```
... in the `.cpp` (we need to keep this second one because of external use).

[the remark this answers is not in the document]
klevzoff: Yeah you're right

src/coreComponents/dataRepository/xmlWrapper.cpp
```diff
@@ -169,6 +195,207 @@ string buildMultipleInputXML( string_array const & inputFileList,
...
+{
+  string const * buffer = document.getOriginalBuffer( filePath );
+  size_t tagEnd = xmlDocument::npos;
+  size_t attOffset = xmlDocument::npos;
+  size_t attLine = xmlDocument::npos;
+  size_t attOffsetInLine = xmlDocument::npos;
+
+  if( isFound() && buffer != nullptr && offset < buffer->size() )
+  {
+    tagEnd = buffer->find( '>', offset );
```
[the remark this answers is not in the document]
MelReyCG: I pushed a fix to:
- ignore attributes and `>` that are inside attribute value quotes,
- search an attribute by searching its name, followed by `=` and `"`
- the default line info of an attribute are now those of the parent node. If something goes wrong here, the user will be redirected to the containing node.

This portion of code surely does not take everything that can appear in an xml into account, but at least, GEOS is now spaceship-proof!

src/coreComponents/dataRepository/xmlWrapper.hpp
```diff
@@ -46,21 +46,220 @@ namespace geos
...
+   * @param attName the name of the attribute to locate
+   * @return an xmlAttributePos object that represents the position of the target node.
+   */
+  xmlAttributePos getAttributeLine( string const & attName ) const;
+};
+
+/// Wrapper class for the type of xml document.
+/// This class exists to intercept file / string loading methods, and to keep the loaded buffers,
+/// in order to retrieve the source file and line of nodes and attributes.
+class xmlDocument : public pugi::xml_document
```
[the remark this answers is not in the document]
MelReyCG: Done!

src/coreComponents/dataRepository/xmlWrapper.hpp
```diff
@@ -46,21 +46,230 @@ namespace geos
...
+  /**
+   * @brief Load document from zero-terminated string. No encoding conversions are applied.
+   * Free any previously loaded xml tree.
+   * Wrapper of pugi::xml_document::loadBuffer() method.
+   * @param contents the string containing the document content
+   * @param loadNodeFileInfo Load the node source file info, allowing getNodePosition() to work.
+   * @param options the parsing options
+   * @return an xmlResult object representing the parsing resulting status.
+   */
+  xmlResult loadString( const pugi::char_t * contents, bool loadNodeFileInfo = false );
```
TotoGaz, line 196:
```suggestion
  xmlResult loadString( string const & contents, bool loadNodeFileInfo = false );
```
?

src/coreComponents/dataRepository/xmlWrapper.hpp
```diff
@@ -46,21 +46,230 @@ namespace geos
...
+  /**
+   * @brief Load document from file. Free any previously loaded xml tree.
+   * Wrapper of pugi::xml_document::loadBuffer() method.
+   * @param path the path of an xml file to load.
+   * @param loadNodeFileInfo Load the node source file info, allowing getNodePosition() to work.
+   * @param options the parsing options
+   * @param encoding the encoding options
+   * @return an xmlResult object representing the parsing resulting status.
+   */
+  xmlResult loadFile( const char * path, bool loadNodeFileInfo = false );
```
TotoGaz, line 207:
```suggestion
  xmlResult loadFile( string const & path, bool loadNodeFileInfo = false );
```
?
etc.

TotoGaz (commented):
Comments passing by.

src/coreComponents/dataRepository/DataContext.hpp
```diff
@@ -0,0 +1,202 @@
...
+protected:
+
+  /// @see getObjectName()
+  string const m_targetName;
+
+};
+
+/// Stores information to retrieve where a target object has been declared in the input source
+/// file (e.g. XML).
+class DataFileContext final : public DataContext
```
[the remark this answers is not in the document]
MelReyCG: I changed these members to `private`.

src/coreComponents/fileIO/vtk/VTKPVDWriter.cpp
```diff
@@ -42,20 +42,20 @@ void VTKPVDWriter::setFileName( string fileName )
...
 void VTKPVDWriter::save() const
 {
-  m_pvdFile.save_file( m_fileName.c_str() );
+  m_pvdFile.saveFile( m_fileName.c_str() );
 }
 
 void VTKPVDWriter::addData( real64 time, string const & filePath ) const
 {
-  auto collectionNode = m_pvdFile.child( "VTKFile" ).child( "Collection" );
+  auto collectionNode = m_pvdFile.getChild( "VTKFile" ).child( "Collection" );
```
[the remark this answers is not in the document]
MelReyCG: I can wrap `xml_node` and `xml_attribute` the same way I wrapped `xml_document`. It also implies to change all `attribute()` to `getAttribute()`, all `name` and `value` to `getName` and `getValue`.
It involves quite a few changes, but nothing impossible. It would uniformize this part of the code and it would be a "true" wrapper (more than a list of `using`).

I think I should use inlines here, or do you think I should continue with separated method in the cpp?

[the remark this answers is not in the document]
TotoGaz: Let's to another PR for this, I think it's best.
w.r.t. the inline policy, if it's onliners, surely `inline` is nice, but I'm fine with your decisions.

src/coreComponents/dataRepository/GroupContext.cpp
```diff
@@ -58,7 +66,15 @@ WrapperContext::WrapperContext( WrapperBase & wrapper ):
...
 {
-  return m_group.getDataContext().toString( m_typeName );
+  ToStringInfo const info = m_group.getDataContext().getToStringInfo();
+  if( info.hasInputFileInfo())
+  {
+    return GEOS_FMT( "{}, attribute {}", m_group.getDataContext().toString(), m_typeName );
+  }
+  else
+  {
+    return GEOS_FMT( "{}->{}", m_group.getDataContext().toString(), m_typeName );
```
TotoGaz, line 76:
```suggestion
  string const tmp = info.hasInputFileInfo() ? "{}, attribute {}" : "{}->{}";
  return GEOS_FMT( tmp, m_group.getDataContext().toString(), m_typeName );
```
?

src/coreComponents/dataRepository/GroupContext.cpp
```diff
@@ -34,22 +34,30 @@ GroupContext::GroupContext( Group & group ):
 
 string GroupContext::toString() const
 {
-  string path;
-  bool foundNearestLine = false;
-  for( Group const * parentGroup = &m_group; parentGroup->hasParent(); parentGroup = &parentGroup->getParent() )
+  std::vector< string > parents;
+  Group const * parentGroup = &m_group;
+  // add all parent names in a path string, until we get some input file info to show
+  bool foundNearestLineInfo = false;
```
TotoGaz, line 40: So do you still need this `bool` here? It seems like it's never modified?
Surely you need to be doing something with it... Is this code doing what you want?

Personnaly, I would have added the extra gathering of `std::vector< ToStringInfo >` before converting to `strings`.

TotoGaz (commented):
It's complicated for me to see the impact of your last commits since you do not appear to be in sync with `develop` or something alike. Can you manage this and ping me back then?

src/coreComponents/dataRepository/GroupContext.cpp
MelReyCG: I changed the output of this method to uniformize it with the `DataFileContext::toString()` output, but I used a ternary as you suggested to shorten the code.

src/coreComponents/dataRepository/GroupContext.cpp
MelReyCG: I also pushed another version of `GroupContext::toString()` that starts by loading all `ToStringInfo`.
It may be better in term of readability.

src/coreComponents/dataRepository/xmlWrapper.hpp
```diff
@@ -46,21 +46,225 @@ namespace geos
...
+/// will modify the original xmlNode object.
 using xmlNode = pugi::xml_node;
 
 /// Alias for the type of an xml attribute.
+/// An xmlAttribute behave as a pointer object: passing it by value to a function which modify it
+/// will modify the original xmlAttribute object.
 using xmlAttribute = pugi::xml_attribute;
 
 /// Alias for the type variant of an xml node.
 using xmlTypes = pugi::xml_node_type;
```
TotoGaz, line 64: Depending on where you want to push your wrapping, I would expect all of these types to be wrapped as well?

src/coreComponents/dataRepository/DataContext.hpp
```diff
@@ -0,0 +1,222 @@
...
+  {
+    return format_to( ctx.out(), dataContext.toString() );
+  }
+
+  /**
+   * @brief Method to parse a dataContext from a string. Not implemented!
+   * @param ctx formatting state consisting of the formatting arguments and the output iterator
+   * @return iterator to the output buffer (leaved unchanged)
+   */
+  constexpr auto parse( format_parse_context & ctx )
```
TotoGaz, line 218: If this is not implemented, can you not define it?
At least, would you kill the simulation in case it gets called?

src/coreComponents/dataRepository/GroupContext.cpp
```diff
@@ -0,0 +1,74 @@
...
+  {
+    parentsInfo.push_back( group->getDataContext().getToStringInfo() );
+  }
+
+  std::ostringstream path;
+  auto lastFileInfo = &*std::find_if( parentsInfo.begin(), parentsInfo.end(),
+                                      []( ToStringInfo const & i ) { return i.hasInputFileInfo(); } );
+  for( auto info = parentsInfo.rbegin(); info != parentsInfo.rend(); ++info )
+  {
+    path << ( &*info != lastFileInfo ?
```
TotoGaz, line 48: Can't you simply compare `lastFileInfo` and `info` without the `&*` here and there?
Or at least
```suggestion
  auto lastFileInfo = std::find_if( parentsInfo.begin(), parentsInfo.end(),
                                    []( ToStringInfo const & i ) { return i.hasInputFileInfo(); } );
  for( auto info = parentsInfo.rbegin(); info != parentsInfo.rend(); ++info )
  {
    path << ( &*info != &*lastFileInfo ?
```
?

src/coreComponents/dataRepository/xmlWrapper.cpp
```diff
@@ -169,6 +194,265 @@ string buildMultipleInputXML( string_array const & inputFileList,
...
+{
+  if( !attName.empty())
+  {
+    size_t searchStart = tagBegin;
+    try
+    {
+      std::smatch m;
+      // we search for a string which is the attribute name followed by an '=', eventually separated by spaces
+      if( std::regex_search( xmlBuffer.cbegin() + searchStart, xmlBuffer.cbegin() + tagEnd,
+                             m, std::regex( attName + "\\s*=\\s*\"" )))
```
TotoGaz, line 376: That does not feel good to regex some `xml` string.
What are you trying the do in the end?

Could a `xpath` query solve your problem?

src/coreComponents/dataRepository/GroupContext.hpp
```diff
@@ -0,0 +1,100 @@
...
+  string toString() const override;
+  /**
+   * @copydoc DataContext::getToStringInfo()
+   */
+  ToStringInfo getToStringInfo() const override;
+};
+
+/// Dedicated implementation of GroupContext for Wrapper.
+/// See DataContext class for more info.
+class WrapperContext final : public GroupContext
```
TotoGaz, line 75: I agree it's a bit questionnable, but could you move `WrapperContext` to its own file?

src/coreComponents/dataRepository/WrapperBase.cpp
```diff
@@ -104,16 +106,42 @@ int WrapperBase::setTotalviewDisplay() const
...
+  xmlWrapper::xmlAttributePos const attPos = nodePos.getAttributeLine( getName() );
+  std::ostringstream oss;
+  string const exStr = ex.what();
+
+  oss << "***** XML parsing error at node ";
+  if( nodePos.isFound() )
+  {
+    oss << "named " << m_parent->getName() << ", attribute " << getName()
+        << " (" << splitPath( attPos.isFound() ? attPos.filePath : nodePos.filePath ).second
+        << ", l." << ( attPos.isFound() ? attPos.line : nodePos.line ) << ").";
```
TotoGaz, line 135:
```suggestion
    string const filePath = attPos.isFound() ? attPos.filePath : nodePos.filePath;
    int const line = attPos.isFound() ? attPos.line : nodePos.line;
    oss << "named " << m_parent->getName() << ", attribute " << getName() << " (" << splitPath( filePath ).second << ", l." << line << ").";
```
?

src/coreComponents/dataRepository/xmlWrapper.hpp
```diff
@@ -46,21 +46,225 @@ namespace geos
...
+  /**
+   * @brief Load document from buffer. Copies/converts the buffer, so it may be deleted or changed
+   * after the function returns. Free any previously loaded xml tree.
+   * Wrapper of pugi::xml_document::loadBuffer() method.
+   * @param contents the buffer containing the document content
+   * @param size the size of the buffer in bytes
+   * @param loadNodeFileInfo Load the node source file info, allowing getNodePosition() to work.
+   * @return an xmlResult object representing the parsing resulting status.
+   */
+  xmlResult loadBuffer( const void * contents, size_t size, bool loadNodeFileInfo = false );
```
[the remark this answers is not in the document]
MelReyCG: I removed `loadBuffer()` in favor of a `loadString()` method that takes a `string_view` parameter (to be able to load `std::string` and raw string).

src/coreComponents/fileIO/vtk/VTKPVDWriter.cpp
```diff
@@ -24,11 +24,11 @@ VTKPVDWriter::VTKPVDWriter( string fileName ):
   m_fileName( std::move( fileName ) )
 {
   // Declaration of XML version
-  auto declarationNode = m_pvdFile.append_child( pugi::node_declaration );
+  auto declarationNode = m_pvdFile.appendChild( pugi::node_declaration );
```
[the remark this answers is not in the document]
MelReyCG: No, I planned this modification in a later PR (as it requires to properly wrap the `xmlNode` type). I would change xmlTypes to this declaration:
```
enum class xmlNodeType
{
  Null = pugi::node_null,               // Empty (null) node handle
  Document = pugi::node_document,       // A document tree's absolute root
  Element = pugi::node_element,         // Element tag, i.e. '<node/>'
  Pcdata = pugi::node_pcdata,           // Plain character data, i.e. 'text'
  Cdata = pugi::node_cdata,             // Character data, i.e. '<![CDATA[text]]>'
  Comment = pugi::node_comment,         // Comment tag, i.e. '<!-- text -->'
  PI = pugi::node_pi,                   // Processing instruction, i.e. '<?name?>'
  Declaration = pugi::node_declaration, // Document declaration, i.e. '<?xml version="1.0"?>'
  Doctype = pugi::node_doctype          // Document type declaration, i.e. '<!DOCTYPE doc>'
};
```

src/coreComponents/dataRepository/xmlWrapper.hpp
MelReyCG: Indeed, I planned to wrap these types in a later PR.

src/coreComponents/dataRepository/Group.hpp
```diff
@@ -1300,14 +1303,43 @@ class Group
...
+  /**
+   * @brief Calls a lambda for the DataContext of this Group and the ones of its children Group and
+   * Wrapper if they can be casted to the requested class type.
+   * @tparam TARGET_DC the requested DataContext class or parent class type.
+   * @tparam LAMBDA the type of functor to call. The functor takes in parameter a DataContext of
+   * TARGET_DC type.
+   * @param func
+   */
+  template< typename TARGET_DC = DataContext, typename FUNC >
+  void forAllDataContext( FUNC func ) const;
```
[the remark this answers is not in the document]
MelReyCG: In `testXMLFile.cpp:184`:
```
  problemManager.forAllDataContext< DataFileContext >( [&]( DataFileContext const & ctx )
  {
    verifyDataFileContext( ctx, xmlDoc, verifiedElements );
  } );
```

src/coreComponents/dataRepository/GroupContext.cpp
MelReyCG: Does the `info.base() != lastFileInfo` syntax would be better ?

src/coreComponents/dataRepository/xmlWrapper.cpp
MelReyCG: I want to determine the attribute line here.

As mentioned in [previous messages](https://github.com/GEOS-DEV/GEOS/pull/2404/#discussion_r1253223108), this portion of code isn't perfect, but takes into account the typical GEOS XMLs. And if the attribute position cannot be determined, the parent node line will be displayed (or in less likely cases a near position of the attribute).

`pugixml` xpaths can only get node and attribute objects. And unfortunately, `pugi::xpath_query` only expose a character offset when an erroneous xpath is evaluated (and the `PUGIXML_NO_EXCEPTIONS` must be enabled).
The code proposed here already get a reliable node character offset from `pugixml`, but `pugi::attribute` don't expose this offset value, that's why I rely on a regex to get the attribute line.

src/coreComponents/dataRepository/GroupContext.cpp
MelReyCG: My bad, this does not give the right result, we would need `( info.base() - 1 ) != lastFileInfo` or `std::prev( info.base() ) != lastFileInfo` here.
To avoid this confusing pointer arithmetic, we should stick to `&*info != &*lastFileInfo`.
Another solutions would be to reverse `parentsInfo` rather than using reverse iterators, or to use `rbegin()` and `rend()` as `std::find_if()` parameters.

src/coreComponents/dataRepository/xmlWrapper.cpp
TotoGaz: OK, I get it better now. I think that `xpath v2` supports `regex`, but `pugixml` only support `v1` 🤷 !
Can you put a little bit more comments around this line? Since it's unusual, I think it's worth documenting the process.

MelReyCG: OK, I'll add some comments here.

MelReyCG:
Can you have a look on this one? https://github.com/GEOS-DEV/GEOS/pull/2404/#discussion_r1303149788 (I can keep my initial proposal but I don't know if you are ok with it).
Apart from this one, yes I think this PR should be merged

src/coreComponents/dataRepository/GroupContext.cpp
TotoGaz: I'm fine with your code if it's working. This `&*` is a bit unusual, but I guess it is what it is.

src/coreComponents/dataRepository/DataContext.cpp
```diff
@@ -0,0 +1,109 @@
...
+DataContext::ToStringInfo::ToStringInfo( string const & targetName, string const & filePath, size_t line ):
+  m_targetName( targetName ),
+  m_filePath( filePath ),
+  m_line( line )
+{}
+DataContext::ToStringInfo::ToStringInfo( string const & targetName ):
+  DataContext::ToStringInfo::ToStringInfo( targetName, "", xmlWrapper::xmlDocument::npos )
+{}
+bool DataContext::ToStringInfo::hasInputFileInfo() const
+{ return !m_filePath.empty() && m_line != xmlWrapper::xmlDocument::npos; }
```
klevzoff, line 46: Could you consider
* defining `hasInputFileInfo` inline in the declaration of `ToStringInfo` in the header;
* removing the constructors and adding a default value in the struct declaration:
```c++
size_t m_line = xmlWrapper::xmlDocument::npos;
```

src/coreComponents/dataRepository/GroupContext.cpp
klevzoff: Dereferencing `lastinfo` is undefined behavior if the `find_if` call doesn't find anything. I wouldn't mind `std::prev( info.base() ) != lastFileInfo`, maybe you can add a short comment on the same line to help the reader, something like `// pointing to the same element`.

src/coreComponents/dataRepository/DataContext.hpp
```diff
@@ -0,0 +1,233 @@
...
+  /**
+   * @brief Method to parse a dataContext from a string. Not implemented!
+   * @param ctx formatting state consisting of the formatting arguments and the output iterator
+   * @return iterator to the output buffer (leaved unchanged)
+   */
+  auto parse( format_parse_context & ctx )
+  {
+    GEOS_ERROR( "DataContext parsing is not implemented." );
+    return ctx.begin();
+  }
```
klevzoff, line 230: `formatter.parse()` is not meant for parsing the target type itself, but for parsing the format string for the type and determining the format options, such as `{:<.2f}`. So anytime you do `GEOS_FMT("{}", dataContext)`, `parse` will be called and throw an error. 

To do a formatter correctly, you have 2 options:
* If you want to strictly only allow empty format strings `{}`, just returning `ctx.begin()` is sufficient - the library will throw an exception if the next character is not a `}`.
* Alternatively, you could inherit from `GEOS_FMT_NS::formatter<std::string>` and automatically reuse its `parse` method. This will give you free support for width/alignment specifiers. Then in format you can just
```c++
    return GEOS_FMT_NS::formatter<std::string>::format( dataContext.toString(), ctx );
```

src/coreComponents/dataRepository/GroupContext.cpp
TotoGaz: Oh, I should have seen that, thank you for checking!

src/coreComponents/dataRepository/Group.hpp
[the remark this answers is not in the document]
TotoGaz: If it's only used in a test, can you find another way to do this?
I'm not too comfortable adding an extra member function to `Group` (and therefore to all our classes or close to be) for the sake of a unit test.

src/coreComponents/dataRepository/DataContext.hpp
MelReyCG: Ok! I will go with this second option.

src/coreComponents/dataRepository/Group.hpp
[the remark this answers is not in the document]
MelReyCG: I removed it in profit of a free function in testXMLFile.cpp

src/coreComponents/dataRepository/GroupContext.hpp
```diff
@@ -0,0 +1,80 @@
...
+  /**
+   * @brief Construct a new GroupContext object
+   * @param group The reference to the Group related to this GroupContext.
+   */
+  GroupContext( Group & group );
+
+  /**
+   * @return the reference to the Group related to this GroupContext.
+   */
+  Group & getGroup() const;
```
TotoGaz, line 50: Could you return a reference to const instead?

TotoGaz (approved):
I think it should be ready to get in.
@klevzoff if you want to check that the fixes you pointed are really fixed, you're welcome!

TotoGaz:
@MelReyCG I've put the PR in the [merge queue](https://github.com/orgs/GEOS-DEV/projects/6/views/1).

PR #2412, opened 2023-04-28
Update ThermoPoroElastic Wellbore Problem

jhuang2601:
This PR updates the tutorial example for [THM wellbore problem](https://geosx-geosx.readthedocs-hosted.com/en/latest/docs/sphinx/advancedExamples/validationStudies/wellboreProblems/thermoPoroElasticWellbore/Example.html) and a better match with the analytical solutions is obtained.

Rebaseline done in https://github.com/GEOS-DEV/integratedTests/pull/327


herve-gross (approved):
Impressive match, that's great.

francoishamon (approved):
Looks great, thanks for updating the test case! 
One thing to check is whether the unit in the temperature legend should become Kelvin instead of degrees Celsius

jhuang2601:
> Looks great, thanks for updating the test case! One thing to check is whether the unit in the temperature legend should become Kelvin instead of degrees Celsius

This example is initialized with zeros (effective stress, pressure and temperature), and their final values are only treated as the perturbation induced by the thermal loading (temperature change) at the wellbore. Therefore, temperature legend can be kept as Celsius for a reasonable physical meaning.

src/docs/sphinx/advancedExamples/validationStudies/wellboreProblems/thermoPoroElasticWellbore/THM_wellbore_displacement.py
```diff
@@ -16,8 +16,8 @@ def main():
 	nNodes = center.shape[1]
 	xCoord = center[0, 0:nNodes, 0]
 	yCoord = center[0, 0:nNodes, 1]	
-	ux_60s = displacement[9, 0:nNodes, 0] # the index for time step at 60s is 9
-	ux_3600s = displacement[23, 0:nNodes, 0] # the index for time step at 3600s is 23
+	ux_60s = displacement[10, 0:nNodes, 0] # the index for time step at 60s is 10
+	ux_3600s = displacement[24, 0:nNodes, 0] # the index for time step at 3600s is 24
```
sytuannguyen, line 20: Well catched! thank you @jhuang2601

sytuannguyen (approved):

Issue #2418, opened 2023-05-02
LC test failures


rrsettgast:
We are failing tests on the various LC platforms. Summary:

on quartz-clang14:
unit tests failures:
```
	167 - testVTKImport (Failed)
	169 - testVTKImport_mpi (Failed)
	185 - testDoxygenCheck (Failed)
```
integrated test failures:
```
    Status     :  TestCase                                           :  Elapsed :  Resources :  TestStep     :  OutFile                                                                                                                                                                                                                               
     ---------- :  -------------------------------------------------- :  ------- :  --------- :  ------------ :  --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
     FAIL RUN   :  acous3D_abc_smoke_01                               :  0:00:01 :  0:00:01   :  geosx        :  wavePropagation/acous3D_abc_smoke_01/acous3D_abc_smoke_01.data wavePropagation/acous3D_abc_smoke_01/acous3D_abc_smoke_01.err                                                                                                          
     FAIL RUN   :  acous3D_abc_smoke_08                               :  0:00:02 :  0:00:17   :  geosx        :  wavePropagation/acous3D_abc_smoke_08/acous3D_abc_smoke_08.data wavePropagation/acous3D_abc_smoke_08/acous3D_abc_smoke_08.err                                                                                                          
     FAIL RUN   :  acous3D_abc_fs_smoke_01                            :  0:00:01 :  0:00:01   :  geosx        :  wavePropagation/acous3D_abc_fs_smoke_01/acous3D_abc_fs_smoke_01.data wavePropagation/acous3D_abc_fs_smoke_01/acous3D_abc_fs_smoke_01.err                                                                                              
     FAIL RUN   :  acous3D_abc_fs_smoke_08                            :  0:00:01 :  0:00:12   :  geosx        :  wavePropagation/acous3D_abc_fs_smoke_08/acous3D_abc_fs_smoke_08.data wavePropagation/acous3D_abc_fs_smoke_08/acous3D_abc_fs_smoke_08.err                                                                                              
     FAIL RUN   :  beamBending_hybridHexPrism_08                      :  0:00:03 :  0:00:26   :  geosx        :  solidMechanicsSSLE/beamBending_hybridHexPrism_08/beamBending_hybridHexPrism_08.data solidMechanicsSSLE/beamBending_hybridHexPrism_08/beamBending_hybridHexPrism_08.err                                                                
[...]

PR #2422, opened 2023-05-04
Fixed wellbore bug related to useCartesianOuterBoundary

francoishamon:
This PR fixes one of the integrated test crashes reported by @rrsettgast in #2418 

It was related to the wrong use of the keyword `useCartesianOuterBoundary`. I just added a check to make sure that the input is correct. That being said, I really don't know how this case was running before (there is was a clear division by 0).

Requires rebaseline


jhuang2601:
I just rerun the existing tutorial example with the keyword `useCartesianOuterBoundary`. 
It still works with most recent develop repo and provides accurate results.
Weird bug for the failure of smoke test.

francoishamon:
> I just rerun the existing tutorial example with the keyword `useCartesianOuterBoundary`. It still works with most recent develop repo and provides accurate results. Weird bug for the failure of smoke test.

Thanks very much @jhuang2601

sytuannguyen (approved):
It looks good to me. Thank you @francoishamon

PR #2426, opened 2023-05-09
Output run start/end/ranks

klevzoff:
A tiny addition to standard GEOS output to make postprocessing of simulation logs easier.

### DIFF
--- a/src/cmake/thirdparty/SetupGeosxThirdParty.cmake
+++ b/src/cmake/thirdparty/SetupGeosxThirdParty.cmake
@@ -710,6 +710,12 @@ if(DEFINED FMT_DIR)
 
     message( " ----> fmt_VERSION = ${fmt_VERSION}")
 
+    get_target_property(includeDirs fmt::fmt INTERFACE_INCLUDE_DIRECTORIES)
+
+    set_property(TARGET fmt::fmt
+                 APPEND PROPERTY INTERFACE_SYSTEM_INCLUDE_DIRECTORIES
+                 ${includeDirs})
+
     set(ENABLE_FMT ON CACHE BOOL "")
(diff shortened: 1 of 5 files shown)


TotoGaz (approved):

klevzoff:
The two CI jobs with GCC 10 that are failing are being hit by this (or similar) compiler bug that is fixed for GCC 11: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95353

`{fmt}` version 9.1 contains some [workarounds](https://github.com/fmtlib/fmt/pull/3054) for similar issues, any objections to trying to update to that and seeing if it helps?

If that doesn't work, can I add `-Wno-stringop-overflow` to CI compiler flags for those jobs? I know this is heavy artillery, but we are still safeguarded by having this warning enabled in other builds.

@TotoGaz @rrsettgast

klevzoff:
> The two CI jobs with GCC 10 that are failing are being hit by this (or similar) compiler bug that is fixed for GCC 11: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95353
> 
> `{fmt}` version 9.1 contains some [workarounds](https://github.com/fmtlib/fmt/pull/3054) for similar issues, any objections to trying to update to that and seeing if it helps?
> 
> If that doesn't work, can I add `-Wno-stringop-overflow` to CI compiler flags for those jobs? I know this is heavy artillery, but we are still safeguarded by having this warning enabled in other builds.
> 
> @TotoGaz @rrsettgast

This was not needed in the end, a simpler workaround was to add `fmt` includes as system includes on GEOS side to ignore warnings, as we do for some other TPLs. That said, I did test with fmt-9.1 and kept the necessary changes in this branch (they work with both old and new versions). I'm not gonna go through the trouble of updating TPLs at this time, but it should be very easy for someone else to do in the future.

PR #2427, opened 2023-05-09
Linear algebra updates

klevzoff:
This is a collection of LAI-related changes extracted from my multiscale branch to be reviewed separately. I will post some comments later highlighting the most important changes.

Rebaseline not needed.


src/coreComponents/linearAlgebra/unitTests/testKrylovSolvers.cpp
```diff
@@ -243,10 +259,17 @@ TYPED_TEST_P( KrylovSolverBlockTest, GMRES )
   this->test( params_GMRES() );
 }
 
+TYPED_TEST_P( KrylovSolverBlockTest, Richardson )
+{
+  this->test( params_GMRES() );
```
klevzoff, line 264:
```suggestion
  this->test( params_Richardson() );
```

src/coreComponents/linearAlgebra/CMakeLists.txt
```diff
@@ -105,19 +108,19 @@ if( ENABLE_HYPRE )
           interfaces/hypre/mgrStrategies/CompositionalMultiphaseReservoirHybridFVM.hpp
           interfaces/hypre/mgrStrategies/HybridSinglePhasePoromechanics.hpp
           interfaces/hypre/mgrStrategies/Hydrofracture.hpp
-          interfaces/hypre/mgrStrategies/LagrangianContactMechanics.hpp      
+          interfaces/hypre/mgrStrategies/LagrangianContactMechanics.hpp
           interfaces/hypre/mgrStrategies/MultiphasePoromechanics.hpp
           interfaces/hypre/mgrStrategies/MultiphasePoromechanicsReservoirFVM.hpp
-	  interfaces/hypre/mgrStrategies/ReactiveCompositionalMultiphaseOBL.hpp
+	        interfaces/hypre/mgrStrategies/ReactiveCompositionalMultiphaseOBL.hpp
```
klevzoff, line 114:
```suggestion
        interfaces/hypre/mgrStrategies/ReactiveCompositionalMultiphaseOBL.hpp
```

src/coreComponents/linearAlgebra/CMakeLists.txt
```diff
@@ -105,19 +108,19 @@ if( ENABLE_HYPRE )
...
           interfaces/hypre/mgrStrategies/SinglePhasePoromechanics.hpp
           interfaces/hypre/mgrStrategies/SinglePhasePoromechanicsEmbeddedFractures.hpp
-          interfaces/hypre/mgrStrategies/SinglePhasePoromechanicsReservoirFVM.hpp	  
+          interfaces/hypre/mgrStrategies/SinglePhasePoromechanicsReservoirFVM.hpp
           interfaces/hypre/mgrStrategies/SinglePhaseReservoirFVM.hpp
           interfaces/hypre/mgrStrategies/SinglePhaseReservoirHybridFVM.hpp
-	  interfaces/hypre/mgrStrategies/ThermalCompositionalMultiphaseFVM.hpp
-          interfaces/hypre/mgrStrategies/ThermalMultiphasePoromechanics.hpp	  	  
-          interfaces/hypre/mgrStrategies/ThermalSinglePhasePoromechanics.hpp	  
+	        interfaces/hypre/mgrStrategies/ThermalCompositionalMultiphaseFVM.hpp
```
klevzoff, line 121:
```suggestion
        interfaces/hypre/mgrStrategies/ThermalCompositionalMultiphaseFVM.hpp
```

src/coreComponents/mesh/DomainPartition.hpp
```diff
@@ -277,6 +277,21 @@ class DomainPartition : public dataRepository::Group
   std::vector< NeighborCommunicator > const & getNeighbors() const
   { return m_neighbors; };
 
+  /**
+   * @brief Get a list of neighbor ranks.
+   * @return Container of neighbor ranks.
+   */
+  std::vector< int > getNeighborRanks() const
```
klevzoff, line 284: This wasn't meant to be in this PR, but let's just merge it anyway, it's used in follow-up multiscale PR

src/coreComponents/physicsSolvers/LinearSolverParameters.cpp
```diff
@@ -20,8 +20,79 @@
...
+    setInputFlag( InputFlags::OPTIONAL ).
+    setDescription( "Schur complement type, options: "
+                    "``" + EnumStrings< LinearSolverParameters::Block::SchurType >::concat( "``, ``" ) + "``" );
+
+  registerWrapper( viewKeyStruct::scalingString(), &m_parameters.scaling ).
+    setDefaultValue( m_parameters.scaling ).
+    setInputFlag( InputFlags::OPTIONAL ).
+    setDescription( "Block scaling type, options: "
+                    "``" + EnumStrings< LinearSolverParameters::Block::Scaling >::concat( "``, ``" ) + "``" );
+}
```
klevzoff, line 94: This is a new way of structuring linear solver parameter inputs: instead of an ever-growing list of prefixed attributes like 
```xml
<LinearSolverParameters
  blockShape="..."
  blockShurType="..."
  blockScaling="..."/>
```
we will start adding nested structure as follows:
```xml
<LinearSolverParameters>
  <Block
    shape="..."
    schurType="..."
    scaling="..."/>
</LinearSolverParameters>
```

This improves readability by indenting parameter blocks according to their logical structure and avoiding repetition of the prefixes. I use this extensively in multiscale work, but will also propose a similar refactoring for AMG parameters.

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.cpp
```diff
@@ -330,12 +330,6 @@ void SinglePhaseBase::initializePostInitialConditionsPreSubGroups()
                                                                MeshLevel & mesh,
                                                                arrayView1d< string const > const & regionNames )
   {
-    FieldIdentifiers fieldsToBeSync;
-    fieldsToBeSync.addElementFields( { fields::flow::pressure::key() },
-                                     regionNames );
-
-    CommunicationTools::getInstance().synchronizeFields( fieldsToBeSync, mesh, domain.getNeighbors(), false );
-
```
klevzoff, line 338: This synchronization was useless because consistency of values is already guaranteed by virtue of ghosted cells being included in sets to which initial conditions are applied.

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseFVM.cpp
```diff
@@ -103,6 +103,25 @@ void SinglePhaseFVM< BASE >::setupSystem( DomainPartition & domain,
...
+SinglePhaseFVM< BASE >::createPreconditioner( DomainPartition & domain ) const
+{
+  LinearSolverParameters const & linParams = m_linearSolverParameters.get();
+  GEOS_UNUSED_VAR( domain );
+  switch( linParams.preconditionerType )
+  {
+    default:
+    {
+      return LAInterface::createPreconditioner( linParams );
+    }
```
klevzoff, line 123: The `switch` statement may seem pointless at the moment, but it's left as a placeholder for adding custom preconditioners (e.g. multiscale)

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.hpp
```diff
@@ -291,8 +276,8 @@ class SolidMechanicsLagrangianFEM : public SolverBase
   MPI_iCommData m_iComm;
   integer m_fixedStressUpdateThermoPoromechanicsFlag;
 
-  /// Rigid body modes
-  array1d< ParallelVector > m_rigidBodyModes;
+  /// Rigid body modes; TODO remove mutable hack
+  mutable array1d< ParallelVector > m_rigidBodyModes;
```
klevzoff, line 280: This is now `mutable` because it's populated on demand in `createPreconditioner()` function which is `const`. This is isn't great, but I don't have a better solution for now.

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.cpp
```diff
@@ -1179,6 +1205,17 @@ SolidMechanicsLagrangianFEM::applySystemSolution( DofManager const & dofManager,
...
 }
 
+void SolidMechanicsLagrangianFEM::solveLinearSystem( DofManager const & dofManager,
+                                                     ParallelMatrix & matrix,
+                                                     ParallelVector & rhs,
+                                                     ParallelVector & solution )
+{
+  // Flip system sign to ensure matrix is positive definite
+  matrix.scale( -1.0 );
+  rhs.scale( -1.0 );
```
klevzoff, line 1215: I needed this at some point for multiscale to work (I'll have to double check to see if still required), but probably not a good idea to apply in the general case.
```suggestion
  // Flip system sign to ensure matrix is positive definite
  matrix.scale( -1.0 );
  rhs.scale( -1.0 );
```

src/coreComponents/physicsSolvers/LinearSolverParameters.cpp
victorapm: This is very nice, specially if we want to expose some MGR options per level...

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.cpp
paveltomin: how about other flow solvers?

src/coreComponents/linearAlgebra/DofManager.cpp
```diff
@@ -905,11 +910,11 @@ void DofManager::setSparsityPatternFromStencil( SparsityPatternView< globalIndex
...
       globalIndex const elemDof = dofNumberView[elemIdx[0]][elemIdx[1]][elemIdx[2]];
-      for( localIndex c = 0; c < numComp; ++c )
+      for( integer c = 0; c < numComp; ++c )
       {
         colDofIndices[c] = elemDof + c;
       }
-      for( localIndex c = 0; c < numComp; ++c )
+      for( integer c = 0; c < numComp; ++c )
       {
         pattern.insertNonZeros( elemDof - rankDofOffset + c, colDofIndices.begin(), colDofIndices.end() );
```
victorapm, line 919: `elemDof` and `rankDofOffset` are of type `globalIndex`, while `c` is `integer`. We should fix the type mismatch

src/coreComponents/linearAlgebra/interfaces/hypre/HypreKernels.hpp
```diff
@@ -210,72 +210,150 @@ makeSortedPermutation( HYPRE_Int const * const indices,
...
 } // namespace internal
 
-template< typename SRC_COLMAP, typename DST_COLMAP >
-void addEntriesRestricted( hypre_CSRMatrix const * const src_mat,
-                           SRC_COLMAP const src_colmap,
-                           hypre_CSRMatrix * const dst_mat,
-                           DST_COLMAP const dst_colmap,
-                           real64 const scale )
+template< typename KERNEL >
+void addMatrixEntries( hypre_ParCSRMatrix const * const src,
```
victorapm, line 214: @castelletto1 would it be a good idea to have this implemented in hypre instead?

src/coreComponents/linearAlgebra/interfaces/hypre/HyprePreconditioner.cpp
```diff
@@ -320,7 +344,7 @@ void HyprePreconditioner::create( DofManager const * const dofManager )
       hypre::mgr::createMGR( m_params, dofManager, *m_precond, *m_mgrData );
       break;
     }
-    case LinearSolverParameters::PreconditionerType::iluk:
+    case LinearSolverParameters::PreconditionerType::ilu:
```
victorapm: Should we have a single `ilu` that represents both `iluk` and `ilut`? If not, should we keep the old name here (`iluk`)?

src/coreComponents/linearAlgebra/interfaces/hypre/HypreUtils.cpp
```diff
@@ -46,6 +48,68 @@ HYPRE_Vector parVectorToVectorAll( HYPRE_ParVector const vec )
   }
 }
 
+HYPRE_Vector parVectorToVector( HYPRE_ParVector const vec, int const targetRank )
```
victorapm, line 51: Maybe we could use https://github.com/hypre-space/hypre/blob/907a2d07b64fe47bdde4540c54665c83ced83a2c/src/parcsr_mv/par_vector.c#L985 instead

Do we need the situation when `targetRank != 0`?

src/coreComponents/linearAlgebra/interfaces/hypre/HypreUtils.cpp
```diff
@@ -173,6 +238,111 @@ HYPRE_Int relaxationDestroy( HYPRE_Solver solver )
   return 0;
 }
 
+/**
+ * @brief Holds data needed by hypre's Chebyshev smoothing.
+ *
+ * See RelaxationData for explanation of the way this structure is used.
+ */
+struct ChebyshevData
```
victorapm, line 246: would it be better to expose the Chebyshev solver in hypre?

src/coreComponents/linearAlgebra/DofManager.cpp
klevzoff: My logic behind these type choices:
  * `globalIndex` - anything that enumerates things (elems, nodes, etc.) in the global mesh
  * `localIndex` - same but in a local mesh partition
  * `integer` - all other integers that don't "scale" with the size of the mesh. In particular, while `localIndex` can be 64-bit in some builds, `integer` is always 32-bit, so it's more suitable to represent small quantities.

By this logic, "small" numbers like number of DoF components in a field or component index should be of type `integer`. They can be freely added to `globalIndex` variables to form new global DoF indices - there is no need to change types here IMO.

src/coreComponents/linearAlgebra/interfaces/hypre/HyprePreconditioner.cpp
klevzoff: So the reason for this change is: I made my input files with the following structure:
```xml
<LinearSolverParameters
  preconditionerType="$prec$"
  amgSmootherType="$smoother$">
  <Multiscale>
    <Smoother
      type="$smoother$"/>
  </Multiscale>
</LinearSolverParameters>
```
which I then invoke with preprocessor:
```
geosx_preprocessed -i <filename> -p prec [amg|multiscale] -p smoother ilu
```
The first value of `$smoother$` maps to `AMG::SmootherType` but the second one is of type `PreconditionerType` (because I use a generic preconditioner interface to instantiate a smoother for multiscale V-cycle). So I needed a unified XML string for `PreconditionerType` (which has `iluk`) and `AMG::SmootherType` (which has `ilu0`) for my input files, so I converged on just `ilu` for both. 

If it's ok to use `iluk` for both, I would be happy to. Either way it's a breaking change for some existing input decks out there. The only way to avoid it would be to do a purely additive change, i.e. support both `iluk|ilu` for one and `ilu0|ilu` for the other.

src/coreComponents/linearAlgebra/interfaces/hypre/HypreUtils.cpp
klevzoff: That would be ideal, and by same reasoning directly exposing other relaxation methods would be useful. But I think we got pretty comfortable maintaining these wrappers on GEOS end, so it's not necessary or urgent.

src/coreComponents/linearAlgebra/interfaces/hypre/HypreUtils.cpp
klevzoff: It's needed to support serial direct solver (`umfpack`) in cases when a target physics may not be present on all MPI ranks (like poromechanics with over/underburden). Because we use `hypre_ParCSRMatrixToCSRMatrixAll` to collect the matrix to a single working rank, and it cannot collect it to a rank where it has 0 rows, we may have to select a working rank other than 0. Now, `umfpack` is only useful for small-scale debugging runs, but I think I also tried it as a coarse-level solver in my multiscale implementation. So ideally I would like not to drop support for `targetRank != 0`.

src/coreComponents/linearAlgebra/DofManager.cpp
victorapm: Makes sense, thanks!

src/coreComponents/linearAlgebra/interfaces/hypre/HyprePreconditioner.cpp
victorapm: Sounds good, I think it's OK to use iluk for ilu0 because ilu0 is a special case here (k = 0)

src/coreComponents/linearAlgebra/interfaces/hypre/HypreUtils.cpp
victorapm: Sounds good, I'm happy to add this in hypre. I believe that's more appropriate because we will be adding new options for Chebyshev soon. If we have the interface implemented there, GEOS doesn't need to bother keeping track with our changes

src/coreComponents/linearAlgebra/interfaces/hypre/HypreUtils.cpp
victorapm: That makes sense! I'd suggest here again to move this functionality to hypre, i.e., implement what is missing in hypre. But, I'll let you guys decide @castelletto1

src/coreComponents/linearAlgebra/interfaces/hypre/HypreKernels.hpp
castelletto1: Yes, ideally all these adding options should be provided by hypre:

https://github.com/GEOS-DEV/GEOS/blob/e3e8d931d3310aec0e77933cdcfac7dfb0173e57/src/coreComponents/linearAlgebra/interfaces/hypre/HypreMatrix.cpp#L897-L927

src/coreComponents/linearAlgebra/interfaces/hypre/HyprePreconditioner.cpp
castelletto1: Yes, let's use `iluk`, and explicitly state that it defaults to `ilu0` when no `k` if provided.

castelletto1: I'm not even sure we can use any type of `ilu` as smoother in AMG for hypre at the moment:

https://github.com/GEOS-DEV/GEOS/blob/e3e8d931d3310aec0e77933cdcfac7dfb0173e57/src/coreComponents/linearAlgebra/interfaces/hypre/HypreUtils.hpp#L361-L375

src/coreComponents/linearAlgebra/interfaces/hypre/HypreUtils.cpp
castelletto1: Having linear algebra functionality in hypre makes sense to me.

src/coreComponents/linearAlgebra/solvers/PreconditionerNull.hpp
```diff
@@ -0,0 +1,62 @@
...
+#ifndef GEOS_LINEARALGEBRA_SOLVERS_PRECONDITIONERNULL_HPP_
+#define GEOS_LINEARALGEBRA_SOLVERS_PRECONDITIONERNULL_HPP_
+
+#include "linearAlgebra/common/PreconditionerBase.hpp"
+
+namespace geos
+{
+
+/**
+ * @brief Common interface for identity preconditioning operator
```
castelletto1: Update description

src/coreComponents/linearAlgebra/solvers/RichardsonSolver.hpp
```diff
@@ -0,0 +1,102 @@
...
+
+namespace geos
+{
+
+/**
+ * @brief Implements right-preconditioned modified Richardson iteration.
+ * @tparam VECTOR type of vectors this solver operates on
+ * @note Richardson is not a Krylov subspace method, but
+ *       for convenience inherits from KrylovSolver; that
+ *       class should really be renamed to IterativeSolver.
```
castelletto1, line 32: Yes, in another PR though. The unit test `testKrylovSolvers` should be renamed accordingly. Unless we want to clear separate Krylov subspace methods from relaxation schemes.

src/coreComponents/linearAlgebra/solvers/RichardsonSolver.hpp
```diff
@@ -0,0 +1,102 @@
...
+#ifndef GEOS_LINEARALGEBRA_SOLVERS_RICHARDSONSOLVER_HPP_
+#define GEOS_LINEARALGEBRA_SOLVERS_RICHARDSONSOLVER_HPP_
+
+#include "linearAlgebra/solvers/KrylovSolver.hpp"
+
+namespace geos
+{
+
+/**
+ * @brief Implements right-preconditioned modified Richardson iteration.
```
castelletto1:
```suggestion
 * @brief Implements preconditioned Richardson iteration.
```
The preconditioner is applied to the residual, so it is left preconditioning. Commonly this scheme is called simply preconditioned Richardson iteration.

src/coreComponents/linearAlgebra/utilities/LinearSolverParameters.hpp
```diff
@@ -167,9 +184,9 @@ struct LinearSolverParameters
       sgs,                ///< Symmetric Gauss-Seidel smoothing
       l1sgs,              ///< l1-Symmetric Gauss-Seidel smoothing
       chebyshev,          ///< Chebyshev polynomial smoothing
-      ilu0,               ///< ILU(0)
+      ilu,                ///< Incomplete LU
       ilut,               ///< Incomplete LU with thresholding
-      ic0,                ///< Incomplete Cholesky
+      ic,                 ///< Incomplete Cholesky
```
castelletto1: If we decide for `iluk`, we should follow the same logic for level-based Incomplete Cholesky variants.

src/coreComponents/linearAlgebra/utilities/LinearSolverParameters.hpp
```diff
@@ -314,14 +333,66 @@ struct LinearSolverParameters
...
   /// Domain decomposition parameters
   struct DD
   {
     integer overlap = 0;   ///< Ghost overlap
   }
   dd;                      ///< Domain decomposition parameter struct
+
+  /// Block preconditioner parameters
+  struct Block
+  {
```
castelletto1, line 347: Should we be verbose and explicitly clarify that all the block preconditions are for 2x2 block systems? In case of additional blocks, are we always going to rely on the LAI or will we eventually extend the current 2x2 block strategy in GEOS?

src/coreComponents/linearAlgebra/utilities/LinearSolverParameters.hpp
```diff
@@ -314,14 +333,66 @@ struct LinearSolverParameters
...
+      FirstBlockUserDefined  ///< User defined preconditioner for the first block
+    };
+
+    /// Type of block row scaling to apply
+    enum class Scaling
+    {
+      None,          ///< No scaling
+      FrobeniusNorm, ///< Equilibrate Frobenius norm of the diagonal blocks
+      UserProvided   ///< User-provided scaling
+    };
```
castelletto1, line 372: hypre can also do block scaling (PR #3682). We should align on a unique strategy at some point.

src/coreComponents/linearAlgebra/utilities/LinearSolverParameters.hpp
```diff
@@ -314,14 +333,66 @@ struct LinearSolverParameters
...
+     */
+    void resize( integer const numBlocks )
+    {
+      subParams.resize( numBlocks );
+      order.resize( numBlocks );
+      for( integer i = 0; i < numBlocks; ++i )
+      {
+        order[i] = i;
+      }
+    }
```
castelletto1, line 393: Can `numBlocks` be arbitrary? Maybe my previous comment on 2x2 block systems is incorrect.

src/coreComponents/unitTests/linearAlgebraTests/testDofManager.cpp
```diff
@@ -644,7 +642,8 @@ TYPED_TEST_P( DofManagerSparsityTest, TPFA_Full )
     { "pressure",
       FieldLocation::Elem,
       DofManager::Connector::Face,
-      2, makeSparsityTPFA }
+      2, makeSparsityTPFA,
+      { {"mesh", "Level0", {} } } }
```
castelletto1, line 646: Why wasn't the region included until now?

castelletto1 (commented):
First batch of comment, more to come

src/coreComponents/unitTests/linearAlgebraTests/testDofManager.cpp
klevzoff: I think the intent was that empty (default-initialized) list of mesh-level-region tuples would mean auto-fill, i.e. include the entire mesh. But looking at the implementation I noticed that this only works correctly for an empty region list (the innermost empty braces), but omitting mesh and level resulted in an empty sparsity pattern, so these tests were silently broken for a while - hard to tell when or how this happened, these tests are pretty old, written in the days when we used to assume a default single mesh and level.

src/coreComponents/linearAlgebra/utilities/LinearSolverParameters.hpp
klevzoff: I would be supportive of anyone working to extend the current 2x2 implementation, though I doubt it will happen, given that MGR covers (or can be extended to cover) all the important use cases. 

There are some comments and assertions in `BlockPrecondtioner` class itself that clarify/enforce the restriction, but I don't know if we need to enforce this in the parameter struct (and how to go about doing this).

src/coreComponents/linearAlgebra/utilities/LinearSolverParameters.hpp
klevzoff: Yep, and I also implemented a rudimentary version of scaling (based on Frobenius norms) in this PR, see `SinglePhasePoromechanics` changes - this is outside (and somewhat duplicating) `BlockPreconditioner` because I needed the fine-level matrix to be well scaled for coupled multiscale.

I don't know how exactly hypre's implementation computes its scaling vector, but if the end result is similar, I'm happy to abandon mine. It'd be nice to have fallback versions for other LA backends somehow, but we can worry about it later.

src/coreComponents/linearAlgebra/interfaces/hypre/HypreKernels.hpp
klevzoff: Resolving for now, will replace with hypre implementations when they are available

src/coreComponents/linearAlgebra/interfaces/hypre/HyprePreconditioner.cpp
klevzoff: Changed both to `AMG::SmootherType::iluk` and `PreconditionerType::iluk`, and similarly `ic` to `ick`. This is a breaking change for user input decks. We don't have a way to do graceful deprecation I think, but hopefully the error message that lists all valid options gives users enough of a clue as to how to update their xmls.

src/coreComponents/linearAlgebra/solvers/PreconditionerNull.hpp
klevzoff: Done

src/coreComponents/linearAlgebra/solvers/RichardsonSolver.hpp
klevzoff: Updated

src/coreComponents/linearAlgebra/utilities/LinearSolverParameters.hpp
klevzoff: Agreed, done

src/coreComponents/linearAlgebra/interfaces/hypre/HypreUtils.cpp
klevzoff: Resolving as there's nothing actionable for this PR, but let me know if I should file an issue/request for hypre.

victorapm: @castelletto1 we have a PR in hypre for exposing Chebyshev as a solver with proper `HYPRE_` APIs. It's under review... I'll let you know when it gets merged

src/coreComponents/linearAlgebra/utilities/LinearSolverParameters.hpp
victorapm: Here's a summarized explanation of how we compute the scaling vector in hypre: https://github.com/hypre-space/hypre/blob/435e042d5355a55ebb460fea6d02846fd0274d1e/src/parcsr_mv/par_csr_matop.c#L7064-L7066

victorapm (approved):
Thank you for all these nice changes!

src/coreComponents/linearAlgebra/common/PreconditionerBase.hpp
```diff
@@ -35,8 +35,6 @@ class PreconditionerBase : public LinearOperator< typename LAI::ParallelVector >
 
   PreconditionerBase() = default;
 
-  virtual ~PreconditionerBase() = default;
```
rrsettgast, line 38: doesn't this still need to be virtual?

src/coreComponents/linearAlgebra/interfaces/direct/SuperLUDist.cpp
```diff
@@ -22,6 +22,7 @@
 #include "codingUtilities/Utilities.hpp"
 #include "common/MpiWrapper.hpp"
 #include "common/Stopwatch.hpp"
+#include "common/TimingMacros.hpp"
```
rrsettgast, line 25: needed?

src/coreComponents/linearAlgebra/utilities/LinearSolverParameters.hpp
klevzoff: I removed anything related to system scaling from `SinglePhasePoromechanics` in favor of Victor's implementation. Still keeping the scaling option in `BlockPreconditioner` for now, in case it's useful in some context other than fine system scaling, e.g. in a coupled coarse-level solve or block-smoother.

src/coreComponents/linearAlgebra/interfaces/MatrixBase.hpp
```diff
@@ -617,6 +619,27 @@ class MatrixBase : public virtual LinearOperator< VECTOR >
     R.multiply( AP, dst );
   }
 
+  /**
+   * @brief Compute the triple product <tt>dst = P1^T * this * P2</tt>
+   * @param P1 the left "prolongation" matrix
+   * @param P2 the right "prolongation" matrix
+   * @param dst the resulting product matrix (will be re-created as needed)
+   */
+  virtual void multiplyPtAP( Matrix const & P1,
```
castelletto1: Should a different name be used? Like this it suggest that a there is a unique prolongation operator.

src/coreComponents/linearAlgebra/interfaces/hypre/HypreMatrix.cpp
```diff
@@ -1070,6 +1069,89 @@ void HypreMatrix::extractDiagonal( HypreVector & dst ) const
...
+  GEOS_LAI_ASSERT( ready() );
+  GEOS_LAI_ASSERT_EQ( localMat.numRows(), numLocalRows() );
+  GEOS_LAI_ASSERT_EQ( localMat.numColumns(), numGlobalCols() );
+
+  hypre::CSRData< true > const diag{ hypre_ParCSRMatrixDiag( unwrapped() ) };
+  hypre::CSRData< true > const offd{ hypre_ParCSRMatrixOffd( unwrapped() ) };
+  HYPRE_BigInt const * const colMap = hypre::getOffdColumnMap( unwrapped() );
+  globalIndex const firstLocalCol = jlower();
+
+  forAll< hypre::execPolicy >( 1,//localMat.numRows(),
```
castelletto1: `1` or `localMat.numRows()`?

src/coreComponents/linearAlgebra/interfaces/MatrixBase.hpp
klevzoff: Sure, the only alternative I can think of is `multiplyP1tAP2` so I'll go with that unless there are other suggestions

src/coreComponents/linearAlgebra/interfaces/hypre/HypreMatrix.cpp
klevzoff: Hmm, good catch!

src/coreComponents/linearAlgebra/interfaces/MatrixBase.hpp
castelletto1: `multiplyP1tAP2` is unambiguous

src/coreComponents/linearAlgebra/solvers/PreconditionerNull.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+#include "linearAlgebra/common/PreconditionerBase.hpp"
+
+namespace geos
+{
+
+/**
+ * @brief Common interface for null (zero) preconditioning operator
+ * @tparam LAI linear algebra interface providing vectors, matrices and solvers
+ */
```
castelletto1, line 30: What is the purpose of the null preconditioner? I would imagine that removing the preconditioner action is achieved by the identity preconditioner. Also, when I read null preconditioner, I tend to think to the null space of an operator.

paveltomin (approved):

src/coreComponents/linearAlgebra/solvers/PreconditionerNull.hpp
klevzoff: Fairly ridiculous purpose to be honest: in multiscale I do
```c++
m_preSmoother = preOrPost == PreOrPost::pre || preOrPost == PreOrPost::both
              ? LAI::createPreconditioner( smoother_params )
              : std::make_unique< PreconditionerNull< LAI > >();

m_postSmoother = preOrPost == PreOrPost::post || preOrPost == PreOrPost::both
               ? LAI::createPreconditioner( smoother_params )
               : std::make_unique< PreconditionerNull< LAI > >();
```
so that later I can call the smoother unconditionally, without checking for null
```c++
auto const applySmoother = [&]( LevelData const & level, Operator const * smoother )
{
  GEOS_MARK_SCOPE( smoother );
  for( integer s = 0; s < m_params.multiscale.smoother.numSweeps; ++s )
  {
    smoother->apply( level.rhs, level.tmp );
    level.sol.axpy( 1.0, level.tmp );
    level.matrix->residual( level.tmp, level.rhs, level.rhs );
  }
};
```
Here having `PreconditionerNull` set `tmp` to 0 means `sol` and `rhs` both will not be modified, which is the intended effect... except it does waste both compute and memory bandwidth and could arguably be just be skipped with a null check :laughing: 

I should really refactor this...

castelletto1: Thanks for the clarification

BASELINE_NOTES.md
```diff
@@ -6,6 +6,10 @@ This file is designed to track changes to the integrated test baselines.
 Any developer who updates the baseline ID in the .integrated_tests.yaml file is expected to create an entry in this file with the pull request number, date, and their justification for rebaselining.
 These notes should be in reverse-chronological order, and use the following time format: (YYYY-MM-DD).
 
+PR #2427 (2025-08-09) <https://storage.googleapis.com/geosx/integratedTests/baseline_integratedTests-pr2427-12699-941dbff.tar.gz>
+=====================
+Change default value of amgNumFunctions from 1 to 3 for solid mechanics solvers. Change in mesh partitioning of PoroElastic_hybridHexPrism_co2 cases due to Scotch version update.
```
paveltomin, line 11: @klevzoff @castelletto1 @victorapm do we need to double-check the `amgNumFunctions` change for bigger models or it is safe to assume positive impact?

klevzoff: The actual value used during solve hasn't changed - it was always 3. But it used to be set in `postInputInitialization()` and for some reason the value didn't make it into restart files. Now the value is already set in the constructor of `SolidMechanicsLagrangianFEM` by the time it is registered with the data repository, and it causes the value to be preserved in restart file.

klevzoff:
Need codeowners approval @rrsettgast

dkachuma (approved):

src/coreComponents/linearAlgebra/common/PreconditionerBase.hpp
klevzoff: It inherits from `LinearOperator` which has a virtual destructor, this makes destructors of the whole hierarchy virtual. I've been trying to gradually remove some boilerplate code like this where it doesn't seem to add much.

src/coreComponents/linearAlgebra/interfaces/direct/SuperLUDist.cpp
klevzoff: There's one use of `GEOS_MARK_FUNCTION` in `apply()`. I instrumented this class a lot more when I was having some performance problems with it, but have since removed most of the markers.

PR #2430, opened 2023-05-10
Feature: Allow negative pressure 

untereiner:


untereiner:
@dkachuma can this PR be reviewed as it is ? Should it be converted to a draft ?

dkachuma:
> @dkachuma can this PR be reviewed as it is ? Should it be converted to a draft ?
I'm not sure about merging this. I think we need to have a discussion about the merits and dangers of allowing negative pressures with the wider community. i think this should be set to draft for now until we get more views/thoughts.

Have you verified that the change fixes the issue with the target model?

untereiner:
> Have you verified that the change fixes the issue with the target model?

Yes it does !

untereiner:
Hello @dkachuma, what about this PR ? Should it be made up to date ? closed ?

paveltomin:
negative pressure is allowed by default for single phase and can be disabled with an option when needed
having negative pressure in compositional is not a good sign and should be avoided at all cost imho

PR #2433, opened 2023-05-11
Fix assert variable scope.

untereiner:


wrtobin (approved):

src/coreComponents/mesh/generators/CellBlockManager.cpp
```diff
@@ -306,7 +306,8 @@ void populateFaceMaps( Group const & cellBlocks,
       NodesAndElementOfFace const & f0 = *first;
       CellBlock const & cb = cellBlocks.getGroup< CellBlock >( f0.blockIndex );
       localIndex const numNodesInFace = cb.getFaceNodes( f0.cellIndex, f0.faceNumber, nodesInFace );
-//      GEOS_ASSERT_EQ( numNodesInFace, f0.numNodes );
+      GEOS_ASSERT_EQ( numNodesInFace, numNodes );
```
TotoGaz, line 309: did you run the integrated tests ?

wrtobin: The assert isn't present in any release build, which is where we run the integratedTests. 

This should only effect behavior in a debug build.

TotoGaz: But how can we be sure that the assertion is legit then?

wrtobin: The assertion is intended to help identify low-hanging issues when running in a debug build typically. Testing assertions like this are really supposed to be part of unit testing rather than regression testing, though I have no idea whether we have unit test coverage of this package.

The `NodesAndElementOfFace` structure has no `numNodes` member, so the previous assertion wasn't valid, which is why it was commented (by me while I was working on the hip/rocm branch). Others also encountered the issue during debug build and proposed this fix for the assert prior to my merge commenting it out.

Issue #2431, opened 2023-05-10
Issue with Python and CMake version < 3.18


untereiner:
**Describe the bug**
At configure step:

```
- Not building pygeosx.
CMake Error at /cmake-3.17.3/share/cmake-3.17/Modules/FindPackageHandleStandardArgs.cmake:261 (message):
  No REQUIRED_VARS specified for FIND_PACKAGE_HANDLE_STANDARD_ARGS()
Call Stack (most recent call first):
  /cmake-3.17.3/share/cmake-3.17/Modules/FindPython/Support.cmake:2437 (find_package_handle_standard_args)
  /cmake-3.17.3/share/cmake-3.17/Modules/FindPython3.cmake:309 (include)
  cmake/thirdparty/SetupGeosxThirdParty.cmake:772 (find_package)
  cmake/CMakeBasics.cmake:49 (include)
  CMakeLists.txt:48 (include)

```

The problem comes with #2265 and the find of python even if `ENABLE_PYGEOSX` is `OFF` in `SetupGeosxThirdParty.cmake`.
Given that [cmake issue](https://gitlab.kitware.com/cmake/cmake/-/issues/20655)  (corrected with cmake v3.18) an empty REQUIRED is necessary.

**To Reproduce**
clone and launch cmake on Pangea2

**Platform (please complete the following information):**
- Machine Pangea2 
- Compiler:  gcc 8.3.0
- GEOSX Version: develop
- CMake version: 3.17.3

XL64:
Same for me, it needs to remove the version range in the find_package()

PR #2437, opened 2023-05-15
fix compatibility with cmake 3.17

XL64:
Fixes issue https://github.com/GEOS-DEV/GEOS/issues/2431

### DIFF
--- a/src/cmake/thirdparty/SetupGeosxThirdParty.cmake
+++ b/src/cmake/thirdparty/SetupGeosxThirdParty.cmake
@@ -755,8 +755,16 @@ endif()
 ################################
 # Python
 ################################
+message(" CMAKE_VERSION ${CMAKE_VERSION} ")
+if( ${CMAKE_VERSION} VERSION_LESS "3.19" )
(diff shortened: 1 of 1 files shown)


src/cmake/thirdparty/SetupGeosxThirdParty.cmake
```diff
@@ -755,8 +755,16 @@ endif()
 ################################
 # Python
 ################################
+message(" CMAKE_VERSION ${CMAKE_VERSION} ")
+if( ${CMAKE_VERSION} VERSION_LESS "3.19" )
+    set( PYTHON_AND_VERSION Python3 )
+    set( PYTHON_OPTIONAL_COMPONENTS)
+else()
```
TotoGaz, line 762: @rrsettgast do you think the `cmake` upgrade will happen soon or it still makes sense to merge this first?

TotoGaz (approved):
@rrsettgast do you think the `cmake` upgrade will happen soon or it still makes sense to merge this first?

jhuang2601 (approved):
Thanks for fixing this compatibility issue and it works on Pangea2.

PR #2439, opened 2023-05-15
Fix regression in config-build

### LINKED ISSUE #2438: Regression in config-build
**Describe the bug**
It is not possible anymore to pass -DGEOSX_TPL_DIR to the config-build.py script and use it in the host-config file.

**To Reproduce**
Try to build on P3 with official host-config file

**Expected behavior**
Complains about missing -DGEOSX_TPL_DIR parameter

- Machine Pangea3
- Compiler: gcc
- GEOSX Version latest

**Additional context**
Regression due to:

```
git show beec4f5b4d -- scripts/
commit beec4f5b4d2cde2ff5e308f74200443c646a804f
Author: Victor A. P. Magri <50467563+victorapm@users.noreply.github.com>
Date:   Tue May 2 10:55:09 2023 -0400

    Expose BoomerAMG options (#2136)

    * Check if using old macro to set spgemm
    * Expose amgInterpolationMaxNonZeros option
    * Expose amgAggressiveInterpType option
    * Expose amgSeparateComponents option
    * Expose amgRelaxWeight and add call to BoomerAMGSetRelaxWt
    * Add LOG_RANK_0 call for hypre's version
    * Add HYPRE_RELEASE_VERSION info
    * Add hypre branch info
    * Move GEOSX Version output to inside the library
    * Use HYPRE_BRANCH_NAME info
    * Update aggressiveNumLevels description
    * Add Enums for AMG interpolation types
    * Add HYPRE_BoomerAMGSetNumPaths call and expose its input parameter
    * Update logLevel meanings
    * Allow modification of default CMake arguments via config-build script
    * changed initialization list to make_tuple
    * Update CompositionalMultiphaseFVM recipe to use Chebyshev smoother on the coarsest MGR level
    * Change default smoother for MGR's coarsest level AMG to Chebyshev
    * Change default smoother for poromechanics AMG's F-relaxation to Chebyshev
    * Add caliper annotation to hypre functions
    * Add CoarseningType enum
    * Update xml according to recent AMG name changes
    * Annotate setupPreconditioningMatrix
    * Updating integratedTests hash
    * Setting a symmetric smoother for CG in testExternalSolvers

    ---------

    Co-authored-by: Francois Hamon <francois.hamon@totalenergies.com>
    Co-authored-by: Nicola Castelletto <38361926+castelletto1@users.noreply.github.com>
    Co-authored-by: Nicola Castelletto <castelletto1@llnl.gov>

diff --git a/scripts/config-build.py b/scripts/config-build.py
index a9a8eb8c1..b3bbaca1c 100755
--- a/scripts/config-build.py
+++ b/scripts/config-build.py
@@ -227,14 +227,14 @@ def main(calling_script, args, unknown_args):
         cmake_line.append("--graphviz=dependency.dot")
         dot_line = "dot -Tpng dependency.dot -o dependency.png"

+    # Append cache file at the end of the command line to make previous argument visible to the cache.
+    cmake_line.append("-C%s" % cache_file)
+
     for unknown_arg in unknown_args:
         if not unknown_arg.startswith('-D'):
             logging.warning("Additional argument '%s' does not start with '-D'. Keeping it nevertheless." % unknown_arg)
         cmake_line.append(unknown_arg)

-    # Append cache file at the end of the command line to make previous argument visible to the cache.
-    cmake_line.append("-C%s" % cache_file)
-
     cmake_line.append(os.path.normpath(os.path.join(scripts_dir, "..", "src")))

     # Dump the cmake command to file for convenience
```

XL64:
Fix issue #2438

### DIFF
--- a/scripts/config-build.py
+++ b/scripts/config-build.py
@@ -206,13 +206,13 @@ def main(calling_script, args, unknown_args):
         cmake_line.append("--graphviz=dependency.dot")
         dot_line = "dot -Tpng dependency.dot -o dependency.png"
 
-    # Append cache file at the end of the command line to make previous argument visible to the cache.
-    cmake_line.append("-C%s" % cache_file)
-
     for unknown_arg in unknown_args:
         if not unknown_arg.startswith('-D'):
             logging.warning("Additional argument '%s' does not start with '-D'. Keeping it nevertheless." % unknown_arg)
         cmake_line.append(unknown_arg)
+
+    # Append cache file at the end of the command line to make previous argument visible to the cache.
+    cmake_line.append("-C%s" % cache_file)
 
     cmake_line.append(os.path.normpath(os.path.join(scripts_dir, "..", "src")))
 


TotoGaz (approved):

scripts/config-build.py
```diff
@@ -206,14 +206,14 @@ def main(calling_script, args, unknown_args):
...
-    # Append cache file at the end of the command line to make previous argument visible to the cache.
-    cmake_line.append("-C%s" % cache_file)
-
     for unknown_arg in unknown_args:
         if not unknown_arg.startswith('-D'):
             logging.warning("Additional argument '%s' does not start with '-D'. Keeping it nevertheless." % unknown_arg)
         cmake_line.append(unknown_arg)
 
+    # Append cache file at the end of the command line to make previous argument visible to the cache.
+    cmake_line.append("-C%s" % cache_file)
```
TotoGaz, line 215: @victorapm Why did you put this line before the `-D` arguments in your PR?
Was there something that you needed done?

victorapm: Oops! This change was not supposed to get pushed yet, sorry for doing that!

At some point, I was experimenting with oversubscribing cmake options that were already defined in the `.cmake` file (HYPRE_DIR to be more precise). So I flipped the append order of `cmake_line` to make that possible. But clearly I wasn't care of side effects...

Do you believe we could have this behavior with the build script? i.e., over-subscribe options defined on the `.cmake` file by default when they are passed via command line to the build script?

TotoGaz: My take it that this script is supposed to be an entry point that wraps `cmake` (which can be complicated to tame if you're not used to it regularly).

If you need something really specific witch `cmake`, I'd recommend that you consider getting the information from the `cmake_cmd` file, generated by `config-build.py`.

For example, on my side, I never use this python script, but `cmake` directly.

victorapm: Good point! Using cmake directly is probably the best way to go in the situation I described

XL64: @victorapm , You can have test if the variable is already defined in your host-config file. This is what I do to check user gave GEOSX_TPL_DIR on Pangea3 and what was broken when the line is moved. Not sure I answer your question correctly, maybe I misunderstood.

victorapm: Thanks, all!

PR #2455, opened 2023-05-17
Years, days, hrs, min, sec time format.

CusiniM:
This small PR introduces another format for the time in the GEOS log. I find it useful for field-scale problems. 

Suggestions are welcome for a better name than "human".

The output looks like this
````
Time: 0 years, 0 days, 0 hrs, 0 minutes, 0 secs, dt: 1000000s, Cycle: 0
Time: 0 years, 11 days, 13 hrs, 46 minutes, 40 secs, dt: 10000000s, Cycle: 1
Time: 0 years, 127 days, 7 hrs, 33 minutes, 20 secs, dt: 50000000s, Cycle: 2
Time: 1 years, 341 days, 0 hrs, 26 minutes, 40 secs, dt: 50000000s, Cycle: 3
Time: 3 years, 189 days, 17 hrs, 20 minutes, 0 secs, dt: 50000000s, Cycle: 4
Time: 5 years, 38 days, 10 hrs, 13 minutes, 20 secs, dt: 50000000s, Cycle: 5
Time: 6 years, 252 days, 3 hrs, 6 minutes, 40 secs, dt: 50000000s, Cycle: 6
Time: 8 years, 100 days, 20 hrs, 0 minutes, 0 secs, dt: 50000000s, Cycle: 7
Time: 9 years, 314 days, 12 hrs, 53 minutes, 20 secs, dt: 50000000s, Cycle: 8
````

IntegratedTests PR: https://github.com/GEOS-DEV/integratedTests/pull/5

### CODE REVIEW

**Comment by cssherman** [comment]
@CusiniM - I've definitely thought about doing this before, but never got around to it!  I think that it would be useful the user could configure this behavior.  For example, they could provide `timestamp="seconds"`, and get the existing output, provide another input like `days` or `years`.  I'd call something like you are presenting here `full` or `year-day-...`

**Comment by cssherman** [comment]
Perhaps an `auto` option, which will trim off higher-level zero timestamp options would be good as well!

**Comment by CusiniM** [comment]
> @CusiniM - I've definitely thought about doing this before, but never got around to it! I think that it would be useful the user could configure this behavior. For example, they could provide `timestamp="seconds"`, and get the existing output, provide another input like `days` or `years`. I'd call something like you are presenting here `full` or `year-day-...`

we can make it customizable. My only question is  to what extent. For example, when specifying `timeOutputFormat="days"` do we want to output:

`Time: 1.5 days `
or 
`Time: 1 days, 12 hours, 0 minutes, 0 seconds`

or do we want to give the option to the user to specify exactly how to break it down so that `years-seconds` would result in
`Time: 0 years,  129600 seconds`
?

**Comment by CusiniM** [comment]
What I have done is pretty simple but it does give the option to choose between the various units and have a `full` output format. It is certainly enough for what I need atm. I think this can be reviewed now.

File: src/coreComponents/events/EventManager.cpp [hunk 1]
```diff
@@ -75,6 +76,10 @@ EventManager::EventManager( string const & name,
     setRestartFlags( RestartFlags::WRITE_AND_READ ).
     setDescription( "Index of the current subevent." );
 
+  registerWrapper( viewKeyStruct::timeOutputFormat(), &m_timeOutputFormat ).
+    setInputFlag( InputFlags::OPTIONAL ).
+    setRestartFlags( RestartFlags::NO_WRITE ).
```
klevzoff, line 81: Despite the `NO_WRITE` flag, check if this still needs a rebaseline. IIRC, each registered wrapper still creates a Conduit node and an HDF5 group in the restart files, regardless of whether data is actually written.

src/coreComponents/events/EventManager.cpp
```diff
@@ -218,4 +223,46 @@ bool EventManager::run( DomainPartition & domain )
...
+{
+  if( m_timeOutputFormat==TimeOutputFormat::full )
+  {    
+    integer const yearsOut   =   m_time / YEAR;
+    integer const daysOut    =  (m_time - yearsOut * YEAR) / DAY;
+    integer const hoursOut   =  (m_time - yearsOut * YEAR - daysOut * DAY) / HOUR;
+    integer const minutesOut =  (m_time - yearsOut * YEAR - daysOut * DAY - hoursOut * HOUR) / MINUTE;
+    integer const secondsOut =   m_time - yearsOut * YEAR - daysOut * DAY - hoursOut * HOUR - minutesOut * MINUTE;
+
+    GEOS_LOG_RANK_0( GEOS_FMT( "Time: {} years, {} days, {} hrs, {} min, {} s, dt: {}s, Cycle: {}", yearsOut, daysOut, hoursOut, minutesOut, secondsOut, m_dt, m_cycle ) );
```
klevzoff, line 236: For consistency, can we also have add a space in `dt: {}s` (here and below)?

klevzoff (approved):
Looks good!

CusiniM:
@jeannepellerin just tagging you to let you know about this since I am guessing it could be useful for users.

src/coreComponents/events/EventManager.cpp
CusiniM: your are right, unfortunately a rebaseline is necessary.

CusiniM:
> This reminds me that the GEOS log assumes users are using a consistent unit system with the second (s) as the unit of time. This will not be true for very large/slow or very small/fast systems. For example, shock modeling typically uses μs/mm/mg. Would it be possible to also have the user specify their simulation time unit to address this?

This is too late for this PR. Can you create an issue about this? I think the choice was to make the code always operate in SI units. It's true that one could pick all parameters in a different unit system (as long as it is consistent) but I am not sure all constitutive models will work appropriately ( I am guessing most but probably not all). If we want to allow for different unit systems I think we should address this a bit more carefully than by just allowing the time units to change.

PR #2457, opened 2023-05-18
Upwind enthalpy flux for sourceFlux b.c. for producers (positive fluxes)

CusiniM:
Whenever a `SoruceFlux` boundary is used to specify a producer, the `enthalpy` flux should be upwinded so that the producer produces fluid with the `enthalpy` of the "perforated" cell. This PR modifies the way the source flux b.c. is applied for the single phase thermal cases so that this is the case. This avoids having non physical solutions due to the fact that one can produce fluid at an enthalpy that does not match the one of the perforated cell. The same thing could (IMO should) be done for component source fluxes in the compositional solver.

The alternative would be to require the user to specify an `enthalpy` flux as well but it can get quite tricky to define. I find this a very simple way to mimic the behavior of a rate constraint well.


src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.cpp
```diff
@@ -1041,25 +1043,78 @@ void SinglePhaseBase::applySourceFluxBC( real64 const time_n,
...
+            localRhs[energyRowIndex] += enthalpy[ei][0] * rhsValue;
+
+            globalIndex dofIndices[2]{pressureDofIndex, temperatureDofIndex};
+            real64 jacobian[2]{rhsValue * dEnthalpy_dPressure[ei][0], rhsValue * dEnthalpy_dTemperature[ei][0]};
+
+            localMatrix.template addToRow< serialAtomic >( energyRowIndex,
+                                                           dofIndices,
+                                                           jacobian,
+                                                           2 );
+          }
```
francoishamon, line 1093: Here I recommend adding some logging (when logLevel is 1 or larger) to let the user know what the code does in the two configurations: 
- Thermal + producing source flux: a message saying that both the mass and energy equations in the cell are modified
- Thermal + injecting source flux: a message saying that only the mass equation in the cell is modified

PR #2460, opened 2023-05-18
Bugfix for {fmt} cmake

wrtobin:


jhuang2601 (approved):
Thanks for fixing the reported problem and it should combine with [PR#2437](https://github.com/GEOS-DEV/GEOS/pull/2437) for compiling GEOS on Pangea2.

TotoGaz (approved):

PR #2464, opened 2023-05-18
Ensure unit tests are enabled by default.

### LINKED ISSUE #2463: Re-enable GEOS unit tests
`GEOSX_ENABLE_TESTS` is not set to `ON` in the host-configs or the CI, so default behavior is to not run GEOS unit tests.
CI is currently running mostly LvArray tests (84 tests instead of 184 tests total).

I tried enabling `GEOSX_ENABLE_TESTS`, and ran into a compilation issue on quartz (duplicate definitions of `RAJAHelper`):

```
/usr/WS1/han12/GEOS/src/coreComponents/common/unitTests/testLifoStorage.cpp:23:8: error: redefinition of 'RAJAHelper'
struct RAJAHelper
       ^
/usr/WS1/han12/GEOS/src/coreComponents/LvArray/src/Macros.hpp:684:8: note: previous definition is here
struct RAJAHelper
       ^
/usr/WS1/han12/GEOS/src/coreComponents/common/unitTests/testLifoStorage.cpp:29:8: error: redefinition of 'RAJAHelper<RAJA::policy::loop::loop_exec>'
struct RAJAHelper< serialPolicy >
       ^~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/WS1/han12/GEOS/src/coreComponents/LvArray/src/Macros.hpp:690:8: note: previous definition is here
struct RAJAHelper< serialPolicy >
       ^
/usr/WS1/han12/GEOS/src/coreComponents/common/unitTests/testLifoStorage.cpp:41:8: error: redefinition of 'RAJAHelper<RAJA::PolicyBaseT<RAJA::Policy::openmp, RAJA::Pattern::forall, RAJA::Launch::undefined, camp::resources::Platform::host, RAJA::policy::omp::Parallel, RAJA::wrapper<RAJA::policy::omp::omp_for_schedule_exec<RAJA::policy::omp::Auto>>>>'
struct RAJAHelper< parallelHostPolicy >
       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/WS1/han12/GEOS/src/coreComponents/LvArray/src/Macros.hpp:701:8: note: previous definition is here
struct RAJAHelper< parallelHostPolicy >
```

wrtobin:
Resolves #2463


wrtobin:
Regardless we're back to the expected 185 unit tests by default, with GEOS_ENABLE_TESTS FALSE allowing suppression of the JUST the geos unit tests.

src/docs/doxygen/GeosxConfig.hpp
```diff
@@ -30,7 +30,7 @@
 #define GEOSX_USE_MPI
 
 /// Enables use of OpenMP (CMake option ENABLE_OPENMP)
-/* #undef GEOSX_USE_OPENMP */
+#define GEOSX_USE_OPENMP
```
CusiniM, line 33: are you sure you want to commit this file?

host-configs/ORNL/frontier-base.cmake
```diff
@@ -23,7 +23,7 @@ set( GEOSX_BUILD_OBJ_LIBS OFF CACHE BOOL "" FORCE )
 set( ENABLE_GTEST_DEATH_TESTS OFF CACHE BOOL "" )
 set( gtest_disable_pthreads ON CACHE BOOL "" )
 
-set( ENABLE_TESTS OFF CACHE BOOL "" FORCE )
+set( GEOS_ENABLE_TESTS OFF CACHE BOOL "" FORCE )
```
CusiniM, line 26: I am guessing they don't work on Frontier?

CusiniM (approved):

scripts/uberenv/packages/geosx/package.py
```diff
@@ -488,9 +488,9 @@ def cmake_args(self):
         # options.append(self.define_from_variant('BUILD_SHARED_LIBS', 'shared'))
 
         # if '~tests~examples~benchmarks' in spec:
-        #     options.append('-DENABLE_TESTS=OFF')
+        #     options.append('-DGEOS_ENABLE_TESTS=OFF')
```
CusiniM, line 491: I have never seen this file tbh.

host-configs/ORNL/frontier-base.cmake
wrtobin: They didn't reliably config / build for a long time, but neither do a ton of our kernels still. At some point I'll do a PR devoted to ensuring they all work now that Frontier is (relatively) stable

src/docs/doxygen/GeosxConfig.hpp
wrtobin: the doxygen/GeosxConfig.hpp just gets added every time someone `git add -u` after doing a build that changes it, so it winds up in a lot of commits.

wrtobin: There's ever chance it should be .gitignore-d but I'm not 100% on that

klevzoff: We have a [way](https://github.com/GEOS-DEV/GEOS/blob/33f3f4d1feb09a0bc81ae3a0b29bc25d68a617b1/src/cmake/GeosxConfig.cmake#L70-L79) to "fix" all cmake variables used to generate this file so it doesn't diff, but the subset of variables currently covered is rather small, and as we keep adding new config macros we usually forget to extend that set.

We've decided we can't ignore this file or remove it from source control because it needs to be current for Doxygen docs to build correctly, whatever those are worth...

TotoGaz: Why do we even commit this file? If it's generated, it should only live in the build dir?
And when it comes to filling in the versions (e.g. for `python` or `hypre`), why don't we build a simple bootstrap executable that could be able to extract those version programmatically? `cmake` would run it during the building process.

klevzoff: We commit it because `readthedocs` doesn't have the environment necessary to generate it (i.e. a compiler toolchain, `cmake` and a `thirdPartyLibs` installation).

As for versions, those are already filled by cmake directly at config time.

TotoGaz: > We commit it because `readthedocs` doesn't have the environment necessary to generate it (i.e. a compiler toolchain, `cmake` and a `thirdPartyLibs` installation).

Can you point me where the information contained in this file is used in `readthedocs`? I could not find it.
E.g. https://github.com/GEOS-DEV/GEOS/blob/develop/src/docs/sphinx/buildGuide/Dependencies.rst is maintained by hand (and not up to date 🤣 )

> As for versions, those are already filled by cmake directly at config time.

By parsing source files, which is not the best for sure. But whatever...

klevzoff: This file is specifically for building doxygen docs hosted on `readthedocs`: https://geosx-geosx.readthedocs-hosted.com/en/latest/doxygen_output/html/_geosx_config_8hpp.html. Not sure how many people are using those, but as long as we support it, it needs to be there, and it better be generated the same way and from the same template as the one that will be generated locally in a developer's build tree.

Parsing source files is not ideal, but to build an executable to extract versions programmatically, we still need a build environment, at least a C++ compiler and a linker... then again, the only reason those version macros exist in `GeosxConfig.hpp` is because we have not implemented version extraction via runtime methods (except for MPI). This can most certainly be improved.

bmhan12:
This configuration section can be updated with the `GEOS_ENABLE_TESTS` option, or maybe a separate PR that updates all the options at once:
https://geosx-geosx.readthedocs-hosted.com/en/latest/docs/sphinx/buildGuide/BuildProcess.html?highlight=ENABLE#configuration-options

wrtobin:
Off yeah that needs some updating. I'll get a documentation PR put together with updates for this and hip/rocm, plus any other name changes.

CusiniM:
I tried to compile on lassen and I get the compilation errors that we see on the CI. My IDE complains about not finding `#include <RAJA/RAJA.hpp>` in that Macro file. I am not sure if that's part of the problem of if it has nothing to do with it.

wrtobin:
@CusiniM My most recent commit fixed the issue for me from lassen, so hopefully the CI will succeed as well.

CusiniM:
Goood news: the CI is passing (thanks @wrtobin!)
bad news: the usual suspect timed out. I ve just restarted it. It s probably taking because of the unit tests though which before were not being compiled... let s see if it makes it. otherwise we may have to only make the geosx target for that job.

src/coreComponents/finiteElement/unitTests/CMakeLists.txt
```diff
@@ -10,8 +10,8 @@ set(testSources
     testH1_Wedge_Lagrange1_Gauss6.cpp
     testH1_Pyramid_Lagrange1_Gauss5.cpp
     testH1_TriangleFace_Lagrange1_Gauss1.cpp
-    testQ3_Hexahedron_Lagrange_GaussLobatto.cpp
-    testQ5_Hexahedron_Lagrange_GaussLobatto.cpp
+    # testQ3_Hexahedron_Lagrange_GaussLobatto.cpp
+    # testQ5_Hexahedron_Lagrange_GaussLobatto.cpp
```
rrsettgast, line 14: Is this intentional?

src/coreComponents/linearAlgebra/unitTests/testVectors.cpp
```diff
@@ -150,7 +150,7 @@ TYPED_TEST_P( VectorTest, copyConstruction )
   using Vector = typename TypeParam::ParallelVector;
 
   Vector x;
-  createAndAssemble< parallelDevicePolicy<> >( 3, x );
+  createAndAssemble< geos::parallelDevicePolicy<> >( 3, x );
```
rrsettgast, line 153: maybe just add a `using namespace geos` at the top of this source?

rrsettgast (approved):

src/coreComponents/finiteElement/unitTests/CMakeLists.txt
wrtobin: The `GaussLobatto` kernels are all disabled with HIP right now due to similar issues encountered when compiling for MI250x

src/coreComponents/linearAlgebra/unitTests/testVectors.cpp
wrtobin: Unfortunately the unit test compile time is pushing our worst case CI back up to 3 hours, it is on its 3rd attempt to pass.

CusiniM: I would suggest adding `--build-exe-only` here: https://github.com/GEOS-DEV/GEOS/blob/f12d9c93cf1bc1c7cfc243ac42067a498c387171/.travis.yml#L179 

coz even if it finally passes, all following PRs will battle with this.

CusiniM: we can turn them back on as soon as we move to `githubactions`.

Issue #2331, opened 2023-03-09
Renaming `geosx` into `geos`


TotoGaz:
Hello guys, since it was decided to rename `geosx` into `geos`, it’s now time to take the appropriate actions and reflect the change in our code and organisations!

I’ve listed below what I could come accros. Please feel free to update the list so we’re sure we’re forgetting nothing.
Of course, when I write _"Change ..."_, it means that we have to deal with the aftermath of the change as well (modifications of the scripts, tell our partners, etc).

I did not think about the best order to do all the actions yet. We need to sort this out and organize ourselves.

_Regarding github_

- [ ] Change the github organisation from `GEOSX` to `GEOS`
- [ ] Change the github repository from `GEOSX` to `GEOS`

While `github` provides forwarding from old name to new name, our `jenkins`/`gitlab`/`travis` will need to be updated accordingly.

_Regarding other organisations_

- [ ] Change the `readthedocs` organisation and project (be careful, the url is hard coded in the C++ code).
- [ ] Change the GCP bucket names that contain `geosx`. 
- [ ] Change the `travis` organisation and repositories.
- [ ] Change the [geosx.org](http://geosx.org/) website.
- [ ] Update the deployment folders that often contain `geosx`.
- [ ] Update the `dockerhub` organisation.

_Regarding source code_

- [ ] Change the occurences of geosx in the C++ `namespace`, `GEOSX_` log/error macros, include guards, and any other variable name containing `geosx`. Be careful of the `uncrustify` issues.
- [ ] Do the same for other code (not C++), like our `cmake` code or our travis anchors.
- [ ] Documentation should not mention `geosx` anymore.
- [ ] Rewrite the _"How does GEOSX relate to GEOS?"_ question!
- [ ] Change the cmake variables to `GEOS_`, update doc accordingly.
- [ ] Change the name of the `geosx` executable. Update `ATS` that uses it. Any other component? Tell the end users!
- [ ] Rename geosxats.
- [ ] Rename `pygeosx`.

Anything else?

This name change will cost us some time 🤣

rrsettgast:
I would say we do the org/repo/readthedocs/travis in one day. We just have to pick the day.

TotoGaz:
> I would say we do the org/repo/readthedocs/travis in one day. We just have to pick the day.

Is it as easy as this, or do we have to contact the support for each of these products?

rrsettgast:
[...]

PR #2466, opened 2023-05-19
Install both `geosx` and `geos` executables.

TotoGaz:
Both `geosx` and `geos` will be available in the installation folder, in order to let the end users adapt their scripts.
Only the `geosx` executable is built in the build folder.

Relates to https://github.com/GEOS-DEV/GEOS/issues/2331


TotoGaz:
@rrsettgast what do you think of this patch?

rrsettgast (changes requested):
Perhaps update documentation at this point?

rrsettgast (approved):
just commit it, and make an issue to follow up the deprecation messages, and the documentation if you wish.

TotoGaz:
I've just added the appropriate entry to https://github.com/GEOS-DEV/GEOS/issues/2331

PR #2468, opened 2023-05-19
Preparatory work for local chop implementation

paveltomin:
Add functionality in DofManager to be able to do local chop for solution deltas.


src/coreComponents/linearAlgebra/DofManager.hpp
```diff
@@ -569,11 +570,11 @@ class DofManager
    * @param scalingFactor a factor to scale vector values by
    * @param mask component selection mask (for partial copy)
    */
-  template< typename FIELD_OP, typename POLICY >
+  template< typename FIELD_OP, typename POLICY, typename SCALING_FACTOR_TYPE >
   void vectorToField( arrayView1d< real64 const > const & localVector,
                       string const & srcFieldName,
                       string const & dstFieldName,
-                      real64 scalingFactor,
+                      SCALING_FACTOR_TYPE const & scalingFactor,
```
francoishamon, line 577: With the new template `SCALING_FACTOR_TYPE`, it is nice to be able to avoid some code duplication. It just breaks a little bit the uniformity of the signature because if we follow the convention used for the other arguments, we should have `scalingFactorName` when it is a string, and just `scalingFactor` when it is a `real64`. So I would not be against more duplication to keep this uniformity. But this is really a minor point, no need to change anything now

francoishamon (approved):
Thanks @paveltomin this is going to be very useful

src/coreComponents/linearAlgebra/DofManager.hpp
francoishamon: It is in the merge queue, so it will be merged after the PRs ahead of it in the merge queue. At this point, there is not much to do. I added the flag "ready to be merged".

PR #2469, opened 2023-05-19
Enable poroelasticity in hydraulic fracture solver

CusiniM:
https://github.com/GEOS-DEV/integratedTests/pull/8


jhuang2601 (approved):
Great job for turning on this modelling capability!

binw1 (approved):
Tested the code with the kgdViscosityDominated_poroelastic model and a 3-layer water injection induced model. Both models exhibited poroelastic effects in the matrix. Especially for the 3-layer model, the pressure in the fracture is now higher than the total stress around the fracture (previously the pressure in the fracture is only higher than the effective stress around the fracture). But in the 3-layer model, negative elementAperture around the fracture tip is still observed before T = 510.4 seconds.

PR #2472, opened 2023-05-23
add missing geosx_core_libs removed in #2453

XL64:


rrsettgast (approved):

XL64:
All test failed, seems to have side effect on other configurations (SHARED GEOSX_CORE), any Idea on that @wrtobin

XL64:
I built on P3 with shared enabled it also works...

XL64:
Ok it was link to an issue on develop, i reran the jenkins test all pass

PR #2476, opened 2023-05-24
Fix compilation bug with the last version of develop

acitrain:
This small PR fix a bug in the compilation of develop after the last two merge inside it.
The last merged file do not respect the new organisation of the folder with fluid solvers so develop does not compile anymore.
This PR was just to put the correct include and make the branch compile.


acitrain:
> Thanks, I have already fixed it in the next PR in the merge queue which I will merge today. I will close this PR.

Ho ok no problem !

PR #2480, opened 2023-05-24
Reduce compilation time by removing some kernel instantiations.

CusiniM:


CusiniM:
it does seem to help a bit.

wrtobin:
Looks like ~7 minutes for the offending CI case, which means a lot when we're usually 5 minutes from the timeout cap.

wrtobin (approved):

PR #2492, opened 2023-05-30
Implement k-value initialisation

dkachuma:
- Uses Wilson correlation
- Adds unit tests


francoishamon (approved):
Looks great, thanks @dkachuma

PR #2495, opened 2023-05-31
Adding list of types dispatch capabilities.

CusiniM:
Needs to be merged after #2496

https://github.com/GEOS-DEV/thirdPartyLibs/pull/236


src/coreComponents/common/TypeDispatch.hpp
```diff
@@ -277,7 +342,38 @@ bool dispatch( TypeList< Ts... > const types,
...
+ * @return true
+ * @return false
+ */
+template< typename LIST, typename LAMBDA, typename ... Ts >
+bool dispatchCombinations( LIST const combinations,
+                           LAMBDA && lambda,
+                           Ts && ... objects )
+{
+  bool const success = internal::dispatchViaTable( combinations,
+                                                   std::make_tuple( std::type_index( typeid(objects))... ),
```
wrtobin, line 369: `createTypeIndexTuple( combinations )`?

wrtobin: Since we're using a variadic parameter here necessitating it is the last parameter, can we restructure the downstream calls to order the arguments similarly? (i.e. lambda 2nd instead of last, type tuple list last in place of `objects` since we process `objects` to get that argument)?

wrtobin (approved):

src/coreComponents/common/TypeDispatch.hpp
```diff
@@ -254,30 +300,38 @@ bool dispatchViaTable( TypeList< Ts... > const types,
...
+//                                                   std::make_tuple( std::type_index( internal::typeIdWrapper(objects))... ) );
+                                                   std::make_tuple( std::type_index( typeid(objects))... ) );
+
+  if( !success )
   {
-    GEOS_ERROR( "Type " << LvArray::system::demangle( type.name() ) << " was not dispatched.\n" <<
-                "Check the stack trace below and revise the type list passed to dispatch().\n" <<
-                "If you are unsure about this error, please report it to GEOSX issue tracker." );
+    std::ostringstream dispatchInput;
+    ( (dispatchInput<<"\n  "<<LvArray::system::demangle( internal::typeIdWrapper( objects ).name() ) ), ...);
```
rrsettgast, line 329: @klevzoff can this be done without the `ostringstream`?

wrtobin: IIRC something like:

`(... + ( "  \n" + LvArray::system::demangle( internal::typeIdWrapper( objects ).name() ) ) ))`

should work, though I think that will construct many temporary strings.

wrtobin: (assuming `.name()` is returning a string and on a character pointer

src/coreComponents/common/TypeDispatch.hpp
```diff
@@ -254,30 +300,38 @@ bool dispatchViaTable( TypeList< Ts... > const types,
...
+template< typename LIST, typename LAMBDA, typename ... Ts >
+bool dispatch( LIST const combinations,
+               LAMBDA && lambda,
+               Ts && ... objects )
 {
-  bool const success = internal::dispatchViaTable( types, type, std::forward< LAMBDA >( lambda ) );
-  if( !success && errorIfTypeNotFound )
+  bool const success = internal::dispatchViaTable( combinations,
+                                                   std::forward< LAMBDA >( lambda ),
+//                                                   std::make_tuple( std::type_index( internal::typeIdWrapper(objects))... ) );
```
CusiniM, line 323:
```suggestion
```

CusiniM: How is this now working without using `internal::typeIdWrapper(objects)`?

wrtobin: The unittest uses the base types, so I would expect that to pass with this, but yeah I'm surprised we're not erroring out when applying this to `WrapperBase` in the other locations. We don't have any mechanism to transparently unwrap the wrapped type, so I would expect we're `typeid()`-ing `WrapperBase` in all the non unit-test locations in this PR.

rrsettgast: I was keeping the failure to work on the error message

PR #2499, opened 2023-06-01
Implemented RCM to reorder degrees of freedom

francoishamon:
This PR implements an option to reorder the degrees of freedom of a given field in the `DofManager`. Currently the only reordering option available is the reverse CutHill-McKee reordering method, and its implementation is borrowed from hypre. The reordering is done once at the beginning of the simulation when we populate the dof number arrays. 

(I did not find a way to perform the reordering after the calls to `computeFieldDimensions` and `createIndexArray`, so both of these functions are now called twice).

The implementation seems to work in parallel and for multiphysics problems. I am going to test it to see if it accelerates the GPU simulations involving ILU (no need to reorder dofs in hypre anymore at every preconditioner call).

Ready for review.


src/coreComponents/linearAlgebra/utilities/ReverseCutHillMcKeeOrdering.cpp
```diff
@@ -0,0 +1,330 @@
...
+  for( i = 0; i < l2; i++ )
+  {
+    marker[level_j[i]] = -1;
+  }
+
+  numLevelsRef = numLevels;
+}
+
+void
+ReverseCutHillMcKeeOrdering::
```
victorapm, line 210: Maybe it makes sense to use `std::sort`?

src/coreComponents/linearAlgebra/utilities/ReverseCutHillMcKeeOrdering.cpp
```diff
@@ -0,0 +1,330 @@
...
+  findNodeWithMinDegree( localIndex const numRows,
+                         arrayView1d< localIndex > const degree,
+                         arrayView1d< localIndex > const marker,
+                         localIndex & rootRef )
+{
+  localIndex i = 0;
+  localIndex minDegree = numRows + 1;
+  localIndex root = 0;
+
+  for( i = 0; i < numRows; i++ )
```
victorapm, line 71: This loop can benefit from threading

src/coreComponents/linearAlgebra/utilities/ReverseCutHillMcKeeOrdering.cpp
```diff
@@ -0,0 +1,330 @@
...
+{
+  localIndex i = 0;
+  localIndex root = 0;
+  localIndex currentNum = 0;
+  localIndex const numRows = perm.size();
+
+  // get the degree for each node
+  array1d< localIndex > degree( numRows );
+  array1d< localIndex > marker( numRows );
+  for( i = 0; i < numRows; i++ )
```
victorapm, line 39: Threading

src/coreComponents/linearAlgebra/DofManager.cpp
```diff
@@ -175,7 +176,107 @@ void forMeshSupport( std::vector< DofManager::FieldSupport > const & support,
...
+    field.numLocalDof  = 0;
+    field.rankOffset   = 0;
+    field.numGlobalDof = 0;
+    field.globalOffset = 0;
+    if( fieldIndex > 0 )
+    {
+      field.blockOffset = 0;
+    }
+    return;
+  }
```
klevzoff, line 224: I like early exit in functions (as opposed to nesting everything else in the `else` branch), but it'd also be nice to not repeat the whole "reset field offsets" block twice. Maybe this function could be refactored a bit by:
1. Isolating `forAll` loop for filling a "trivial" permutation into its own (free or member) function, and RCM-specific code into its own (member) function.
2. `computePermutation()` keeps all common "setup" code (up to and including `computeFieldDimensions()` + allocate `permutation`), followed by a `switch` on `field.reorderingType`, followed by common "teardown" code (reset `field` and return `permutation`).

src/coreComponents/linearAlgebra/DofManager.cpp
```diff
@@ -367,6 +468,12 @@ void DofManager::addField( string const & fieldName,
   addField( fieldName, location, components, support );
 }
 
+void DofManager::enableLocalReordering( string const & fieldName )
+{
+  FieldDescription & field = m_fields[getFieldIndex( fieldName )];
+  field.reorderingType = LocalReorderingType::ReverseCutHillMcKee;
```
klevzoff, line 474: I quite like that you made reordering type an `enum` instead of a `bool`! I know we only have a single non-trivial reordering type, but would it make it more explicit and visible at the call site if we passed the desired `LocalReorderingType` value into this function? This way one could also "disable" it by passing a value of `None`.

src/coreComponents/linearAlgebra/utilities/ReverseCutHillMcKeeOrdering.hpp
```diff
@@ -0,0 +1,165 @@
...
+
+#ifndef GEOS_LINEARALGEBRA_UTILITIES_REVERSECUTHILLMCKEEORDERING_HPP_
+#define GEOS_LINEARALGEBRA_UTILITIES_REVERSECUTHILLMCKEEORDERING_HPP_
+
+#include "common/DataTypes.hpp"
+
+namespace geos
+{
+
+class ReverseCutHillMcKeeOrdering
```
klevzoff, line 27: Would you consider making this a namespace with a single entry point (`computePermutation`), and placing all `private` functions as free functions in the `.cpp` file (keep `static` on them to avoid unneeded external linkage)?

klevzoff (approved):
Very nice implementation, I'm glad the approach worked! Just a few comments in addition to @victorapm's.

src/coreComponents/linearAlgebra/utilities/ReverseCutHillMcKeeOrdering.cpp
```diff
@@ -0,0 +1,330 @@
...
+  localIndex row = 0;
+  localIndex minDegree = 0;
+  localIndex levDegree = 0;
+  localIndex numLevels = 0;
+  localIndex newNumLevels = 0;
+  localIndex root = rootRef;
+
+  // at most numRows levels
+  array1d< localIndex > level_i( numRows+1 );
+  array1d< localIndex > level_j( numRows );
```
dkachuma, line 106: I'm not sure if these 2 are allocations. But if they are, I would allocate once in `computePermutation` and pass those arrays as parameters for each call to `findPseudoPeripheralNode`. Maybe there's a threading aspect I'm missing.

victorapm (approved):

src/coreComponents/linearAlgebra/utilities/ReverseCutHillMcKeeOrdering.cpp
francoishamon: Done, I added a RAJA loop with `parallelHostPolicy`

src/coreComponents/linearAlgebra/utilities/ReverseCutHillMcKeeOrdering.cpp
francoishamon: Here I left the `qsort` for now because it seems specialized to handle the presence of the permutation, if I understood correctly. However, your comment made me think that I could use `std::swap` instead of the custom swap that was there before

src/coreComponents/linearAlgebra/utilities/ReverseCutHillMcKeeOrdering.hpp
francoishamon: Done, I placed all the private functions as free functions in the cpp file.

src/coreComponents/linearAlgebra/DofManager.cpp
francoishamon: Done, this function is now called `setLocalReorderingType` and takes the reordering type as input

src/coreComponents/linearAlgebra/DofManager.cpp
francoishamon: Done, I split the function as you suggested

src/coreComponents/linearAlgebra/DofManager.cpp
```diff
@@ -175,7 +176,107 @@ void forMeshSupport( std::vector< DofManager::FieldSupport > const & support,
...
+    // reset the offsets, since they will be recomputed with the proper number of components
+    // this is important to get the right reordering for multiphysics problems
+    field.numLocalDof  = 0;
+    field.rankOffset   = 0;
+    field.numGlobalDof = 0;
+    field.globalOffset = 0;
+    if( fieldIndex > 0 )
+    {
+      field.blockOffset = 0;
+    }
```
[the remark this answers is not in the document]
francoishamon: Done

src/coreComponents/linearAlgebra/DofManager.cpp
```diff
@@ -175,7 +176,107 @@ void forMeshSupport( std::vector< DofManager::FieldSupport > const & support,
 
 } // namespace
 
-void DofManager::createIndexArray( FieldDescription const & field )
+
+void DofManager::computePermutation( FieldDescription & field,
+                                     array1d< localIndex > & permutation )
```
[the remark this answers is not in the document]
francoishamon: Done

src/coreComponents/linearAlgebra/DofManager.cpp
```diff
@@ -1406,11 +1513,24 @@ void DofManager::reorderByRank()
...
 
+  std::map< string, array1d< localIndex > > permutations;
+
+  // First loop: compute the local permutation
+  for( FieldDescription & field : m_fields )
+  {
+    // compute local permutation of dofs, if needed
+    array1d< localIndex > permutation;
+    permutations[field.name] = permutation;
+    computePermutation( field, permutations.at( field.name ) );
```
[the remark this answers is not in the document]
francoishamon: Done

src/coreComponents/linearAlgebra/utilities/ReverseCutHillMcKeeOrdering.cpp
francoishamon: Here I added a RAJA loop as well

src/coreComponents/linearAlgebra/utilities/ReverseCutHillMcKeeOrdering.cpp
victorapm: Good! Makes sense!

src/coreComponents/linearAlgebra/DofManager.cpp
```diff
@@ -173,9 +174,114 @@ void forMeshSupport( std::vector< DofManager::FieldSupport > const & support,
...
+  array1d< localIndex > permutation( numLocalDofs( field.name ) );
+
+  // if no reordering is requesting, we just return the identity permutation
+  if( field.reorderingType == LocalReorderingType::None )
+  {
+    fillTrivialPermutation( permutation );
+  }
+  else
+  {
+    fillTrivialPermutation( permutation );
```
klevzoff, line 216: Is a trivial permutation required as input for the RCM routine? I hadn't studied the algorithm in enough detail, but my previous comments were based on an assumption you wouldn't need to fill it in this branch.

francoishamon: Yes, given the way I implemented it, I have to call `createIndexArray` in a first pass to generate the degrees of freedom (assuming `numComps = 1`) here: 
https://github.com/GEOS-DEV/GEOS/blob/e021fa7b3f65b5d129c221db16c9b280eb46a3e0/src/coreComponents/linearAlgebra/DofManager.cpp#L239-L243
I did this to try to reuse the existing code as much as possible, but this is a little bit annoying I agree with you

PR #2502, opened 2023-06-01
Remove GeosxConfig.hpp file that should not be in the repo

paveltomin:
Remove src/docs/doxygen/GeosxConfig.hpp


klevzoff:
We actually keep this file in the repo on purpose. It is needed to build [doxygen docs](https://geosx-geosx.readthedocs-hosted.com/en/latest/doxygen_output/html/_geosx_config_8hpp.html) on `readthedocs`, which lacks the full build environment needed to generate this header (e.g. `cmake` and dependencies).

We have a ["solution"](https://github.com/GEOS-DEV/GEOS/blob/04841c14859d1145bdef77a11d05d97d6fd0a998/src/cmake/GeosxConfig.cmake#L70-L79) for avoiding the diffs to this file between platforms, but it has not been updated with the addition of new macro definitions. I would suggest turning this PR into a fix for that rather than a deletion of the header. If anyone has a better long-term solution, I'm interested!

untereiner:
I was wondering, why are config values like the build type in the documentation ? Doxygen doesn’t need them for the generation process. Am I missing something ?

klevzoff:
> I was wondering, why are config values like the build type in the documentation ? Doxygen doesn’t need them for the generation process. Am I missing something ?

The value of this `GEOSX_CMAKE_BUILD_TYPE` macro is used in code in one [place](https://github.com/GEOS-DEV/GEOS/blob/04841c14859d1145bdef77a11d05d97d6fd0a998/src/coreComponents/common/initializeEnvironment.cpp#L192).

Basically, we want that header to be more or less an exact copy of the one a developer will find in their build directory and included everywhere in GEOS (up to the values that we "fix"). The doxygen build will not fail without it, but will be "incomplete" in some sense. E.g. someone inspecting source on doxygen won't be able to navigate to `GeosxConfig.hpp` or jump to the documentation of one of these macros they encounter in code.

We just need to apply the "fix" for the missing macros and forget about this problem. We may also want to consider splitting the "dependency version" macros into a separate header and not store that one in the repo, since those are likely to change often.

untereiner:
> Basically, we want that header to be more or less an exact copy of the one a developer will find in their build directory and included everywhere in GEOS (up to the values that we "fix")

This file is only to let a developer (who somehow reads the online doxygen doc) know that the macro definitions like `GEOSX_USE_XXX` exists ?

PR #2508, opened 2023-06-05
Added a warning for porosity above 1

### LINKED ISSUE #2505: Error message for wrong porosity range
Hello Everyone,

We were working on a CO2 injection model in a tetrahedral grid with one injection well.
This case was run with:

- PANGEA-2 machine

- gcc 8.3.0 compiler

- the Geos version 2e19665

Even though the porsity was wrong (2.5 actually instead of 0.25), the run on Geos worked.
So, as users a error message would be a great help for wrong porosity range.
You'll find below our simulation  folder  (the logs "job_GEOS_954119.out" is for Porosity=2.5 and "job_GEOS_513781.out" for porosity =0.25).
[Test1.zip](https://github.com/GEOS-DEV/GEOS/files/11636365/Test1.zip)

francoishamon:
This PR adds a warning when GEOS detects a porosity above 1 at initialization.

Fixes #2505


rrsettgast:
Shouldn't this be "corrected" (i.e. set to `1`) and then issue the warning that it was corrected?

francoishamon:
> Shouldn't this be "corrected" (i.e. set to `1`) and then issue the warning that it was corrected?

Yes, that is a possibility as well. But there is a common practice in reservoir simulation that consists in having cells with very large pore volumes on the sides of the model to impose boundary conditions (it is an alternative to aquifer boundary conditions). In GEOS, this would be done by forcing a very large porosity in the boundary cells.  To make sure that this is still doable, I think it is good to just issue a warning and not correct the value by default.

rrsettgast:
> > Shouldn't this be "corrected" (i.e. set to `1`) and then issue the warning that it was corrected?
> 
> Yes, that is a possibility as well. But there is a common practice in reservoir simulation that consists in having cells with very large pore volumes on the sides of the model to impose boundary conditions (it is an alternative to aquifer boundary conditions). In GEOS, this would be done by forcing a very large porosity in the boundary cells. To make sure that this is still doable, I think it is good to just issue a warning and not correct the value by default.

But it is a physically unattainable configuration to have a porosity > 1? What would be the difference between a `porosity==1` vs `porosity > 1`?

dkachuma:
A porosity of more than 1 is not physically possible. The idea is that people want to have a large volume at the boundary without having too many cells or changing the cell geometry - it's a cheap way (in terms of setup and computation) of adding volume to the model. In traditional simulators, a pore volume multiplier is used for this whilst keeping the actual cell porosity within physical bounds. I suppose the idea of a pore volume multiplier is not in the scope of GEOS so the only option for the user to do this kind of workflow is the porosity.

Of course porosity values of more than 1 cause some models to fail eg. Carman-Kozeny. But as long as the user has been warned that should be enough I think.

dkachuma (approved):
Looks good

PR #2513, opened 2023-06-07
Use 32bits coordinates inside wave propagators

XL64:
In order to reduce memory usage in wave propagation kernels on GPU we use 32 bits coordinates instead of 64 bits.

Integrated tests PR: https://github.com/GEOS-DEV/integratedTests/pull/25


src/coreComponents/mesh/NodeManager.hpp
```diff
@@ -431,6 +443,7 @@ class NodeManager : public ObjectManagerBase
   //START_SPHINX_REFPOS
   /// reference position of the nodes
   array2d< real64, nodes::REFERENCE_POSITION_PERM > m_referencePosition;
+  array2d< real32, nodes::REFERENCE_POSITION_PERM > m_referencePosition32;
```
CusiniM, line 446: I think it would be better to avoid "polluting" the `NodeManager` an keep everything inside the wave equations physics solvers. Hopefully, we will get to the needed refactoring  soon enough for this not to be needed any more.

TotoGaz: Yes, for the moment, if you need 32 bits positions, please perform the conversion inside your solvers. This shouldn't be a problem and that way we try to keep the architecture as it is while we try to refactor.

TotoGaz (changes requested):
Please move conversion to your solvers.

XL64:
@TotoGaz are you ok with the change ?

rrsettgast:
rather than hard-code the floating point type, wouldn't it be better to make a solver-specific alias for this? So for instance, since you have encapsulated the change inside the wave solvers, in `WaveSolverBase.hpp`, you could have something like:

```c++
using wsCoordType = real32;
```
and just have `wsCoordType` replace all the places that you have replace `real64` with `real32`. Then you can switch back and forth by changing one line of code.

francoishamon:
Hello @acitrain @sframba @XL64 
I am moved out the PR of the merge queue to give you some time to address @rrsettgast's comment.
Let me know if you want me to re-do the integrated tests rebaseline later

Issue #2465, opened 2023-05-18
[EPIC] PVT modeling


francoishamon:
## Description
The goal of this issue is to keep track of all developments related to PVT modeling capabilities for CO2 in depleted reservoirs.

## Work Scope

Basic first steps to port PVTPackage capabilities in GEOS:
1. Create and (unit) test functions to solve Rachford-Rice equation (@francoishamon)
2. Create and (unit) test functions to model cubic EOS (@francoishamon)
3. Add initialization heuristics for K-values (@dkachuma)
4. Add and (unit) test the negative two-phase flash loop (@dkachuma)

Steps to test the new functions in `PVTDriver`

5. Create a class derived from `MultiFluidBase` to read an input and call the new flash functions

Steps to make the new code usable for simulations of CO2 injection in depleted reservoirs:

6. Add Soreide Whitson modification to the cubic EOS functions (for the water phase)
7. Add phase identification in the two-phase flash to identify the water phase
8. Add property calculations for density (CO2 with impurities, etc) and viscosity (LBC and others)
9. Compute finite-difference derivatives

Steps to make the code efficient:

10. Modify two-phase flash to do non-iterative K-value flash
11. Compute analytical derivatives
12. Explore Newton bisection in Rachford-Rice

## Issues in this EPIC
- https://github.com/GEOS-DEV/GEOS/pull/1726
- https://github.com/GEOS-DEV/GEOS/pull/2451
- https://github.com/GEOS-DEV/GEOS/pull/2456

paveltomin:
_Compute finite-difference derivatives_ - would you consider using AD instead? Flash is local, so AD operations are much easier.

PR #2520, opened 2023-06-09
Implement Cubic EOS derivatives

dkachuma:
- Adds derivative computation for mixture coefficients and compressibility factor
- Adds unit tests to compare calculated values with finite difference values

#2465


francoishamon (approved):
Looks great, thanks @dkachuma !

PR #2523, opened 2023-06-13
UML diagrams for redesign of infrastructure.

rrsettgast:
This "PR" contains some uml diagrams that are proposals for infrastructure redesign. It is split into 4 main concpets:

1. `Body` is an object to represent the physical body.
2. `Mesh` describes a discretized geometry for a portion of the `Body`. There can be multiple overlapping `Mesh` objects for a single `Body`
    - This contains only geometry...so far
    - NO field data
    - Should we hold Property data?
3. `Discretization` is intended to hold data for the numerical method applied to the `Mesh`. There can be multiple `Discretization` of a single `Mesh`.
    - As the UML shows, we are talking about a `Discretization` type for `FEM`, `Stencil`, and `Particle` discretization. 
4. `ShivaInterface` is a placeholder for our Discretization library. It holds:
    - `FiniteElementInterface`...which is an interface for applying FEM
    - `StencilInterface`.... interface for stencil based methods
    - `ParticleInterface`... interface for particle methods


untereiner:
Quick questions
- is Body a connected component ? Then could we have multiple connected components in the same simulation ?
- should `Mesh` not only describe the topology, the conbinatorics of the discretization ? You could then `embed` this discretization into a geometrical space. It means you could have multiple dimensions of discretizations: 0d for points, 1d for edges, 2D for faces, etc and multiple dimensions for embeddings: embed a 2D object into xD space.
- Incidence and adjacence relations between cells have to be written for each mesh type and each dimension ? ( it is maybe already the case. I didn’t check)

untereiner:
Another question: Is it planned that the combinatorics of the mesh "body" can change during the simulation (cutting / tearing / fracturing / ... ) ?

rrsettgast:
> Quick questions
> 
> * is Body a connected component ? Then could we have multiple connected components in the same simulation ?

I am not sure what definition of "connected component" you are using. I assume you mean it as it is used in graph theory by which a single "component" would be each collection of connected vertices? 

> * should `Mesh` not only describe the topology, the conbinatorics of the discretization ? You could then `embed` this discretization into a geometrical space. It means you could have multiple dimensions of discretizations: 0d for points, 1d for edges, 2D for faces, etc and multiple dimensions for embeddings: embed a 2D object into xD space.

`Mesh` contains the information to describe the geometry of the body. We haven't got to how exactly we will represent embedded surfaces yet, but they certainly need some sort of a geometry level representation. 


> * Incidence and adjacence relations between cells have to be written for each mesh type and each dimension ? ( it is maybe already the case. I didn’t check)
Well, if I understand the question, for a Cartesian IJK mesh the relations would be implicit. For an unstructured mesh type these relationships may or may not need to be defined in order to use the mesh to perform a simulation. However, we should have the facilities to create all of the relations.

rrsettgast:
> Another question: Is it planned that the combinatorics of the mesh "body" can change during the simulation (cutting / tearing / fracturing / ... ) ?

I would prefer that the mesh geometry stay intact, and we handle cutting/tearing/fracturing through manipulations of the discretization that is overlayed on the mesh geometry. For example if we were to split a node/vertex, it would have the same reference position, so there is no need to store that data twice. Also the creation of a new node may need to know its parent, and then we have an unbalanced relationship between parent and child objects (as in the existing/old code). It may be better to handle this through allocation of new "nodes" which refer to a single "vertex" and have a modified connectivity at a discretization rather than the geometric mesh.

untereiner:
> I am not sure what definition of "connected component" you are using. I assume you mean it as it is used in graph theory by which a single "component" would be each collection of connected vertices?

Yes, a collection of connected cells (vertices, edges, faces, volumes, etc) such as it exists a path between any two cells in a subspace X is connected component. 


> `Mesh` contains the information to describe the geometry of the body. We haven't got to how exactly we will represent embedded surfaces yet, but they certainly need some sort of a geometry level representation.

When I worked with meshes, I usually made a clean separation between the representation of the structure of the object (its “topology”) and the representation of its shape (its “embedding”, usually in some Rn ). That's what I meant. 


> I would prefer that the mesh geometry stay intact, and we handle cutting/tearing/fracturing through manipulations of the discretization that is overlayed on the mesh geometry. 

That's exactly what I thought but it not exactly what comes next. 

> For example if we were to split a node/vertex, it would have the same reference position, so there is no need to store that data twice. 

I agree for the duplication but when you split a node with the consequence of disconnecting a part of the mesh then you should also split the vertex into two because their two. Handling changes of the discretization without bugs is not so easy. I worked on this subject in the past.

>Also the creation of a new node may need to know its parent, and then we have an unbalanced relationship between parent and child objects (as in the existing/old code). It may be better to handle this through allocation of new "nodes" which refer to a single "vertex" and have a modified connectivity at a discretization rather than the geometric mesh.

When you write `parent` you mean `cells` (tetrahedra, etc ?). That's exactly why I suggest a clear separation between `topology` and `embedding`. That's really dependent of the skills of your data structure to handle seamlessly these changes in the discretization.

Is there room for suggestions ?

rrsettgast:
There is always room for suggestions.

untereiner:
It's a bit stiff to start like that but I do not know how to start in an other way....
Some years ago I worked with a data model called combinatorial maps. It is a research object that has implementations in the computer graphics community [here](https://github.com/cgogn/CGoGN_3) and [here](https://doc.cgal.org/latest/Combinatorial_map/index.html). It is also used by the image processing community. There is also a book that explains the model [here](https://www.routledge.com/Combinatorial-Maps-Efficient-Data-Structures-for-Computer-Graphics-and/Damiand-Lienhardt/p/book/9780367378356) and applications.

The most important thing that was not tackled by these communities is its versatility as engineered data structures.

For GEOS it would allow:
- optimized data structures depending on the data (structured or not)
- factor out from the classes all neighborhood queries
- by definition access to all neighborhoods of all cells defined in a mesh whether its dimension. Time and space complexity of these queries will depend on the concrete implementations of the model (see diagram below).
- consistent definition in any dimension (0d: points ; 1d: graphs; 2d: surfaces; 3d: volumes; etc)
- separate the body (physical behavior) from the mesh (topological queries)

Here is a start. I tried to simply the diagram but I assume it seems to be odd. I will complete it when I have time.
If you are interested I can explain more someday.


mesh

acitrain:
Hi @rrsettgast 
As we discuss yesterday, on our side we have two "main" needs for our future scope of work: 

- For future DG work, we need the posibility of using variable polynomial orders inside the DG part 

- And the second stuff is the possibility to define a variable on a nodeSet and not only on the NodeManager as it is done now (I think there is an old issue from 2018 about this but I don't manage to find it....)
This will allow us to define our SEm solution on noe per region for exemple (it will be useful when wi will use different physics in different regions)

Thank you (in advance) for all this work !

TotoGaz:
@untereiner That would be great if you could gather some information and make a presentation one Thursday or another. When do you think you would be ready?

Issue #2526, opened 2023-06-14
Allow compilation GEOS with external components


untereiner:
**What is the requested feature?**
Allows to build a version of GEOS with features that may (in a very long term) or may not be merged into the main project.
The source files would be in another repository and compiled with geos components.

**Is your request related to a specific problem?**
I'm frustrated about my stale branch about RESQML

**Describe the solution you'd like**
The cleanest solution would be to have installed targets and find them.
I investigated it a bit and I had difficulties with BLT.
I have another proposal in a PR linked to the issue.

PR #2527, opened 2023-06-14
Feature in-tree external components

untereiner:
A proposal to solve issue #2526


untereiner:
@TotoGaz Could you take a look please ?

TotoGaz (approved):

src/cmake/GeosxOptions.cmake
```diff
@@ -127,7 +127,7 @@ blt_append_custom_compiler_flag( FLAGS_VAR CMAKE_CXX_FLAGS
                                )
 
 blt_append_custom_compiler_flag( FLAGS_VAR CMAKE_CXX_FLAGS_DEBUG
-                                 GNU "-Wno-unused-parameter -Wno-unused-variable"
+                                 GNU "-Wno-unused-parameter -Wno-unused-variable -Wno-dangling-reference"
```
rrsettgast, line 130: What was this required for? i.e. where is the dangling reference that this avoids?

rrsettgast (approved):
@untereiner Thanks for picking this up. I don't think anyone has used this since we ditched PAMELA and moved PVTPackage as a submodule inside the `src` directory. Have you confirmed that this works with an external package that is called at runtime?

untereiner:
@rrsettgast Yes I confirm. I am using RESQML inputs/outputs with a kind of [plugin](https://github.com/geosiris-technologies/geos-resqml/)

rrsettgast:
> @rrsettgast Yes I confirm. I am using RESQML inputs/outputs with a kind of [plugin](https://github.com/geosiris-technologies/geos-resqml/)

Great. Thanks again.

PR #2537, opened 2023-06-19
Add MGR strategy for SolidMechanicsEmbeddedFractures

CusiniM:
This PR:

- [x] Defines an MGR strategy for SolidMechanicsEmbeddedFractures
- [x] Fixes AMG parameters for staticCond + AMG solution startegy
- [x] Adds 2 Sneddon tests that only rely on a curve check step and employ an iterative solver:
   - [x] A case that uses static condensation implemented in GEOS + boomer AMG
   - [x] A case that employs MGR to perform the static condensation

integratedTests PR: https://github.com/GEOS-DEV/integratedTests/pull/27


victorapm (approved):
Just looked at the linear solver additions and they look good. Thanks!!

PR #2540, opened 2023-06-21
Fixed the use of globalIds in parallel

francoishamon:
This PR uses some code from @TotoGaz to fix the use of globalIds in parallel (currently does not work in `develop`)

This PR also modifies an integrated test to test this feature


TotoGaz (approved):

src/coreComponents/mesh/generators/VTKUtilities.cpp
```diff
@@ -593,44 +593,64 @@ findNeighborRanks( std::vector< vtkBoundingBox > boundingBoxes )
...
-  vtkSmartPointer< vtkDataSet > mesh;
-  bool globalIdsAvailable = loadedMesh.GetPointData()->GetGlobalIds() != nullptr
-                            && loadedMesh.GetCellData()->GetGlobalIds() != nullptr;
+  {
+    // Add global ids on the fly if needed
+    int const me = hasGlobalIds( mesh );
+    int everyone;
+    MpiWrapper::allReduce( &me, &everyone, 1, MPI_MAX, MPI_COMM_GEOSX );
+
+    if( everyone and not me )
```
klevzoff, line 609: Would this be ok?
```suggestion
    int const anyone = MpiWrapper::max( me );

    if( anyone && !me )
```
(I don't object to use of `and`, `not`, `or` etc. keywords in c++ in general, this is just not the style GEOS uses everywhere else, even in this PR)

francoishamon: Yes, I will fix it

TotoGaz: Yes, we'll fix this.

For the context, @francoishamon did cherry pick this in one of my branches.
So I'll fix it in my branch to make my merge easier, if you do not mind!

src/coreComponents/mesh/generators/VTKUtilities.cpp
```diff
@@ -593,44 +593,64 @@ findNeighborRanks( std::vector< vtkBoundingBox > boundingBoxes )
...
+    {
+      mesh->GetPointData()->SetGlobalIds( vtkIdTypeArray::New() );
+      mesh->GetCellData()->SetGlobalIds( vtkIdTypeArray::New() );
```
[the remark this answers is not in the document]
TotoGaz: Yes, it's most likely something like this.
For fracture meshes which do not span the whole ranks, the situation can be even more troublesome.
Global indices are not created/spread if there's no element on the rank.
Maybe we could have spread an empty field in the `vtk` partitioning, I did not have a look at it.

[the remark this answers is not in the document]
klevzoff: Ah, this makes total sense

PR #2541, opened 2023-06-21
Implement Quasi-Newton for HF problems

CusiniM:
Updated baselines: GEOS-DEV/integratedTests/pull/92


src/coreComponents/finiteVolume/SurfaceElementStencil.hpp
```diff
@@ -427,7 +427,7 @@ SurfaceElementStencilWrapper::
                   real64 (& dWeight_dVar1 )[maxNumConnections][2],
                   real64 (& dWeight_dVar2 )[maxNumConnections][2][3] ) const
 {
-  real64 sumOfTrans = 0.0;
+  real64 sumOfTrans = 1e-15;
```
CusiniM, line 430: why do you need this change? can it actually be 0?

src/coreComponents/constitutive/contact/ContactBase.hpp
```diff
@@ -247,7 +247,8 @@ GEOS_FORCE_INLINE
 real64 ContactBaseUpdates::computeHydraulicAperture( real64 const aperture,
                                                      real64 & dHydraulicAperture_dAperture ) const
 {
-  return m_apertureTable.compute( &aperture, &dHydraulicAperture_dAperture );
+  dHydraulicAperture_dAperture = 1;
+  return aperture;
```
CusiniM, line 251: I don't think we won't to allow for 0 hydraulic aperture anyways. In a contact problem it would result in a row of zeros in case of closed fracture. I think that in the other solver I simply added a constant that defines minimum value for the hydraulic aperture https://github.com/GEOS-DEV/GEOS/blob/958e393cc4633b8cd97f830b655a1c2106fde918/src/coreComponents/physicsSolvers/multiphysics/poromechanicsKernels/SinglePhasePoromechanicsFractures.hpp#L72

This should not really matter for quasi-newton though. right?

src/coreComponents/finiteVolume/SurfaceElementStencil.hpp
Guotong-Ren: if the aperture = 0 is the initial condition, the sumOfTrans will be zero and cause an invalid operation in division. So I just give it a small enough number to avoid division by zero.

src/coreComponents/constitutive/contact/ContactBase.hpp
Guotong-Ren: Yes, I don't think adding artificial aperture matters Q-N. But in the KGD test, the performance of QN is a bit different (worse) than the code I tested several months ago. In future, we can test the sensitivity of minimumHydraulicAperture to the nonlinear solver. BTW, is the minimumHydraulicAperture assigned from "defaultAperture" in input xml?

binw1:
Instead of directly changing all nonlinear iterations to be Quasi-Newton iteration, could you add a switch so user can choose from the input file to either use original Newton iteration or this Quasi-Newton iteration? Thanks. @CusiniM @Guotong-Ren

yylicvx:
> Instead of directly changing all nonlinear iterations to be Quasi-Newton iteration, could you add a switch so user can choose from the input file to either use original Newton iteration or this Quasi-Newton iteration? Thanks. @CusiniM @Guotong-Ren

That will be good!

paveltomin:
@Guotong-Ren what should we do with this PR?

src/coreComponents/constitutive/contact/ContactBase.hpp
paveltomin: @Guotong-Ren I removed that code change. First - there could be a numerical problem with zero row that Matteo mentioned, and second - hydraulic aperture responsible for some physics, right? So we better keep the mapping there.

src/coreComponents/finiteVolume/SurfaceElementStencil.hpp
paveltomin: There are few other places with similar treatment, but I am not 100% sure what is the right way to deal with those division by zero.

src/coreComponents/finiteVolume/SurfaceElementStencil.hpp
```diff
@@ -475,6 +475,7 @@ SurfaceElementStencilWrapper::
       real64 const harmonicWeight   = t0*t1 / sumOfTrans;
       real64 const arithmeticWeight = 0.25 * (t0+t1);
 
+      m_meanPermCoefficient = 0;
```
paveltomin, line 478: @Guotong-Ren not sure i fully understand it here - you want to switch from harmonic to arithmetic? does it help with convergence?

paveltomin: > this allows the fluid going into the newly added fracture elements. Otherwise, harmonic average will always result 0 trans if new fracture element's aperture = 0

Very good point, @CusiniM can we frame this change properly?

> BTW, i have a local version merged with Matteo's latest changes on total stress sif. I will push it to this branch.

Thanks and let me know if you will get any conflicts. Sorry for messing up this branch...

CusiniM: In reality, the aperture can never be 0. The hydraulic aperture is the one used here and it's never 0.

Guotong-Ren: I think what you assume is the default aperture is a number larger than zero?

src/coreComponents/finiteVolume/SurfaceElementStencil.hpp
```diff
@@ -624,7 +625,8 @@ SurfaceElementStencilWrapper::
     localIndex const esr =  m_elementSubRegionIndices[iconn][k];
     localIndex const ei  =  m_elementIndices[iconn][k];
 
-    m_weights[iconn][k] = m_weights[iconn][k] / hydraulicAperture[er][esr][ei];
+    real64 const apperture = std::max( hydraulicAperture[er][esr][ei], 1e-15 );
```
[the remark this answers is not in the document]
paveltomin: yeah, i was just chasing some floating point exception here...

src/coreComponents/finiteVolume/SurfaceElementStencil.hpp
paveltomin: Somehow that change seems to fix failing case from docs: https://github.com/GEOS-DEV/GEOS/issues/2632
@CusiniM and @Guotong-Ren could you please double check?

paveltomin:
@CusiniM @Guotong-Ren @binw1 can we merge it as is?

paveltomin (approved):

paveltomin:
@CusiniM @Guotong-Ren is it ok if i move it into merge queue?

Guotong-Ren (approved):

paveltomin:
tests are ok, just the new flag presence difference

paveltomin:
switched kgdToughnessDominated to use QN

PR #2542, opened 2023-06-22
Compositional multiphase fluid properties

dkachuma:
Creates a utility for the calculation of compositional multiphase properties
- Molar density
- Mass density

Adds unit tests to compare against values generated by PVTPackage

#2465


src/coreComponents/constitutive/fluid/multifluid/compositional/functions/CompositionalProperties.cpp
```diff
@@ -0,0 +1,145 @@
...
+                                                   real64 const compressibilityFactor,
+                                                   real64 & molarDensity )
+{
+
+  real64 vEos = gasConstant * temperature * compressibilityFactor / pressure;
+  real64 vCorrected = vEos;
+
+  for( integer ic = 0; ic < numComps; ++ic )
+  {
+    vCorrected = vCorrected + composition[ic] * ( volumeShift[ic] * temperature + volumeShift[ic] );
```
jafranc, line 42: I am puzzled by unit homogeneity ...

src/coreComponents/constitutive/unitTests/TestFluid.hpp
```diff
@@ -67,6 +67,7 @@ class TestFluid
     createArray( testFluid->criticalTemperature, components, Fluid::Tc, Fluid::data );
     createArray( testFluid->acentricFactor, components, Fluid::Ac, Fluid::data );
     createArray( testFluid->molecularWeight, components, Fluid::Mw, Fluid::data );
+    createArray( testFluid->volumeShift, std::vector< real64 >( NC, 0.0 ) );
```
jafranc, line 70: Can we not optionally refactor that for data to be append from `Fluid::data` as the others even though it is only zeros ?

src/coreComponents/constitutive/unitTests/testCompositionalProperties.cpp
```diff
@@ -0,0 +1,536 @@
...
+  }
+
+  // Compares the calculated molar density derivatives against numerical calculated
+  // finite difference values
+  void testMolarDensityDerivative( CompositionalPropertiesTestData< NC > const & data )
+  {
+    real64 const pressure = std::get< 0 >( data );
+    real64 const temperature = std::get< 1 >( data );
+    array1d< real64 > composition;
+    TestFluid< NC >::createArray( composition, std::get< 2 >( data ));
```
jafranc, line 85: Should we refactor that as it will be repeated in all tests to out the inputs and expected output ?

src/coreComponents/constitutive/unitTests/testCompositionalProperties.cpp
```diff
@@ -0,0 +1,536 @@
...
+
+    // Compare against numerical derivatives
+    // -- Pressure derivative
+    real64 const dp = 1.0e-4 * pressure;
+    computeMolarDensity( pressure-dp, temperature, composition, currentMolarDensity );
+    fdDerivative = -(currentMolarDensity - molarDensity) / dp;
+    checkDerivative( dMolarDensity_dp, fdDerivative, "Molar density left pressure derivative" );
+    computeMolarDensity( pressure+dp, temperature, composition, currentMolarDensity );
+    fdDerivative = (currentMolarDensity - molarDensity) / dp;
+    checkDerivative( dMolarDensity_dp, fdDerivative, "Molar density right pressure derivative" );
```
jafranc, line 110: Optionally we could also refactor these left and right test for all _constitutive/unitTests_, it would help readability

jafranc (approved):
Great work ! Thanks

src/coreComponents/constitutive/fluid/multifluid/compositional/functions/CompositionalProperties.cpp
dkachuma: Nice catch. I think the idea in PVT package is to have 2 values with the second one dependent on temperature. But I'm not quite sure this was coded correctly. I think I will remove this and stick to the formula that is better known without the temperature dependency.

src/coreComponents/constitutive/unitTests/testCompositionalProperties.cpp
dkachuma: This is a good idea. I will do a follow up PR to address this across all the unit tests that I have added that do this type of numerical derivative test.

PR #2545, opened 2023-06-23
logLevel

paveltomin:


src/coreComponents/functions/TableFunction.hpp
```diff
@@ -373,6 +385,10 @@ TableFunction::KernelWrapper::interpolateLinear( IN_ARRAY const & input ) const
     }
     else if( input[dim] >= coords[coords.size() - 1] )
     {
+      if(m_logLevel > 0)
+      {
+        GEOS_WARNING("Table " << m_tableName << ": argument value << " << input[dim] << " is above the maximum range limit " << coords[coords.size() - 1] );
```
klevzoff, line 392: I know this is still a draft, but I thought I'd chime in early.

We cannot log anything from within kernels. The mere presence of `printf` statements in code (even when not invoked) has a very detrimental effect on CUDA compilation (e.g. allocating additional registers). It might be acceptable to have these in the CPU compilation path, but you would need to guard each one with `#ifdef __CUDA_ARCH__`, and ideally do the same for `m_logLevel` variable (making sure it's never captured into a GPU kernel to avoid increasing argument size yet again).

We could come up with an abstraction that encapsulates these details (something that is 0 bytes/no-op on GPU but prints on CPU). But to me it's not obvious that printing potentially thousands of warning messages is even useful or desirable. Perhaps detecting this condition at runtime but only reporting in aggregate (by atomically writing a warning flag in some predetermined location and checking it after the kernel launch) is a possible workaround?

Thoughts? @rrsettgast @wrtobin @CusiniM @francoishamon

wrtobin: We could go a similar direction as the LVARRAY_ERROR_IF macros and deal with the different compilation contexts when defining the macro itself. 

As it is I had to disable it completely on HIP as the printf, as you say, causes a lot of issues in compilation and causes many kernels to fail to compile with HIP.

Preventing erroneous capture of the log-level variable is really what would require additional #ifdef guarding, and we would want to be certain additionally that strings like `m_tableName` are also never used in a way they would be visible on the device compile pass.

wrtobin: flagging a kernel "return status" with a flag variable and logging once the kernel completes (in the case of warnings) is in my mind the most direct route to get the type of behavior we probably want

wrtobin: This is aside from the kernel concerns:

In general for logging I think we should prefer to use `GEOS_FMT` to produce our logging strings instead of assuming we'll be using streams under the hood of our logging macros.

MelReyCG:
Hi @paveltomin, I may have worked on the same subject in the following PR: #2552 
I used another approach were I pre-validate the input on the PVT `TableFunction` before the kernel in launched, therefore avoiding  to log a warning from inside of the kernel.
Also, I throw an error by default, which can be disabled by an attribute. The goal is to warn the user with certainty (without writing a huge log).

Don't hesitate to review it!

paveltomin:
> Hi @paveltomin, I may have worked on the same subject in the following PR: #2552 I used another approach were I pre-validate the input on the PVT `TableFunction` before the kernel in launched, therefore avoiding to log a warning from inside of the kernel. Also, I throw an error by default, which can be disabled by an attribute. The goal is to warn the user with certainty (without writing a huge log).
> 
> Don't hesitate to review it!

Thanks for the information and invitation to review.
I've seen you PR and it looks good to me, good luck with merging it!

PR #2546, opened 2023-06-23
Some formulation options

paveltomin:
Few experimental options for flow solver:

`useTotalMassEquation` - enabled by default, when disabled - no total mass equation will be formed

`useSimpleAccumulation` - disabled by default, when enabled - the accumulation term will be assembled using simple form: `PV * rho_c`, mathematically equivalent to the current version, numerically might be slightly different. Simplifies accumulation linearization.

`minCompDens` - default value 1e-10 (same as before), allows experiments with minimum component ammount.

Also added zero diagonal check in accumulation linearization.

Packed flags into bit flags to pass into kernels.

Changed `co2_flux_3d.xml` to use new flags.

Rebaseline PR: https://github.com/GEOS-DEV/integratedTests/pull/56


paveltomin:
review anyone?

paveltomin:
review anyone? all changes are flag-protected and rebaseline required only because of the new flags introduced in xml

klevzoff (approved):
Nice and clean implementation! Sorry this took so long to review. Just a couple minor comments on `BitFlags`.

src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseBaseKernels.hpp
```diff
@@ -720,6 +743,9 @@ class ElementBasedAssemblyKernel
         real64 const dPhaseCompAmount_dP = dPhaseAmount_dP * phaseCompFrac[ip][ic]
                                            + phaseAmount * dPhaseCompFrac[ip][ic][Deriv::dP];
 
+        GEOS_ASSERT_MSG( ic == 0 && LvArray::math::abs( dPhaseCompAmount_dP ) < minDensForDivision,
+                         "Zero diagonal in Jacobian" );
```
paveltomin, line 747: @klevzoff is it allowed or bad?

klevzoff: Having an assertion in kernel is fine, keep in mind it is only enabled in debug builds. You want to negate the condition here though, `GEOS_ASSERT` works opposite of `GEOS_ERROR_IF`.

paveltomin: Just corrected, thanks!

PR #2554, opened 2023-06-29
Sequential flow-mechanics follow-up, iteration 2

francoishamon:
This PR modifies the sequential flow-mechanics solution strategy:
- [X] Switches the order of the solvers: first flow then mechanics
- [X] Averages the porosity over quadrature points
- [X] Changes the formulation of the porosity computation in `BiotPorosity`

This PR fixes some small bugs found while running simulations:
- Disabled porosity update after initialization (should not be called, as we don't want initialization displacements to be used to compute porosity)
- Made sure that statistics are always displayed (which is not the case when a `PeriodicEvent` has an `endTime` equal to the `maxTime` of the simulation)
- Allows the initialization procedure for sequential simulations.
- Adds checks for target regions in the poromechanics solvers.

Rebaseline done in https://github.com/GEOS-DEV/integratedTests/pull/20


src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.cpp
```diff
@@ -702,7 +702,10 @@ void SinglePhaseBase::implicitStepComplete( real64 const & time,
 
       CoupledSolidBase const & porousSolid =
         getConstitutiveModel< CoupledSolidBase >( subRegion, subRegion.template getReference< string >( viewKeyStruct::solidNamesString() ) );
-      porousSolid.saveConvergedState();
+      if( !m_keepFlowVariablesConstantDuringInitStep )
+      {
+        porousSolid.saveConvergedState();
```
paveltomin, line 707: if it skips saving, what is going to be stored there and does it matter or not?

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoromechanics.cpp
```diff
@@ -213,9 +213,6 @@ void MultiphasePoromechanics::assembleSystem( real64 const GEOS_UNUSED_PARAM( ti
 
   solidMechanicsSolver()->getMaxForce() = LvArray::math::max( mechanicsMaxForce, poromechanicsMaxForce );
 
-  // tell the flow solver that this is a stress initialization step
-  flowSolver()->keepFlowVariablesConstantDuringInitStep( m_performStressInitialization );
```
paveltomin, line 217: so this you just moved to a more appropriate place, right? not sure i follow the logic here...

paveltomin (approved):

frankfeifan (approved):
This PR is looking good. Thanks, Francois! I now realize why the first time step has relatively poor convergence after initialization.

src/coreComponents/constitutive/solid/CompressibleSolid.hpp
```diff
@@ -70,7 +74,7 @@ class CompressibleSolidUpdates : public CoupledSolidUpdates< NullModel, PORO_TYP
                                            real64 const & oldHydraulicAperture,
                                            real64 const & newHydraulicAperture ) const
   {
-    m_porosityUpdate.updateFromPressureAndTemperature( k, q, pressure, 0.0, 0.0, 0.0 );
+    m_porosityUpdate.updateFromPressureAndTemperature( k, q, pressure, 0.0, 0.0, 0.0, 0.0, 0.0 );
```
paveltomin, line 77: you sure about zeros here, would "pressure" be more appropriate or it does not matter?

francoishamon: I just followed the logic used before in this function because the zeros are not going to be used to update the porosity [here](https://github.com/GEOS-DEV/GEOS/blob/f45f89423c0020e471f88fc340cdc40524ff78ab/src/coreComponents/constitutive/solid/porosity/PressurePorosity.hpp#L52). But you are right, we should pass `pressure` here even if the function will not use it, it will be easier to understand and less error prone if someone ever changes the implementation of `PressurePorosity`.

src/coreComponents/constitutive/solid/porosity/BiotPorosity.hpp
```diff
@@ -87,7 +87,7 @@ class BiotPorosityUpdates : public PorosityBaseUpdates
                                                real64 & dPorosity_dTemperature ) const
   {
     real64 const biotSkeletonModulusInverse = (m_biotCoefficient[k] - m_referencePorosity[k]) / m_grainBulkModulus;
-    real64 const porosityThermalExpansion = 3 * m_thermalExpansionCoefficient[k] * m_biotCoefficient[k];
+    real64 const porosityThermalExpansion = 3 * m_thermalExpansionCoefficient[k] * ( m_biotCoefficient[k] - m_referencePorosity[k] );
```
paveltomin, line 90: you sure about this whole thing?

francoishamon: I am following the book by Coussy (2004):

Screen Shot 2023-07-03 at 9 40 22 AM

where the thermal expansion of the porosity is given by:

Screen Shot 2023-07-03 at 9 40 57 AM

and in particular equations (4.21c) and (4.44)

paveltomin: makes sense, thank you
@binw1 helped me to double check and it looks good

binw1: @francoishamon We further reviewed the formula here:
porosity = porosity_n
               + biotCoefficient * meanStressIncrement / bulkModulus // change due to stress increment
               + biotSkeletonModulusInverse * deltaPressureFromBeginningOfTimeStep // change due to pressure increment
               - porosityThermalExpansion * deltaTemperatureFromBeginningOfTimeStep // change due to temperature increment
               + fixedStressPressureCoefficient * deltaPressureFromLastIteration // fixed-stress pressure term
               + fixedStressTemperatureCoefficient * deltaTemperatureFromLastIteration; // fixed-stress temperature term

And it seems the first term on the RHS should be biotCoefficient * (volumetricStrain_k - volumetricStrain_n), could you please double check it? Thanks.

frankfeifan: > @francoishamon We further reviewed the formula here: porosity = porosity_n + biotCoefficient * meanStressIncrement / bulkModulus // change due to stress increment + biotSkeletonModulusInverse * deltaPressureFromBeginningOfTimeStep // change due to pressure increment - porosityThermalExpansion * deltaTemperatureFromBeginningOfTimeStep // change due to temperature increment + fixedStressPressureCoefficient * deltaPressureFromLastIteration // fixed-stress pressure term + fixedStressTemperatureCoefficient * deltaTemperatureFromLastIteration; // fixed-stress temperature term
> 
> And it seems the first term on the RHS should be biotCoefficient * (volumetricStrain_k - volumetricStrain_n), could you please double check it? Thanks.

Hi @binw1, I think it should be `meanStressIncrement` here instead of the strain as we are using the fixed stress algorithm for the sequential solver. Note that this `computePorosity` function is only used for the sequential solver. 

@francoishamon maybe it is more precise if we change `meanStressIncrement` to `meanTotalStressIncrement` to indicate that we are using the fixed total stress algorithm.

francoishamon: @binw1  and @frankfeifan regarding this comment and [the following comment](https://github.com/GEOS-DEV/GEOS/pull/2554#discussion_r1253474355): the two options are doing the same thing unless I am missing something, we just have to decide which one we prefer:

Option 1: pass the mean total stress increment
```
porosity = porosity_n 
+ biotCoefficient * meanTotalStressIncrement / bulkModulus // change due to stress increment 
+ biotSkeletonModulusInverse * (pressure - pressure_n) // change due to pressure increment 
- porosityThermalExpansion * (temperature - temperature_n) // change due to temperature increment 
+ fixedStressPressureCoefficient * (pressure - pressure_n) // fixed-stress pressure term 
+ fixedStressTemperatureCoefficient * (temperature - temperature_n); // fixed-stress temperature term
```
where
```
meanTotalStressIncrement = meanEffectiveStressIncrement 
- biotCoefficient * (pressure_k - pressure_n ) 
- 3 alpha ( temperature_k  - temperature_n ) 
```

Option 2: pass the mean effective stress increment
```
porosity = porosity_n 
+ biotCoefficient * meanEffectiveStressIncrement / bulkModulus // change due to stress increment 
+ biotSkeletonModulusInverse * (pressure - pressure_n) // change due to pressure increment 
- porosityThermalExpansion * (temperature - temperature_n) // change due to temperature increment 
+ fixedStressPressureCoefficient * (pressure_k - pressure_n) // fixed-stress pressure term 
+ fixedStressTemperatureCoefficient * (temperature_k - temperature_n); // fixed-stress temperature term
```
where
```
meanEffectiveStressIncrement = bulkModulus * ( strainIncrement[0] + strainIncrement[1] + strainIncrement[2] )
```

and I agree with @frankfeifan that we can be more precise on the naming by using the terms `effective` or `total` to qualify the stress.

My preference goes to Option 2 because it makes it easier to understand the time levels of pressure/temperature terms (rather than have pressure and temperature buried in the stress).

paveltomin: Option 2 (implementation in this branch) is like that:
```
porosity = porosity_n
               + biotCoefficient * meanStressIncrement_k / bulkModulus // change due to stress increment (at the previous sequential
                                                                       // iteration)
               + biotSkeletonModulusInverse * (pressure - pressure_n) // change due to pressure increment
               - porosityThermalExpansion * (temperature - temperature_n) // change due to temperature increment
               + fixedStressPressureCoefficient * (pressure - pressure_k) // fixed-stress pressure term
               + fixedStressTemperatureCoefficient * (temperature - temperature_k); // fixed-stress temperature term
```
the last two terms are different from what @francoishamon wrote above. Both options are good, I just slightly prefer Option 2 for historical reasons.

francoishamon: Yes, sounds good, it is what I implemented in the current version of this branch. I just fixed what is written in my previous message, sorry about the confusion :)

Guotong-Ren (approved):
I was confused whether the "meanStressIncrement" is "total" or "effective". As long as it is "effective", Option 2 is good for me.

francoishamon:
> I was confused whether the "meanStressIncrement" is "total" or "effective". As long as it is "effective", Option 2 is good for me.

I see. After this PR, it becomes `effective` and is called `meanEffectiveStressIncrement`.

src/coreComponents/constitutive/solid/porosity/BiotPorosity.hpp
frankfeifan: > @binw1 and @frankfeifan regarding this comment and [the following comment](https://github.com/GEOS-DEV/GEOS/pull/2554#discussion_r1253474355): the two options are doing the same thing unless I am missing something, we just have to decide which one we prefer:
> 
> Option 1: pass the mean total stress increment
> 
> ```
> porosity = porosity_n 
> + biotCoefficient * meanTotalStressIncrement / bulkModulus // change due to stress increment 
> + biotSkeletonModulusInverse * (pressure - pressure_n) // change due to pressure increment 
> - porosityThermalExpansion * (temperature - temperature_n) // change due to temperature increment 
> + fixedStressPressureCoefficient * (pressure - pressure_n) // fixed-stress pressure term 
> + fixedStressTemperatureCoefficient * (temperature - temperature_n); // fixed-stress temperature term
> ```
> 
> where
> 
> ```
> meanTotalStressIncrement = meanEffectiveStressIncrement 
> - biotCoefficient * (pressure_k - pressure_n ) 
> - 3 alpha ( temperature_k  - temperature_n ) 
> ```
> 
> Option 2: pass the mean effective stress increment
> 
> ```
> porosity = porosity_n 
> + biotCoefficient * meanEffectiveStressIncrement / bulkModulus // change due to stress increment 
> + biotSkeletonModulusInverse * (pressure - pressure_n) // change due to pressure increment 
> - porosityThermalExpansion * (temperature - temperature_n) // change due to temperature increment 
> + fixedStressPressureCoefficient * (pressure - pressure_k) // fixed-stress pressure term 
> + fixedStressTemperatureCoefficient * (temperature - temperature_k); // fixed-stress temperature term
> ```
> 
> where
> 
> ```
> meanEffectiveStressIncrement = bulkModulus * ( strainIncrement[0] + strainIncrement[1] + strainIncrement[2] )
> ```
> 
> and I agree with @frankfeifan that we can be more precise on the naming by using the terms `effective` or `total` to qualify the stress.
> 
> My preference goes to Option 2 because it makes it easier to understand the time levels of pressure/temperature terms (rather than have pressure and temperature buried in the stress).

Yeah. I also agree that Option 2 is clearer and better.

francoishamon:
> This all looks great to me, thanks Francois! Out of curiosity, did you do much testing to see if loosening the sequential Newton tolerances still gave matching answers to FIM? I will likely start testing these shortly as well.

Hi Ryan, not much testing at all for the sensitivity of the results to the tolerance. I just made sure that the sequential solver matches the analytical solution for two problems using a tight tolerance:

Single-phase thermo-poro-mechanics

temperature

Mandel

mandel_sequential

I pushed the input files I used to the repository. But this is all I did. So testing the solver at a looser tolerance as you suggest is a great idea, as well as making sure that the outer convergence is good.

paveltomin:
FYI @sohailwaziri

PR #2555, opened 2023-06-30
Github Actions - Check for all job completion

bmhan12:
This PR:
- Creates a job like in current Travis CI that passes when all other jobs have passed. Makes it easier to have GitHub actions as a required workflow for PR's.

### DIFF
--- a/.github/workflows/ci_tests.yml
+++ b/.github/workflows/ci_tests.yml
@@ -1,5 +1,5 @@
 name: GEOS CI 
-on: pull_request
+on: push
 
 # Cancels in-progress workflows for a PR when updated
 concurrency:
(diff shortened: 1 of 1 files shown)


rrsettgast (approved):

.github/workflows/ci_tests.yml
```diff
@@ -216,3 +216,11 @@ jobs:
...
         tar czf ${GEOSX_BUNDLE} --directory=${TMP_DIR} ${GEOSX_EXPORT_DIR}
         CLOUDSDK_PYTHON=python3 gsutil cp -a public-read ${GEOSX_BUNDLE} gs://${GCP_BUCKET}/
+
+  # Convenience job - passes when all other jobs have passed.
+  check_that_all_jobs_succeeded:
+    runs-on: ubuntu-22.04
+    needs: [check_pull_request_is_not_a_draft, check_pull_request_is_assigned, check_submodules, code_style, documentation, linux_builds]
+    steps:
+    - name: Success
+      run: "true"
```
TotoGaz, line 226: This is strange that we cannot call the worklow name by itself.

TotoGaz (approved):

.github/workflows/ci_tests.yml
bmhan12: It's not ideal... 

spack has a similar all-check job like this because you cannot add the whole workflow: https://github.com/spack/spack/blob/26a98f4c14598fb21423904799ae5cd01fa939d0/.github/workflows/ci.yaml#L81-L86

TotoGaz: so be it!

PR #2562, opened 2023-07-10
fix a few typos and `404`s

tbeltzun:
I guess the title says it all ...


src/coreComponents/LvArray
TotoGaz, line 1: What is this submodule update? (as well a the others)

TotoGaz (approved):

PR #2564, opened 2023-07-10
Added checks for displacement BCs and delayed well transmissibility calculations

### LINKED ISSUE #2290: Improve boundary condition error checking

francoishamon:
This PR makes three small changes to address issues found in recent simulations:
- Adds a warnings if displacement boundary conditions are missing (fixes #2290).
- Delays the computation of the well transmissibility to make sure that permeability has been properly updated by the flow solver (needed when the Karman-Kozeny model is used).
- Enables total rate control for single-phase producers

Ready for review, rebaseline done in https://github.com/GEOS-DEV/integratedTests/pull/21


src/coreComponents/physicsSolvers/multiphysics/CoupledReservoirAndWellsBase.hpp
```diff
@@ -191,12 +192,84 @@ class CoupledReservoirAndWellsBase : public CoupledSolver< RESERVOIR_SOLVER, WEL
...
+                     real64 const & dt,
+                     DomainPartition & domain ) override
+  {
+    Base::implicitStepSetup( time_n, dt, domain );
+
+    // we delay the computation of the transmissibility until the last minute
+    // because we want to make sure that the permeability has been updated (in the flow solver)
+    // this is necessary for some permeability models (like Karman-Kozeny) that do not use the imported permeability
+    // ultimately, we may want to use this mechanism to update the well transmissibility at each time step (if needed)
+    if( !m_isWellTransmissibilityComputed )
```
joshua-white, line 213: Is it ever possible for this condition to be silently skipped?  Do we need to check the flag value at some point to confirm true?

joshua-white (approved):
Looks great. Thanks for fixing this!

src/coreComponents/physicsSolvers/multiphysics/CoupledReservoirAndWellsBase.hpp
francoishamon: Yes, good idea, I added this check before assembly
https://github.com/GEOS-DEV/GEOS/blob/f43f5dd3211ceb468d5ce51e0d4cf57b350de002/src/coreComponents/physicsSolvers/multiphysics/CompositionalMultiphaseReservoirAndWells.cpp#L282-L285
to make sure that the computation is not silently ignored

Issue #1994, opened 2022-08-01
Semi-Automated explicit instantiation of templated functions in separate compilation units


rrsettgast:
**What is the requested feature?**
We have various functions that get called after going through a series of "pass thru" template, which are essentially a large manual `if/elseif/.../else` block for calling compile time instantiations of functions from a runtime interface. This is all well and good, except that it results in huge compilation units that compile all permutations of the pass-thru. This has two problems: 1) long compile times that are not able to use parallelism, 2) large memory requirements for these bloated compilation units.

example:
```c++
template< typename LAMBDA >
void constitutiveUpdatePassThru( MultiFluidBase const & fluid,
                                 LAMBDA && lambda )
{
  ConstitutivePassThruHandler< DeadOilFluid,
                               BlackOilFluid,
#ifdef GEOSX_USE_PVTPackage
                               CompositionalMultiphaseFluid,
#endif
                               CO2BrinePhillipsFluid,
                               CO2BrineEzrokhiFluid,
                               CO2BrinePhillipsThermalFluid /*, // if I uncomment the two models at the same time, the compiler segfaults on
                                                               Lassen!
                                                               CO2BrineEzrokhiThermalFluid*/>::execute( fluid, std::forward< LAMBDA >( lambda ) );
}
```
and
```c++
template< typename TYPE, typename ... TYPES >
struct ConstitutivePassThruHandler< TYPE, TYPES... >
{
  template< typename BASE, typename LAMBDA >
  static void execute( BASE & relation, LAMBDA && lambda )
  {
    using Derived = add_const_if_t< TYPE, std::is_const< BASE >::value >;

    if( dynamicCast< Derived * >( &relation ) )
    {
      lambda( static_cast< Derived & >( relation ) );
    }
    else
    {
      ConstitutivePassThruHandler< TYPES... >::execute( relation, std::forward< LAMBDA >( lambda ) );
    }
  }
};
```

Ideally, the `lambda` function will call another function that is templated on the strong type of `fluid` that is passed to `execute`, which is passed to `lambda` eventually. For instance, `PVTDriver.cpp has :
```c++
  constitutiveUpdatePassThru( baseFluid, [&] ( auto & selectedFluid )
  {
    using FLUID_TYPE = TYPEOFREF( selectedFluid );
    runTest< FLUID_TYPE >( selectedFluid, m_table );
  } );
```
We need to have cmake create a bunch of compilation units that look like:
```c++
[...]

PR #2565, opened 2023-07-10
[WIP] Configure-time specification of which kernels to instantiate for dispatching

wrtobin:
- [X] CMake functionality for parsing JSON kernel spec and generating source files from templates.
- [ ] Add kernel instantiation template source files


rrsettgast:
hi @johnbowen42 Sorry. I was supposed to do this a couple of weeks ago. @wrtobin had started the task we discussed, but is pretty overloaded. Perhaps you would be ok picking this up?


Here is an issue with some discussion on the matter:
https://github.com/GEOS-DEV/GEOS/issues/1994

wrtobin:
Just grabbed the template source files and default kernels json, which IIRC I was unhappy with the overall format of and was in the middle of modifying / reducing which is why it wasn't already committed. 
The kernel template instantiation source file generation should be bare-bones operable after that commit.

klevzoff:
@johnbowen42 I've just pushed my changes, the code is now in a state where it will compile, link and run (at least it does for me :) ). 

It supports reduced kernel specs for kernels under `solidMechanics/kernels` and `multiphysics/poromechanicsKernels`. For some of those under `fluidFlow`, it generates translation units for instantiations, but does not fully support reduced kernel specs because the dispatch on the caller side is not modified and expects the full set of instantiations to be linked. As we discussed, this maybe worthwhile addressing in a follow up PR.

PR #2568, opened 2023-07-11
Support wells in sequential poromechanics + 3 other PRs

paveltomin:
Make wells compatible with poromechanics, example for single-phase case:
image

Mechanical initialization would not work when mechanics-only solver is used because bulkDensity is 0, but works ok if coupled solver is applied (documented here https://github.com/GEOS-DEV/GEOS/issues/2698)

Rebase: https://github.com/GEOS-DEV/integratedTests/pull/59

Merged with https://github.com/GEOS-DEV/GEOS/pull/2766, https://github.com/GEOS-DEV/GEOS/pull/2687, and https://github.com/GEOS-DEV/GEOS/pull/2754 to speedup merging


inputFiles/poromechanics/PoroElastic_staircase_singlephase_3d_seq.xml
```diff
@@ -0,0 +1,344 @@
...
+  <Solvers>
+
+    <SinglePhaseReservoirPoromechanics
+      name="reservoirPoromechanics"
+      solidSolverName="linearElasticity"
+      reservoirAndWellsSolverName="reservoirSystem"
+      logLevel="1"
+      targetRegions="{ channel, barrier, wellRegion1, wellRegion2 }">
+      <NonlinearSolverParameters
+        couplingType="Sequential"
```
joshua-white, line 13: I know you are just testing right now, but we should eventually use the xml include functionality to have a `PoroElastic_staircase_singlephase_3d_base.xml` and then two concise downstream files, `_fim` and `_seq`.  We are only really changing two lines.

paveltomin:
Just checked - zero bulk density issue is not related to presence of wells, it happens in flow-mechanics sequential too.

paveltomin:
> Just checked - zero bulk density issue is not related to presence of wells, it happens in flow-mechanics sequential too.

documented the wrong behavior here https://github.com/GEOS-DEV/GEOS/issues/2698

inputFiles/poromechanics/PoroElastic_staircase_co2_3d_sequential.xml
```diff
@@ -0,0 +1,398 @@
+<?xml version="1.0" ?>
```
joshua-white, line 1: Is it possible to put some of these blocks in a base XML file to avoid redundancy?

inputFiles/poromechanics/PoroElastic_staircase_co2_3d_sequential.xml
```diff
@@ -0,0 +1,398 @@
...
+      name="reservoirSystem"
+      solidSolverName="linearElasticity"
+      reservoirAndWellsSolverName="reservoirAndWells"
+      logLevel="1"
+      targetRegions="{ channel, barrier, wellRegion1, wellRegion2 }">
+      <NonlinearSolverParameters
+        couplingType="Sequential"
+        subcycling="1"
+        newtonTol="1.0e-8"
+        lineSearchAction="None"
```
joshua-white, line 16: Is this disabled because it doesn't work or just not needed?

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoromechanics.cpp
```diff
@@ -37,59 +37,92 @@ namespace geos
...
+}
+
+template< typename FLOW_SOLVER >
+MultiphasePoromechanics< FLOW_SOLVER >::MultiphasePoromechanics( const string & name,
+                                                                 Group * const parent )
   : Base( name, parent ),
   m_isThermal( 0 )
 {
-  registerWrapper( viewKeyStruct::stabilizationTypeString(), &m_stabilizationType ).
+  this->registerWrapper( viewKeyStruct::stabilizationTypeString(), &m_stabilizationType ).
```
joshua-white, line 77: Are you inserting `this->` here as a stylistic preference, or is it strictly required here?  I don't know what our code style guidelines say, but we should be consistent.

joshua-white (commented):
This looks good to me.  May be worth a quick walkthrough at a dev meeting.

inputFiles/poromechanics/PoroElastic_staircase_co2_3d_sequential.xml
paveltomin: yes i will definitely do that before merging, thanks!

inputFiles/poromechanics/PoroElastic_staircase_co2_3d_sequential.xml
paveltomin: just not needed and also my personal preference about line search

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoromechanics.cpp
paveltomin: for me it just does not compile without this because of the template introduction
i believe i took that "this" way from some other solver code, see https://github.com/search?q=repo%3AGEOS-DEV%2FGEOS%20this-%3EregisterWrapper&type=code

joshua-white: ahh, right, definitely required then.

inputFiles/poromechanics/PoroElastic_staircase_co2_3d_sequential.xml
paveltomin: sorry, clarification here, coupled solver currently does not support line search https://github.com/GEOS-DEV/GEOS/blob/5b20fd397c152fadc4e7f12b6209580221295634/src/coreComponents/physicsSolvers/multiphysics/CoupledSolver.hpp#L565

inputFiles/poromechanics/PoroElastic_staircase_co2_3d_sequential.xml
paveltomin: done

inputFiles/poromechanics/PoroElastic_staircase_singlephase_3d_seq.xml
paveltomin: done

paveltomin:
> This looks good to me. May be worth a quick walkthrough at a dev meeting.

I think I addressed all your comments, can you have another look and approve if everything looks ok?

joshua-white (approved):
Looks good.  I think a few more duplicated xml blocks can be moved to the base file, but I could be wrong.  Regardless, I think this is ready to merge.

CusiniM (approved):
I find it quite hard to remember which solve is which since the names are similar but putting reservoir before or after changes the coupling.  However, I guess that we don't really have a choice atm. It would be good if we could add a check in case someone specified `couplingType="Sequential" ` for a solver `SinglePhasePoromechanicsReservoir`, which, if I am not mistaking, only supports FIM coupling.

paveltomin:
> I find it quite hard to remember which solve is which since the names are similar but putting reservoir before or after changes the coupling. However, I guess that we don't really have a choice atm. It would be good if we could add a check in case someone specified `couplingType="Sequential" ` for a solver `SinglePhasePoromechanicsReservoir`, which, if I am not mistaking, only supports FIM coupling.

I agree it is very confusing. This PR is just to enable the functionality and we can address the details later.
SinglePhasePoromechanicsReservoir does work in sequential way, it just does what is not very expected from user point of view because it is solving reservoir-flow+mechanics followed by solving wells and outer loop on top of that. However it is completely clear behavior from the coupled solver logic.
I suggest we merge it, start using and then straighten the input/code organization.

PR #2569, opened 2023-07-12
Added a function to VTKUtilities to avoid empty ranks after redistributeByKdTree

francoishamon:
This PR tries to overcome a blocker that @zhishuaizhang and @jafranc have faced when loading 2D meshes with `VTKMesh`. They faced a pathological case in which some ranks (sometimes, many ranks) have zero elements after the call to `redistributeByKdTree`, which breaks the code.

This PR adds a function to detect this case and make sure that the simulation can proceed.
Ranks are divided by M `donors` (ranks with 2 elements or more) and N `recipients` (ranks without any elements).
- If M > N, the N donors with the largest number of elements donate half of their elements to the recipients.
- Otherwise, I tried a brute force approach in which the donors send one element to each recipient. Unfortunately, this case seems to happen frequently for large number of ranks in the reproducer sent by @zhishuaizhang. When this approach is used there is now a message telling the user to increase `partitionRefinement` to make sure that the partitioning is balanced after the call to ParMetis.

### DIFF
--- a/src/coreComponents/mesh/generators/VTKUtilities.cpp
+++ b/src/coreComponents/mesh/generators/VTKUtilities.cpp
@@ -640,6 +640,115 @@ vtkSmartPointer< vtkDataSet > manageGlobalIds( vtkSmartPointer< vtkDataSet > mes
   return output;
 }
 
+/**
+ * @brief This function tries to make sure that no MPI rank is empty
+ *
+ * @param[in] mesh a vtk grid
+ * @param[in] comm the MPI communicator
+ * @return the vtk grid redistributed
+ */
+vtkSmartPointer< vtkDataSet >
+ensureNoEmptyRank( vtkDataSet & mesh,
+                   MPI_Comm const comm )
+{
+  // step 1: figure out who is a donor and who is a recipient
+  globalIndex const numElems = mesh.GetNumberOfCells();
+  globalIndex const numProcs = MpiWrapper::commSize( comm );
+
+  array1d< globalIndex > elemCounts( numProcs );
+  MpiWrapper::allGather( numElems, elemCounts, comm );
+
+  SortedArray< globalIndex > recipientRanks;
+  array1d< globalIndex > donorRanks;
+  array1d< globalIndex > donorNumElems;
+  recipientRanks.reserve( numProcs );
+  donorRanks.reserve( numProcs );
+  donorNumElems.reserve( numProcs );
+
+  for( globalIndex iRank = 0; iRank < numProcs; ++iRank )
+  {
+    if( elemCounts[iRank] == 0 )
+    {
+      recipientRanks.insert( iRank );
+    }
+    else if( elemCounts[iRank] > 1 ) // need at least two elems to be a donor
+    {
+      donorRanks.emplace_back( iRank );
+      donorNumElems.emplace_back( elemCounts[iRank] );
+    }
+  }
+
+  // step 2: at this point, we need to determine the donors and which cells they donate
+
+  // First we sort the donor in order of the number of elems they contain
+  std::stable_sort( donorRanks.begin(), donorRanks.end(),
+                    [&donorNumElems] ( auto & i1, auto & i2 )
+  { return donorNumElems[i1] < donorNumElems[i2]; } );
+
+  // Then, if my position is "i" in the donorRanks array, I will send half of my elems to the i-th recipient
+  globalIndex const myRank = MpiWrapper::commRank();
+  auto const myPosition =
+    LvArray::sortedArrayManipulation::find( donorRanks.begin(), donorRanks.size(), myRank );
+  bool const isDonor = myPosition != donorRanks.size();
+
+  // step 3: my rank was selected to donate cells, let's proceed
+  // we need to make a distinction between two configurations
+
+  array1d< globalIndex > newParts( numElems );
+  newParts.setValues< parallelHostPolicy >( myRank );
+
+  // step 3.1: donorRanks.size() >= recipientRanks.size()
+  // we use a strategy that preserves load balancing
+  if( isDonor && donorRanks.size() >= recipientRanks.size() )
+  {
+    if( myPosition < recipientRanks.size() )
+    {
+      globalIndex const recipientRank = recipientRanks[myPosition];
+      for( globalIndex iElem = std::floor( numElems/2.0 ); iElem < numElems; ++iElem )
+      {
+        newParts[iElem] = recipientRank; // I donate half of my cells
+      }
+    }
+  }
+  // step 3.2: donorRanks.size() < recipientRanks.size()
+  // this is the unhappy path: we don't care anymore about load balancing at this stage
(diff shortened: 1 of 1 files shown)


src/coreComponents/mesh/generators/VTKUtilities.cpp
```diff
@@ -640,6 +640,115 @@ vtkSmartPointer< vtkDataSet > manageGlobalIds( vtkSmartPointer< vtkDataSet > mes
...
+      donorNumElems.emplace_back( elemCounts[iRank] );
+    }
+  }
+
+  // step 2: at this point, we need to determine the donors and which cells they donate
+
+  // First we sort the donor in order of the number of elems they contain
+  std::stable_sort( donorRanks.begin(), donorRanks.end(),
+                    [&donorNumElems] ( auto & i1, auto & i2 )
+  { return donorNumElems[i1] < donorNumElems[i2]; } );
```
klevzoff, line 686: Are you sure this shouldn't just be
```suggestion
  std::stable_sort( donorRanks.begin(), donorRanks.end(),
                    [&elemCounts] ( auto i1, auto i2 )
  { return elemCounts[i1] < elemCounts[i2]; } );
```
since `i1` and `i2` here are both in the range `[0..numProcs)`, not `[0..numDonors)`? If I read this correctly, you won't need `donorNumElems` at all.

src/coreComponents/mesh/generators/VTKUtilities.cpp
```diff
@@ -640,6 +640,115 @@ vtkSmartPointer< vtkDataSet > manageGlobalIds( vtkSmartPointer< vtkDataSet > mes
...
+  newParts.setValues< parallelHostPolicy >( myRank );
+
+  // step 3.1: donorRanks.size() >= recipientRanks.size()
+  // we use a strategy that preserves load balancing
+  if( isDonor && donorRanks.size() >= recipientRanks.size() )
+  {
+    if( myPosition < recipientRanks.size() )
+    {
+      globalIndex const recipientRank = recipientRanks[myPosition];
+      for( globalIndex iElem = std::floor( numElems/2.0 ); iElem < numElems; ++iElem )
```
klevzoff, line 707: I think this is equivalent (rounds down for odd numbers, if that was the intent)?
```suggestion
      for( localIndex iElem = numElems/2; iElem < numElems; ++iElem )
```

src/coreComponents/mesh/generators/VTKUtilities.cpp
```diff
@@ -640,6 +640,115 @@ vtkSmartPointer< vtkDataSet > manageGlobalIds( vtkSmartPointer< vtkDataSet > mes
...
+ * @param[in] comm the MPI communicator
+ * @return the vtk grid redistributed
+ */
+vtkSmartPointer< vtkDataSet >
+ensureNoEmptyRank( vtkDataSet & mesh,
+                   MPI_Comm const comm )
+{
+  // step 1: figure out who is a donor and who is a recipient
+  globalIndex const numElems = mesh.GetNumberOfCells();
+  globalIndex const numProcs = MpiWrapper::commSize( comm );
```
klevzoff, line 656: I think we prefer to use `int` or `integer` to represent rank/partition numbers (they don't need to be 64-bit), here and below. Also `numElems` and related quantities should probably be `localIndex`, since we're describing local mesh parts.
```suggestion
  localIndex const numElems = LvArray::integerConverstion< localIndex >( mesh.GetNumberOfCells() );
  int const numProcs = MpiWrapper::commSize( comm );
```

src/coreComponents/mesh/generators/VTKUtilities.cpp
```diff
@@ -640,6 +640,115 @@ vtkSmartPointer< vtkDataSet > manageGlobalIds( vtkSmartPointer< vtkDataSet > mes
...
+ *
+ * @param[in] mesh a vtk grid
+ * @param[in] comm the MPI communicator
+ * @return the vtk grid redistributed
+ */
+vtkSmartPointer< vtkDataSet >
+ensureNoEmptyRank( vtkDataSet & mesh,
+                   MPI_Comm const comm )
+{
+  // step 1: figure out who is a donor and who is a recipient
```
klevzoff, line 654: I would maybe propose a caliper mark here so may later get an idea of how much this redistribution step contributes to the overall VTK mesh generation time
```suggestion
{
  GEOS_MARK_FUNCTION;

  // step 1: figure out who is a donor and who is a recipient
```

src/coreComponents/mesh/generators/VTKUtilities.cpp
```diff
@@ -640,6 +640,115 @@ vtkSmartPointer< vtkDataSet > manageGlobalIds( vtkSmartPointer< vtkDataSet > mes
...
+    globalIndex firstRecipientPosition = 0;
+    for( globalIndex iRank = 0; iRank < myPosition; ++iRank )
+    {
+      firstRecipientPosition += elemCounts[iRank] - 1;
+    }
+    if( firstRecipientPosition < recipientRanks.size() )
+    {
+      bool const isLastDonor = myPosition == donorRanks.size() - 1;
+      globalIndex const lastRecipientPosition = firstRecipientPosition + numElems - 2;
+      GEOS_THROW_IF( isLastDonor && ( lastRecipientPosition < recipientRanks.size() - 1 ),
```
klevzoff, line 727: Sorry for the super minor nit, this might be just a tad easier to read if you treat `lastRecipientPosition` as a typical C++/STL-style one-past-the-end iterator
```suggestion
      globalIndex const lastRecipientPosition = firstRecipientPosition + numElems - 1;
      GEOS_THROW_IF( isLastDonor && ( lastRecipientPosition < recipientRanks.size() ),
```

klevzoff (approved):
Great work! Just some minor comments

src/coreComponents/mesh/generators/VTKUtilities.cpp
francoishamon: Thanks for the catch, I fixed it and removed `donorNumElems`

src/coreComponents/mesh/generators/VTKUtilities.cpp
francoishamon: Done

src/coreComponents/mesh/generators/VTKUtilities.cpp
francoishamon: Done, I updated the types in the new function

src/coreComponents/mesh/generators/VTKUtilities.cpp
francoishamon: Done

src/coreComponents/mesh/generators/VTKUtilities.cpp
francoishamon: Done

francoishamon: Done

PR #2571, opened 2023-07-12
Remove travis

TotoGaz:
- Removes travis.
- Removes the epic GHA.
- Fix the `geos` installation folder so it uses the hash.


bmhan12 (approved):
Should also rename the `travis_build_and_test.sh` script, or maybe combine it with the `ci_build_and_test.sh` script?

The scripts:
https://github.com/GEOS-DEV/GEOS/blob/991327d913654ce589c80b17c58ea6d8146cc7b6/scripts/ci_build_and_test.sh#L35

https://github.com/GEOS-DEV/GEOS/blob/develop/scripts/travis_build_and_test.sh

TotoGaz:
> Should also rename the `travis_build_and_test.sh` script, or maybe combine it with the `ci_build_and_test.sh` script?

I've quickly looked at it. I should definitely rename it, but I'm not sure the two scripts can be merged.

bmhan12:
> I've quickly looked at it. I should definitely rename it, but I'm not sure the two scripts can be merged.
Renaming works for me.

bmhan12 (approved):

TotoGaz:
@jhuang2601 was able to validate that we the auto deployments are working.
@rrsettgast I think we can have this PR merged: we're running travis no more. Are you OK?

jhuang2601 (approved):
Great job!

PR #2577, opened 2023-07-14
Logger compilation error

herve-gross:
Remove `template` from the two function calls to a logger (causes issues compiling on my system). Not sure how this passed CI, though.


TotoGaz (approved):

TotoGaz:
> Not sure how this passed CI, though.

Most likely your compiler is stricter!

wrtobin:
Whoops didn't realize this was already open. I merged a fix for this 20 minutes ago or so.

PR #2584, opened 2023-07-17
Modified parallel plates permeability

karimifard:
This PR modifies the parallel plates permeability to be consistent with other models. The permeability is now (aperture^2)/12 instead of (aperture^3)/12. The discretization for fractures is modified accordingly to account for this change.


src/coreComponents/finiteVolume/StencilBase.hpp
```diff
@@ -82,16 +85,19 @@ class StencilWrapperBase : public TRAITS
...
   StencilWrapperBase( typename TRAITS::IndexContainerType const & elementRegionIndices,
                       typename TRAITS::IndexContainerType const & elementSubRegionIndices,
                       typename TRAITS::IndexContainerType const & elementIndices,
+                      typename TRAITS::WeightContainerType const & partialWeights,
                       typename TRAITS::WeightContainerType const & weights ):
     m_elementRegionIndices( elementRegionIndices.toViewConst() ),
     m_elementSubRegionIndices( elementSubRegionIndices.toViewConst() ),
     m_elementIndices( elementIndices.toViewConst() ),
-    m_weights( weights.toViewConst() )
+    m_partialWeights( partialWeights.toView() ),
```
CusiniM, line 99: are we going to need `partialWeights` also for the `CellElementTPFAStencil`? If not, I think that we could define it directly only in the stencils involving fractures.

src/coreComponents/constitutive/permeability/ParallelPlatesPermeability.hpp
```diff
@@ -44,19 +44,10 @@ class ParallelPlatesPermeabilityUpdate : public PermeabilityBaseUpdate
...
-                                 oldHydraulicAperture*newHydraulicAperture*newHydraulicAperture +
-                                 newHydraulicAperture*newHydraulicAperture*newHydraulicAperture ) / 12;
-
-    real64 const dPerm  = 0.25 * ( oldHydraulicAperture*oldHydraulicAperture +
-                                   2*oldHydraulicAperture*newHydraulicAperture +
-                                   3*newHydraulicAperture*newHydraulicAperture ) / 12;
+    GEOS_UNUSED_VAR( oldHydraulicAperture );
 
+    real64 const perm  = newHydraulicAperture*newHydraulicAperture / 12.0;
+    real64 const dPerm = newHydraulicAperture / 6.0;
```
CusiniM, line 50: we should make sure that the convergence of the hydraulic fracturing benchmarks does not get worse by doing this. If it does we may need a switch.

src/coreComponents/finiteVolume/StencilBase.hpp
karimifard: I may be wrong but I think they all use the same interface. And also we thought we may keep it this way in case it's useful for large deformation. But I am open to limit it to fracture stencils if it doesn't require too many changes.

src/coreComponents/constitutive/permeability/ParallelPlatesPermeability.hpp
karimifard: Yes, we will see how it behaves.

TotoGaz (commented):
LGTM

PR #2588, opened 2023-07-18
Add reordering steps in Box definition

jafranc:
Add a comparison/reordering step in ``Box`` definition so that the input can specify any of the 2 possible diagonals.


src/coreComponents/mesh/simpleGeometricObjects/Box.cpp
```diff
@@ -65,6 +65,13 @@ void Box::postProcessInput()
   LvArray::tensorOps::add< 3 >( m_boxCenter, m_max );
   LvArray::tensorOps::scale< 3 >( m_boxCenter, 0.5 );
 
+  // reordering min and max to fit former interface but so that user can input any of the two diagonals
```
karimifard, line 68: // ... any of four diagonals?

Maybe add a GEOS message to tell the user about the swap.

francoishamon (approved):

karimifard (approved):

src/coreComponents/mesh/simpleGeometricObjects/Box.cpp
```diff
@@ -65,6 +65,16 @@ void Box::postProcessInput()
   LvArray::tensorOps::add< 3 >( m_boxCenter, m_max );
   LvArray::tensorOps::scale< 3 >( m_boxCenter, 0.5 );
 
+  // reordering min and max to fit former interface but so that user can input any of the four diagonals
+  for( int i = 0; i < m_max.SIZE; ++i )
+  {
+    if( m_max[i]<m_min[i] )
+    {
+      std::swap( m_max[i], m_min[i] );
+      GEOS_WARNING( GEOS_FMT( "Reordering box definition for {} component as {} < {} ", i, m_max[i], m_min[i] ));
```
francoishamon, line 74: Sorry @jafranc I missed the warning at the first review. If I remember well `GEOS_WARNING` tends to be overwhelming for many-rank simulations (there is too much output to the log because each rank issues the warning). 

Maybe we could use: `GEOS_LOG_RANK_0` here instead?

Issue #2595, opened 2023-07-21
Update tutorial example for CO2 injection


jhuang2601:
Duo to recent change, following tutorial example needs to be updated:

- https://geosx-geosx.readthedocs-hosted.com/en/latest/docs/sphinx/basicExamples/co2Injection/Example.html

Here is the error message for running the existing case
```
***** LOCATION: /shared/scratch1/Users/j0551570/geosx/IntegratedTest/GEOS/src/coreComponents/mesh/WellElementRegion.cpp:66
***** Controlling expression (should be false): matchedPerforations != numPerforationsGlobal
***** Rank 0: Invalid mapping perforation-to-element in well wellRegion. This happens when GEOSX cannot match a perforation with a reservoir element. There are two common reasons for this error:
 1- The most common reason for this error is that a perforation is on a section of  the well polyline located outside the domain.
 2- This error can also happen if a perforation falls on a mesh face or a mesh vertex. Please try to move the perforation slightly (to the interior of the perforated cell) to see if it fixes the problem. 
```

PR #2596, opened 2023-07-21
Update tutorial example for CO2 injection

jhuang2601:
This small PR fixes the [reported issue ](https://github.com/GEOS-DEV/GEOS/issues/2595).

Somehow, the smoke test for this tutorial example is not included in the integratedTests and hence this unexpected issue.


herve-gross (approved):
Thank you for fixing this.

francoishamon (approved):

PR #2597, opened 2023-07-24
use specific implementation ti get available host memory on MacOS

XL64:
This fixes a compilation error on MacOS where the method to obtain the current memory available on host doesn't not build because _SC_AVPHYS_PAGES is not available.

### DIFF
--- a/src/coreComponents/common/getAvailableMemory.hpp
+++ b/src/coreComponents/common/getAvailableMemory.hpp
@@ -1,0 +1,56 @@
+#ifndef GET_AVAILABLE_MEMORY_HPP
+#define GET_AVAILABLE_MEMORY_HPP
+
+#include <sys/stat.h>
+#if defined __MACH__
+#include <sys/types.h>
+#include <sys/sysctl.h>
+
+#include <mach/host_info.h>
+#include <mach/mach_host.h>
+#include <mach/task_info.h>
+#include <mach/task.h>
+#endif
+
+namespace geos
+{
+  /**
+   * @brief Retieves current available memory on host
+   * @return the available memory in bytes.
+   */
+  static inline size_t getAvailableMemory() 
+  {
+#if defined(__APPLE__) && defined(__MACH__)
+    int mib[6]; 
+    mib[0] = CTL_HW;
+    mib[1] = HW_PAGESIZE;
+
+    int pagesize;
+    size_t length;
+    length = sizeof( pagesize );
+    if ( sysctl( mib, 2, &pagesize, &length, NULL, 0 ) < 0 )
+    {
+      fprintf( stderr, "getting page size" );
+    }
+
+    mach_msg_type_number_t count = HOST_VM_INFO_COUNT;
+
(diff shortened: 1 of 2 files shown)


herve-gross (approved):
The code now compiles without issues on my Mac. Thanks for fixing this!

src/coreComponents/common/getAvailableMemory.hpp
```diff
@@ -0,0 +1,56 @@
...
+
+  vm_statistics_data_t vmstat;
+  if( host_statistics( mach_host_self(), HOST_VM_INFO, ( host_info_t ) &vmstat, &count ) != KERN_SUCCESS )
+  {
+    fprintf ( stderr, "Failed to get VM statistics." );
+  }
+
+  return vmstat.free_count * pagesize;
+#else
+  GEOS_ERROR_IF( percent > 100, "Error, percentage of memory should be smallerer than -100, check lifoOnHost (should be greater that -100)" );
```
klevzoff, line 46: Copied here by accident?

src/coreComponents/common/getAvailableMemory.hpp
```diff
@@ -0,0 +1,56 @@
...
+  return (size_t)sysconf( _SC_AVPHYS_PAGES ) *(size_t) sysconf( _SC_PAGESIZE );
+#endif
+}
+}
+
+// remove these definitions from mach/boolean.h that can conflict with GEOS code (eg. InputFlags::FALSE)
+#undef TRUE
+#undef FALSE
```
klevzoff, line 54: Could you instead encapsulate this implementation and all system-specific includes in a separate translation unit (`.cpp`)? I think `LvArray/system.[h/c]pp` would be a good place to expose this functionality, if you can spare the effort to open an LvArray PR.

src/coreComponents/common/getAvailableMemory.hpp
```diff
@@ -0,0 +1,56 @@
...
+#include <mach/task.h>
+#endif
+
+namespace geos
+{
+/**
+ * @brief Retieves current available memory on host
+ * @return the available memory in bytes.
+ */
+static inline size_t getAvailableMemory()
```
klevzoff, line 21: Maybe indicate in the function name that this is for host/system memory?

src/coreComponents/common/getAvailableMemory.hpp
XL64: Indeed.

src/coreComponents/common/getAvailableMemory.hpp
XL64: Good idea.

src/coreComponents/common/getAvailableMemory.hpp
XL64: https://github.com/GEOS-DEV/LvArray/pull/299

paveltomin:
@herve-gross still needed or close?

herve-gross:
This was fixed. We can close it.

PR #2598, opened 2023-07-24
Github Actions - Separate CUDA builds

bmhan12:
This PR:
- Separates cuda build jobs from the other linux jobs.


CusiniM:
I wonder if it wouldn't make sense to run at least 1 cuda build as part of the first step. This way, if that one fails, we don't waste resources running all the other cuda builds.

rrsettgast (approved):

rrsettgast:
> I wonder if it wouldn't make sense to run at least 1 cuda build as part of the first step. This way, if that one fails, we don't waste resources running all the other cuda builds.

I think we can think of this as a phase1/phase2 where we can move builds between the phases? So at any point we can move something from phase2 to phase1 to make it builds every time.

.github/workflows/ci_tests.yml
```diff
@@ -221,10 +188,111 @@ jobs:
...
+        docker cp -a ${CONTAINER_NAME}:${GEOSX_TPL_DIR}/.. ${TMP_DIR}/${GEOSX_EXPORT_DIR}
+        GEOSX_BUNDLE=${TMP_DIR}/${GEOSX_EXPORT_DIR}.tar.gz
+        tar czf ${GEOSX_BUNDLE} --directory=${TMP_DIR} ${GEOSX_EXPORT_DIR}
+        CLOUDSDK_PYTHON=python3 gsutil cp -a public-read ${GEOSX_BUNDLE} gs://${GCP_BUCKET}/
+
+  large_cuda_builds:
+    name: ${{ matrix.name }}
+#    runs-on: ubuntu-22.04
+    runs-on: ${{ matrix.os }}
+    needs: [check_pull_request_is_not_a_draft, linux_builds]
```
TotoGaz, line 205:
```suggestion
    needs: [linux_builds]
```

TotoGaz (approved):

PR #2599, opened 2023-07-24
Well skin factor

kimtaeho07:
Implemented skin factor as possible input to perforation - minor edits to detailed instructions from Francois.

integratedTests PR: https://github.com/GEOS-DEV/integratedTests/pull/61


src/.vscode/tasks.json
```diff
@@ -0,0 +1,28 @@
+{
```
CusiniM, line 1: I don't think this file should be in the repository.

src/coreComponents/mesh/PerforationFields.hpp
```diff
@@ -73,6 +73,14 @@ DECLARE_FIELD( wellTransmissibility,
                WRITE_AND_READ,
                "For each perforation, well transmissibility" );
 
+DECLARE_FIELD( wellSkin,
+               "wellSkin",
```
CusiniM, line 77: here and in all functions names, can we use `skinFactor` instead of `skin`

src/coreComponents/mesh/PerforationData.hpp
```diff
@@ -177,6 +177,21 @@ class PerforationData : public ObjectManagerBase
...
+   * @return list of perforation well skin factors
+   */
+  arrayView1d< real64 const > getWellSkin() const { return m_wellSkin; }
+
+
+  /**
+   * @brief Get perforation well skin factors.
+   * @return list of perforation well skin factors
+   */
+  arrayView1d< real64 > getWellSkin() { return m_wellSkin; }
```
CusiniM, line 192: is this getter needed?

CusiniM (commented):
I have a few comments/questions:
-  I would prefer using `skinFactor` instead of `skin`. 
- I don't fully understand why we need to define the `skinFactor` both in the wellGenerator and on the perforations. Which of the two is exposed to the user? It seems to me that the field that it is actually used in the calculations is the one in `PerforationData`.

src/.vscode/tasks.json
kimtaeho07: Yes, will delete this.

src/coreComponents/mesh/PerforationFields.hpp
kimtaeho07: Sounds good, will change accordingly.

src/coreComponents/mesh/PerforationData.hpp
kimtaeho07: The dataflow that I got from doing this was:

1. XML input is passed into `Perforation`
2. `Perforation` passes to `InternalWellGenerator`
3. `InternalWellGenerator` passes to `LineBlock`
4. `LineBlock` passes to `WellElementSubRegion`
5. `WellElementSubRegion` passes to `PerforationData`

The `PerforationData::getWellSkin()` is needed as part of step 5. Hard setting the transmissibility at the perforation follows the same data flow.

CusiniM (approved):

CusiniM:
@francoishamon does this look good to you? can we get it ready to be merged?

francoishamon (approved):
Looks very good, thanks @kimtaeho07

paveltomin:
@kimtaeho07 clean up and move into merge queue?

PR #2605, opened 2023-07-27
Initialization of coupled solver is added

sohailwaziri:
The modifications are added to initialize the coupled solver. Without initialization, flow residual is nonzero during the flow solver initialization in a sequential setting.


paveltomin (approved):

ryar9534:
I am confused why this is necessary. It looks to me like this should be doing the same thing as what is already here: https://github.com/GEOS-DEV/GEOS/blob/a736739324b230d7bc34ff94ddcd4fefe0aff961/src/coreComponents/physicsSolvers/multiphysics/CoupledSolver.hpp#L374C26-L374C26

I feel like I must be missing something.

paveltomin:
> I am confused why this is necessary. It looks to me like this should be doing the same thing as what is already here: https://github.com/GEOS-DEV/GEOS/blob/a736739324b230d7bc34ff94ddcd4fefe0aff961/src/coreComponents/physicsSolvers/multiphysics/CoupledSolver.hpp#L374C26-L374C26
> 
> I feel like I must be missing something.

`implicitStepSetup` is indeed called for each subsolver, however, implicitStepSetup is missing and for example the function below is not called for sequential:
```
void SinglePhasePoromechanics::implicitStepSetup( real64 const & time_n,
                                                  real64 const & dt,
                                                  DomainPartition & domain )
{
  flowSolver()->keepFlowVariablesConstantDuringInitStep( m_performStressInitialization );
  Base::implicitStepSetup( time_n, dt, domain );
}
```
The critical thing in that function is `flowSolver()->keepFlowVariablesConstantDuringInitStep( m_performStressInitialization )` that enables correct initialization behavior.

ryar9534:
> > I am confused why this is necessary. It looks to me like this should be doing the same thing as what is already here: https://github.com/GEOS-DEV/GEOS/blob/a736739324b230d7bc34ff94ddcd4fefe0aff961/src/coreComponents/physicsSolvers/multiphysics/CoupledSolver.hpp#L374C26-L374C26
> > I feel like I must be missing something.
> 
> `implicitStepSetup` is indeed called for each subsolver, however, implicitStepSetup is missing and for example the function below is not called for sequential:
> 
> ```
> void SinglePhasePoromechanics::implicitStepSetup( real64 const & time_n,
>                                                   real64 const & dt,
>                                                   DomainPartition & domain )
> {
>   flowSolver()->keepFlowVariablesConstantDuringInitStep( m_performStressInitialization );
>   Base::implicitStepSetup( time_n, dt, domain );
> }
> ```
> 
> The critical thing in that function is `flowSolver()->keepFlowVariablesConstantDuringInitStep( m_performStressInitialization )` that enables correct initialization behavior.

Ah ok, I was missing that the derived classes of CoupledSolver have updated implicitStepSetups, thanks!

ryar9534 (approved):

paveltomin:
@castelletto1, @CusiniM, @francoishamon - could you please have a look and help us to merge this PR?

Guotong-Ren (approved):

jhuang2601:
I've tested this PR with SPE10 case and with this fix, case with sequential coupling scheme can converge to the corresponding full coupling one.

image

jhuang2601 (approved):

CusiniM (approved):

Issue #1834, opened 2022-03-20
Restart file invalid after a group is removed


klevzoff:
**Describe the bug**
If one or more groups are removed via `deregisterGroup`, the restart data written for some of the remaining sibling groups becomes garbage.

In practice, this can occurs after a call to `removeUnusedNeighbors()` here:
https://github.com/GEOSX/GEOSX/blob/aa0d5aea5a0d38607094b478aed3d6060e8005d6/src/coreComponents/mesh/mpiCommunications/CommunicationTools.cpp#L770

If any ranks' `NeighborData` is actually removed by this call (which can happen, in particular, when using `VTKMeshGenerator` which produces a slightly generous initial neighbor list), the simulation will proceed fine, but attempting to restart from any checkpoint will cause an exception similar to the following to be thrown:
```
{
  "file": "/home/geosx/thirdPartyLibs/build-gcc8-release/conduit/src/conduit/src/libs/conduit/conduit_node.cpp",
  "line": 13181,
  "message": "Node::as_signed_int() const -- DataType empty at path Problem/domain/MeshBodies/mesh1/meshLevels/Level0/nodeManager/neighborData/1/matchedPartitionBoundaryObjects/__sizedFromParent__ does not equal expected DataType int32"
}
```

**To Reproduce**
A minimal reproducer unit test with more detailed explanation is provided in https://github.com/GEOSX/GEOSX/pull/1833

Another way to trigger is to run (from `build-<config>-release/bin` directory):
```
mpirun -n 5 ./geosx -i ../../inputFiles/compositionalMultiphaseWell/staircase_co2_wells_hybrid_3d.xml
# let it finish
mpirun -n 5 ./geosx -i ../../inputFiles/compositionalMultiphaseWell/staircase_co2_wells_hybrid_3d.xml -r staircase_co2_wells_hybrid_3d_restart_000000012
```
(or any other restart produced by first command)

**Expected behavior**
Restart should be written correctly under any modifications of the group tree.

**Additional information**
  * Going forward, the bug will affect many parallel simulations using VTK mesh import. I wasn't able to get to the bottom of it, but it's rather important that we fix it.
  * For now, restart checking has been temporarily disabled for `staircase_co2_wells_hybrid_3d` test here: https://github.com/GEOSX/integratedTests/pull/206. We should re-enable it after the bug is fixed.
[...]

PR #2608, opened 2023-07-31
Fix a bug in MappedVector::erase

klevzoff:
Resolves https://github.com/GEOS-DEV/GEOS/issues/1834


francoishamon (approved):
Works great, thanks!

PR #2610, opened 2023-08-03
feat: Add freeOnDevice calls on Wave Propagation Solvers

acitrain:
This PR add the call to free some array from GPU memory using freeOnDevice method added in the following LvArray PR : 
https://github.com/GEOS-DEV/LvArray/pull/298


paveltomin:
@acitrain is it still active?

acitrain:
> @acitrain is it still active?

yes it should.. @jiemeng-total are you going to use this pull request for the free on device etc stuf?

wrtobin (approved):
can you remove the commented-out line? if its not needed its not needed.

@wrtobin Sure, thanks for pointing it out!

sframba (approved):

PR #2612, opened 2023-08-04
Adding strain integration option for DAS measurement in the elastic wave propagator

sframba:
This PR adds another way to compute a Distributed Acoustic Sensing (DAS) signal in the Elastic Wave Propagator SEM kernel. Specifically, the following formula is implemented:
$\mathrm{DAS}(t)=\sum_{i,j=1}^3F_i F_j \frac{\partial u_i}{\partial x_j}$,
where $F_i$ is the $i$-th component of the fiber direction (unit vector) and $u_i$ is the $i$-th component of the displacement field. This formulation is mathematically equivalent to the current one, but it is more accurate for shorter gauge length.

The DAS option in the `ElasticWaveEquationSEM` solver is modified as follows: the `useDAS` option now allows to select the displacement difference method (the old method, when `useDAS` is set equal to `"2"`) or the strain integration method, when `useDas="1"`. Setting `useDAS="0"` disables DAS computation, as per current behavior. 

Furthermore, the computation of the average strain vector along the gauge length is done via sampling some equispaced points along the fiber. The number of points can be set via the `linearDASSamples` field. The default value is `5`.


acitrain (approved):

sframba:
Rebaseline PR: [PR 89](https://github.com/GEOS-DEV/integratedTests/pull/89)

CusiniM (commented):
Wouldn't it be better to use an `enum` instead of an `integer` for the input option so that it is more user-friendly? Like it is done here https://github.com/GEOS-DEV/GEOS/blob/cb4e99b3ca3e86fb08017b291e87d054c3d8006d/src/coreComponents/linearAlgebra/utilities/LinearSolverParameters.hpp#L38 or LSParameters...

sframba:
> Wouldn't it be better to use an `enum` instead of an `integer` for the input option so that it is more user-friendly? Like it is done here
> 
> https://github.com/GEOS-DEV/GEOS/blob/cb4e99b3ca3e86fb08017b291e87d054c3d8006d/src/coreComponents/linearAlgebra/utilities/LinearSolverParameters.hpp#L38
> 
> or LSParameters...

Done

PR #2629, opened 2023-08-18
Add issue to dashboard using GHA

TotoGaz:


.github/workflows/add_issue_to_dashboard.yml
```diff
@@ -0,0 +1,16 @@
...
+
+jobs:
+  add-to-project:
+    name: Add issue to project
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/add-to-project@v0.5.0
+        with:
+          project-url: https://github.com/orgs/GEOS-DEV/projects/3
+          github-token: ${{ secrets.ADD_TO_PROJECT_PAT }}
```
TotoGaz, line 16: @bmhan12 Do you know which authentication token I should be putting here?

bmhan12: > @bmhan12 Do you know which authentication token I should be putting here?

Going over the add-to-project documentation ([link](https://github.com/actions/add-to-project#inputs)), looks to be a GEOS/GEOS_DEV(?) Personal Access Token with specific permissions? As far as I know, we do not have one currently in our secrets.

TotoGaz: Yes, I wanted to know if we had such a thing already.
If not, who could create one and store it in our secrets?

TotoGaz:
Done using the new feature of the project workflows instead of using a GHA.

PR #2635, opened 2023-08-26
Build/modify apple builds

rrsettgast:
- [x] fix apple builds
- [x] move apple hostconfigs


CusiniM (approved):
I have tried it and it works!

PR #2637, opened 2023-08-28
feat: choose the number of output vtu files

untereiner:


Algiane:
I am interested by this feature too: thanks for the PR.

untereiner:
It works as it is but you should be aware that their is no correlation for time-dependent outputs. It means that:
- it is correct to see a property for each time step 
- computing something between two timesteps is not ok because the merging can change the order of the elements.

It should be possible to keep an _absolute order_ by outputting the `localToGlobalMap` attribute

untereiner:
To be more precise. It is not possible to keep the coherence of rank merging between each time step. This means that a cell can be in different multi blocks at each time step. 

How to keep the consistency between the time steps, meaning: Get a properly from the same cell for each time step ?
I propose to use `localToGlobalMap` to reference the cells (and access to properties) because it is stable and across time.

herve-gross:
@untereiner is this ready for reviews? Can you explain why indices are shuffled across time steps?

untereiner:
@herve-gross yes it is ready. The indices are not shuffled.
Currently there is an isomorphism between the ranks in memory and the ranks written in a multiblock.
There is not possible to enforce which ranks that will be merged. Then for each time step you will have different ranks that will be merged into blocks.
So, if you go through the timesteps with paraview the resulting color map will be ok because it loads and renders each timestep independently.
But we observed that computing a difference between two timesteps will result in incoherent results because filters are applied per block.

paveltomin:
Waiting on code owner review from @MelReyCG, @rrsettgast, and/or @wrtobin.

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp
```diff
@@ -1188,13 +1185,13 @@ void VTKPolyDataWriterInterface::writeVtmFile( integer const cycle,
...
 
-      int const mpiSize = MpiWrapper::commSize();
+      // int const mpiSize = MpiWrapper::commSize();
 
       auto addElementRegion = [&]( ElementRegionBase const & region )
       {
         std::vector< string > const blockPath{ meshBody.getName(), meshLevel.getName(), region.getCatalogName(), region.getName() };
         string const regionPath = joinPath( meshPath, region.getName() );
-        for( int i = 0; i < mpiSize; i++ )
+        for( const auto & i : m_targetProcessesId.at( region.getName()) )
```
rrsettgast, line 1194: Is this to avoid writing empty regions to ranks that don't contain any elements in them?

rrsettgast (approved):

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp
untereiner: @rrsettgast yes it is

paveltomin:
requires rebaseline because of newly added `numberOfTargetProcesses`

src/coreComponents/fileIO/Outputs/VTKOutput.cpp
```diff
@@ -56,6 +56,11 @@ VTKOutput::VTKOutput( string const & name,
     setInputFlag( InputFlags::OPTIONAL ).
     setDescription( "Level detail plot. Only fields with lower of equal plot level will be output." );
 
+  registerWrapper( viewKeysStruct::numberOfTargetProcesses, &m_numberOfTargetProcesses ).
+    setApplyDefaultValue( MpiWrapper::commSize() ).
+    setInputFlag( InputFlags::OPTIONAL ).
+    setDescription( "Aggregate output data to be written." );
```
MelReyCG: I would expect something like:
```suggestion
    setDescription( "Number of output aggregate files to be written." );
```

src/coreComponents/fileIO/Outputs/VTKOutput.cpp
```diff
@@ -101,6 +106,7 @@ void VTKOutput::postInputInitialization()
   m_writer.setFieldNames( m_fieldNames.toViewConst() );
   m_writer.setLevelNames( m_levelNames.toViewConst() );
   m_writer.setOnlyPlotSpecifiedFieldNamesFlag( m_onlyPlotSpecifiedFieldNames );
+  m_writer.setNumberOfTargetProcesses( m_numberOfTargetProcesses );
```
MelReyCG, line 116: Valid values seem to be between 1 & MPI::commSize(), what do you think of adding `GEOS_ERROR()`s or at least `GEOS_WARNING()`s.
Source: https://vtk.org/doc/nightly/html/classvtkAggregateDataSetFilter.html#:~:text=void%20vtkAggregateDataSetFilter%3A%3ASetNumberOfTargetProcesses

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.hpp
```diff
@@ -210,7 +219,7 @@ class VTKPolyDataWriterInterface
   void clearData();
 
 
-private:
+protected:
```
MelReyCG: I may I missed something, but does anything inherits from `VTKPolyDataWriterInterface`?

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp
```diff
@@ -1279,15 +1279,50 @@ void VTKPolyDataWriterInterface::writeUnstructuredGrid( string const & path,
...
+  MpiWrapper::gather( &localCommRank,
+                      1,
+                      globalValues.data(),
+                      1,
+                      0,
+                      MPI_COMM_GEOS );
+
+  if( MpiWrapper::commRank() == 0 )
+  {
+    globalValues.erase( std::remove_if( globalValues.begin(),
```
MelReyCG, line 1321: I would add a small comment for that:
```suggestion
    // any rank that does not hold data do not participate in the output
    globalValues.erase( std::remove_if( globalValues.begin(),
```

MelReyCG (commented):
minor comments

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.hpp
untereiner: can be removed here because it is kept their: https://github.com/GEOS-DEV/GEOS/pull/2957/files#r1742425368

PR #2649, opened 2023-09-02
Add solver timers for logLevel > 1

paveltomin:
In order to simplify the testing and track performance without running with "-t", add few simple timers for solvers.
Example output:
```
Cleaning up events
compflow, number of time steps: 20
compflow, number of successful nonlinear iterations: 62
compflow, number of successful linear iterations: 62
compflow, number of time step cuts: 0
compflow, number of discarded nonlinear iterations: 0
compflow, number of discarded linear iterations: 0
compflow: apply solution time = 0.0222572 s (min), 0.026937007 s (max)
compflow: assemble time = 0.202956569 s (min), 0.209872378 s (max)
compflow: convergence check time = 0.002731362 s (min), 0.00538078 s (max)
compflow: linear solver time = 6.747940005 s (min), 6.752180354 s (max)
compflow: linear solver create time = 0.040424459 s (min), 0.041963672 s (max)
compflow: linear solver setup time = 6.528474797 s (min), 6.52893558 s (max)
compflow: linear solver solve time = 0.175283717 s (min), 0.175608148 s (max)
compflow: update state time = 0.028253674 s (min), 0.031888834 s (max)

```


paveltomin:
Example:
```
...
Cleaning up events
compflow, number of time steps: 10
compflow, number of successful nonlinear iterations: 25
compflow, number of successful linear iterations: 258
compflow, number of time step cuts: 0
compflow, number of discarded nonlinear iterations: 0
compflow, number of discarded linear iterations: 0
compflow: assemble time              00:00:10.563
compflow: convergence check time     00:00:00.165
compflow: linear solver time         00:00:58.983
compflow: apply solution time        00:00:00.236
compflow: update state time          00:00:01.272
Umpire            HOST sum across ranks:    5.5 GB
Umpire            HOST         rank max:  499.3 MB
Finished at 2023-09-01 19:55:58
total time            00:01:18.054
initialization time   00:00:04.620
run time              00:01:12.627

```

victorapm:
Hi @paveltomin,

thank you for the work you've put on this! I've gone through the changes, and while I appreciate the simplification proposed, I have a few concerns about tradeoff between the value brought by these changes and code maintenance. In one side, this makes it a bit simpler to parse the new output and get access to the timings which we, as developers, generally want to know. On the other hand, this adds new code to be maintained by the team, even if minimal, while the output of this new code is already fully accessible by an already existing approach ("-t" option with caliper). In addition, I remember about a discussion in the past with other developers in the direction of prioritizing caliper for gathering timings instead of using GEOS' internal timers. 

In short, I'm on the fence here about what to do and would like to hear from @rrsettgast and @klevzoff what they think.

Again, thank you for proposing these changes!

rrsettgast:
@paveltomin Can we just replace `Timer` with `Stopwatch`?
https://github.com/GEOS-DEV/GEOS/blob/f36934094ffb2ddc5a6d2f8b36a016c40ceb2e7f/src/coreComponents/common/Stopwatch.hpp#L31

paveltomin:
> @paveltomin Can we just replace `Timer` with `Stopwatch`?
> https://github.com/GEOS-DEV/GEOS/blob/f36934094ffb2ddc5a6d2f8b36a016c40ceb2e7f/src/coreComponents/common/Stopwatch.hpp#L31

Thank you, I was not aware of it, but it seems that Stopwatch is not doing += for the result, and just assigns instead, no?
I can move Timer class into a separate file if that is the issue...

paveltomin:
anyone?

rrsettgast:
yeah. sorry. please move it into its own file, and in the same component as stopwatch. We can see about joining them at a later date.

paveltomin:
> yeah. sorry. please move it into its own file, and in the same component as stopwatch. We can see about joining them at a later date.

ok done

paveltomin:
review anyone?

src/coreComponents/physicsSolvers/SolverBase.cpp
```diff
@@ -1236,6 +1317,18 @@ void SolverBase::cleanup( real64 const GEOS_UNUSED_PARAM( time_n ),
...
+
+  if( getLogLevel() > 0 )
+  {
+    for( auto & timer : m_timers )
+    {
+      real64 const time = std::chrono::duration< double >( timer.second ).count();
+      real64 const minTime = MpiWrapper::min( time );
+      real64 const maxTime = MpiWrapper::max( time );
+      if( maxTime > 0 )
+        GEOS_LOG_RANK_0( GEOS_FMT( getName() + ": {} time = {} s (min), {} s (max)", timer.first, minTime, maxTime ) );
```
klevzoff, line 1329: Consider making solver name part of the format string, also adding missing braces
```suggestion
      if( maxTime > 0 )
      {
        GEOS_LOG_RANK_0( GEOS_FMT( "{}: {} time = {} s (min), {} s (max)", getName(), timer.first, minTime, maxTime ) );
      }
```

klevzoff (approved):
I generally share the same concern as @victorapm about duplicating Caliper functionality in an ad-hoc manner. In addition, it's tied to an already overloaded `logLevel` setting. 

At the same time, I acknowledge that visually processing Caliper output is not easy and we don't have a better solution. Hence I think this is good to merge.

rrsettgast (approved):

PR #2652, opened 2023-09-05
fix: add early check for invalid `toMetricPrefixString` input

tbeltzun:
Fix https://github.com/GEOS-DEV/GEOS/issues/2651.

Use [`fpclassify`](https://en.cppreference.com/w/cpp/numeric/math/fpclassify) to avoid `-Werror=float-equal`.


rrsettgast:
@tbeltzun can you post what the output looks like with the zero case?

castelletto1:
Is this ready to be merged? It seems quite a minor fix

paveltomin:
@rrsettgast simple fix, merge?

src/coreComponents/codingUtilities/StringUtilities.cpp
```diff
@@ -73,6 +73,9 @@ string removeStringAndFollowingContent( string const & str,
 template< typename T >
 string toMetricPrefixString( T const & value )
 {
+  if( std::fpclassify( value ) == FP_ZERO )
+    return " 0.0  ";
+
```
wrtobin:
```suggestion
  if( std::fpclassify( value ) == FP_ZERO )
  {
    return " 0.0  ";
  }

```

rrsettgast (approved):

PR #2659, opened 2023-09-07
Github Actions - Run CUDA jobs only when read to merge

bmhan12:
This PR:
- ~~Duplicates the current CUDA job, each with different dependencies ("needs" in Github Actions terminology):~~
  - ~~Waits for Linux jobs to complete first before starting~~
  - ~~Runs immediately if PR is labeled "flag: ready to be merged"~~
 - ~~(This duplication is necessary because Github Actions [does not support optional needs](https://stackoverflow.com/questions/65844033/conditional-needs-in-github-action) like CMake)~~
 - Runs CUDA jobs only if labeled "flag: ready to be merged"
 - Makes non-debug CUDA jobs run on non-paid runners 
 - Refactors the matrix jobs to use a Github Actions [reusable workflow](https://docs.github.com/en/actions/using-workflows/reusing-workflows) to cut down on duplication.

Example of run with label "flag: ready to be merged":
~~https://github.com/GEOS-DEV/GEOS/actions/runs/6114497477~~
https://github.com/GEOS-DEV/GEOS/actions/runs/6279500154

~~Example of run without the flag:
https://github.com/GEOS-DEV/GEOS/actions/runs/6178832418~~

### DIFF
--- a/.github/workflows/ci_tests.yml
+++ b/.github/workflows/ci_tests.yml
@@ -159,69 +159,6 @@ jobs:
       OS: ${{ matrix.OS }}
     secrets: inherit
 
-  # Note: cuda_builds_merge_not_ready and cuda_builds_merge_ready jobs
-  # should have identical matrix.include entries.
-  # If "flag: ready to be merged" PR label not found, cuda jobs waits for
-  # linux jobs to pass before running.
-  cuda_builds_merge_not_ready:
-    name: ${{ matrix.name }}
-    needs: [check_pull_request_is_not_a_draft, linux_builds]
-    if: ${{ !(contains( github.event.pull_request.labels.*.name, format('flag{0} ready to be merged', ':'))) }}
-    strategy:
-
-      # In-progress jobs will not be cancelled if there is a failure
-      fail-fast : false
-      matrix:
-        include:
-          - name: Ubuntu CUDA debug (20.04, clang 10.0.0 + gcc 9.4.0, open-mpi 4.0.3, cuda-11.8.89)
-            DOCKER_REPOSITORY: geosx/ubuntu20.04-clang10.0.0-cuda11.8.89
-            OS: Runner_4core_16GB
-            CMAKE_BUILD_TYPE: Debug
-            BUILD_AND_TEST_ARGS: "--disable-unit-tests --build-exe-only --disable-schema-deployment"
-            ENABLE_HYPRE: ON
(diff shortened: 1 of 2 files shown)


.github/workflows/ci_tests.yml
```diff
@@ -10,11 +10,16 @@ env:
...
+
   # Matrix jobs will be cancelled if PR is a draft.
   # PR status must be "Open" to run CI.
   check_pull_request_is_not_a_draft:
     # We use the most recent ubuntu distribution available in Github Actions to ensure maximum support of google cloud's sdk.
     runs-on: ubuntu-22.04
+    # Workaround due to github actions variable scope restrictions passing to
+    # another job's env
+    outputs:
+      GEOSX_TPL_TAG: ${{ env.GEOSX_TPL_TAG}}
```
TotoGaz, line 22: Do you need to remove `GEOSX_TPL_TAG: 238-63` line 10 then?

rrsettgast (approved):

.github/workflows/ci_tests.yml
```diff
@@ -241,66 +205,112 @@ jobs:
...
-      if: matrix.GCP_BUCKET
-      uses: 'google-github-actions/setup-gcloud@v1'
-      with:
-        version: '>= 363.0.0'
-
-    - name: Print environment
-      run: printenv
+      # In-progress jobs will not be cancelled if there is a failure
+      fail-fast : false
+      matrix:
```
TotoGaz, line 235: Is it possible to make an additional workflow "template" (like `build_and_test.yml`) which would call `build_and_test.yml` backend, with one unique source of parameters? (a template of template...).
Then we could call this template if the `if` branch and in the `if not` as well without any duplication.
If not, then I guess that this is what it is.

.github/workflows/ci_tests.yml
bmhan12: > Do you need to remove GEOSX_TPL_TAG: 238-63 line 10 then?

It is still necessary.
The expression here is to pass `GEOSX_TPL_TAG` to the reusable workflow.

.github/workflows/ci_tests.yml
TotoGaz: OK, let's keep it like it is. `yaml` is the death of computer science.

TotoGaz (approved):

PR #2660, opened 2023-09-07
Dissipation

rasimHZ:
Implemented and Validated Dissipation Based Continuation Method

Tests update: https://github.com/GEOS-DEV/integratedTests/pull/39


paveltomin (approved):

src/coreComponents/physicsSolvers/NonlinearSolverParameters.hpp
```diff
@@ -272,6 +346,27 @@ class NonlinearSolverParameters : public dataRepository::Group
 
   /// Flag to specify whether subcycling is allowed or not in sequential schemes
   integer m_subcyclingOption;
+
```
paveltomin, line 349: I personally prefer long list of parameters to be grouped in a struct

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.cpp
```diff
@@ -138,20 +139,46 @@ void CompositionalMultiphaseFVM::assembleFluxTerms( real64 const dt,
...
+          DissipationCompositionalMultiphaseFVMKernels::
+            FaceBasedAssemblyKernelFactory::
+            createAndLaunch< parallelDevicePolicy<> >( m_numComponents,
+                                                       m_numPhases,
+                                                       dofManager.rankOffset(),
+                                                       elemDofKey,
+                                                       m_hasCapPressure,
+                                                       getName(),
+                                                       mesh.getElemManager(),
+                                                       //subRegion,
```
paveltomin, line 153: remove

src/coreComponents/physicsSolvers/fluidFlow/DissipationCompositionalMultiphaseFVMKernels.hpp
```diff
@@ -0,0 +1,472 @@
...
+ */
+
+/**
+ * @file DissipationCompositionalMultiphaseFVMKernels.hpp
+ */
+
+#ifndef GEOS_PHYSICSSOLVERS_FLUIDFLOW_DISSIPATIONCOMPOSITIONALMULTIPHASEFVMKERNELS_HPP
+#define GEOS_PHYSICSSOLVERS_FLUIDFLOW_DISSIPATIONCOMPOSITIONALMULTIPHASEFVMKERNELS_HPP
+
+#include "physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseFVMKernels.hpp"
```
paveltomin, line 22: no need for path here? just #include "IsothermalCompositionalMultiphaseFVMKernels.hpp"

src/coreComponents/physicsSolvers/fluidFlow/DissipationCompositionalMultiphaseFVMKernels.hpp
```diff
@@ -0,0 +1,472 @@
...
+   * @param[in] dissCompFlowAccessor accessor for wrappers registered by the solver needed for dissipation
+   * @param[in] multiFluidAccessor accessor for wrappers registered by the multifluid model
+   * @param[in] dissMultiFluidAccessor accessor for wrappers registered by the multifluid model needed for dissipation
+   * @param[in] capPressureAccessors accessor for wrappers registered by the cap pressure model
+   * @param[in] permeabilityAccessors accessor for wrappers registered by the permeability model
+   * @param[in] porosityAccessors accessor for wrappers registered by the porosity model
+   * @param[in] relPermAccessors accessor for wrappers registered by the relative permeability model needed for dissipation
+   * @param[in] dt time step size
+   * @param[inout] localMatrix the local CRS matrix
+   * @param[inout] localRhs the local right-hand side vector
```
paveltomin, line 132: add the rest of DBC params

src/coreComponents/physicsSolvers/fluidFlow/DissipationCompositionalMultiphaseFVMKernels.hpp
```diff
@@ -0,0 +1,472 @@
...
+                           real64 const omega,
+                           integer const curNewton,
+                           integer const continuation,
+                           integer const miscible,
+                           real64 const kappamin,
+                           real64 const contMultiplier )
+    : Base( numPhases,
+            rankOffset,
+            hasCapPressure,
+            //0.0,                  // no C1-PPU
```
paveltomin, line 159: remove commented

src/coreComponents/physicsSolvers/fluidFlow/DissipationCompositionalMultiphaseFVMKernels.hpp
```diff
@@ -0,0 +1,472 @@
...
+   * @param[in] permeabilityAccessors accessor for wrappers registered by the permeability model
+   * @param[in] porosityAccessors accessor for wrappers registered by the porosity model
+   * @param[in] relPermAccessors accessor for wrappers registered by the relative permeability model needed for dissipation
+   * @param[in] dt time step size
+   * @param[inout] localMatrix the local CRS matrix
+   * @param[inout] localRhs the local right-hand side vector
+   */
+  FaceBasedAssemblyKernel( integer const numPhases,
+                           globalIndex const rankOffset,
+                           integer const hasCapPressure,
```
paveltomin, line 136: i am pretty sure that some of params can be dropped, e.g. related to cap pressure

src/coreComponents/physicsSolvers/fluidFlow/DissipationCompositionalMultiphaseFVMKernels.hpp
```diff
@@ -0,0 +1,472 @@
...
+  template< typename POLICY, typename STENCILWRAPPER >
+  static void
+  createAndLaunch( integer const numComps,
+                   integer const numPhases,
+                   globalIndex const rankOffset,
+                   string const & dofKey,
+                   integer const hasCapPressure,
+                   string const & solverName,
+                   ElementRegionManager const & elemManager,
+                   //ElementSubRegionBase const & subRegion,
```
paveltomin, line 425: remove commented lines

paveltomin:
@jafranc, @francoishamon, @dkachuma i think it's ready for you to have a look and let @rasimHZ know what you think

src/coreComponents/physicsSolvers/NonlinearSolverParameters.cpp
```diff
@@ -149,6 +149,36 @@ NonlinearSolverParameters::NonlinearSolverParameters( string const & name,
...
+    setApplyDefaultValue( 1 ).
+    setInputFlag( InputFlags::OPTIONAL ).
+    setDescription( "Factor by which DBC flux is multiplied" );
+
+  registerWrapper( viewKeysStruct::continuationDBCString(), &m_continuationDBC ).
+    setApplyDefaultValue( 1 ).
+    setInputFlag( InputFlags::OPTIONAL ).
+    setDescription( "Flag for enabling continuation parameter" );
+
+  registerWrapper( viewKeysStruct::miscibleDBCString(), &m_miscibleDBC ).
```
jafranc, line 167: Can't we deduce that from the fluid model in use and `m_useDBC` ?

src/coreComponents/physicsSolvers/NonlinearSolverParameters.cpp
```diff
@@ -149,6 +149,36 @@ NonlinearSolverParameters::NonlinearSolverParameters( string const & name,
     setApplyDefaultValue( 0 ).
     setDescription( "Flag to decide whether to iterate between sequentially coupled solvers or not." );
 
+  registerWrapper( viewKeysStruct::useDBCString(), &m_useDBC ).
```
jafranc, line 152: Minor concern but I think there is a recommendation to use full length name for readability in GEOS.

src/coreComponents/physicsSolvers/NonlinearSolverParameters.hpp
jafranc: I agree with @paveltomin on this.

src/coreComponents/physicsSolvers/fluidFlow/DissipationCompositionalMultiphaseFVMKernels.hpp
```diff
@@ -0,0 +1,469 @@
...
+    m_curNewton( curNewton ),
+    m_continuationDBC( continuation ),
+    m_miscibleDBC( miscible ),
+    m_kappaminDBC( kappamin ),
+    m_contMultiplierDBC( contMultiplier ),
+    m_volume( dissCompFlowAccessors.get( fields::elementVolume {} ) ),
+    m_porosity_n( porosityAccessors.get( fields::porosity::porosity_n {} ) )
+  {}
+
+  struct StackVariables : public Base::StackVariables
```
jafranc, line 187: Is it really needed if none variable is added to the struct ? isn't a `using StackVariables = Base::StackVariables` possible ?

src/coreComponents/physicsSolvers/fluidFlow/DissipationCompositionalMultiphaseFVMKernels.hpp
```diff
@@ -0,0 +1,469 @@
...
+        else
+        {
+          for( int mp = 0; mp < m_curNewton; mp++ ) kappaDBC *= m_contMultiplierDBC;
+          kappaDBC = std::max( kappaDBC, m_kappaminDBC );
+        }
+      }
+      /// Step 2. Collect all contributions
+      real64 poreVolume_n = 0; // Pore volume contribution
+      real64 trans = 0; // Transmissibility contribution
+      for( integer ke = 0; ke < numFluxSupportPoints; ++ke )
```
jafranc, line 270: IIUC we are assuming TPFA, hence the `0.5` prefactor. It is not as such a strong assumption but can we comment about it just for maintenance purpose

src/coreComponents/physicsSolvers/fluidFlow/DissipationCompositionalMultiphaseFVMKernels.hpp
```diff
@@ -0,0 +1,469 @@
...
+      }
+
+      // bias towards x and y direction for miscible
+      real64 directional_coef;
+      if( m_miscibleDBC )
+      {
+        directional_coef = 100.0;
+        if( LvArray::math::abs( grad_depth ) != 0 )
+        {
+          if( 1000.f / LvArray::math::abs( grad_depth * grad_depth ) < 100 )
```
jafranc, line 297: 1. I am not sure I understand the hard-coded threshold criteria here
2. What if gravity is tilted ?
I guess this should be let to the user to specify `m_miscibleBias` as a tensor and this would spare you a `m_miscibleDBC` definition and few if-s

src/coreComponents/physicsSolvers/fluidFlow/DissipationCompositionalMultiphaseFVMKernels.hpp
```diff
@@ -0,0 +1,469 @@
...
+    {
+      GEOS_UNUSED_VAR( k_up, potGrad, phaseFlux, dPhaseFlux_dP, dPhaseFlux_dC, er_up, esr_up, ei_up );
+
+      /// Storing dissipation flux and its derivatives locally
+      real64 dissFlux[numComp]{};
+      real64 dDissFlux_dP[numFluxSupportPoints][numComp]{};
+      real64 dDissFlux_dC[numFluxSupportPoints][numComp][numComp]{};
+      real64 fluxPointCoef[numFluxSupportPoints] = {1.0, -1.0}; // for gradients
+
+      real64 viscosityMult[3] = {1.0, 1.0, 1.0}; // for viscosity
```
jafranc, line 250: Is local multipliers `viscosityMult` likely to be altered to another value than `{1.,1.,1.}` ?

jafranc (changes requested):
Good work, maybe needs some parameters clean-up and documentation. 
High level, do you foresee a use that would be outside of Flow model ? 
If yes and as it is a non-linear technic, it would sound logic to me to try to have it on the Nonlinear Solution side of *GEOS*

inputFiles/compositionalMultiphaseFlow/dbc/buckleyLeverett_1d/reference_output/buckleyLeverett_benchmark_DBC_log.txt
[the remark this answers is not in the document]
paveltomin: yeah, those logs formally should be removed once proper baselines are generated
although in general having baseline logs is super helpful

src/coreComponents/physicsSolvers/NonlinearSolverParameters.cpp
paveltomin: good question, @francoishamon can you point me how to do that?

src/coreComponents/physicsSolvers/NonlinearSolverParameters.cpp
paveltomin: let's start with "GEOS" then haha

src/coreComponents/physicsSolvers/fluidFlow/DissipationCompositionalMultiphaseFVMKernels.hpp
paveltomin: this is a bit work in progress to be honest... needs to be generalized and improved

paveltomin:
> Good work, maybe needs some parameters clean-up and documentation. High level, do you foresee a use that would be outside of Flow model ? If yes and as it is a non-linear technic, it would sound logic to me to try to have it on the Nonlinear Solution side of _GEOS_

I don't think it makes sense to use it outside of flow solver, do you suggest moving the setting parameters into the solver itself instead of nonlinear params?

jafranc:
> > Good work, maybe needs some parameters clean-up and documentation. High level, do you foresee a use that would be outside of Flow model ? If yes and as it is a non-linear technic, it would sound logic to me to try to have it on the Nonlinear Solution side of _GEOS_
> 
> I don't think it makes sense to use it outside of flow solver, do you suggest moving the setting parameters into the solver itself instead of nonlinear params?

I was thinking if it can be treated as a generic treatment in 
https://github.com/GEOS-DEV/GEOS/blob/f37e593da54d6a8b13d29431808dfa33e6f2900a/src/coreComponents/physicsSolvers/SolverBase.cpp#L771-L774
if the reciepe was generic enough to not rely on low-level physics parameters.

paveltomin:
> > > Good work, maybe needs some parameters clean-up and documentation. High level, do you foresee a use that would be outside of Flow model ? If yes and as it is a non-linear technic, it would sound logic to me to try to have it on the Nonlinear Solution side of _GEOS_
> > 
> > 
> > I don't think it makes sense to use it outside of flow solver, do you suggest moving the setting parameters into the solver itself instead of nonlinear params?
> 
> I was thinking if it can be treated as a generic treatment in
> 
> https://github.com/GEOS-DEV/GEOS/blob/f37e593da54d6a8b13d29431808dfa33e6f2900a/src/coreComponents/physicsSolvers/SolverBase.cpp#L771-L774
> 
> 
> if the reciepe was generic enough to not rely on low-level physics parameters.

no, it's not general, if you familiar with @ryar9534 work on stabilization, dbc is in the same class of methods imho

src/coreComponents/physicsSolvers/fluidFlow/DissipationCompositionalMultiphaseFVMKernels.hpp
```diff
@@ -0,0 +1,384 @@
...
+      real64 directional_coef = 1.0;
+      if( m_miscibleDBC )
+      {
+        directional_coef = 100.0;
+        if( LvArray::math::abs( grad_depth ) > 0.0 )
+        {
+          real64 const d2 = LvArray::math::abs( grad_depth * grad_depth );
+          if( 1000.0 / d2 < 100.0 )
+            directional_coef = 1000.0 / d2;
+        }
```
dkachuma, line 235: Can you please add a comment about the value 1000 and the cutoff value of 100.

src/coreComponents/physicsSolvers/NonlinearSolverParameters.hpp
paveltomin: i moved those params into CompositionalMultiphaseFVM and grouped them into a struct

paveltomin:
buckleyLeverett_1d

no DBC

buckleyLeverett_1d, number of time steps: 24
buckleyLeverett_1d, number of successful nonlinear iterations: 979
buckleyLeverett_1d, number of successful linear iterations: 979

with DBC

buckleyLeverett_1d_DBC, number of time steps: 24
buckleyLeverett_1d_DBC, number of successful nonlinear iterations: 230
buckleyLeverett_1d_DBC, number of successful linear iterations: 230


grav_seg_1d

no DBC

grav_seg_1d, number of time steps: 27
grav_seg_1d, number of successful nonlinear iterations: 898
grav_seg_1d, number of successful linear iterations: 898
grav_seg_1d, number of time step cuts: 3
grav_seg_1d, number of discarded nonlinear iterations: 300
grav_seg_1d, number of discarded linear iterations: 300

with DBC

grav_seg_1d_DBC, number of time steps: 18
grav_seg_1d_DBC, number of successful nonlinear iterations: 275
grav_seg_1d_DBC, number of successful linear iterations: 275
grav_seg_1d_DBC, number of time step cuts: 1
grav_seg_1d_DBC, number of discarded nonlinear iterations: 100
grav_seg_1d_DBC, number of discarded linear iterations: 100


bottom_layer_SPE10

no DBC

bottom_layer_SPE10, number of time steps: 18
bottom_layer_SPE10, number of successful nonlinear iterations: 130
bottom_layer_SPE10, number of successful linear iterations: 130

with DBC

bottom_layer_SPE10, number of time steps: 18
bottom_layer_SPE10, number of successful nonlinear iterations: 99
bottom_layer_SPE10, number of successful linear iterations: 99

src/coreComponents/physicsSolvers/NonlinearSolverParameters.hpp
```diff
@@ -96,6 +97,14 @@ class NonlinearSolverParameters : public dataRepository::Group
...
     static constexpr char const * subcyclingOptionString()               { return "subcycling"; }
+
+    static constexpr char const * useDBCString()                  { return "useDBC"; }
+    static constexpr char const * omegaDBCString()                { return "omegaDBC"; }
+    static constexpr char const * continuationDBCString()         { return "continuationDBC"; }
+
+    static constexpr char const * miscibleDBCString()             { return "miscibleDBC"; }
+    static constexpr char const * kappaminDBCString()             { return "kappaminDBC"; }
+    static constexpr char const * contMultiplierDBCString()       { return "contMultiplierDBC"; }
   } viewKeys;
```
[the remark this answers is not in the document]
paveltomin: Good point, i decided to move them into CompositionalMultiphaseFVM since the methos is multiphase FVM specific

src/coreComponents/physicsSolvers/fluidFlow/DissipationCompositionalMultiphaseFVMKernels.hpp
```diff
@@ -0,0 +1,469 @@
...
+      real64 fluxPointCoef[numFluxSupportPoints] = {1.0, -1.0}; // for gradients
+
+      real64 viscosityMult[3] = {1.0, 1.0, 1.0}; // for viscosity
+
+      /// Step 1. Calculate the continuation parameter based on the current Newton iteration
+      real64 kappaDBC = 1.0; // default value
+
+      if( m_continuationDBC ) // if continuation is enabled
+      {
+        if( m_curNewton >= 5 )
```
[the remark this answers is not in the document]
rasimHZ: Good point, @jafranc, will do. @dkachuma, to your comment: the short answer is no. At least in Jiamin Jiang's original paper on DBC, this heuristic was fixed.

src/coreComponents/physicsSolvers/fluidFlow/DissipationCompositionalMultiphaseFVMKernels.hpp
rasimHZ: @dkachuma The point of this "directional coefficient" was to add less dissipation flux in the direction of gravity and more in all other directions motivated by the case of CO2 injection with plume formation. These hard-coded values of 1000 and 100 were picked by some tuning and as expected are not robust. Further research/testing is needed to generalize this idea, however, due to short time for the internship, I left these values in the code.

src/coreComponents/physicsSolvers/fluidFlow/DissipationCompositionalMultiphaseFVMKernels.hpp
rasimHZ: @jafranc The idea behind these viscosity multipliers was to include the effect of the phase viscosities  (different dissipation flux for different phases). However, the viscosities of the phases from the previous timestep (time n) are not stored in GEOS (at least to my knowledge), so I couldn't add them (and so kept them at 1.0). I've tried different values for these multipliers (by hard coding), but did not see an order 1 impact on the nonlinear convergence. And then, of course, more testing and sensitivity is to be done to have an adaptive tuning for them.

src/coreComponents/physicsSolvers/fluidFlow/DissipationCompositionalMultiphaseFVMKernels.hpp
rasimHZ: Done.

src/coreComponents/physicsSolvers/fluidFlow/DissipationCompositionalMultiphaseFVMKernels.hpp
```diff
@@ -0,0 +1,469 @@
...
+                                           real64 const (&dPhaseFlux_dP)[2],
+                                           real64 const (&dPhaseFlux_dC)[2][numComp] )
+    {
+      GEOS_UNUSED_VAR( k_up, potGrad, phaseFlux, dPhaseFlux_dP, dPhaseFlux_dC, er_up, esr_up, ei_up );
+
+      /// Storing dissipation flux and its derivatives locally
+      real64 dissFlux[numComp]{};
+      real64 dDissFlux_dP[numFluxSupportPoints][numComp]{};
+      real64 dDissFlux_dC[numFluxSupportPoints][numComp][numComp]{};
+      real64 fluxPointCoef[numFluxSupportPoints] = {1.0, -1.0}; // for gradients
```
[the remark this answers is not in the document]
rasimHZ: @francoishamon could you please have a look?

src/coreComponents/physicsSolvers/fluidFlow/DissipationCompositionalMultiphaseFVMKernels.hpp
rasimHZ: @jafranc  added comment in the code. Generalizing to modify from input file left as future work.

src/coreComponents/physicsSolvers/fluidFlow/DissipationCompositionalMultiphaseFVMKernels.hpp
rasimHZ: @dkachuma comment added in the code

paveltomin:
@jafranc and @dkachuma could you please have another look?

src/coreComponents/physicsSolvers/fluidFlow/DissipationCompositionalMultiphaseFVMKernels.hpp
```diff
@@ -0,0 +1,395 @@
...
+
+      real64 viscosityMult[3] = {1.0, 1.0, 1.0}; // for viscosity
+
+      /// Step 2. Collect all contributions
+      real64 poreVolume_n = 0.0; // Pore volume contribution
+      real64 trans = 0.0; // Transmissibility contribution
+      for( integer ke = 0; ke < numFluxSupportPoints; ++ke )
+      {
+        trans = std::max( stack.transmissibility[connectionIndex][ke], trans );
```
jafranc, line 213: This I believe is supposed to bring you sign for your fluxes with pair of positive of negative values depending on norm convention. See. here

https://github.com/GEOS-DEV/GEOS/blob/4d2011e8f50680bb5a7dbeff29f58d6bf8e67239/src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseFVMKernelUtilities.hpp#L151

src/coreComponents/physicsSolvers/fluidFlow/DissipationCompositionalMultiphaseFVMKernels.hpp
```diff
@@ -0,0 +1,395 @@
...
+        if( LvArray::math::abs( grad_depth ) > 0.0 )
+        {
+          real64 const d2 = LvArray::math::abs( grad_depth * grad_depth );
+          if( 1000.0 / d2 < 100.0 )
+            directional_coef = 1000.0 / d2;
+        }
+      }
+
+      // multiplier with all contributions
+      real64 const multiplier = m_kappaDBC * m_omegaDBC * trans / poreVolume_n * m_dt * potential_gradient * directional_coef;
```
jafranc, line 250: Following above comment `trans` should be abs closer to this point

jafranc (approved):
Provided usage of signed `transmissivity` and gathering harcoded values into `static` class var. Good to me.

src/coreComponents/physicsSolvers/fluidFlow/DissipationCompositionalMultiphaseFVMKernels.hpp
```diff
@@ -0,0 +1,395 @@
...
+      {
+        m_kappaDBC = kappamin;
+      }
+      else
+      {
+        for( int mp = 0; mp < curNewton; mp++ )
+        {
+          m_kappaDBC *= contMultiplier;
+        }
+        m_kappaDBC = std::max( m_kappaDBC, kappamin );
```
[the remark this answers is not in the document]
rasimHZ: @jafranc 

> IIRC it is sequential only, disregard the comment if it is indeed, maybe add a comment about that fact. If not, this should be replace by parallel safe version.

m_kappaDBC is computed based on the current newton iteration number which is the same across all ranks. Should be fine for parallel (tested, works fine for parallel runs).

src/coreComponents/physicsSolvers/fluidFlow/DissipationCompositionalMultiphaseFVMKernels.hpp
```diff
@@ -0,0 +1,395 @@
...
+        for( integer ic = 0; ic < numComp; ++ic )
+        {
+          localIndex const er  = seri[ke];
+          localIndex const esr = sesri[ke];
+          localIndex const ei  = sei[ke];
+
+          // composition gradient contribution to the dissipation flux
+          dissFlux[ic] += coef * m_compFrac[er][esr][ei][ic];
+
+          dDissFlux_dP[ke][ic] = 0; // Pavel: TODO ?
```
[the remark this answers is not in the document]
rasimHZ: > Doubling Pavel's comment, is it always 0, lumped or todo ? Please adjust 287-288 accordingly

Removed Pavel's TODO, because there is no dependence on pressure at n+1, since m_compFrac is global component fraction (z_c)

src/coreComponents/physicsSolvers/fluidFlow/DissipationCompositionalMultiphaseFVMKernels.hpp
rasimHZ: > This I believe is supposed to bring you sign for your fluxes with pair of positive of negative values depending on norm convention. See. here
> 
> https://github.com/GEOS-DEV/GEOS/blob/4d2011e8f50680bb5a7dbeff29f58d6bf8e67239/src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseFVMKernelUtilities.hpp#L151

fixed.

src/coreComponents/physicsSolvers/fluidFlow/DissipationCompositionalMultiphaseFVMKernels.hpp
rasimHZ: > Following above comment `trans` should be abs closer to this point

fixed

src/coreComponents/physicsSolvers/fluidFlow/DissipationCompositionalMultiphaseFVMKernels.hpp
```diff
@@ -0,0 +1,395 @@
...
+    m_compFrac( dissCompFlowAccessors.get( fields::flow::globalCompFraction {} ) ),
+    m_omegaDBC( omega ),
+    m_miscibleDBC( miscible )
+  {
+    /// Step 1. Calculate the continuation parameter based on the current Newton iteration
+    m_kappaDBC = 1.0;   // default value
+
+    if( continuation )   // if continuation is enabled
+    {
+      if( curNewton >= 5 )
```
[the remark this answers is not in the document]
rasimHZ: done

> Can we before merge gather these hardcoded values into `static` member var located in the same place so future revision is easier

PR #2662, opened 2023-09-07
Modification of trailing face check in SIF calculation to allow for intersecting fractures

frankfeifan:
**Issue in the current code:**  when one crack intersects or is about to intersect with another crack, it is possible to have trailing faces that contain more than two unpinched nodes, which is out of those [conditions](https://github.com/GEOS-DEV/GEOS/blob/3e0456d4f7e7877e2b60497a8c9bc527b7212171/src/coreComponents/physicsSolvers/surfaceGeneration/SurfaceGenerator.cpp#L2996C12-L2996C12) that we consider in `calculateNodeAndFaceSif`. 

In this PR, I simply change the [if-statement](https://github.com/GEOS-DEV/GEOS/blob/3e0456d4f7e7877e2b60497a8c9bc527b7212171/src/coreComponents/physicsSolvers/surfaceGeneration/SurfaceGenerator.cpp#L2925) that checks if the SIF calculation is needed for a specific trailing face. With this change, we don't calculate SIF for those trailing faces with more than two unpinched nodes, and thus circumvent the above issue in the intersecting fracture problem.


CusiniM (approved):
Let's just make sure that we run all integrated tests.

frankfeifan:
> Let's just make sure that we run all integrated tests.

Yes. Just checked that it passes all integrated tests.

rrsettgast (approved):

PR #2678, opened 2023-09-12
Implemented diffusion and (preliminary) dispersion flux 

francoishamon:
This PR adds the functions and constitutive models to simulation diffusion and dispersion in the compositional solver. 
The formulation follows the description of the SPE11 benchmark.

Screenshot 2023-09-13 at 12 17 49 AM

Diffusion is fully implemented:
- [X] Constant diffusion constitutive model
- [X] Diffusion flux kernel

Dispersion is only partially implemented and disabled for now:
- [X] Linear isotropic dispersion constitutive model
- [X] Dispersion flux kernel

The computation of the velocity term |u_alpha| in the dispersion flux is not yet implemented and will be addressed by @jafranc in a separate PR. It requires some attention if we want to implement it without duplicating a lot of code.

Rebaseline done in https://github.com/GEOS-DEV/integratedTests/pull/50


src/coreComponents/constitutive/diffusion/ConstantDiffusion.hpp
```diff
@@ -0,0 +1,112 @@
...
+  } viewKeys;
+
+protected:
+
+  virtual void postProcessInput() override;
+
+private:
+
+  /// default diffusivity in the subRegion
+  R1Tensor m_diffusivityComponents;
```
jafranc, line 103: Idk if I remember correctly but wasn't there a policy not to use `R1Tensor` anymore ?

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -178,6 +184,22 @@ void CompositionalMultiphaseBase::registerDataOnMesh( Group & meshBodies )
       {
         m_hasCapPressure = true;
       }
+
+      // If at least one region has a diffusion model, consider it enabled for all
+      string const diffusionName = getConstitutiveName< DiffusionBase >( subRegion );
+      if( !diffusionName.empty() )
+      {
+        GEOS_ERROR( "Dispersion is not supported yet, please remove this model from this XML file" );
```
paveltomin, line 192: other way around, right?

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -789,8 +832,25 @@ void CompositionalMultiphaseBase::initializeFluidState( MeshLevel & mesh,
...
+      DiffusionBase const & diffusionMaterial = getConstitutiveModel< DiffusionBase >( subRegion, diffusionName );
+      arrayView1d< real64 const > const temperature = subRegion.template getField< fields::flow::temperature >();
+      diffusionMaterial.initializeTemperatureState( temperature );
     }
+    if( m_hasDispersion )
+    {
+      string const & dispersionName = subRegion.template getReference< string >( viewKeyStruct::dispersionNamesString() );
+      DispersionBase const & dispersionMaterial = getConstitutiveModel< DispersionBase >( subRegion, dispersionName );
+      GEOS_UNUSED_VAR( dispersionMaterial );
+      // TODO: compute the total velocity here
```
paveltomin, line 850: why total velocity? should be phase, no?

src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseFVMKernels.hpp
```diff
@@ -850,6 +854,721 @@ class FaceBasedAssemblyKernelFactory
...
+
+            // add contributions of the derivatives of upwind coefficient wrt pressure/component fractions
+            real64 const dUpwindCoefficient_dP =
+              m_referencePorosity[er_up][esr_up][ei_up] *
+              m_phaseDiffusivityMultiplier[er_up][esr_up][ei_up][0][ip] *
+              ( m_dPhaseDens[er_up][esr_up][ei_up][0][ip][Deriv::dP] * m_phaseVolFrac[er_up][esr_up][ei_up][ip]
+                + m_phaseDens[er_up][esr_up][ei_up][0][ip] * m_dPhaseVolFrac[er_up][esr_up][ei_up][ip][Deriv::dP] );
+            dDiffusionFlux_dP[k_up][ic] += dUpwindCoefficient_dP * compFracGrad;
+
+            applyChainRule( numComp,
```
paveltomin, line 1144: sorry, i don't understand that applyChainRule here, why can't we use m_dPhaseDens[er_up][esr_up][ei_up][0][ip][Deriv::dC + jc] directly below?

inputFiles/compositionalMultiphaseWell/staircase_co2_wells_3d.xml
```diff
@@ -182,6 +182,13 @@
     <ConstantPermeability
       name="rockPerm"
       permeabilityComponents="{ 2.0e-16, 2.0e-16, 2.0e-16 }"/>
+
+    <ConstantDiffusion
+      name="diffusion"
+      phaseNames="{ gas, water }"
+      defaultPhaseDiffusivityMultipliers="{ 20, 1 }"
+      diffusivityComponents="{ 1e-9, 1e-9, 1e-9 }"/>
```
paveltomin, line 190: is diffusivity tensor?

paveltomin (approved):
looks ok

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
francoishamon: Thanks for the catch, you are right, I fixed it

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
francoishamon: Yes, if we stick to the SPE11 benchmark description, it must be the phase velocity. @jafranc will handle this part. 

In the current framework, if we use the phase velocity to compute a dispersivity tensor, we have to call `computeWeights` twice in the kernel, which is a little bit annoying. So initially I thought we could use the total velocity instead to avoid this. But @jafranc has a better idea than using the total velocity here and will test it soon to see if it is accurate.

In the meantime, I updated the test that you highlighted to mention phase velocities and not total velocity.

inputFiles/compositionalMultiphaseWell/staircase_co2_wells_3d.xml
francoishamon: Yes, I agree with your suggestion. For now, the diffusivity (and the thermal conductivity) use exactly the same terminology as the permeability since they play the same role in the code. That is why I reused the term `components` here. 

But if we all agree that it is a good time to switch from `components` to `tensor`, then let's do it once for all for permeability, thermal conductivity, and diffusivity (in a separate PR possibly). One thing to keep in mind is that this is going to break a lot of xml files.

src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseFVMKernels.hpp
francoishamon: We need to use the chain rule here because m_dPhaseDens[er_up][esr_up][ei_up][0][ip][Deriv::dC + jc] is computed by the fluid model as the derivative of phase density (rho_l) with respect to component fractions (z_c), and we need to assemble derivatives of the mass conservation equation with respect to component densities (rho_c).

So, we have to first compute (drho_l/drho_c) = (drho_l/dz_c)*(dz_c/drho_c) and then use the result to compute the derivative of the diffusive flux with respect to component densities.

paveltomin: why don't we do it right after fluid model compute once and for all? it's really confusing...

francoishamon: Yes, we can definitely do it as another state update step after calling the fluid model. There is quite a bit of code to modify, but it would simplify a lot of things (and possibly avoid future implementation mistakes)

jafranc (approved):
Sounds good !! Thank you @francoishamon

src/coreComponents/constitutive/diffusion/ConstantDiffusion.cpp
```diff
@@ -0,0 +1,81 @@
...
+{
+  return DiffusionBase::deliverClone( name, parent );
+}
+
+void ConstantDiffusion::allocateConstitutiveData( dataRepository::Group & parent,
+                                                  localIndex const numConstitutivePointsPerParentIndex )
+{
+  DiffusionBase::allocateConstitutiveData( parent, numConstitutivePointsPerParentIndex );
+
+  for( localIndex ei = 0; ei < parent.size(); ++ei )
```
CusiniM, line 50: I would move this to a function called `initialize() `and write a device kernel for it.

src/coreComponents/constitutive/diffusion/DiffusionBase.hpp
```diff
@@ -0,0 +1,195 @@
...
+
+protected:
+
+  virtual void postProcessInput() override;
+
+  /// phase names read from input
+  string_array m_phaseNames;
+
+  /// array of phase-wise default diffusion multipliers
+  array1d< real64 > m_defaultPhaseDiffusivityMultiplier;
```
CusiniM, line 177: what's the physical meaning of a diffusivity multiplier? Is it because diffusivity varies by phase? If that's the case why not sizing it based on the number of phases?

CusiniM (approved):

src/coreComponents/constitutive/diffusion/DiffusionBase.hpp
francoishamon: Yes, the diffusivity varies by phase:

Screenshot 2023-09-18 at 11 34 58 PM

The main reason why it is implemented with 1) a unique diffusion tensor and 2) diffusivity multipliers is the `computeWeights` function here: 
https://github.com/GEOS-DEV/GEOS/blob/39f6373ea92a0501c256960783b5e942c3c3ebfa/src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseFVMKernels.hpp#L1064-L1068
Right now I call `computeWeights` only once and then I "scale" it for each phase with the phase multipliers to get the phase diffusivity. 

But we could also remove the phase multipliers and either 
1. Call `computeWeights` `n_p` times (once for each phase)
2. Call `computeWeights` once on a phase-based diffusivity tensor

In both cases, we would need to add a `computeWeights` function that takes a:
```
CoefficientAccessor< arrayView4d< real64 const > > const & coefficient
```
and refactor slightly the existing function to avoid code duplication. But that would be cleaner and also very useful for dispersion which will have the same issue. Let me know what you think, I would be glad to try it out

src/coreComponents/constitutive/diffusion/ConstantDiffusion.cpp
francoishamon: Will do!

src/coreComponents/constitutive/diffusion/DiffusionBase.cpp
```diff
@@ -0,0 +1,95 @@
...
+  ConstitutiveBase::allocateConstitutiveData( parent, numConstitutivePointsPerParentIndex );
+
+  for( localIndex ei = 0; ei < parent.size(); ++ei )
+  {
+    // NOTE: enforcing 1 quadrature point
+    for( localIndex q = 0; q < 1; ++q )
+    {
+      for( integer ip = 0; ip < numFluidPhases(); ++ip )
+      {
+        m_phaseDiffusivityMultiplier[ei][q][ip] = m_defaultPhaseDiffusivityMultiplier[ip];
```
dkachuma, line 86: I don't understand how we are linking the phase names in `m_phaseNames` to the indices here? For instance, if the user changes the ordering of the `phaseNames` field, do we expect the `defaultPhaseDiffusivityMultiplier` to apply to the phases according to the new ordering?

dkachuma (approved):

src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseFVMKernels.hpp
klevzoff: As designed, fluid model owns its output quantities and derivatives, and always computes the derivatives w.r.t. the input variables of the fluid model interface, i.e. `z_c`. Thus the model is independent of the specific variable formulation implemented in a solver (as there could be multiple solvers using the same fluid model). Applying the chain rule in-place (i.e. overwriting `drho_l/dz_c` with `drho_l/drho_c` in memory) would be seen as an anti-pattern in this design as the solver shouldn't overwrite data it doesn't "own", while allocating and storing a separate copy of `drho_l/drho_c` would be prohibitively expensive in terms of memory footprint (once this is done for all quantities). Hence we apply the chain rule on the fly where needed, but suggestions are welcome.

src/coreComponents/constitutive/diffusion/ConstantDiffusion.hpp
klevzoff: We only stopped using `R1Tensor` for intermediate variables in kernels due to some concerns about how well they optimize IIRC. We do use `R1Tensor` for reading from input for things that are "vector"-like (e.g. geometric points, permeability components, gravity, etc). This has the benefit of better conveying the intent to the user (via rST/schema), as well automatic validation of input size. That said, if the input of full-tensor quantities is addressed, this may change and we might just make them all `array1d` to allow inputting a variable number of values (1, 3 or 6). So I'm not necessarily suggesting reverting the change, just bringing it up for context.

src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseFVMKernels.hpp
paveltomin: I am talking about solver code outside of fluid model. My expectation is that derivatives will be consistent for all quantities and there is no need to remember that specifically for this one we need to chain rule.

src/coreComponents/constitutive/diffusion/ConstantDiffusion.hpp
jafranc: Thanks, I wasn't sure at all I recall it correctly. My bad sorry @francoishamon

src/coreComponents/constitutive/diffusion/DiffusionBase.hpp
CusiniM: I am also not so sure that we want to use harmonic averaging for the diffusivity so maybe we will wan to do that.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -662,8 +703,9 @@ real64 CompositionalMultiphaseBase::updateFluidState( ObjectManagerBase & subReg
   updateRelPermModel( subRegion );
   updatePhaseMobility( subRegion );
   updateCapPressureModel( subRegion );
-  // note: for now, thermal conductivity is treated explicitly, so no update here
 
+  // note1: for now, thermal conductivity is treated explicitly, so no update here
+  // note2: for now, diffusion and dispersion are also treated explicitly
```
paveltomin, line 708: @jafranc @francoishamon is it even true?

PR #2680, opened 2023-09-14
Update CoupledSolver.hpp to log statistics for subsolvers

sohailwaziri:
Statistics for nonlinear subsolvers in sequential scheme are not reported in the current version of develop branch. Relevant code is added so that subsolvers' statistics are recorded and reported.

Rebaseline: https://github.com/GEOS-DEV/integratedTests/pull/41


paveltomin (approved):

castelletto1 (approved):

sohailwaziri:
Output For sequentially coupled (SC) and fully coupled schemes (FC):
FC_output
SC_output

Guotong-Ren (approved):

src/coreComponents/physicsSolvers/multiphysics/CoupledSolver.hpp
```diff
@@ -425,6 +438,11 @@ class CoupledSolver : public SolverBase
 
       if( isConverged )
       {
+        // Save Time step statistics for the subsolvers
+        forEachArgInTuple( m_solvers, [&]( auto & solver, auto )
+        {
+          solver->getSolverStatistics().saveTimeStepStatistics();
```
TotoGaz, line 444: This will most likely require a rebaseline. Can you double check?

paveltomin: This only affects screen output, should not cause any diffs in integratedTests

TotoGaz: You may be modifying a member
```
  registerGroup( groupKeyStruct::solverStatisticsString(), &m_solverStatistics );
```
that may be written as well (and compared). But I confess I really don't know.

@CusiniM Should we proceed? I'm next in line so I'm fine fixing it if it appears something's needed.

CusiniM: I really think that it will require a rebaseline. It's fine if you want to do a single one for both PRs though.

TotoGaz: OK then. How should we proceed?

CusiniM: well, if you don't mind dealing with the diffs introduced by this PR when you rebase for yours I think you can merge this one and then go ahead and work on the rebaseline.

PR #2695, opened 2023-09-21
Refactor fluid unit tests

dkachuma:
- Adds a utility function for testing numerical derivatives against analytical values
- Follow up from review of #2542

#2465


paveltomin (approved):
the changes look good, hope you will manage to fix the failing builds

src/coreComponents/constitutive/unitTests/testCompositionalProperties.cpp
```diff
@@ -96,33 +89,32 @@ class CompositionalPropertiesTestDataTestFixture : public ::testing::TestWithPar
...
-    fdDerivative = (currentMolarDensity - molarDensity) / dp;
-    checkDerivative( dMolarDensity_dp, fdDerivative, "Molar density right pressure derivative" );
+    internal::testNumericalDerivative(
+      pressure, dp, dMolarDensity_dp,
+      [this, & t=temperature, & zmf=composition]( real64 const p ) -> real64 {
+      return computeMolarDensity( p, t, zmf );
+    } );
+
     // -- Temperature derivative
     real64 const dt = 1.0e-6 * temperature;
```
jafranc, line 99: very minor, but can we take advantage of this PR to note `dT` rather than `dt` 
```suggestion
    real64 const dT = 1.0e-6 * temperature;
```

jafranc (approved):
Great work ! Thank you @dkachuma, it is so easier to read through now

Issue #2631, opened 2023-08-21
Failing ci tests in Debug mode


Algiane:
**Describe the bug**
The assertion that `numNodesInFace` is equal to `numNodes` seems erroneous. To my understanding, numNodes is the total number of nodes in mesh while `numNodesInFaces` is the number of nodes per face.
It seems that this assertion has been commented and uncommented multiple times since one year.

**To Reproduce**
1. Build geos in Debug mode
2. Run the `make test` command
3. See failure 

For example the `testDofManager` test fails.

**Expected behavior**
All test should pass in Debug mode (for me it fixes 28 from 30 failing tests).

**Platform (please complete the following information):**
- Machine: Mac M1 Max
- Compiler: Clang 14.0.0
- GEOSX Version: 068dcde698fb4a1a25d6f4ab08066de4e898a5b2

I would suggest to delete this assertion (I will attach a PR for that) and to add at least one github-ci config in Debug mode.

Algiane:
I don't have the rights to create new branches so please find attached a patch containing a fix proposal.
[0001-Remove-1-erroneous-assert-issue-2631.patch](https://github.com/GEOS-DEV/GEOS/files/12402527/0001-Remove-1-erroneous-assert-issue-2631.patch)

Best

klevzoff:
IIRC we discovered at some point that out CI always builds with `-DNDEBUG` which disables most assertions (even in `Debug` builds), and I don't think anyone has come up with a fix yet?

Algiane:
Hi,

Thanks for your help.

If I am not wrong:
  1. hdf5 exports the `NDEBUG` flag in its `HDF5_DEFINITIONS` CMake variable (https://cmake.org/cmake/help/latest/module/FindHDF5.html);
  2. then `conduit` (v0.8.2) registers the hdf5 library using the `blt_register_library` macro and export the content of `HDF5_DEFINITIONS`(see [conduit-0.8.2/src/cmake/thirdparty/SetupHDF5.cmake](https://github.com/LLNL/conduit/blob/72f58ef894cf627d00258e813a43752975e2c02b/src/cmake/thirdparty/SetupHDF5.cmake#L326-L329));
[...]

PR #2699, opened 2023-09-22
Remove an erroneous assertion  (#2631)

Algiane:
This PR:
  - removes an assertion who makes geos crash when it is built in Debug mode (see issue #2631).
   According to geos code and doc:
    - `numNodesInFace` is the number of nodes along the face of an element ([`getFaceNodes` method](https://github.com/GEOS-DEV/GEOS/blob/52354b4874c44d1132717a49732f2d9888d5241b/src/coreComponents/mesh/generators/CellBlock.hpp#L95-L100)),
    - `numNodes = lowestNodeToFaces.size();  (see [numNodes setting](https://github.com/GEOS-DEV/GEOS/blob/52354b4874c44d1132717a49732f2d9888d5241b/src/coreComponents/mesh/generators/CellBlockManager.cpp#L285))
    - `lowestNodeToFaces` is of size `m_numNodes` (that is the number of nodes on the partition) (see [lowestNodeToFaces allocation](https://github.com/GEOS-DEV/GEOS/blob/52354b4874c44d1132717a49732f2d9888d5241b/src/coreComponents/mesh/generators/CellBlockManager.cpp#L428C1-L428C1))
  - develops a little the associated code documentation.

Note that this erroneous assertion is not detected by the ci due to the transfer of the `NDEBUG` flag of the hdf5 build toward the geos compilation flags through Conduit (see TPL PR [#243](https://github.com/GEOS-DEV/thirdPartyLibs/pull/243) )

Closes issue #2631

### DIFF
--- a/src/coreComponents/mesh/generators/CellBlockManager.cpp
+++ b/src/coreComponents/mesh/generators/CellBlockManager.cpp
@@ -292,8 +292,7 @@ void populateFaceMaps( Group const & cellBlocks,
   GEOS_ERROR_IF_NE( faceToBlocks.size( 1 ), 2 );
 
   // loop over all the nodes.
-  forAll< parallelHostPolicy >( numNodes, [ numNodes,
-                                            uniqueFaceOffsets,
+  forAll< parallelHostPolicy >( numNodes, [ uniqueFaceOffsets,
                                             lowestNodeToFaces,
                                             faceToNodes,
                                             faceToCells,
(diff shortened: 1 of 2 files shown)


Algiane:
mmm, I don't understant because the Ubuntu-debug config was previously working...
The config that was failing was the Ubuntu-CUDA-debug (triggered only if all other configs are successful)...

Issue #2701, opened 2023-09-22
Compilation fails with NVCC on some PowerPC machines


sframba:
**Describe the bug**
Compilation fails on some PowerPC machines with `nvcc`. This is the case on the Pangea3 Power9 platform for exemple.
The cause is a compiler bug that causes the fmt library to malfunction. Specifically, if:
- Using fmt versions `10.0` and above;
- On a PowerPC machine (although it could happen on other platforms)
- WIth a compiler for which the following substitution in `fmt/include/core.h` fails: 
``typename Context::template formatter_type<T>().format( std::declval<const T&>(), std::declval<Context&>())``
then the `has_const_formatter` check fails and the compilation stops.
The bug was inadvertently introduced by [PR 2449](https://github.com/GEOS-DEV/GEOS/pull/2449).

This `fmt` issue [is known](https://github.com/fmtlib/fmt/issues/2746), and a workaround was [introduced](https://github.com/fmtlib/fmt/commit/70de324aa801eaf52e94c402d526a3849797c620) but later [removed](https://github.com/fmtlib/fmt/commit/466e0650ec2d153d255a40ec230eb77d7f1c3334).

**To Reproduce**
Take the current `develop` branch of `GEOS` (as of 09/22/2023) and compile. The compilation fails at `coreComponents/functions/TableFunctions.cpp` line 69. `fmt` provides the following information:

``"Cannot format an argument. To make type T formattable provide a formatter<T> specialization"``

**Expected behavior**
Compilation succeeds.

**Platform (please complete the following information):**
Pangea3 and possibly other Power machines. GEOS latest version.

sframba:
Solved by [PR 2702](https://github.com/GEOS-DEV/GEOS/pull/2702)

sframba:
Compilation issue appears again, after PR [#2756](https://github.com/GEOS-DEV/GEOS/pull/2756).

PR #2702, opened 2023-09-22
Solve compilation issue with NVCC on some PowerPC machines 

sframba:
This PR solves [issue 2701](https://github.com/GEOS-DEV/GEOS/issues/2701), at least on the Pangea3 configuration.

The solve involves overriding the `has_const_formatter` test for the specific case of `DataContext` with format context `fmt::format_context`. This is the smallest override that can solve the issue.

There is no risk with this fix, since the `const` fomatter is defined in the same file, just above the workaround.
If at some point in the future the problem is solved at the `fmt` library level, this workaround can be removed at the same time as the library upgrade.

If similar issues arise due to `const` classes being passed to `GEOS_FMT`, the scope of this workaround can be widened accordingly, to encompass the new cases and avoid new compilation issues.


src/coreComponents/dataRepository/DataContext.hpp
```diff
@@ -223,4 +223,23 @@ struct GEOS_FMT_NS::formatter< geos::dataRepository::DataContext > : GEOS_FMT_NS
...
+// The standard definition of the has_const_formatter check of fmt fails due to a compiler bug, see the issue below: 
+// https://github.com/fmtlib/fmt/issues/2746
+// The workaround was originally implemented in fmt:
+// https://github.com/fmtlib/fmt/commit/70de324aa801eaf52e94c402d526a3849797c620
+// but later removed: 
+// https://github.com/fmtlib/fmt/commit/466e0650ec2d153d255a40ec230eb77d7f1c3334
+// This workaround provides a specialization of the const formatter check for the DataContext object.
+// The formatter is defined within this file, and therefore the check is not needed.
+// The scope of the check override is as small as possible to solve the current issue.
+#ifdef GEOS_USE_CUDA
```
MelReyCG, line 236: As the compilation on Pecan GPU works, the problem doesn't seem to come from NVCC. It seems to come from the Power9 compilation target.
If no define are declared by the compiler, it would be possible to add a specific define for this workaround (`FMT_CONST_FORMATTER_WORKAROUND` ?)

src/coreComponents/dataRepository/DataContext.hpp
```diff
@@ -223,4 +223,23 @@ struct GEOS_FMT_NS::formatter< geos::dataRepository::DataContext > : GEOS_FMT_NS
   }
 };
 
+// The following workaround is needed to fix compilation with NVCC on some machines.
```
MelReyCG, line 226: I think the error message from FMT could be added in the comment so it will be easier if the error shows up again.

src/coreComponents/dataRepository/DataContext.hpp
sframba: I think the workaround tag is a good idea--it will allow to activate/deactivate it on a case-by-case basis.

src/coreComponents/dataRepository/DataContext.hpp
sframba: 👍

MelReyCG (approved):
For my part, I think this fix is well implemented.

TotoGaz (approved):

PR #2705, opened 2023-09-22
More fixes in SurfaceGenerator to allow for intersecting cracks

frankfeifan:
This PR fixes two new issues encountered in modeling intersecting cracks: 

1. A corner case that has one crack trailing face with two unpinched nodes intersecting with another existing fracture. For this crack face, it will enter the Sif calculation, but the existing criterion in the code cannot find the trailing node for its convex pinched node. The (temporary) fix is to skip the Sif calculation for this kind of trailing face. 
2. When selecting the right Sif for the nodes and faces, we need to check whether the node or the face has any Sif calculated, because we skipped the Sif calculation for some crack faces on intersection.


TotoGaz:
You're a hero dude!

CusiniM (approved):

frankfeifan:
I will also add an integrated test to ensure the capability will not be affected by new features.

rrsettgast (approved):

PR #2714, opened 2023-09-27
Minor docs updates

paveltomin:
- multiphaseFlowWithWells: point to `deadOilEgg_benchmark.xml` instead of `deadOilEgg_base_iterative.xml`, add url for GEOSXDATA
- co2Injection: correct `SimpleCo2InjTutorial.xml` into `simpleCo2InjTutorial.xml`
- thermalConsolidation: add `_fim` suffix for filename


src/docs/sphinx/basicExamples/multiphaseFlowWithWells/Example.rst
```diff
@@ -27,14 +27,14 @@ This example is based on the XML file located at
 
 .. code-block:: console
 
-  ../../../../../inputFiles/compositionalMultiphaseWell/benchmarks/Egg/deadOilEgg_base_iterative.xml
+  ../../../../../inputFiles/compositionalMultiphaseWell/benchmarks/Egg/deadOilEgg_benchmark.xml
```
jafranc, line 30:
```suggestion
  inputFiles/compositionalMultiphaseWell/benchmarks/Egg/deadOilEgg_benchmark.xml
```
By convention, I think getting rid of the relative path operators would be better

jafranc (approved):
Looks great ! thank you for the fixes @paveltomin

src/docs/sphinx/basicExamples/multiphaseFlowWithWells/Example.rst
paveltomin: I agree but those relative paths are everywhere in the docs. @herve-gross could you please confirm what is the right way?

jhuang2601 (approved):

PR #2721, opened 2023-09-28
add `NDEBUG` test - add `build_test` target - fix asserts

### LINKED ISSUE #2631: Failing ci tests in Debug mode

tbeltzun:
#  addition of a `NDEBUG` test

Add a test for the HDF5 propagated flags as explained in https://github.com/GEOS-DEV/GEOS/issues/2631 and  https://github.com/GEOS-DEV/GEOS/pull/2699 or https://github.com/GEOS-DEV/thirdPartyLibs/pull/248.

In `Debug` CMake configuration, if the `DNDEBUG` flag is propagated by any dependencies from the TPLs, this test should fail.

# addition of a `build_test` target 

It is well known (CMake limitation) that running `make test` doesn't build the tests and only tries to run them (see [here](https://stackoverflow.com/questions/733475/cmake-ctest-make-test-doesnt-build-tests)).

There is no dedicated target to enable building all the unitests executables at once apart from running `make all` (which I want to avoid because it builds targets that I don't need from the C++ side).

This PR add a `build_test` target such that running `make geosx build_test test` builds geosx, the unitests, and runs the test suite.

# remove erroneous assertion

Authored by @Algiane.

Description duplicated from https://github.com/GEOS-DEV/GEOS/pull/2699:
  - removes an assertion who makes geos crash when it is built in Debug mode (see issue #2631).
   According to geos code and doc:
    - `numNodesInFace` is the number of nodes along the face of an element ([`getFaceNodes` method](https://github.com/GEOS-DEV/GEOS/blob/52354b4874c44d1132717a49732f2d9888d5241b/src/coreComponents/mesh/generators/CellBlock.hpp#L95-L100));
    - `numNodes = lowestNodeToFaces.size();`  (see [numNodes setting](https://github.com/GEOS-DEV/GEOS/blob/52354b4874c44d1132717a49732f2d9888d5241b/src/coreComponents/mesh/generators/CellBlockManager.cpp#L285));
    - `lowestNodeToFaces` is of size `m_numNodes` (that is the number of nodes on the partition) (see [lowestNodeToFaces allocation](https://github.com/GEOS-DEV/GEOS/blob/52354b4874c44d1132717a49732f2d9888d5241b/src/coreComponents/mesh/generators/CellBlockManager.cpp#L428C1-L428C1)).
  - develops a little the associated code documentation.

~~Note that this erroneous assertion is not detected by the ci due to the transfer of the `NDEBUG` flag of the hdf5 build toward the geos compilation flags through Conduit (see TPL PR [#243](https://github.com/GEOS-DEV/thirdPartyLibs/pull/243) and Geos PR #2700 where the assertion is executed .)~~

Fixes #2700.  Closes #2631.

# enhance `CMakeLists.txt` consistency

- correct indentation and spaces (inconsistent within a file);
- replace `set(dependencyList ${dependencyList} geosx_core)` patterns with `list(APPEND dependencyList geosx_core)`, which is more concise;
- use `unset(something)` instead of `set(something)`, for clarity;
- remove tabs in `CMakeLists.txt`.


wrtobin (approved):
These both seem like sensible additions, and a good way to test against erroneous `NDEBUG` symbol propagation.

src/coreComponents/unitTests/toolchain/testNDEBUG.cpp
```diff
@@ -0,0 +1,29 @@
...
+   * This test guards against spurious propagation of -DNDEBUG preprocessor flag (HDF5 in our case),
+   * which has the effect of disabling LvArray assertions:
+   * instead, we always check that this test fails in CMake Debug mode, whilst it should always
+   * pass in RelWithDebInfo or Release builds.
+   */
+#ifdef NDEBUG
+  return 0;
+#else
+  return 1;
+#endif
```
TotoGaz, line 27: What is this test actually testing? I do not understand how you retrieve the value (`0` or `1`) and make use of it?
Also, could you use the `googletest` fwk to make it a unit test like the others?

scripts/ci_build_and_test_in_container.sh
```diff
@@ -80,11 +80,16 @@ if [[ "$*" == *--test-documentation* ]]; then
 fi
 
 # "Make" target check (builds geosx executable target only if true)
-# Use one process to prevent out-of-memory error
+# Reduce n° of processes to prevent out-of-memory error
+if [ ${CMAKE_BUILD_TYPE} == 'Debug' ]; then
+  np=2
+else
+  np=$(nproc)
+fi
```
TotoGaz, line 88: Do you think you could be more precise by using job pools https://cmake.org/cmake/help/latest/prop_gbl/JOB_POOLS.html

Also, is it for all `Debug` builds or only `CUDA` build (e.g.)?
Maybe you can configure the amount of cores used per build directly in the `yaml` files using the pools?

src/coreComponents/CMakeLists.txt
```diff
@@ -123,3 +123,29 @@ if( GEOS_ENABLE_TESTS )
...
+    get_property(current_targets DIRECTORY ${dir} PROPERTY BUILDSYSTEM_TARGETS)
+    foreach(target ${current_targets})
+      if(${target} MATCHES "(test|example).+")
+        list(APPEND ${targets} ${target})
+      endif()
+    endforeach()
+endmacro()
+
+get_all_test_targets(unittests_targets)
+add_custom_target(unittests COMMAND ${CMAKE_CTEST_COMMAND} DEPENDS ${unittests_targets})
```
TotoGaz, line 150: Can you better explain what you're trying to achieve here? I'm not sure to get it.

scripts/ci_build_and_test_in_container.sh
TotoGaz: We use `ninja` for all our CI builds.
Using the `JOB_POOLS` can provide a more precise way not to run too many processes in the same time.
If the problem is because there are too many RAM-consuming link editions (e.g.), why should we limit the compilation tasks?

src/coreComponents/unitTests/toolchain/testToolchain.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+#include <gtest/gtest.h>
+
+TEST( Toolchain, NDEBUGfromTPls )
+{
+  /*
+   * This test guards against spurious propagation of -DNDEBUG preprocessor flag (HDF5 in our case),
+   * which has the effect of disabling LvArray assertions:
+   * instead, we always check that this test fails in CMake Debug mode, whilst it should always
+   * pass in RelWithDebInfo or Release builds.
```
TotoGaz: For the moment on GHA, this test is OK whatever the `cmake` build type.
But the comments states that the test should be KO for debug builds.
Is it intended?

Also, a test is supposed to be OK, whatever the context.
If a test KO, how are you supposed to pass the CI?

src/coreComponents/CMakeLists.txt
TotoGaz: OK I better understand now.
Is it possible to add those comments into the `cmake` code so the direct intention is clear.
Also, would a name like `build_tests` (or `build_test` if we want to keep the symmetry with `test`) be clearer?

IIUC, the pattern is to walk through all the folders and find targets which look alike tests, based on naming conventions.
Would the best way to do this be to add a function `geos_add_test` which would call `blt_add_test` as we do now, but also add the test to the custom target `unittest`/`build_tests`.
This would make the registration explicit and safer?

Other than that, is it possible to retrieve the tests which are included in the `test` target?
If we could, then we would not have to look for those?

src/coreComponents/unitTests/toolchain/testToolchain.cpp
tbeltzun: > For the moment on GHA, this test is OK whatever the cmake build type.

No, the test passes only thanks to https://github.com/GEOS-DEV/thirdPartyLibs/pull/243.
If you point to an older TPL tag in this branch, the test fails as expected on GHA (see e.g. the checks for an older commit: https://github.com/GEOS-DEV/GEOS/actions/runs/6458645160/job/17532777983).

scripts/ci_build_and_test_in_container.sh
tbeltzun: I'll try to use `JOB_POOLS` (inspiration from https://www.scivision.dev/cmake-ninja-job-pool-limited-memory/).

The OOM failure occur at **compile time**, not **link time** (see e.g. https://github.com/GEOS-DEV/GEOS/actions/runs/6277002577/job/17137641986).

tbeltzun: Tentative fix in 397a3c5bf574d5c088dd070d64e13345a6fcc0e2.

src/coreComponents/CMakeLists.txt
tbeltzun: > Also, would a name like build_tests (or build_test if we want to keep the symmetry with test) be clearer?

`build_test` is probably a better naming yes.

> IIUC, the pattern is to walk through all the folders and find targets which look alike tests, based on naming conventions.

Exactly, a simple regex `(test|example).+`.

> Would the best way to do this be to add a function geos_add_test which would call blt_add_test as we do now, but also add the test to the custom target unittest/build_tests.
> This would make the registration explicit and safer?

This would probably be safer, at the expensive of adding another indirection level, and making the `CMake` process more complex to understand.

I've had a look into the implementation `blt_add_test` (https://github.com/LLNL/blt/blob/develop/cmake/BLTMacros.cmake#L397-L473), and wish there was a global list of test populated whenever `blt_add_test` is called: this feature doesn't seem to exist.

> Other than that, is it possible to retrieve the tests which are included in the test target?

I don't think this would be possible. Or maybe through a hack like `make help` which should list all targets, and then parsing this, which would not be bullet proof.

tbeltzun: Implemented in https://github.com/GEOS-DEV/GEOS/pull/2721/commits/c10f098948478fa7bd1c488cb3dc22236db9a878.

tbeltzun: Another downside to this approach is that `build_test` doesn't generate the executables for the submodules such as `LvArray` or `blt` anymore (w.r.t. using a regex).

src/coreComponents/unitTests/toolchain/testToolchain.cpp
TotoGaz: OK, I understand that on top of the test you have added a `cmake` rule that, depending on the build type (`Debug` or else), will analyze if we want the test to fail or not.

Could you consider gathering all the information needed in the test so it's clearer?

Would something like
```c++
if constexpr( std::string_view( GEOSX_CMAKE_BUILD_TYPE ) == std::string_view( "Debug" ) )
{
# ifdef NDEBUG
    SUCCEED();  // RelWithDebInfo or Release builds
# else
    FAIL();  // Debug builds
# endif
}
```
let you protect the behavior you want without having to add a specific configuration on `CMakeLists.txt`?
Or even this
```c++
if constexpr( std::string_view( GEOSX_CMAKE_BUILD_TYPE ) == std::string_view( "Debug" ) )
{
# ifdef NDEBUG
    bool constexpr hasNDEBUG = true;
# else
    bool constexpr hasNDEBUG = false;
# endif

  static_assert( hasNDEBUG, "NDEBUG not working as expected blah blah" );
}
```
which yells at compile time?

That would also let anyone call `testToolchain` without `ctest` and have the proper validation.
And there would be no need to get the information from two different locations to understand the complete pattern.

src/CMakeLists.txt
```diff
@@ -22,7 +22,13 @@ else()
 endif()
 set( CMAKE_ENABLE_EXPORTS ON )
 
-#set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE "${CMAKE_COMMAND} -E time")
+# ninja only: use `JOB_POOLS` to prevent OOM errors in CI (CUDA)
+set_property(GLOBAL PROPERTY JOB_POOLS one_jobs=1 two_jobs=2)
+
+# populated by `geos_add_test` (private list of all geos tests)
+set_property( GLOBAL PROPERTY geos_tests_list )
```
TotoGaz: In the `geos_add_test` macro I could see a `geos_tests_exe_list` list being used (actually at the end of this file too). But no `geos_tests_list`. I know that `cmake` can do crazy things, so I'm not sure. Is it intended?

src/docs/sphinx/developerGuide/Contributing/UnitTests.rst
```diff
@@ -29,12 +29,12 @@ What makes this such a good test is that it depends on very little other than ke
 
 MPI
 ---
-Often times it makes sense to write a unit test that is meant to be run with multiple MPI ranks. This can be accomplished by simply adding the ``NUM_MPI_TASKS`` parameter to ``blt_add_test`` in the CMake file. For example
+Often times it makes sense to write a unit test that is meant to be run with multiple MPI ranks. This can be accomplished by simply adding the ``NUM_MPI_TASKS`` parameter to ``geos_add_test`` in the CMake file. For example
```
TotoGaz, line 32: Great thx!

src/coreComponents/unitTests/wavePropagationTests/CMakeLists.txt
```diff
@@ -1,30 +1,23 @@
-#
 # Specify list of tests
-#
-
 set( gtest_geosx_tests
-	testWavePropagation.cpp
-        testWavePropagationAcousticFirstOrder.cpp
-   )
+	   testWavePropagation.cpp
```
TotoGaz: Warning, I think there's a tab here.

src/coreComponents/unitTests/toolchain/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+
+    geos_add_test( NAME ${test_name}
+                   COMMAND ${test_name} )
+
+    # NOTE: we explicitly depend on internal headers in order to check for the spurious addition of -DNDEBUG flag
+    target_include_directories( ${test_name} PUBLIC ${CMAKE_SOURCE_DIR}/coreComponents )
+
+  if(CMAKE_BUILD_TYPE STREQUAL "Debug")
+    set_tests_properties(${test_name} PROPERTIES WILL_FAIL TRUE)
+  endif()
```
TotoGaz: I think that we should try not to rely on this modifier.
Furthermore, if someone adds a test on line 2, that person will surely spend time understanding that there's a `negate` automatically applied to all the tests during the loop.

src/CMakeLists.txt
```diff
@@ -22,7 +22,13 @@ else()
 endif()
 set( CMAKE_ENABLE_EXPORTS ON )
 
-#set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE "${CMAKE_COMMAND} -E time")
+# ninja only: use `JOB_POOLS` to prevent OOM errors in CI (CUDA)
+set_property(GLOBAL PROPERTY JOB_POOLS one_jobs=1 two_jobs=2)
```
TotoGaz: Are you using this eventually or was this part of your pool experiment?

scripts/ci_build_and_test_in_container.sh
TotoGaz: I do not know what the best solution is about that...

Were you able to locate where/when the issue was?
Would using a job pool of `1` make sense?
Could it be possible to run the `ninja ...` in different steps, maybe triggering the compilation(s) unit(s) which is/are at stake first (like `ninja path/to/my_big_file.cpp.o path/to/my_other_big_file.cpp.o`) and then a std `ninja` with no proc restriction?

src/cmake/GeosxMacros.cmake
```diff
@@ -54,3 +54,36 @@ macro( geosx_add_code_checks )
     endif()
 
 endmacro( geosx_add_code_checks )
+
+##------------------------------------------------------------------------------
+## geos_add_test( NAME       [name]
+##                COMMAND    [command]
+##                EXECUTABLE [executable] )
+##
+## Adds a test to the project, remaining arguments are forwarded to `blt_add_test`
```
TotoGaz, line 63: Can you mention which global property you are using (`geos_tests_exe_list`) in the doc please?

TotoGaz (commented):
Thank you for your efforts

src/CMakeLists.txt
tbeltzun: Good catch, thanks. I prefer to initialize the global property here, although it doesn't seem to be needed in the end.

src/CMakeLists.txt
tbeltzun: This should have been removed by https://github.com/GEOS-DEV/GEOS/pull/2721/commits/b9ca046ef21c533a82681c873173b54c14f30f61. Using `JOB_POOLS` doesn't work at the moment (doesn't fix CUDA builds). ~~Left eventually for another PR~~. Cleanup done in https://github.com/GEOS-DEV/GEOS/pull/2721/commits/ef8a19facb6e1b1108eac768a7ab63f5f034db79.

EDIT: Fixed by using a larger runner for CUDA Debug as mentioned in https://github.com/GEOS-DEV/GEOS/pull/2721#issuecomment-1812504278.

scripts/ci_build_and_test_in_container.sh
tbeltzun: Uh, this is a lot of time consuming test to make. CUDA compiler crashing in `Debug` mode because running out of memory is abnormal. ~~Left as an improvement in another PR.~~

EDIT: Fixed by using `Runner_8core_32GB` as mentioned in https://github.com/GEOS-DEV/GEOS/pull/2721#issuecomment-1812504278.

src/cmake/GeosxMacros.cmake
tbeltzun: Done in 7fcf73e8332636a2dd1eeaee4a94021a64951b80.

src/coreComponents/unitTests/wavePropagationTests/CMakeLists.txt
tbeltzun: Nice catch, fixed.

src/coreComponents/unitTests/toolchain/CMakeLists.txt
tbeltzun: I understand that negating the logic can seem odd. However, I believe that it is the only way to check for false positives as it was the case before (`Debug` configuration is and has been untested for a long time). This check will prevent future regressions.

src/coreComponents/unitTests/toolchain/testToolchain.cpp
tbeltzun: I think that it increses the complexity and worsens readability.

I don't understand requiring further changes.
I've conformed to using the `gtest` framework as requested 3 weeks ago (https://github.com/GEOS-DEV/GEOS/pull/2721#discussion_r1368932681): this is exactly what I do using `SUCCEED` or `FAIL`, together with a standard `main` and I believe this test is compact and readable.

tbeltzun:
Thanks @TotoGaz for the thorough review.
I think I have addressed all the comments (please comment if you have more questions).
I'm toggling the `ready to be merged` flag to see if we pass CUDA jobs as it was the case after https://github.com/GEOS-DEV/GEOS/pull/2721/commits/f1860eff3fae3af3974b39149645e936c009966a.

tbeltzun:
Using `nproc=1` doesn't work anymore for `CUDA` in `Debug` mode, let's try to add some swap, this might have something to do with the docker process (when no `--memory` flag is passed, it should be limited only by the host memory).

tbeltzun:
It seems we can use alternative larger runners https://github.com/GEOS-DEV/GEOS/blob/c60be255c18de9b623e8b6c398a14d7444336831/.github/workflows/ci_tests.yml#L198

Ref: https://docs.github.com/en/actions/using-github-hosted-runners/about-larger-runners/about-larger-runners#machine-sizes-for-larger-runners.

Trying a larger memory runner now (`Runner_8core_32GB` and running on a lower number of cores).

tbeltzun:
@Algiane, now that CI seems to go forward (OOM issue fixed), we have a new issue arising from `silo` https://github.com/GEOS-DEV/GEOS/actions/runs/6878226393/job/18707464236?pr=2721#step:6:6863:
```
/usr/bin/ld: /opt/GEOS/GEOS_TPL-248-88-0f0b026/silo/lib/libsiloh5.a(alloc.o): relocation R_X86_64_32S against `.bss' can not be used when making a shared object; recompile with -fPIC
```

~~`-D CMAKE_POSITION_INDEPENDENT_CODE=${CMAKE_POSITION_INDEPENDENT_CODE}` seems to be propagated to `silo`, but is set to `ON / 1` ?~~

EDIT: this was a red herring, we still build silo with autogen, I got confused by HDF5.

EDIT: probably fixed by https://github.com/GEOS-DEV/thirdPartyLibs/pull/248/commits/907382bf0c6e7215d7c96120f9ee37c38037637d.

Algiane:
> I'm toggling the `ready to be merged` flag to see if we pass CUDA jobs as it was the case after [f1860ef](https://github.com/GEOS-DEV/GEOS/commit/f1860eff3fae3af3974b39149645e936c009966a).

I have this issue too: it would be convenient to have a "run only Cuda jobs" flag to trigger only the Cuda jobs and to try to fix them without wasting resources. I will try to propose this feature in another PR.

Algiane:
> It seems we can use alternative larger runners
> 
> https://github.com/GEOS-DEV/GEOS/blob/c60be255c18de9b623e8b6c398a14d7444336831/.github/workflows/ci_tests.yml#L198
> 
> Ref: https://docs.github.com/en/actions/using-github-hosted-runners/about-larger-runners/about-larger-runners#machine-sizes-for-larger-runners.
> 
> Trying a larger memory runner now (`Runner_8core_32GB` and running on a lower number of cores).

Thanks for your help an investigations: I am happy that you solved the compilation issue with Cuda.

Just as a short feedback, from what I have seen, adding assertions to the code causes ptxas to consume a huge amount of memory.
I have founded [this post](https://forums.developer.nvidia.com/t/nvcc-out-of-memory-ptxas-consumes-all-the-system-memory/18204/2) and [this one](https://forums.developer.nvidia.com/t/host-memory-eaten-up-during-compile-time/19786/2) that may be related.

I can reproduce this behaviour locally inside the Docker image built by the TPLs. I am not sure if it is related to the versions of Cuda / nvcc that are used or to the fact that we run inside a Docker container...

Anyway : I think that it is not linked to the TPLs or Geos...

tbeltzun:
Finally CI is green, nice work @Algiane .

@TotoGaz, we do not plan to further work on this PR, so we think it is ready (together with https://github.com/GEOS-DEV/thirdPartyLibs/pull/248) for a final round of review.
Thanks for your time.

Algiane (approved):

Algiane:
Thanks for all the work that you have done on this PR @tbeltzun ;-).

src/coreComponents/unitTests/toolchain/testToolchain.cpp
TotoGaz: > this test is compact and readable

No, this current test is not compact: it spans two files to be understood.
_Both_ `testToolChain.cpp` and the corresponding `CMakeLists.txt` are involved in the pattern.
(Plus the "negate" pattern that will negate all the tests in the folder, not only this one).

Is the code below equivalent to the full logic on the test?

```c++
bool constexpr isDebug = std::string_view( GEOSX_CMAKE_BUILD_TYPE ) == std::string_view( "Debug" );

#ifdef NDEBUG 
bool constexpr hasNDEBUG = true;
#else
bool constexpr hasNDEBUG = false;
#endif

ASSERT_EQ( !isDebug, hasNDEBUG );  // or static_assert(...)
```

scripts/ci_build_and_test.sh
```diff
@@ -15,23 +15,32 @@ BUILD_DIR=${GITHUB_WORKSPACE}
 # Since this information is repeated twice, we use a variable.
 BUILD_DIR_MOUNT_POINT=/tmp/GEOSX
 
+# We need to keep track of the building container (hence the `CONTAINER_NAME`)
+# so we can extract the data from it later (if needed). Another solution would have been to use a mount point,
+# but that would not have solved the problem for the TPLs (we would require extra action to copy them to the mount point).
+CONTAINER_NAME=geosx_build
+
+dargs=()  # docker arguments
```
TotoGaz: Nice! There was a strong rewriting of the ci scripts, but I didn't use bash arrays. I should have and I'll surely get inspired from your patch!

.github/workflows/ci_tests.yml
```diff
@@ -195,7 +195,7 @@ jobs:
         include:
           - name: Ubuntu CUDA debug (20.04, clang 10.0.0 + gcc 9.4.0, open-mpi 4.0.3, cuda-11.8.89)
             DOCKER_REPOSITORY: geosx/ubuntu20.04-clang10.0.0-cuda11.8.89
-            OS: Runner_4core_16GB
+            OS: Runner_8core_32GB
```
rrsettgast: why are you selecting the 8 core machine here?

rrsettgast (approved):

src/coreComponents/unitTests/toolchain/testToolchain.cpp
tbeltzun: > GEOSX_CMAKE_BUILD_TYPE

The downside of using this variable initialized by `CMake`  in [GeosxConfig.hpp.in](https://github.com/GEOS-DEV/GEOS/blob/2ada27d219211a0b5d452b734736f47f0adc131f/src/coreComponents/common/GeosxConfig.hpp.in#L114) is that it adds another level of complexity for one to understand this test.

> Plus the "negate" pattern that will negate all the tests in the folder, not only this one

This was only to conform to the existing tests structure, even if admittedly adding another test would require to change this pattern.

> Is the code below equivalent to the full logic on the test?

I will check this today locally and make sure the test fails using an older version of the TPLs.

.github/workflows/ci_tests.yml
tbeltzun: We had CI OOM errors on ubuntu + CUDA builds (and locally for @Algiane on a [local docker on MacOS](https://github.com/GEOS-DEV/GEOS/pull/2721#issuecomment-1814144127)) for the compilation of a single file ! See e.g. [this failed job](https://github.com/GEOS-DEV/GEOS/actions/runs/6863715326/job/18664043266#step:6:2489).
Switching to a 32Gb runner seems to work, and doesn't require workarounds like https://github.com/GEOS-DEV/GEOS/pull/2721#issuecomment-1812473253 or https://github.com/GEOS-DEV/GEOS/pull/2721#discussion_r1368943929 .

tbeltzun: Side note: we are still finalizing cosmetic changes (see https://github.com/GEOS-DEV/GEOS/pull/2721/files#r1396404470): I've added a `DO NOT MERGE !` label since I need the `ci: ready to be merged label` for running the CUDA jobs, I will remove when the PR is ready.

src/coreComponents/unitTests/toolchain/testToolchain.cpp
tbeltzun: I can confirm that the rewrite in https://github.com/GEOS-DEV/GEOS/pull/2721/commits/441527b985fab0e9e9dfb8ccdb3bb60c78e06f84 works as expected locally using https://github.com/GEOS-DEV/thirdPartyLibs/pull/248 (added test **passes** as expected in Debug), **and** the unpatched TPLs (added test **fails** as expected in Debug).

I'm checking this in CI now ...

Does these changes suit you ?

tbeltzun:
Ready on my side.

rrsettgast:
@TotoGaz Are you ok with this ?

TotoGaz (approved):
LGTM, thank you for your dedication and fruitful discussion @tbeltzun @Algiane !
I feel disappointed that we have to buy a bigger machine for the build, but I guess it is what it is, we'll have to figure how to improve the situation!

PR #2557, opened 2023-07-04
[WIP] acoustic - elastic coupling (SEM, 2ⁿᵈ order wave equation)

tbeltzun:
This PR is a WIP for the acoustic - elastic coupling in SEM for the second order wave equation.

PRs required:
- https://github.com/GEOS-DEV/GEOS/pull/2548: required for the coupling terms;
- https://github.com/GEOS-DEV/GEOS/pull/2642: high order ricker wavelets;
- https://github.com/GEOS-DEV/GEOS/pull/2641: `--output` cli flag fix;
- https://github.com/GEOS-DEV/GEOS/pull/2723;
- https://github.com/GEOS-DEV/GEOS/pull/2722;
- https://github.com/GEOS-DEV/GEOS/pull/2721;
- https://github.com/GEOS-DEV/GEOS/pull/2718;
- https://github.com/GEOS-DEV/GEOS/pull/2729;
- https://github.com/GEOS-DEV/GEOS/pull/2730.

Linked issues:
- https://github.com/GEOS-DEV/GEOS/issues/2592: high order solution visualization (nodes and elements);
- https://github.com/GEOS-DEV/GEOS/issues/166: a workaround using node sets is currently used.

- [x] normals signs
- [ ] test slanted interface & domain swap
- [x] validate against `gar6more3d`
- [ ] investigate
```
VTKOutput: Warning! 1 field name specified in `fieldNames` is/are not registered anywhere.
VTKOutput: Warning! `pressure_np1` is not registered anywhere and will not be displayed.
```
- [x] investigate vtk crash (duplicate of https://github.com/GEOS-DEV/GEOS/issues/1529#issuecomment-892818346)

PR #2722, opened 2023-09-28
support high order hex Q2 output in VTK

tbeltzun:
Q2 GLL nodes coincide with a regular grid, so handling Q2 nodes is un-ambiguous in VTK.

Credit for this PR goes to @Algiane .


klevzoff (approved):
LGTM

PR #2723, opened 2023-09-28
IO: filter mesh levels for VTK

tbeltzun:
Add the ability to filter mesh levels to avoid saving un-needed data (as e.g. in the case of high-order SEM visualization).

A rebaseline is probably needed since this PR changes the VTK output from our wave solvers.

Rebaseline: https://github.com/GEOS-DEV/integratedTests/pull/67


src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp
```diff
@@ -1156,14 +1156,24 @@ void VTKPolyDataWriterInterface::writeVtmFile( integer const cycle,
...
-      {
         return;
+
+      string const & meshLevelName = meshLevel.getName();
+
+      if( !m_levelNames.empty())
+      {
+        if( m_levelNames.find( meshLevelName ) == m_levelNames.end())
+          return;
       }
```
TotoGaz, line 1167:
```suggestion
      if( m_levelNames.find( meshLevelName ) == m_levelNames.end())
        return;
```

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp
```diff
@@ -1290,15 +1301,20 @@ void VTKPolyDataWriterInterface::write( real64 const time,
...
```
TotoGaz, line 1312:
```suggestion
      if( m_levelNames.find( meshLevelName ) == m_levelNames.end())
        return;
```

TotoGaz (approved):

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp
tbeltzun: I'm afraid that this changes the logic. If the array is empty we want to output all levels. 
`std::find` returns an iterator pointing to the end of the array when it is empty.
In this case we do **not** want an early return.

TotoGaz: Oh yes, you are right, sorry about that.
Can you then update the comments, descriptions and doxygens by adding something like "_Empty set means that all the level will be saved_" or something like that?
I'm not a huge fan of having `empty` means `all` because it means that the interface can't be easily used programmatically, but I guess we can live with this for this case here!

TotoGaz (approved):

TotoGaz:
Hi @tbeltzun there are conflict in the `rst` files. Can you fix those then we can merge your PR

src/coreComponents/schema/docs/VTK.rst
```diff
@@ -6,8 +6,13 @@ Name                        Type                    Default  Description
...
+<<<<<<< HEAD
+levelNames                  string_array            {}       Names of the mesh levels to output. By default (empty array) all levels are used.                                                                                                        
+name                        string                  required A name is required for any non-unique nodes                                                                                                                                              
+=======
 logLevel                    integer                 0        Log level                                                                                                                                                                                
 name                        groupName               required A name is required for any non-unique nodes                                                                                                                                              
+>>>>>>> develop
```
CusiniM: can you regenerate these rst and remove these conflict markers?

src/coreComponents/schema/schema.xsd
```diff
@@ -1981,8 +1981,13 @@ the relative residual norm satisfies:
 		<xsd:attribute name="fieldNames" type="groupNameRef_array" default="{}" />
 		<!--format => Output data format.  Valid options: ``binary``, ``ascii``-->
 		<xsd:attribute name="format" type="geos_vtk_VTKOutputMode" default="binary" />
+<<<<<<< HEAD
+		<!--levelNames => Names of mesh levels to output.-->
+		<xsd:attribute name="levelNames" type="string_array" default="{}" />
+=======
 		<!--logLevel => Log level-->
 		<xsd:attribute name="logLevel" type="integer" default="0" />
+>>>>>>> develop
```
CusiniM: same for the schema

tbeltzun: Thanks, done.

src/coreComponents/schema/docs/VTK.rst
tbeltzun: Thanks, done.

Issue #2720, opened 2023-09-28
CI: can't finalize PR


tbeltzun:
**Describe the bug**

What is this `${{ matrix.name }}` skipped job ? Does it correspond to `cuda_builds_merge_ready` ?

It seems to prevent the final check from suceeding:

![ci](https://github.com/GEOS-DEV/GEOS/assets/129868353/f01ac117-bd29-4be4-9ab6-38908e807ae0)

@acitrain pointed out that this might have something to do with https://github.com/GEOS-DEV/GEOS/pull/2659 (cc @bmhan12 ): but despite adding the `flag: ready to be merged` (which imho is misleading: why do we have to mark a PR `ready to be merged` before being `ready for review` and before CI passes ?!?), I can't seem to make CI green.

It also seems to be a waste of resources if we have to re-run all the jobs if we forgot to add the `flag: ready to be merged` label: restarting the failed jobs does **not** seem to trigger the CUDA builds.

PR #2726, opened 2023-09-28
Github Actions - Make CUDA jobs check and PR draft check dynamic

bmhan12:
This PR:
- Uses Github REST API to dynamically check for "ready for merge" PR label to run CUDA jobs.
  - Can now rerun workflows with skipped CUDA jobs to run CUDA tests.
- Uses Github REST API to dynamically check for PR draft status.
  - Can now rerun cancelled workflows with draft PR status.

Workflow with rerun working to trigger CUDA jobs: https://github.com/GEOS-DEV/GEOS/actions/runs/6342440554
(Can ignore the failed 2nd rerun attempt, 50X error from Docker Hub website)

Workflow with rerun working to for PR draft status: https://github.com/GEOS-DEV/GEOS/actions/runs/6410501990

Related to #2720

### DIFF
--- a/.github/workflows/ci_tests.yml
+++ b/.github/workflows/ci_tests.yml
@@ -23,9 +23,12 @@ jobs:
     steps:
     - name: Check that the PR is not a draft (cancel rest of jobs otherwise)
       run: |
-        echo "Is PR a draft?"
-        echo ${{ toJSON(github.event.pull_request.draft) }}
-        if [[ ${{ toJSON(github.event.pull_request.draft) }} == true ]]; then "false" ; else "true"; fi
+        draft_status=$(curl -H "Accept: application/vnd.github+json" \
+           https://api.github.com/repos/${{ github.repository }}/pulls/${{ github.event.number }} \
+           | jq '.draft')
+        echo "Is PR a draft? (Check PR's draft status )"
+        echo $draft_status
+        if [[ $draft_status == false ]]; then "true" ; else "false"; fi
 
   # PR must be assigned to be merged.
   # This job will fail if this is not the case.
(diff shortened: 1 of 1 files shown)


.github/workflows/ci_tests.yml
```diff
@@ -43,6 +43,23 @@ jobs:
...
+    steps:
+    - name: Check that the PR is ready to be merged
+      run: |
+        labels=$(curl -H "Accept: application/vnd.github+json" \
+           https://api.github.com/repos/${{ github.repository }}/pulls/${{ github.event.number }} \
+           | jq '.labels')
+        ready_label="ready to be merged"
+        echo "Is PR Ready to be Merged? (Check for $ready_label label)"
+        echo "Labels found are: $labels"
+        if [[ $labels == *"$ready_label"* ]]; then "true" ; else "false"; fi
```
TotoGaz, line 61: We were condensing our request as 
```
    exit $(curl -sS -H "Travis-API-Version: 3" \
           -X GET https://api.travis-ci.com/build/$TRAVIS_BUILD_ID/jobs | \
           jq '[ .jobs[] | select( (.stage.name == "non_blocking_failures") and (.allow_failure == true) and (.state != "passed")) ] | length')
```
Do you think we could use the same pattern?

TotoGaz: Also, what do you think of checking that the PR's `draft` the same way?

TotoGaz: > > Do you think we could use the same pattern?
> 
> I don't think so, as I am using the Github API instead of Travis's.

Couldn't we get the `draft` status for [this service](https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#get-a-pull-request)?

Issue #2633, opened 2023-08-24
`readthedocs` does not respect bullet points anymore and latex is broken


TotoGaz:
Our documentation on `readthedocs` does not respect the bullet points anymore. For example, the code
```
Supported Formats
=================

GEOS provides features to run simulations on unstructured meshes.
It uses VTK_ to read the external meshes and its API to write
it into the GEOS mesh data structure.

The supported mesh elements for volume elements consist of the following:

- 4-node tetrahedra,
- 5-node pyramids,
- 6-node wedges,
- 8-node hexahedra,
- n-gonal prisms (n = 7, ..., 11).

The mesh can be divided in several regions.
These regions are intended to support different physics
```
[generates](https://geosx-geosx.readthedocs-hosted.com/en/latest/coreComponents/mesh/docs/Mesh.html#supported-formats)

![image (7)](https://github.com/GEOS-DEV/GEOS/assets/49004943/c2556d8a-012c-4e7b-a5fd-37563be2a529)

This is true everywhere in the doc, not only for the example I provide.

sytuannguyen:
There is also an issue with the mathematical formulas as shown below:

![image](https://github.com/GEOS-DEV/GEOS/assets/45685596/9434dd00-9695-4e51-a930-ca35416d4728)

drmichaeltcvx:
Have you tried

```
* item 1
* item 2
* item 3
```

* item 1
* item 2
* item 3

or insert html directly

```
<ul>  
 <li>First item</li>  
 <li>Second item</li>  
 <li>Third item</li>  
 <li>Fourth item</li>
</ul>
```
<ul>  
 <li>First item</li>  
 <li>Second item</li>  
 <li>Third item</li>  
 <li>Fourth item</li>
</ul>

https://www.markdownguide.org/basic-syntax/

PR #2727, opened 2023-09-28
ReadTheDocs equations

bmhan12:
This PR:
- Attempts to fix readthedocs sphinx equations
  -  Looks to be somewhat related to installing pydata-sphinx-theme, when readthedocs theme is already present/targeted for use?

ReadTheDocs link to this PR: https://geosx-geosx--2727.com.readthedocs.build/en/2727/
Related to #2633

### DIFF
--- a/src/docs/sphinx/requirements.txt
+++ b/src/docs/sphinx/requirements.txt
@@ -8,4 +8,3 @@ pandas
 # using plantuml for diagrams
 sphinxcontrib-plantuml
 sphinx-argparse
-pydata-sphinx-theme


CusiniM (approved):
seems to fix it! Thank you so much

src/docs/sphinx/requirements.txt
```diff
@@ -8,4 +8,3 @@ pandas
 # using plantuml for diagrams
 sphinxcontrib-plantuml
 sphinx-argparse
-pydata-sphinx-theme
```
TotoGaz, line 11: Do you know who added this package and why?
We should checking that the feature it was allowing still works.

CusiniM: so, can this be merged?

bmhan12: > so, can this be merged?

I think so. 
This change should only affect sphinx on ReadTheDocs, and not sphinx on LC/local sphinx builds where pydata-sphinx-theme was inteded for use.

PR #2730, opened 2023-09-29
rewrite wave propagation damping kernels

tbeltzun:
Fix OOB error in the computation of damping kernels while developing the coupled acousto-elastic solver.

Loop over elements (then on face per element) instead of all domain faces to pre-compute the boundary damping terms.

Should be a noop regarding integrated tests (or maybe up to an numerical epsilon w.r.t. summation order).


sframba (approved):
Looks good to me

PR #2733, opened 2023-10-01
Update QuickStart.rst

Oluwatobi-PM:
The phrase "four processes" was changed to "four processors"

### DIFF
--- a/src/docs/sphinx/QuickStart.rst
+++ b/src/docs/sphinx/QuickStart.rst
@@ -344,7 +344,7 @@ An alternative is to set the path ``GEOSX_TPL_DIR`` via a cmake command line opt
    python scripts/config-build.py -hc host-configs/your-platform.cmake -bt Release -D GEOSX_TPL_DIR=/full/path/to/thirdPartyLibs
 
 We highly recommend using full paths, rather than relative paths, whenever possible.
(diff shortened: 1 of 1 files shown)


jhuang2601 (approved):

src/docs/sphinx/QuickStart.rst
```diff
@@ -344,7 +344,7 @@ An alternative is to set the path ``GEOSX_TPL_DIR`` via a cmake command line opt
    python scripts/config-build.py -hc host-configs/your-platform.cmake -bt Release -D GEOSX_TPL_DIR=/full/path/to/thirdPartyLibs
 
 We highly recommend using full paths, rather than relative paths, whenever possible.
-The parallel ``make -j 4`` will use four processes for compilation, which can substantially speed up the build if you have a multi-processor machine.
+The parallel ``make -j 4`` will use four processors for compilation, which can substantially speed up the build if you have a multi-processor machine.
```
TotoGaz, line 347: Hello, I'm not sure that `processors` is the correct wording here.
Processors can have multiple cores, and `ninja` will use several of them, but not the full processor.
Maybe `will simultaneously use four processes for compilation` is better?

paveltomin: @Oluwatobi-PM could you please reply to the comment above?

PR #2734, opened 2023-10-02
Fixed bug in VTKUtilities related to sortedArrayManipulation::find

francoishamon:
This PR fixes a bug that I introduced in PR #2569 

The bug was coming from a misunderstanding of how `sortedArrayManipulation::find` works. Say that we search for the value `myRank` in a `array1d` `donorRanks`. To detect that `myRank` is contained in `donorRanks`, the correct check is:
```
auto const myPosition =
    LvArray::sortedArrayManipulation::find( donorRanks.begin(), donorRanks.size(), myRank );
bool const isDonor = ( myPosition != donorRanks.size() ) && ( donorRanks[myPosition] == myRank );
```
and it is not enough to just check:
```
  auto const myPosition =
    LvArray::sortedArrayManipulation::find( donorRanks.begin(), donorRanks.size(), myRank );
  bool const isDonor = myPosition != donorRanks.size();
```

See for instance the [implementation](https://github.com/GEOS-DEV/LvArray/blob/86af5bfb5204d94dd7b1f4bcf98a37b8de9a630c/src/sortedArrayManipulation.hpp#L457) of `contains` in LvArray

Ready for review


src/coreComponents/mesh/generators/VTKUtilities.cpp
```diff
@@ -808,7 +808,7 @@ ensureNoEmptyRank( vtkSmartPointer< vtkDataSet > mesh,
   integer const myRank = MpiWrapper::commRank();
   auto const myPosition =
     LvArray::sortedArrayManipulation::find( donorRanks.begin(), donorRanks.size(), myRank );
-  bool const isDonor = myPosition != donorRanks.size();
+  bool const isDonor = ( myPosition != donorRanks.size() ) && ( donorRanks[myPosition] == myRank );
```
klevzoff, line 811: Hmm, I did miss it in the review, but should we even be using `sortedArrayManipulation::find()` which does a binary search on non-sorted data? I think linear search via `std::find` is appropriate here.

jafranc:
```suggestion
  auto const pos = std::find( donorRanks.begin(), donorRanks.end(), myRank);
  auto const myPosition = std::distance(donorRanks.begin(), pos);
   bool const isDonor = ( myPosition != donorRanks.size() ) && ( *pos == myRank );
```

then ?

klevzoff: No need to compare `*pos == myRank` since `std::find` looks for an exact match, not lower bound, and `pos` is not dereferencable when value is not found. More like
```
   auto const pos = std::find( donorRanks.begin(), donorRanks.end(), myRank);
   bool const isDonor = ( pos != donorRanks.end() );
   ...
   if( isDonor )
   {
     auto const myPosition = std::distance(donorRanks.begin(), pos);
     ...
   }
```

francoishamon: Done, thanks for the suggestion!

klevzoff (approved):

PR #2737, opened 2023-10-04
Output well rates to a simple csv file (in addition to the already implemented screen output)

paveltomin:
Simplify output logic: well solver log level decides if rates will be printed, not individual wellcontrols log level.

Example for multiphase:
```
time [s],bhp [Pa],total rate [kg/s],total reservoir volumetric rate [rm3/s],phase0 reservoir volumetric rate [rm3/s],phase1 reservoir volumetric rate [rm3/s]
10000,3.9e+07,-5.64728,-0.00673907,-0.00673178,-7.28919e-06
18688.6,3.9e+07,-7.77964,-0.0092858,-0.00928569,-1.07831e-07
30517.4,3.9e+07,-9.30733,-0.0111093,-0.0111093,-8.34709e-10
46705.8,3.9e+07,-10.2528,-0.0122378,-0.0122378,-3.83896e-12
67397.6,3.9e+07,-10.7294,-0.0128067,-0.0128067,-1.22283e-14
93818.7,3.9e+07,-10.9481,-0.0130677,-0.0130677,-2.73262e-17
122569,3.9e+07,-11.0276,-0.0131626,-0.0131626,-2.73262e-17
154433,3.9e+07,-11.0597,-0.0132009,-0.0132009,-4.50455e-22
```

Example for singlephase:
```
time [s],bhp [Pa],total rate [kg/s],total reservoir volumetric rate [rm3/s]
5000,593524,-0.00010003,-1e-07
10000,509522,-0.000100026,-1e-07
15000,500000,-8.90821e-05,-8.90597e-08
20000,500000,-8.25133e-05,-8.24925e-08
25000,500000,-7.82704e-05,-7.82507e-08
30000,500000,-7.50506e-05,-7.50317e-08
35000,500000,-7.2538e-05,-7.25198e-08
40000,500000,-7.05746e-05,-7.05569e-08
45000,500000,-6.90565e-05,-6.90391e-08
50000,500000,-6.78982e-05,-6.78812e-08
```

Rates are printed at every time step, which is also very common practice in reservoir simulators. Each well in a separate file.


src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
```diff
@@ -235,6 +235,23 @@ void CompositionalMultiphaseWell::registerDataOnMesh( Group & meshBodies )
...
+      {
+        string const wellControlsName = wellControls.getName();
+        string const massUnit = m_useMass ? "kg" : "mol";
+        integer const useSurfaceConditions = wellControls.useSurfaceConditions();
+        string const conditionKey = useSurfaceConditions ? "surface" : "reservoir";
+        string const unitKey = useSurfaceConditions ? "s" : "r";
+        integer const numPhase = m_numPhases;
+        // format: time,bhp,total_rate,total_vol_rate,phase0_vol_rate,phase1_vol_rate,...
+        std::ofstream outputFile( m_ratesOutputDir + "/" + wellControlsName + ".csv" );
+        outputFile << "time [s],bhp [Pa],total rate ["<<massUnit<<"/s],total " << conditionKey << " volumetric rate ["<<unitKey<<"m3/s]";
```
jafranc, line 250: Is this thread safe ?

src/coreComponents/physicsSolvers/fluidFlow/wells/WellSolverBase.cpp
```diff
@@ -69,6 +70,17 @@ WellSolverBase::~WellSolverBase() = default;
 void WellSolverBase::postProcessInput()
 {
   SolverBase::postProcessInput();
+
+  // create dir for rates output
+  if( getLogLevel() > 0 )
+  {
+    if( MpiWrapper::commRank() == 0 )
+    {
+      makeDirsForPath( m_ratesOutputDir );
```
jafranc, line 79: Just for discussion not as a review note, as we are C++17 compliant, is there any plan/advantage to use `std::filesystem::path` or specific path structure @TotoGaz @klevzoff  ?

jafranc (approved):
If no problem with the multirank writing in file, looks good to me. Thank you @paveltomin

paveltomin:
> If no problem with the multirank writing in file, looks good to me. Thank you @paveltomin

thank you
my understanding is that there will be unique rank writing into a file, other ranks will not write in the same file

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
paveltomin: my hope is the same - we are writing here separate file for each `wellControls`, so hopefully it's safe
would appreciate if @klevzoff can have a look and correct me if i am wrong

klevzoff: We don't parallelize across multiple wells (at least not yet), so it would even be safe to write to the same file

klevzoff: Looking at another comment by @jafranc, I guess this was about MPI/multiprocess safety. Yeah, you want only one rank to write data for each well. Regardless of how the well is partitioned across ranks, each rank has each well object (even if it has no locally present segments). Only one rank will be considered the nominal "owner" of the well and be responsible for assembling the control well equation, and it's usually the one that owns the well head IIRC. This is probably the rank you want writing well rates. This is determined via `subRegion.isLocallyOwned()`, and `printRates()` already checks that condition. You just need to check it here when writing the file header too.

src/coreComponents/physicsSolvers/fluidFlow/wells/WellSolverBase.cpp
klevzoff: We encourage using it for new code, but there is no concerted effort to convert all existing path-handling code

untereiner:
I understand your point on TimeHistory but this seems a bit hacky. 
Using `getLogLevel > 0` is a bit off-hand, you'll get it even if you don't ask for it.

I think a more viable solution is to finish the job in #2369 to be able to write everything into hdf5 files.
Then, you will be able to load your data frames in python directly from hdf5 files using h5py.

@wrtobin did you try the integrated tests of #2369  after the OS upgrade ?

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
paveltomin: added the check, that you

paveltomin:
> I understand your point on TimeHistory but this seems a bit hacky. Using `getLogLevel > 0` is a bit off-hand, you'll get it even if you don't ask for it.
> I think a more viable solution is to finish the job in #2369 to be able to write everything into hdf5 files. Then, you will be able to load your data frames in python directly from hdf5 files using h5py.

Just my opinion: using TimeHistory when manually editing xml is painful to output well rates. One has to add event for collection, event for output, task for collection, finally output for timehistory. I might be wrong here but seems that for each well  some of the listed before have to be repeated. Not user friendly at all, considering that in addition to that users have to heal with hdf5...

I can create a separate flag for rates output in the file, I just wanted to avoid doing rebaseline step which is another GEOS pain.

PR #2739, opened 2023-10-04
uncrustify: add spacing (assignments, operators, ...)

tbeltzun:
Stylistic code changes are controversial so I'll try to explain why I strongly believe that these changes are necessary.

There are currently sources of indentation inconsistencies even within a single source file (non exhaustive examples):
<details>

https://github.com/GEOS-DEV/GEOS/blob/fdb653369a4dec0c92bb873cf75cb6794f66d739/src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEMKernel.hpp#L508

https://github.com/GEOS-DEV/GEOS/blob/fdb653369a4dec0c92bb873cf75cb6794f66d739/src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.cpp#L846

https://github.com/GEOS-DEV/GEOS/blob/fdb653369a4dec0c92bb873cf75cb6794f66d739/src/coreComponents/finiteElement/elementFormulations/H1_Pyramid_Lagrange1_Gauss5.hpp#L500

https://github.com/GEOS-DEV/GEOS/blob/fdb653369a4dec0c92bb873cf75cb6794f66d739/src/coreComponents/physicsSolvers/wavePropagation/WaveSolverUtils.hpp#L49

</details>

I don't have a strong opinion on what style should be used and this is usually an endless source of debate (flame wars).
However, **being inconsistent in a code base** is imho even worse than having strict formatting rules.

1. To that purpose I claim that adding at least a space before and after operators and assignments improves readability and consistency, with the benefit of minimizing changes (w.r.t. the `force` uncrustify flag for example):

```c++
a=array[ 3*a+1 ];  // before
a = array[ 3 * a + 1 ];  // enforce spaces around operators and assignements
```

2. ~~Because of these added extra spacing within brackets, removing leading and trailing spaces in that case makes sense to me (also enhances consistency) `a = array[3 * a + 1];  // remove leading and trailing spaces`~~

3. ~~A last minor change tbd is allowing one liners for `if`s and `while`s: `c++` being a verbose language, saving vertical space and condensing code vertically is useful on common `16:9` or large aspect ratio screens:~~

```c++
if (cond)
  return;  // before

```c++
if (cond) return;  // after
````

### CODE REVIEW

**Comment by paveltomin** [comment]
Thank you, @tbeltzun, I support spacing enforcement.
One small request from my side, would it be possible also to enforce adding new line in situations like that:
```
      forAll< serialPolicy >( 1, [&numPhase,
                                  &useSurfaceConditions,
                                  &currentBHP,
                                  connRate,
                                  &currentTotalVolRate,
                                  currentPhaseVolRate,
                                  &iwelemRef, &wellControlsName,
                                  &massUnit] ( localIndex const )
```
I was expecting `&wellControlsName,` to be moved into next line after formatting but it does not.

CusiniM:
> Stylistic code changes are controversial so I'll try to explain why I strongly believe that these changes are necessary.
> 
> There are currently sources of indentation inconsistencies even within a single source file (non exhaustive examples):
> 
> https://github.com/GEOS-DEV/GEOS/blob/fdb653369a4dec0c92bb873cf75cb6794f66d739/src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEMKernel.hpp#L508
> 
> https://github.com/GEOS-DEV/GEOS/blob/fdb653369a4dec0c92bb873cf75cb6794f66d739/src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.cpp#L846
> 
> https://github.com/GEOS-DEV/GEOS/blob/fdb653369a4dec0c92bb873cf75cb6794f66d739/src/coreComponents/finiteElement/elementFormulations/H1_Pyramid_Lagrange1_Gauss5.hpp#L500
> 
> https://github.com/GEOS-DEV/GEOS/blob/fdb653369a4dec0c92bb873cf75cb6794f66d739/src/coreComponents/physicsSolvers/wavePropagation/WaveSolverUtils.hpp#L49
> 
> I don't have a strong opinion on what style should be used and this is usually an endless source of debate (flame wars). However, **being inconsistent in a code base** is imho even worse than having strict formatting rules.
> 

I  agree with this statement.

> To that purpose I claim that adding at least a space before and after operators and assignments improves readability and consistency, with the benefit of minimizing changes (w.r.t. the `force` uncrustify flag for example).
> 
Sounds good! 

> Because of these added extra spacing within brackets, removing leading and trailing spaces in that case makes sense to me (also enhances consistency).
> 

So, re you proposing to have:
`(a  + b)` instead of `( a + b )` ?  If that's what you are proposing I don't have a strong opinion about it but I agree that it would be nice to be consistent. Unless others have strong opinions about it I would pick either one or the other and apply it code wide.

> A last minor change tbd is allowing one liners for `if`s and `while`s: `c++` being a verbose language, saving vertical space and condensing code vertically is useful on common `16:9` or large aspect ratio screens.

I am not a huge fan of this. I personally prefer to always see the curly brackets following `if `statements on separate lines (event if it's a single line). It's not a very strong preference though and I would once again support trying to enforce consistency.

rrsettgast:
> Regarding the failing tests, I believe we shouldn't check formatting for submodules such as [LvArray](https://github.com/GEOS-DEV/LvArray) or [hdf5_interface](https://github.com/GEOS-DEV/hdf5_interface).
> 
> Those checks should be handled within these submodule repositories.

We need to setup independent CI for submodules :)

rrsettgast:
@tbeltzun Thanks for taking this on. Coincidently I have a modified `uncrusitfy.cfg` file in `shiva`. The spacing inconsistency was allowed because the expressions can get really long. Code formatting is a double edged sword sometimes. Also there are modifications to `uncrustify` that need to happen in some cases, and it is a really hard code to modify imo. For example, there needs to be a fix for some of the lambda indentation issues that have come up with new specifiers...and templated lambdas. That said, I think the spacing is an easy one to address. I think we just accept the "spaces" everywhere format and move on.

wrtobin (commented):
The "spaces everywhere" is typically my stylistic preference anyway, and as @CusiniM mentions consistency is preferable, so I'm fine with approving this once the "spaces everywhere" version is pushed.

untereiner (approved):

PR #2647, opened 2023-09-01
Verification of viscoplastic deformation with a relaxation problem

jhuang2601:
Following recent development of [viscoplastic solver](https://github.com/GEOS-DEV/GEOS/pull/2611),  this PR 

- [x] documents the analytical verification of viscoplastic deformation with relaxation test
image

- [x] adds a smoke test to the integratedTests
       https://github.com/GEOS-DEV/integratedTests/pull/42

PR #2741, opened 2023-10-05
Fix readthedocs build

rrsettgast:
https://github.com/GEOS-DEV/GEOS/pull/2647 broke the display formatting for our docs....need to undo whatever changes cause the issue.

- [ ] Fix the current build

### DIFF
--- a/src/docs/sphinx/advancedExamples/validationStudies/viscoplasticity/Index.rst
+++ b/src/docs/sphinx/advancedExamples/validationStudies/viscoplasticity/Index.rst
@@ -13,7 +13,6 @@ Viscoplasticity
    ViscoExtendedDruckerPrager/Example
    ModifiedCamClay/Example	
    ViscoModifiedCamClay/Example
-   RelaxationTest/Example		   
 
 
 


cssherman:
@rrsettgast - I don't see anything in #2647 that would break the RTD build.  Looking at the raw html, it seems like we are using the 'classic' backup:

```html
<script type="application/json" id="READTHEDOCS_DATA">{"ad_free": false, "api_host": "https://readthedocs.com", 
"builder": "sphinx", "canonical_url": null, "docroot": "/src/", "features": {"docsearch_disabled": false}, 
"global_analytics_code": "UA-17997319-2", "language": "en", "page": "index", "programming_language": 
"cpp", "project": "geosx-geosx", "proxied_api_host": "/_", "source_suffix": ".rst", "subprojects": {}, 
"theme": "classic", "user_analytics_code": "", "version": "latest"}</script>
```

In the conf.py file, we are likely triggering one of the import exceptions:

https://github.com/GEOS-DEV/GEOS/blob/develop/src/conf.py#L154-L176

Looking at https://github.com/GEOS-DEV/GEOS/blob/develop/src/docs/sphinx/requirements.txt#L1-L14 , we don't seem to include **sphinx-rtd-theme** .  They did warn us a while ago that they are cleaning up their build images, so I'm guessing that they aren't including this package by default in their build environment.

cssherman:
> **sphinx-rtd-theme**

And honestly, we should probably remove this (and maybe other) try/except block from the conf.py file.  I'd consider publishing the fall-back 'classic' theme a failure, and the ImportError exception **should** be caught by the CI.

cssherman:
@rrsettgast - I've tested manually including the theme install, and am still getting some odd results.  I also seem to struggle with this theme locally.  I'll create a PR that uses the pydata theme for you to look at.  In all of my testing, this theme is much more configurable and resistant to theme-related errors.

PR #2742, opened 2023-10-06
Adding rtd theme to doc requirements file

cssherman:


cssherman:
It seems like the VM's are using a fairly old version of sphinx and the themes:
Running Sphinx v5.3.0

src/conf.py
```diff
@@ -308,22 +250,19 @@
 
 # Additional stuff for the LaTeX preamble.
 latex_elements['preamble'] = '\\usepackage{amsmath}\n\\usepackage{amssymb}\n\\usepackage[retainorgcmds]{IEEEtrantools}\n'
-
-imgmath_image_format='svg'
+imgmath_image_format='png'
```
TotoGaz, line 253: I would have kept `svg`. Why did you change?

TotoGaz (approved):

src/conf.py
```diff
@@ -99,7 +99,7 @@
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
 extensions = [
-    'sphinx.ext.imgmath',
```
cssherman, line 102: @TotoGaz - in order for the light/dark theme to work properly with equations, I've turned off imgmath.  The built-in sphinx version seems to work just fine, tbh.

src/conf.py
cssherman: This is a residual setting that i missed when removing the imgmath extension

rrsettgast:
not a fan of this theme. the rtd theme still works for raja and all the other LLNL codes. can't we just figure out what is going wrong instead of changing themes?

cssherman:
@rrsettgast - Here's the latest version that is now building correctly: https://geosx-geosx.readthedocs-hosted.com/en/bugfix-sherman-fixdocbuild/   (Switching to/from pydata is just a couple of lines)

rrsettgast (approved):
Thanks!!

PR #1888, opened 2022-04-22
Use Gitlab CI for nightly testing

bmhan12:
This PR:

- Adds nightly testing of GEOSX using Gitlab CI.
  - Gitlab CI does not allow CI runs from a branch without an initial `.gitlab-ci.yml` in `develop `(and adding one directly to develop through Gitlab UI breaks mirroring...), so example output is viewable through a GEOSX fork and ~~test~~ nightlyTests repository.
    - Gitlab CI output from a GEOSX fork using similar `.gitlab-ci.yml` (difference is it updates a test repository instead of nightlyTests repository): https://lc.llnl.gov/gitlab/han12/GEOSX/-/pipelines
      - Site is only viewable if you have LC machine access to login
      - ~~Example test repository where nightlyTests output is written:  https://github.com/bmhan12/test~~
      - EDIT: Output is now written to nightlyTests repository: https://github.com/GEOSX/nightlyTests
- ~~Changes `.gitmodules` paths to absolute urls from relative paths (Gitlab CI had issues using relative paths)~~

Related PR:
GEOSX/nightlyTests#6

PR #2750, opened 2023-10-11
Simple LC build scripts for testing TPL PRs.

wrtobin:
Modified the TPL helper scripts.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -120,7 +120,7 @@ stages:
   stage: build
   variables:
     ALLOC_COMMAND: "salloc -N1 -ppdebug"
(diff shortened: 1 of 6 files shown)


wrtobin:
@bmhan12 Is there a compelling reason not to swap our git submodules to absolute URLs instead of the current relative URLs which require the set-url stuff at the top of the .gitlab-ci.yml?

bmhan12:
> @bmhan12 Is there a compelling reason not to swap our git submodules to absolute URLs instead of the current relative URLs which require the set-url stuff at the top of the .gitlab-ci.yml?

@TotoGaz pointed out in a previous discussion this would require https: https://github.com/GEOS-DEV/GEOS/pull/1888#discussion_r860237193

Blog post on merits/demerits: https://blog.tremily.us/posts/Relative_submodules/

PR #2751, opened 2023-10-11
Fix EDFM cases from multiphaseFlowFractures

paveltomin:
Looks like after https://github.com/GEOS-DEV/GEOS/pull/1925 EmbeddedSurfaceGenerator requires finite element discretization and the run crashes when finite volume is specified.
To fix that - modified the case by adding finite element discretization. Relaxed tolerance for one case to avoid convergence problems.


CusiniM:
are these not part of the `integratedTests` suite?

CusiniM (approved):

paveltomin:
> are these not part of the `integratedTests` suite?

for some reason no:
```
decks = [TestDeck(name='deadOil_fractureMatrixFlow_edfm_horizontalFrac', 
                  description='Multiphase flow EDFM horizontal frac.', 
                  partitions=((1, 1, 1), (2, 1, 1), (4, 1, 1)), 
                  restart_step=0, 
                  check_step=23,
                  restartcheck_params=RestartcheckParameters(**restartcheck_params)), 
        TestDeck(name='deadOil_fractureMatrixFlow_edfm_inclinedFrac',
                description='Multiphase flow EDFM inclined frac.', 
                partitions=((1, 1, 1), (2, 2, 1)), 
                restart_step=0, 
                check_step=29,
                restartcheck_params=RestartcheckParameters(**restartcheck_params))]
```
no sure why but please don't ask me add them :)

PR #2760, opened 2023-10-12
Activate `sccache` in the `CI/CD`

TotoGaz:
When the same job is triggered multiple times, `sccache` is able to download all the `.o` files from its cache for CPU builds.
It happens that `sccache` is experiencing hard time with our cuda build (eee the `Non-cacheable calls`):
```
Compile requests                    819
Compile requests executed            35
Cache hits                           35
Cache hits (C/C++)                   35
Cache misses                          0
Cache timeouts                        0
Cache read errors                     0
Forced recaches                       0
Cache write errors                    0
Compilation failures                  0
Cache errors                          0
Non-cacheable compilations            0
Non-cacheable calls                 784
Non-compilation calls                 0
Unsupported compiler calls            0
Average cache write               0.000 s
Average compiler                  0.000 s
Average cache read hit            0.144 s
Failed distributed compilations       0

Non-cacheable reasons:
multiple input files                784

Cache location                  gcs, name: geos-dev, prefix: /sccache/
Version (client)                0.5.4
```
Still, it's working good for CPU builds, and thus the PR is an improvement.

Relates to https://github.com/GEOS-DEV/thirdPartyLibs/pull/245


TotoGaz:
@bmhan12 @rrsettgast this is ready for review.
There's still something we need to understand for CUDA builds (see description of the PR), but it's working good for CPU builds.
I think it's still an improvement.

TotoGaz:
@bmhan12 @rrsettgast this is ready for review and merge. This could speed up some CPU targets first.

bmhan12 (approved):
Thanks @TotoGaz !

PR #2762, opened 2023-10-16
Add a few solver options plus some useful output for hydrofrac solver

paveltomin:
- option `minNormalizer` to change minimum normalizer value for residual evaluations
- add scaling and solution check for single phase flow solver, keeping old default behavior
- output aperture changes and min max values
- max pressure change output and number of negative values together with minimum value


src/coreComponents/physicsSolvers/NonlinearSolverParameters.hpp
```diff
@@ -272,6 +273,9 @@ class NonlinearSolverParameters : public dataRepository::Group
 
   /// Flag to specify whether subcycling is allowed or not in sequential schemes
   integer m_subcyclingOption;
+
+  /// Value used to make sure that residual normalizers are not too small when computing residual norm
+  real64 m_minNormalizer = 1e-12;
```
CusiniM, line 278: which residual was giving us issues?

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.cpp
```diff
@@ -58,6 +58,17 @@ SinglePhaseBase::SinglePhaseBase( const string & name,
...
     setDescription( "Temperature" );
+
+  this->registerWrapper( viewKeyStruct::maxPressureChangeString(), &m_maxPressureChange ).
+    setSizedFromParent( 0 ).
+    setInputFlag( InputFlags::OPTIONAL ).
+    setApplyDefaultValue( -1.0 ).   // disabled by default
+    setDescription( "Maximum (absolute) pressure change in a Newton iteration" );
+
+  this->registerWrapper( viewKeyStruct::allowNegativePressureString(), &m_allowNegativePressure ).
+    setApplyDefaultValue( 0 ). // by default not allowed
```
CusiniM, line 69: doesn't this happen quite often in hydrafrac simulations? Maybe we should have the hydrofrac solver change this value to 1 by default.

src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp
```diff
@@ -328,6 +345,21 @@ void HydrofractureSolver< POROMECHANICS_SOLVER >::updateDeformationForCoupling(
...
+  minHydraulicAperture  = MpiWrapper::min( minHydraulicAperture );
+  maxHydraulicAperture  = MpiWrapper::max( maxHydraulicAperture );
+
+  GEOS_LOG_LEVEL_RANK_0( 1, GEOS_FMT( "        {}: Max aperture change: {} m, max hydraulic aperture change: {} m",
+                                      this->getName(), fmt::format( "{:.{}f}", maxApertureChange, 6 ), fmt::format( "{:.{}f}", maxHydraulicApertureChange, 6 ) ) );
+  GEOS_LOG_LEVEL_RANK_0( 1, GEOS_FMT( "        {}: Min aperture: {} m, max aperture: {} m",
+                                      this->getName(), fmt::format( "{:.{}f}", minAperture, 6 ), fmt::format( "{:.{}f}", maxAperture, 6 ) ) );
+  GEOS_LOG_LEVEL_RANK_0( 1, GEOS_FMT( "        {}: Min hydraulic aperture: {} m, max hydraulic aperture: {} m",
+                                      this->getName(), fmt::format( "{:.{}f}", minHydraulicAperture, 6 ), fmt::format( "{:.{}f}", maxHydraulicAperture, 6 ) ) );
+
```
CusiniM, line 362: This output is useful! thank you

CusiniM (approved):

src/coreComponents/physicsSolvers/NonlinearSolverParameters.hpp
paveltomin: I am not 100% solid about this specific option (if this is the right solution) but what I observed for flow residual in hydrafracture cases - since the fracture volume is so tiny, the residual can go crazy because of normalization, so maybe better just to check the absolute value... more for experimenting for now.

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.cpp
paveltomin: yes it happens routinely there and hydrofrac solver does enable that flag https://github.com/GEOS-DEV/GEOS/blob/7025b62207f5dad87edd8bb8d0f03df2be4ef46e/src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp#L175, let me clarify the comment

paveltomin:
@rrsettgast @CusiniM this PR requires rebaselines because of the new flags in xml, what should we do now?

rrsettgast:
My mistake. There are run failures too. Can you fix those and I will merge it.

paveltomin:
> My mistake. There are run failures too. Can you fix those and I will merge it.

ok let me work on that...

PR #2765, opened 2023-10-17
DRAFT: CSV TimeHistory output

wrtobin:
Adding a basic ASCII CSV file output format for TimeHistory.

Each rank which actually collects data will write it's own file: `<filename>.<mpi_rank>.csv`

And each piece of data collected will have an individual file as well, to avoid having to totally rewrite the file at any point.

Very much WIP, needs tests/testing etc.


rrsettgast:
@wrtobin We should have a python post processor for this. Closing.

wrtobin:
Yeah this was mostly as an interim solution while the TH issues were worked on. I think we have enough mechanisms for I/O and the TH work is being staged in a branch so this shouldn't be necessary anymore.

PR #2767, opened 2023-10-18
feature/crook5/meshGeneration

cmcrook5:
Resolves conflicts with updates to develop branch regarding changes to mesh generation affecting particle mesh generators and managers.

This pull request contains work from [feature/povolny1/particleManager](https://github.com/GEOS-DEV/GEOS/pull/1756)

IntegratedTests: [feature/crook5/meshGeneration](https://github.com/GEOS-DEV/integratedTests/pull/51)


rrsettgast (approved):

src/coreComponents/mesh/generators/MeshGeneratorBase.cpp
```diff
@@ -47,15 +48,26 @@ MeshGeneratorBase::CatalogInterface::CatalogType & MeshGeneratorBase::getCatalog
...
+  }
+  else
+  {
+    CellBlockManager & cellBlockManager = parent.registerGroup< CellBlockManager >( keys::cellManager );
 
-  return cellBlockManager;
+    fillCellBlockManager( cellBlockManager, partition );
+
+    this->attachWellInfo( cellBlockManager );
+  }
```
TotoGaz, line 70: This part of the code should not we aware of neither `MeshBody` nor `MeshLevel`, as they basically depend on the whole `geos`.
It took us lots of efforts (@XL64 and myself) to slowly but surely remove these dependencies from the geometry world. Is it possible to keep this in the `MeshManager` part?

PR #2769, opened 2023-10-18
Create compositional multiphase fluid

dkachuma:
- Creates a native implementation of `CompositionalMultiphaseFluid`
- There is currently no implementation behind this to keep the PR smaller. The required values are currently hard coded constants. This PR focusses on the definition of the catalog entries that will face the user. The current purpose it to ensure that the mechanism is in place to be able to run this through the `PVTDriver` and check that the hard-coded values are output. Follow up PRs will link the actual static calculation kernels.
- Renames the implementation that depends on `PVTPackage` to `CompositionalMultiphaseFluidPVTPackage`
- Introduces 2 fluid types `CompositonalTwoPhaseFluidSoaveRedlichKwong` and `CompositonalTwoPhaseFluidPengRobinson`
- Removes the `equationsOfState` input. This is implied from the element tag.
- Adds `componentCriticalVolume` which is optional. When not supplied this is calculated using Ihmels' (2010) correlation.


src/coreComponents/constitutive/fluid/multifluid/compositional/CompositionalMultiphaseFluidUpdates.hpp
```diff
@@ -0,0 +1,431 @@
...
+  else
+  {
+    for( integer ic = 0; ic < numComp; ++ic )
+    {
+      compMoleFrac[ic] = composition[ic];
+    }
+  }
+
+  // 2. Compute phase fractions and phase component fractions
+  m_flash.compute( pressure,
```
paveltomin: it overall looks ok, following current GEOS practices for fluids
but i still do not fully understand why flash compute and phase computes can't become separate kernels?
would simplify the code imho and avoid zillions of template functions
@rrsettgast @klevzoff @francoishamon need you opinion please

paveltomin: maybe i could try to implement what i am talking about for a simple fluid to see how it would look...

joshua-white: Yes, a dummy design would be helpful.  For the moment I suggest we keep this PR moving forward so we can get baselines in the code, but we can always revisit the design.

src/coreComponents/constitutive/CMakeLists.txt
```diff
@@ -271,14 +286,11 @@ set( dependencyList ${parallelDeps} events dataRepository functions denseLinearA
 if( ENABLE_PVTPackage )
     set( constitutive_headers
          ${constitutive_headers}
-         fluid/multifluid/compositional/CompositionalMultiphaseFluid.hpp
-         )
+         fluid/multifluid/compositional/CompositionalMultiphaseFluidPVT.hpp )
```
joshua-white: Suggest we rename this to `CompositionalMultiphaseFluidPVTPackage` to be obvious what this is.

src/coreComponents/constitutive/fluid/multifluid/compositional/CompositionalMultiphaseFluid.cpp
```diff
@@ -66,45 +62,61 @@ CompositionalMultiphaseFluid::CompositionalMultiphaseFluid( string const & name,
     setDescription( "Table of binary interaction coefficients" );
 }
 
-integer CompositionalMultiphaseFluid::getWaterPhaseIndex() const
+template< typename FLASH, typename PHASE1, typename PHASE2, typename PHASE3 >
+integer CompositionalMultiphaseFluid< FLASH, PHASE1, PHASE2, PHASE3 >::getWaterPhaseIndex() const
 {
-  string const expectedWaterPhaseNames[] = { "water" };
-  return PVTProps::PVTFunctionHelpers::findName( m_phaseNames, expectedWaterPhaseNames, viewKeyStruct::phaseNamesString() );
+  return -1;
```
joshua-white: Is there a plan yet for how we handle the water phase index in the new design?

src/coreComponents/constitutive/fluid/multifluid/compositional/CompositionalMultiphaseFluid.cpp
```diff
@@ -121,81 +133,34 @@ void CompositionalMultiphaseFluid::postProcessInput()
...
 {
-  return KernelWrapper( *m_fluid,
-                        m_phaseTypes,
+  //auto phaseModels = std::make_tuple((m_phases)...);
+  return KernelWrapper( *m_componentProperties,
+                        *m_flash,
+                        *m_phase1,
+                        *m_phase2,
+                        *m_phase3,
                         m_componentMolarWeight,
```
joshua-white, line 166: Should this be grouped together with `m_componentProperties`?

src/coreComponents/constitutive/fluid/multifluid/compositional/CompositionalMultiphaseFluid.cpp
```diff
@@ -208,7 +173,71 @@ CompositionalMultiphaseFluid::createKernelWrapper()
...
+  m_componentProperties = std::make_unique< compositional::ComponentProperties >(
+    m_componentCriticalPressure,
+    m_componentCriticalTemperature,
+    m_componentCriticalVolume,
+    m_componentAcentricFactor,
+    m_componentVolumeShift,
+    m_componentBinaryCoeff );
+
+  m_flash = std::make_unique< FLASH >( getName() + '_' + FLASH::catalogName(),
+                                       m_componentNames,
```
joshua-white: Ditto.  Should this also be kept with `m_componentProperties`?

src/coreComponents/constitutive/fluid/multifluid/compositional/CompositionalMultiphaseFluidUpdates.hpp
```diff
@@ -0,0 +1,431 @@
...
+    m_phase3.density.compute( pressure,
+                              temperature,
+                              phaseCompFrac[2].toSliceConst(),
+                              phaseDens[2],
+                              phaseMassDens[2],
+                              m_useMass );
+  }
+
+  // 4. Calculate the phase viscosities
+  m_phase1.viscosity.compute( pressure,
```
joshua-white: The LBC viscosity model will require the phaseDensity as an argument as well.  Can put this now or wait until the LBC PR.

src/coreComponents/constitutive/fluid/multifluid/compositional/CompositionalMultiphaseFluidUpdates.hpp
```diff
@@ -0,0 +1,431 @@
...
+                                phaseVisc[2],
+                                m_useMass );
+  }
+
+  // 5. if mass variables used instead of molar, perform the conversion
+
+  if( m_useMass )
+  {
+
+    // unfortunately here, we have to copy the molecular weight coming from PVT package...
```
joshua-white: remove?  legacy comment.

src/coreComponents/constitutive/fluid/multifluid/compositional/CompositionalMultiphaseFluidUpdates.hpp
klevzoff: The original design was influenced primarily by `PVTPackage` which provided a single entry point for all calculations. This does not need to be the case any more.

However a secondary reason concerns performance, especially on GPUs. Split kernel design introduces additional latency and requires input data to be brought from DRAM into SM registers multiple times. Fused kernels are typically much more efficient in that sense and also allow intermediate computation results can be reused in-kernel without storing them into memory. This may be less of a concern if 90% of simulation time is taken by the linear solver.

In addition, our current approach to running these kernels is to have the solver do it rather than the model. This allows fluid compute to be fused into another kernel with a single level of model dispatch, as is the case with Dirichlet boundary conditions in `CompositionalMultiphaseFVM`, which avoids the need to allocate memory to store all fluid properties at those BC locations. I don't see an easy way to do this with split kernel approach, but I'm open to experiments.

src/coreComponents/constitutive/fluid/multifluid/compositional/CompositionalMultiphaseFluidUpdates.hpp
```diff
@@ -0,0 +1,431 @@
...
+    real64 phaseMolecularWeight[maxNumPhase]{};
+    for( integer ip = 0; ip < numPhase; ++ip )
+    {
+      phaseMolecularWeight[ip] = 1.0;
```
joshua-white, line 256: Is it okay that 1.0 is assigned here, not the phase MW?  Just checking.

src/coreComponents/constitutive/fluid/multifluid/compositional/CompositionalMultiphaseFluidUpdates.hpp
```diff
@@ -0,0 +1,431 @@
...
+  }
+
+  // 6. Compute total fluid mass/molar density
+
+  computeTotalDensity< maxNumComp, maxNumPhase >( phaseFrac,
+                                                  phaseDens,
+                                                  totalDens );
+}
+
+template< typename FLASH, typename PHASE1, typename PHASE2, typename PHASE3 >
```
joshua-white, line 273: I have never been a fan of this design where we have separate but nearly identical implementations, but with slightly different argument types.  I think this one (or perhaps the other one) only gets used in the hydrostatic initialization call?  Would be nice to remove one of them to avoid keeping redundant code aligned.  Also, I think the viscosity call has no impact when being used in the initialization call.

src/coreComponents/constitutive/fluid/multifluid/compositional/models/CompositionalDensity.hpp
```diff
@@ -0,0 +1,133 @@
...
+                                          real64 const & temperature,
+                                          arraySlice1d< real64 const, USD1 > const & phaseComposition,
+                                          real64 & molarDensity,
+                                          real64 & massDensity,
+                                          bool useMass ) const
+{
+  GEOS_UNUSED_VAR( pressure, temperature, useMass );
+  GEOS_UNUSED_VAR( phaseComposition );
+
+  massDensity = 1000.0;
```
joshua-white, line 100: Hardcoded values?

src/coreComponents/constitutive/fluid/multifluid/compositional/models/CompositionalDensity.hpp
```diff
@@ -0,0 +1,133 @@
...
+                                          real64 & molarDensity,
+                                          real64 & massDensity,
+                                          arraySlice1d< real64, USD3 > const & dMolarDensity,
+                                          arraySlice1d< real64, USD3 > const & dMassDensity,
+                                          bool useMass ) const
+{
+  GEOS_UNUSED_VAR( pressure, temperature, useMass );
+  GEOS_UNUSED_VAR( phaseComposition, dPhaseComposition );
+
+  massDensity = 1000.0;
```
joshua-white, line 120: Hardcoded values?

src/coreComponents/constitutive/fluid/multifluid/compositional/models/ConstantViscosity.hpp
```diff
@@ -0,0 +1,129 @@
...
+void ConstantViscosityUpdate::compute( real64 const & pressure,
+                                       real64 const & temperature,
+                                       arraySlice1d< real64 const, USD1 > const & phaseComposition,
+                                       real64 & viscosity,
+                                       bool useMass ) const
+{
+  GEOS_UNUSED_VAR( pressure, temperature, useMass );
+  GEOS_UNUSED_VAR( phaseComposition );
+
+  viscosity = 0.001;
```
joshua-white, line 100: Hardcoded value?

src/coreComponents/constitutive/fluid/multifluid/compositional/models/ConstantViscosity.hpp
```diff
@@ -0,0 +1,129 @@
...
+                                       arraySlice1d< real64 const, USD1 > const & phaseComposition,
+                                       arraySlice2d< real64 const, USD2 > const & dPhaseComposition,
+                                       real64 & viscosity,
+                                       arraySlice1d< real64, USD3 > const & dViscosity,
+                                       bool useMass ) const
+{
+  GEOS_UNUSED_VAR( pressure, temperature, useMass );
+  GEOS_UNUSED_VAR( phaseComposition, dPhaseComposition );
+
+  viscosity = 0.001;
```
joshua-white, line 121: Hardcoded value?

joshua-white (commented):
This design looks great to me.  I just see a few hardcoded placeholders for the static update functions to fill in.  It would also be nice to remove that redundant update version that is only used for initialization, but perhaps that is a separate PR.

src/coreComponents/constitutive/fluid/multifluid/compositional/CompositionalMultiphaseFluid.cpp
dkachuma: Currently I'm not sure how the `phaseNames` input fits into all this. I haven't put this in since my current implementations are 2-phase with liquid and vapour. Once I start doing implementations with water, this will have to be worked out.

src/coreComponents/constitutive/fluid/multifluid/compositional/CompositionalMultiphaseFluid.cpp
dkachuma: The `m_componentMolarWeight` and `m_componentNames` members are part of the base `MultiFluidBase` class. So initially I thought it wouldn't be a good idea to lump them in. But having thought about it, I don't see why not.

src/coreComponents/constitutive/fluid/multifluid/compositional/CompositionalMultiphaseFluidUpdates.hpp
dkachuma: I will add this now.

src/coreComponents/constitutive/fluid/multifluid/compositional/models/ConstantViscosity.hpp
dkachuma: The actual implementation in the static functions is not "linked" yet. So all values are hardcoded constants - to ensure that the PVTDriver runs and outputs these values. The idea was to show the design layout and then, in subsequent PRs, link to the actual implementation.

src/coreComponents/constitutive/fluid/multifluid/compositional/CompositionalMultiphaseFluidUpdates.hpp
dkachuma: Definitely not. All calculations are hard coded constants. Initial PR is to generate the object and kernels that will run. Subsequent PRs will link to actual calculation functions.

src/coreComponents/constitutive/fluid/multifluid/compositional/CompositionalMultiphaseFluidUpdates.hpp
dkachuma: I will presently carry on as is and welcome ideas on how to rework this.

src/coreComponents/constitutive/fluid/multifluid/compositional/CompositionalMultiphaseFluidUpdates.hpp
dkachuma: This looks like it needs to be a concerted effort across the different fluid model types. I will add a task to #2465 to follow up on this.

dkachuma:
Outstanding issues:

- Hard-coded constant values: Subsequent PRs will link the values to the actual static calculation functions.
- Removal of possibly redundant/duplicate interface: Has to be done across fluid models in a follow up task.

paveltomin (approved):

joshua-white (approved):

joshua-white:
I went ahead and added the "ready to be merged" flag.  I assume no rebaselines are required.

dkachuma:
> I went ahead and added the "ready to be merged" flag. I assume no rebaselines are required.

No. Shouldn't require rebaselines.

paveltomin:
> > I went ahead and added the "ready to be merged" flag. I assume no rebaselines are required.
> 
> No. Shouldn't require rebaselines.

most probably it does require rebaseline because you changed registerWrapper in some places and registered something in catalog?

dkachuma:
> most probably it does require rebaseline because you changed registerWrapper in some places and registered something in catalog?

But these are new types which shouldn't exist yet in any of the tests.

joshua-white:
> > most probably it does require rebaseline because you changed registerWrapper in some places and registered something in catalog?
> 
> But these are new types which shouldn't exist yet in any of the tests.

The situation will be obvious when we try to run the integrated tests :-)  @dkachuma, can you (have you) run on Quartz and Lassen?  We've had a few instances recently of PRs getting in without confirming tests.

dkachuma:
> The situation will be obvious when we try to run the integrated tests :-) @dkachuma, can you (have you) run on Quartz and Lassen? We've had a few instances recently of PRs getting in without confirming tests.

I have run the tests locally on pecan. But as of yet I don't have access to Quartz or Lassen.

castelletto1:
> > The situation will be obvious when we try to run the integrated tests :-) @dkachuma, can you (have you) run on Quartz and Lassen? We've had a few instances recently of PRs getting in without confirming tests.
> 
> I have run the tests locally on pecan. But as of yet I don't have access to Quartz or Lassen.

Tests pass on Quartz. New baselines are not needed.

PR #2777, opened 2023-10-24
Use streak as GitHub Actions runner for cuda CI

rrsettgast:


wrtobin:
@rrsettgast What is the current state of things on Streak? What compile times were we seeing for different `-j` arguments? Do you have a CPU replacement ETA (procurement impossible difficulty)

PR #2784, opened 2023-10-26
Define a single quadrature point for FaceElements

CusiniM:
Currently, there is no need to create a FE for `FaceElementsSubRegions` since we don't do any FE calculation on them. 

In the future, we may want to define 2D spaces for them but it's not easy since they can heterogeneous. 

In any case, this PR allows to run `solidMechanics` simulations on external meshes with conforming fractures. 

Rebaseline https://github.com/GEOS-DEV/integratedTests/pull/78


TotoGaz (approved):

src/coreComponents/mainInterface/ProblemManager.cpp
```diff
@@ -949,6 +949,20 @@ map< std::tuple< string, string, string, string >, localIndex > ProblemManager::
                   numQuadraturePointsInList = std::max( numQuadraturePointsInList, numQuadraturePoints );
                 } );
               } );
+              
+              // For now FaceElementSubRegions do not have a FE type associated with them. They don't need one for now and
+              // it would have to be a heterogeneous one coz they are usually heterogeneous subregions.
+              elemRegion.forElementSubRegions< FaceElementSubRegion >( [&]( auto & subRegion )
```
TotoGaz:
```suggestion
              elemRegion.forElementSubRegions< FaceElementSubRegion >( [&]( FaceElementSubRegion const & subRegion )
```

src/coreComponents/mainInterface/ProblemManager.cpp
```diff
@@ -913,7 +913,7 @@ map< std::tuple< string, string, string, string >, localIndex > ProblemManager::
 
             if( feDiscretization != nullptr )
             {
-              elemRegion.forElementSubRegions< CellElementSubRegion, FaceElementSubRegion >( [&]( auto & subRegion )
+              elemRegion.forElementSubRegions< CellElementSubRegion >( [&]( auto & subRegion )
```
TotoGaz, line 928:
```suggestion
              elemRegion.forElementSubRegions< CellElementSubRegion >( [&]( auto & CellElementSubRegion )
```
and you may be able to simplify into
```c++
fe = subRegion.registerWrapper< finiteElement::FiniteElementBase >( discretizationName, std::move( newFE ) ).
               setRestartFlags( dataRepository::RestartFlags::NO_WRITE ).reference();
```
and remove
```c++
using SUBREGION_TYPE = TYPEOFREF( subRegion );
```
?

TotoGaz (approved):

src/coreComponents/mainInterface/ProblemManager.cpp
CusiniM: we can but we will likely have to put it back coz this fix should be temporary. I just don't know what elemType is appropriate for the `FaceElements` since they are heterogeneous. And I don't think we want to allocate always the max number of quadrature points. Sounds like a huge waste.

TotoGaz: > we can but we will likely have to put it back coz this fix should be temporary. I just don't know what elemType is appropriate for the `FaceElements` since they are heterogeneous.

A lot of water will have passed under the bridge when we have a solution for that.
I think that nobody's working on that subject?

> And I don't think we want to allocate always the max number of quadrature points. Sounds like a huge waste.

Not such a huge waste since fractures have little number of elements w.r.t. 3d elements. 🤷

paveltomin: > Not such a huge waste since fractures have little number of elements w.r.t. 3d elements. 🤷

I would be really careful about that assumption, see examples coming out of Kinetix https://www.software.slb.com/products/kinetix, it could be a lot of fractures

TotoGaz: Do you have figures? I can only see pictures, I'd be interested in knowing a little more.
But in that case, we'll have to manage this. Still, I don't think  anyone's working on the subject 🤣 🤷

paveltomin: > Do you have figures? I can only see pictures, I'd be interested in knowing a little more. But in that case, we'll have to manage this. Still, I don't think anyone's working on the subject 🤣 🤷

check for example
https://onepetro.org/REE/article/26/03/888/516635/A-Hybrid-Embedded-Discrete-Fracture-Model-and-Dual?searchresult=1

PR #2792, opened 2023-11-01
Fixes for brine density computes, output internally constructed PVT tables for CO2 fluid into CSV files

paveltomin:
Two critical fixes:
- fix for pressure units in calculatePureWaterDensity (bug introduced in https://github.com/GEOS-DEV/GEOS/pull/2686), thanks @jhuang2601 
- fix for phase molar weight

Output activated through logLevel > 0

Examples:
image
image
image

Needed for users to check props correctness and consistency.

Rebaseline: https://github.com/GEOS-DEV/integratedTests/pull/68


src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/BrineEnthalpy.cpp
```diff
@@ -73,21 +73,34 @@ real64 michaelidesBrineEnthalpy( real64 const & T,
...
+                             array1d< real64 > const & enthalpies,
+                             bool const printTable )
 {
 
   localIndex const nTemperatures = tableCoords.nTemperatures();
 
+  std::ofstream table_file;
+  if( printTable )
+  {
+    table_file.open( "BrineEnthalpy.csv" );
```
dkachuma: Could we potentially have multiple fluids writing to the same file?
It would also be nice to capture the output directory specified by `-o outDir`. This way we can handle different models running in the same directory.

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/CO2Enthalpy.cpp
```diff
@@ -281,8 +291,12 @@ CO2Enthalpy::calculateCO2Enthalpy( PTTableCoordinates const & tableCoords,
     {
       real64 const TC = tableCoords.getTemperature( j );
       enthalpies[j*nPressures+i] = helmholtzCO2Enthalpy( TC, densities[j*nPressures+i] ) + referenceEnthalpy;
+      if( i == 0 && printTable )
```
dkachuma: Instead of mixing the calculation and the printing, would it be better to have a separate `print()` function on these models that then gets called by the encompassing fluid?

untereiner:
Hello @rrsettgast, @TotoGaz. Is the good old plain text file the way to go now to output things ?

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/EzrokhiBrineDensity.hpp
```diff
@@ -234,20 +238,42 @@ void EzrokhiBrineDensityUpdate::compute( real64 const & pressure,
...
   dValue[Deriv::dT] = dValueCoef * exponent_dTemperature + waterDensity_dTemperature * exponentPowered;
 
   // here, we multiply common part of derivatives by specific coefficients
+  real64 const dValue_dPhaseComp = dValueCoef * exponent_dPhaseComp;
   dValue[Deriv::dC+m_CO2Index] = dValue_dPhaseComp * phaseComposition[m_waterIndex] * dPhaseComposition[m_CO2Index][Deriv::dC+m_CO2Index];
   dValue[Deriv::dC+m_waterIndex] = dValue_dPhaseComp * ( -phaseComposition[m_CO2Index] ) * dPhaseComposition[m_waterIndex][Deriv::dC+m_waterIndex];
 
+  if( !useMass )
+  {
+    real64 const MT =
```
paveltomin: @klevzoff @joshua-white @dkachuma @jafranc could you please have a look if this way make sense?

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/PhillipsBrineDensity.hpp
```diff
@@ -243,37 +244,41 @@ void PhillipsBrineDensityUpdate::compute( real64 const & pressure,
...
-                        + dConc_dPres
-                        - dConcDensVol_dPres / m_componentMolarWeight[m_waterIndex];
-    dValue[Deriv::dT] = densityDeriv[1]  / m_componentMolarWeight[m_waterIndex]
-                        + dConc_dTemp
-                        - dConcDensVol_dTemp  / m_componentMolarWeight[m_waterIndex];
-    dValue[Deriv::dC+m_CO2Index] = dConc_dComp[m_CO2Index]
-                                   - dConcDensVol_dComp[m_CO2Index] / m_componentMolarWeight[m_waterIndex];
-    dValue[Deriv::dC+m_waterIndex] = dConc_dComp[m_waterIndex]
-                                     - dConcDensVol_dComp[m_waterIndex] / m_componentMolarWeight[m_waterIndex];
+    real64 const MT =
```
paveltomin: @klevzoff @joshua-white @dkachuma @jafranc also here

klevzoff: Looks right to me

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/BrineEnthalpy.cpp
```diff
@@ -73,21 +73,34 @@ real64 michaelidesBrineEnthalpy( real64 const & T,
 
 void calculateBrineEnthalpy( PTTableCoordinates const & tableCoords,
                              real64 const & m,
-                             array1d< real64 > const & enthalpies )
+                             array1d< real64 > const & enthalpies,
+                             bool const printTable )
```
klevzoff: Instead of interleaving table calculations with output, would it be feasible to have a single function
```
TableFunction::print( std::ostream & os, std::vector<string> const & columnNames )
```
that is called _after_ the table is populated?

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/BrineEnthalpy.cpp
```diff
@@ -73,21 +73,34 @@ real64 michaelidesBrineEnthalpy( real64 const & T,
...
 {
 
   localIndex const nTemperatures = tableCoords.nTemperatures();
 
+  std::ofstream table_file;
+  if( printTable )
+  {
+    table_file.open( "BrineEnthalpy.csv" );
+    table_file << "T[C],Enthalpy" << std::endl;
```
klevzoff: Consider using `\n` instead of `std::endl`, since there is no need to flush the stream after every line

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/BrineEnthalpy.cpp
paveltomin: considered that but honestly i have troubles dealing with table format and how values are stored there...

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/BrineEnthalpy.cpp
paveltomin: no one:

Sergey: no need to flush

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/EzrokhiBrineDensity.hpp
```diff
@@ -234,20 +238,42 @@ void EzrokhiBrineDensityUpdate::compute( real64 const & pressure,
...
+      dPhaseComposition[m_CO2Index][Deriv::dT] * m_componentMolarWeight[m_CO2Index];
+    dValue[Deriv::dT] = (dValue[Deriv::dT] - value * dMT_dTemp) / MT; // value is already divided by MT
+    real64 const dMT_dC_CO2 =
+      dPhaseComposition[m_waterIndex][Deriv::dC+m_CO2Index] * m_componentMolarWeight[m_waterIndex] +
+      dPhaseComposition[m_CO2Index][Deriv::dC+m_CO2Index] * m_componentMolarWeight[m_CO2Index];
+    dValue[Deriv::dC+m_CO2Index] = (dValue[Deriv::dC+m_CO2Index] - value * dMT_dC_CO2) / MT; // value is already divided by MT
+    real64 const dMT_dC_water =
+      dPhaseComposition[m_waterIndex][Deriv::dC+m_waterIndex] * m_componentMolarWeight[m_waterIndex] +
+      dPhaseComposition[m_CO2Index][Deriv::dC+m_waterIndex] * m_componentMolarWeight[m_CO2Index];
+    dValue[Deriv::dC+m_waterIndex] = (dValue[Deriv::dC+m_waterIndex] - value * dMT_dC_water) / MT; // value is already divided by MT
```
dkachuma: I would be in favour of replacing these with a loop.

jafranc: What about making this type of repetitive operations internal to the derivative struct ? (In an other PR I think)

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/PhillipsBrineDensity.hpp
jafranc: Can we think of a refactor (maybe left for later) in which the duplicated blocks are factored out in one place ? I start finding difficult to maintain that as little details will eventually diverged

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/PureWaterProperties.cpp
```diff
@@ -31,14 +31,16 @@ namespace PVTProps
...
 TableFunction const *
 PureWaterProperties::makeSaturationViscosityTable( string const & functionName,
-                                                   FunctionManager & functionManager )
+                                                   FunctionManager & functionManager,
+                                                   bool const printTable )
 {
   array1d< array1d< real64 > > temperatures;
   array1d< real64 > viscosities;
 
+  integer const nValues = 26;
```
jafranc, line 39: thanks :)

src/coreComponents/constitutive/fluid/multifluid/reactive/ReactiveFluidDriver.cpp
```diff
@@ -219,7 +219,7 @@ void ReactiveFluidDriver::runTest( FLUID_TYPE & fluid, arrayView2d< real64 > con
   }
 
   TableFunction const * waterDensityTable =
-    constitutive::PVTProps::PureWaterProperties::makeSaturationDensityTable( "helpTable", FunctionManager::getInstance() );
+    constitutive::PVTProps::PureWaterProperties::makeSaturationDensityTable( "helpTable", FunctionManager::getInstance(), false );
```
jafranc: Isn't it useful to output `PureWaterTable`. IMO, we might want to even if it is always the same and we now it is right. Just out of keeping consistency among the calls with higher `logLevel`

src/coreComponents/unitTests/constitutiveTests/testCO2BrinePVTModels.cpp
```diff
@@ -364,7 +364,8 @@ std::unique_ptr< MODEL > makePVTFunction( string const & filename,
       pvtFunction = std::make_unique< MODEL >( strs[1],
                                                strs,
                                                componentNames,
-                                               componentMolarWeight );
+                                               componentMolarWeight,
+                                               true ); // print PVT tables
```
jafranc, line 368: Do you then want to add test on the file generated ? Test that you can read back and compared to memory value ?

jafranc (approved):
Mix a bit fix and output concern but looks fine to me.
Thanks for the fix @paveltomin, looks great.
For the file output, I let architects decide if it is what it is wanted or not. Find it useful as you are.
If so, a follow up refactor trying @klevzoff suggestion of a unique `printer` sounds great.

src/coreComponents/unitTests/constitutiveTests/testCO2BrinePVTModels.cpp
paveltomin: not really to be honest

src/coreComponents/unitTests/fluidFlowTests/testThermalCompMultiphaseFlow.cpp
```diff
@@ -308,7 +308,7 @@ class ThermalCompositionalMultiphaseFlowTest : public ::testing::Test
 
   static real64 constexpr time = 0.0;
   static real64 constexpr dt = 1e4;
-  static real64 constexpr eps = std::numeric_limits< real64 >::epsilon();
+  static real64 constexpr eps = 10*std::numeric_limits< real64 >::epsilon();
```
paveltomin: @klevzoff @joshua-white please let me know if this is ok
i believe there is an accuracy issue with numerical derivatives here
the derivative is 0 and new code gives ~1e-19 while old code and numerical gives ~1e-10
checked the values in excel
image
trimmed accuracy gives ~1e-10 which probably confirms the accuracy issue?

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/PhillipsBrineDensity.hpp
paveltomin: moved common parts into PVTFunctionBaseUpdate

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/EzrokhiBrineDensity.hpp
paveltomin: did the loops and moved into PVTFunctionBaseUpdate

paveltomin:
@klevzoff @jafranc Same division by pure component molar weight here:
https://github.com/GEOS-DEV/GEOS/blob/c7add464015944d2deb86a652feb9b5e9508ba3d/src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/SpanWagnerCO2Density.hpp#L176
and here
https://github.com/GEOS-DEV/GEOS/blob/c7add464015944d2deb86a652feb9b5e9508ba3d/src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/CO2Enthalpy.hpp#L144
but CO2 phase is currently pure CO2 component, is it safe to keep it as is then or generalize? @dkachuma ?

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/BrineEnthalpy.cpp
paveltomin: ok, should be better in the latest version, thanks!

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/BrineEnthalpy.cpp
paveltomin: i changed and now file name contains fluid name etc, should be ok for now
running in the same directory might be an issue though

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/CO2Enthalpy.cpp
paveltomin: did that, thanks for good suggestion

src/coreComponents/functions/TableFunction.cpp
```diff
@@ -180,6 +180,49 @@ void TableFunction::checkCoord( real64 const coord, localIndex const dim ) const
...
+    os << "\n";
+    for( integer i = 0; i < N0; i++ )
+    {
+      os << coords0[i];
+      for( integer j = 0; j < N1; j++ )
+      {
+        os << "," << m_values[ j*N0 + i ];
+      }
+      os << "\n";
+    }
```
klevzoff: I see that you went for the 2D layout of the table, with coordinates spread across both rows and columns. It's a valid choice, but does not generalize above 2D. If we ever extend this to arbitrary dimensions (printing at least 3D tables might be useful), I would prefer if all tables followed the same format, e.g.
```
NameX [UnitX],NameY [UnitY],NameZ [UnitZ],NameV [UnitV]
x[0],y[0],z[0],v[0][0][0]
x[1],y[0],z[0],v[1][0][0]
x[2],y[0],z[0],v[2][0][0]
...
x[NX-1],y[0],z[0],v[NX-1][0][0]
x[0],y[1],z[0],v[0][1][0]
x[1],y[1],z[0],v[1][1][0]
...
x[NX-1],y[NY-1],z[NZ-1],v[NX-1][NY-1][NZ-1]
```
This is easy to consume in any tool (Excel, Matlab) and make plots since you can assign different columns to different axis. The order of coordinates/values doesn't really matter, but "Fortran" order feels most natural here.

paveltomin: ok, i kept that layout for 2d tables but for all other dimensions it's your way
hope it's ok

jhuang2601 (approved):
Thanks for fixing the bug!

src/coreComponents/codingUtilities/UnitTestUtilities.hpp
```diff
@@ -87,14 +87,14 @@ ::testing::AssertionResult checkRelativeErrorFormat( const char *, const char *,
 {
   real64 const delta = std::abs( v1 - v2 );
   real64 const value = std::max( std::abs( v1 ), std::abs( v2 ) );
-  if( delta > absTol && delta > relTol * value )
+  if( delta > absTol && delta > relTol * (value + 1.0) )
```
paveltomin, line 90: @joshua-white @klevzoff is it ok like that?

joshua-white: Do you need both conditions?  Personally I would drop the `absTol` one.  In the situation where `value` is very large in magnitude, a large magnitude `delta` may still be indicative of a good solution.  I'd like to hear @klevzoff's opinion though.

paveltomin: dropping `absTol` completely breaks `testMultiFluid` parts of which are currently not checking anything (and that was for a while, see https://github.com/GEOS-DEV/GEOS/pull/278), because the default absTol is a huge number and AND condition in the check

klevzoff: `absTol` was meant to allow us to ignore benign diffs below machine precision when the values are also small in magnitude (in this case `relTol` often fails). Not sure about the default value though, seems like it should've been 0, i.e. strict checking by default?

The proposed change can introduce false negatives for values << 1, for example
```
value = 1e-3
delta = 1e-8
relTol = 1e-6
absTol = 1e-14
```
would previously fail the test (`1e-8 > 1e-9 == true`), but will now pass (`1e-8 > 1.001e-6 == false`).

Maybe we just need more flexibility w.r.t. customizing tolerances in tests? I.e. instead of using the same `absTol` and `relTol` for all numerical derivative checking, we should be specifying them separately for pressure vs density derivs and separately for each property tested, on a per-model basis. Then we could use `absTol` to simply ignore spurious diffs like the one @paveltomin is dealing with. Does this seem implementable?

joshua-white: I think the following two conditions
```
delta > absTol
&&
delta > relTol*value
```
should be essentially equivalent to the single condition
```
delta > relTol * (value + absTol / relTol)
```
One is simply a smoothed version of the other.  Using @klevzoff's example values, we would pass the same test.  Instead choosing 1.0 for the last coefficient implies `absTol = relTol`.

So maybe either implementation is fine.  I think the core issue was that the chosen tolerances are stringent.  In the finite difference test, the derivative estimation is only O(h), where h is something like `sqrt(epsilon)`.  We currently have `relTol = epsilon`.  Perhaps we should switch to the centered difference approximation, which is O(h^2) and we can maybe keep stringent tolerances?

paveltomin: i just want to merge those bug fixes to be honest
not that much interested in dealing with unit tests

jhuang2601:
This PR went through the `ci: run integrated tests` and requires rebaseline. 
```
ATS END OF RUN STARTED December 08, 2023 18:42:45
    Status     : TestCase                                        : Directory                                                           : Elapsed : Resources : TestStep  
    ---------- : ----------------------------------------------- : ------------------------------------------------------------------- : ------- : --------- : ----------
    FAIL CHECK : staircase_co2_wells_3d_01                       : compositionalMultiphaseWell/staircase_co2_wells_3d_01               : 0:00:19 : 0:00:19   : restartcheck
    FAIL CHECK : class09_pb3_smoke_3d_01                         : compositionalMultiphaseWell/class09_pb3_smoke_3d_01                 : 0:00:09 : 0:00:09   : restartcheck
    FAIL CHECK : thermalLeakyWell_smoke_3d_01                    : compositionalMultiphaseFlow/thermalLeakyWell_smoke_3d_01            : 0:00:39 : 0:00:39   : restartcheck
    FAIL CHECK : PoroElastic_staircase_co2_3d_sequential_01      : poroElasticCoupling/PoroElastic_staircase_co2_3d_sequential_01      : 0:00:54 : 0:00:54   : restartcheck
    FAIL CHECK : co2_flux_3d_01                                  : compositionalMultiphaseFlow/co2_flux_3d_01                          : 0:00:10 : 0:00:10   : restartcheck
    FAIL CHECK : ThermoPoroElastic_staircase_co2_smoke_01        : thermoPoromechanics/ThermoPoroElastic_staircase_co2_smoke_01        : 0:00:45 : 0:00:45   : restartcheck
    FAIL CHECK : PoroElastic_hybridHexPrism_co2_fim_3d_01        : poroElasticCoupling/PoroElastic_hybridHexPrism_co2_fim_3d_01        : 0:08:21 : 0:08:21   : restartcheck
    FAIL CHECK : PoroElastic_permeableFault_smoke_01             : poroElasticCoupling/PoroElastic_permeableFault_smoke_01             : 0:00:05 : 0:00:05   : restartcheck
    FAIL CHECK : staircase_co2_wells_hybrid_3d_01                : compositionalMultiphaseWell/staircase_co2_wells_hybrid_3d_01        : 0:00:39 : 0:00:39   : restartcheck
    FAIL CHECK : co2_hybrid_1d_01                                : compositionalMultiphaseFlow/co2_hybrid_1d_01                        : 0:00:04 : 0:00:04   : restartcheck
    FAIL CHECK : PoroElastic_impermeableFault_smoke_01           : poroElasticCoupling/PoroElastic_impermeableFault_smoke_01           : 0:00:05 : 0:00:05   : restartcheck
    FAIL CHECK : co2_hybrid_1d_02                                : compositionalMultiphaseFlow/co2_hybrid_1d_02                        : 0:00:02 : 0:00:04   : restartcheck
    FAIL CHECK : PoroElastic_staircase_co2_3d_fim_01             : poroElasticCoupling/PoroElastic_staircase_co2_3d_fim_01             : 0:00:34 : 0:00:34   : restartcheck
    FAIL CHECK : co2_thermal_2d_01                               : thermalMultiphaseFlow/co2_thermal_2d_01                             : 0:00:05 : 0:00:05   : restartcheck
    FAIL CHECK : PoroElastic_hybridHexPrism_co2_sequential_3d_01 : poroElasticCoupling/PoroElastic_hybridHexPrism_co2_sequential_3d_01 : 0:07:30 : 0:07:30   : restartcheck
    ---------- : ----------------------------------------------- : ------------------------------------------------------------------- : ------- : --------- : ----------
```


Move it to the merge queue.

TotoGaz:
@jhuang2601 Thanks for you help. 😉 
Did you check that the discrepancies reported by the tool are legitimate? If you download the bundle (`integratedTests....tar.gz` and the end of the log), you can find a `test_result.html` that will let you investigate this.
Ping me, I can show you if you want!

paveltomin:
> @jhuang2601 Thanks for you help. 😉 Did you check that the discrepancies reported by the tool are legitimate? If you download the bundle (`integratedTests....tar.gz` and the end of the log), you can find a `test_result.html` that will let you investigate this. Ping me, I can show you if you want!

all co2 cases are expected to fail, need to prepare new baselines

PR #2796, opened 2023-11-02
Tutorial example for defining and importing faults and fractures with enhanced feature of  `mesh_doctor` 

jhuang2601:
This PR:

- [x] Review the document to run `mesh_doc` to generate the mesh for T-shape frac
- [x] Prepare a python script to handle the mesh generation to assign global IDs
- [x] Instead of using `InternalMesh`, rerun the existing [tutorial example](https://geosx-geosx.readthedocs-hosted.com/en/latest/docs/sphinx/advancedExamples/validationStudies/faultMechanics/intersectFrac/Example.html) with externally imported vtk mesh and check against the corresponding analytical solution
- [ ] Document the workflow and update the tutorial example


untereiner:
FYI https://github.com/GEOS-DEV/GEOS/pull/2896

rrsettgast:
@jhuang2601 Can we get this merged?

PR #2797, opened 2023-11-03
Remove unused variables

dkachuma:
Removes a couple of unused variables that cause compilation failures on some systems.


src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.cpp
```diff
@@ -1331,7 +1331,6 @@ bool SinglePhaseBase::checkSystemSolution( DomainPartition & domain,
       arrayView1d< globalIndex const > const dofNumber = subRegion.getReference< array1d< globalIndex > >( dofKey );
       arrayView1d< integer const > const ghostRank = subRegion.ghostRank();
       arrayView1d< real64 const > const pres = subRegion.getField< fields::flow::pressure >();
-      arrayView1d< real64 const > const mob = subRegion.getField< fields::flow::mobility >();
```
TotoGaz, line 1334: Do you need to unregister the field totally, or is the field used in other places?

TotoGaz (approved):

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.cpp
dkachuma: > Do you need to unregister the field totally, or is the field used in other places?

The field is used in some instances. Just not this one.

PR #2799, opened 2023-11-03
Use ComponentProperties in calculation functions

dkachuma:
- Uses the newly created ComponentProperties which bundles the component properties into one object instead of passing the list of properties independently.
- Changes `NegativeTwoPhaseFlash` to not be specific to the cubic EOS model.

#2465


src/coreComponents/constitutive/fluid/multifluid/compositional/functions/CubicEOSPhaseModel.hpp
```diff
@@ -44,6 +48,8 @@ struct PengRobinsonEOS
       ? 0.37464 + 1.54226 * omega - 0.26992 * omega * omega
       : 0.3796 + 1.485 * omega - 0.164423 * omega * omega + 0.016666 * omega * omega * omega;
   }
+
+  static constexpr char const * catalogName(){ return "PengRobinson"; }
```
paveltomin, line 52: I am not sure one needs `catalogName` for those things, could you help me to understand it here?

src/coreComponents/constitutive/fluid/multifluid/compositional/functions/KValueInitialization.hpp
```diff
@@ -20,13 +20,17 @@
 #define GEOS_CONSTITUTIVE_FLUID_MULTIFLUID_COMPOSITIONAL_FUNCTIONS_KVALUEINITIALIZATION_HPP_
 
 #include "common/DataTypes.hpp"
+#include "constitutive/fluid/multifluid/compositional/models/ComponentProperties.hpp"
```
paveltomin, line 23: did you forget to commit this file?

paveltomin: oh no, it's already there in the repo, sorry

paveltomin (approved):

src/coreComponents/constitutive/fluid/multifluid/compositional/functions/CubicEOSPhaseModel.hpp
dkachuma: There was a point to this. I believe I have a coupled of commits missing. Let me check my local branch.

dkachuma: Added one remaining commit. The idea is the negative two phase flash is no longer tied to the cubic eos. So the names are moved into the cubic eos objects themselves. These are used down the line in CompositionalMultiphaseFluid to give names to catalog entries.

src/coreComponents/constitutive/unitTests/testCompositionalProperties.cpp
```diff
@@ -195,11 +197,9 @@ class CompositionalPropertiesTestDataTestFixture : public ::testing::TestWithPar
   real64 computeMolarDensity( real64 const pressure, real64 const temperature,
                               arrayView1d< real64 const > const & composition ) const
   {
-    auto criticalPressure = m_fluid->getCriticalPressure();
-    auto criticalTemperature = m_fluid->getCriticalTemperature();
-    auto acentricFactor = m_fluid->getAcentricFactor();
-    auto volumeShift = m_fluid->getVolumeShift();
-    real64 const binaryInteractionCoefficients = 0.0;
+    auto const componentProperties = this->m_fluid->createKernelWrapper();
```
paveltomin, line 200: this `createKernelWrapper` looks a bit ambiguous, this is only in tests, right?

paveltomin (approved):

src/coreComponents/constitutive/unitTests/testCompositionalProperties.cpp
dkachuma: Yes it's not really a kernel wrapper. But it is for the unit test only.

PR #2804, opened 2023-11-06
Address the issue #2781: add a feature for visualize cells' global index

### LINKED ISSUE #2781: Export Global Cell Index out to VTK output
We would like to visualize the Global Cell index used in GEOS by printing these values out as a property on the VTK output.

This is helpful when importing output properties on an input mesh with a different indexing system.

There is a flag in the XML input file for this feature (`outputGlobalCellsIndex=true` or something equivalent) and the Global Cell index is part of the possible properties to visualize in the VTK output.

sytuannguyen:
This PR addresses the issue #2781 by adding a registerWrapper for cells' global index.

### DIFF
--- a/src/coreComponents/mesh/CellElementSubRegion.cpp
+++ b/src/coreComponents/mesh/CellElementSubRegion.cpp
@@ -297,6 +297,9 @@ void CellElementSubRegion::
   calculateElementCenterAndVolume( localIndex const k,
                                    arrayView2d< real64 const, nodes::REFERENCE_POSITION_USD > const & X ) const
 {
+  // TODO: Compute the global cell index
+  m_elementGlobalIndex[k] = 0;
+
   auto getElementCoordinatesaAndComputeElementCenter = [k, X, this]( auto & XLocal )
   {
     LvArray::tensorOps::fill< 3 >( m_elementCenter[k], 0 );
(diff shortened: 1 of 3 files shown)


untereiner:
is the global cell index not stored in `localToGlobalMap` (and `globalToLocalMap` if you want the reverse) ?

sytuannguyen:
> is the global cell index not stored in `localToGlobalMap` (and `globalToLocalMap` if you want the reverse) ?

Indeed, that function will be used to get the global indexes from local ones. Then we add a registerWrapper to output them for visualization as suggested in the issue #2781 by @herve-gross.

untereiner:
ok. but you can already use the `localToGlobalMap` array in a VTKOutput like:
```
    <VTK
      name="vtkOutput"
      fieldNames="{localToGlobalMap}"/>
```

 I am actually using it. What I am saying is that `elementGlobalIndex` will be an array that stores for each element it's global index, exactly what is stored in `localToGlobalMap`.

sytuannguyen:
> ok. but you can already use the `localToGlobalMap` array in a VTKOutput like:
> 
> ```
>     <VTK
>       name="vtkOutput"
>       fieldNames="{localToGlobalMap}"/>
> ```
> 
> I am actually using it. What I am saying is that `elementGlobalIndex` will be an array that stores for each element it's global index, exactly what is stored in `localToGlobalMap`.

I just tried your method, it works perfectly. @untereiner Thank you very much! @herve-gross The issue #2781 is solved then. Do we need a documentation to show this capability?

PR #2806, opened 2023-11-07
Fixing unused var

castelletto1:


paveltomin:
https://github.com/GEOS-DEV/GEOS/pull/2797

PR #2807, opened 2023-11-07
feat: Adding dispersion model for SPE11

jafranc:
Continuing the work initiated in #2678, the `LinearIsotropicDispersion` constitutive class init the `BaseDispersion`'s `m_dispersivity` cell-wise diagonal tensor field to the sole longitudinal dispersion coefficient.

In order to get a cell-centered vectorial velocity, we enrich `CellElementStencilTPFAWrapper` of an `initVelocity` and a `computeVelocity` functions as the geometry is easily accessible from this point. They are responsible for filling in a `m_phaseVelocity` fields held by `MultiFluidBase` (as for now).

The call site is located after phaseFlux computation in `isothermalCompositionalMultiphaseFVMKernel`, as the reconstruction relies on this values.

The reconstruction is done as a 3d linear interpolation of the normal of the faces scaled by the `phaseFlux` value. In order to do that, in `CellElementStencilTPFAWrapper::computeVelocity(...)`, after the flux values has been divided by the surface (which is recovered from the `halfWeight`), the `distanceToFace` is pulled from already existing vector and ratio-ed with the sum of the distances on this axis (which as been also memoized and pass as parameters).

On _deadoil_2ph_staircase_gravity_segregation_3d.xml_, this is per gas and water phase the kind of vector fields reconstructed
velocity_staircase

Step to get there:
- [x] Reconstruct a velocity vector field
- [x] Discuss the reconstruction algorithm
- [x] Opt. Refactor reconstruction in more appropriate place
- [x] Introduce velocity vector as a `field` 
- [x] Opt. Attach it to a more appropriate object than `MultiFluidBase`
- [x] Decide on the introduction derivatives or lagging the velocity field
- [x] Write a unit test for reconstruction
- [x] Test it in intergated Tests suite  

The implementation of `computeWeights()` taking phase indexed 4d tensor for dispersivity's transmissivity computation is discussed more in #2918.


src/coreComponents/finiteVolume/TwoPointFluxApproximation.cpp
```diff
@@ -184,10 +186,20 @@ void TwoPointFluxApproximation::computeCellStencil( MeshLevel & mesh ) const
...
+      //cumulating signed distance to from face to cell center to form denom in cell-wise linear interpolation
+      real64 absCellToFaceVec[3];
+        for (int dir = 0; dir < 3; ++dir) {
+            absCellToFaceVec[dir] = LvArray::math::abs(cellToFaceVec[ke][dir]);
+        }
+
+      LvArray::tensorOps::add<3>( m_globalCellToFace[stencilCellsGlobalIndex[ke]], absCellToFaceVec );
+
+//      real64 const c2fDistance = LvArray::tensorOps::normalize< 3 >( cellToFaceVec[ke] );
+        real64 const c2fDistance = LvArray::tensorOps::l2Norm< 3 >( cellToFaceVec[ke] );
```
jafranc: @francoishamon Not sure I broke something there, but I really needed the true `cellToFaceVec[ke]` here and not a normalized version that is returned by `LvArray::tensorOps::normalize< 3 >( cellToFaceVec[ke] )`. Then I changed it for `LvArray::tensorOps::l2Norm< 3 >( cellToFaceVec[ke] )` which IMO produce the distance too but left the input vector unchanged. Should I revert ?

paveltomin (approved):

src/coreComponents/unitTests/finiteVolumeTests/CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 # Specify list of tests
 set( gtest_geosx_tests
-     testMimeticInnerProducts.cpp )
+     testMimeticInnerProducts.cpp
```
CusiniM, line 3: we really need to stop adding `unitTests` to this folder.

src/coreComponents/finiteVolume/CellElementStencilTPFA.hpp
```diff
@@ -215,84 +272,204 @@ GEOS_HOST_DEVICE
...
+  }
+
+  averageWeights( iconn, weight, dWeight_dVar, halfWeight );
+
 }
 
+GEOS_HOST_DEVICE
+inline void
+CellElementStencilTPFAWrapper::
+  computeVelocity( localIndex iconn,
```
CusiniM: I don't love this function being defined in here. The stencil is also used for non advective fluxes (heat conduction for example) and this does not make much sense at that point. This function only uses the stencil to get the connectivity. I would make a free function that uses an instance of the object instead of a method in here.

paveltomin:
@jafranc is https://github.com/GEOS-DEV/GEOS/pull/2918 a hard requirement here?

PR #2811, opened 2023-11-09
Include water vapourisation in CO2/Brine flash

dkachuma:
Adds the possibility to have water vapourised in the gas phase for CO<sub>2</sub>-brine models.
- The vapourised water is expressed as a "solubility" which will be in a pre-calculated table similar to the gas solution in water.
- A two-phase-two-component "flash" is then carried out to determine the split of the phases.

This is a precursor to implementation of the Spycher & Pruess model.

At the moment, the water vapourisation table is not filled in - it is all zeros. This will be done in a follow up PR.

There are some numerical changes requiring a rebaseline. This is reflected in the changes to the unit tests in for the pure CO<sub>2</sub> composition. The numerical derivative test is now also limited by an absolute difference to avoid fails due to numerical noise for very small derivative values ~1e-8.

The implementation is based on the following:

- From the tables, we will read the solubility of CO<sub>2</sub> in moles of CO<sub>2</sub> per kg of water and also the amount of water vapourised in moles of water per kg of CO<sub>2</sub>.
- The solubility values are converted into mole/mole values by multiplying by the molecular weight. This gives the solubility of CO<sub>2</sub> in water $a_{c}(P,T)$ and vaporised water $a_{w}(P,T)$ as functions of pressure and temperature. 
- Suppose we have 1 mole of fluid mixture. This will be composed of $z_{co2}$ moles of CO<sub>2</sub> and $z_{wat}$ moles of water. Now let $x_{co2}$ be the number of moles of the CO<sub>2</sub> that are dissolved in the liquid phase and $y_{wat}$ the moles of water are vapourised in the gas phase. In such a case
```math
\begin{align}
x_{co2} &= a_c\left( z_{wat} - y_{wat} \right)  \\
y_{wat} &= a_w\left( z_{co2} - x_{co2} \right)
\end{align}
```  
or  
```math
\begin{bmatrix}
1 & -a_c\\ -a_w & 1
\end{bmatrix}
\begin{bmatrix}
x_{co2}\\y_{wat} 
\end{bmatrix}
=
\begin{bmatrix}
a_c z_{wat}\\a_w z_{co2}
\end{bmatrix}
```
 - This is solved for $x_{co2}$ and $y_{wat}$ provided $a_ca_w\neq1$.
 - The liquid and vapour fractions are then
 ```math
 \begin{align}
L &= x_{co2} + z_{wat} - y_{wat} \\
V &= y_{wat} + z_{co2} - x_{co2}
\end{align}
```

Rebasline https://github.com/GEOS-DEV/integratedTests/pull/79

### CODE REVIEW

File: src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/CO2Solubility.cpp [hunk 1]
```diff
@@ -262,6 +262,35 @@ TableFunction const * makeSolubilityTable( string_array const & inputParams,
...
+  // initialize the (p,T) coordinates
+  PTTableCoordinates tableCoords;
+  PVTFunctionHelpers::initializePropertyTable( inputParams, tableCoords );
+
+  // Currently initialise to all zeros
+
+  array1d< real64 > values( tableCoords.nPressures() * tableCoords.nTemperatures() );
+  values.zero();
+
+  string const tableName = functionName + "_waterVaporization_table";
```
jafranc, line 277: Maybe to be consistent, the name of the dissolution table should be altered from `_table` to `_co2Dissolution_table`

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/CO2Solubility.cpp
```diff
@@ -262,6 +262,35 @@ TableFunction const * makeSolubilityTable( string_array const & inputParams,
   }
 }
 
+TableFunction const * makeVapourisationTable( string_array const & inputParams,
+                                              string const & functionName,
+                                              FunctionManager & functionManager )
+{
```
jafranc, line 267: Maybe some TODO to be inserted there as this is where we want to check inputs

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/CO2Solubility.hpp
```diff
@@ -229,112 +241,169 @@ CO2SolubilityUpdate::compute( real64 const & pressure,
...
 
-  // Y = C/W = z/(1-z)
-  real64 Y = 0.0;
-  real64 dY_dCompFrac[2]{};
+  real64 const z_co2 = compFraction[m_CO2Index];
+  real64 const z_wat = compFraction[m_waterIndex];
 
-  if( compFraction[m_CO2Index] > 1.0 - minForDivision )
+  real64 const determinant = 1.0 - co2Solubility*watSolubility;
+  real64 invDeterminant = 0.0;
```
jafranc, line 193: Should we throw here if `determinant` is null ?

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/CO2Solubility.hpp
```diff
@@ -229,112 +241,169 @@ CO2SolubilityUpdate::compute( real64 const & pressure,
...
+      invDeterminantDeriv[ic] = invDeterminant*invDeterminant*(co2Solubility*watSolubilityDeriv[ic] + watSolubility*co2SolubilityDeriv[ic]);
+    }
   }
   else
   {
-    real64 const oneMinusCompFracInv = 1.0 / (1.0 - compFraction[m_CO2Index]);
-    Y = compFraction[m_CO2Index] * oneMinusCompFracInv;
-    dY_dCompFrac[m_CO2Index] = oneMinusCompFracInv * oneMinusCompFracInv;
-    dY_dCompFrac[m_waterIndex] = 0.0;
+    invDeterminant = 1.0 / minForDivision;
```
jafranc: Or better add a warning here as this is numerical tricks I guess

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/CO2Solubility.hpp
```diff
@@ -229,112 +241,169 @@ CO2SolubilityUpdate::compute( real64 const & pressure,
...
+    // Pressure and temperature derivatives
+    for( integer ic = 0; ic < 2; ic++ )
     {
-      phaseCompFraction.value[m_phaseLiquidIndex][ic] = compFraction[ic];
+      x_co2Deriv[ic] = co2SolubilityDeriv[ic] * (z_wat - watSolubility * z_co2) * invDeterminant
+                       - co2Solubility * watSolubilityDeriv[ic] * z_co2 * invDeterminant
+                       + co2Solubility * (z_wat - watSolubility * z_co2) * invDeterminantDeriv[ic];
+    }
+    // Composition derivatives
+    x_co2Deriv[2+m_CO2Index] = -co2Solubility * watSolubility * invDeterminant;
```
jafranc: if the `2` here because of `numComp` being constexpr hardcoded to 2 ? better use `numComp` then to help maintaining

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/CO2Solubility.hpp
```diff
@@ -229,112 +241,169 @@ CO2SolubilityUpdate::compute( real64 const & pressure,
...
+  {
+    x_co2 = 0.0;
+  }
 
-      for( localIndex jc = 0; jc < 2; ++jc )
-      {
-        phaseCompFraction.derivs[m_phaseLiquidIndex][ic][Deriv::dC+jc] = (ic == jc ) ? 1.0 : 0.0;
-        phaseCompFraction.derivs[m_phaseGasIndex][ic][Deriv::dC+jc] = 0.0;
-      }
+  real64 y_wat = watSolubility * (z_co2 - x_co2);
```
jafranc, line 222: IIUC this is now substitution, Idk but is there/will there be an argument (e.g. based on the fraction values) to solve the co2 and substitute the water or will solve co2 then substitute water be equivalent and then this way can be hardcoded as it makes no difference ?

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/CO2Solubility.hpp
```diff
@@ -229,112 +241,169 @@ CO2SolubilityUpdate::compute( real64 const & pressure,
...
   else
   {
-    // two-phase flow
+    y_wat = 0.0;
+  }
 
-    // 1) Compute phase fractions
+  // Liquid and vapour phase fractions
+  real64 const L = x_co2 + z_wat - y_wat;
+  real64 const V = y_wat + z_co2 - x_co2; // = 1 - L;
```
jafranc, line 242:
```suggestion
  real64 const V = 1 - L;// y_wat + z_co2 - x_co2;
```
Not for sure but this should be less numerical round-off sensible, isn't it ?

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/CO2Solubility.hpp
```diff
@@ -229,112 +241,169 @@ CO2SolubilityUpdate::compute( real64 const & pressure,
...
+    // Two phases
+
+    // 1) Compute phase fractions and derivatives
+
+    real64 const dL_dP = x_co2Deriv[0] - y_watDeriv[0];
+    real64 const dL_dT = x_co2Deriv[1] - y_watDeriv[1];
+    real64 const dL_dzco2 = x_co2Deriv[2+m_CO2Index] - y_watDeriv[2+m_CO2Index];
+    real64 const dL_dzwat = x_co2Deriv[2+m_waterIndex] + 1.0 - y_watDeriv[2+m_waterIndex];
+
+    real64 const dV_dP = y_watDeriv[0] - x_co2Deriv[0];
```
jafranc: Shouldn't we leverage also here the `V=1-L` rule ?

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/CO2Solubility.hpp
```diff
@@ -229,112 +241,169 @@ CO2SolubilityUpdate::compute( real64 const & pressure,
...
+    phaseFraction.derivs[m_phaseGasIndex][Deriv::dT] = dV_dT;
+    phaseFraction.derivs[m_phaseGasIndex][Deriv::dC+m_CO2Index] = dV_dzco2;
+    phaseFraction.derivs[m_phaseGasIndex][Deriv::dC+m_waterIndex] = dV_dzwat;
+
+    // 2) Compute phase component fractions and derivatives
+    real64 const invL2 = 1.0 / (L*L);
+    real64 const invV2 = 1.0 / (V*V);
+
+    phaseCompFraction.value[m_phaseLiquidIndex][m_CO2Index] = x_co2 / L;
+    phaseCompFraction.derivs[m_phaseLiquidIndex][m_CO2Index][Deriv::dP] = (x_co2Deriv[0]*L - x_co2*dL_dP)*invL2;
```
jafranc: Those hard-coded `0` and `1` would benefit from using similar struct as `Deriv::dP` to not get lost at some point

jafranc: Note for later, we could benefit from some third-party utilities working with `.val` `.derivs` struct to differentiate those long error-prone derivatives. A step toward AD, so it saves us a lots of lines here and there

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/CO2Solubility.hpp
```diff
@@ -229,112 +241,169 @@ CO2SolubilityUpdate::compute( real64 const & pressure,
...
+    LvArray::forValuesInSlice( phaseCompFraction.derivs, setZero );
 
-    phaseFraction.value[m_phaseGasIndex] = 1.0 - phaseFraction.value[m_phaseLiquidIndex];
+    // 1) Compute phase fractions
 
-    phaseFraction.derivs[m_phaseGasIndex][Deriv::dP] = -phaseFraction.derivs[m_phaseLiquidIndex][Deriv::dP];
-    phaseFraction.derivs[m_phaseGasIndex][Deriv::dT] = -phaseFraction.derivs[m_phaseLiquidIndex][Deriv::dT];
-    phaseFraction.derivs[m_phaseGasIndex][Deriv::dC+m_CO2Index] = -phaseFraction.derivs[m_phaseLiquidIndex][Deriv::dC+m_CO2Index];
-    phaseFraction.derivs[m_phaseGasIndex][Deriv::dC+m_waterIndex] = -phaseFraction.derivs[m_phaseLiquidIndex][Deriv::dC+m_waterIndex];
+    phaseFraction.value[activePhase] = 1.0;
```
jafranc, line 322: Shouldn't we set `phaseFraction.derivs` too ? or is it zero and then already set ?

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/CO2Solubility.hpp
```diff
@@ -229,112 +241,169 @@ CO2SolubilityUpdate::compute( real64 const & pressure,
...
-    // phaseCompFraction does not depend on globalComponentFraction
-
+    // Setup default values which will be overridden for the active phase
+    phaseCompFraction.value[m_phaseGasIndex][m_CO2Index] = 1.0;
+    phaseCompFraction.value[m_phaseLiquidIndex][m_waterIndex] = 1.0;
+    // Set the global composition as the composition of the active phase
+    for( integer ic = 0; ic < numComps; ++ic )
+    {
+      phaseCompFraction.value[activePhase][ic] = compFraction[ic];
+      phaseCompFraction.derivs[activePhase][ic][Deriv::dC+ic] = 1.0;
```
jafranc, line 332: Is this for sure ?

jafranc (approved):
Great work ! Thanks @dkachuma.
Looking forward to the following PRs

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/CO2Solubility.cpp
dkachuma: The inputs should be are checked in `PVTFunctionHelpers::initializePropertyTable`.

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/CO2Solubility.hpp
dkachuma: I still can't figure out why this doesn't make sense physically. In any case the values that we have a re unlikely to get us into this situation. So I will add a throw.

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/CO2Solubility.hpp
dkachuma: No. This is because they come in after `dP` and `dT`. I will change to consistently used `Deriv::dP` etc everywhere.

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/CO2Solubility.hpp
dkachuma: As long as we're in the single phase regime, `phaseFraction` will not change so the derivatives will be zero.

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/CO2Solubility.hpp
dkachuma: Yes. Again valid as long as we remain single phase.

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/CO2Solubility.hpp
dkachuma: > Those hard-coded `0` and `1` would benefit from using similar struct as `Deriv::dP` to not get lost at some point

To be replaced to use everywhere `Deriv`. I have also favoured the use of loops which is hopefully less error prone.

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/CO2Solubility.hpp
dkachuma: The condition `L=1-V` assumes that `z_co2+z_wat=1` which is true in actual simulations. However, the actual partial derivatives in this case do not use this relationship. Allowing `z_co2` and `z_wat` to vary independently like this should make us more robust to whichever composition is eliminated when these derivatives are propagated down the pipeline. Also it ensures that the numerical derivative tests are simpler to implement.

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/CO2Solubility.hpp
dkachuma: Yes this is a substitution. This is because the value of x_co2 can potentially be truncated to ensure a minimum value of 0 - negative values can occur if z_wat is very small. Such cases would reduce to single phase anyway.

As for which is best to start solving for, my take is that it will be a fairly stable 2x2 linear system with `determinant ~= 1` so it really shouldn't matter in most cases.

dkachuma:
The integrated tests showed these failures
```
---------- : ----------------------------------------------- : ------------------------------------------------------------------- : ----------
Status     : TestCase                                        : Directory                                                           : TestStep  
---------- : ----------------------------------------------- : ------------------------------------------------------------------- : ----------
FAIL CHECK : co2_thermal_2d_01                               : thermalMultiphaseFlow/co2_thermal_2d_01                             : restartcheck
FAIL CHECK : staircase_co2_wells_3d_01                       : compositionalMultiphaseWell/staircase_co2_wells_3d_01               : restartcheck
FAIL CHECK : class09_pb3_smoke_3d_01                         : compositionalMultiphaseWell/class09_pb3_smoke_3d_01                 : restartcheck
FAIL CHECK : thermalLeakyWell_smoke_3d_01                    : compositionalMultiphaseFlow/thermalLeakyWell_smoke_3d_01            : restartcheck
FAIL CHECK : PoroElastic_staircase_co2_3d_sequential_01      : poroElasticCoupling/PoroElastic_staircase_co2_3d_sequential_01      : restartcheck
FAIL CHECK : co2_flux_3d_01                                  : compositionalMultiphaseFlow/co2_flux_3d_01                          : restartcheck
FAIL CHECK : ThermoPoroElastic_staircase_co2_smoke_01        : thermoPoromechanics/ThermoPoroElastic_staircase_co2_smoke_01        : restartcheck
FAIL CHECK : PoroElastic_hybridHexPrism_co2_fim_3d_01        : poroElasticCoupling/PoroElastic_hybridHexPrism_co2_fim_3d_01        : restartcheck
FAIL CHECK : PoroElastic_permeableFault_smoke_01             : poroElasticCoupling/PoroElastic_permeableFault_smoke_01             : restartcheck
FAIL CHECK : staircase_co2_wells_hybrid_3d_01                : compositionalMultiphaseWell/staircase_co2_wells_hybrid_3d_01        : restartcheck
FAIL CHECK : co2_hybrid_1d_01                                : compositionalMultiphaseFlow/co2_hybrid_1d_01                        : restartcheck
FAIL CHECK : PoroElastic_impermeableFault_smoke_01           : poroElasticCoupling/PoroElastic_impermeableFault_smoke_01           : restartcheck
FAIL CHECK : co2_hybrid_1d_02                                : compositionalMultiphaseFlow/co2_hybrid_1d_02                        : restartcheck
FAIL CHECK : PoroElastic_staircase_co2_3d_fim_01             : poroElasticCoupling/PoroElastic_staircase_co2_3d_fim_01             : restartcheck
FAIL CHECK : PoroElastic_hybridHexPrism_co2_sequential_3d_01 : poroElasticCoupling/PoroElastic_hybridHexPrism_co2_sequential_3d_01 : restartcheck
---------- : ----------------------------------------------- : ------------------------------------------------------------------- : ----------
 ```
The cases `PoroElastic_impermeableFault_smoke_01` and `PoroElastic_permeableFault_smoke_01` appear to be failures due to some other issue since they don't even have CO2-brine.

There are 3 types of changes:
- The introduction of an auto generated table e.g. `fluid_CO2Solubility_waterVaporization_table`.
- A change to name of a generated table e.g. rename from `fluid_CO2Solubility_table` to `fluid_CO2Solubility_co2Dissolution_table`.
- A numerical change in phase properties for a phase that does not exist. With the new flash, the composition of a non-existent phase is simply set to a single component having 100%. This changes the properties of the phase but they have no further impact on the result.

As far as I can tell, these changes are expected and OK.

PR #2815, opened 2023-11-10
Update `logLevel` usage in MGR

victorapm:
- Fix behavior of `logLevel = 0` (shouldn't print anything)
- Change logic for printing point markers array

Depends on https://github.com/GEOS-DEV/thirdPartyLibs/pull/249


castelletto1 (approved):
Thank you, Victor.

src/coreComponents/linearAlgebra/interfaces/hypre/HypreMGR.cpp
```diff
@@ -67,7 +67,7 @@ void hypre::mgr::createMGR( LinearSolverParameters const & params,
   {
     GEOS_LOG_RANK_0( GEOS_FMT( "        MGR preconditioner: numComponentsPerField = {}", numComponentsPerField ) );
   }
-  if( params.logLevel >= 4 )
+  if( params.logLevel >= 1024 )
```
CusiniM, line 70: why such a high value for this?

victorapm: 1 up to 128 are needed by hypre. I put 1024 just to be on the safe side (in case we add new `print_level` codes to MGR in the future)

rrsettgast (approved):

PR #2817, opened 2023-11-10
Read field directly from file

paveltomin:


paveltomin:
@klevzoff @CusiniM is this something legal or I am on the wrong path here?

klevzoff:
We don't want to support something like this because there is no way to make it work robustly. GEOS makes no guarantees about numbering of mesh entities (cells, nodes), especially in parallel, but any attempt to consume per-entity values directly from a file has to assume a particular order of values. TableFunction is a robust way to map values via coordinates for Cartesian grids, while unstructured datasets have their own way of managing attached data (e.g. vtk files with fields).

paveltomin:
I see, thanks.
_GEOS makes no guarantees about numbering_  - can you elaborate what does it mean? There should be some global id for cells, no?

klevzoff:
> _GEOS makes no guarantees about numbering_ - can you elaborate what does it mean? There should be some global id for cells, no?

Yes, but there the specific assignment of these IDs is an implementation detail. Like, `InternalMeshGenerator` currently assigns global IDs in z-y-x order, but it has been proposed to change it to x-y-z, which would be a breaking change if our input files relied on it.

In general, relying on IDs/ordering that you don't explicitly control seems like a bad idea to me. For the same reason, we don't specify `TimeHistory` collection via explicit cell IDs, but by geometric regions (e.g. boxes). It's just very brittle to use IDs in presence of mesh parallel partitioning and/or refinement.

Though I should probably clarify, the above is just my opinion, others may disagree. Perhaps we could start from the use case and come up with a robust solution that may or may not involve changes in GEOS? Is it structured or unstructured mesh? If structured, is it rectilinear or skewed/slanted?

PR #2820, opened 2023-11-15
Implement Spycher-Pruess solubility tables

dkachuma:
Populates CO2 solubility and water vapourisation tables using the formulation from [Spycher et al. (2003)](https://doi.org/10.1016/S0016-7037(03)00273-4)
* Switch from Duan & Sun (2003) model to Spycher et al. (2003) model done by adding an extra parameter to the flash file.
* Values are precalculated for a range of pressure and temperature and stored as a 2D table which is used in the "flash" (CO2Solubility).
* Currently salinity is passed but not yet used. This will have to be extended using [Spycher & Pruess (2005)
](https://doi.org/10.1016/j.gca.2005.01.015) to include salinity.


src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/CO2Solubility.cpp
```diff
@@ -321,8 +322,28 @@ CO2Solubility::CO2Solubility( string const & name,
   string const expectedWaterPhaseNames[] = { "Water", "water", "Liquid", "liquid" };
   m_phaseLiquidIndex = PVTFunctionHelpers::findName( phaseNames, expectedWaterPhaseNames, "phaseNames" );
 
-  m_CO2SolubilityTable = makeSolubilityTable( inputParams, m_modelName, FunctionManager::getInstance() );
-  m_WaterVapourisationTable = makeVapourisationTable( inputParams, m_modelName, FunctionManager::getInstance() );
+  string const solubilityModel = 10 < inputParams.size() ? inputParams[10] : "DuanSun";
+  FunctionManager & functionManager = FunctionManager::getInstance();
+
+  if( solubilityModel == "SpycherPruess" )
```
jafranc: would go for `enum` compare rather than string, but as long as it works :)
see `ENUM_STRINGS` IIRC

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/CO2Solubility.cpp
```diff
@@ -321,8 +322,28 @@ CO2Solubility::CO2Solubility( string const & name,
...
+  {
+    std::pair< TableFunction const *, TableFunction const * > tables =
+      CO2SolubilitySpycherPruess::makeSolubilityTables( inputParams, m_modelName, functionManager );
+    m_CO2SolubilityTable = tables.first;
+    m_WaterVapourisationTable = tables.second;
+  }
+  else if( solubilityModel == "DuanSun" )
+  {
+    m_CO2SolubilityTable = makeSolubilityTable( inputParams, m_modelName, functionManager );
```
jafranc: Wouldn't it be better to have `CO2Solubility` to be pure virtual and each model in a class, hence `CO2SolubilitySpycherPruess` and `CO2SolubilityDuanSun`

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/CO2Solubility.cpp
```diff
@@ -321,8 +322,28 @@ CO2Solubility::CO2Solubility( string const & name,
...
+    m_WaterVapourisationTable = makeVapourisationTable( inputParams, m_modelName, functionManager );
+  }
+  else
+  {
+    GEOS_THROW( GEOS_FMT( "The CO2Solubility model {} is not known."
+                          " This should be either 'DuanSun' or 'SpycherPruess'.", solubilityModel ),
```
jafranc: Would go as @MelReyCG did to list enum values if possible

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/PVTFunctionHelpers.hpp
```diff
@@ -19,6 +19,8 @@
 #include "common/DataTypes.hpp"
 #include "common/Units.hpp"
 
+#include "codingUtilities/StringUtilities.hpp"
```
jafranc, line 22: Probably should be in the cpp if the only change in the file

src/coreComponents/unitTests/constitutiveTests/CMakeLists.txt
```diff
@@ -1,9 +1,11 @@
 # Specify list of tests
 set( gtest_geosx_tests
+     testCapillaryPressure.cpp
+     testCO2BrinePVTModels.cpp
```
jafranc, line 4: Why `testCO2BrinePVTModel` is moved out of the flag `ENABLE_PVTPackage` ?

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/CO2SolubilitySpycherPruess.cpp
```diff
@@ -0,0 +1,285 @@
...
+      real64 const y_H2O = (1.0 - B)/(1.0/A - B);
+      real64 const x_CO2 = B*(1.0 - y_H2O);
+
+      // Calculate the solubility
+      co2Values[j*nPressures+i] = x_CO2/((1.0 - x_CO2)*molarMassH2O);
+      h2oValues[j*nPressures+i] = y_H2O/((1.0 - y_H2O)*molarMassCO2);
+    }
+  }
+
+  // Truncate negative solubility and warn
```
jafranc: This, for instance, could go in abstract base class

jafranc (approved):
Great addition to GEOS capability.
Thank you @dkachuma ! Just few comments on class hierarchy

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/PVTFunctionHelpers.hpp
dkachuma: This was a little strange to me. `StringUtilities` is used in the `.hpp` file itself. It means that any file that uses `PVTFunctionHelpers.hpp` has to also include `StringUtilities.hpp` otherwise the compilation fails.

src/coreComponents/unitTests/constitutiveTests/CMakeLists.txt
dkachuma: Not sure why this was added there because it doesn't depend on PVTPackage. It should still be runnable without PVTPackage.

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/CO2Solubility.cpp
dkachuma: I considered adding another solubility class. However, CO2Solubility is used as a template parameter in [CO2BrineFluid](https://github.com/GEOS-DEV/GEOS/blob/cb4e99b3ca3e86fb08017b291e87d054c3d8006d/src/coreComponents/constitutive/fluid/multifluid/CO2Brine/CO2BrineFluid.cpp#L419-L437). Adding a different class likely means adding a new instantiation of the template and probably a new catalog entry. This is probably OK but I didn't want to add to the proliferation of catalog types especially considering that the change in functionality is quite minimal.

In any case I think I should refactor CO2Solubility and clearly separate the Duan&Sun part from the Spycher part.

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/CO2Solubility.cpp
```diff
@@ -390,28 +234,18 @@ CO2Solubility::CO2Solubility( string const & name,
...
-  if( solubilityModel == "SpycherPruess" )
-  {
-    std::pair< TableFunction const *, TableFunction const * > tables =
-      CO2SolubilitySpycherPruess::makeSolubilityTables( inputParams, m_modelName, functionManager );
-    m_CO2SolubilityTable = tables.first;
-    m_WaterVapourisationTable = tables.second;
-  }
-  else if( solubilityModel == "DuanSun" )
+  SolubilityModel solubilityModel = SolubilityModel::DuanSun;   // Default solubility model
+  if( inputParams[1] == EnumStrings< SolubilityModel >::toString( SolubilityModel::Tables ) )
```
jafranc, line 240: Not sure but wouldn't `EnumStrings< SolubilityModel >::fromString(s)` throw if pass any other values than the one defined ? If so, it will spare you the if statement and the silent defaulting

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/CO2Solubility.cpp
```diff
@@ -30,337 +28,183 @@ namespace geos
...
+  FunctionManager & functionManager = FunctionManager::getInstance();
+  constitutive::PVTProps::PTTableCoordinates tableCoords;
 
-  for( localIndex i = 0; i < nPressures; ++i )
+  // Check solubility model for explicit table input
+  if( solubilityModel == constitutive::PVTProps::CO2Solubility::SolubilityModel::Tables )
   {
-    real64 const P = tableCoords.getPressure( i ) / P_Pa_f;
+    // The default table is a table with all zeros unless the name is explicitly provided
+    tableCoords.appendPressure( 1.0e5 ).appendPressure( 1.0e8 )
```
jafranc, line 69: Just out of curiousity/documentation, why these hard-coded values ? Is this all model range ?

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/CO2Solubility.cpp
```diff
@@ -30,337 +28,183 @@ namespace geos
...
-      return getSolubilityTable( inputTableName, functionManager );
+      badValueTable
+        << std::setw( 15 ) << badValues( row, 0 ) << " "
+        << std::setw( 15 ) << badValues( row, 1 ) << " "
+        << std::setw( 23 ) << badValues( row, 2 ) << " "
+        << std::setw( 15 ) << badValues( row, 3 ) << " "
+        << "\n";
     }
+    GEOS_LOG_RANK_0( GEOS_FMT( "CO2Solubility: {} negative solubility values encountered. These will be truncated to zero.\n{}",
+                               badCount, badValueTable.str() ) );
```
jafranc:
```suggestion
    GEOS_LOG_RANK_0( GEOS_FMT( "CO2Solubility: {} negative solubility values encountered. These will be truncated to zero.\n Check out report table {} with max {} values",
                               badCount, badValueTable.str(), maxBad ) );
```

jafranc (approved):
Great refactoring/extension !! 
Thank you @dkachuma

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/CO2Solubility.cpp
dkachuma: These are used to create a zero value table. So they need only have a valid range. I have added a comment to that effect.

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/CO2Solubility.cpp
dkachuma: Yes `EnumStrings::fromString` would throw. However the idea is that the user doesn't have to specify a value in which case we default to `DuanSun` - this preserves current behaviour.

TotoGaz:
Does this require rebaselines?

dkachuma:
> Does this require rebaselines?

No it shouldn't.

PR #2822, opened 2023-11-15
feat: Introduce full tensor permeability

CusiniM:
@castelletto1 should be usable but I have not tested it yet.


src/coreComponents/finiteVolume/BoundaryStencil.hpp
```diff
@@ -204,8 +204,9 @@ BoundaryStencilWrapper::
   auto const compute = [&]
   {
     real64 faceConormal[3];
-    LvArray::tensorOps::hadamardProduct< 3 >( faceConormal, coef, faceNormal );
+    LvArray::tensorOps::Ri_eq_symAijBj< 3 >( faceConormal, coef, faceNormal )
```
CusiniM:
```suggestion
    LvArray::tensorOps::Ri_eq_symAijBj< 3 >( faceConormal, coef, faceNormal );
```

src/coreComponents/constitutive/permeability/ConstantPermeability.cpp
```diff
@@ -28,12 +28,40 @@ namespace constitutive
...
+  registerWrapper( viewKeyStruct::symmetricFullPermeabilityTensorString(), &m_symmetricFullPermeabilityTensor ).
+    setInputFlag( InputFlags::OPTIONAL ).
+    setRestartFlags( RestartFlags::NO_WRITE ).
+    setDescription( "xx, yy and zz, xz, yz, xy components of a symmetric permeability tensor." );
+}
+
+void ConstantPermeability::postProcessInput()
+{
+  GEOS_ERROR_IF( m_diagonalPermeabilityTensor[0] < 0.0 && m_symmetricFullPermeabilityTensor[0] < 0.0, 
+  "Either a diagonal permeability tensor or a full tensor must be provided.");
```
CusiniM:
```suggestion
  "Either a diagonal permeability tensor or a symmetric full tensor must be provided.");
```

castelletto1:
I confirm it works. I ran an incompressible single-phase flow quarter five spot problem without gravity on two domains. The first case, the domain is aligned with the Cartesian axes and the permeability tensor is diagonal. In the second case, the domain is obtained by rotating the aligned domain and consistently updating the permeability tensor, which becomes full tensor. In both cases the grid is K-orthogonal and TPFA is correct. The computed Jacobian matrices are the same. Below two snapshots with details for the same cell (Id 52) in both domains.

aligned

rotated

I temporarily added xml and vtk files to reproduce the results in `./inputFiles/singlePhaseFlow/fullTensorPermeability`
We may end up adding a new smoke test.

@karimifard, @jafranc, @francoishamon: have a look and share if you think this is headed in the correct direction.

src/coreComponents/constitutive/permeability/ConstantPermeability.cpp
```diff
@@ -54,9 +82,8 @@ void ConstantPermeability::allocateConstitutiveData( dataRepository::Group & par
   {
     for( localIndex q = 0; q < numQuad; ++q )
     {
-      m_permeability[ei][q][0] =  m_permeabilityComponents[0];
-      m_permeability[ei][q][1] =  m_permeabilityComponents[1];
-      m_permeability[ei][q][2] =  m_permeabilityComponents[2];
+      for ( int c=0; c < 6; c++)
+      m_permeability[ei][q][c] =  m_symmetricFullPermeabilityTensor[c];
```
jafranc: Shouldn't we enforce positiveness of diagonal component here ?

CusiniM: We should really throw an error earlier if someone inputs a negative value but, for now, I am using the default negative value to distinguish whether a diagonal or a full tensor was provided so I need to improve that logic fist.

CusiniM: One option would be to simply provide 2 R1Tensors one for the main diagonal and one for the off diagonal terms and have the off diagonal components default to 0.0 if not provided.

src/coreComponents/finiteVolume/BoundaryStencil.hpp
```diff
@@ -204,8 +204,9 @@ BoundaryStencilWrapper::
   auto const compute = [&]
   {
     real64 faceConormal[3];
-    LvArray::tensorOps::hadamardProduct< 3 >( faceConormal, coef, faceNormal );
+    LvArray::tensorOps::Ri_eq_symAijBj< 3 >( faceConormal, coef, faceNormal );
     weight = LvArray::tensorOps::AiBi< 3 >( cellToFace, faceConormal );
+    // Matteo: this one also has to change! 
     LvArray::tensorOps::hadamardProduct< 3 >( dWeight_dCoef, cellToFace, faceNormal );
```
jafranc, line 211: I am not sure about how this one should change.

src/coreComponents/constitutive/permeability/ConstantPermeability.cpp
jafranc: Just noticed that the order would be imposed by `LvArray::Ri_eq_symAijBj` implicitly :/

CusiniM: yes, and it's different from the one used by `vtk`.

jafranc:
> I confirm it works. I ran an incompressible single-phase flow quarter five spot problem without gravity on two domains. The first case, the domain is aligned with the Cartesian axes and the permeability tensor is diagonal. In the second case, the domain is obtained by rotating the aligned domain and consistently updating the permeability tensor, which becomes full tensor. In both cases the grid is K-orthogonal and TPFA is correct. The computed Jacobian matrices are the same. Below two snapshots with details for the same cell (Id 52) in both domains.
> 
> aligned
> 
> rotated
> 
> I temporarily added xml and vtk files to reproduce the results in `./inputFiles/singlePhaseFlow/fullTensorPermeability` We may end up adding a new smoke test.
> 
> @karimifard, @jafranc, @francoishamon: have a look and share if you think this is headed in the correct direction.

I let @karimifard and @francoishamon, who are more expert than I am, tell if it is what they had in mind. 
It looks like a right start and not adding too much complexity. There will be some bookkeeping traps along the road though and I like @CusiniM 's idea to have one tensor for the diag and one for off-diag so the full tensor case is use as the base and off-diag zeroed out by default.

paveltomin (approved):

paveltomin:
@castelletto1 can we just merge this?

src/coreComponents/finiteVolume/BoundaryStencil.hpp
castelletto1: https://github.com/GEOS-DEV/GEOS/pull/2822#issuecomment-3169306412

This should be addressed first

paveltomin: @castelletto1 do you know how it should be changed?

rrsettgast:
@copilot resolve the merge conflicts in this pull request

PR #2801, opened 2023-11-03
Reorganization of the GHA files

TotoGaz:
The integrated tests can be run as part of the workflow when the label `ci: run integrated tests` is associated to the PR.

A bundle of the new reference files and of the logs is sent to the cloud and made available for download to the developer. (_For the sake of merging, this step deactivated in this PR._).

For the moment, integrated tests are only requested to be run on 2 procs. This can be changed easily.

Important note: while the integrated tests can be _run_, they are not _passing_:
```
    Status     : TestCase                                      : Directory                                                         : Elapsed : Resources : TestStep  
    ---------- : --------------------------------------------- : ----------------------------------------------------------------- : ------- : --------- : ----------
    FAIL RUN   : Sneddon_embeddedFrac_benchmark_01             : embeddedFractures/Sneddon_embeddedFrac_benchmark_01               : 0:00:03 : 0:00:03   : geos      
    ---------- : --------------------------------------------- : ----------------------------------------------------------------- : ------- : --------- : ----------
    FAIL CHECK : PoroElastic_permeableFault_smoke_01           : poroElasticCoupling/PoroElastic_permeableFault_smoke_01           : 0:00:04 : 0:00:04   : restartcheck
    FAIL CHECK : PoroElastic_impermeableFault_smoke_01         : poroElasticCoupling/PoroElastic_impermeableFault_smoke_01         : 0:00:04 : 0:00:04   : restartcheck
    FAIL CHECK : grav_seg_c1ppu_hyst_01                        : compositionalMultiphaseFlow/grav_seg_c1ppu_hyst_01                : 0:00:05 : 0:00:05   : restartcheck
    FAIL CHECK : ThermoPoroElastic_consolidation_sequential_02 : thermoPoromechanics/ThermoPoroElastic_consolidation_sequential_02 : 0:00:16 : 0:00:32   : restartcheck
    FAIL CHECK : grav_seg_c1ppu_hyst_02                        : compositionalMultiphaseFlow/grav_seg_c1ppu_hyst_02                : 0:00:03 : 0:00:06   : restartcheck
    ---------- : --------------------------------------------- : ----------------------------------------------------------------- : ------- : --------- : ----------
```

Also:
- The GHA `yaml` files now only manage input data and job dependencies.
In the end, the job informations are later provided to `ci_build_and_test_in_container.sh` directly.
Script `ci_build_and_test.sh`, which was previously calling `ci_build_and_test_in_container.sh` could be removed.
[...]

PR #2825, opened 2023-11-15
Updating error checking in python setup script

### LINKED ISSUE #2690: Cannot `make ats_run` on Pecan
On Pecan, I can only partially compile `geosx_python_tools` with most recent develop repo , `setup_ats_environment` was somehow NOT compiled along with other python tools.

```
[100%] Built target geosx
(py3) [j0551570@pecan-1 build-pecan_cpu2-release]$ make geosx_python_tools
Generating ../../bin/preprocess_xml, ../../bin/format_xml
Checking the python target...
Attempting to install packages directly in target python environment...
  /shared/data1/Users/j0551570/IntegratedTests/GEOS/scripts/../src/coreComponents/python/modules/geosx_mesh_tools_package
  /shared/data1/Users/j0551570/IntegratedTests/GEOS/scripts/../src/coreComponents/python/modules/geosx_xml_tools_package
  /shared/data1/Users/j0551570/IntegratedTests/GEOS/scripts/../src/coreComponents/python/modules/hdf5_wrapper_package
  /shared/data1/Users/j0551570/IntegratedTests/GEOS/scripts/../src/coreComponents/python/modules/pygeosx_tools_package
  /shared/data1/Users/j0551570/IntegratedTests/GEOS/scripts/../integratedTests/scripts/geos_ats_package
Linking key scripts to bin directory...
  preprocess_xml
/home/j0551570/.conda/envs/py3/bin/preprocess_xml
    (found preprocess_xml as /home/j0551570/.conda/envs/py3/bin/preprocess_xml)
  format_xml
/home/j0551570/.conda/envs/py3/bin/format_xml
    (found format_xml as /home/j0551570/.conda/envs/py3/bin/format_xml)
  convert_abaqus
/home/j0551570/.conda/envs/py3/bin/convert_abaqus
    (found convert_abaqus as /home/j0551570/.conda/envs/py3/bin/convert_abaqus)
  run_geos_ats
/home/j0551570/.conda/envs/py3/bin/run_geos_ats
/home/j0551570/.local/bin/run_geos_ats
/home/j0551570/local/bin/run_geos_ats
    (could not find where run_geos_ats is installed)
  setup_ats_environment
/home/j0551570/.conda/envs/py3/bin/setup_ats_environment
/home/j0551570/.local/bin/setup_ats_environment
/home/j0551570/local/bin/setup_ats_environment
    (could not find where setup_ats_environment is installed)
  activate
/home/j0551570/.conda/envs/py3/bin/activate
/home/j0551570/.local/bin/activate
/home/j0551570/local/bin/activate
    (could not find where activate is installed)
  python
/home/j0551570/.conda/envs/py3/bin/python
    (found python as /home/j0551570/.conda/envs/py3/bin/python)
Linking additional scripts to the bin directory...
Done!
Built target geosx_python_tools
```

`setup_ats_environment` cannot be found in my folder and hence following failure on `make ats_run` or `make ats_environment` was reported.

```
(py3) [j0551570@pecan-1 build-pecan_cpu2-release]$ make ats_environment
Built target geosx_python_tools
Generating geos_ats.sh
make[3]: /shared/data1/Users/j0551570/IntegratedTests/GEOS/build-pecan_cpu2-release/bin/setup_ats_environment: Command not found
make[3]: *** [integratedTests/geos_ats.sh] Error 127
make[2]: *** [integratedTests/CMakeFiles/ats_environment.dir/all] Error 2
make[1]: *** [integratedTests/CMakeFiles/ats_environment.dir/rule] Error 2
make: *** [ats_environment] Error 2
(py3) [j0551570@pecan-1 build-pecan_cpu2-release]$ which setup_ats_environment
/usr/bin/which: no setup_ats_environment in (/home/j0551570/.conda/envs/py3/bin:/hrtc/apps/scripting/anaconda3/x86_64/2020.02/condabin:/hrtc/apps/devtools/cmake/x86_64/centos7/3.26.4/bin:/appli_MTS/TOOLS/condacommon/Scripts:/shared/data1/Users/j0265258/pub/scripts:/home/j0497482/survival_kit/UtilsTL:/home/j0497482/bin:.:/usr/lib64/qt-3.3/bin:/usr/local/bin:/usr/bin:/usr/local/sbin:/usr/sbin:/opt/ibutils/bin:/apps/slurm/x86/20.11.9-pecan/bin:/apps/slurm/x86/20.11.9-pecan/sbin)
```

The required `virtualenv` package is installed.
```
(py3) [j0551570@pecan-1 geosx_mesh_tools_package]$ python -m virtualenv --version
virtualenv 20.4.6 from /home/j0551570/.conda/envs/py3/lib/python3.8/site-packages/virtualenv/__init__.py
```

- Machine: Pecan 
- GEOSX Version: develop `52354b4`
- cmake/3.26.4 gcc/8.2.0 openmpi/4.0.1 virtualenv/20.4.6 python/3.8

cssherman:


cssherman:
@jhuang2601 - Could you try setting up the testing environment using this branch?  These changes will also detect an environment variable VERBOSE to print additional information if the default values are unclear:

`VERBOSE=1 make ats_environment`

scripts/setupPythonEnvironment.bash
```diff
@@ -109,10 +124,25 @@ then
...
+            fi
+            
+            if [[ $RES =~ "HTTP error" ]]
+            then
+                echo "Could not access package dependencies:"
+                echo $RES
+                exit 1
+            fi
+
             if [[ $RES =~ "Error" ]]
```
TotoGaz: Is there an error code sent by `pip` when it fails?
If so, capturing the `$?` shell variable just after the call seems like the most reliable way to solve this.

cssherman: It seems like pip just uses the unspecified error code, so we will still need to do some string parsing to figure out the details of any failures:

```
  ERROR: HTTP error 404 while getting https://github.com/LLNL/ATS/archive/refs/tags/7.0.105_badurl.tar.gz
ERROR: Could not install requirement ats@ https://github.com/LLNL/ATS/archive/refs/tags/7.0.105_badurl.tar.gz from https://github.com/LLNL/ATS/archive/refs/tags/7.0.105_badurl.tar.gz (from geos-ats==0.2.0) because of HTTP error 404 Client Error: Not Found for url: https://codeload.github.com/LLNL/ATS/tar.gz/refs/tags/7.0.105_badurl for URL https://github.com/LLNL/ATS/archive/refs/tags/7.0.105_badurl.tar.gz
WARNING: You are using pip version 22.0.4; however, version 23.3.1 is available.
You should consider upgrading via the '/usr/tce/bin/python3 -m pip install --upgrade pip' command.
[sherman@quartz1148:geos_ats_package]$ echo $?
1
```

cssherman: (Because in some cases, we want to continue trying to install the package in a virtual environment if we detect an error)

TotoGaz: > (Because in some cases, we want to continue trying to install the package in a virtual environment if we detect an error)

What is the use case for that?
At first thought, I would kill the process at the first error popping (providing the complete error message and asking the end-user to fix the error).

cssherman: > > (Because in some cases, we want to continue trying to install the package in a virtual environment if we detect an error)
> 
> What is the use case for that? At first thought, I would kill the process at the first error popping (providing the complete error message and asking the end-user to fix the error).

That's reasonable.  This fall-back option (where we try again to build in a virtual environment) could be replaced with a request to the user to re-try with the correct arguments.

And TBH, with recent changes to pip defaults, the user-privilege errors that would require a virtualenv be created don't seem to show up often now.

cssherman: @TotoGaz - with regards to the HTTPS connectivity issues we've seen, do you think we should just throw an error and tell users to setup the tests on a machine with access to github?  Or, should we try to add an option to install the ATS package from a local source as a fall-back?  (This second option may need to happen in the setup script being modified here, since it isn't a common pattern for setup.cfg files.)

TotoGaz: Managing error and fallbacks, etc... is really complicated. Especially in shell 😨 
Providing a script that can do it for us is fine, but if it fails, I would print an URL on readthedocs (or a github issue) where we can detail workarounds and hints to help.
Now, I do not know what that implies in term of `cmake` dependencies. But maybe we can revisit some elements as well?

cssherman: @TotoGaz - I've updated the script and tried to simplify the processing/error checking.  I've also removed the (never used) code to create a miniconda environment, since I don't think that we should maintain this.

scripts/setupPythonEnvironment.bash
```diff
@@ -19,42 +19,51 @@ declare -a LINK_SCRIPTS=("preprocess_xml"
 
 
 # Read input arguments
-PYTHON_TARGET="$(which python)"
+PYTHON_TARGET="$(which python3)"
```
TotoGaz: How does this relate to some `Python3_ROOT_DIR` which could have been provided?
And 👍 for the 3 suffix!

scripts/setupPythonEnvironment.bash
```diff
@@ -144,31 +120,53 @@ then
...
+        echo "Virtual environment path:"
+        echo $VIRTUAL_PATH
+    fi
+
     mkdir -p $VIRTUAL_PATH/$VIRTUAL_NAME
     $PYTHON_TARGET -m virtualenv --system-site-packages $VIRTUAL_PATH/$VIRTUAL_NAME
 
-    # Install packages
-    echo "Installing packages..."
     PYTHON_TARGET=$VIRTUAL_PATH/$VIRTUAL_NAME/bin/python
```
TotoGaz: It looks like we're swinging between the `python3` of the `PATH`, the `Python3_ROOT_DIR` if defined, and a `virtualenv` created on the fly.

Could we ask the user to explicitly define `Python3_ROOT_DIR` and let him pass this information to the script as a parameter? (If not provided, we die with an explicit message)
That way, he can use whatever env system he wants and we get rid of any troubleshooting on our side? (Surely logging which python is being used can help).

TotoGaz: E.g. I've added
```shell
if [[ "${RUN_INTEGRATED_TESTS}" = true ]]; then
  echo "Running the integrated tests has been requested."
  # We install the python environment required by ATS to run the integrated tests.
  or_die apt-get install -y virtualenv python3-dev python-is-python3
  ATS_PYTHON_HOME=/tmp/run_integrated_tests_virtualenv
  or_die virtualenv ${ATS_PYTHON_HOME}
  ATS_CMAKE_ARGS="-DATS_ARGUMENTS=\"--machine openmpi --ats openmpi_mpirun=/usr/bin/mpirun --ats openmpi_args=--allow-run-as-root --ats openmpi_procspernode=2 --ats openmpi_maxprocs=2\" -DPython3_ROOT_DIR=${ATS_PYTHON_HOME}"
fi
```
in the ci file [here](https://github.com/GEOS-DEV/GEOS/pull/2801/files#diff-0cef284cfdb08fea23f36d5d918fe37db5c5af609b2837dc27253a354dc1e641). (see `DPython3_ROOT_DIR` at the end).
Maybe not the cleanest, but I'm consistent that way.

scripts/setupPythonEnvironment.bash
cssherman: Usually, cmake passes its version of Python3_Executable via the `-p` argument of this script.  This which statement was put in place to give a reasonable guess for the default option when used in a standalone mode.

scripts/setupPythonEnvironment.bash
cssherman: > It looks like we're swinging between the `python3` of the `PATH`, the `Python3_ROOT_DIR` if defined, and a `virtualenv` created on the fly.
> 
> Could we ask the user to explicitly define `Python3_ROOT_DIR` and let him pass this information to the script as a parameter? (If not provided, we die with an explicit message) That way, he can use whatever env system he wants and we get rid of any troubleshooting on our side? (Surely logging which python is being used can help).

That's fair.  If you want, we could remove the default (from the path) and the virtualenv construction method.  These are all hold-overs from historical development/testing, and are rarely used now.

TotoGaz: I guess that the script would be simplified into: give me an environment in which I can install my python packages.
If I can't, here's the error message, deal with it and start again.

scripts/setupPythonEnvironment.bash
```diff
@@ -19,42 +19,42 @@ declare -a LINK_SCRIPTS=("preprocess_xml"
...
-    -p|--python_target)
+    -p|--python-target)
     PYTHON_TARGET="$2"
     shift # past argument
     ;;
-    -m|--miniconda_build)
-    MINICONDA_BUILD="$2"
-    shift # past argument
-    ;;
-    -v|--virtual_path)
```
TotoGaz, line 42: When running `cmake`, there's still an error message complaining about `virtualenv` not being installed (for `pygeos`). Do you think it's still relevant?  (I do not know what it's about)

scripts/setupPythonEnvironment.bash
```diff
@@ -91,84 +85,47 @@ then
...
-            if [[ $RES =~ "Error" ]]
-            then
-                echo "  (cannot install target packes directly)"
-                INSTALL_VIRTUAL=true
-                break
-            fi
-        else
-            echo "Could not find target package: $p"
+            INSTALL_MSG=$($PYTHON_TARGET -m $PIP_CMD install $p 2>&1)
+            INSTALL_RC=$?
```
TotoGaz, line 99: Maybe get `INSTALL_RC=$?` out of the `if`/`then`?
As you wish, having the error code the line just after is also good!

TotoGaz (approved):

scripts/setupPythonEnvironment.bash
cssherman: I did try that, but apparently the op code gets reset when it exits the if block.  Thanks for your suggestions!

scripts/setupPythonEnvironment.bash
cssherman: This check shows up on this line:

https://github.com/GEOS-DEV/GEOS/blob/develop/src/coreComponents/python/CMakeLists.txt#L3-L20

As this warning suggests, building `pygeos` requires that the target environment have the virtualenv package installed.  As a final step for this PR, I'll move this check inside the `ENABLE_PYGEOSX` check and change it to an error.

src/coreComponents/python/CMakeLists.txt
```diff
@@ -4,19 +4,19 @@ if ( Python3_EXECUTABLE )
      # Select the version of python to target
      if( ENABLE_PYGEOSX )
           set( PYTHON_POST_EXECUTABLE ${CMAKE_BINARY_DIR}/lib/PYGEOSX/bin/python CACHE PATH "" FORCE )
```
cssherman, line 6: @TotoGaz - I've moved the check for the virtualenv package to the pygeosx block here, since it isn't a firm requirement for the environment setup script anymore.

src/coreComponents/python/CMakeLists.txt
```diff
@@ -30,8 +30,13 @@ if ( Python3_EXECUTABLE )
...
 
-     add_custom_target( geosx_python_tools
-                        DEPENDS ${GEOSX_PYTHON_TOOLS_BINS} )
+     if( ENABLE_PYGEOSX )
+          add_custom_target( geosx_python_tools
+                             DEPENDS pygeosx ${GEOSX_PYTHON_TOOLS_BINS} )
+     else()
+          add_custom_target( geosx_python_tools
+                             DEPENDS ${GEOSX_PYTHON_TOOLS_BINS} )
+     endif()
```
cssherman, line 39: @TotoGaz @rrsettgast - I've added pygeosx as a dependency of geosx_python_tools when PYGEOSX is active.  I've also removed the (now uneccesary) error messages to the user in the setup script instructing to manually run this step if they missed it.

src/coreComponents/python/CMakeLists.txt
TotoGaz: Thx. @sframba as you are using `pygeosx` a lot, is it OK for you?

cssherman:
@CusiniM - Could you test these changes to the python setup script on OSX (or suggest someone else to do so)?

PR #2854, opened 2023-11-29
Fix Rachford-Rice solver

dkachuma:
- Fixes a bug on the SSI portion of the Rachford-Rice solver.
- Adds a unit test that follows this path.


joshua-white (approved):
Nice.  The other fixes and cleanups you added look good too.

src/coreComponents/constitutive/fluid/multifluid/compositional/functions/RachfordRice.hpp
```diff
@@ -98,48 +98,37 @@ struct RachfordRice
...
+      else if( 0.0 < funcXMin * funcXMid )
       {
         xMin = xMid;
-        recomputeMax = false;
-        recomputeMin = true;
+        funcXMin = funcXMid;
       }
 
-      currentError = LvArray::math::min( LvArray::math::abs( funcXMax - funcXMin ),
+      currentError = LvArray::math::min( LvArray::math::abs( funcXMid ),
```
joshua-white, line 131: We should maybe take a careful look at the `currentError` metric and potential scaling issues here and in the Newton loop below.  I think the absolute error metric is mostly okay here in SSI since everything should be hopefully order 1 near the root, but the error metric in the newton loop below at line 148 should be fixed.  If `newtonValue` approaches zero, the error may appear very large when it is not.  I suggest we use a very similar error measure in both SSI and newton.  `deltaNewton` serves the same purpose as `xMax-xMin`, with an absolute value checking having the meaning of the precision in phase saturation we need, always bounded between 0 and 1.   Maybe we should add a function value check to Newton too.  To be safe, I might also scale the function evaluation by the function evaluated at the midpoint (x=0.5) or some other starting guess, in case the function values have crazy scaling.  That one may benefit from a relative error check versus an absolute error check.  What do you think?

joshua-white: On second thought, maybe both the `deltaNewton` and `functionValue` checks should be absolute, with no relative scaling at all.  There may be situations where the initial guess for `functionValue` is near perfect.
```
SSI:  currentError = min( abs( funcXMid ), abs( xMax - xMin ) );
Newt: currentError = min( abs( newtonValue ), abs( deltaNewton ) );
```

dkachuma: > 
> ```
> Newt: currentError = min( abs( newtonValue ), abs( deltaNewton ) );
> ```

I would suppose we want
```
Newt: currentError = min( abs( evaluate(newtonValue) ), abs( deltaNewton ) );
```
i.e. the function value at the newtonValue not the newtonValue itself.

joshua-white: Whoops, yes, exactly

src/coreComponents/constitutive/fluid/multifluid/compositional/functions/RachfordRice.hpp
```diff
@@ -151,12 +140,11 @@ struct RachfordRice
...
 
     while( ( currentError > newtonTolerance ) && ( newtonIteration < maxNewtonIterations ) )
     {
-      real64 const deltaNewton = -evaluate( kValues, feed, presentComponentIds, newtonValue )
-                                 / evaluateDerivative( kValues, feed, presentComponentIds, newtonValue );
-      currentError = LvArray::math::abs( deltaNewton ) / LvArray::math::abs( newtonValue );
+      real64 const deltaNewton = -funcNewton / evaluateDerivative( kValues, feed, presentComponentIds, newtonValue );
 
       // test if we are stepping out of the [xMin;xMax] interval
       if( newtonValue + deltaNewton < xMin )
```
joshua-white, line 150:
```suggestion
      if( newtonValue + deltaNewton < xMin )
      {
        deltaNewton = 0.5 * ( xMin - newtonValue );
      }
      else if( newtonValue + deltaNewton > xMax )
      {
        deltaNewton = 0.5 * ( xMax - NewtonValue );
      }
       
      newtonValue = newtonValue + deltaNewton;
```

This way, when the `deltaNewton` is checked in `currentError` below, it is accurately reflecting the `deltaNewton` used in case of chopping at the bounds.

TotoGaz:
@dkachuma it looks like you did not run the integrated tests using the CI. You're sure we do not need any rebaseline?

TotoGaz:
@dkachuma I've run the integrated tests and no rebaseline is needed.
You were in the merge queue and then no more; what happened? I've put you back on top.
Ping me if you want me to merge it.

PR #2861, opened 2023-11-30
Time step cut support for sequential solver

### LINKED ISSUE #2858: Sequential solver does not support time step cut
If outer loop does not converge, the solver just terminates with the following error:
```
  Iteration 20: outer-loop convergence check
        ( Rflow ) = ( 3.25e-04 )        ( Rwell ) = ( 1.67e-10 )        ( Rsolid ) = ( 2.03e-12 )        ( R ) = ( 3.25e-04 )
***** ERROR
***** LOCATION: /data/rpo_ptls/GEOSX/residual_flash/GEOS/GEOS/src/coreComponents/physicsSolvers/multiphysics/CoupledSolver.hpp:457
***** Controlling expression (should be false): !isConverged
***** Rank 209: SEAM (SEAM_seq.xml, l.4): sequentiallyCoupledSolverStep did not converge!
```

paveltomin:
Currently the simulation would stop if outer loop didn't converge. This PR adds time step cut logic.
Bonus:
- print next dt in sub-timestepping process
- add setNextDtBasedOnNewtonIter for coupled solver to avoid subsolvers doing a lot of iters
- when pressure change is computed - switch from relative to absolute for pres < 1


paveltomin:
tests are ok, does not need rebaseline

jhuang2601 (approved):

PR #2872, opened 2023-12-06
Use GHA's `if:` instead of dedicated jobs

TotoGaz:
This prevents from having specific jobs acting as gates to `label`-driven jobs.
Those jobs (_e.g._ `are_integrated_tests_requested`) may appear `KO` and the whole workflow would come red because an optional job was not triggered.
Using GHA's native `if:` pattern in the `yaml` keeps the workflow green when the integrated tests are not triggered.

@jhuang2601


.github/workflows/ci_tests.yml
```diff
@@ -220,16 +189,15 @@ jobs:
       ENABLE_TRILINOS: OFF
       GCP_BUCKET: geosx/integratedTests
       RUNS_ON: ubuntu-22.04
-
+        
   # If the 'ci: ready to be merged' PR label is found, the cuda jobs run immediately along side linux jobs.
   # Note: CUDA jobs should only be run if PR is ready to merge.
```
bmhan12, line 194: Same comment and note should be added for the `run_integrated_tests` job.

bmhan12 (approved):

.github/workflows/ci_tests.yml
TotoGaz: done

jhuang2601 (approved):

rrsettgast (approved):

PR #2873, opened 2023-12-07
use streak.llnl.gov as runner for GHA

rrsettgast:
This PR uses streak to execute one CUDA build and execute unit tests on the GPU. Two tests are disabled `testLifoStorage` and `testExternalSolvers` are disabled on streak due to failure of those tests on GPU with unified memory disabled. Follow up PR's will be needed to fix and reenable those tests.


src/coreComponents/constitutive/solid/SolidBase.hpp
```diff
@@ -98,7 +98,7 @@ class SolidBaseUpdates
   arrayView1d< real64 const > const m_thermalExpansionCoefficient;
 
   /// Flag to disable inelasticity
-  const bool & m_disableInelasticity;
+  const bool m_disableInelasticity;
```
rrsettgast, line 101: This was a bug on heterogeneous memory systems. @klevzoff This was the problem with the constitutive unit tests from the remove `array1d<string>`

src/coreComponents/linearAlgebra/unitTests/testLinearAlgebraUtils.hpp
```diff
@@ -192,6 +192,8 @@ void compute2DLaplaceOperator( MPI_Comm comm,
 
   // Construct the 2D Laplace matrix
   laplace2D.create( matrix.toViewConst(), matrix.numRows(), comm );
+
+//  laplace2D.write( "matrix.txt", LAIOutputFormat::NATIVE_ASCII);
```
rrsettgast: keep this because it will be needed when we fix this test on GPU.

scripts/ci_build_and_test_in_container.sh
```diff
@@ -232,7 +252,11 @@ fi
 
 # Run the unit tests (excluding previously ran checks).
 if [[ "${RUN_UNIT_TESTS}" = true ]]; then
-  or_die ctest --output-on-failure -E "testUncrustifyCheck|testDoxygenCheck"
+  if [[ ${HOSTNAME} == 'streak.llnl.gov' ]]; then
+    or_die ctest --output-on-failure -E "testUncrustifyCheck|testDoxygenCheck|testLifoStorage|testExternalSolvers"
```
rrsettgast, line 256: This is to exclude the two failing unit tests on GPU systems so we can merge this and follow up with fix PR's.

scripts/ci_build_and_test_in_container.sh
```diff
@@ -143,6 +143,26 @@ EOT
...
+        filename_no_ext="${filename%.*}"
+        new_filename="${DOCKER_CERTS_DIR}/${filename_no_ext}.crt"
+        cp "$file" "$new_filename"
+        echo "Copied $filename to $new_filename"
+      fi
+    done
+    update-ca-certificates 
+    # gcloud config set core/custom_ca_certs_file cert.pem'
+    
+    NPROC=8
```
rrsettgast, line 160: this is hard coded right now for streak.

.github/workflows/build_and_test.yml
```diff
@@ -110,6 +119,11 @@ jobs:
           script_args+=(--sccache-credentials $(basename ${GOOGLE_GHA_CREDS_PATH}))
         fi
 
+        if [ ${{ inputs.RUNS_ON }} == 'self-hosted' ]; then
+          RUNNER_CERTIFICATES_DIR=/etc/pki/ca-trust/source/anchors/
+          mkdir -p ${GITHUB_WORKSPACE}/certificates
+          cp ${RUNNER_CERTIFICATES_DIR}/*.crt* ${GITHUB_WORKSPACE}/certificates
+        fi
```
TotoGaz, line 126: Could we use the `DOCKER_RUN_ARGS` input to add `-v /etc/pki/ca-trust/source/anchors/.../bundle.crt:/etc/pki/ca-trust/.../llnl.bundle.crt:ro`.

That way the certificate would be at the proper location in read-only, with no copy required from within the container.

Then inside of the docker container simply update the certificates list (we can do it whatever the case : it's fast and could do not harm?). Or if we really want it, we can add an option.

.github/workflows/build_and_test.yml
```diff
@@ -126,11 +140,19 @@ jobs:
...
 
         script_args+=(--cmake-build-type ${{ inputs.CMAKE_BUILD_TYPE }})
         script_args+=(${{ inputs.BUILD_AND_TEST_CLI_ARGS }})
 
+        # SPLIT_DOCKER_REPOSITORY=(${DOCKER_REPOSITORY//// })
+        # CONTAINER_NAME=geosx_build_${SPLIT_DOCKER_REPOSITORY[1]}_${GITHUB_SHA:0:7}
+        # if [ "$(docker ps -aq -f name=${CONTAINER_NAME})" ]; then
+        #   docker rm -f ${CONTAINER_NAME}
+        # fi
+        # docker_args+=(--name ${CONTAINER_NAME})
```
TotoGaz: Please remove the commented attempts.

.github/workflows/build_and_test.yml
```diff
@@ -156,6 +179,12 @@ jobs:
             echo "Download the bundle at https://storage.googleapis.com/${{ inputs.GCP_BUCKET }}/${DATA_BASENAME}"
           fi
         fi
+
+        # Remove the container and the workspace to avoid any conflict with the next run.
+        echo github.workspace = ${{ github.workspace }}
+        #rm -rf ${{ github.workspace }}/*
+        #docker rm -f ${CONTAINER_NAME}
```
TotoGaz, line 188: So do you still need to clean up?

scripts/ci_build_and_test_in_container.sh
```diff
@@ -256,8 +280,14 @@ if [[ "${RUN_INTEGRATED_TESTS}" = true ]]; then
   or_die tar cfM ${DATA_EXCHANGE_DIR}/${DATA_BASENAME_WE}.tar --directory ${GEOS_SRC_DIR}    --transform "s/^integratedTests/${DATA_BASENAME_WE}\/repo/" integratedTests
   or_die tar rfM ${DATA_EXCHANGE_DIR}/${DATA_BASENAME_WE}.tar --directory ${GEOSX_BUILD_DIR} --transform "s/^integratedTests/${DATA_BASENAME_WE}\/logs/" integratedTests
   or_die gzip ${DATA_EXCHANGE_DIR}/${DATA_BASENAME_WE}.tar
+
+  # want to clean the integrated tests folder to avoid polluting the next build.
+  or_die integratedTests/geos_ats.sh -a clean
 fi
 
+# Cleaning the build directory.
+or_die ninja clean
```
TotoGaz, line 289: This is most likely superfluous since the build dir is in `/tmp` which is washed away with the `--rm` option of `docker run`. You can keep it if you want, since this may change.

scripts/ci_build_and_test_in_container.sh
```diff
@@ -256,8 +280,14 @@ if [[ "${RUN_INTEGRATED_TESTS}" = true ]]; then
   or_die tar cfM ${DATA_EXCHANGE_DIR}/${DATA_BASENAME_WE}.tar --directory ${GEOS_SRC_DIR}    --transform "s/^integratedTests/${DATA_BASENAME_WE}\/repo/" integratedTests
   or_die tar rfM ${DATA_EXCHANGE_DIR}/${DATA_BASENAME_WE}.tar --directory ${GEOSX_BUILD_DIR} --transform "s/^integratedTests/${DATA_BASENAME_WE}\/logs/" integratedTests
   or_die gzip ${DATA_EXCHANGE_DIR}/${DATA_BASENAME_WE}.tar
+
+  # want to clean the integrated tests folder to avoid polluting the next build.
+  or_die integratedTests/geos_ats.sh -a clean
```
TotoGaz, line 285: With @cssherman 's patch to build the baselines out of source, cleaning the integrated tests will surely not be mandatory anymore, since they disappear with the `docker run --rm`.
Also, we will be able to mount the source `read-only`!

scripts/ci_build_and_test_in_container.sh
TotoGaz, line 256: I'd add a `EXCLUDE_UNIT_TESTS_RE` input in GHA, defaulting to `testUncrustifyCheck|testDoxygenCheck`, so we do not have to hardcode anything.
Combined to giving the number of CPUs as a CLI arguments as well (`NPROC` can be provided through GHA, or defaulted to the results of `nproc` is absent), then we'll remove the need to hard code the name of the self hosted runner.

.github/workflows/ci_tests.yml
```diff
@@ -201,6 +201,8 @@ jobs:
       ENABLE_TRILINOS: OFF
       GCP_BUCKET: geosx/integratedTests
       RUNS_ON: ubuntu-22.04
+#      RUNS_ON: self-hosted
+#      DOCKER_RUN_ARGS: "--cpus=8 --memory=128g"
```
TotoGaz: Remove?

.github/workflows/build_and_test.yml
rrsettgast: We certainly can...and probably should. Would you mind if we did this in a follow up PR. We are getting another workstation to run integratedTests on the CPU, and we will require a generic treatment at that time. I would rather just get this merged as is, and in separate PR's:

1.  clean up and generalize for the new workstation,
2. fix the failing unit tests on GPU, and
3. get integrated tests diffs worked out.

we can do all of these steps independent of each other IMO.

.github/workflows/build_and_test.yml
rrsettgast: Ahhh....I intentionally kept this in case we want to specify the container name and manual cleanup. It seems that adding the `--rm` argument to `docker run` takes care of the cleanup...BUT there are cases when we don't want to clean up. For instance if we remove the `--rm` the container remains and we can run the container from its saved state....which is handy for debugging.

.github/workflows/build_and_test.yml
rrsettgast: see above

scripts/ci_build_and_test_in_container.sh
rrsettgast: I think you are right.

scripts/ci_build_and_test_in_container.sh
rrsettgast: I think this was before the `--rm` command. I can try without.

scripts/ci_build_and_test_in_container.sh
rrsettgast: I would, but I actually don't think we should be excluding unit tests once they are fixed. This is just a hack to get us to the next step (passing tests), after witch this goes away. I actually do want to make it as hard as possible to exclude specific tests 😁

.github/workflows/ci_tests.yml
rrsettgast: this will be used in the follow on the `integratedTests` pr, but it isn't hard to find elsewhere. I can remove it.

.github/workflows/build_and_test.yml
TotoGaz: OK, I'm good with it. We need to be sure we actually have time to revisit this.

.github/workflows/build_and_test.yml
TotoGaz: If you want to keep it, could you indicate your intent as a comment?

.github/workflows/build_and_test.yml
TotoGaz: Then can you add a small comment about it?

scripts/ci_build_and_test_in_container.sh
TotoGaz: Oh, I think @cssherman 's PR has not been merged already. Am I wrong?

.github/workflows/build_and_test.yml
TotoGaz: Also, wouldn't it be simpler to remove the `--rm` argument if we want to keep the docker container?

scripts/ci_build_and_test_in_container.sh
TotoGaz: This makes sense. I kind of prefer having an easy way to do thing with a Cerberus at the code review, but it's OK.
What happens, is that the people maintaining the CI are not the same maintaining the tests or code. So the CI guys gets some ugly code he's not responsible for.

.github/workflows/ci_tests.yml
TotoGaz: Yes, they are just below, lines 251-252.
That would let the commented parts in `build_and_test.yml` only.

scripts/ci_build_and_test_in_container.sh
CusiniM: Can you create an issue pointing at this line though? If someone other than you fixes the tests I am pretty sure they'll forget to remove this.

scripts/ci_build_and_test_in_container.sh
CusiniM: `clean` not `veryclean` ?  I think it will still be better to clean up the integratedTests directory. One may update the version of the baselines... and the clean also removes the tests results which I think we do want to remove IMO. Unless the rm options cleans all that for us.

TotoGaz: An option can be to mount the workspace read-only in the container, then copy it to a temporary and work in it (waiting for an out-of-source integrated tests which should solve the last item which touches the src dir).
Then everything would be cleaned by the `docker run --rm`.

Eventually, we'll be able to skip the workspace copy, but not right now.

CusiniM (approved):

src/coreComponents/constitutive/solid/SolidBase.hpp
CusiniM: is there any way to automatically check for this kind of bugs? Any way to at least get a warning whenever a reference is captured by value in a raja `forAll`?

sframba:
@rrsettgast as for the `testLifoStorage`, we plan on fixing it on CUDA in the not-too-distant future. It's supposed to work but it's broken right now.
Btw we had `testExternalSolvers` fail on Pangea3 too, with one single test failing (`Hypre/SolverTestLaplace2D/0.GMRES_ILU`, see [this discussion](https://github.com/GEOS-DEV/thirdPartyLibs/pull/253#issuecomment-1869700641) with @victorapm) . We are available to do more tests if needed

scripts/ci_build_and_test_in_container.sh
rrsettgast: well....that is a good enough reason.

PR #2878, opened 2023-12-08
feat: Simplification & consolidation of CellElementRegion definition

### LINKED ISSUE #2223: Simplify definition of regions in the XML
As a user, I want to define easily regions from their name in the subsurface model.

A non exhaustive list of subitems are:
   - the different cell types are managed by GEOSX 
   - the user manages  meaninful regions names that correspond to the subsurface model instead of an Attribute value
   - the user can create regions that do not have a physical meaning (rock type, fluid type) but e.g. border between countries
   

The XML file is very complex to write and there is no check of its validity.
GEOSX might crash without notice and no error message.

Check of validity of the XML 
Clear error messages to the user
Replace entries that have a meaning for the developers by entries that have a meaning for the users to define Regions

MelReyCG:
This PR comes after the #3247.

------------------

In order to simplify the `cellBlocks` referencing and thus simplify user workflows, this PR propose the following functionalities:

------------------

1. Possibility to reference all `cellBlock` from a `regionAttribute` at once, without listing all primitives:
```XML
    <CellElementRegion
      name="burdens"
      cellBlocks="{ 1, 4, 5 }"
      materialList="{}" />
```
instead of:
```XML
    <CellElementRegion
      name="burdens"
      cellBlocks="{ 1_tetrahedra, 1_pyramids, 1_hexahedra, ...
                    4_tetrahedra, 4_pyramids, 4_hexahedra, ...
                    5_tetrahedra, 5_pyramids, 5_hexahedra, ... }"
      materialList="{}" />
```

------------------

2. Possibility to reference all cellBlocks with `fnmatch` patterns:
```XML
    <CellElementRegion
      name="reservoir"
      cellBlocks="{ * }"
      materialList="{}" />
```
```XML
    <CellElementRegion
      name="reservoir"
      cellBlocks="{ [1-4]_* }"
      materialList="{}" />
```

------------------

3. This PR also improved the `cellBlock` errors:
- Improved the error when a `cellBlock` is referenced twice
- Added an error when the `cellBlockAttributeValues`, `cellBlocks`, `cellBlocksMatch` attributes are mixed (allowing only one way to select `cellblocks`).
- Improved the error when a referenced `cellBlock` is missing,
- Improved the error when failing to select `cellBlocks`.
See the form and more details here: [new cellBlocks errors.pdf](https://github.com/GEOS-DEV/GEOS/files/13615703/new.cellBlocks.errors.pdf) (not up to date).

------------------

This PR also:
- Updates the `Sphinx` documentation page as it was incomplete, by providing the whole necessary python package list and a part to describe how to generate the `.rst` class tables.
- Adds some `.rst` tables that were not previously pushed (forgotten?).


inputFiles/compositionalMultiphaseWell/benchmarks/Class09Pb3/class09_pb3_hystRelperm_direct_base.xml
```diff
@@ -51,7 +51,7 @@
   <ElementRegions>
     <CellElementRegion
       name="reservoir"
-      cellBlocks="{ 1_hexahedra }"
+      cellBlocks="{ reservoirCells }"
```
TotoGaz: Is this the name of the "region" defined in the `VTK` mesh?

src/coreComponents/mesh/ElementRegionManager.cpp
```diff
@@ -152,6 +154,59 @@ void ElementRegionManager::generateMesh( CellBlockManagerABC const & cellBlockMa
   } );
 
 }
+void ElementRegionManager::checkSubRegionRegistering( Group const & cellBlocks )
+{
+  // Each cellBlock must be referenced by exactly one sub region.
```
TotoGaz: Wasn't this already granted by design?

src/coreComponents/mesh/CellElementRegion.cpp
```diff
@@ -35,15 +45,108 @@ CellElementRegion::CellElementRegion( string const & name, Group * const parent
...
+    return n == 1;
+  }
+  else
+  {
+    return false;
+  }
+}
+
+void registerSubRegion( CellElementRegion const & region, Group & elementSubRegions,
+                        CellBlockABC const & cellBlock )
```
TotoGaz: This looks like this should be a member function.

src/coreComponents/mesh/generators/CellBlockABC.hpp
```diff
@@ -52,12 +52,23 @@ class CellBlockABC : public dataRepository::Group
...
+  /**
+   * @return a string specifying the region attribute
+   */
+  virtual string_view getRegionName() const = 0;
+
+  /**
+   * @brief Set this cellBlock regionName.
+   * @param regionName A string specifying the region attribute.
+   */
+  virtual void setRegionName( string_view regionName ) = 0;
```
TotoGaz: There's no reason for a setter to be defined here since this is exposed to consumers that should only `get`.

src/coreComponents/mesh/generators/CellBlockABC.hpp
```diff
@@ -52,12 +52,23 @@ class CellBlockABC : public dataRepository::Group
...
+   * @return the type of element in this subregion
    *
    * See class FiniteElementBase for possible element type.
    */
   virtual ElementType getElementType() const = 0;
 
+  /**
+   * @return a string specifying the region attribute
+   */
+  virtual string_view getRegionName() const = 0;
```
TotoGaz: A `CellBlock` is mainly a container of cells. There's no reason why a `CellBlock` should have knowledge of some `region` (I guess a compound of `CellBlocks`) above it.

src/coreComponents/mesh/generators/CellBlockManager.hpp
```diff
@@ -165,10 +165,11 @@ class CellBlockManager : public CellBlockManagerABC
 
   /**
    * @brief Registers and returns a cell block of name @p name.
-   * @param name The name of the created cell block.
+   * @param name The name of the cell block containing region (typically the region attribute).
+   * @param name The unique name of the cell block Group.
    * @return A reference to the new cell block. The CellBlockManager owns this new instance.
    */
-  CellBlock & registerCellBlock( string const & name );
+  CellBlock & registerCellBlock( string const & regionName, string const & groupName );
```
TotoGaz: ~~If you intend to group `CellBlock` at that level of the architecture (to be discussed), how do you manage `CellBlocks` that will be used for multiple regions?~~ EDIT: this is not allowed.
Important: the `mesh/generator` "module" is not business oriented: it has no idea of regions, solvers, etc. It's just serving a mesh, and we must think on how we want this to be served. But in the end, it's only a mesh.

src/coreComponents/constitutive/docs/Constitutive.rst
```diff
@@ -49,7 +49,7 @@ A typical ``<Constitutive>`` and ``<ElementRegions>`` block will look like:
       <!--Add water to the material list for region 1-->
       <CellElementRegion
          name="region1"
-         cellBlocks="{ hexahedra, wedges, tetrahedra, pyramids }"
+         cellBlocks="{ all }"
```
TotoGaz: **If we want to go that way**, personally I'm in favor of being able to use `fnmatch` and/or `regex` to help with this.
That way, the `fnmatch` wildcard `*` can be used to gather all the cell blocks, but we can do other less obvious selections if we have to.

Ideally, calling the attributes provided by the mesh instances looks like an ideal solution (like we did for `yaml` [here](https://github.com/GEOS-DEV/GEOS/pull/2988), but I do not know to which extent we can have this merged separately).

This may remove the need of any selection by the user?

It looks like you may already allow this kind of pattern, but what do you think of having something _in the spirit_ of
```yaml
mesh:
  internal:
    element_type: hexahedra
    x_range: [ 0, 0.5, 1 ]
    y_range: [ 0, 0.4, 1 ]
    z_range: [ 0, 0.7, 1 ]
    nx: [ 5, 5 ]
    ny: [ 5, 5 ]
    nz: [ 5, 5 ]
    regions:
      mapping:
        domain0:
          - [ 0, 0, 0 ]
          - [ 1, 0, 0 ]
        domain1:
          - [ 0, 1, 1 ]
          - [ 1, 1, 1 ]
```
(note the `regions` entry), or
```yaml
mesh:
  vtk:
    file_name: test.vtm
    regions:
      block_name: main
      attribute: marker
      mapping:
        reservoir: 0
        overburden: 1
        underburden: 2
    fractures:
      block_name: fracture
```
(again, note the `regions` entry).

Then the `CellElementRegion` could only select the `regions` it wants (`reservoir`, `domain1`, ...) without having to care about the cellblocks/sub-regions anymore...
In `CellBlockManagerABC`, adding an accessor linking the region name to the proper collection of `CellBlockABC` would do the job I guess?

Doing so, maybe could be even entirely hide the concept of sub-regions to the end user?
I would completely remove the old way to select cell blocks: it can only bring complication.

TotoGaz (commented):
Good that you can start working on this.
You may be interested in https://github.com/GEOS-DEV/GEOS/pull/1963 where we discuss how we can select objects in `geos`.

inputFiles/compositionalMultiphaseWell/benchmarks/Class09Pb3/class09_pb3_hystRelperm_direct_base.xml
MelReyCG: Oh, I didn't noticed the `VTKMesh` version, I just saw the `InternalMesh` one (in `class09_pb3_smoke_3d.xml`).
I'll use the syntax to select all `cellBlocks` here.

src/coreComponents/mesh/ElementRegionManager.cpp
MelReyCG: Currently, referencing a `cellBlocks` in multiple `CellElementRegions` is not supported by `ElementRegionManager::getCellBlockToSubRegionMap()` :
```
***** Rank 0: burdens (doubledsubreg2.xml, l.205), subregion 1_tetrahedra: Cell block at index 3 is mapped to more than one subregion.

Frame 0: geos::ElementRegionManager::getCellBlockToSubRegionMap(geos::CellBlockManagerABC const&) const 
Frame 1: geos::ElementRegionManager::generateMesh(geos::CellBlockManagerABC const&) 
Frame 2: geos::ProblemManager::generateMeshLevel(geos::MeshLevel&, geos::CellBlockManagerABC const&, geos::dataRepository::Group const*, LvArray::ArrayView<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, 1, 0, int, LvArray::ChaiBuffer> const&) 
Frame 3: geos::ProblemManager::generateMesh() 
Frame 4: geos::ProblemManager::problemSetup() 
Frame 5: geos::GeosxState::initializeDataRepository() 
...
```

I agree that this feature should be supported, but I would like to leave that refactoring for later. In the meantime, my goal here is to output a proper error when a user do something that is not allowed by the current implementation.

Would you prefer that I add "TODOs" to keep that in mind ?

src/coreComponents/constitutive/docs/Constitutive.rst
MelReyCG: So in the XML input, we would have something like this ?
```xml
  <Mesh>
    <VTKMesh
      logLevel="1"
      name="mesh"
      file="03c_hexdom_10m.vtu"
      regionAttribute="marker">

      <CellBlocks>
        <CellBlock name="reservoirBlock" attributeValues="{ 2 }" />
        <CellBlock name="burdensBlock" attributeValues="{ 1, 3 }" />

        <!-- This one could "re-select" some cells without causing any issue (and be used for stats as an example) -->
        <CellBlock name="allBlocks" attributeValues="{ * }" />
      </CellBlocks>
    </VTKMesh>
  </Mesh>

  <ElementRegions>
    <CellElementRegion
      name="reservoir"
      cellBlocks="{ reservoirBlock }"
      materialList="{ rock, fluid, relperm }" />
    <CellElementRegion
      name="burdens"
      cellBlocks="{ burdensBlock }"
      materialList="{ rock }" />
  </ElementRegions>
```
(and the `InternalMesh` XML would not change ?)

TotoGaz: The problem with having a nested `CellBlocks` entry is that it implies a sub-`Group` in the `VTKMesh`, which may not be what we want. So we should most probably stick to a "hand-made map". (Plus, I would not call it `CellBlocks` but something like `Regions`, but we can discuss that later on).

Despite this small details, yes, that could be the idea.

`InternalMesh` would also have to change in order to cherry-pick which cells blocks we want in which region (currently there's an implicit naming convention that allows that, you can look at the `staircase` cases). But we should manage a proper default.

EDIT: And I would change `cellBlocks` in the `CellElementRegion` nodes as well. But another detail.

@castelletto1 Any thoughts?

MelReyCG: We should indeed remove this `CellBlocks` nested node to have directly the `CellBlock` nodes in `VTKMesh`, as it is done with the `InternalWell` nodes.

inputFiles/singlePhaseFlow/FieldCaseTutorial3_base.xml
```diff
@@ -5,12 +5,12 @@
...
     <CellElementRegion
       name="Reservoir"
-      cellBlocks="{ 2_tetrahedra }"
+      cellBlocks="{ 2 }"
       materialList="{ water, rock }"/>
 
     <CellElementRegion
       name="Burden"
-      cellBlocks="{ 1_tetrahedra, 3_tetrahedra }"
+      cellBlocks="{ 1, 3 }"
```
TotoGaz, line 13: How is this working. Shouldn't is use `cellBlockAttributeValuesString` instead?

inputFiles/compositionalMultiphaseWell/simpleCo2InjTutorial_base.xml
```diff
@@ -58,7 +58,7 @@
   <ElementRegions>
     <CellElementRegion
       name="reservoir"
-      cellBlocks="{ cellBlock }"
+      cellBlocks="{ all }"
```
TotoGaz: I do not quite get the interest of having a dedicated keyword when we can have `cellBlocksMatch="{ * }"` with the same result and less corner casees?

src/coreComponents/mesh/CellElementRegion.cpp
```diff
@@ -35,15 +72,157 @@ CellElementRegion::CellElementRegion( string const & name, Group * const parent
...
+  CellElementSubRegion & subRegion =
+    elementSubRegions.registerGroup< CellElementSubRegion >( cellBlock.getName() );
+
+  subRegion.copyFromCellBlock( cellBlock );
+}
+
+string CellElementRegion::getCellBlockAttributeValue( string_view cellBlockName )
+{
+  size_t const sep = cellBlockName.find( '_' );
+  return sep != string::npos ? string( cellBlockName.substr( 0, sep ) ) : string( "" );
```
TotoGaz: Can't you use the `stringutils::tokenize` the `cellBblockName` and return the first one if `size > 1` ?

src/coreComponents/mesh/ElementRegionManager.cpp
```diff
@@ -152,6 +154,78 @@ void ElementRegionManager::generateMesh( CellBlockManagerABC const & cellBlockMa
...
+  // region that contain a cellBlock (name)
+  std::map< string, CellElementRegion const * > cellBlocksRegion;
+
+  // Let's find out which CellElementSubRegion is using each source cellBlock.
+  forElementRegions< CellElementRegion >( [&]( CellElementRegion & region ) {
+    region.forElementSubRegions< CellElementSubRegion >( [&]( CellElementSubRegion & subRegion ) {
+      string const & subRegionName = subRegion.getName();
+
+      // if no region already referenced the cellBlock
+      if( cellBlocksRegion.find( subRegionName ) == cellBlocksRegion.end() )
```
TotoGaz: Use `std::map::insert` instead?

src/coreComponents/mesh/CellElementRegion.cpp
```diff
@@ -35,15 +72,157 @@ CellElementRegion::CellElementRegion( string const & name, Group * const parent
...
+      {
+        // if the pattern matches the tested cellBlock name
+        if( fnmatch( matchPattern.c_str(), cellBlock.getName().c_str(), 0 ) == 0 )
+        {
+          matching = true;
+          selectedCellBlocks.insert( cellBlock.getName() );
+        }
+      } );
+
+      // a) check for the case where the user is requesting attribute values
```
TotoGaz: Disputable, but what about a
```c++
if( matching ) continue;
```
?

src/coreComponents/mesh/ElementRegionManager.cpp
```diff
@@ -152,6 +154,78 @@ void ElementRegionManager::generateMesh( CellBlockManagerABC const & cellBlockMa
   } );
 
 }
+void ElementRegionManager::checkSubRegionRegistering( CellBlockManagerABC const & cellBlockManager )
```
TotoGaz: In `CellElementRegion::computeSelectedCellBlocks`, you already compute the `std::set` of the selected `cellBlocks`. In the same time, could you gather the `std::set` of all the available `cellBlocks` and check that `available == selected`?
For the duplication of selection, instead of storing the selected cell blocks in a flat `std::set`, use a `std::map` that binds the `matcher` to the `std::set` that matches. Then you have a finer grain that you can use to verify that `std::set_intersection` is empty before merging?

inputFiles/singlePhaseFlow/FieldCaseTutorial3_base.xml
MelReyCG: Indeed, I forgot to change this example!

inputFiles/compositionalMultiphaseWell/simpleCo2InjTutorial_base.xml
MelReyCG: I forgot to change this one too!

I just looked in all examples, and I'm wondering if I should change all exemples that want to select all mesh in a region to `cellBlocksMatch="{ * }"`. What do you think?

MelReyCG: Similar question, should I change all examples to use `cellBlockAttributeValues` when it is possible? (like in `inputFiles/poromechanics/nonlinearAcceleration/smallEggModel/smallEggModel.xml`)

src/coreComponents/mesh/CellElementRegion.cpp
MelReyCG: Nice suggestion to look into `stringutilities`, I found `stringutilities::removeStringAndFollowingContent()` that seems perfect here.

src/coreComponents/mesh/ElementRegionManager.cpp
MelReyCG: Would `emplace` do the job?
```c++
string const & subRegionName = subRegion.getName();
const auto [regionIterator, isNewRegion] = cellBlocksRegion.emplace( subRegionName, &region );
// if no region already referenced the cellBlock
if( !isNewRegion )
{
  CellElementRegion const & otherRegion = iterator->second;
```

src/coreComponents/mesh/CellElementRegion.cpp
MelReyCG: I usually try to avoid them but here it seems bound to the errors.
And what do you think of a:
```c++
if( !matching )
{
  // error checking
}
```
It adds the brackets lines but prevent anyone to fail to write code under this error checking part. I hope I'm being clear but the idea is that I try to avoid `continue` / `break` as much as possible in long loops.

TotoGaz: That does not really matter. It's just that the `matching` boolean was filled but it's usage was a bit hidden.
As you wish.

inputFiles/compositionalMultiphaseWell/simpleCo2InjTutorial_base.xml
TotoGaz: @castelletto1 I think you should have a look at this PR (or at least at what it implies) since it can change a lot our `xml` input files (for the better :wink:).
Could you answer @MelReyCG 's question (above in this thread) about spreading the changes to our integrated tests? I think that should be a good idea, but I'm not 100% sure.

@MelReyCG could you update the PR description so we can use it for the announce of the change (I think the people have the right to know!).
Could you also tell me where the `all` keyword comes from? Did you add it yourself? I did not find it in your PR. 🤷

src/coreComponents/mesh/CellElementRegion.cpp
TotoGaz: Great, I did not know about it 🤣

src/coreComponents/mesh/ElementRegionManager.cpp
TotoGaz: Yes, it should be equivalent. The idea is that you'll search only once (and if you insert twice, then you exhibit an error so we don't car).

src/coreComponents/mesh/ElementRegionManager.cpp
MelReyCG: Do you think that I should store the selected cellblocks in a dedicated `std::set` attribute, and access to it in `ElementRegionManager::checkSubRegionRegistering` ?

inputFiles/compositionalMultiphaseWell/simpleCo2InjTutorial_base.xml
MelReyCG: About the `all`, I just forgot to change the `cellBlocks="{ all }"` to `cellBlocksMatch="{ * }"` in that example.

src/coreComponents/mesh/ElementRegionManager.cpp
TotoGaz: I would have to double check, but I'm not sure that you have to store them (as an attribute).
Once you've done your allocation (through the function), keep track of it in the function and perform the check at the exact same moment?

src/coreComponents/codingUtilities/StringUtilities.hpp
```diff
@@ -180,8 +180,8 @@ string_view trimSpaces( string_view str );
  * @param[in] strToRemove the string to search for in the line
  * @return the new (truncated) string
  */
-string removeStringAndFollowingContent( string const & str,
-                                        string const & strToRemove );
+string_view removeStringAndFollowingContent( string_view str,
+                                             string_view strToRemove );
```
TotoGaz: Please update the doc by adding that we get a `view` onto the subpart of interest of `str`.
So if you modify `str` then the returned result will be modified as well.
In case that's an issue, then we should copy.

inputFiles/compositionalMultiphaseWell/simpleCo2InjTutorial_base.xml
TotoGaz: So I've discussed with @castelletto1 and yes, please modify the integrated tests accordingly.

MelReyCG: @TotoGaz I did changed the doc and all tests in the GEOS repo accordingly 👍

MelReyCG: I think we need a `require baseline` label here but I am not sure.

src/coreComponents/mesh/ElementRegionManager.cpp
MelReyCG: So:
- moved the duplication detection as soon as possible, in `CellElementRegion::generateMesh()`,
- to keep track of the other matching cellblock, changed the `CellElementRegion::generateMesh()` signature to compare / modify a `std::map< string, CellElementRegion const * > & cellBlocksMatchers` between calls,
- at `ElementRegionManager`, we now only have the "orphan" check.

src/coreComponents/LvArray
TotoGaz, line 1: Maybe a merge to finish?

src/docs/sphinx/developerGuide/Contributing/Sphinx.rst
```diff
@@ -5,24 +5,26 @@ Sphinx Documentation
 Generating the documentation
 ====================================
 
-- To generate the documentation files, you will need to install Sphinx using
+- To generate the documentation files, you will need to install Sphinx using :
```
TotoGaz:
```suggestion
- To generate the documentation files, you will need to install Sphinx using:
```

src/docs/sphinx/developerGuide/Contributing/Sphinx.rst
```diff
@@ -5,24 +5,26 @@ Sphinx Documentation
...
 
   .. code-block:: sh
 
-   sudo apt install python-sphinx
+    pip install sphinx
+    pip install sphinx-design sphinx-argparse sphinxcontrib-plantuml sphinxcontrib.programoutput sphinx_rtd_theme
+    pip install scipy
```
TotoGaz:
```suggestion
    python -m pip install sphinx
    python -m pip install sphinx-design sphinx-argparse sphinxcontrib-plantuml sphinxcontrib.programoutput sphinx_rtd_theme
    python -m pip install scipy
```

src/docs/sphinx/developerGuide/Contributing/Sphinx.rst
```diff
@@ -5,24 +5,26 @@ Sphinx Documentation
...
 
-  Then you can generate the documentation files with the following command
+- Then you can generate the documentation files with the following command :
```
TotoGaz:
```suggestion
- Then you can generate the documentation files with the following command:
```

src/docs/sphinx/developerGuide/Contributing/Sphinx.rst
```diff
@@ -35,3 +37,20 @@ like those describing a specific class, can instead be found in ``docs`` subdire
...
+
+To describe the GEOS classes, some parts of the Sphinx documentation are automatically filled with
+generated ``.rst`` tables. Those tables are filled by the registered ``Group`` & ``Wrapper``
+configuration (with ``Group::registerWrapper()``, ``Wrapper::setDescription()``, ...).
+
+When modifying a ``Wrapper`` configuration, one should call ``make geosx_update_rst_tables`` to
+generate the tables, and should then include the updated / added tables in its PR, exactly the
+same way the ``schema.xsd`` is maintained with ``make geosx_generate_schema``.
+
+To make these commands work, the folowing line is needed in the host-config :
```
TotoGaz:
```suggestion
To make these commands work, the folowing line is needed in the host-config:
```

src/coreComponents/unitTests/wellsTests/testReservoirCompositionalMultiphaseMSWells.cpp
```diff
@@ -118,7 +118,7 @@ char const * xmlInput =
     </NumericalMethods>
     <ElementRegions>
       <CellElementRegion name="Region1"
-                         cellBlocks="{cb1}"
+                         cellBlocksMatch="{ * }"
```
TotoGaz: Make sure to keep a few tests with the legacy `cellBlocks="{cb1}"` so we do not break the feature.

src/coreComponents/mesh/CellElementRegion.cpp
```diff
@@ -35,16 +73,163 @@ CellElementRegion::CellElementRegion( string const & name, Group * const parent
...
+  // For now, subRegion name must be the same as the cellBlock (so we can match them and reference them in errors).
+  CellElementSubRegion & subRegion =
+    elementSubRegions.registerGroup< CellElementSubRegion >( cellBlock.getName() );
+
+  subRegion.copyFromCellBlock( cellBlock );
+}
+
+string CellElementRegion::getCellBlockAttributeValue( string_view cellBlockName )
+{
+  return string( stringutilities::removeStringAndFollowingContent( cellBlockName, "_" ) );
```
TotoGaz: Isn't this supposed to be an integer?

src/coreComponents/mesh/docs/Mesh.rst
```diff
@@ -189,84 +189,154 @@ The mesh block has the following syntax:
...
 
 .. figure:: mesh_multi.png
    :align: center
    :width: 500
 
-In the example presented above, the mesh is is composed of two regions (*Top* and *Bot*).
-Each region contains 4 ``cellBlocks``.
+In GEOS, there are three different ways to select ``cellBlocks`` in a ``CellElementRegion``:
+- Using ``cellBlockAttributeValues``, a list of the desired ``regionAttribute`` values.
+  I.e. ``"{ 1, 2 }"`` selects the { 1_tetrahedra, 1_pyramid, 2_tetrahedra, 2_pyramid... } cellBlocks.
```
TotoGaz:
```suggestion
  I.e. ``"{ 1, 2 }"`` selects the ``{ 1_tetrahedra, 1_pyramid, 2_tetrahedra, 2_pyramid... }`` cellBlocks.
```

src/coreComponents/mesh/docs/Mesh.rst
```diff
@@ -189,84 +189,154 @@ The mesh block has the following syntax:
...
+- Using ``cellBlocks``, a list of the desired ``cellBlocks`` names from the mesh to contain in this
+  CellElementRegion. Naming is detailed bellow.
+- Using ``cellBlocksMatch``, a list of `fnmatch patterns <https://metacpan.org/pod/File::FnMatch>`_ to match cellBlock names to add them in this CellElementRegion.
+  I.e. ``{ 1_* }`` selects the { 1_tetrahedra, 1_pyramid, 1_hexahedra... } cellBlocks, ``{ * }`` selects every elements.
```
TotoGaz:
```suggestion
  I.e. ``{ 1_* }`` selects the ``{ 1_tetrahedra, 1_pyramid, 1_hexahedra... }`` cellBlocks, ``{ * }`` selects every elements.
```

src/coreComponents/mesh/CellElementRegion.cpp
```diff
@@ -35,16 +73,163 @@ CellElementRegion::CellElementRegion( string const & name, Group * const parent
...
-    CellElementSubRegion & subRegion = elementSubRegions.registerGroup< CellElementSubRegion >( cellBlockName );
-    CellBlockABC const & source = cellBlocks.getGroup< CellBlockABC >( subRegion.getName() );
-    subRegion.copyFromCellBlock( source );
+    attributeValues.insert( CellElementRegion::getCellBlockAttributeValue( cellBlockName ) );
   }
+  return attributeValues;
+}
+
+std::set< string >
+CellElementRegion::computeSelectedCellBlocks( std::set< string > const & cellBlocksNames ) const
```
TotoGaz: You have errors checks at a _lot_ of different places and eventually I really have hard time following the flow of what's going on.
Could you simplify the code into having a `stream/iterable` of `CellBlockABC` that you filter in a generic way using a `predicate` that you'll build from the user input.
I think that will make the code a lot simpler without degrading the use UX.
Otherwise, I think we're good. 😉

src/coreComponents/mesh/CellElementRegion.cpp
MelReyCG: I though I could use strings to store the region attribute for these reasons:
- we do not really need integers here, there are no numerical operation at all,
- to avoid unnecessary parsing / formatting operations,
- to support text region names in the future.

src/coreComponents/unitTests/wellsTests/testReservoirCompositionalMultiphaseMSWells.cpp
MelReyCG: Of course!
I am testing all configurations at `src\coreComponents\unitTests\meshTests\testElementRegions.cpp.cpp:126`, but as it is only a mid-level test, I can keep an explicit syntax a high-level test like this one.

src/docs/sphinx/developerGuide/Contributing/Sphinx.rst
MelReyCG: I also forgot a `l` at `following`, Oops!

paveltomin:
@MelReyCG are you planning something similar for wells?

MelReyCG:
I have not planned it yet, but I would be interested by what you have in mind on how they could be simplified.

paveltomin:
> I have not planned it yet, but I would be interested by what you have in mind on how they could be simplified.

I don't have a solid idea to be honest, just feel that well definition is very repetitive in case of multiple wells, somewhat similar to what you addressed in this PR

paveltomin:
It still shows as "Review required"

paveltomin:
@MelReyCG and there is some conflict now, please resolve

paveltomin:
@MelReyCG review needed

src/coreComponents/mesh/CellElementRegionSelector.hpp
```diff
@@ -0,0 +1,116 @@
...
+   * @brief Construct a new CellElementRegionSelector.
+   * @param cellBlocks a Group containing all the available cell-blocks.
+   */
+  CellElementRegionSelector( dataRepository::Group const & cellBlocks );
+
+  /**
+   * @brief Select the user-requested mesh cell-blocks for the specified cell-blocks.
+   * @throw an InputError if the selection is incoherent (already selected cell-block, no selection...).
+   * @param region the region for which we want to select the cell-blocks.
+   * @return the selected cell-blocks.
```
TotoGaz:
```suggestion
   * @return the selected cell-blocks names.
```

src/coreComponents/mesh/CellElementRegionSelector.cpp
```diff
@@ -0,0 +1,251 @@
...
+  }
+  return oneByOneCellBlocks;
+}
+
+
+template< typename T >
+std::set< T > toStdSet( arrayView1d< T const > const & array )
+{
+  return std::set< T >( array.begin(), array.end() );
+}
```
TotoGaz: Consider making it a `lambda` directly in `CellElementRegionSelector::selectRegionCellBlocks` but as you want.

src/coreComponents/mesh/CellElementRegionSelector.cpp
```diff
@@ -0,0 +1,251 @@
...
+  // only one selection method allowed for now
+  GEOS_THROW_IF( ( !requestedAttributeValues.empty() + !requestedMatchPatterns.empty() + !requestedCellBlock.empty() ) != 1,
+                 GEOS_FMT( "{}: Only one setting must be used to select cellBlocks ({}, {} or {}).",
+                           region.getDataContext(),
+                           ViewKeys::sourceCellBlockNamesString(),
+                           ViewKeys::cellBlockAttributeValuesString(),
+                           ViewKeys::cellBlockMatchPatternsString() ),
+                 InputError );
+
+  std::set< string > const matchPatterns = getFNMatchPatterns( region, requestedAttributeValues, requestedMatchPatterns );
```
TotoGaz: IIUC, you are building the `matchPatterns` based on `requestedAttributeValues` xor `requestedMatchPatterns` (because you do not allow both to coexist).
Could you add a `doxygen` for this function?
Could you also rename it something like `buildFNMatchPatterns` which emphasizes a little processing involved? That would have helped me.

src/coreComponents/mesh/CellElementRegionSelector.cpp
```diff
@@ -0,0 +1,251 @@
...
+  std::set< string > const matchedCellBlocks = getFNMatchSelection( region, matchPatterns );
+  std::set< string > cellBlocksSelection = getOneByOneSelection( region, requestedCellBlock );
+  cellBlocksSelection.insert( matchedCellBlocks.begin(), matchedCellBlocks.end() );
```
TotoGaz: Are you relying on the fact that `cellBlocksSelection` and `matchedCellBlocks` will be exclusively empty to be sure it's one or the other?

src/coreComponents/mesh/CellElementRegionSelector.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+  selectRegionCellBlocks( region, requestedAttributeValues, cellBlocksSelection );
```
TotoGaz: Why do you have to give `requestedAttributeValues` again if you already used it line 169 at the `getFNMatchPatterns` call?

src/coreComponents/mesh/CellElementRegionSelector.cpp
```diff
@@ -0,0 +1,251 @@
...
+  GEOS_ERROR_IF( cellBlock == nullptr,
+                 GEOS_FMT( "{}: Unexpected error, {} not found.\nAvailable cell-block list: {{ {} }}",
+                           region.getWrapperDataContext( ViewKeys::sourceCellBlockNamesString() ),
+                           cellBlockName, stringutilities::join( m_cellBlockNames, ", " ) ) );
+
+  return *cellBlock;
+}
+
+
+void CellElementRegionSelector::checkForNoOrphanCellBlocks() const
```
TotoGaz: To find orphan cell blocks, could it be easier to loop over the `set` of the available cell blocks, and then check if it matches or not? If a cell block does not match, it's an orphan, otherwise you store it?

EDIT: Maybe could you create a `cell block name -> regions` that you would initiate with all the cell blocks available.
Then every time a cell block is used, you populate the `regions` container.
In the end, you would just have to post process the `map` to find inconsistencies.
- The `regions` container is empty? -> orphan.
- Its size is > 1? -> duplicate.

That would easier?

src/coreComponents/mesh/CellElementRegion.cpp
```diff
@@ -35,17 +74,34 @@ CellElementRegion::CellElementRegion( string const & name, Group * const parent
...
+
+  m_cellBlockNames.clear();
+  m_cellBlockNames.insert( 0, selectedCellBlocks.cbegin(), selectedCellBlocks.cend() );
+
+  for( string const & cellBlockName : m_cellBlockNames )
   {
-    CellElementSubRegion & subRegion = elementSubRegions.registerGroup< CellElementSubRegion >( cellBlockName );
-    CellBlockABC const & source = cellBlocks.getGroup< CellBlockABC >( subRegion.getName() );
-    subRegion.copyFromCellBlock( source );
+    registerSubRegion( cellBlockSelector.getCellBlock( *this, cellBlockName ) );
```
TotoGaz: We have a cyclic dependency here. `CellElementRegionSelector` depends on `CellElementRegion` which depends on `CellElementRegionSelector`.

src/coreComponents/mesh/CellElementRegionSelector.hpp
```diff
@@ -0,0 +1,116 @@
...
+   * @todo For now, multiple regions per cell is not supported (ElementRegionManager::getCellBlockToSubRegionMap()).
+   *       We could refactor the CellElementRegion & Mesh classes so regions are mapped to cell-blocks IN the mesh (and potencially
```
TotoGaz:
```suggestion
   *       We could refactor the CellElementRegion & Mesh classes so regions are mapped to cell-blocks IN the mesh (and potentially
```

TotoGaz (commented):
It is better that you have added a `CellElementRegionSelector`, that makes the process clearer and does not bloat `CellElementRegion` with something which is only used during the initialization phase.

src/coreComponents/mesh/CellElementRegionSelector.cpp
MelReyCG: At this point, this is not a problem. The prohibition on using multiple parameters at the same time is implemented in `CellElementRegionSelector::selectRegionCellBlocks()`.

MelReyCG: I did this because, as I understood it, I had to avoid making branches, and the information would have to cascade through the same "pipes".

src/coreComponents/mesh/CellElementRegionSelector.cpp
MelReyCG: I implemented it that way, I did not address all your other comments for now. Don't hesitate to tell me if I go in the right direction.

MelReyCG: @TotoGaz

MelReyCG:
@paveltomin I share your feel, I've been told that someone else is working on the wells so I wait to see if the xml format changes.
I made changes on this PR following the lastest Thomas review, I'll merge again with develop when I'm done with it.

src/coreComponents/mesh/CellElementRegionSelector.cpp
MelReyCG: I prefer to keep the collected attribute values to do a more complete error report later in `checkSelectionConsistency()` (previously `checkOrphanCellBlocks()`). I've renamed this `selectRegionCellBlocks()` method to `registerRegionSelection()` as its goal is to effectively register the selection for the region. I've also renamed the other `selectRegionCellBlocks()` homolog to `buildRegionCellBlocksSelection()`. I think this is a lot clearer now, naming things is not that easy ! :D

MelReyCG:
Ok, I think I addressed all your comments @TotoGaz.

src/coreComponents/mesh/CellElementRegionSelector.cpp
```diff
@@ -0,0 +1,265 @@
...
+  // user region attribute matching patterns creation
+  for( integer attributeValue : requestedAttributeValues )
+  {
+    string attributeValueStr = std::to_string( attributeValue );
+    // if attribute value does not exist in the mesh
+    GEOS_THROW_IF( m_regionAttributeOwners.count( attributeValueStr ) == 0,
+                   GEOS_FMT( "{}: Region attribute value '{}' not found.\nAvailable region attribute list: {{ {} }}",
+                             region.getWrapperDataContext( ViewKeys::cellBlockAttributeValuesString() ),
+                             attributeValueStr,
+                             joinMapKeys( m_regionAttributeOwners, ", " ) ),
```
TotoGaz:
```suggestion
                             stringutilities::join( mapKeys< std::set >( map ), ", " ) ),
```

src/coreComponents/mesh/CellElementRegionSelector.cpp
```diff
@@ -0,0 +1,265 @@
...
+    m_cellBlocksOwners[cellBlockName].push_back( &region );
+  }
+}
+
+
+std::set< string >
+CellElementRegionSelector::buildRegionCellBlocksSelection( CellElementRegion const & region )
+{
+  auto toStdSet = []( auto const & array ) {
+    using ArrayElementType = std::decay_t< decltype( *array.data() ) >;
```
TotoGaz: 😎

src/coreComponents/mesh/CellElementRegionSelector.cpp
```diff
@@ -0,0 +1,265 @@
...
+                           ViewKeys::sourceCellBlockNamesString(),
+                           ViewKeys::cellBlockAttributeValuesString(),
+                           ViewKeys::cellBlockMatchPatternsString() ),
+                 InputError );
+
+  std::set< string > const matchPatterns = buildMatchPatterns( region, requestedAttributeValues, requestedMatchPatterns );
+  std::set< string > const matchedCellBlocks = getMatchingCellblocks( region, matchPatterns );
+  std::set< string > cellBlocksSelection = requestedCellBlock;
+  cellBlocksSelection.insert( matchedCellBlocks.begin(), matchedCellBlocks.end() );
+  verifyRequestedCellBlocks( region, cellBlocksSelection );
```
TotoGaz: Can you have both `requestedCellBlock` and `matchedCellBlocks` non empty (regarding the exclusive unique selection method you want)?
I would say that one has to be empty?

src/coreComponents/mesh/CellElementRegionSelector.cpp
```diff
@@ -0,0 +1,265 @@
...
+  {
+    if( owningRegions.size() == 0 )
+    {
+      orphanRegionAttributes.insert( attributeValueStr );
+    }
+    GEOS_THROW_IF( owningRegions.size() > 1,
+                   GEOS_FMT( "The region attribute '{}' has been referenced in multiple {}:\n- {}\n- {}",
+                             attributeValueStr, CellElementRegion::catalogName(),
+                             getDataContextsStr( owningRegions ) ),
+                   InputError );
```
TotoGaz: Would you like to wait for all the "errors" to be found before you log them all?

src/coreComponents/mesh/CellElementRegionSelector.cpp
```diff
@@ -0,0 +1,265 @@
...
+  {
+    if( owningRegions.size() == 0 )
+    {
+      orphanCellBlockNames.insert( cellBlockName );
+    }
+    GEOS_THROW_IF( owningRegions.size() > 1,
+                   GEOS_FMT( "The cellBlock '{}' has been referenced in multiple {}:\n- {}\n- {}",
+                             cellBlockName, CellElementRegion::catalogName(),
+                             getDataContextsStr( owningRegions ) ),
+                   InputError );
```
TotoGaz: Same

src/coreComponents/mesh/ElementRegionManager.cpp
```diff
@@ -119,12 +120,26 @@ void ElementRegionManager::setSchemaDeviations( xmlWrapper::xmlNode schemaRoot,
...
-  {
-    elemRegion.generateMesh( cellBlockManager.getCellBlocks() );
-  } );
+  { // cellBlocks loading
+    Group const & cellBlocks = cellBlockManager.getCellBlocks();
+    CellElementRegionSelector cellBlockSelector{ cellBlocks };
+    this->forElementRegions< CellElementRegion >( [&]( CellElementRegion & elemRegion )
+    {
+      std::set< string > selectedCellBlocks = cellBlockSelector.buildRegionCellBlocksSelection( elemRegion );
+      elemRegion.clearCellBlockNames();
```
TotoGaz: For info, why do you need to clean? It may already contain something?

TotoGaz (approved):

src/coreComponents/mesh/CellElementRegionSelector.cpp
MelReyCG: Indeed, with the constraints on the multiple inputs, only one would be filled.
- This code would not change if the inputs constraints would change,
- Plus, I did this because, as I understood it, I had to avoid making branches, and the information would have to cascade through the same "pipes".

src/coreComponents/mesh/CellElementRegionSelector.cpp
MelReyCG: I propose adding a `stringutilities::joinLambda( iteratorReadingLambda, list, delimiter )` function because I oftenly find myself creating new local functions to create vectors/sets to feed the `stringutilities::join()` function. I would prefer to give a lambda to give the way to obtain the values from the iterator.

src/coreComponents/mesh/ElementRegionManager.cpp
MelReyCG: Indeed, the goal here were to clear all user input lists (`m_cellBlockAttributeValues`, `m_cellBlockMatchPatterns`, `m_cellBlockNames`) and take into account the selection built here (to store it in `m_cellBlockNames`).
I renamed this method to `clearCellBlockInputs()`, that may be clearer.

src/coreComponents/mesh/CellElementRegionSelector.cpp
MelReyCG: Good suggestion!

MelReyCG:
@TotoGaz I don't know if you still can, but could you look at my answers of you last review comments?

src/coreComponents/mesh/CellElementRegionSelector.cpp
TotoGaz: I'm not too much into it. There's a `join` version that takes iterators. We should find a way to provide the appropriate iterator instead. Maybe that will be easier with `ranges`? Meanwhile I would keep the extraction logic out of `join`?

src/coreComponents/mesh/CellElementRegionSelector.cpp
TotoGaz: Can you add a comment about it and why then?

src/coreComponents/mesh/ElementRegionManager.cpp
TotoGaz: OK, that was not obvious to me. Could you add a little comment?

TotoGaz:
@CusiniM @castelletto1 You may want to look at this because it can simplify a lot how we handle cell blocks (somehow hiding them, even if it's not entirely the case).

src/coreComponents/mesh/CellElementRegionSelector.cpp
MelReyCG: I think something like `std::views::keys` would be perfect here (but it's only in `C++20` or `range-v3`), but what about the iteration of name / `DataContext` in groups arrays?
I would like to understand what worries you about this approach. To me, giving a functor is near to give iterators, the goal is to give a way to get the data. If we give a functor to `joinLambda()`, is the extraction logic really coupled?

src/coreComponents/mesh/CellElementRegionSelector.cpp
MelReyCG: Done!

src/coreComponents/mesh/CellElementRegionSelector.cpp
MelReyCG: @TotoGaz

TotoGaz: You have a stream of `data`, then you can `transform` it (using `std::views::keys` or any other process for `DataContext` I imagine), then you consume it using the `join`. Your approach would merge the transformation and the consumption.
Or maybe I misunderstand your idea. Do you want to provide a functor that will return a new `string` every time you call it? How do you see the things? I'd say it's not very different from providing an iterator? Could you write pseudo code of what you want to do?

MelReyCG: @TotoGaz Thanks for your answers.

I've implemented it so you can see here:
`src/coreComponents/codingUtilities/StringUtilities.hpp:116`
```c++
/**
 * @brief Join strings or other printable objects returned by a formatter functor.
 * @tparam CONTAINER         type of container to join
 * @tparam S                 type of delimiter, usually char, char const * or string
 * @tparam LAMBDA            type of formatter functor, usually `[]( auto it ) -> string`
 * @param iteratorFormatter  formatter function to get a formattable value from an iterator of the container
 * @param container          container to join
 * @param delim              delimiter used to glue together strings
 * @return a string containing input values concatenated with a delimiter
 */
template< typename CONTAINER, typename S = char, typename LAMBDA >
string joinLamda( LAMBDA iteratorFormatter, CONTAINER const & container, S const & delim = S() )
```

It allows this kind of things:
`src/coreComponents/mesh/CellElementRegionSelector.cpp:124`
```c++
                   GEOS_FMT( "{}: No cellBlock named '{}'.\nAvailable cellBlock list: {{ {} }}",
                             region.getWrapperDataContext( ViewKeys::sourceCellBlockNamesString() ),
                             requestedCellBlockName,
                             stringutilities::joinLamda( []( auto pair ) { return pair->first; },
                                                         m_cellBlocksOwners, ", " ) ),
```
(we can also get the names/data-context of a `Group` list)

As `join()` can take a container or iterator to access the data, `joinLambda()` would take a functor that would give access to the data. I don't know if this is compliant with best practices?

Note that I can put the functor into a dedicated local variable to separate a bit more.

About `std::views`, I'd really like to use them but `C++20` or the `range-v3` library is needed.

_edit: clarifying comments / param names_

MelReyCG:
@rrsettgast 
According to the tests I did, the resulting restart `.hdf5` files are different if we change the ordering of the `cellBlocks`  list. As this PR reorders the cellblocks (sorting them alphabetically by using a temporary `std::set`), this result in failing the `integratedTest`.

The advantage is that reordering the `cellBlocks` will no longer have an effect on the simulation results.

MelReyCG:
Just leaving a note here, the only CI steps that is legitimately failing is the `integratedTest` one. The CUDA builds were working but crash now because of a CI bug since I updated my branch.

paveltomin (approved):

CusiniM (approved):

CusiniM:
this PR is missing one codeowner review.

src/coreComponents/common/format/StringUtilities.hpp
```diff
@@ -66,14 +66,57 @@ string join( IT first, IT last, S const & delim = S() )
...
+ * @tparam S                 type of delimiter, usually char, char const * or string
+ * @tparam LAMBDA            type of formatter functor, usually `[]( auto it ) -> string`
+ * @param iteratorFormatter  formatter function to get a formattable value from a IT iterator
+ * @param first              iterator to start of the range
+ * @param last               iterator past-the-end of the range
+ * @param delim              delimiter used to glue together strings
+ * @return a string containing input values concatenated with a delimiter
+ */
+template< typename IT, typename S, typename LAMBDA >
+string joinLamda( IT first, IT last, S const & delim, LAMBDA iteratorFormatter )
```
untereiner: _joinLambda_
_funcFormatter_ instead of _iteratorFormatter_  because the parameter is a function not an iterator

src/coreComponents/common/format/StringUtilities.hpp
```diff
@@ -66,14 +66,57 @@ string join( IT first, IT last, S const & delim = S() )
...
+ * @tparam CONTAINER         type of container to join
+ * @tparam S                 type of delimiter, usually char, char const * or string
+ * @tparam LAMBDA            type of formatter functor, usually `[]( auto it ) -> string`
+ * @param iteratorFormatter  formatter function to get a formattable value from an iterator of the container
+ * @param container          container to join
+ * @param delim              delimiter used to glue together strings
+ * @return a string containing input values concatenated with a delimiter
+ */
+template< typename CONTAINER, typename S, typename LAMBDA >
+string joinLamda( CONTAINER const & container, S const & delim, LAMBDA iteratorFormatter )
```
untereiner: _joinLambda_
_funcFormatter_ instead of _iteratorFormatter_  because the parameter is a function not an iterator

src/coreComponents/mesh/CellElementRegionSelector.cpp
```diff
@@ -0,0 +1,259 @@
...
+#include <optional>
+
+
+namespace geos
+{
+using namespace dataRepository;
+using ViewKeys = CellElementRegion::viewKeyStruct;
+
+
+std::optional< string > getCellBlockAttributeValue( string_view cellBlockName )
```
untereiner: separator can be a parameter of this function and a member of CellElementRegionSelector class

untereiner (approved):

MelReyCG:
I think it is good @untereiner

dkachuma (approved):

CusiniM:
This is still missing one codeowner review

src/coreComponents/unitTests/meshTests/box_hybrid_mesh.vtu
rrsettgast, line 1: that is a large file. quite large for a unit test.

src/coreComponents/mesh/CellElementRegionSelector.hpp
rrsettgast, line 1: It seems that this could be a more generic capability. We would want to use `fnmatch` for multiple specifications in the input file. So instead of a `CellElementRegionSelector`, we would have a generic `MatchSelector` that could be applied to the contents of a `Group`....either subgroups, wrappers, or sets.

src/coreComponents/mesh/CellElementRegion.cpp
```diff
@@ -17,16 +17,54 @@
...
+                              viewKeyStruct::cellBlockMatchPatternsString() ) );
+
+  registerWrapper( viewKeyStruct::cellBlockAttributeValuesString(), &m_cellBlockAttributeValues ).
+    setInputFlag( InputFlags::OPTIONAL ).
+    setDescription( GEOS_FMT( "List of regionAttribute values for which we want to add the cells in this {}.\n"
+                              "I.e. {{ 1, 2 }} selects the {{ 1_tetrahedra, 1_pyramid, 2_tetrahedra, 2_pyramid... }} cellBlocks.\n"
+                              "This setting cannot be used simultaneously with {} nor {}.",
+                              catalogName(),
+                              viewKeyStruct::sourceCellBlockNamesString(),
+                              viewKeyStruct::cellBlockMatchPatternsString() ) );
```
rrsettgast: With the match pattern option, is this option redundant?

rrsettgast (approved):
I have put in a question about a generic matching capability, but I do have another question about whether or not we should have different xml attributes for the input. By this I mean `cellBlocks` vs `cellBlocksMatch`. Couldn't we just have a single `cellBlocksMatch` input and evaluate the `fnmatch` on that to generate the contents of `cellBlocks`. This way we don't have multiple xml keywords. I think it would work?

src/coreComponents/unitTests/meshTests/box_hybrid_mesh.vtu
MelReyCG: I clipped it in a mesh file I added in this folder : `box_hybrid_mesh_hybrid.vtu`
It is now around 14x smaller in size, and the unit test is 2x faster.

src/coreComponents/mesh/CellElementRegionSelector.hpp
MelReyCG: I think it is the subject of a dedicated refactoring PR but the functionality is interesting. I think the `CellElementRegionSelector` would be still needed to keep the selecting logic out of the cell-blocks management / storing classes.
I added it to our to-do list (@arng40 @paloma-martinez).

src/coreComponents/mesh/CellElementRegion.cpp
MelReyCG: I don't think so because it hides the cell-blocks concept, however I merged it too (see my following message).

MelReyCG:
@rrsettgast 
Following your suggestion, I merged the 3 `cellBlocks`, `cellBlocksMatch`, `cellBlockAttributeValues` attributes behaviour in the original `cellBlocks` one. It is now able to take all types of inputs (exact names, fnmatch patterns, region attribute).
Here are some supported exemples: 
```xml
      <CellElementRegion name="everything" materialList="{ }" cellBlocks="{ * }" />
```
```xml
      <CellElementRegion name="overburden" materialList="{ }" cellBlocks="{ 1, 5 }" />
      <CellElementRegion name="reservoir" materialList="{ }" cellBlocks="{ 2, 6 }" />
```
```xml
      <CellElementRegion name="mixed" materialList="{ }"
                         cellBlocks="{ 1, 2_hexahedra, 2_tetrahedra, 2_pyramids, [5-6]_* }" />
```
```xml
      <CellElementRegion name="legacy" materialList="{ }"
                         cellBlocks="{ 1_hexahedra, 1_tetrahedra, 1_pyramids, 5_hexahedra, 5_tetrahedra, 5_pyramids }" />
```

paveltomin:
> I have not planned it yet, but I would be interested by what you have in mind on how they could be simplified.

how about this situation:
```
<Solvers gravityVector="{0.0, 0.0, -9.81}">
<MultiphasePoromechanicsReservoir name="poromechanicsReservoirSolver" poromechanicsSolverName="poromechanicsSolver" wellSolverName="wellSolver" initialDt="1e6" logLevel="1" targetRegions="{ Domain, WELL.W001, WELL.W002, WELL.W003, WELL.W004, WELL.W005 }">
```
it would be nice to be able to use patterns like that:
```
<Solvers gravityVector="{0.0, 0.0, -9.81}">
<MultiphasePoromechanicsReservoir name="poromechanicsReservoirSolver" poromechanicsSolverName="poromechanicsSolver" wellSolverName="wellSolver" initialDt="1e6" logLevel="1" targetRegions="{ Domain, WELL.* }">
```
i am sure users would appreciate

MelReyCG:
Sure! We will add a next PR will propose this pattern feature to more fields.

MelReyCG:
By using a modified version of `restart_check.py` that allowed me to display the text arrays and indicate that they were only reordered, I think I can safely say that we can rebaseline, and add this in `BASELINE_NOTES.md`:
> PR #2878 (2024-10-17)
=====================
Sorted region cellBlocks names alphabetically. Therefore affected ordering of: faceManager/elemSubRegionList, nodeManager/elemList, nodeManager/elemSubRegionList, SurfaceElementSubRegion::fractureElementsToCellSubRegions, field::perforation::reservoirElementSubregion.

I also changed the tag in `.integrated_tests.yaml`.

I would say that this PR is now ready to be merged.

MelReyCG:
I will open a PR to propose those changes to `restart_check.py`.

MelReyCG:
By using a modified version of `restart_check.py` that allowed me to display the text arrays and indicate that they were only reordered, I think I can safely say that we can rebaseline, and add this in `BASELINE_NOTES.md`:
> PR #2878 (2024-10-17)
=====================
Sorted region cellBlocks names alphabetically. Therefore affected ordering of: faceManager/elemSubRegionList, nodeManager/elemList, nodeManager/elemSubRegionList, SurfaceElementSubRegion::fractureElementsToCellSubRegions, field::perforation::reservoirElementSubregion.

I also changed the tag in `.integrated_tests.yaml`. I would say that this PR should be now ready to be merged but it is the first time I do a rebaseline, so don't hesitate to tell me if I did not that properly.


[I opened a PR](https://github.com/GEOS-DEV/geosPythonPackages/pull/43) to propose those changes to `restart_check.py`.

MelReyCG:
🥳

PR #2884, opened 2023-12-09
Fix issue #2882 by using output dir in wells rates output

### LINKED ISSUE #2882: Well output rates folder located in run directory not the output directory
**Describe the bug**
Running the integrated tests, a bunch of folders with well rates are left over when the ATS veryclean method is called.  I assume this is because they are not output to the usual results folder specified by the command line `-o` option.

paveltomin:
Fix for https://github.com/GEOS-DEV/GEOS/issues/2882


TotoGaz (approved):

TotoGaz:
Are you sure there's no implications on the integrated tests? (in the end, it's related to ATS).

paveltomin:
> Are you sure there's no implications on the integrated tests? (in the end, it's related to ATS).

Sure that there isn't that csv output is not used in ATS to check the results

PR #2894, opened 2023-12-14
Optimization of higher-order Gauss-Lobatto kernels

sframba:
This PR optimizes the GL finite element formulation on hexahedral meshes, by: 
- using symmetries and properties of the GL basis functions to simplify considerably the calculation of the jacobian matrix;
- pre-computing the values of the derivatives of basis functions at quadrature points.

With the improved performance obtained in this way, the `atomic_add` operations can become a bottleneck. Thus, this PR also modifies the wave propagation solvers to only do the atomic add once per element instead of once per quadrature point.
Here is a comparison of run times for a sample acoustic case with 1 million cells (`max time=0.5s`, `dt=1ms`), on 1 GPU, at orders $1-5$:

| order | run time in s (`develop`) | run time in s (this PR) |
|-------|-------------------------|----------------------|
| 1 | 1.302 | 1.086 |
| 2 | 14.482 | 7.224 |
| 3 | 126.184 | 30.318 |
| 4 | 812.539 | 170.739 |
| 5 | 3417.655 | 618.230 |


jiemeng-total (approved):
The modifications look good to me. The performance improvement is impressive!

sframba:
Tested acoustic and elastic at 1nd and 2nd order, for Q1 and Q2, results are ok.
@Bertbk do you have an acoustic VTI reference test case?

src/coreComponents/finiteElement/elementFormulations/Qk_Hexahedron_Lagrange_GaussLobatto.hpp
```diff
@@ -162,107 +277,127 @@ class Qk_Hexahedron_Lagrange_GaussLobatto final : public FiniteElementBase
...
+/**
+ * @brief Calculate the shape functions derivatives wrt the physical
+ *   coordinates.
+ * @param q Index of the quadrature point.
+ * @param X Array containing the coordinates of the mesh support points.
+ * @param stack Variables allocated on the stack as filled by @ref setupStack.
+ * @param gradN Array to contain the shape function derivatives for all
+ *   support points at the coordinates of the quadrature point @p q.
+ * @return The determinant of the parent/physical transformation matrix.
+ */
```
TotoGaz: Be careful w.r.t. the indentation?

src/coreComponents/physicsSolvers/wavePropagation/WaveSolverUtils.hpp
```diff
@@ -308,10 +308,10 @@ struct WaveSolverUtils
                                         arrayView2d< wsCoordType const, nodes::REFERENCE_POSITION_USD > const nodeCoords,
                                         real64 (& coordsOnRefElem)[3] )
   {
-    real64 xLocal[FE_TYPE::numNodes][3]{};
-    for( localIndex a = 0; a < FE_TYPE::numNodes; ++a )
+    real64 xLocal[8][3]{};
```
TotoGaz, line 312: What was the problem with `FE_TYPE::numNodes`?

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEMKernel.hpp
```diff
@@ -315,12 +315,13 @@ struct DampingMatrixKernel
         if( facesDomainBoundaryIndicator[f] == 1 && freeSurfaceFaceIndicator[f] != 1 )
         {
           constexpr localIndex numNodesPerFace = FE_TYPE::numNodesPerFace;
-          real64 xLocal[ numNodesPerFace ][ 3 ];
-          for( localIndex a = 0; a < numNodesPerFace; ++a )
+          real64 xLocal[ 4 ][ 3 ];
```
TotoGaz, line 318: What don't you change `numNodesPerFace` line 329 then?

TotoGaz (commented):
LGTM.
Some questions though on why you hard code values instead of using the variables.
@castelletto1 you're surely better positioned to review this PR.

src/coreComponents/physicsSolvers/wavePropagation/ElasticWaveEquationSEMKernel.hpp
sframba: It is needed later when we compute the actual damping term (this needs to be computed for all d.o.f., including higher-order nodes). To make it more readable, I have moved the definition of `numNodesPerFace` next to where it is actually used. I did the same for other similar definitions

src/coreComponents/physicsSolvers/wavePropagation/WaveSolverUtils.hpp
sframba: This PR changes the way the jacobian matrix is computed for the GL hexahedra. It used to be computed based on all the nodes in the mesh element (including higher order nodes), which lead to a higher algorithmic complexity for higher orders. With this PR, only the actual mesh nodes are needed (which are always 8, independent of order). Thus, we don't need to use `FE_TYPE::numNodes` (which is equal to $(k+1)^3$ at order $k$) anymore, since we are only using the 8 mesh vertices no matter $k$.
Note that since these are all SEM kernels, they can only work on hexahedra, so the number of (mesh) nodes per face or element is fixed and cannot be changed (these kernels cannot be used on tetrahedra, or any other mesh element type).

src/coreComponents/finiteElement/elementFormulations/Qk_Hexahedron_Lagrange_GaussLobatto.hpp
sframba: done -- indentation is more consistent now

acitrain (approved):

sframba:
@TotoGaz @castelletto1 do you think this PR is good to go?

src/coreComponents/physicsSolvers/wavePropagation/WaveSolverUtils.hpp
TotoGaz: OK; can you put some part of this comment into the code as a comment?

TotoGaz (approved):

sframba:
Rebaseline PR is here: https://github.com/GEOS-DEV/integratedTests/pull/87

PR #2896, opened 2023-12-14
Migrate python packages

### LINKED ISSUE #2851: Python package for pre/post-processing tools
**Describe the issue**
I would like to pre-and post-process data for geos, maybe add new algorithms... But, I do not want to clone the whole geos repository to have access to these modules.
I think that some python modules in `src/coreComponents/python` (e.g. mesh_doctor) could benefit from an extractiion of this repository into a specific one dedicated to a python package.
I could also lead to less burden for the geos repository

**Proposed cleanup**
Extract the modules in a new repository dedicated to a python package that will ultimately be distributed on pypi.
This package could include the algorithms, cli, paraview plugins, etc..

Package naming suggestions:
- geostk (stands for GEOS Tool Kit)
- _(open to suggestions, feel free to add yours)_

**Additional context**
n.a.

cssherman:
This PR removes old copies of the python packages that have been moved to their own repository: https://github.com/GEOS-DEV/geosPythonPackages .  The `setupPythonEnvironment.bash` script will now fetch the latest version of these packages from the internet.

integratedTests PR: https://github.com/GEOS-DEV/integratedTests/pull/77


cssherman:
@rrsettgast - for this change to work smoothly, we'll also want to move the python tool documentation to the new repo.  I've done so here: https://github.com/GEOS-DEV/geosPythonPackages/pull/1 .  We should be able to treat this repo as a subproject in readthedocs (see https://docs.readthedocs.io/en/stable/subprojects.html).  Any thoughts/concerns?

cssherman:
Here are the new separate docs for the python tools: https://geosx-geospythonpackages.readthedocs-hosted.com/en/latest/
As a final step, I'll need to go through, remove old files, and update links to the new docs.

src/CMakeLists.txt
```diff
@@ -229,6 +229,59 @@ install( FILES ${CMAKE_BINARY_DIR}/schema.xsd
...
+
+    # Check for the virtualenv package
+    execute_process(
+        COMMAND ${Python3_EXECUTABLE} -c "import virtualenv"
+        RESULT_VARIABLE VIRTUALENV_AVAILABLE
+    )
+
+    if (NOT ${VIRTUALENV_AVAILABLE} EQUAL 0)
+        message(WARNING "The \"virtualenv\" package was not found in the target python environment (${Python3_EXECUTABLE}).  This package may be required to build PYGEOSX or the python development environment.")
+    endif()
```
TotoGaz, line 257: Hello, is this still the case?

src/CMakeLists.txt
```diff
@@ -229,6 +229,59 @@ install( FILES ${CMAKE_BINARY_DIR}/schema.xsd
...
+                     )
+
+    add_custom_target( geosx_format_all_xml_files
+                       COMMAND bash ${CMAKE_SOURCE_DIR}/../scripts/formatXMLFiles.bash -g ${CMAKE_BINARY_DIR}/bin/format_xml ${CMAKE_SOURCE_DIR} ${CMAKE_SOURCE_DIR}/../examples
+                       WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
+                       DEPENDS geosx_xml_tools
+                     )
+
+else()
+    message(WARNING "Building the GEOSX python tools requires Python >= 3.7.")
```
TotoGaz: Are we checking against `Python_VERSION` and `Python_VERSION_MAJOR`?

scripts/setupPythonEnvironment.bash
```diff
@@ -3,8 +3,9 @@
 
 # Configuration
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
-PACKAGE_DIR=$SCRIPT_DIR/../src/coreComponents/python/modules
+PACKAGE_DIR=$(mktemp -d)
```
TotoGaz: Is this really is a `PACKAGE_DIR` or more `TMP_CLONE_DIR`?
Also, does it make sense to optionally provide a `--tmpdir` as CLI argument that could be the `CMAKE_BUILD_DIR`? I think that it could be nice no to leave the `build dir`.

scripts/setupPythonEnvironment.bash
```diff
@@ -69,6 +70,11 @@ shift # past argument or value
 done
 
 
+# Clone the package directory
+echo "Cloning the GEOS python package repository..."
+git clone --depth 1 --branch main https://github.com/GEOS-DEV/geosPythonPackages.git $PACKAGE_DIR
```
TotoGaz: If possible, I would put the clone very next to the definition of `PACKAGE_DIR`? That way, we can connect to the `TARGET_PACKAGES` as well?
Also, so you want to add `--single-branch`?

untereiner:
@cssherman Could this new repo have its own CI for the docs instead of fetching it in this repository ?

src/CMakeLists.txt
cssherman: I don't think that we are doing this explicitly anywhere in the cmake.  It's definitely worth doing though

src/CMakeLists.txt
cssherman: It is for the way that we build pygeosx, but for the rest of the packages we can sometimes get by without it

scripts/setupPythonEnvironment.bash
cssherman: Sure, I was just avoiding the re-name.  I agree that a CLI arg would be a good idea here

cssherman:
> @cssherman Could this new repo have its own CI for the docs instead of fetching it in this repository ?

Yes, that's what I'm working towards.  The RTD platform allows sub-projects, with different release schedules per repo

scripts/setupPythonEnvironment.bash
```diff
@@ -84,13 +76,47 @@ then
...
 
+# Check for a predefined package directory
+echo "Checking for python packages..."
+if [[ -z "${PACKAGE_DIR}" ]]
+then
+    echo "Cloning the GEOS python package repository..."
+    TMP_CLONE_DIR=$(mktemp -d)
+    PACKAGE_DIR=$TMP_CLONE_DIR/geosPythonPackages
+    git clone --depth 1 --branch main --single-branch https://github.com/GEOS-DEV/geosPythonPackages.git $PACKAGE_DIR
+elif [ ! -d "${PACKAGE_DIR}/geosx_xml_tools_package" ]
```
cssherman, line 102: This is just a quick check to see if the directory is close to correct.  There are later error messages that will be printed if other packages are broken/missing.

scripts/setupPythonEnvironment.bash
cssherman: Sounds good.  I've restructured the setup script to match this behavior

src/CMakeLists.txt
cssherman: I've updated the cmake to explicitly check the python version (if it is found) and throw an error if it is below 3.7.0

scripts/setupPythonEnvironment.bash
```diff
@@ -84,18 +91,36 @@ then
...
 for p in "${TARGET_PACKAGES[@]}"
 do
     if [ -d "$p" ]
     then
         echo "  $p"
+
+        # Try installing the package
         if $VERBOSE
-            INSTALL_MSG=$($PYTHON_TARGET -m $PIP_CMD install $p)
+            INSTALL_MSG=$($PYTHON_TARGET -m $PIP_CMD install $PACKAGE_DIR/$p)
```
TotoGaz, line 120: Is it possible to install directly from the `url` (like `python -m pip install https://github.com/GEOS-DEV/geosPythonPackages.git/...`)?
That would prevent us from removing by hand the temporary which is always a scary part.

src/CMakeLists.txt
```diff
@@ -229,6 +229,65 @@ install( FILES ${CMAKE_BINARY_DIR}/schema.xsd
...
+                     )
+
+    add_custom_target( geosx_format_all_xml_files
+                       COMMAND bash ${CMAKE_SOURCE_DIR}/../scripts/formatXMLFiles.bash -g ${CMAKE_BINARY_DIR}/bin/format_xml ${CMAKE_SOURCE_DIR} ${CMAKE_SOURCE_DIR}/../examples
+                       WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
+                       DEPENDS geosx_xml_tools
+                     )
+
+else()
+    message(WARNING "Building the GEOSX python tools requires Python >= 3.6.")
```
TotoGaz, line 286:
```suggestion
    message(WARNING "Building the GEOSX python tools requires Python >= 3.6. No version was provided.")
```

TotoGaz (approved):

CusiniM (approved):

CusiniM:
why does this require baselines ?

src/CMakeLists.txt
```diff
@@ -229,6 +229,65 @@ install( FILES ${CMAKE_BINARY_DIR}/schema.xsd
...
+    set( GEOSX_PYTHON_TOOLS_BINS
+         "${CMAKE_BINARY_DIR}/bin/preprocess_xml"
+         "${CMAKE_BINARY_DIR}/bin/format_xml" )
+
+    add_custom_command( OUTPUT ${GEOSX_PYTHON_TOOLS_BINS}
+                        COMMAND bash ${CMAKE_SOURCE_DIR}/../scripts/setupPythonEnvironment.bash -p ${PYTHON_POST_EXECUTABLE} -b ${CMAKE_BINARY_DIR}/bin
+                        WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
+                      )
+
+    add_custom_target( geosx_python_tools
```
CusiniM, line 269: shouldn't we call these `geos_python_tools` at some point?

src/CMakeLists.txt
CusiniM:
```suggestion
    message(WARNING "Building the GEOS python tools requires Python >= 3.6. No version was provided.")
```

PR #2899, opened 2023-12-18
Only pack err and data files on CI/CD

TotoGaz:


TotoGaz:
@cssherman Do you confirm that this is outdated?

PR #2907, opened 2023-12-21
Add HYPRE warning

victorapm:
Add warning message to the simulation log for the case when only GEOS is configured with GPU support, but not hypre.

I'm not sure if I put the warning in the right place, so feel free to modify. Thanks!


wrtobin (approved):

rrsettgast:
Do we want a warning or error or Compiler time error?

sframba:
> Do we want a warning or error or Compiler time error?

For what it's worth, compiling with cuda support for GEOS but without cuda for hypre is definitely possible and everything seems to work well. We have been doing it for a while to avoid issues with some missing hypre features on cuda (although I deduce from this PR that everything is fine now?)

wrtobin:
Allowing the GEOS+GPU - HYPRE+CPU build combo is very useful in debugging failures on new platforms, I definitely don't want to drop support for it.

jhuang2601 (approved):

paveltomin:
move to merge queue?

PR #2909, opened 2023-12-22
feat: Add option for automatic timestep computation in wave solvers

acitrain:
This PR add the possibility of compute automatically the timestep for wave propagation solvers.

How to use it?
- You add the correct switch in your `XML` (`timestepStabilityLimit`=1) 
- You launch GEOS
- If the requested dt for the solver (`forceDt`) is higher than the CFL condition limit, substeps are used to remain below the condition
Also, an option is added to input source wavelets using `TableFunction`s, with the option `sourceWaveletTableNames`
There also a routine inside the PR to be wrapped using Pygeosx


src/coreComponents/physicsSolvers/wavePropagation/AcousticFirstOrderWaveEquationSEM.hpp
```diff
@@ -151,6 +151,8 @@ class AcousticFirstOrderWaveEquationSEM : public WaveSolverBase
    */
   virtual void applyPML( real64 const time, DomainPartition & domain ) override;
 
+  virtual real64 computeTimeStep( real64 & dtOut ) override;
```
sframba: Add doc for the method

src/coreComponents/physicsSolvers/wavePropagation/WaveSolverUtils.hpp
```diff
@@ -327,6 +327,34 @@ struct WaveSolverUtils
...
+                          real64 & res )
+  {
+
+    RAJA::ReduceSum< parallelDeviceReduce, real64 > tmp( 0.0 );
+    forAll< EXEC_POLICY >( size, [=] GEOS_HOST_DEVICE ( localIndex const a )
+    {
+      tmp+= vector1[a]*vector2[a];
+    } );
+
+    res = tmp.get();
```
sframba: I think there should be an extra MPI::sum to compute the global value over all MPI ranks. Or is it already done by the ReduceSum? Have you tried printing the value computed by each MPI rank?

src/coreComponents/physicsSolvers/wavePropagation/AcousticFirstOrderWaveEquationSEM.cpp
```diff
@@ -337,13 +339,41 @@ void AcousticFirstOrderWaveEquationSEM::initializePostInitialConditionsPreSubGro
...
+                                                                             nodeManager.size(),
+                                                                             X,
+                                                                             elemsToNodes,
+                                                                             density,
+                                                                             mass );
+
+
+          real64 globaldt = MpiWrapper::min( dtCompute );
+
+          printf( "dt=%f\n", globaldt );
```
sframba: use `GEOS_LOG_RANK_0` instead of `printf`

src/coreComponents/physicsSolvers/wavePropagation/AcousticFirstOrderWaveEquationSEM.cpp
```diff
@@ -337,13 +339,41 @@ void AcousticFirstOrderWaveEquationSEM::initializePostInitialConditionsPreSubGro
...
+
+          exit( 2 );
```
sframba: We should exit more gracefully, abruptly calling `exit` here might lead to unfinalized I/O and other undefined behavior. For example, could you use the `exitFlag` of the `EventManager`?

src/coreComponents/physicsSolvers/wavePropagation/AcousticFirstOrderWaveEquationSEMKernel.hpp
```diff
@@ -321,6 +321,355 @@ struct DampingMatrixKernel
...
+    lambdaNew = dotProductPPaux/normP;
+
+    real64 normPaux = 0.0;
+    WaveSolverUtils::dotProduct( sizeNode, pAuxView, pAuxView, normPaux );
+    forAll< EXEC_POLICY >( sizeNode, [=] GEOS_HOST_DEVICE ( localIndex const a )
+    {
+      pView[a] = pAuxView[a]/( normPaux );
+    } );
+
+    //Step 3: Do previous algorithm until we found the max eigenvalues
```
sframba: Can't you merge the previous step in this `do` loop and avoid this (sizeable) copy/paste?

src/coreComponents/physicsSolvers/wavePropagation/AcousticFirstOrderWaveEquationSEMKernel.hpp
```diff
@@ -321,6 +321,355 @@ struct DampingMatrixKernel
...
+
+      numberIter++;
+
+
+    }
+    while (counter < 10 && numberIter < nIterMax);
+
+    GEOS_THROW_IF( numberIter> nIterMax, "Power Iteration algorithm does not converge", std::runtime_error );
+
+    real64 dt = 1.99/sqrt( abs( lambdaNew ));
```
sframba: 1.99 because you are using a 0.5% margin with respect to the CFL condition? Please add a small comment to explain

src/coreComponents/physicsSolvers/wavePropagation/AcousticFirstOrderWaveEquationSEMKernel.hpp
```diff
@@ -321,6 +321,355 @@ struct DampingMatrixKernel
...
+    for( localIndex a = 0; a < sizeNode; ++a )
+    {
+      pView[a] = (real64)rand()/(real64) RAND_MAX;
+    }
+
+    //Step 1: Normalize randomized pressure
+    real64 normP= 0.0;
+    WaveSolverUtils::dotProduct( sizeNode, pView, pView, normP );
+
+    forAll< EXEC_POLICY >( sizeNode, [=] GEOS_HOST_DEVICE ( localIndex const a )
```
sframba: Can't you use the stiffness kernel directly to compute K.x instead of adding all this code? It seems like tere should be a lot of overlap with the usual computation of a forward step that the kernel already does...

src/coreComponents/physicsSolvers/wavePropagation/AcousticWaveEquationSEM.cpp
```diff
@@ -317,13 +320,87 @@ void AcousticWaveEquationSEM::initializePostInitialConditionsPreSubGroups()
...
+                                                                             nodeManager.size(),
+                                                                             nodeCoords,
+                                                                             elemsToNodes,
+                                                                             mass );
+
+          real64 globaldt = MpiWrapper::min( dtCompute );
+
+          printf( "dt=%f\n", globaldt );
+
+          exit( 2 );
```
sframba: same comments as above

src/coreComponents/physicsSolvers/wavePropagation/AcousticFirstOrderWaveEquationSEM.cpp
```diff
@@ -337,13 +339,41 @@ void AcousticFirstOrderWaveEquationSEM::initializePostInitialConditionsPreSubGro
...
+
+          dtCompute = kernelT.template launch< EXEC_POLICY, ATOMIC_POLICY >( elementSubRegion.size(),
+                                                                             nodeManager.size(),
+                                                                             X,
+                                                                             elemsToNodes,
+                                                                             density,
+                                                                             mass );
+
+
+          real64 globaldt = MpiWrapper::min( dtCompute );
```
sframba: Why are you taking the Min over all MPI ranks ? The deltaT (and the CFL condition) should be global, obtained by the Rayleigh quotient using the global fields, there shouldn't be a "local" timestep per rank, no?

sframba (commented):
Thanks Aurélien! 
I have two main comments:
- The CFL factor should be global, not local to an MPI rank. Thus
  - The scalar products in `WaveSolverUtils` should include an MPI::sum to get a global value;
  - There should be no MPI::min call on the delta T, you should have a single global value
- There is a very large amount of code that essentially computes M^(-1).K.x. If I understand correctly, this should be extremely similar to what we do in a single forward time step. We should be able to refactor this and reuse the already available code, there is no need to add these many lines (and computations)

tbeltzun:
Would there be a way to specify a zero timestep (standing for automatic computation for the timestep) in the xml file, and reuse that time step automatically in the computation, without having to modify xmls and relaunch the code ?

sframba (approved):

CusiniM (approved):

src/coreComponents/physicsSolvers/wavePropagation/shared/WaveSolverBase.cpp
```diff
@@ -191,6 +185,18 @@ WaveSolverBase::WaveSolverBase( const std::string & name,
     setSizedFromParent( 0 ).
     setDescription( "Flag that indicates whether the receiver is local to this MPI rank" );
 
+  registerWrapper( viewKeyStruct::timestepStabilityLimitString(), &m_timestepStabilityLimit ).
```
rrsettgast, line 188: Shouldn't this always be used unless `forceDt` is set?

rrsettgast (approved):

src/coreComponents/physicsSolvers/wavePropagation/shared/WaveSolverBase.cpp
acitrain: In fact in this PR we are closer to the logic of the non-linear solvers: if you put a forceDt too big for the CFL condition the internal loop of the solver will do sub-steps to respect the CFL condition. 
We decided to put this feature as an option to let the user be able to tests some dt on is own without getting into that routine.

CusiniM:
I have approved it but two things about this PR:
1. please don't put PRs  in the merge queue until they have all codeowners approvals.
2. IMO, the idea of `forceDt` is for the timestep size to be fixed. To let the solver adjust the timestep size one can just set a maxTiimestep size for the event instead of forcing it. So, I would argue that, if the `forceDt` is too large it's fine for the simulation to fail instead of substepping. I know that solver base sort of does this too but we have talked about removing it coz it is actually a lot easier to let the `EventManager` be the only object managing timestepping.

Issue #2823, opened 2023-11-15
PVT module broken on Pangea 3 - no reservoir simulation possible


sframba:
**Describe the bug**
Starting from [PR 2552](https://github.com/GEOS-DEV/GEOS/pull/2552), reservoir simulation is broken on Pangea3.
The `testPVT` unit test fails, and reservoir simulations fail at the beginning of the time loop.

**To Reproduce**
1. compile GEOS on Pangea 3
2. the `testPVT` unit test fails

TotoGaz:
Hello @sframba this is really surprising. Can you share some logs?

sframba:
@TotoGaz The original issue is probably fixed by #2812, by adding `GEOS_HOST_DEVICE` on conversion methods in `Units.hpp`. 
Unfortunately, another very similar issue appears after #2772 . But It seems to be tied to a direct access to private members of the class `PVTDriver` from device code. Hopefully it will be easy to fix

PR #2910, opened 2023-12-22
Fix compilation issues and PVT Driver unit test on Pangea3

sframba:
Currently, GEOS is broken on Pangea3 due to a few issues with compilation and with the `PVTDriver` module's tests.
- This PR closes issue #2823, by removing direct access to some private class members that are not allowed from device code.
- Furthermore, this PR updates the Pangea3 host config file following the upgrade of `cuda` to version `11.5` on Pangea3. In particular, 
  - ~~Hypre is by default compiled without CUDA acceleration (not completely supported);~~ hypre with cuda acceleration seems to be mostly working now
  - The LvArray benchmarks are disabled since they do not compile with this version of `cuda`(solution suggested by @corbett5), and 
  - A flag is addedd to allow host constexpr function call from device code.

Overall, this PR restores compilation and functionality on Pangea3. 
There is a companion thirdPartyLibs PR, [#253](https://github.com/GEOS-DEV/thirdPartyLibs/pull/253), which updates and makes the host config files consistent.


dkachuma (approved):
Looks good to me. Thanks.

TotoGaz (approved):

src/coreComponents/constitutive/fluid/multifluid/PVTDriverRunTest.hpp
```diff
@@ -64,14 +64,16 @@ void PVTDriver::runTest( FLUID_TYPE & fluid, arrayView2d< real64 > const & table
   // note: column indexing should be kept consistent with output file header below.
 
   integer const numSteps = m_numSteps;
+  integer const outputCompressibility = m_outputCompressibility;
+  integer const outputPhaseComposition = m_outputPhaseComposition;
   using ExecPolicy = typename FLUID_TYPE::exec_policy;
   forAll< ExecPolicy >( composition.size( 0 ),
-                        [this, numPhases, numComponents, numSteps, kernelWrapper,
```
CusiniM, line 69: ah yeah, capturing this by value won't work if run on device.

PR #2916, opened 2023-12-25
Fix issue #2871

### LINKED ISSUE #2871: Brooks Corey Capillary Pressure: Pore Size Distribution Index Checked for < 1, should be < 0
Description: GEOS checks that the pore size distribution index value for the Brooks Corey capillary pressure model is not less than 1; however, values less than 1 are common and should be allowed. The code should likely check that this value is not less than 0 instead, which wouldn't be an acceptable parameter value. 

To Reproduce: Run a GEOS simulation using a Brooks Corey capillary pressure model with a `phaseCapPressureExponentInv` value for either the gas or water phase that is less than 1. This produces the error:

```
***** LOCATION: /qfs/people/haag359/Codes/GEOS/src/coreComponents/constitutive/capillaryPressure/BrooksCoreyCapillaryPressure.cpp:91
***** Controlling expression (should be false): m_phaseCapPressureExponentInv[ip] < 1.0
***** Rank 0: BrooksCoreyCapillaryPressure capPressure: invalid value at phaseCapPressureExponentInv[0]
Expected m_phaseCapPressureExponentInv[ip] >= 1.0
  m_phaseCapPressureExponentInv[ip] = 0.88
  1.0 = 1
``` 

Platform:
- Machine: PNNL Deception
- Compiler: gcc/g++ 10.3.0
- GEOSX Version -- Branch: feature/white/lbcViscosity, Commit: d4ec4501f8ced76ada6c1df4fb851701786582db

Suggested fix:
Edit lines 91-93 of BrooksCoreyCapillaryPressure.cpp to read as
```
      GEOS_THROW_IF_LT_MSG( m_phaseCapPressureExponentInv[ip], 0.0,
                            errorMsg( viewKeyStruct::phaseCapPressureExponentInvString() ),
                            InputError );
```

paveltomin:
Fix https://github.com/GEOS-DEV/GEOS/issues/2871

### DIFF
--- a/src/coreComponents/constitutive/capillaryPressure/BrooksCoreyCapillaryPressure.cpp
+++ b/src/coreComponents/constitutive/capillaryPressure/BrooksCoreyCapillaryPressure.cpp
@@ -88,7 +88,7 @@ void BrooksCoreyCapillaryPressure::postProcessInput()
 
     if( m_phaseTypes[ip] != CapillaryPressureBase::REFERENCE_PHASE )
     {
-      GEOS_THROW_IF_LT_MSG( m_phaseCapPressureExponentInv[ip], 1.0,
+      GEOS_THROW_IF_LE_MSG( m_phaseCapPressureExponentInv[ip], 0.0,
                             errorMsg( viewKeyStruct::phaseCapPressureExponentInvString() ),
                             InputError );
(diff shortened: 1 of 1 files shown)


jafranc (approved):

jhuang2601 (approved):
Thanks for this fix.

PR #2918, opened 2023-12-30
refactor: diffusion and tpfa weights calculation to enable phase indexed arrays (4d)

jafranc:
Like what was discuss [here](https://github.com/GEOS-DEV/GEOS/pull/2678#discussion_r1329317111) about allowing phase-indexed tensor in the tpfa transmissivities/weights calculation to be able to keep consistency with usual transmissivity computed on intrinsinc permeability.

This is also used in #2807 to enable phase indexed dispersivity.

For now it refactor main steps of weights computation into `computeWeightBase()` and `averageWeights()` refactor weights blending between harmonic and arithmetic average. This makes it easy to refactor further if physical phenomenon ( convection vs diffusion or dipersion) requires other blending rules.

Eventually, as commented in the code, the main stride manipulation that occurs in the new overload of `computeWeights()` taking 4d-arrays, should be refactored once @klevzoff 's LvArray refactorization PR is merged in.


src/coreComponents/finiteVolume/SurfaceElementStencil.hpp
```diff
@@ -116,12 +116,30 @@ class SurfaceElementStencilWrapper : public StencilWrapperBase< SurfaceElementSt
...
    * @param[out] dWeight_dVar derivative of the weights w.r.t to the variable
    */
   GEOS_HOST_DEVICE
+  void computeWeights( localIndex const iconn,
+                       localIndex const ip,
+                       CoefficientAccessor< arrayView4d< real64 const > > const & coefficient,
+                       CoefficientAccessor< arrayView4d< real64 const > > const & dCoeff_dVar,
+                       real64 ( & weight )[maxNumConnections][2],
+                       real64 ( & dWeight_dVar )[maxNumConnections][2] ) const
+  { GEOS_UNUSED_VAR( iconn, ip, coefficient, dCoeff_dVar, weight, dWeight_dVar ); };
```
CusiniM: If necessary for diffusion this function should also be implemented for the other stencil types. All the assembly kernels should work with any stencil type so that all physics can be used in the faults/fractures. I don't think any special treatment is needed.

jafranc: @CusiniM I did a first pass at refactoring stencils to fit the proposed formalism. I did it quite blindly and second pass should come up with some more efficient refactoring. (There is quite some overloads in `SurfaceElementStencil`, `EmbeddedSurfaceToCellStencil` and `FaceElementToCellStencil`). Please review and let me know your feedback. Two point prior though:

1. I did not include 
```
GEOS_HOST_DEVICE
inline void
SurfaceElementStencilWrapper::
  computeWeights( localIndex iconn,
                  CoefficientAccessor< arrayView3d< real64 const > > const & coefficient1,
                  CoefficientAccessor< arrayView3d< real64 const > > const & coefficient1Multiplier,
                  CoefficientAccessor< arrayView1d< real64 const > > const & coefficient2,
                  R1Tensor const & unitGravityVector,
                  real64 ( & weight1 )[maxNumPointsInFlux],
                  real64 ( & weight2 )[maxNumPointsInFlux],
                  real64 ( & geometricWeight )[maxNumPointsInFlux] ) const
```
or

```
GEOS_HOST_DEVICE
inline void
SurfaceElementStencilWrapper::
  computeWeights( localIndex iconn,
                  CoefficientAccessor< arrayView3d< real64 const > > const & coefficient,
                  CoefficientAccessor< arrayView3d< real64 const > > const & coefficientMultiplier,
                  R1Tensor const & gravityVector,
                  real64 (& weight)[maxNumConnections][2] ) const
```
as I did not find usage for second and only find usage of first in propant solver. These seems special cases but could be refactor _if needed_.

2. I might have overdone in adding an overload for 
```
GEOS_HOST_DEVICE
inline void
FaceElementToCellStencilWrapper::computeWeights( const localIndex iconn,
                                                 const localIndex ip,
                                                 const CoefficientAccessor< arrayView4d< const real64 > > & coefficient,
                                                 const CoefficientAccessor< arrayView4d< const real64 > > & dCoeff_dVar1,
                                                 const CoefficientAccessor< arrayView4d< const real64 > > & dCoeff_dVar2,
                                                 real64 (& weight)[1][2],
                                                 real64 (& dWeight_dVar1)[1][2],
                                                 real64 (& dWeight_dVar2)[1][2] ) const
```
This might never be used with phase-indexed coefficient, so let me know and I'll revert it

src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseFVMKernels.hpp
```diff
@@ -1125,14 +1118,22 @@ class DiffusionDispersionFaceBasedAssemblyKernel : public FaceBasedAssemblyKerne
...
-                                                     stack.transmissibility[connectionIndex][1] };
-
         //***** calculation of flux *****
         // loop over phases, compute and upwind phase flux and sum contributions to each component's flux
         for( integer ip = 0; ip < m_numPhases; ++ip )
         {
-
+          // first, compute the transmissibilities at this face
+          // note that the dispersion tensor is lagged in iteration
+          m_stencilWrapper.computeWeights( iconn,
```
CusiniM, line 1156: By moving this inside the connected elem loops are we looping twice over the connected elems? I think the `SurfaceElementStencil` will loop again because with the star-delta method it needs to know all half transmissibilities to compute the one between two connected cells.

src/coreComponents/finiteVolume/SurfaceElementStencil.hpp
```diff
@@ -327,59 +389,104 @@ SurfaceElementStencilWrapper::
...
+  }
+}
 
-      real64 dHarmonic[2];
-      dHarmonic[0] = ( dt0 * t1 * sumOfTrans - dt0 * t0 * t1 ) / ( sumOfTrans * sumOfTrans );
-      dHarmonic[1] = ( t0 * dt1 * sumOfTrans - dt1 * t0 * t1 ) / ( sumOfTrans * sumOfTrans );
+    GEOS_HOST_DEVICE
+    inline void
+    SurfaceElementStencilWrapper::
+    computeWeights( localIndex iconn,
```
CusiniM: isn't this method the same for all stencils? I feel that the only method that needs to be specialized is `computeWeightsBase`.

src/coreComponents/finiteVolume/SurfaceElementStencil.hpp
```diff
@@ -559,19 +652,99 @@ SurfaceElementStencilWrapper::
...
+
+  dWeight_dVar[connexionIndex][0] =   ( m_meanPermCoefficient * dHarmonic[0] + (1 - m_meanPermCoefficient) * dArithmetic[0] );
+  dWeight_dVar[connexionIndex][1] = -( m_meanPermCoefficient * dHarmonic[1] + (1 - m_meanPermCoefficient) * dArithmetic[1] );
+
+
+}
+
+GEOS_HOST_DEVICE
+inline void
+SurfaceElementStencilWrapper::averageWeights( const localIndex iconn,
```
CusiniM: I suspect this method is also the same for all stencils.

src/coreComponents/finiteVolume/SurfaceElementStencil.hpp
CusiniM: If I understand correctly what's happening, I think all stencils need to be able to deal with a coefficient that depends on the phase index. However, I think they all deal with it in the same way so most things can be implemented in the base class and what needs to be specialized is the method `computeWeightsBase`

src/coreComponents/finiteVolume/SurfaceElementStencil.hpp
jafranc: Not sure I understand here. Is that all stencil get a 
```
GEOS_HOST_DEVICE
inline void
EmbeddedSurfaceToCellStencilWrapper::
  computeWeights( localIndex iconn,
                  CoefficientAccessor< arrayView3d< real64 const > > const & coefficient,
                  CoefficientAccessor< arrayView3d< real64 const > > const & dCoeff_dVar,
                  real64 ( & weight )[1][2],
                  real64 ( & dWeight_dVar )[1][2] ) const;
``` 
and should then get the phase-indexed extension ? If so, it is unfortunately not `SurfaceElementStencil` 's case as it has `maxNumConnections` set to 6.

Otherwise there is quite some overloads of weight calcultation that take up to two variable deps.

src/coreComponents/finiteVolume/SurfaceElementStencil.hpp
jafranc: Agree on factoring up possible duplicate.

jafranc:
Re-booting this PR as useful to #3026 and #2807,
Making an intermediate abstraction `ToCell` on the way to `FaceToCellElement` and `EmbeddedElementToCell` as most of them can be factored.

jafranc:
For now I am unsure though how to factor up this

https://github.com/GEOS-DEV/GEOS/blob/6919bfb04393cae7ff7a336a5ae635dbd2118f59/src/coreComponents/finiteVolume/FaceElementToCellStencil.hpp#L244-L263

As unlinke halfWeight in TPFA, the treatment is non-symmetrical for `t0` and `t1` ?
Another question is there still a use of MPFA sources ?

CusiniM:
> For now I am unsure though how to factor up this
> 
> https://github.com/GEOS-DEV/GEOS/blob/6919bfb04393cae7ff7a336a5ae635dbd2118f59/src/coreComponents/finiteVolume/FaceElementToCellStencil.hpp#L244-L263
> 
> As unlinke halfWeight in TPFA, the treatment is non-symmetrical for `t0` and `t1` ? Another question is there still a use of MPFA sources ?

is that ever used? I am not sure what for... It would just return the harmonic average of the geometric weights? I really don't remember much. Maybe @karimifard ? 

No, I think the conclusion was that we would use mimetic to do mpfa. It also means we could simplify a bit our stencil classes if we accept some loss of generality.

karimifard:
> > For now I am unsure though how to factor up this
> > https://github.com/GEOS-DEV/GEOS/blob/6919bfb04393cae7ff7a336a5ae635dbd2118f59/src/coreComponents/finiteVolume/FaceElementToCellStencil.hpp#L244-L263
> > 
> > As unlinke halfWeight in TPFA, the treatment is non-symmetrical for `t0` and `t1` ? Another question is there still a use of MPFA sources ?
> 
> is that ever used? I am not sure what for... It would just return the harmonic average of the geometric weights? I really don't remember much. Maybe @karimifard ?
> 
> No, I think the conclusion was that we would use mimetic to do mpfa. It also means we could simplify a bit our stencil classes if we accept some loss of generality.

I also recall that the plan was to remove MPFA and use only mimetic.

jafranc:
> > > For now I am unsure though how to factor up this
> > > https://github.com/GEOS-DEV/GEOS/blob/6919bfb04393cae7ff7a336a5ae635dbd2118f59/src/coreComponents/finiteVolume/FaceElementToCellStencil.hpp#L244-L263
> > > 
> > > As unlinke halfWeight in TPFA, the treatment is non-symmetrical for `t0` and `t1` ? Another question is there still a use of MPFA sources ?
> > 
> > 
> > is that ever used? I am not sure what for... It would just return the harmonic average of the geometric weights? I really don't remember much. Maybe @karimifard ?
> > No, I think the conclusion was that we would use mimetic to do mpfa. It also means we could simplify a bit our stencil classes if we accept some loss of generality.
> 
> I also recall that the plan was to remove MPFA and use only mimetic.

Great :-)
I would then restart with that in mind and propose a refactoring that can unlock #2807 and #3026 :)

PR #2919, opened 2024-01-02
Make possible to provide input PVT tables instead of internally computed

### LINKED ISSUE #2810: Table input for CO2 solubility
Would be useful to have an option to provide CO2 solubility table directly instead of internal correlation

paveltomin:
Although a bit convoluted, it addresses https://github.com/GEOS-DEV/GEOS/issues/2810

integrateTests PR: https://github.com/GEOS-DEV/integratedTests/pull/91


src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/CO2Solubility.cpp
```diff
@@ -265,22 +270,27 @@ TableFunction const * makeVapourisationTable( string_array const & inputParams,
...
   string const tableName = functionName + "_waterVaporization_table";
+
   if( functionManager.hasGroup< TableFunction >( tableName ) )
   {
-    return functionManager.getGroupPointer< TableFunction >( tableName );
+    TableFunction * const vapourisationTable = functionManager.getGroupPointer< TableFunction >( tableName );
+    vapourisationTable->initializeFunction();
+    vapourisationTable->setDimUnits( { units::Pressure, units::TemperatureInC } );
+    vapourisationTable->setValueUnits( units::Solubility );
+    vapourisationTable->setInterpolationMethod( TableFunction::InterpolationType::Linear );
```
dkachuma: I would imagine we wouldn't want to override the user specified interpolation method. Same for all the other tables.

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/CO2Solubility.cpp
```diff
@@ -248,7 +248,12 @@ TableFunction const * makeSolubilityTable( string_array const & inputParams,
   string const tableName = functionName + "_co2Dissolution_table";
```
dkachuma, line 282: I think we should allow the user to give the table a name instead of these internally generated names. One use case would be where I want to use the same table across say 3 fluids.

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/CO2Solubility.cpp
paveltomin: just though being consistent with internal table interpolation is better, but you have a point, let me drop `setInterpolationMethod`

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/CO2Solubility.cpp
paveltomin: just not quite sure how to pull down the table name here without changing too much :)

paveltomin: @dkachuma do you have idea how it should look like?

dkachuma: @paveltomin I put a comment on the associated issue #2810. As for how to implement it, let me draft a change.

dkachuma (approved):

paveltomin:
tests are ok, rebase needed because of the new field `solubilityTableNames`

paveltomin:
now it has `Error: Formal parameter space overflowed (4224 bytes required, max 4096 bytes allowed)` in CUDA builds...

PR #2623, opened 2023-08-17
Modified parallel plates permeability and updated discretization

karimifard:
- Parallel plates permeability is now defined as ($a^2 /12$) instead of ($a^3 /12$) to be consistent with the other permeability models and the thermal discretization.
- All TPFA discretizations are updated to be consistent with this change, this includes the update of stencil weights during the simulation when the hydraulic aperture changes.
- In addition, the discretization for flow exchange between fracture and matrix is updated to account for fracture properties.
- Small bug fix in Carman Kozeny permeability formula.

IntegratedTests PR: https://github.com/GEOS-DEV/integratedTests/pull/37

PR #2922, opened 2024-01-04
Fixed issue #2853

### LINKED ISSUE #2853: Hydrofracture solver performance degradation after PR #2623
Hello,

I noticed that the `hydrofracture` solver has become worse in performance after the commit 94e338c. The convergence speed is relatively slow. I am not sure which PR after commit 94e338c caused this issue.

I have compiled the branch `develop latest` and  branch `develop 94e338c` based on ubuntu 22.04 and measured the time consumed by case `kgdViscosityDominated_poroelastic_benchmark.xml` as follows (`export OMP_NUM_THREADS=1 mpirun -np 9 geos -i kgdViscosityDominated_poroelastic_benchmark.xml -x 3 -y 3 -z 1` ):

| kgdViscosityDominated_poroelastic_benchmark.xml  |  Branch `develop latest` elasped time | Branch `develop 94e338c` elasped time |
| :-- | :--: | :--: |
| The first 1 second                                                        |  134 s                                                      | 97 s                                                        |
| The first 10 seconds                                                     |  1016 s                                                    | 620 s                                                     |

The results show that the `develop 94e338c` branch is faster than the `develop latest` by about 50%. 

At first I suspected that commit 6d397f7 was the problem because it changed `HydrofractureSolver.cpp` a lot, but I tested it and found the performance degradation was not related to commit 6d397f7. I suspect that there might be some modifications after commit 94e338c and before commit 6d397f7 that affect the performance of the simulation.

I suggest that performance testing is necessary as well when submitting a new PR.

liyangrock:
This PR fixed issue #2853 . 

The nonlinear system solver (`solveNonlinearSystem`) may invoke the `updateState` function, which can modify the stencil weights at each newton iteration. In this PR, the stencil weights are updated only once in `implicitStepComplete` function at the end of each solver step.

### DIFF
--- a/src/coreComponents/physicsSolvers/fluidFlow/FlowSolverBase.cpp
+++ b/src/coreComponents/physicsSolvers/fluidFlow/FlowSolverBase.cpp
@@ -736,7 +736,7 @@ void FlowSolverBase::prepareStencilWeights( DomainPartition & domain ) const
     FiniteVolumeManager const & fvManager = numericalMethodManager.getFiniteVolumeManager();
     FluxApproximationBase const & fluxApprox = fvManager.getFluxApproximation( getDiscretizationName() );
     ElementRegionManager::ElementViewAccessor< arrayView1d< real64 const > > hydraulicAperture =
-      mesh.getElemManager().constructViewAccessor< array1d< real64 >, arrayView1d< real64 const > >( fields::flow::hydraulicAperture::key() );
+      mesh.getElemManager().constructViewAccessor< array1d< real64 >, arrayView1d< real64 const > >( fields::flow::aperture0::key() );
 
     fluxApprox.forStencils< SurfaceElementStencil, FaceElementToCellStencil, EmbeddedSurfaceToCellStencil >( mesh, [&]( auto & stencil )
     {
--- a/src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp
+++ b/src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp
@@ -856,13 +856,7 @@ void HydrofractureSolver< POROMECHANICS_SOLVER >::updateState( DomainPartition &
 
   Base::updateState( domain );
 
-  // remove the contribution of the hydraulic aperture from the stencil weights
-  flowSolver()->prepareStencilWeights( domain );
-
   updateHydraulicApertureAndFracturePermeability( domain );
-
-  // update the stencil weights using the updated hydraulic aperture
-  flowSolver()->updateStencilWeights( domain );
 
   forDiscretizationOnMeshTargets( domain.getMeshBodies(), [&] ( string const &,
                                                                 MeshLevel & mesh,
@@ -878,6 +872,20 @@ void HydrofractureSolver< POROMECHANICS_SOLVER >::updateState( DomainPartition &
       flowSolver()->updateFluidState( subRegion );
     } );
   } );
+}
+
+template< typename POROMECHANICS_SOLVER >
+void HydrofractureSolver< POROMECHANICS_SOLVER >::implicitStepComplete( real64 const & time_n,
+                                                                        real64 const & dt,
+                                                                        DomainPartition & domain )
+{
+  Base::implicitStepComplete( time_n, dt, domain );
+
+  // remove the contribution of the hydraulic aperture from the stencil weights
+  flowSolver()->prepareStencilWeights( domain );
+
+  // update the stencil weights using the updated hydraulic aperture
+  flowSolver()->updateStencilWeights( domain );
 }
 
 template< typename POROMECHANICS_SOLVER >
--- a/src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.hpp
+++ b/src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.hpp
@@ -112,6 +112,11 @@ public:
 
   virtual void updateState( DomainPartition & domain ) override final;
 
+  virtual void implicitStepComplete( real64 const & time_n,
+                                     real64 const & dt,
+                                     DomainPartition & domain ) override final;
+
+
   /**@}*/
 
   void updateHydraulicApertureAndFracturePermeability( DomainPartition & domain );
--- a/src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoromechanicsConformingFractures.cpp
+++ b/src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoromechanicsConformingFractures.cpp
@@ -736,13 +736,7 @@ void SinglePhasePoromechanicsConformingFractures::updateState( DomainPartition &
 
   Base::updateState( domain );
 
-  // remove the contribution of the hydraulic aperture from the stencil weights
-  poromechanicsSolver()->flowSolver()->prepareStencilWeights( domain );
-
   updateHydraulicApertureAndFracturePermeability( domain );
-
-  // update the stencil weights using the updated hydraulic aperture
-  poromechanicsSolver()->flowSolver()->updateStencilWeights( domain );
 
   forDiscretizationOnMeshTargets( domain.getMeshBodies(), [&] ( string const &,
                                                                 MeshLevel & mesh,
@@ -763,6 +757,20 @@ void SinglePhasePoromechanicsConformingFractures::updateState( DomainPartition &
       }
     } );
   } );
+}
+
+void SinglePhasePoromechanicsConformingFractures::implicitStepComplete( real64 const & time_n,
+                                                                        real64 const & dt,
+                                                                        DomainPartition & domain )
+{
+  Base::implicitStepComplete( time_n, dt, domain );
+
+  // remove the contribution of the hydraulic aperture from the stencil weights
+  poromechanicsSolver()->flowSolver()->prepareStencilWeights( domain );
+
+  // update the stencil weights using the updated hydraulic aperture
+  poromechanicsSolver()->flowSolver()->updateStencilWeights( domain );
+
 }
 
 void SinglePhasePoromechanicsConformingFractures::updateHydraulicApertureAndFracturePermeability( DomainPartition & domain )
--- a/src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoromechanicsConformingFractures.hpp
+++ b/src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoromechanicsConformingFractures.hpp
@@ -114,9 +114,9 @@ public:
   virtual void updateState( DomainPartition & domain ) override final;
 
 
-  // virtual void implicitStepComplete( real64 const & time_n,
-  //                                    real64 const & dt,
-  //                                    DomainPartition & domain ) override;
+  virtual void implicitStepComplete( real64 const & time_n,
+                                     real64 const & dt,
+                                     DomainPartition & domain ) override final;
 
   bool resetConfigurationToDefault( DomainPartition & domain ) const override final;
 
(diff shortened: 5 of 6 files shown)


src/coreComponents/physicsSolvers/fluidFlow/FlowSolverBase.cpp
```diff
@@ -736,7 +736,7 @@ void FlowSolverBase::prepareStencilWeights( DomainPartition & domain ) const
     FiniteVolumeManager const & fvManager = numericalMethodManager.getFiniteVolumeManager();
     FluxApproximationBase const & fluxApprox = fvManager.getFluxApproximation( getDiscretizationName() );
     ElementRegionManager::ElementViewAccessor< arrayView1d< real64 const > > hydraulicAperture =
-      mesh.getElemManager().constructViewAccessor< array1d< real64 >, arrayView1d< real64 const > >( fields::flow::hydraulicAperture::key() );
+      mesh.getElemManager().constructViewAccessor< array1d< real64 >, arrayView1d< real64 const > >( fields::flow::aperture0::key() );
```
CusiniM, line 739: `aperture0` is not the aperture at the previous time step. I don't think this makes sense. I guess that what you would like is to use `aperture_n`.

paveltomin: aperture0 is aperture_n, no?
https://github.com/GEOS-DEV/GEOS/blob/decc28110acf58053a0d6911f5ea7a4d9d7f76d1/src/coreComponents/physicsSolvers/fluidFlow/FlowSolverBaseFields.hpp#L147

CusiniM:
The issue is that this will effectively change the solution. I can see how it would help with convergence because it removes a nonlinearity but it is essentially treating the main source of coupling explicitly.

src/coreComponents/physicsSolvers/fluidFlow/FlowSolverBase.cpp
liyangrock: `aperture0` is the aperture at the previous time step. It is updated here in `SinglePhaseBase::implicitStepSetup` function.
https://github.com/GEOS-DEV/GEOS/blob/decc28110acf58053a0d6911f5ea7a4d9d7f76d1/src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.cpp#L664-L666

liyangrock:
> The issue is that this will effectively change the solution. I can see how it would help with convergence because it removes a nonlinearity but it is essentially treating the main source of coupling explicitly.

To fully account for the nonlinearity, we need to include the derivatives of `m_weights` w.r.t displacement and pressure, since it is a function of `aperture`. However, I think this is not essential.
https://github.com/GEOS-DEV/GEOS/blob/decc28110acf58053a0d6911f5ea7a4d9d7f76d1/src/coreComponents/finiteVolume/SurfaceElementStencil.hpp#L483-L484

The updating scheme of `m_weights` in `SurfaceElementStencil.hpp` is lagging, just as stated by @karimifard in https://github.com/GEOS-DEV/GEOS/issues/2853#issuecomment-1856491359

>  Essentially the update of the aperture in the flow discretization is lagging and the Jacobian accounts only for the impact of the aperture on permeability variation.

karimifard:
Yes, it is lagging but it's happening within the nonlinear iterations so when the solution has converged at a given time step all quantities are up to date.

CusiniM:
> > The issue is that this will effectively change the solution. I can see how it would help with convergence because it removes a nonlinearity but it is essentially treating the main source of coupling explicitly.
> 
> To fully account for the nonlinearity, we need to include the derivatives of `m_weights` w.r.t displacement and pressure, since it is a function of `aperture`. However, I think this is not essential. https://github.com/GEOS-DEV/GEOS/blob/decc28110acf58053a0d6911f5ea7a4d9d7f76d1/src/coreComponents/finiteVolume/SurfaceElementStencil.hpp#L483C1-L484C89
> 

We do account for derivatives of the transmissibility w.r.t. displacement and pressure in the flux terms. The only thing that was done was to lag part of the update by 1 iteration  but, as pointed out by Mohammad, at the end of the solution procedure for a timestep everything is up to date. 

> The updating scheme of `w_weights` in `SurfaceElementStencil.hpp` is lagging, just as stated by @karimifard in [#2853 (comment)](https://github.com/GEOS-DEV/GEOS/issues/2853#issuecomment-1856491359)
> 
> > Essentially the update of the aperture in the flow discretization is lagging and the Jacobian accounts only for the impact of the aperture on permeability variation.

liyangrock:
Updating the stencil weights at the end of each Newton iteration is a standard Newton iteration, not lagging. 
$$J(y_k^{n+1})(y_{k+1}^{n+1} - y_k^{n+1}) = -R(y_k^{n+1})$$

$k$ is the Newton iteration count and $n$ is the time increment count.

If we try to use the standard Newton iteration, we should account for the contribution of the derivatives of `m_weights` (not stencil `weights`) in `dt0_dvar1`, `dt1_dvar1`, `dt0_dvar2`, and `dt1_dvar2` as well, since the `m_weights` is a function of `aperture`.

https://github.com/GEOS-DEV/GEOS/blob/decc28110acf58053a0d6911f5ea7a4d9d7f76d1/src/coreComponents/finiteVolume/SurfaceElementStencil.hpp#L483-L484

https://github.com/GEOS-DEV/GEOS/blob/decc28110acf58053a0d6911f5ea7a4d9d7f76d1/src/coreComponents/finiteVolume/SurfaceElementStencil.hpp#L497-L498

CusiniM:
> Updating the stencil weights at the end of each Newton iteration is a standard Newton iteration, not lagging. J(ykn+1)(yk+1n+1−ykn+1)=−R(ykn+1)
> 
> k is the Newton iteration count and n is the time increment count.
> 
> If we try to use the standard Newton iteration, we should account for the contribution of the derivatives of `m_weights` (not stencil `weights`) in `dt0_dvar1`, `dt1_dvar1`, `dt0_dvar2`, and `dt1_dvar2` as well, since the `m_weights` is a function of `aperture`.
> 
> https://github.com/GEOS-DEV/GEOS/blob/decc28110acf58053a0d6911f5ea7a4d9d7f76d1/src/coreComponents/finiteVolume/SurfaceElementStencil.hpp#L483-L484
> 
> https://github.com/GEOS-DEV/GEOS/blob/decc28110acf58053a0d6911f5ea7a4d9d7f76d1/src/coreComponents/finiteVolume/SurfaceElementStencil.hpp#L497-L498

Yes, we are all well aware of this. The choice was made for a few reasons. 

My initial point, though, was that this PR goes in the complete opposite direction by introducing an explicit update of the dependency.

liyangrock:
> > Updating the stencil weights at the end of each Newton iteration is a standard Newton iteration, not lagging. J(ykn+1)(yk+1n+1−ykn+1)=−R(ykn+1)
> > k is the Newton iteration count and n is the time increment count.
> > If we try to use the standard Newton iteration, we should account for the contribution of the derivatives of `m_weights` (not stencil `weights`) in `dt0_dvar1`, `dt1_dvar1`, `dt0_dvar2`, and `dt1_dvar2` as well, since the `m_weights` is a function of `aperture`.
> > https://github.com/GEOS-DEV/GEOS/blob/decc28110acf58053a0d6911f5ea7a4d9d7f76d1/src/coreComponents/finiteVolume/SurfaceElementStencil.hpp#L483-L484
> > 
> > https://github.com/GEOS-DEV/GEOS/blob/decc28110acf58053a0d6911f5ea7a4d9d7f76d1/src/coreComponents/finiteVolume/SurfaceElementStencil.hpp#L497-L498
> 
> Yes, we are all well aware of this. The choice was made for a few reasons.
> 
> My initial point, though, was that this PR goes in the complete opposite direction by introducing an explicit update of the dependency.

Yes. In this PR, the surface element stencil `weights` is updated by old hydraulic aperture `aperture0`, while the transmissibility component $\frac{w^2}{12}$ is still updated by current hydraulic aperture `aperture`.

src/coreComponents/physicsSolvers/fluidFlow/FlowSolverBase.cpp
CusiniM: ah yeah, my bad. We should change the name tbh so that it is the same as all other variables. I would do it separately, anyways.

CusiniM:
> > > Updating the stencil weights at the end of each Newton iteration is a standard Newton iteration, not lagging. J(ykn+1)(yk+1n+1−ykn+1)=−R(ykn+1)
> > > k is the Newton iteration count and n is the time increment count.
> > > If we try to use the standard Newton iteration, we should account for the contribution of the derivatives of `m_weights` (not stencil `weights`) in `dt0_dvar1`, `dt1_dvar1`, `dt0_dvar2`, and `dt1_dvar2` as well, since the `m_weights` is a function of `aperture`.
> > > https://github.com/GEOS-DEV/GEOS/blob/decc28110acf58053a0d6911f5ea7a4d9d7f76d1/src/coreComponents/finiteVolume/SurfaceElementStencil.hpp#L483-L484
> > > 
> > > https://github.com/GEOS-DEV/GEOS/blob/decc28110acf58053a0d6911f5ea7a4d9d7f76d1/src/coreComponents/finiteVolume/SurfaceElementStencil.hpp#L497-L498
> > 
> > 
> > Yes, we are all well aware of this. The choice was made for a few reasons.
> > My initial point, though, was that this PR goes in the complete opposite direction by introducing an explicit update of the dependency.
> 
> Yes. In this PR, the surface element stencil `weights` is updated by old hydraulic aperture `aperture0`, while the transmissibility component w212 is still updated by current hydraulic aperture `aperture`.

If it really solves convergence issues for some of those test cases we can maybe keep it as optional by introducing a flag.

liyangrock:
I have tested the `pknViscosityDominated_benchmark.xml` with this PR on my side and the results match well.

The MPI parameters I used for the case is as follows:
`export OMP_NUM_THREADS=2`
`mpirun -np 45 geos -i ./pknViscosityDominated_benchmark.xml -x 3 -y 3 -z 5`

**Following figure is the results from this PR `hydrofractureSolver` and it took about `01h13m04s` to complete the run**
Figure_1

**Following figure is the results from `hydrofractureSolver` before merging PR #2623  and it took about `01h44m03s` to complete the run**
Figure_oldSolver



**`hydrofractureSolver` from the latest develop branch has convergence issues (performance degradation #2853 ) and requires frequent time cutback for most time increment.**





Please note that I changed the `penaltyStiffness` value from `1.0e0` to `1.0e12`, otherwise the simulation would stop around `60s`.

https://github.com/GEOS-DEV/GEOS/blob/741b6ab459c01266627b518df67769dd0dbf6700/inputFiles/hydraulicFracturing/pknViscosityDominated_base.xml#L55

castelletto1:
Any updates on this PR?

liyangrock:
> Any updates on this PR?

No further updates. I believe this Pull Request should resolve the performance problem. If required, I can modify this Pull Request to address any merge conflicts.

paveltomin:
> > Any updates on this PR?
> 
> No further updates. I believe this Pull Request should resolve the performance problem. If required, I can modify this Pull Request to address any merge conflicts.

Since it is going to be hard to agree what way is the right, how about we make an option (flag) to switch the way the weights are updated.
@CusiniM @karimifard what do you think?

karimifard:
> > > Any updates on this PR?
> > 
> > 
> > No further updates. I believe this Pull Request should resolve the performance problem. If required, I can modify this Pull Request to address any merge conflicts.
> 
> Since it is going to be hard to agree what way is the right, how about we make an option (flag) to switch the way the weights are updated. @CusiniM @karimifard what do you think?

I am fine if we have a flag to choose how the stencils are updated.

paveltomin:
@liyangrock are you willing to implement a flag to switch between current and new way ?

liyangrock:
> @liyangrock are you willing to implement a flag to switch between current and new way ?

Hi @paveltomin , I originally thought that you were planning to adopt new weight coefficients, so I didn't update this PR. I will implement a flag to switch between the two methods.

liyangrock:
Hi @paveltomin, I’ve submitted a new PR to tackle issue #2853, please see #3108.

PR #2924, opened 2024-01-05
Fix the naming of the fluid heat capacity 

### LINKED ISSUE #2923: Volumetric Heat Capacity Values are Off for THM Examples
In some of the thermoporomechanics examples, the specified volumetric heat capacities look incorrect (for example, see: https://github.com/GEOS-DEV/GEOS/blob/develop/inputFiles/thermoPoromechanics/ThermoPoroElastic_consolidation_base.xml#L49-L68 ).

In these, the rock phase has a value equal to 1.672e5, and the fluid has a value of 1.672e2.  Looking at some online references, reasonable values for rock are about 3e6, and for water is about 4e6.  This is obviously off from the reality.   Looking at the formulation we use for the THM model with @frankfeifan , we suspect that the existing models are effectively ignoring everything aside from conduction. 

In addition, I've found that when these values are fixed, that the solvers can struggle to converge, and report very large residuals for Renergy.

To address this issue, we should do the following:

1) Use appropriate values for the volumetric heat capacity (or at least ones withe the correct ratio between solid/fluid)
2) Run the examples, and address any scaling problems that arise (we may need to address the MGR recipe)

frankfeifan:
This PR changes the naming of heat capacity in the fluid model as previous `volumetricHeatCapacity` is misleading. We are actually using the heat capacity for a specific fluid mass (unit: J/K/kg). 

This PR is also related to the issue #2923 

integratedTests PR: https://github.com/GEOS-DEV/integratedTests/pull/81


jhuang2601:
What is the correct unit and name of the Heat Capacity in `SolidInternalEnergy`?

frankfeifan:
> What is the correct unit and name of the Heat Capacity in `SolidInternalEnergy`?

For `SolidInternalEnergy`, the unit of heat capacity should be J/K/m^3 according to the calculation [here.](https://github.com/GEOS-DEV/GEOS/blob/39e75de52fbb5b36d862ddb7adc95c16250771e1/src/coreComponents/physicsSolvers/fluidFlow/ThermalSinglePhaseBaseKernels.hpp#L221) So the name is correct for the solid model.

jhuang2601:
> > What is the correct unit and name of the Heat Capacity in `SolidInternalEnergy`?
> 
> For `SolidInternalEnergy`, the unit of heat capacity should be J/K/m^3 according to the calculation [here.](https://github.com/GEOS-DEV/GEOS/blob/39e75de52fbb5b36d862ddb7adc95c16250771e1/src/coreComponents/physicsSolvers/fluidFlow/ThermalSinglePhaseBaseKernels.hpp#L221) So the name is correct for the solid model.

Thanks @frankfeifan for this clarification. For the existing [tutorial example](https://geosx-geosx.readthedocs-hosted.com/en/latest/docs/sphinx/advancedExamples/validationStudies/thermoPoromechanics/thermalConsolidation/Example.html), heat capacity for solid should be 1.672e5 J/K/m^3, which was provided by reference work.

jhuang2601 (approved):

frankfeifan:
> > > What is the correct unit and name of the Heat Capacity in `SolidInternalEnergy`?
> > 
> > 
> > For `SolidInternalEnergy`, the unit of heat capacity should be J/K/m^3 according to the calculation [here.](https://github.com/GEOS-DEV/GEOS/blob/39e75de52fbb5b36d862ddb7adc95c16250771e1/src/coreComponents/physicsSolvers/fluidFlow/ThermalSinglePhaseBaseKernels.hpp#L221) So the name is correct for the solid model.
> 
> Thanks @frankfeifan for this clarification. For the existing [tutorial example](https://geosx-geosx.readthedocs-hosted.com/en/latest/docs/sphinx/advancedExamples/validationStudies/thermoPoromechanics/thermalConsolidation/Example.html), heat capacity for solid should be 1.672e5 J/K/m^3, which was provided by reference work.

Yes. I think the values of heat capacity for both fluid and solid in this tutorial should be reasonable, maybe a bit smaller than their real values from experiment, e.g., the heat capacity of water is around 4e3 J/K/kg. But I think it should be fine just for model verification. 

The issue here is that the current name of fluid heat capacity is misleading, and its unit is actually different from that of the solid (J/K/kg v.s. J/K/m^3). That's why the heat capacity values for the solid and fluid are very different in the tutorial example.

sytuannguyen:
@frankfeifan I would propose to consider Specific Heat Capacity in place of Heat Capacity.

frankfeifan:
> @frankfeifan I would propose to consider Specific Heat Capacity in place of Heat Capacity.

Yes. `specificHeatCapacity` is a better option. I have updated the code accordingly.

CusiniM (approved):

cssherman (approved):

PR #2927, opened 2024-01-08
Use `LvArray::math` functions

### LINKED ISSUE #2926: Compilation issue
**Describe the bug**

compilation issue on several files
` error: using integer absolute value function 'abs' when argument is of floating point type [-Werror,-Wabsolute-value] ... : note: use function 'std::abs' instead
`
**To Reproduce**
clone & compile using gcc 13.2.1

**Expected behavior**
Compilation...

**Screenshots**
n.a.

- Machine: unix laptop
- Compiler: gcc (GCC) 13.2.1 20230801
- GEOSX Version: develop branch

**Additional context**
n.a.

untereiner:
closes #2926


tbeltzun (approved):

src/coreComponents/physicsSolvers/wavePropagation/ElasticFirstOrderWaveEquationSEMKernel.hpp
```diff
@@ -305,9 +305,9 @@ struct DampingMatrixKernel
           for( localIndex q = 0; q < numNodesPerFace; ++q )
           {
             real32 const aux = density[e] * m_finiteElement.computeDampingTerm( q, xLocal );
-            real32 const localIncrementx = density[e] * (velocityVp[e] * abs( nx ) + velocityVs[e] * sqrt( pow( ny, 2 ) + pow( nz, 2 ) ) ) * aux;
-            real32 const localIncrementy = density[e] * (velocityVp[e] * abs( ny ) + velocityVs[e] * sqrt( pow( nx, 2 ) + pow( nz, 2 ) ) ) * aux;
-            real32 const localIncrementz = density[e] * (velocityVp[e] * abs( nz ) + velocityVs[e] * sqrt( pow( nx, 2 ) + pow( ny, 2 ) ) ) * aux;
+            real32 const localIncrementx = density[e] * (velocityVp[e] * std::abs( nx ) + velocityVs[e] * sqrt( pow( ny, 2 ) + pow( nz, 2 ) ) ) * aux;
```
TotoGaz: Don't you think you should be using `LvArray::math::abs` instead

untereiner: Why? Someone might change one day the define real32 ?

untereiner: What about sqrt and pow? I don’t see any `using namespace Lvarray::math`. That would be _koenig’s lookup aware_ by the way

TotoGaz: > Why? Someone might change one day the define real32 ?

Because the `LvArray::math` functions are supposed to be `GPU` proof. Which makes sense in a `__device__` function.

CusiniM:
This does not seem to be `ready to be merged`. Neither @TotoGaz nor @rrsettgast have approved it. And it looks like @TotoGaz had some comments.

src/coreComponents/physicsSolvers/wavePropagation/ElasticFirstOrderWaveEquationSEMKernel.hpp
klevzoff: `std::abs` will not compile on device. Sadly, `LvArray::math::abs` [does not provide](https://github.com/GEOS-DEV/LvArray/blob/develop/src/math.hpp#L401-L439) correct overload for floating point values either (it dispatches to `::abs(int)` on device which causes truncation).

You have 2 options:
  * use `fabsf` (no namespace)
  * or implement correct overloads in `LvArray/src/math.hpp` (dispatch to `::fabsf` for `float` and `::fabs` for `double`) and use them.

`sqrt`, `pow`, `fabs`, etc. are functions in the global namespace inherited from the C standard. On device compilation trajectory, device compilers substitute their own versions, so these are always safe to use. The value of `LvArray::math` versions is in generic code, since they can usually dispatch appropriately depending on argument type (e.g. to `fabsf` vs `fabs` vs `abs`).

TotoGaz: Thanks @klevzoff !
Do you know why `LvArray::math::abs` does not offer the proper impl? I did not check prior to my comment, do I thought it was OK.

klevzoff: I don't, I also assumed it was correct, but after looking at the source code, it doesn't seem to be, although it's possible I missed something. Moreover, searching through GEOS, we seem to be using that function on floating point values in quite a few places (including device kernels), so perhaps through some compiler magic it does work correctly in the end? But at least looking at [CUDA docs](https://docs.nvidia.com/cuda/cuda-math-api/index.html), `::abs(int)` seems to be the only version of `abs` (those are C functions, so they don't even have overloads). @corbett5 do you happen to know?

corbett5: No that is a good point and if correct seems like a significant oversight on my part. I'll modify the tests real quick.

corbett5: So I think `LvArray::math::abs` works as expected. The following produces

```c++
#include <RAJA/RAJA.hpp>

template< typename POLICY, typename INDEX_TYPE, typename LAMBDA >
inline void forall( INDEX_TYPE const max, LAMBDA && body )
{
  RAJA::forall< POLICY >( RAJA::TypedRangeSegment< INDEX_TYPE >( 0, max ), std::forward< LAMBDA >( body ) );
}

template< typename POLICY >
void fooBar()
{
  forall< POLICY >( 1, [] __host__ __device__ ( int )
  {
    double x = - M_PI;
    printf("x = %f, ::abs(x) = %f\n", x, ::abs(x));
  } );
}

int main( int argc, char * * argv )
{
  printf("On CPU\n");
  fooBar< RAJA::seq_exec >();

  printf("On GPU\n");
  fooBar< RAJA::cuda_exec< 32 > >();
  
  return 0;
}
```

```
On CPU
x = -3.141593, ::abs(x) = 3.141593
On GPU
x = -3.141593, ::abs(x) = 3.141593
```

Looking at the [cuda docs](https://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH.html#group__CUDA__MATH) they do supply overloads for most common functions, which I assume includes `abs`.

corbett5: On further inspection compiling with CUDA automatically introduces `abs` (with the appropriate overloads) and `printf` into the namespace whether compiling for device or not. The following works as expected on Lassen, but fails to compile on Quartz.

```c++
int main( int argc, char * * argv )
{
  double x = -3.14;
  printf("x = %f, abs(x) = %f\n", x, abs(x));
  
  return 0;
}
```

Furthermore when I add `#include <cmath>` and `#include <cstdio>` with clang 14.0.6 on Quartz I get warnings telling me that I'm using the wrong `abs` function.

CusiniM: Okay, so should we just use `LvArray::math::abs` everywhere?  

The only thing I do not understand is why the code was not giving us errors for the `cuda` jobs with `std::abs`. I would have expected the usual... `calling a host function...`

corbett5: Yes, that is my suggestion. And I would suggest using that for all math functions.

If you look at the last sentence note [here](https://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH.html#group__CUDA__MATH) my guess is that `std::abs` is one of those functions.

untereiner: Thank you all for these answers. I may be wrong but for me there is a need of deep knowledge of the different compilers/ std libs to be able to know if the good overload will be in scope or not. I remember there was already an issue about abs. Is it possible or even desirable to have a uniform API for these math functions ?

corbett5: `LvArray::math` is intended to provide the uniform API. The idea is that `LvArray::math::foo` should be a drop in replacement for `std::foo` with the added benefit of working on gpu devices (not just with CUDA). It also has the added benefit of providing host wrappers for things like `sincos` that CUDA provides.

untereiner: Ok! Uniformisation by replacing also sqrt, pow (others ?) would help to not call std versions I think

CusiniM: > Yes, that is my suggestion. And I would suggest using that for all math functions.
> 
> If you look at the sentence note [here](https://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH.html#group__CUDA__MATH) my guess is that `std::abs` is one of those functions.

ah yeah, makes sense.

untereiner: @TotoGaz @klevzoff is last commit acceptable ? If a pow function is added to lvarray it will be used first.

TotoGaz (approved):
This LGTM. I'm waiting for @klevzoff 's expertise though!

untereiner:
as a side note, if this solution is acceptable it should be added to another places.

src/coreComponents/physicsSolvers/wavePropagation/ElasticFirstOrderWaveEquationSEMKernel.hpp
klevzoff: It's all good by me as long as it compiles on all platforms :)

Note that adding `using namespace LvArray::math` does not necessarily mean the `LvArray` version will be called, it just adds another set of templates/overloads in the set and lets overload resolution pick the winner. For example, given
```c++
double pow(double base, double exp);
```
in the global namespace coming from C stdlib, and
```c++
template<typename T>
T pow(T base, T exp);
```
in `LvArray::math`, the first one will be picked (since a non-template perfect match wins over a template). The end result is the same though. The details about when C stdlib functions are added or not to the global namespace are [notoriously](https://stackoverflow.com/a/2588103) [murky](https://stackoverflow.com/a/32606280) and even less clear with CUDA (we seem to get more than is documented). The whole situation is a mess :) The one thing we should always avoid is using `std::` version in kernel code, which this PR correctly does.

klevzoff (approved):

src/coreComponents/physicsSolvers/wavePropagation/ElasticFirstOrderWaveEquationSEMKernel.hpp
corbett5: I don't know if this PR (or anywhere else for that matter) has `using namespace LvArray::math`, but for exactly that reason it should be avoided. Using `LvArray::math::pow` is not a elegant solution but at least there is zero ambiguity.

TotoGaz: OK, so let's wait for @untereiner to explicitly use `LvArray::math`.

Maybe `using LvArray::math::pow;`, `using LvArray::math::abs;`... can help at having the formula a bit short?

In the long term, maybe could we have a `geos::math` namespace that would allow us to forward the `LvArray::math` namespace, so its usage will be more natural?

untereiner: Thank you @klevzoff. I thought the opposite: if you bring functions from a namespace into scope then they would be prioritized. Maybe the C versions should be put in a namespace and should not hang around and pollute the global one.

rrsettgast (approved):

PR #2932, opened 2024-01-12
Add log for opened deck filepath

arng40:
Added in the log : 
- opened xml filepath 
- included xml filepath

```
Opened XML file: /home/adudes/total/codes/GEOS/inputFiles/compositionalMultiphaseWell/simpleCo2InjTutorial_smoke.xml
Included additionnal XML file: /home/adudes/total/codes/GEOS/inputFiles/compositionalMultiphaseWell/simpleCo2InjTutorial_base.xml
```


MelReyCG (approved):
It will help the users to keep track of the source of their logs.
Seems OK to me.

src/coreComponents/mainInterface/ProblemManager.cpp
```diff
@@ -186,6 +186,13 @@ void ProblemManager::parseCommandLineInput()
   OutputBase::setOutputDirectory( outputDirectory );
 
   string & inputFileName = commandLine.getReference< string >( viewKeys.inputFileName );
+
+  for( string xmlFile : opts.inputFileNames )
```
TotoGaz:
```suggestion
  for( string const & xmlFile : opts.inputFileNames )
```

src/coreComponents/mainInterface/ProblemManager.cpp
```diff
@@ -186,6 +186,13 @@ void ProblemManager::parseCommandLineInput()
...
+  {
+    Path::pathPrefix() = getAbsolutePath( xmlFile );
```
TotoGaz: Having `Path::pathPrefix` returning a mutable reference to the global variable looks smells wrong.
I know it's not your PR, but could you add a `Path::setPathPrefix` and remove the mutable calls here and there (as well as the mutable version of the function)?

EDIT: So why are you changing the `Path::pathPrefix` to a different value for each file? Only one file will have its `prefix` correct and the others not? How is this working?

TotoGaz (commented):
Hi @arng40 thank you for your job and welcome!

src/coreComponents/mainInterface/ProblemManager.cpp
MelReyCG: Oops, I did not see that. He should not have modified `Path::pathPrefix()` here as it was not his intention (he just would like to simplify the path (get rid of `./`, `../`).
A local variable would be sufficient here, but getting rid of the mutable calls would be nice too!

@TotoGaz Under the hood, when multiple input XML are used, it create a `composite_input.xml` file, but the working dir is still the current directory from where GEOS was started.

arng40:
Thank you @TotoGaz for reviewing and advice, I've taken into account what you've said

arng40:
My PR seems to pass the integrated tests
```
ATS END OF RUN STARTED January 16, 2024 16:39:30
2024-01-16T17:50:22.7610108Z     Status     : TestCase                              : Directory                                                 : Elapsed : Resources : TestStep  
2024-01-16T17:50:22.7611466Z     ---------- : ------------------------------------- : --------------------------------------------------------- : ------- : --------- : ----------
2024-01-16T17:50:22.7613070Z     FAIL CHECK : PoroElastic_impermeableFault_smoke_01 : poroElasticCoupling/PoroElastic_impermeableFault_smoke_01 : 0:00:04 : 0:00:04   : restartcheck
2024-01-16T17:50:22.7615077Z     FAIL CHECK : PoroElastic_permeableFault_smoke_01   : poroElasticCoupling/PoroElastic_permeableFault_smoke_01   : 0:00:04 : 0:00:04   : restartcheck
2024-01-16T17:50:22.7616770Z     ---------- : ------------------------------------- : --------------------------------------------------------- : ------- : --------- : ----------
```

MelReyCG (approved):

TotoGaz (approved):

PR #2941, opened 2024-01-17
Activate water vapourisation table

dkachuma:
Activates the water vapourisation table in CO2-Brine models. This was commented out because it was failing to compile on GPU systems. The issue was that with the extra table `CO2BrineFluid::KernelWrapper`, became too large to be captured in a lambda for kernel launch.

This is resolved by a change in `LvArray` ([PR311](https://github.com/GEOS-DEV/LvArray/pull/311)) which removes a member variable in `ArrayView` and reduces the memory enough to manage the compilation.


jafranc (approved):
Thank you for enabling this feature. :+1:

PR #2942, opened 2024-01-17
Make useSimpleAccumulation=1 default for isothermal

paveltomin:
When `useSimpleAccumulation=1` accumulation term is assembled as `PV * rho_c` instead of  looping over phases.
Much shorter, cleaner, and less dependencies on props derivatives.

https://github.com/GEOS-DEV/integratedTests/pull/85


src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -185,6 +185,8 @@ void CompositionalMultiphaseBase::postProcessInput()
                         getWrapperDataContext( viewKeyStruct::solutionChangeScalingFactorString() ) <<
                         ": The solution change scaling factor must be smaller or equal to 1.0" );
 
+  if( m_isThermal )
```
dkachuma: I would add a message here if the user tries to set `useSimpleAccumulation` on a thermal simulation.

dkachuma (commented):
I can see that the original `computeAccumulation` ([IsothermalCompositionalMultiphaseBaseKernels.hpp:769-774](https://github.com/GEOS-DEV/GEOS/blob/pt/simple-acc-default/src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseBaseKernels.hpp#L769-L774)) does a non-zero diagonal check albeit in debug only. Does the new `computeAccumulationSimple` also guarantee this? I think the assert at [IsothermalCompositionalMultiphaseBaseKernels.hpp:799-800](https://github.com/GEOS-DEV/GEOS/blob/pt/simple-acc-default/src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseBaseKernels.hpp#L799-L800) is not correctly formed and will never hit. I also think the check should be performed after the accumulation is done.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
paveltomin: good point but, is there a way to tell if user tries or useSimpleAccumulation=1 just came as default value?

paveltomin:
> I can see that the original `computeAccumulation` ([IsothermalCompositionalMultiphaseBaseKernels.hpp:769-774](https://github.com/GEOS-DEV/GEOS/blob/pt/simple-acc-default/src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseBaseKernels.hpp#L769-L774)) does a non-zero diagonal check albeit in debug only. Does the new `computeAccumulationSimple` also guarantee this? I think the assert at [IsothermalCompositionalMultiphaseBaseKernels.hpp:799-800](https://github.com/GEOS-DEV/GEOS/blob/pt/simple-acc-default/src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseBaseKernels.hpp#L799-L800) is not correctly formed and will never hit. I also think the check should be performed after the accumulation is done.

Good catch, thanks
I corrected the check and it should work fine on the flight, but maybe worth to unify that zero diag check, let me have another look

paveltomin:
> > I can see that the original `computeAccumulation` ([IsothermalCompositionalMultiphaseBaseKernels.hpp:769-774](https://github.com/GEOS-DEV/GEOS/blob/pt/simple-acc-default/src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseBaseKernels.hpp#L769-L774)) does a non-zero diagonal check albeit in debug only. Does the new `computeAccumulationSimple` also guarantee this? I think the assert at [IsothermalCompositionalMultiphaseBaseKernels.hpp:799-800](https://github.com/GEOS-DEV/GEOS/blob/pt/simple-acc-default/src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseBaseKernels.hpp#L799-L800) is not correctly formed and will never hit. I also think the check should be performed after the accumulation is done.
> 
> Good catch, thanks I corrected the check and it should work fine on the flight, but maybe worth to unify that zero diag check, let me have another look

Just unified the check, I hope it looks good now

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
jafranc:
```suggestion
  if( m_isThermal && m_useSimpleAccumulation==1)
  {
        GEOS_LOG_RANK_0("simpleAccumulation is not yet implemented for thermal simulation. Switched to phase sum accumulation.");
        m_useSimpleAccumulation=0;
  }
```

jafranc (approved):
Great ! Thanks

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
paveltomin: ok, I just used Jacques' version

dkachuma (approved):

PR #2944, opened 2024-01-18
Misc cleanup of CI and README

rrsettgast:
- [x] Cleanup badges in README.md
- [x] fix ci_test.yml for non-pr runs
- [x] add badge for actions status


.github/workflows/ci_tests.yml
```diff
@@ -58,19 +56,22 @@ jobs:
...
     runs-on: ubuntu-22.04
     steps:
-    - name: Check that the PR is assigned
+    - name: If this is a PR, Check that it is assigned
       run: |
-        echo "There are ${{ needs.is_pull_request_a_draft.outputs.NUM_ASSIGNEES }} on this PR."
-        ${{ needs.is_pull_request_a_draft.outputs.NUM_ASSIGNEES > 0 }}
+        if [[ ${{github.event_name}} != 'pull_request' ]]; then exit 0 ; fi
+        pr_json=$(curl -H "Accept: application/vnd.github+json" https://api.github.com/repos/${{ github.repository }}/pulls/${{ github.event.number }})
+        export NUM_ASSIGNEES=$(echo ${pr_json} | jq '.assignees | length')
```
TotoGaz:
```suggestion
       NUM_ASSIGNEES=$(echo ${pr_json} | jq '.assignees | length')
```

.github/workflows/ci_tests.yml
```diff
@@ -58,19 +56,22 @@ jobs:
 
   # PR must be assigned to be merged.
   # This job will fail if this is not the case.
-  is_pull_request_assigned:
-    needs: [is_pull_request_a_draft]
+  if_not_unassigned_pull_request:
```
TotoGaz, line 59: Maybe names like `unassigned_pr_guard` or `draft_pr_guard` can be nice as well?

TotoGaz (approved):

bmhan12 (approved):

.github/workflows/ci_tests.yml
rrsettgast: yeah...I struggled with a name...but this seemed pretty descriptive of the result. It is true if this action is "not on an unassigned pull request"....otherwise "it is an unassigned pull request" and it is a fail.

I think we should consider a more intrusive restructure of the yaml s.t. we have a "query" phase to get all the variables...then tests on those variables for things like (is_draft_pr, is_unassigned_pr, run_cuda_tests) then have the testing blocks depend on those instead of inferred....for instance the cuda tests only run if the label exists, but we have a test for a run_cuda label, then the tests depend on that. This way we can see the workflow better.

PR #2946, opened 2024-01-19
Use output dir (option "-o") for printing out PVT tables

paveltomin:
Somehow it got missed in https://github.com/GEOS-DEV/GEOS/pull/2884


src/coreComponents/functions/CMakeLists.txt
```diff
@@ -27,7 +27,7 @@ if( ENABLE_MATHPRESSO )
 endif()
 
 
-set( dependencyList ${parallelDeps} codingUtilities dataRepository )
+set( dependencyList ${parallelDeps} codingUtilities dataRepository fileIO )
```
CusiniM, line 30: I wish we didn't need to add additional dependencies but I guess this one should not be a big issue.

CusiniM (approved):

PR #2897, opened 2023-12-14
Regions for sequential poromechanics solver

paveltomin:
The goal is to support flow solver to be not applied for some of the regions:
```
    <CompositionalMultiphaseReservoirPoromechanics
      name="reservoirSystem"
      solidSolverName="linearElasticity"
      reservoirAndWellsSolverName="reservoirAndWells"
      logLevel="1"
      targetRegions="{ channel }">
      <NonlinearSolverParameters
        couplingType="Sequential"
        subcycling="1"
        newtonTol="1.0e-8"
        lineSearchAction="None"
        newtonMaxIter="100"/>
      <LinearSolverParameters
        directParallel="0"/>
    </CompositionalMultiphaseReservoirPoromechanics>
    
    <SolidMechanicsLagrangianSSLE
      name="linearElasticity"
      timeIntegrationOption="QuasiStatic"
      logLevel="1"
      discretization="FE1"
      targetRegions="{ channel, **barrier**  }"/>

    <CompositionalMultiphaseReservoir
      name="reservoirAndWells"
      flowSolverName="twoPhaseFlow"
      wellSolverName="compositionalMultiphaseWell"
      logLevel="1"
      targetRegions="{ channel, wellRegion1, wellRegion2 }">
      <NonlinearSolverParameters
        newtonTol="1.0e-8"
        lineSearchAction="None"
        newtonMaxIter="40"/>
      <LinearSolverParameters
        directParallel="0"/>
    </CompositionalMultiphaseReservoir>
```

PR #2766, opened 2023-10-17
Fix bulk modulus assignment and remove magic if from fixed stress porosity

paveltomin:
Results for PoroElastic_Mandel_benchmark:

BEFORE
```
SinglePhaseFlow, number of time steps: 406
SinglePhaseFlow, number of successful nonlinear iterations: 5228
SinglePhaseFlow, number of successful linear iterations: 5228
SinglePhaseFlow, number of time step cuts: 0
SinglePhaseFlow, number of discarded nonlinear iterations: 0
SinglePhaseFlow, number of discarded linear iterations: 0
SinglePhaseFlow: apply solution time = 0.415142593 s (min), 0.415142593 s (max)
SinglePhaseFlow: assemble time = 12.163587556 s (min), 12.163587556 s (max)
SinglePhaseFlow: convergence check time = 0.296807929 s (min), 0.296807929 s (max)
SinglePhaseFlow: linear solver create time = 0.564239068 s (min), 0.564239068 s (max)
SinglePhaseFlow: linear solver setup time = 4.399641692 s (min), 4.399641692 s (max)
SinglePhaseFlow: linear solver solve time = 0.371110135 s (min), 0.371110135 s (max)
SinglePhaseFlow: linear solver total time = 5.47389087 s (min), 5.47389087 s (max)
SinglePhaseFlow: update state time = 0.296379135 s (min), 0.296379135 s (max)
lagsolve, number of time steps: 406
lagsolve, number of successful nonlinear iterations: 5227
lagsolve, number of successful linear iterations: 5227
lagsolve, number of time step cuts: 0
lagsolve, number of discarded nonlinear iterations: 0
lagsolve, number of discarded linear iterations: 0
lagsolve: apply solution time = 0.694335835 s (min), 0.694335835 s (max)
lagsolve: assemble time = 46.35469781 s (min), 46.35469781 s (max)
lagsolve: convergence check time = 0.55177106 s (min), 0.55177106 s (max)
lagsolve: linear solver create time = 8.371216151 s (min), 8.371216151 s (max)
lagsolve: linear solver setup time = 128.323329863 s (min), 128.323329863 s (max)
lagsolve: linear solver solve time = 9.865611708 s (min), 9.865611708 s (max)
lagsolve: linear solver total time = 148.527633954 s (min), 148.527633954 s (max)
lagsolve: update state time = 0.001346909 s (min), 0.001346909 s (max)
poroSolve, number of time steps: 406
poroSolve, number of successful nonlinear iterations: 5227
[...]

PR #2952, opened 2024-01-24
Make sequential poromechanics to conserve mass

### LINKED ISSUE #2845: Explicit coupling scheme for poromechanics
Needed for benchmarking purposes, some draft is here https://github.com/GEOS-DEV/GEOS/pull/2646

paveltomin:
Has to go after https://github.com/GEOS-DEV/GEOS/pull/2906 and https://github.com/GEOS-DEV/GEOS/pull/2897

With changes in this PR explicit coupling seems to work fine and I am able to match FIM reference for SEAM case:
image
image
while simulation times:
```
SEQ           total time            05h44m08s (20648.672443899 s)
FIM           total time            1d, 07h38m18s (113898.859705466 s)
```

Also: force outer loop in poromechanics initialization

New baselines: https://github.com/GEOS-DEV/integratedTests/pull/111


src/coreComponents/finiteVolume/TwoPointFluxApproximation.cpp
```diff
@@ -554,8 +555,9 @@ void TwoPointFluxApproximation::initNewFractureFieldsDFM( MeshLevel & mesh,
     }
     fluidPressure_n[newElemIndex] = fluidPressure[newElemIndex];
 #if ALLOW_CREATION_MASS==0
-    // set the initial density of the face element to 0 to enforce mass conservation ( i.e. no creation of mass)
-    dens[newElemIndex] = 0.0;
+    // set the initial mass/density of the face element to 0 to enforce mass conservation ( i.e. no creation of mass)
```
paveltomin: @CusiniM not sure about this piece, it does not seem even compile when ALLOW_CREATION_MASS is 0

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.cpp
```diff
@@ -244,6 +247,34 @@ void SinglePhaseBase::updateFluidModel( ObjectManagerBase & dataGroup ) const
...
+
+  SingleFluidBase & fluid =
+    getConstitutiveModel< SingleFluidBase >( subRegion, subRegion.getReference< string >( viewKeyStruct::fluidNamesString() ) );
+  arrayView2d< real64 const > const density = fluid.density();
+  arrayView2d< real64 const > const density_n = fluid.density_n();
+
+  forAll< parallelDevicePolicy<> >( subRegion.size(), [=] GEOS_HOST_DEVICE ( localIndex const ei )
+  {
+    mass[ei] = porosity[ei][0] * ( volume[ei] + deltaVolume[ei] ) * density[ei][0];
+    if( isZero( mass_n[ei] ) ) // this is a hack for hydrofrac cases
```
paveltomin, line 273: @CusiniM this is also tricky, not sure how to properly do that

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.hpp
Guotong-Ren, line 1: what is the change for here? is the modification to address the bulk modulus property in the first iteration of the iterative coupling?

src/coreComponents/finiteVolume/TwoPointFluxApproximation.cpp
Guotong-Ren: i think in the fracture propagation, it provides different options to initialize the new element. At least in FP, new cells are always filled with fluid once generated.

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.hpp
paveltomin: Sorry, I don't see where are you pointing to?
The bulk modulus issue was addressed in https://github.com/GEOS-DEV/GEOS/pull/2766 and technically got merged into develop branch with https://github.com/GEOS-DEV/GEOS/pull/2568

jhuang2601 (approved):
It works well on SEAM CO2 case as expected!

src/coreComponents/finiteVolume/TwoPointFluxApproximation.cpp
CusiniM: yes, it does offer different options but it was done through macros. This option was basically never used. This should be fixed in https://github.com/GEOS-DEV/GEOS/pull/2960 which we should review and merge.

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.cpp
CusiniM: I think this should also be taken care in the initialization of new fractures. Why is the if statement necessary? isn't `mass_n` always equal to` porosity_n * volume_n * density_n` ?

src/coreComponents/finiteVolume/TwoPointFluxApproximation.cpp
paveltomin: Is it ok if merge it like that for now and refactor after https://github.com/GEOS-DEV/GEOS/pull/2960 is merged?

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.cpp
paveltomin: There is kind of cyclic dependency here https://github.com/GEOS-DEV/GEOS/blob/cb20b88c5e2d9afcf0b727bd8abb70f2c5366772/src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.cpp#L618

- `mass_n` is saved from `mass` in `saveConvergedState`
- if I understand correctly `updatePorosityAndPermeability` needs `saveConvergedState` to be called before (see comment here https://github.com/GEOS-DEV/GEOS/blob/cb20b88c5e2d9afcf0b727bd8abb70f2c5366772/src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.cpp#L646)
- `updateMass` needs to be called the last to update `mass`

I am still not exactly sure how to untangle that...

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -626,17 +631,17 @@ real64 CompositionalMultiphaseBase::updatePhaseVolumeFraction( ObjectManagerBase
   return maxDeltaPhaseVolFrac;
 }
 
-void CompositionalMultiphaseBase::updateFluidModel( ObjectManagerBase & dataGroup ) const
+void CompositionalMultiphaseBase::updateFluidModel( ElementSubRegionBase & subRegion ) const
```
CusiniM: I would have kept this an `ObjectManagerBase` though coz the method could work to update these fields on a `FaceManger` for example.

Issue #2704, opened 2023-09-22
[EPIC] Unit system


MelReyCG:
Issue general information
----------------------------
**What is the requested feature?**
- A unit system for GEOS.

**Is your request related to a specific problem?**

1. Some data structures, as the `TableFunction`, hold data and does not know its unit, so related reports or error message can be misleading.
2. GEOS users cannot specify which unit they want for the various input & output data.
3. The code could be more self-explaining if variables could be declared as `Pressure`, `Distance` or `Viscosity` type rather than `real64`.
4. There is a lack of information on the temperature units (`Kelvin` or `Celsius`) at different places in the code.

----------------------------
- [x] **I.** Add a simple unit system to GEOS so when we store unqualified data, we can add the wanted unit metadata to it.
- GEOS PR that need this system: #2552
----------------------------
- [ ] **II.** Add a statically typed unit system.
- We could add or inspire from a library like [mp-units](https://github.com/mpusz/mp-units).
- We need an extensive, simple and light solution, the goal is not to complexify the code nor to affect the performances.
- We need to be able to statically or dynamically add an unit to a value or a container.
----------------------------
- [ ] **III.** Expose a unit system configuration for GEOS input / output.
- Add a `<UnitSystem/>` node to `<Problem/>`
- Add a `preset` attribute to `UnitSystem`, which would allow to quickly use a unit system preset as a base. These preset would be documented in the user guide, and would be added by the different teams working on GEOS.
- Add `<UnitSystem/>` children node for each characteristics. 
As an exemple : `<Time inputFiles="years" stdout="days" outputFiles="s" />`
Or, as a more uniform exemple : `<Time units="d" />`

_Note: I tagged this issue as an epic, don't hesitate to tell me if it is not applicable._

rrsettgast:
@MelReyCG  Lets pick this back up.
As a reference, the way we did units in the old GEOS as to have them be strictly an input/output concept. 

- There was a `unit` block with the default system. 
- Any numerical values were allowed to contain a unit at the end of the input string. These would be separated with a space, allowing parsing of the units upon processing of the XML.
- Values with a unit other than the default system would be converted to the default units, values with no unit specified would be assumed to be in the default unit. We generally used `SI` as the default units.
- The internals of the code had no concept of units.
- outputs could be converted to a output unit system. 

[...]

PR #2954, opened 2024-01-25
Adding SourceFluxStatistics

### LINKED ISSUE #3011: Output `SourceFlux` statistics,
This issue is part of the #2277 EPIC.

MelReyCG:
The goal of this PR is to add a `SourceFluxStatistics` component which aggregate the producted mass & rate on a `SourceFlux` set.
- Log Level 1 outputs the sum of all `SourceFluxStatistics` produced rate & mass,
- Log Level 2 details values for each SourceFluxStatistics,
- Log Level 3 details values for each region.

Here is the typical log output for a multi-phase simulation, with a log level of 2 :
```
Rank 0: SourceFluxStatistics timeStepFluxStats (of sourceFlux, in all_regions): Producing on 1 elements
Rank 0: SourceFluxStatistics timeStepFluxStats (of sourceFlux, in all_regions): Produced mass = [-280.5, 0] kg
Rank 0: SourceFluxStatistics timeStepFluxStats (of sourceFlux, in all_regions): Production rate = [-0.561, 0] kg/s
Rank 0: SourceFluxStatistics timeStepFluxStats (of sinkFlux, in all_regions): Producing on 1 elements
Rank 0: SourceFluxStatistics timeStepFluxStats (of sinkFlux, in all_regions): Produced mass = [0, 31] kg
Rank 0: SourceFluxStatistics timeStepFluxStats (of sinkFlux, in all_regions): Production rate = [0, 0.062] kg/s
Rank 0: SourceFluxStatistics timeStepFluxStats (of flux_set, in all_regions): Producing on 2 elements
Rank 0: SourceFluxStatistics timeStepFluxStats (of flux_set, in all_regions): Produced mass = [-280.5, 31] kg
Rank 0: SourceFluxStatistics timeStepFluxStats (of flux_set, in all_regions): Production rate = [-0.561, 0.062] kg/s
```

Note: If the `PeriodicEvent` that execute the `SourceFluxStatistics` has a frequency such as multiple time-step are encountered, the statistics are aggregated (masses -> sum, rates -> mean).

I've also added a test which checks if the statistics output of the `SourceFluxStatistics` and `SinglePhaseStatistics::RegionStatistics::totalMass` are consistent with the provided production rates table.
The test checks if the statistics are right over the whole simulation, but also for each timesteps. It will crash if a timestep does not run. Thus, it tests if the `PeriodicEvent` functions as expected (is it the only test to do so ?).


paveltomin:
please double check that logic is consistent with useMass flag (default is 0)

paveltomin:
> please double check that logic is consistent with useMass flag (default is 0)

and time step chop

untereiner (approved):

MelReyCG:
@paveltomin

Timestep cuts were tested by the `checkMultiPhaseFluxStatistics` test (this simulation has 3 sub-timesteps). I added a guard (`testInputs.requireSubTimeStep`) so it will keep being tested with timesteps.

About `useMass`, the `checkMultiPhaseFluxStatistics` simulation does not converge anymore when I set it to 1. I am probably missing how to manage this component. Are the `SourceFlux` units in mole by default ? I'll investigate tomorrow.

Side question : do you think I should add CSV export for this component ?

paveltomin:
> About `useMass`, the `checkMultiPhaseFluxStatistics` simulation does not converge anymore when I set it to 1. I am probably missing how to manage this component. Are the `SourceFlux` units in mole by default ? I'll investigate tomorrow.

That's the tricky part, up to my understanding SourceFlux units depend on useMass flag. If useMass=0 then units are moles, if useMass=1 then units are mass.

> Side question : do you think I should add CSV export for this component ?

I found CSV output for flow and mechanics very convenient and use it all the time, also I think @jhuang2601 started using it. It would be nice to have it for SourceFlux too yes, but it can be done in a separate PR.

paveltomin:
> Timestep cuts were tested by the `checkMultiPhaseFluxStatistics` test (this simulation has 3 sub-timesteps). I added a guard (`testInputs.requireSubTimeStep`) so it will keep being tested with timesteps.

You test one new code with another new code :)

Your current implementation is tricky and only valid for constant flux value because if I understood correctly the statistics is being accumulated over the newton iterations while the real value that matters is the final one, when the iterations converged.
I don't see any zeroing-out of accumulated quantities when time step cut happens but I am probably blind. My suspicion for time step cut is similar - only works because the flux value is constant, but fundamental logic behind is slippery.

MelReyCG:
@paveltomin 

I've added the CSV export and try to make the code of `SourceFluxStatsAggregator::WrappedStats::gatherTimeStepStats()` clearer. Do you think this method is missing something about timestep cuts? Or do you think I should go for another way to keep track of data accumulation.

> You test one new code with another new code :)

I did not understand if you are thinking that is a good thing.
I made a typo it the comment you quoted -> What I did was to add a check in my unit test to ensure that sub-timesteps are being tested.

paveltomin:
> I've added the CSV export and try to make the code of `SourceFluxStatsAggregator::WrappedStats::gatherTimeStepStats()` clearer. Do you think this method is missing something about timestep cuts? Or do you think I should go for another way to keep track of data accumulation.

Thanks for CSV and if you say timestep cut is ok, I trust that, thanks for checking and refining the code

> > You test one new code with another new code :)
> 
> I did not understand if you are thinking that is a good thing. I made a typo it the comment you quoted -> What I did was to add a check in my unit test to ensure that sub-timesteps are being tested.

I just mean that one could potentially do the same mistake in the main code and in the test code, so unit test may not help catching it

Would you mind adding that output into one of the xmls?

MelReyCG:
@paveltomin 

> I just mean that one could potentially do the same mistake in the main code and in the test code, so unit test may not help catching it

Oh I see!
To explain my test, I prepared some expected data and a xml input and check that the simulation outputs statistics that are consistent with the expected ones, so the test is not dependent of the internal code. I also chek if the count of timestep is as expected. And now I check if at least one timestep cut occurs, in order to see if everything goes right in that case too.

> Would you mind adding that output into one of the xmls?

Which XMLs output are you talking about ?

paveltomin:
> > Would you mind adding that output into one of the xmls?
> 
> Which XMLs output are you talking about ?

I mean add this new SourceFluxStatistics output to one of the existing XMLs

MelReyCG:
Ok I will add one or more `SourceFluxStatistics` in the exemples.

`useMass` has indeed an influence on the `SourceFlux` units 👍 :
- I fixed the `SourceFluxStatistics` and added a (temporary) test to compare with mass and mole,
- I added descriptions of the wrapper about the fluxes units.
- I added a way of getting the *amount of fluid unit* on `FlowSolverBase`, and uniformized a bit with the other statistics classes.
- I verified the `SinglePhaseStatistics` units.

I don't know why, but the same test simulation in mass and mole are behaving differently (`MultiPhaseFluxStatisticsTestMol` has a timestep cut whereas `MultiPhaseFluxStatisticsTestMass` haven't any), here is the log :
[test2.txt](https://github.com/GEOS-DEV/GEOS/files/14361746/test2.txt)

paveltomin (approved):

paveltomin:
> I don't know why, but the same test simulation in mass and mole are behaving differently (`MultiPhaseFluxStatisticsTestMol` has a timestep cut whereas `MultiPhaseFluxStatisticsTestMass` haven't any), here is the log : [test2.txt](https://github.com/GEOS-DEV/GEOS/files/14361746/test2.txt)

this is normal, since the amount injected can be very different, solver can be more stressed

MelReyCG:
Ok I see. Thanks a lot for your inputs and review !

src/coreComponents/unitTests/fluidFlowTests/testFluidStatistics.cpp
```diff
@@ -0,0 +1,1199 @@
...
+                                discretization="fluidTPFA"
+                                targetRegions="{ reservoir }"
+                                temperature="366.483"
+                                useMass="1"
+                                logLevel="1" >
+      <NonlinearSolverParameters newtonTol="1.0e-6"
+                                 newtonMaxIter="8"
+                                 allowNonConverged="1" />
+      <LinearSolverParameters solverType="gmres"
+                              preconditionerType="amg"
```
paveltomin: should be mgr i guess

src/coreComponents/unitTests/fluidFlowTests/testFluidStatistics.cpp
```diff
@@ -0,0 +1,1197 @@
+/*
```
paveltomin: could you please rename it to test**Flow**Statistics.cpp?

src/coreComponents/unitTests/fluidFlowTests/testFluidStatistics.cpp
```diff
@@ -0,0 +1,1197 @@
...
+<Problem>
+
+  <Solvers>
+    <SinglePhaseFVM name="testSolver"
+                    discretization="singlePhaseTPFA"
+                    targetRegions="{ reservoir }" >
+
+      <NonlinearSolverParameters newtonTol="1.0e-6"
+                                 newtonMaxIter="40"
+                                 allowNonConverged="1" />
```
paveltomin: remove

src/coreComponents/unitTests/fluidFlowTests/testFluidStatistics.cpp
MelReyCG: It is indeed a better name

paveltomin (approved):

src/coreComponents/fieldSpecification/SourceFluxStatistics.hpp
dkachuma, line 1: This looks like a task that should belong to `physicsSolvers` and not quite a `fieldSpecification`.

src/coreComponents/fieldSpecification/SourceFluxStatistics.cpp
```diff
@@ -0,0 +1,364 @@
...
+
+SourceFluxStatsAggregator::SourceFluxStatsAggregator( const string & name,
+                                                      Group * const parent ):
+  Base( name, parent )
+{
+  getWrapper< integer >( Group::viewKeyStruct::logLevelString() ).
+    appendDescription( GEOS_FMT( "\n- Log Level 1 outputs the sum of all {0}(s) produced rate & mass,\n"
+                                 "- Log Level 2 details values for each {0},\n"
+                                 "- Log Level 3 details values for each region,\n"
+                                 "- Log Level 4 details values for each sub-region.",
```
dkachuma: I wouldn't go to the sub-region level. I would stop at the region level.

src/coreComponents/fieldSpecification/SourceFluxStatistics.cpp
```diff
@@ -0,0 +1,364 @@
...
+                                                              MeshLevel & mesh,
+                                                              arrayView1d< string const > const & )
+  {
+    registerWrappedStats( mesh, viewKeyStruct::fluxSetWrapperString(), viewKeyStruct::allRegionWrapperString() );
+
+    for( string const & fluxName : m_fluxNames )
+    {
+      registerWrappedStats( mesh, fluxName, viewKeyStruct::allRegionWrapperString() );
+
+      mesh.getElemManager().forElementRegions( [&]( ElementRegionBase & region )
```
dkachuma, line 109: If the `logLevel` is low enough, then we don't need to create these wrappers.

dkachuma (approved):

src/coreComponents/fieldSpecification/SourceFluxStatistics.cpp
```diff
@@ -0,0 +1,364 @@
...
+    outputFile.close();
+  }
+}
+
+bool SourceFluxStatsAggregator::execute( real64 const GEOS_UNUSED_PARAM( time_n ),
+                                         real64 const GEOS_UNUSED_PARAM( dt ),
+                                         integer const GEOS_UNUSED_PARAM( cycleNumber ),
+                                         integer const GEOS_UNUSED_PARAM( eventCounter ),
+                                         real64 const GEOS_UNUSED_PARAM( eventProgress ),
+                                         DomainPartition & domain )
```
TotoGaz, line 189: Hi, I'd be interested in decoupling the gathering/aggregation of the data from its output.
Here we impose `log` and `csv`, for example. There's no reason not to use the `CSV` for other types of data, and to impose `CSV` for the source flux data.

@MelReyCG , I had a quick chat with @arng40 about that.

src/coreComponents/fieldSpecification/SourceFluxStatistics.cpp
```diff
@@ -0,0 +1,364 @@
...
+    {
+      outputFile << GEOS_FMT( "{},{},{},{}",
+                              stats.getStatsPeriodStart(), stats.stats().m_elementCount,
+                              stats.stats().m_producedMass, stats.stats().m_productionRate ) << std::endl;
+    }
+    outputFile.close();
+  }
+}
+
+bool SourceFluxStatsAggregator::execute( real64 const GEOS_UNUSED_PARAM( time_n ),
```
TotoGaz, line 184: Hi, I'd be interested in decoupling the gathering/aggregation of the data from its output.
Here we impose `log` and `csv`, for example. There's no reason not to use the `CSV` for other types of data, and to impose `CSV` for the source flux data.

@MelReyCG , I had a quick chat with @arng40 about that.

TotoGaz (commented):
That looks like a nice feature.
A question about the display though.

src/coreComponents/fieldSpecification/SourceFluxStatistics.cpp
MelReyCG: Just to notify everyone about the conclusion of our chat on this subject:
- There are currently a lot of these log / CSV output in the statistics classes,
- Statistics output must be unified, and not strictly bound to CSV,
- This output refactoring is not the subject of this PR, but **is a task that I have got in my roadmap**.
- An `Outputs` new node could be created to output those stats, maybe something like:
```XML
<Outputs>
  <StatsOutput
    name="mySourceFluxesOutput"
    source="mySourceFluxesStats"
    outputFile="sourceFluxStats.csv" /> <!-- could be json or so -->
</Outputs>
```

src/coreComponents/fieldSpecification/SourceFluxStatistics.cpp
MelReyCG: I use them to collect stats from the `SinglePhaseBase::applySourceFluxBC()` methods that have no knowledge of the regions. Can I keep them to collect the data or should I proceed differently?

src/coreComponents/fieldSpecification/SourceFluxStatistics.hpp
MelReyCG: So the solvers would expose the producted rates & masses for each `SourceFlux`, and then the `SourceFluxStatistics` would access to it?

src/coreComponents/common/Units.hpp
```diff
@@ -122,6 +138,14 @@ constexpr inline std::string_view getSymbol( Unit unit )
...
     case Distance:        return "m";
     case Time:            return "s";
+    case Viscosity:       return "Pa*s";
+    case Enthalpy:        return "J/kg";
+    case Density:         return "kg/m3";
+    case Solubility:      return "g/L";
+    case Mass:            return "kg";
+    case Mole:            return "mol";
+    case MassRate:        return "kg/s";
+    case MoleRate:        return "mol/s";
```
CusiniM, line 148: Units are not really strongly enforced in GEOS. One could use any set of units as long as they are consistent. For example a user can decide to specify pressures in MPa an adjust all other properties. I would say that we either properly introduce a unit parser or these units can easily be false.

MelReyCG: About this header, it is in S.I. by default (following the GEOS convention). Some units can be in other systems as you can see for the temperatures (°C rather than °K), and if someone want to add more, I think it should come with conversion functions.
I strongly advise to add comments on all non-S.I. variables / method, it would be less error-prone and would help when implementing a parsing unit system or when adding new outputs.

About adding unit parsing system, it is on my roadmap. Feel free to add suggestions on [my EPIC](https://github.com/GEOS-DEV/GEOS/issues/2704).

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseStatistics.cpp
```diff
@@ -305,7 +304,7 @@ void CompositionalMultiphaseStatistics::computeRegionStatistics( real64 const ti
                                         subRegionImmobilePhaseMass.toView(),
                                         subRegionComponentMass.toView() );
 
-    ElementRegionBase & region = elemManager.getRegion( subRegion.getParent().getParent().getName() );
+    ElementRegionBase & region = elemManager.getRegion( ElementRegionBase::getParentRegion( subRegion ).getName() );
```
TotoGaz, line 307: That's better!

src/coreComponents/unitTests/testingUtilities/TestingTasks.hpp
```diff
@@ -0,0 +1,80 @@
...
+{
+
+
+/**
+ * @brief This Task allows to do checks each timestep during the simulation by calling a provided functor.
+ * To be executed, it must be - as every task - declared in the provided xmlInput within the Tasks node and called by an even.
+ * As this Group is designed for developpers and not user oriented, it should not appear in the documentation nor in the xsd.
+ * Question: could this task be used in the integratedTests ?
+ */
+class TimeStepChecker : public TaskBase
```
TotoGaz, line 37: @CusiniM What do you think of this. I believe that's interesting.

src/coreComponents/physicsSolvers/SolverBase.hpp
```diff
@@ -734,6 +734,7 @@ class SolverBase : public ExecutableGroup
   virtual bool registerCallback( void * func, const std::type_info & funcType ) final override;
 
   SolverStatistics & getSolverStatistics() { return m_solverStatistics; }
+  SolverStatistics const & getSolverStatistics() const { return m_solverStatistics; }
```
TotoGaz, line 737: Who's using the non-const version?
If you can, can you remove it or use a setter instead?

src/coreComponents/fieldSpecification/SourceFluxStatistics.cpp
```diff
@@ -0,0 +1,361 @@
...
+                              "Use \"all\" to target all {0}.",
+                              SourceFluxBoundaryCondition::catalogName() ) );
+}
+
+void SourceFluxStatsAggregator::postProcessInput()
+{
+  Base::postProcessInput();
+
+  FieldSpecificationManager & fsManager = FieldSpecificationManager::getInstance();
+  if( m_fluxNames.size() == 1 && m_fluxNames[0] == "all" )
```
TotoGaz:
```suggestion
  if( m_fluxNames.size() == 1 && m_fluxNames[0] == "*" )
```
Even if we're not doing the `fnmatch` in this PR, would using `*` make sense?

src/coreComponents/fieldSpecification/SourceFluxStatistics.cpp
```diff
@@ -0,0 +1,361 @@
...
+    {
+      fluxStats.stats() = StatData();
+
+      forAllRegionStatsWrappers( meshLevel, fluxStats.getFluxName(),
+                                 [&] ( ElementRegionBase & region, WrappedStats & regionStats )
+      {
+        regionStats.stats() = StatData();
+
+        forAllSubRegionStatsWrappers( region, regionStats.getFluxName(),
+                                      [&] ( ElementSubRegionBase &, WrappedStats & subRegionStats )
```
TotoGaz, line 207: 4 nested lambda functions, I think that you tie the record 🤣 😉

src/coreComponents/fieldSpecification/SourceFluxStatistics.cpp
```diff
@@ -0,0 +1,361 @@
...
+  real64 const timeDivisor = m_statsPeriodDT > 0.0 ? 1.0 / m_statsPeriodDT : 0.0;
+  for( int ip = 0; ip < m_periodStats.getPhaseCount(); ++ip )
+  {
+    real64 periodMass = m_periodStats.m_timeStepMass[ip] + m_periodStats.m_periodPendingMass[ip];
+    m_stats.m_producedMass[ip] = periodMass;
+    m_stats.m_productionRate[ip] = periodMass * timeDivisor;
+  }
+
+  // combine period results from all MPI ranks
+  m_stats.mpiReduce();
```
TotoGaz, line 329: You need to be sure that this is called by all the ranks. Is it the case?

src/coreComponents/fieldSpecification/SourceFluxStatistics.hpp
```diff
@@ -0,0 +1,420 @@
...
+     * @param elementCount number of cell elements concerned by this instance
+     */
+    void gatherTimeStepStats( real64 currentTime, real64 dt,
+                              arrayView1d< real64 const > const & producedMass,
+                              integer elementCount );
+
+    /**
+     * @brief Finalize the period statistics of each timestep gathering and render data over all mpi ranks.
+     * The result can be read by the data() assessor.
+     */
```
TotoGaz: Be a little more explicit about the `MPI` call that needs to be done by everyone. Maybe add a `@note` ?

src/coreComponents/fieldSpecification/SourceFluxStatistics.hpp
```diff
@@ -0,0 +1,420 @@
...
+    /// @return The key for statistics wrapper name that targets all fluxes of the set
+    constexpr inline static string_view fluxSetWrapperString() { return "flux_set"; }
+  };
+
+protected:
+
+  /**
+   * @copydoc Group::postProcessInput()
+   */
+  void postProcessInput() override;
```
TotoGaz: Can you move this `protected` function between the `public` and the `private`?
+ Are you sure that you need it to be `protected`?

src/coreComponents/fieldSpecification/SourceFluxStatistics.hpp
```diff
@@ -0,0 +1,420 @@
...
+ * @file SourceFluxStatistics.hpp
+ */
+
+#ifndef SRC_CORECOMPONENTS_PHYSICSSOLVERS_FLUIDFLOW_SOURCEFLUXSTATISTICS_HPP_
+#define SRC_CORECOMPONENTS_PHYSICSSOLVERS_FLUIDFLOW_SOURCEFLUXSTATISTICS_HPP_
+
+#include "physicsSolvers/FieldStatisticsBase.hpp"
+#include "physicsSolvers/fluidFlow/FlowSolverBase.hpp"
+
+class SourceFluxBoundaryCondition;
```
TotoGaz: Do you need that?

TotoGaz (approved):

src/coreComponents/fieldSpecification/SourceFluxStatistics.cpp
MelReyCG: I agree, these `"all"` proposals came to my mind as it is possible to use that as `setNames` in field specifications.

src/coreComponents/physicsSolvers/SolverBase.hpp
MelReyCG: It is used at `src/coreComponents/physicsSolvers/multiphysics/CoupledSolver.hpp:431, 491, 506`, I think the non-const version is useful, should I remove the const one?

src/coreComponents/fieldSpecification/SourceFluxStatistics.cpp
MelReyCG: Indeed, I tested that all ranks send & combine how much the `SourceFlux` produced / injected (possibly 0).

src/coreComponents/fieldSpecification/SourceFluxStatistics.hpp
MelReyCG: I set the class as `final` and removed the protected qualificator.

src/coreComponents/fieldSpecification/SourceFluxStatistics.hpp
MelReyCG: Oops, I don't need this forward declaration anymore.

MelReyCG:
I am currently working on the parallel implementation, so please wait before merging this PR.

paveltomin:
> I am currently working on the parallel implementation, so please wait before merging this PR.

I moved it into "Approved" to avoid confusion

MelReyCG:
@paveltomin Sorry, I posted on the wrong PR, I was talking about the transmissibility one, this one works fine!

PR #2957, opened 2024-01-26
feat: Add ExternalDataRepository

### LINKED ISSUE #2953: Handle input data from a complex source
**Describe the issue**
For now, one input mesh is read from from one vtu file.
Their are also surfaces embedded in this vtu file using the potpourri ability of this file format.

I am considering more complex files format where data is (or can be) organized semantically. 
There is of course RESQML but we can also consider VTK composite meshes such as [vtkPartitionedDataSetCollection](https://vtk.org/doc/nightly/html/classvtkPartitionedDataSetCollection.html)

It is not fairly possible to add a new node in `Problem`. It would require a new manager and to find a place in the order of the managers.

**Proposed cleanup**
I suggest for now to add a new abstract level above `MeshGeneratorBase` that would allow a new subtree of classes to handle such cases
I am already using it with the branch feature/untereiner/wip. I will propose a cleaned version

**Additional context**

untereiner:
May close #2953 

More explanations:

There below is the situation today:
- Every child of MeshGeneratorBase has to fulfill its API even if not compatible
- ExternalMeshGeneratorBase has to have a file as input

```mermaid
classDiagram
namespace GEOS {

  class Group 

 class WellGeneratorABC

 class WellGeneratorBase
  
  class MeshGeneratorBase {
    generateMesh( Group & parent, SpatialPartition & partition ): void 
    importFieldOnArray( Block block, ....) : void virtual
    freeResources(): virtual void 
    getVolumicFieldsMapping(): std::map< string, string > const &
    getSurfacicFieldsMapping(): std::map< string, string > const & 
  }

  class ExternalMeshGeneratorBase {
    m_filePath: Path 
    m_translate: R1Tensor
    m_scale: R1Tensor
    m_volumicFieldsToImport: array1d< string >
    m_volumicFieldsInGEOS: array1d< string >
    m_surfacicFieldsToImport: array1d< string > 
    m_surfacicFieldsInGEOS: array1d< string >
  }

  class InternalMeshGenerator 
}

Group <|-- MeshGeneratorBase
MeshGeneratorBase <|-- ExternalMeshGeneratorBase
MeshGeneratorBase <|-- InternalMeshGenerator
Group <|-- WellGeneratorABC 
WellGeneratorABC <|-- WellGeneratorBase
```

What I expect after this PR:
- A new class (`ExternalDataRepository`) with almost no constraints on what it expects allows me to create a proper API for another module (example: [RESQML](https://github.com/geosiris-technologies/geos-resqml/blob/main/src/EnergyMLDataObjectRepository.hpp))
- A new class (`VTKHierarchicalDataRepository`) to add the capability to load complex meshes with semantics ([example](inputFiles/singlePhaseFlow/FieldCaseTutorial3_composite_smoke.xml) provided on singlePhaseFlow)
- A new class (`Region`) to add the capability to load sets of cells (vtkDataSets) from a complex mesh data format (VTK or RESQML) and give it a id for internal purpose (aka regionAttribute for vtu files)

```mermaid
classDiagram
direction LR
namespace GEOS {

class Group

class ExternalDataRepository {
  virtual open(); void
}

class MeshGeneratorBase {
  generateMesh( Group & parent, SpatialPartition & partition ): void 
  importFieldOnArray( Block block, ....) : void virtual
  freeResources(): virtual void 
  getVolumicFieldsMapping(): std::map< string, string > const &
  getSurfacicFieldsMapping(): std::map< string, string > const & 
}

class ExternalMeshGeneratorBase {  
  m_translate: R1Tensor
  m_scale: R1Tensor
  m_volumicFieldsToImport: array1d< string >
  m_volumicFieldsInGEOS: array1d< string >
  m_surfacicFieldsToImport: array1d< string > 
  m_surfacicFieldsInGEOS: array1d< string >
}

class VTKMeshGenerator {
  m_filePath: Path 
}

class InternalMeshGenerator 

class VTKHierarchicalDataRepository

class MeshComponentBase

class WellGeneratorBase

class Region

}

Group <|-- ExternalDataRepository
Group <|-- MeshGeneratorBase
MeshGeneratorBase <|-- ExternalMeshGeneratorBase
MeshGeneratorBase <|-- InternalMeshGenerator
ExternalMeshGeneratorBase <|-- VTKMeshGenerator
ExternalDataRepository <|-- VTKHierarchicalDataRepository
Group <|-- MeshComponentBase
MeshComponentBase <|-- WellGeneratorBase
MeshComponentBase <|-- Region


namespace GEOS-RESQML {

class EnergyMLDataObjectRepository {
common::DataObjectRepository * getData()
COMMON_NS::AbstractObject* getDataObject(string const & id)
COMMON_NS::AbstractObject* getDataObjectByTitle(string const & name)

}


class EpcDocumentRepository

class RESQMLMesh 
}

ExternalDataRepository <|-- EnergyMLDataObjectRepository
EnergyMLDataObjectRepository <|-- EpcDocumentRepository
MeshGeneratorBase <|-- RESQMLMesh

```


paveltomin:
hi @untereiner 
i am not sure how adding `MeshBase` layer helps here, most probably i am missing soemthing
could you please add some description explaining that?
thank you

untereiner:
Hi @paveltomin,
I can link to the [RESQML plugin](https://github.com/geosiris-technologies/geos-resqml). Adding this layer helps me creating a child node in _Mesh_ by subclassing something more abstract that _MeshGeneratoreBase_

The entry point of the RESQML C++ library is a repository of 3D objects that I want to put in the deck.

I can also link an example deck if needed.

Also, VTK has file formats storing composite data (grids and surfaces) with complex relationships ( see [here](https://github.com/geosiris-technologies/geos-resqml)). I claim that these files should be considered as « data repositories » to extract from them what’s needed where it’s needed. So, this MeshBase layer could also benefit to them

paveltomin:
> Hi @paveltomin, I can link to the [RESQML plugin](https://github.com/geosiris-technologies/geos-resqml). Adding this layer helps me creating a child node in _Mesh_ by subclassing something more abstract that _MeshGeneratoreBase_
> 
> The entry point of the RESQML C++ library is a repository of 3D objects that I want to put in the deck.
> 
> I can also link an example deck if needed.
> 
> Also, VTK has file formats storing composite data (grids and surfaces) with complex relationships ( see [here](https://github.com/geosiris-technologies/geos-resqml)). I claim that these files should be considered as « data repositories » to extract from them what’s needed where it’s needed. So, this MeshBase layer could also benefit to them

thanks for the explanations!

paveltomin (approved):

src/coreComponents/mesh/generators/MeshGeneratorBase.hpp
```diff
@@ -41,7 +38,7 @@ class CellBlockManager;
  *  @brief The MeshGeneratorBase class provides an abstract base class implementation for different mesh types.
  *	   The MeshGeneratorBase is the Group specialization for different type of mesh handling.
  */
-class MeshGeneratorBase : public dataRepository::Group
+class MeshGeneratorBase : public MeshBase
```
TotoGaz: Doing so, you make the (yet to become) `generators` module depend on `mesh` while we have been willing to de the exact opposite: `generators` would become a "somehow" independent "source" of meshing informations.

src/coreComponents/mesh/MeshManager.cpp
```diff
@@ -43,15 +43,15 @@ MeshManager::~MeshManager()
 Group * MeshManager::createChild( string const & childKey, string const & childName )
 {
   GEOS_LOG_RANK_0( "Adding Mesh: " << childKey << ", " << childName );
-  std::unique_ptr< MeshGeneratorBase > solver = MeshGeneratorBase::CatalogInterface::factory( childKey, childName, this );
-  return &this->registerGroup< MeshGeneratorBase >( childName, std::move( solver ) );
+  std::unique_ptr< MeshBase > mesh = MeshBase::CatalogInterface::factory( childKey, childName, this );
```
TotoGaz: Now `MeshManager` depends on `MeshBase` and `MeshGeneratorBase`.
Couldn't we have one single concept and use it? I'd say that would make the things clearer.

TotoGaz (commented):
Given the 2 PRs (https://github.com/GEOS-DEV/GEOS/pull/3068 and https://github.com/GEOS-DEV/GEOS/pull/3063) that are deeply revamping this area of the code, this is unfortunately almost granted that this work (and all the work attached relying on it) will become obsolete quite soon.

src/coreComponents/mesh/generators/MeshGeneratorBase.hpp
untereiner: I can rename it _GeneratorBase_. But that’s not the path you are drawing. I would be happy to revamp my work in a sibling modèle of the generator one called « data repository »

MelReyCG:
@untereiner does it modify [this graph in the docs](https://geosx-geosx.readthedocs-hosted.com/en/latest/coreComponents/mesh/docs/meshDeveloperGuide.html) ?

untereiner:
@MelReyCG it adds a level in the class hierarchy, so yes. The PRs of Thomas are modifying these files. I don’t know how to move this forward.

untereiner:
Hi @rrsettgast, I put diagrams to explain a bit more what this PR is expected to improve

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.hpp
```diff
@@ -210,7 +210,7 @@ class VTKPolyDataWriterInterface
   void clearData();
 
 
-private:
+protected:
```
rrsettgast, line 222: Is the increase of visibility required? I am not understanding from the review what child calls these?

src/coreComponents/mesh/MeshBase.hpp
```diff
@@ -0,0 +1,78 @@
...
+
+namespace geos
+{
+
+/**
+ *  @class MeshBase
+ *  @brief The MeshBase class provides an abstract base class implementation for different mesh types.
+ *	   The MeshBase is the Group specialization for different type of mesh handling.
+ */
+class MeshBase : public dataRepository::Group
```
rrsettgast: So `MeshBase` is currently just to hold the catalog definition/instantiation? Is there plans for specifying a pure virtual interface here?

src/coreComponents/mesh/MeshManager.cpp
rrsettgast: Is there an answer to this question? To me, if `MeshBase` and `MeshGeneratorBase` have different responsibilities and `MeshGeneratorBase` is intended to generate something derived from `MeshBase`, then I don't see a problem....but perhaps I don't understand.

rrsettgast (commented):
Hi @untereiner .
Would it be better to rename the proposed `MeshBase` to `MeshGeneratorBase`, and rename the current `MeshGeneratorBase` to something more specific to its use case?

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.hpp
untereiner: You are right. This is needed to subclass PolyDataWriterInterface to export data into a VTK compatible one. Here [RESQMLWriterInterface](https://github.com/geosiris-technologies/geos-resqml/blob/main/src/RESQMLWriterInterface.hpp) inherits of PolyDataWriterInterface. It's a way to avoid recoding transformation from GEOS internal structures to RESQML data structures since I have function for VTK to RESQML

src/coreComponents/mesh/MeshBase.hpp
untereiner: For now no. It is to avoid implementing the API of MeshGeneratorBase

untereiner:
> Hi @untereiner . Would it be better to rename the proposed `MeshBase` to `MeshGeneratorBase`, and rename the current `MeshGeneratorBase` to something more specific to its use case?

I am open to suggestions ! 
I don’t have any good idea since MeshGeneratorBase is really ment to be a parent of a specific family of generators.

untereiner:
@rrsettgast I completely changed the design. This one is closer and cleaner to what the architecture really is. Think of `EpcDocumentRepository` as a database with an API you query to get what's inside (meshes, properties, geometric objects, etc.)  
Here is an excerpt of a simulation deck:

```xml
<Problem xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">

  <Solvers>
   ....
  </Solvers>

  <ExternalDataRepository>
    <EpcDocumentRepository
      logLevel="2"
      name="plop"
      files="{ myfile.epc }"/>
  </ExternalDataRepository> 
  

  <Mesh>
    
    <RESQMLMesh
      logLevel="2"
      name="mesh"
      repositoryName="plop"
      uuid="5cff7c6a-b5d7-441f-b981-bc77e91a144d">

      <Region
        name="Region 1"
        uuid="0b807d31-540a-45e7-a898-b75b05e358a4"        
      />
      <Region
        name="Region 2"
        uuid="1facf29e-25ec-4c6c-b404-595dc84eb29c"
      />
      <Region
        name="Region 3"
        uuid="4b3ce11e-6b9a-4497-b430-d06226a0491c"
      />
      <Region
        name="Region 4"
        uuid="31e184e2-7894-4408-bb57-d0ec10f1dcbf"
      />
      <Region
        name="Region 5"
        uuid="4efccb61-13f4-4583-8f3f-8303be28ea88"
      />

      <Property
        name="rockPermeability_permeability"
        uuid="c2f3faae-4a88-4906-95ff-01e2bbc95c5f"
        title="PERM"
      />
      <Property
        name="rockPorosity_referencePorosity"
        uuid="f678acb4-eecf-4dd9-8a02-2cd7a64179eb"
        title="PHIT"
      />
      <Property
        name="netToGross"
        uuid="c27a314a-2ead-4ed6-8e79-4087cf1557bd"
        title="NTG"
      />

      <Surface uuid="" title=""/>
      <Surface uuid="" title=""/>

    </RESQMLMesh>

  </Mesh>
   ...
</Problem>
```

I think this kind queryable external data repository will also be usefull with `vtkParitionedDataSetCollections` because you have to use a `DataAssembly` to query the hierarchical organization of partitioned-datasets.

rrsettgast:
@untereiner Can we discuss this at the meeting tomorrow?

untereiner:
@rrsettgast I am not available at the meeting time. Can we meet at another time or next week ?

untereiner:
Here is an insight for [VTK complex datasets](https://discourse.paraview.org/t/transition-from-multiblock-to-partitioned-datasets-in-vtk-and-paraview-a-complete-guide/12061)

```xml
<Problem xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">

  <Solvers>
   ....
  </Solvers>

  <ExternalDataRepository>
    <VTKHierarchicalDataRepository
      logLevel="2"
      name="plop"
      files="{ myfile.vtpc }"/>
  </ExternalDataRepository> 
  

  <Mesh>
    
    <VTKMesh
      logLevel="2"
      name="mesh"
      repositoryName="plop"
      path="/Root/grid">
     
     <InternalWell
        logLevel="1"	
        name="wellInjector"
        wellRegionName="well"
        wellControlsName="wellControls"
        repositoryName="plop"
        path="/Root/wells/wellInjector">
         
        <Perforation
           name="wellInjector_perf0"
           distanceFromHead="5"/>

     </InternalWell>

    </VTKMesh>

  </Mesh>
...
</Problem>
```

rrsettgast:
> @rrsettgast I am not available at the meeting time. Can we meet at another time or next week ?

Hello @untereiner. Next week is fine. Thanks

src/coreComponents/mesh/ExternalDataRepositoryBase.hpp
```diff
@@ -0,0 +1,77 @@
...
+
+  /**
+   * @brief Accessor for the singleton Catalog object
+   * @return a static reference to the Catalog object
+   */
+  static CatalogInterface::CatalogType & getCatalog();
+
+  /**
+   * @brief
+   *
```
MelReyCG: I would write the documentation at the abstraction level and use `@copydoc` in the child class(es).

src/coreComponents/mesh/generators/VTKUtilities.cpp
```diff
@@ -579,6 +579,25 @@ AllMeshes loadAllMeshes( Path const & filePath,
   return AllMeshes( main, faces );
 }
 
+//TODO handle more general cases: multiple partitions
+AllMeshes loadPartitions( vtkSmartPointer< vtkPartitionedDataSet > & partitions )
```
MelReyCG: If I understand correctly, `loadPartitions()` doesn't work if multiple partitions are loaded on a rank. Is it possible to detect that and `GEOS_ERROR()` if it is the case?

src/coreComponents/mesh/generators/VTKMeshGenerator.cpp
```diff
@@ -75,6 +80,40 @@ VTKMeshGenerator::VTKMeshGenerator( string const & name,
...
+
+  registerWrapper( viewKeyStruct::meshPathString(), &m_meshPath ).
+    setInputFlag( InputFlags::OPTIONAL ).
+    setDescription( "path of the VTK mesh in the repository" );
+}
+
+void VTKMeshGenerator::postInputInitialization()
+{
+  GEOS_ERROR_IF( !this->m_filePath.empty() && (!m_repositoryName.empty() || !m_meshPath.empty()),
+                 getName() << ": Access to the mesh via file or repository are mutually exclusive. "
```
MelReyCG: I would use `getDataContext()` here and in the 2 other `GEOS_ERROR_IF` as they are errors related to the input source file.

+ what happens if I don't fill any of the three attributes?

src/coreComponents/mesh/generators/VTKMeshGenerator.cpp
```diff
@@ -75,6 +80,40 @@ VTKMeshGenerator::VTKMeshGenerator( string const & name,
...
+  GEOS_ERROR_IF ( !m_repositoryName.empty() && m_meshPath.empty(),
+                  getName() << ": using the repository mode, you must supply a path in the repository for the mesh." );
+
+  GEOS_ERROR_IF ( m_repositoryName.empty() && !m_meshPath.empty(),
+                  getName() << ": using a path in the repository for the mesh, you must supply a repository mode." );
+
+  if( !m_repositoryName.empty())
+  {
+    ExternalDataRepositoryManager & externalDataManager = this->getGroupByPath< ExternalDataRepositoryManager >( "/Problem/ExternalDataRepository" );
+    // objectRepository.
```
MelReyCG: ?

MelReyCG (commented):
a few remarks (review to be continued)

untereiner:
ExternalDataRepository is too close to the data repository name used internally by GEOS:
Here are some suggestions
Choose a comment emoji to cast your vote

- :heart:   ExternalDataSource
- :rocket: :  ExternalDataStream
- :tada: :  ExternalDataInput

rrsettgast (approved):

untereiner:
@rrsettgast I will update the branch and change for `ExternalDataSource` then I should be ok for merging

wrtobin (changes requested):
It seems like we still need to rename `ExternalDataRepository` > `ExternalDataSource`

src/coreComponents/mesh/generators/VTKMeshGenerator.cpp
```diff
@@ -86,10 +120,68 @@ void VTKMeshGenerator::fillCellBlockManager( CellBlockManager & cellBlockManager
   vtkSmartPointer< vtkMultiProcessController > controller = vtk::getController();
   vtkMultiProcessController::SetGlobalController( controller );
 
-  GEOS_LOG_RANK_0( GEOS_FMT( "{} '{}': reading mesh from {}", catalogName(), getName(), m_filePath ) );
+  // GEOS_LOG_RANK_0( GEOS_FMT( "{} '{}': reading mesh from {}", catalogName(), getName(), m_filePath ) );
```
wrtobin:
```suggestion
```

wrtobin:
Nothing jumps out at me in the implementation that needs changes, really just the naming.

src/CMakeLists.txt
```diff
@@ -80,7 +80,7 @@ add_subdirectory( thirdparty )
 ################################
 # Add components
 ################################
-unset( externalComponentsLinkList )
+# unset( externalComponentsLinkList )
```
paveltomin: @rrsettgast is this ok?

src/docs/doxygen/GeosxConfig.hpp
```diff
@@ -135,7 +135,7 @@
 /* #undef chai_VERSION */
 
 /// Version information for adiak
-#define adiak_VERSION ..
+/* #undef adiak_VERSION */
```
paveltomin: @rrsettgast and this, ok?

paveltomin:
> It seems like we still need to rename `ExternalDataRepository` > `ExternalDataSource`

i renamed it, could you please have another look?

paveltomin:
@wrtobin could you please have another look?

wrtobin (approved):

src/coreComponents/mesh/generators/Region.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+  registerWrapper( viewKeyStruct::pathInRepositoryString(), &m_pathInRepository ).
+    setInputFlag( InputFlags::REQUIRED ).
+    setDescription( "Path of the dataset in the repository" );
+}
+
+Region::~Region()
+{}
+
+// Region::CatalogInterface::CatalogType & Region::getCatalog()
```
paveltomin: remove

src/coreComponents/mesh/generators/Region.hpp
```diff
@@ -0,0 +1,76 @@
...
+namespace geos
+{
+
+/**
+ * @brief Region parameters with Group capabilities
+ *
+ * This class has dataRepository::Group capabilities to allow for XML input.
+ *
+ */
+class Region : public MeshComponentBase
```
paveltomin, line 35: @rrsettgast @CusiniM is ok to call it Region?

src/docs/doxygen/GeosxConfig.hpp
rrsettgast: that is fine....but can you just switch it back?

src/coreComponents/mesh/generators/Region.hpp
rrsettgast: Region seems like it is too generic? How does this "Region" differ from the other "Region"?

src/CMakeLists.txt
rrsettgast: It is fine...but why is this needed?

paveltomin: reverted

src/docs/doxygen/GeosxConfig.hpp
paveltomin: reverted

PR #2960, opened 2024-01-27
Initialization of fields in new fracture cells in HydroFracture solver 

frankfeifan:
In this PR, we 

- [x] Move the code that initializes fields in new fracture cells from `TwoPointsFluxApproximation` to `HydroFractureSolver`

https://github.com/GEOS-DEV/integratedTests/pull/107


src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp
```diff
@@ -965,6 +973,245 @@ void HydrofractureSolver< POROMECHANICS_SOLVER >::setUpDflux_dApertureMatrix( Do
   } );
 }
 
+template< typename POROMECHANICS_SOLVER >
+void HydrofractureSolver< POROMECHANICS_SOLVER >::initializeNewFractureFields( DomainPartition & domain )
+{
+#if !defined(SET_CREATION_DISPLACEMENT)
```
CusiniM: can we take advantage of this PR to remove all these macros? We can have two different initialization strategies for now if we are still testing which one works best but let's guard it with  proper flag. 
Also, I don't think we saw that much benefit from initializing the displacement.

src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp
```diff
@@ -965,6 +973,245 @@ void HydrofractureSolver< POROMECHANICS_SOLVER >::setUpDflux_dApertureMatrix( Do
...
+  #endif
+
+  #if ALLOW_CREATION_MASS==0
+    arrayView1d< real64 > const dens = fractureSubRegion.getReference< array1d< real64 > >( "density_n" );
+  #endif
+
+  #if SET_CREATION_PRESSURE==1
+    arrayView1d< real64 > const fluidPressure_n = fractureSubRegion.getField< fields::flow::pressure_n >();
+    arrayView1d< real64 > const fluidPressure = fractureSubRegion.getField< fields::flow::pressure >();
+    // Set the new face elements to some unphysical numbers to make sure they get set by the following routines.
```
CusiniM: feels like we can do something better than this.

CusiniM (approved):
We can clean this up even further. For example, I don't think we need any of those macros. For now, can we just merge this so that it is at least clear where the initialization occurs?

src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp
```diff
@@ -965,6 +983,182 @@ void HydrofractureSolver< POROMECHANICS_SOLVER >::setUpDflux_dApertureMatrix( Do
...
+          }
+          if( zeroDisp )
+          {
+            aperture[newElemIndex] = 0;
+          }
+        }
+
+        if( this->getLogLevel() > 1 )
+        {
+          printf( "New elem index, init aper, init press = %4d, %4.2e, %4.2e \n",
```
paveltomin: GEOS_LOG?

paveltomin (approved):
looks good, definitely better than before and change in the right direction
thanks!

src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp
paveltomin: Fully support to remove the macroses, they are leading to dead code which doesn't even compile

paveltomin:
When are you planning to merge this PR?

src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp
CusiniM: yes, absolutely.

CusiniM:
> When are you planning to merge this PR?

as soon as possible

src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp
CusiniM: honestly we should never output stuff inside a kernel but this kernel has to be on host anyways so I think it's fine.

PR #2964, opened 2024-01-29
Out of place ATS

cssherman:
This PR moves ats files from the integratedTests repo into the main GEOS repo.  The new test baselines will have a 1:1 structure to the inputFiles directory.  We are planning to address changes to baseline storage in a follow-up PR.

geosPythonPackages PR:
https://github.com/GEOS-DEV/geosPythonPackages/pull/6

integratedTests PR:
https://github.com/GEOS-DEV/integratedTests/pull/88


host-configs/LLNL/quartz-base.cmake
```diff
@@ -58,5 +58,12 @@ set(MKL_LIBRARIES ${MKL_ROOT}/lib/intel64/libmkl_intel_lp64.so
 
 # ATS
 set(ATS_ARGUMENTS "--machine slurm36"  CACHE STRING "")
+set(USER $ENV{USER} CACHE STRING "")
+set(ATS_WORKING_DIR "/p/lustre2/${USER}/integratedTests/${CONFIG_NAME}"  CACHE PATH "")
+set(ATS_BASELINE_DIR "/p/lustre2/${USER}/integratedTests/baselines"  CACHE PATH "")
```
cssherman: These are setting where we want to run the tests and where we want to look for baselines.  Currently, there are no default values set, which triggers a cmake error later on.

host-configs/LLNL/quartz-base.cmake
```diff
@@ -58,5 +58,12 @@ set(MKL_LIBRARIES ${MKL_ROOT}/lib/intel64/libmkl_intel_lp64.so
...
+
+# Temporary argument for python module change testing
+set(GEOS_PYTHON_PACKAGES_BRANCH "feature/sherman/outOfPlaceATS" CACHE STRING "" FORCE)
```
cssherman: This is a cmake variable that can override the branch of geosPythonPackages during the build.  We'll want to have a procedure in place for handling this when we make potentially breaking changes to the CI

CusiniM (commented):
Thanks Chris! I think having the ats files in the same folder as the xml without those soft links is great. A few things we could work on:

- Naming of the ats files (it's all over the place)
- Enforce smoke and benchmark naming for all xmls so that we can then easily create a filter that only runs `smoke` tests and one that only runs `benchmarks`. 
- consolidate to a single ats file per folder? 
- add a field `file_name` to `TestDeck`

host-configs/LLNL/quartz-base.cmake
CusiniM: makes sense. We'll need that cmake target to reinstall the packages though.

host-configs/LLNL/quartz-base.cmake
CusiniM: This seems like a reasonable default for LC systems. Otherwise we could simply define a the geos build 
directory as a default if that fixes the cmake errors.

inputFiles/compositionalMultiphaseFlow/benchmarks/SPE10/SPE10.ats
```diff
@@ -0,0 +1,19 @@
+import geos_ats
```
TotoGaz, line 1: So are you adding new tests? Or are you basically "removing" the links with the file they are pointing to?

inputFiles/efemFractureMechanics/embeddedFractures_mechanics.ats
```diff
@@ -0,0 +1,53 @@
...
+        restart_step=0,
+        check_step=1,
+        curvecheck_params=CurveCheckParameters(**curvecheck_params)),
+    TestDeck(name="SneddonRotated_smoke",
+             description='Sneddon with inclined fracture',
+             partitions=((1, 1, 1), (2, 2, 1)),
+             restart_step=0,
+             check_step=1,
+             restartcheck_params=RestartcheckParameters(**restartcheck_params))
+]
```
TotoGaz: Maybe we can use `yapf` on `ats` files as well?

inputFiles/main.ats
```diff
@@ -0,0 +1,21 @@
...
+glue(outputs='')
+glue(constitutive_models='')
+
+# Recursively search the child directories and add them to ATS
+root_path = Path(ats.tests.AtsTest.getOptions().get("atsRootDir"))
+Nr = len(root_path.parts)
+for path in root_path.rglob('*.ats'):
+    if (path.name != 'main.ats'):
+        glue(collection=path.parts[Nr + 1])
+        source(path.absolute())
```
TotoGaz, line 21:
```suggestion
def main():
    # Set default ats parameters
    glue(collection='')
    glue(checks='')
    glue(solvers='')
    glue(outputs='')
    glue(constitutive_models='')

    # Recursively search the child directories and add them to ATS
    root_path = Path(ats.tests.AtsTest.getOptions().get("atsRootDir"))
    Nr = len(root_path.parts)
    for path in root_path.rglob('*.ats'):
        if (path.name != 'main.ats'):
            glue(collection=path.parts[Nr + 1])
            source(path.absolute())


if __name__ == "__main__":
    main()          
```
Does it make sense? Is it working or is there black magic?

host-configs/LLNL/quartz-base.cmake
TotoGaz: Is it also working if we put and hash?

src/CMakeLists.txt
```diff
@@ -291,12 +296,78 @@ endif()
...
+                       DEPENDS ats_environment
+                     )
+endif()
+
+
+# Python formatting
+if ( ENABLE_YAPF )
+    set( integrated_tests_python_sources )
+    file( GLOB_RECURSE integrated_tests_python_sources "${CMAKE_SOURCE_DIR}/../inputFiles/*.py" )
+     set( integrated_tests_ats_sources )
```
TotoGaz: warning indent.

src/CMakeLists.txt
```diff
@@ -291,12 +296,78 @@ endif()
...
+    file( GLOB_RECURSE integrated_tests_ats_sources "${CMAKE_SOURCE_DIR}/../inputFiles/*.ats" )
```
TotoGaz, line 367: 😎

src/CMakeLists.txt
```diff
@@ -291,12 +296,78 @@ endif()
...
+    endif() 
+
+
+     # Setup testing
+    set( ATS_SCRIPT
+        "${CMAKE_BINARY_DIR}/integratedTests/geos_ats.sh"
+       )
+
+    add_custom_command( OUTPUT ${ATS_SCRIPT}
+                        COMMAND ${CMAKE_BINARY_DIR}/bin/setup_ats_environment ${CMAKE_SOURCE_DIR}/.. ${CMAKE_BINARY_DIR} ${ATS_WORKING_DIR} ${ATS_BASELINE_DIR} ${ATS_ARGUMENTS}
```
TotoGaz: At some moment, maybe it'll be useful to have named options like `--src ...`

host-configs/LLNL/quartz-base.cmake
cssherman: That shouldn't be hard to hard to do

TotoGaz (approved):
Did you test your PR on the CI?

Also, something that can be really interesting is to mount the source directory read only `--volume=${GITHUB_WORKSPACE}:${GITHUB_WORKSPACE_MOUNT_POINT}` from https://github.com/GEOS-DEV/GEOS/blob/develop/.github/workflows/build_and_test.yml#L116 
Normally, we should not touch the source folder anymore so it should be OK.

@rrsettgast that could be a very nice way not to bother about cleaning after the build: when the `docker` container gets removed with the `--rm` option, then we're good.

inputFiles/compositionalMultiphaseFlow/benchmarks/SPE10/SPE10.ats
cssherman: I'm just about ready to start testing on the CI.  The last steps are to figure out how we want to fetch the baselines and the default path for things on the CI.

I'm not adding/removing any tests here.  Essentially, the script that I wrote did the following:

1) Find any .ats files in the integrated tests repo
2) Find any corresponding .xml file links in the same repo, resolving their absolute paths
3) For each set of (absolute) test directories, create a new test file with correct xml names (or a flag to indicate that a file isn't present in the target directory)

This was followed by some manual cleanup of invalid tests.

cssherman: We 100% should go through and further curate the test files

inputFiles/efemFractureMechanics/embeddedFractures_mechanics.ats
cssherman: Yeah, already done.  Something like this works in the inputFiles directory: `python -m yapf -i $(find -name "*.ats")`

inputFiles/main.ats
cssherman: I'm not entirely sure how ATS calls these files, so I'm not sure how it would react to a __main__ guard

host-configs/LLNL/quartz-base.cmake
cssherman: @TotoGaz - This is forwarded to this git clone arg, so a hash should work: https://git-scm.com/docs/git-clone#Documentation/git-clone.txt---branchltnamegt

PR #2866, opened 2023-12-04
Solution increments convergence criteria for sequential solver

paveltomin:
- add `SolutionIncrements` option for sequential convergence
- rename `saveIterationState` into `saveSequentialIterationState` and unify its logic (update `_k` fields after outer iteration is done, not in mapSolutionBetweenSolvers)

Bonus:
- `lineSearchStartingIteration` - defines iteration number from which line search becomes active (line search can be too conservative at first Newton iterations where residual does not necessarily decrease)
- separate timer for line search to track its specific cost
- shift "Last LinSolve(iter,res)" output down after linear solve
- pull `maxDeltaPhaseVolFrac = MpiWrapper::max( maxDeltaPhaseVolFrac )` out of subregion loop (observed big overhead for SEAM when number of even empty subregions is high)
- unify some registerField between CompositionalMultiphaseBase and SinglePhaseBase -> FlowSolverBase
- move deltaPressure update into "complete step" to avoid lagging from last time step

Updated baselines: GEOS-DEV/integratedTests/pull/96

PR #2967, opened 2024-01-30
Refactor contact solvers using inheritance from a SolidMechanics solver

paveltomin:
The goal of the PR is to detangle the contact solver design a bit to be able to allow multiphase poromechanics with contacts, wells, and sequential poromechanics with contacts later.

Main idea is to inherit contact solver from solid mechanics solver and add second template argument in poromechanics solver so that instead of:

`SinglePhasePoromechanicsConformingFractures = CoupledSolver< SinglePhasePoromechanics< SinglePhaseBase >, LagrangianContactSolver >`

now we have

`SinglePhasePoromechanicsConformingFractures = SinglePhasePoromechanics< SinglePhaseBase, LagrangianContactSolver >
`

Note that these changes simplify the input files for contact problems and should allow us to easily employ sequential coupling with contact and extend these fracture poromechanics solvers to coupling with wells and to multiphase flow. 

We are also renaming one of the solvers:

`LagrangianContactSolver` -> `SolidMechanicsLagrangeContact`

https://github.com/GEOS-DEV/integratedTests/pull/97

Coauthored by @CusiniM


src/coreComponents/physicsSolvers/contact/ContactSolverBase.hpp
```diff
@@ -53,19 +52,19 @@ class ContactSolverBase : public SolverBase
 
   string const & getContactRelationName() const { return m_contactRelationName; }
 
-  string const & getFractureRegionName() const { return m_fractureRegionName; }
+  string const & getFractureRegionName() const { return m_fractureRegionNames[0]; }
```
paveltomin: TODO double check

src/coreComponents/physicsSolvers/contact/LagrangianContactSolver.hpp
```diff
@@ -21,6 +21,8 @@
 #define GEOS_PHYSICSSOLVERS_CONTACT_LAGRANGIANCONTACTSOLVER_HPP_
 
 #include "physicsSolvers/contact/ContactSolverBase.hpp"
+#include "../../linearAlgebra/DofManager.hpp"
```
paveltomin: TODO cleanup

src/coreComponents/physicsSolvers/contact/ContactSolverBase.hpp
```diff
@@ -53,19 +52,30 @@ class ContactSolverBase : public SolverBase
 
   string const & getContactRelationName() const { return m_contactRelationName; }
```
CusiniM: I think we can remove this.

src/coreComponents/physicsSolvers/contact/ContactSolverBase.hpp
```diff
@@ -82,35 +92,53 @@ class ContactSolverBase : public SolverBase
...
-
-  /// fracture region name
-  string m_fractureRegionName;
-
-  /// pointer to the solid mechanics solver
-  SolidMechanicsLagrangianFEM * m_solidSolver;
+  void setFractureRegions( dataRepository::Group const & domain );
 
   /// contact relation name string
   string m_contactRelationName;
```
CusiniM: remove

inputFiles/lagrangianContactMechanics/ContactMechanics_SingleFracCompression_base.xml
```diff
@@ -6,13 +6,11 @@
     gravityVector="{0.0, 0.0, 0.0}">
     <LagrangianContact
       name="lagrangiancontact"
-      solidSolverName="lagsolve"
+      timeIntegrationOption="QuasiStatic"
```
CusiniM, line 9: we should make this the only option for contact solvers.

src/coreComponents/physicsSolvers/contact/ContactSolverBase.hpp
CusiniM: yeah for now, there a few places where we basically only handle a single fracture Region. I don't think it's a big deal. But maybe we can change this function to: `getUniqueFractureRegionName()`

src/coreComponents/physicsSolvers/contact/LagrangianContactSolver.cpp
```diff
@@ -65,33 +64,28 @@ LagrangianContactSolver::LagrangianContactSolver( const string & name,
     setInputFlag( InputFlags::REQUIRED ).
     setDescription( "Name of the stabilization to use in the lagrangian contact solver" );
 
-  m_linearSolverParameters.get().mgr.strategy = LinearSolverParameters::MGR::StrategyType::lagrangianContactMechanics;
-  m_linearSolverParameters.get().mgr.separateComponents = false;
-  m_linearSolverParameters.get().mgr.displacementFieldName = solidMechanics::totalDisplacement::key();
-  m_linearSolverParameters.get().dofsPerNode = 3;
+  LinearSolverParameters & linSolParams = m_linearSolverParameters.get();
+  linSolParams.mgr.strategy = LinearSolverParameters::MGR::StrategyType::lagrangianContactMechanics;
+  linSolParams.mgr.separateComponents = true;
```
CusiniM, line 69: I don't think this can be applied since we have AMG as a coarse solver only. Not too sure what the effect of this would be.

src/coreComponents/physicsSolvers/contact/LagrangianContactSolver.hpp
```diff
@@ -21,6 +21,8 @@
 #define GEOS_PHYSICSSOLVERS_CONTACT_LAGRANGIANCONTACTSOLVER_HPP_
 
 #include "physicsSolvers/contact/ContactSolverBase.hpp"
+#include "linearAlgebra/DofManager.hpp"
+#include "common/DataTypes.hpp"
```
CusiniM: do we really need these 2 includes?

src/coreComponents/physicsSolvers/contact/SolidMechanicsEmbeddedFractures.cpp
```diff
@@ -142,26 +130,19 @@ void SolidMechanicsEmbeddedFractures::resetStateToBeginningOfStep( DomainPartiti
...
+  SolidMechanicsLagrangianFEM::implicitStepComplete( time_n, dt, domain );
 
   forDiscretizationOnMeshTargets( domain.getMeshBodies(), [&] ( string const &,
                                                                 MeshLevel & mesh,
                                                                 arrayView1d< string const > const & )
   {
 
     ElementRegionManager & elemManager = mesh.getElemManager();
-    SurfaceElementRegion & region = elemManager.getRegion< SurfaceElementRegion >( m_fractureRegionName );
+    SurfaceElementRegion & region = elemManager.getRegion< SurfaceElementRegion >( m_fractureRegionNames[0] );
```
CusiniM:
```suggestion
    SurfaceElementRegion & region = elemManager.getRegion< SurfaceElementRegion >( getUniqueFractureRegionName() );
```

src/coreComponents/physicsSolvers/contact/SolidMechanicsEmbeddedFractures.cpp
```diff
@@ -296,7 +272,7 @@ void SolidMechanicsEmbeddedFractures::assembleSystem( real64 const time,
   {
     NodeManager const & nodeManager = mesh.getNodeManager();
     ElementRegionManager & elemManager = mesh.getElemManager();
-    SurfaceElementRegion & region = elemManager.getRegion< SurfaceElementRegion >( m_fractureRegionName );
+    SurfaceElementRegion & region = elemManager.getRegion< SurfaceElementRegion >( m_fractureRegionNames[0] );
```
CusiniM:
```suggestion
    SurfaceElementRegion & region = elemManager.getRegion< SurfaceElementRegion >( getUniqueFractureRegionName() );
```

src/coreComponents/physicsSolvers/contact/SolidMechanicsEmbeddedFractures.cpp
```diff
@@ -695,7 +668,7 @@ void SolidMechanicsEmbeddedFractures::updateJump( DofManager const & dofManager,
...
```
CusiniM:
```suggestion
    SurfaceElementRegion & region = elemManager.getRegion< SurfaceElementRegion >( getUniqueFractureRegionName() );
```

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoromechanicsConformingFractures.cpp
```diff
@@ -199,61 +179,33 @@ void SinglePhasePoromechanicsConformingFractures::assembleSystem( real64 const t
...
+                                                                                     CRSMatrixView< real64, globalIndex const > const & localMatrix,
+                                                                                     arrayView1d< real64 > const & localRhs )
 {
   GEOS_UNUSED_VAR( time_n, dt );
 
-  /// 3. assemble Force Residual w.r.t. pressure and Fluix mass residual w.r.t. displacement
+  /// 3. assemble Force Residual w.r.t. pressure and Flow mass residual w.r.t. displacement
+
+  Base::assembleElementBasedTerms( time_n, dt, domain, dofManager, localMatrix, localRhs );
+  this->solidMechanicsSolver()->getMaxForce() = 0.0; // this is to be consistent with old version
```
CusiniM:
```suggestion
  this->solidMechanicsSolver()->getMaxForce() = 0.0; 
 /// TODO: this is to be consistent with old version but it should be changed. We likely want to scale the residual for contact the same way we scale the mechanics one.
```

CusiniM (approved):
Just a couple of minor things to fix then we should merge this.

inputFiles/lagrangianContactMechanics/ContactMechanics_SingleFracCompression_base.xml
IsaacJu-debug: So in my local branch, the main difference from `LagrangianContact` solver is how to compute the quadrature points on fracture faces. Do we need to have an option in `LagrangianContact` solver to see what options to use for evaluating quadrature points? Or, we should do something to find the correct discretization method to dispatch 2D elements, something like we do dispatch 3D.

paveltomin: > we should make this the only option for contact solvers.

added a check for that

paveltomin: > So in my local branch, the main difference from `LagrangianContact` solver is how to compute the quadrature points on fracture faces. Do we need to have an option in `LagrangianContact` solver to see what options to use for evaluating quadrature points? Or, we should do something to find the correct discretization method to dispatch 2D elements, something like we do dispatch 3D.

@CusiniM ?

CusiniM: I mean yes and yes... I would not worry about FE types on the faces at this point. It's a non-trivial amount of work and I would wait for Shiva to be more mature.

src/coreComponents/physicsSolvers/contact/ContactSolverBase.hpp
paveltomin: how about throwing an error (for now) when there is more than one fracture region?

src/coreComponents/physicsSolvers/contact/LagrangianContactSolver.cpp
paveltomin: i didn't change that part and not sure i understand what you mean exactly
one tricky thing is that now the parent class is assigning some of these parameters and i will double check that there is no conflict here

src/coreComponents/physicsSolvers/contact/LagrangianContactSolver.hpp
paveltomin: removed, thanks

src/coreComponents/physicsSolvers/contact/LagrangianContactSolver.cpp
CusiniM: these calls should overwrite whatever the parent class assigns though. I think I had added an integratedTest using MGR for the embedded case. If that one works I think we should be okay.

paveltomin: just to document:

LagrangianContactSolver (child)
```
  linSolParams.mgr.strategy = LinearSolverParameters::MGR::StrategyType::lagrangianContactMechanics;
  linSolParams.mgr.separateComponents = true;
  linSolParams.mgr.displacementFieldName = solidMechanics::totalDisplacement::key();
  linSolParams.dofsPerNode = 3;
```

SolidMechanicsLagrangianFEM (parent)
```
  linParams.isSymmetric = true;
  linParams.dofsPerNode = 3;
  linParams.amg.separateComponents = true;
```

it does not seem to affect the results:
```
Rank 0 is comparing /tmp/geos-build/integratedTests/workingDir/poromechanicsFractures/PoroElastic_efem-edfm_inclinedFrac_smoke_01/0to1_restart_000000010/rank_0000000.hdf5 with /tmp/geos/integratedTests/poromechanicsFractures/PoroElastic_efem-edfm_inclinedFrac_smoke_01/0to1_restart_000000010/rank_0000000.hdf5 
...
Error: /Problem/Solvers/fractureMechSolver/LinearSolverParameters/amgSeparateComponents/__values__
	Scalar values of types int32 and int32 differ: 1, 0.
```
but no solution diffs are detected

paveltomin:
there are 3 additional cases that started failing restart check after merging https://github.com/GEOS-DEV/GEOS/commit/c11ad8ecd2d0d0aba5cd0661eb6e72951252409d
cases are:
- grav_seg_c1ppu_hyst_01
- fractureMatrixFlowWithGravity_edfm_verticalFrac_smoke_01 and fractureMatrixFlowWithGravity_edfm_verticalFrac_smoke_02
and the diffs are in `deltaPressure`, not sure why, seems they were rebased
example:
```
Error: /Problem/domain/MeshBodies/mesh/meshLevels/Level0/ElementRegions/elementRegionsGroup/region/elementSubRegions/cellBlock/deltaPressure
	Arrays of types float64 and float64 have 100 values of which 1 fail both the relative and absolute tests.
		Max absolute difference is at index (42,): value = -8628.164332123473, base_value = -8628.134468052536
		Max relative difference is at index (33,): value = -28.418695440515876, base_value = -28.388831375166774

```
 `deltaPressure` computations were indeed changed in https://github.com/GEOS-DEV/GEOS/pull/2866, namely the compute was moved from `implicitStepSetup` to `implicitStepComplete` to avoid lagging `deltaPressure` output by one time step

src/coreComponents/physicsSolvers/contact/ContactSolverBase.hpp
paveltomin: added an error here https://github.com/GEOS-DEV/GEOS/blob/4fd813d740c65a17257e280ff669fea8782598e5/src/coreComponents/physicsSolvers/contact/ContactSolverBase.cpp#L113
hope it is temporary ok, until we further clean up the fracture region logic

CusiniM:
> there are 3 additional cases that started failing restart check after merging [c11ad8e](https://github.com/GEOS-DEV/GEOS/commit/c11ad8ecd2d0d0aba5cd0661eb6e72951252409d) cases are:
> 
>     * grav_seg_c1ppu_hyst_01
> 
>     * fractureMatrixFlowWithGravity_edfm_verticalFrac_smoke_01 and fractureMatrixFlowWithGravity_edfm_verticalFrac_smoke_02
>       and the diffs are in `deltaPressure`, not sure why, seems they were rebased
>       example:
> 
> 
> ```
> Error: /Problem/domain/MeshBodies/mesh/meshLevels/Level0/ElementRegions/elementRegionsGroup/region/elementSubRegions/cellBlock/deltaPressure
> 	Arrays of types float64 and float64 have 100 values of which 1 fail both the relative and absolute tests.
> 		Max absolute difference is at index (42,): value = -8628.164332123473, base_value = -8628.134468052536
> 		Max relative difference is at index (33,): value = -28.418695440515876, base_value = -28.388831375166774
> ```
> 
> `deltaPressure` computations were indeed changed in #2866, namely the compute was moved from `implicitStepSetup` to `implicitStepComplete` to avoid lagging `deltaPressure` output by one time step

are these also failing on develop? If it's only the `deltaPressure` I don't think we should be worried.

src/coreComponents/physicsSolvers/contact/LagrangianContactSolver.cpp
CusiniM: Yeah, I don't think it does anything if it's not the first MGR level F-relaxation.

PR #2968, opened 2024-01-31
refactor: Remove all instances of array1d<string>

### LINKED ISSUE #1766: remove alias for `string_array` etc.
So what's our policy about `string_array`? Should we use it or `array1d< string >`?

_Originally posted by @TotoGaz in https://github.com/GEOSX/GEOSX/pull/1682#discussion_r801208733_

We typically have stopped using all of these aliases. It provides very little benefit to have them at this point, so I propose they are all removed.

rrsettgast:
In an attempt to identify some errors we are seeing on cuda systems, I removed all `array1d<string>` from the code and replaced it with `string_array` which aliases `std::vector<string>`....it remains to be seen if this fixes the problem.

The error can be seen here:
https://github.com/GEOS-DEV/GEOS/actions/runs/7674651265/job/20919616872?pr=2873


src/coreComponents/constitutive/fluid/multifluid/CO2Brine/CO2BrineFluid.cpp
```diff
@@ -226,9 +226,9 @@ void CO2BrineFluid< PHASE1, PHASE2, FLASH >::createPVTModels()
 
   // TODO: get rid of these external files and move into XML, this is too error prone
   // For now, to support the legacy input, we read all the input parameters at once in the arrays below, and then we create the models
-  array1d< array1d< string > > phase1InputParams;
+  array1d< string_array > phase1InputParams;
   phase1InputParams.resize( 3 );
-  array1d< array1d< string > > phase2InputParams;
+  array1d< string_array > phase2InputParams;
```
wrtobin: we should presumably not nest these in `array1d`s?

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/PhaseModel.hpp
```diff
@@ -55,7 +55,7 @@ struct PhaseModel
    * @param[in] componentMolarWeight molar weights of the components
    */
   PhaseModel( string const & phaseModelName,
-              array1d< array1d< string > > const & inputParams,
+              array1d< string_array > const & inputParams,
```
wrtobin: ditto

src/coreComponents/constitutive/fluid/multifluid/reactive/ReactiveBrineFluid.cpp
```diff
@@ -131,7 +131,7 @@ void ReactiveBrineFluid< PHASE > ::createPVTModels()
 
   // TODO: get rid of these external files and move into XML, this is too error prone
   // For now, to support the legacy input, we read all the input parameters at once in the arrays below, and then we create the models
-  array1d< array1d< string > > phase1InputParams;
+  array1d< string_array > phase1InputParams;
```
wrtobin: ditto

src/coreComponents/mesh/generators/InternalMeshGenerator.cpp
```diff
@@ -165,7 +165,11 @@ void InternalMeshGenerator::postProcessInput()
     if( m_elementType.size() == 1 )
     {
       string const elementType = m_elementType[0];
-      m_elementType.resizeDefault( m_numElePerBox.size(), elementType );
+      m_elementType.resize( m_numElePerBox.size() );
+      for( size_t i = 1; i < m_elementType.size(); ++i )
```
wrtobin: could we use the `equate` you added here?

src/coreComponents/mesh/generators/InternalMeshGenerator.cpp
```diff
@@ -198,7 +202,11 @@ void InternalMeshGenerator::postProcessInput()
       if( m_regionNames.size() == 1 )
       {
         string const regionName = m_regionNames[0];
-        m_regionNames.resizeDefault( numBlocks, regionName );
+        m_regionNames.resize( numBlocks );
+        for( size_t i = 1; i < m_regionNames.size(); ++i )
```
wrtobin: could we use the `equate` you added here?

src/coreComponents/physicsSolvers/wavePropagation/AcousticElasticWaveEquationSEM.hpp
```diff
@@ -175,8 +175,8 @@ class AcousticElasticWaveEquationSEM : public CoupledWaveSolver< AcousticWaveEqu
   virtual void initializePostInitialConditionsPreSubGroups() override;
 
   SortedArray< localIndex > m_interfaceNodesSet;
-  arrayView1d< string const > m_acousRegions;
-  arrayView1d< string const > m_elasRegions;
+  string_array const * m_acousRegions;
+  string_array const * m_elasRegions;
```
wrtobin: do we need raw pointers here? if not we might prefer a weak pointer type or a reference wrapper?

src/coreComponents/unitTests/dataRepositoryTests/testRestartBasic.cpp
```diff
@@ -107,9 +107,9 @@ using TestTypes = ::testing::Types< int,
                                     std::vector< int >,
                                     // std::vector< string > bufferOps currently can't pack this
                                     array1d< double >,
-                                    array1d< string >,
+                                    string_array,
                                     array1d< array1d< double > >,
-                                    array1d< array1d< string > >,
+                                    array1d< string_array >,
```
wrtobin: again do we want this nesting?

wrtobin (approved):

TotoGaz:
For info, what's the problem?

wrtobin:
There was some motivation coming from an error on frontier, but IIRC we've since moved past that.

Regardless, there is never a reason (and should never be, by design) we would need to move strings between memory spaces, which is one of the primary purposes of the lvarray arrays. String array1ds have also been the most likely case to fail in lvarray unit tests when making changes to lvarray itself, so if we don't need to support that usage it will probably save a bit of dev time.

corbett5:
> There was some motivation coming from an error on frontier, but IIRC we've since moved past that.
> 
> Regardless, there is never a reason (and should never be, by design) we would need to move strings between memory spaces, which is one of the primary purposes of the lvarray arrays. String array1ds have also been the most likely case to fail in lvarray unit tests when making changes to lvarray itself, so if we don't need to support that usage it will probably save a bit of dev time.

On the other hand I can think of twp reasons for the status-quo. First is that everything is stored as an `LvArray:Array` and we don't need to support `std::vector`. The second is that as long as we're still using `LvArray::Array<LvArray::Array>`, we want those `LvArray::Array<std::string>` unit tests to pass and should work to maintain them.

Finally from an LvArray perspective if we want to only support trivial types and remove support for everything else that would be a major breaking change. It would certainly simplify the code a lot and make it easier to maintain but I'm also sure someone exists who would like to have a multi-dimensional array of non-trivial types.

wrtobin:
Randy also pointed to streak (hardware-wise its essentially 1 node of lassen we're working using as a CI runner) which is having cuda memory errors with lvarray string arrays: https://github.com/GEOS-DEV/GEOS/actions/runs/7674651265/job/20919616872?pr=2873

Definitely agree with uniformity from a design perspective, though if we're encountering errors supporting a usage we don't need I'm fine losing uniformity to resolve the error.

The frontier issue seems to persist despite the experimental branch, so I think don't think we actually need to push for any changes to LvArray to drop support for non-trivial types (though I stand by that the string-typed lvarray unit tests are often what needs to be fixed when making changes and it could be a developer time savings to drop them -- though that would certainly be a breaking change).

klevzoff:
> The error can be seen here:
https://github.com/GEOS-DEV/GEOS/actions/runs/7674651265/job/20919616872?pr=2873

I don't see any erros in that log that have anything to do with `array<string>`? There's a bunch of deallocation errors for `array<double>` though. They all begin after a segfault in `testLifoStorage`, so I would start with fixing that one first.

paveltomin:
@rrsettgast abandon?

rrsettgast:
@wrtobin What do you think? I still think maybe we should do this even though it wasn't the culprit I was looking for

paveltomin:
> @wrtobin What do you think? I still think maybe we should do this even though it wasn't the culprit I was looking for

@wrtobin @rrsettgast ping

rrsettgast:
@corbett5 This PR was an attempt to fix a problem we were having on GPU systems. There were some stack traces that indicated that `Array<string>` was causing some sort of memory error....but it ended up not being the issue. Still...wondering if we should disallow `LvArray::Array<string>` as they shouldn't really ever go onto the GPU.

corbett5:
> @corbett5 This PR was an attempt to fix a problem we were having on GPU systems. There were some stack traces that indicated that `Array<string>` was causing some sort of memory error....but it ended up not being the issue. Still...wondering if we should disallow `LvArray::Array<string>` as they shouldn't really ever go onto the GPU.

Certainly I agree that `LvArray::Array<string>` should never go on the GPU. It's not clear to me whether the benefit of  using `std::vector<string>` instead of `Array<string>` to communicate that these objects shouldn't be on the GPU outweighs the potential complication of having to write functions that take a `std::vector` as well as `LvArray::Array`. I think either choice is fine.

CusiniM:
@rrsettgast can we close this one?

src/coreComponents/mesh/generators/InternalMeshGenerator.cpp
rrsettgast: there is an overload for this

src/coreComponents/mesh/generators/InternalMeshGenerator.cpp
rrsettgast: same

paveltomin (approved):

CusiniM (approved):
it's kinda hard to review but seems fine to me.

src/coreComponents/mesh/generators/InternalMeshGenerator.cpp
```diff
@@ -194,12 +193,11 @@ void InternalMeshGenerator::postInputInitialization()
...
       numBlocks *= m_nElems[i].size();
     }
-    if( numBlocks != m_regionNames.size() )
+    if( numBlocks != LvArray::integerConversion< localIndex >( m_regionNames.size()) )
     {
       if( m_regionNames.size() == 1 )
       {
-        string const regionName = m_regionNames[0];
-        m_regionNames.resizeDefault( numBlocks, regionName );
+        m_regionNames.resize(numBlocks, m_regionNames[0]);
```
paveltomin, line 200: this line becomes `m_regionNames.resize( numBlocks, m_regionNames[0] )` (with spaces) after `make uncrustify_style`

PR #2969, opened 2024-01-31
remove RESQML support from GEOS

untereiner:
Companion PR https://github.com/GEOS-DEV/thirdPartyLibs/pull/258


untereiner:
@TotoGaz I don't know how to deal with the spack recipes.  Should I remove them too ? They could be necessary for building geos with spack with the RESQML external component

bmhan12:
@untereiner I can remove them from spack if they are no longer being used.

TotoGaz:
That would be nice, thank you @bmhan12 
@untereiner does it make sense to keep the `spack` recipe if we're not using/validating it anymore? Is the `git` history enough if we want to revive it?

untereiner:
Thank you @bmhan12 ! If think the git history is enough. I am wondering how to play with these recipes with what I call an « external component » eg: added at compile time but in another repository

TotoGaz:
> I am wondering how to play with these recipes with what I call an « external component » eg: added at compile time but in another repository

You mean `spack` recipes? I've never been using `spack` so I do not know.

untereiner:
Yes, I ment spack recipes.

untereiner:
@TotoGaz This one should be easy

TotoGaz (approved):

PR #2974, opened 2024-02-02
Trying to fix pressure spikes for fractures (VTK)

paveltomin:
Running single phase case in parallel produces pressure spikes:
image
These spikes are result of application of fixed pressure boundary condition like that:
```
        <FieldSpecification
                name="source4"
                objectPath="ElementRegions"
                fieldName="pressure"
                scale="5.0e7"
                setNames="{ source4 }"/>
```
The BC is wrongly applied because there is an element with zero number of nodes (in fracture region) and this element ends up in BC set.

This PR fixes zero-nodes element assignment to a set and adds a secondary call for set construction after setupGhosts is performed.

Single phase case will run fine without pressure spikes:
image
Multiphase case also runs.


src/coreComponents/mesh/generators/VTKUtilities.cpp
```diff
@@ -1646,6 +1646,8 @@ CellMapType buildCellMap( vtkDataSet & mesh, string const & attributeName )
   // Now, actually split into groups according to region attribute, if present
   vtkDataArray * const attributeDataArray =
     vtkDataArray::FastDownCast( mesh.GetCellData()->GetAbstractArray( attributeName.c_str() ) );
+  GEOS_LOG_RANK_0_IF( !attributeName.empty() && attributeDataArray == nullptr,
```
paveltomin, line 1649: this thing is not relevant to the main topic of the PR

src/coreComponents/mesh/ElementRegionManager.cpp
```diff
@@ -244,7 +244,9 @@ void ElementRegionManager::buildSets( NodeManager const & nodeManager )
       {
         localIndex const numNodes = subRegion.numNodesPerElement( k );
```
paveltomin, line 245: @TotoGaz side question - how it that going to work for heterogenous subregion?

TotoGaz:
> The BC is wrongly applied because there is an element with zero number of nodes (in fracture region) and this element ends up in BC set.
> This PR fixes zero-nodes element assignment to a set, but does not resolve the core issue - existence of elements like that.

Thank you for your initiative. In `geos` fracture elements do not rely on their nodes but point to nodes from "standard" 3d cells. When ghosting, that forces me to tweak the algorithm, which, in this precise case, is not working. I'm trying to understand the root cause. Or maybe I should find a workaround to wait for me to rewrite the ghosting. 🤷

CusiniM:
> > The BC is wrongly applied because there is an element with zero number of nodes (in fracture region) and this element ends up in BC set.
> > This PR fixes zero-nodes element assignment to a set, but does not resolve the core issue - existence of elements like that.
> 
> Thank you for your initiative. In `geos` fracture elements do not rely on their nodes but point to nodes from "standard" 3d cells. When ghosting, that forces me to tweak the algorithm, which, in this precise case, is not working. I'm trying to understand the root cause. Or maybe I should find a workaround to wait for me to rewrite the ghosting. 🤷

irrespective of this issue, Pavel's fix seems appropriate. I know that an element without any node does not necessarily mean much but, if one existed like in this case, I don't think it should automatically be added to all sets. 

If the fracture element has zero nodes, I am a bit surprised that we can solve though coz I would expect it to be disconnected from everything else which should lead to a singular system...

TotoGaz:
> irrespective of this issue, Pavel's fix seems appropriate. I know that an element without any node does not necessarily mean much but, if one existed like in this case, I don't think it should automatically be added to all sets.

I'm not sure: if such a thing happens, we can argue that the simulation is in an inconsistent state and should be stopped.

> If the fracture element has zero nodes, I am a bit surprised that we can solve though coz I would expect it to be disconnected from everything else which should lead to a singular system...

If it's connected to nothing with values prescribed by BC, then it may only generate a hole in the mesh.

CusiniM:
> > irrespective of this issue, Pavel's fix seems appropriate. I know that an element without any node does not necessarily mean much but, if one existed like in this case, I don't think it should automatically be added to all sets.
> 
> I'm not sure: if such a thing happens, we can argue that the simulation is in an inconsistent state and should be stopped.
> 
> > If the fracture element has zero nodes, I am a bit surprised that we can solve though coz I would expect it to be disconnected from everything else which should lead to a singular system...
> 
> If it's connected to nothing with values prescribed by BC, then it may only generate a hole in the mesh.

Agreed but this PR should exactly prevent it from happening since it's removing the element from the B.C. set so it's kind of mysterious how it can actually be solved. My guess is that the element is at least connected to something.

paveltomin:
> > > irrespective of this issue, Pavel's fix seems appropriate. I know that an element without any node does not necessarily mean much but, if one existed like in this case, I don't think it should automatically be added to all sets.
> > 
> > 
> > I'm not sure: if such a thing happens, we can argue that the simulation is in an inconsistent state and should be stopped.
> > > If the fracture element has zero nodes, I am a bit surprised that we can solve though coz I would expect it to be disconnected from everything else which should lead to a singular system...
> > 
> > 
> > If it's connected to nothing with values prescribed by BC, then it may only generate a hole in the mesh.
> 
> Agreed but this PR should exactly prevent it from happening since it's removing the element from the B.C. set so it's kind of mysterious how it can actually be solved. My guess is that the element is at least connected to something.

I am sorry for confusion here.
There is actually no element without nodes when simulation starts. The nodes are attached during setupGhosts.
But the problem is that BC sets are constructed before that and that element with not yet attached nodes getting added into the set.
What this PR does:
1) prevents adding no-nodes element into BC (and initial conditions etc) sets.
2) calls the sets building procedure one more time after the setupGhosts, so now we have all elements with nodes and can finally add them into sets.
Would be happy to discuss and clarify more.

TotoGaz:
So I've been testing your PR on a simple case.

When there's no ghosting issues, the parallel results I get on my cube are identical to serial.
In configurations where I have the spike, your PR deleted the spike but the serial and parallel results are different, while it's true that the parallel results are now good looking.

As a summary, it looks to me that this PR is dangerous to merge as is, because wrong results are still wrong and would be far more complicated to detect.

paveltomin:
> So I've been testing your PR on a simple case.
> 
> When there's no ghosting issues, the parallel results I get on my cube are identical to serial. In configurations where I have the spike, your PR deleted the spike but the serial and parallel results are different, while it's true that the parallel results are now good looking.
> 
> As a summary, it looks to me that this PR is dangerous to merge as is, because wrong results are still wrong and would be far more complicated to detect.

I see, thanks for testing
Just curious - how big is the difference, could you please share?

TotoGaz:
I think you can try this case, which runs OK until rank 5. I'm using it to debug.
[debug_ghosting.tar.gz](https://github.com/GEOS-DEV/GEOS/files/14146879/debug_ghosting.tar.gz)

paveltomin:
> I think you can try this case, which runs OK until rank 5. I'm using it to debug. [debug_ghosting.tar.gz](https://github.com/GEOS-DEV/GEOS/files/14146879/debug_ghosting.tar.gz)

this function https://github.com/GEOS-DEV/GEOS/blob/d306d3f8ac8aac8a3cdf6778e3041bbd7ab4375d/src/coreComponents/mesh/generators/VTKFaceBlockUtilities.cpp#L274
does not work properly specifically for mpi=5
there are no proper edges yet to map to and the ones being picked are wrong (they only have one collocated node, while as far as i understand it should be two)

TotoGaz:
> this function does not work properly specifically for mpi=5
> there are no proper edges yet to map to and the ones being picked are wrong (they only have one collocated node, while as far as i understand it should be two)

That could be a good trail to follow. Can you try to fix it so we see how it goes?
In the idea, something like and we see how it goes
```c++
auto const res = std::max_element( edgeCount.cbegin(), edgeCount.cend(), comp );
if( res.second == 2 )
{
    face2dToEdge[i] = LvArray::integerConversion< localIndex >( res->first );
}
else
{
    face2dToEdge[i] = -1;
}
```

PR #2976, opened 2024-02-02
Add solubility table input to CO2-Brine model

dkachuma:
This is what I had in mind. Having done it however, I realise it's rather convoluted. Especially if we want to do it for everything including density, viscosity and enthalpy. Not sure how good an idea this is. Maybe a simple implied name is best.


src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/CO2Solubility.hpp
```diff
@@ -86,7 +84,7 @@ class CO2SolubilityUpdate final : public FlashModelBaseUpdate
   TableFunction::KernelWrapper m_CO2SolubilityTable;
 
   /// Table with water vapourisation as a function (P,T)
-  //TableFunction::KernelWrapper m_WaterVapourisationTable;
+  TableFunction::KernelWrapper m_WaterVapourisationTable;
```
dkachuma, line 87: This currently won't build until LvArray PR311 is merged.

paveltomin:
thanks a lot, @dkachuma, it looks ok actually, let me think about other tables
by the way - did you change the flash constructor in any of your PRs?
having a structure with flash parameters would simplify a lot, instead of passing those strings

paveltomin:
@dkachuma how about this - we will add this option to provide solubility table name, and we will for now keep hard-coded name option for all other tables?
up to now users only requested the solubility table input, and we can add other props one-by-one later
would that work for you?

dkachuma:
> did you change the flash constructor in any of your PRs? having a structure with flash parameters would simplify a lot, instead of passing those strings

No - it's all passed as strings. I did a hack where the second element in the array of strings shows that explicit tables have been defined. I suppose we could add a second constructor to deal with this case. I agree with having a structure being a good option. Ultimately I would say we can put the input in the xml instead of these separate text files.

dkachuma:
> @dkachuma how about this - we will add this option to provide solubility table name, and we will for now keep hard-coded name option for all other tables?

That's OK I suppose.

paveltomin:
> > @dkachuma how about this - we will add this option to provide solubility table name, and we will for now keep hard-coded name option for all other tables?
> 
> That's OK I suppose.

Ok, thanks, I just merged your code into my branch

PR #2981, opened 2024-02-08
ci: pygeosx testing

bmhan12:
This PR adds pygeosx unit tests to Github Actions, specfically images with Ubuntu 22 to satisfy the python version requirement (>=3.10) .

Related to GEOS-DEV/thirdPartyLibs#259


scripts/ci_build_and_test_in_container.sh
```diff
@@ -259,6 +263,12 @@ if [[ "${RUN_UNIT_TESTS}" = true ]]; then
   fi
 fi
 
+if [[ "${RUN_PYGEOSX_TESTS}" = true ]]; then
+  or_die ninja pygeosx
+  or_die ninja geosx_python_tools
+  or_die ninja pygeosx_unit_tests
+fi
```
TotoGaz: I think this should go aside the `if [[ "${BUILD_EXE_ONLY}" = true ]]; then` block, before the `sccache` block for example.

.github/workflows/build_and_test.yml
```diff
@@ -140,6 +143,14 @@ jobs:
         docker_args+=(-e ENABLE_HYPRE_DEVICE=${ENABLE_HYPRE_DEVICE:-CPU})
         docker_args+=(-e ENABLE_TRILINOS=${ENABLE_TRILINOS:-ON})
 
+
+        # pygeosx testing
+        ENABLE_PYGEOSX=${{ inputs.ENABLE_PYGEOSX }}
+        docker_args+=(-e ENABLE_PYGEOSX=${ENABLE_PYGEOSX:-OFF})
```
TotoGaz: I wanted to reduce the number of environment variables and favor direct CLI options.
Do you think we could have the shell script have some kind of `--pygeosx` flag that would trigger `pygeos`?
That would make a long list of options indeed, but in the end, it helps us to be more explicit about what we want.

untereiner:
Should `pygeosx` be renamed `pygeos` ? Would it be the good moment to do it ?

TotoGaz:
> Should `pygeosx` be renamed `pygeos` ? 

Yes it should https://github.com/GEOS-DEV/GEOS/issues/2331

> Would it be the good moment to do it ?

I guess that would be MAKUTU to steer the change @sframba

.github/workflows/build_and_test.yml
bmhan12: > Do you think we could have the shell script have some kind of --pygeosx flag that would trigger pygeos?

Yes, adjusted the script to pass a ` -DENABLE_PYGEOSX=ON` option instead of setting an environment variable.

cssherman (approved):
Looks good to me

bmhan12:
This PR has been updated (post-spack changes), and is ready for review.

paveltomin:
@rrsettgast review?

paveltomin (approved):

rrsettgast (approved):

PR #2983, opened 2024-02-09
add pragmas to supress warnings associated with compiler bugs

rrsettgast:
after https://github.com/GEOS-DEV/LvArray/pull/311 we get compiler warnings on quartz...as a result of compiler bugs. 

https://github.com/GEOS-DEV/LvArray/pull/316


castelletto1 (approved):
Looks good to me thanks. Let's open an issue to investigate the `SuperLU_DIST` problem when building GEOS with GPU support.

CusiniM (commented):
so, the `directParallel` solver issue on the GPUs is a separate one right?

rrsettgast:
> so, the `directParallel` solver issue on the GPUs is a separate one right?

yes

PR #2986, opened 2024-02-10
Fix guards for gcc warning pragmas

rrsettgast:
https://github.com/GEOS-DEV/LvArray/pull/317


herve-gross (approved):
It compiles again, thank you.

PR #2989, opened 2024-02-15
Refactor wave solvers: separation of time and space update inside wave solvers

acitrain:
This PR contains the second part of the refactoring of wave solvers. 
It focuses on the extraction of the time-scheme from the cpp files. The time scheme are now included in their dedicated files (one for each physics) abd only called inside the cpp.


src/coreComponents/physicsSolvers/wavePropagation/AcousticVTIWaveEquationSEM.cpp
```diff
@@ -586,54 +587,13 @@ real64 AcousticVTIWaveEquationSEM::explicitStepInternal( real64 const & time_n,
     addSourceToRightHandSide( cycleNumber, rhs );
 
     /// calculate your time integrators
-    real64 const dt2 = dt*dt;
```
tbeltzun, line 589: Use `real64 const dt2 = pow( dt, 2 );` for consistency with previous code ?

src/coreComponents/physicsSolvers/wavePropagation/AcousticTimeSchemeSEMKernel.hpp
```diff
@@ -0,0 +1,168 @@
...
+        if( lateralSurfaceNodeIndicator[a] != 1 && bottomSurfaceNodeIndicator[a] != 1 )
+        {
+          // Interior node, no boundary terms
+          p_np1[a] /= mass[a]/dt2;
+          q_np1[a] /= mass[a]/dt2;
+        }
+        else
+        {
+          // Boundary node
+          p_np1[a] += damping_p[a]*p_nm1[a]/dt/2;
```
tbeltzun, line 138: Can't we use an auxiliary variable `real32 const dt_2 = dt / 2` to avoid unnecessary divisions ?

Could we also enforce spacing around operators for consistency ?

src/coreComponents/physicsSolvers/wavePropagation/AcousticVTIWaveEquationSEM.cpp
```diff
@@ -586,54 +587,13 @@ real64 AcousticVTIWaveEquationSEM::explicitStepInternal( real64 const & time_n,
...
-          // Hand-made Inversion of 2x2 matrix
-          real32 coef_pp = mass[a]/dt2;
-          coef_pp += damping_p[a]/dt/2;
-          real32 coef_pq = damping_pq[a]/dt/2;
-
-          real32 coef_qq = mass[a]/dt2;
-          coef_qq += damping_q[a]/2/dt;
-          real32 coef_qp = damping_qp[a]/dt/2;
-
-          real32 det_pq = 1/(coef_pp * coef_qq - coef_pq*coef_qp);
```
tbeltzun, line 629: Unconsistent spacing.

tbeltzun (approved):

src/coreComponents/physicsSolvers/wavePropagation/AcousticTimeSchemeSEMKernel.hpp
rrsettgast: @acitrain It is better to do something like specifying a variable outside the loop to represent the inverse here.

src/coreComponents/physicsSolvers/wavePropagation/AcousticVTIWaveEquationSEM.cpp
rrsettgast: no...we should not have a call to `pow()` if it can be avoided.

rrsettgast (approved):

PR #2994, opened 2024-02-20
Remove rock from fracture materialList in HydrofractureSolver

paveltomin:
It should not be there and it is confusing to have it

https://github.com/GEOS-DEV/integratedTests/pull/98


jhuang2601 (approved):

CusiniM (approved):
There used to be a reason but I don't remember it. In any case, if everything works this is clearly better.

PR #2998, opened 2024-02-23
Enable high order for fwi in acoustic wave solver

acitrain:
This PR generalize the formula for gradient computation inside the acoustic second order wave solver to high orders

The main change is the computation of the number of nodes per elements instead of hardcoded it to 8


rrsettgast:
@sframba Is this deprecated or superseded?

acitrain:
We can close it, in fact the formula was right at the beginning so no need to change it

PR #3019, opened 2024-03-04
feat: Adding log parts

arng40:
This PR will be merged after : https://github.com/GEOS-DEV/GEOS/pull/3537 

This PR aim to improve the readibility of the logs by adding log separation the differents parts of the log and 
the first step for displaying specific log or timestep we want.

The log separation identified are 
-   **Input Initialisation** : `ProblemManager::postInputInitializationRecursive()`;
-   **Mesh generation** : `ProblemManager::generateMesh()`;
-   **NumericalMethods**: `ProblemManager::applyNumericalMethods()`;
-   **Mesh data registration** : `ProblemManager::egisterDataOnMeshRecursive()`;
-  **Group & subgroups initialization functions** :  `ProblemManager::initialize()`;
-  **Each cycle time-step** : `EventManager::run()`
-  **Cleanup** : `basicCleanup() `

Example of an input :

``` 
####################################################################################################
##                                            TIMESTEP                                            ##
####################################################################################################
##  - Time        : 00h00m00s out of 01h31m40s (0% completed)                                     ##
##                  0 s / 5500 s                                                                  ##
##  - Delta Time  : 00h08m20s (500 s)                                                             ##
##  - Cycle       : 0                                                                             ##

        ( Rflow ) = ( 4.82e-04 )        ( Rwell ) = ( 4.74e-03 )        ( R ) = ( 4.76e-03 )
        Last LinSolve(iter,res) = (   1, 1.64e-11 )
        compositionalMultiphaseFlow: Max pressure change = 19297.019 Pa (before scaling)
        compositionalMultiphaseFlow: Max component density change = 0.346 kg/m3 (before scaling)
        reservoirSystem: Global solution scaling factor = 1
        compositionalMultiphaseFlow: Max phase volume fraction change = 0.0004
[...]
Rank 0: Writing out restart file at ./staircase_co2_wells_3d_restart_000000004/rank_0000000.hdf5

##  - Cycle        : 0                                                                            ##
##  - N substeps   : 1                                                                            ##
##  - substep dts  : 500 s                                                                        ##
####################################################################################################
##                                        End of TIMESTEP                                         ##
####################################################################################################
```


src/coreComponents/codingUtilities/Section.hpp
```diff
@@ -0,0 +1,124 @@
...
+   * @brief Set the minimal width of a row
+   * @param minWidth The minimal width of the table
+   */
+  void setMinWidth( integer const & minWidth );
+
+  /**
+   * @brief Draw the first part of the section. It include the title and optionnaly, the description(s);
+   * @param os An output stream (by default, std::cout)
+   */
+  void begin( std::ostream & os = std::cout );
```
MelReyCG, line 64: Prefer something like `beginSection()` or `start()`, begin/end is a reserved keyword for iterators.

MelReyCG: Prefer something like `beginSection()` or `start()`, begin/end is a reserved keyword for iterators.

src/coreComponents/codingUtilities/Section.hpp
```diff
@@ -0,0 +1,124 @@
...
+{
+public:
+
+  Section();
+
+  /**
+   * @brief Set the name of the section
+   * @param m_sectionTitle The name of the section
+   */
+  void setName( string_view m_sectionTitle );
```
MelReyCG: For me, it feels strange that the name of a section is not `const` and not given at construction:
- what is the purpose of an instance?
- what happens if the name is reset between begin() and end()?

src/coreComponents/codingUtilities/Section.hpp
```diff
@@ -0,0 +1,124 @@
...
+
+  /**
+   * @brief Add a description to the section by concatening a description name and descriptions values.
+   * @param descriptionName The description name
+   * @param args Descriptions values to be aligned.
+   * Descriptions values can be be any types and will be aligned
+   */
+  template< typename ... Args >
+  void addDescription( string const & descriptionName, Args const & ... args );
```
MelReyCG: Currently this works, but if someone want to show the description at the `end()`, it will have the same strange behaviour I mentionned at `setName()`.
You can document `end()` implementation to never show non-const attributes, or set the description as constant too.

src/coreComponents/codingUtilities/Section.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+  auto it = std::max_element( rowsDescription.begin(),
+                              rowsDescription.end(),
+                              []( auto const & a, auto const & b ) {
+    return a.size() < b.size();
+  } );
+
+  string const maxDescriptionSize = *it;
+
+  integer const maxDescriptionLength = integer( maxDescriptionSize.length()) + m_marginBorder * 2 + m_nbSpecialChar * 2;
```
MelReyCG: long line + be careful of space before closing parenthesis

src/coreComponents/codingUtilities/Section.hpp
```diff
@@ -0,0 +1,124 @@
...
+
+private:
+
+  /**
+   * @brief Compute the max string size (m_rowLength) between title and the description(s)
+   * @param m_sectionTitle The table title
+   * @param descriptions The descriptions vector
+   * @return The max row length of the section
+   */
+  void computeMaxRowSize( string_view m_sectionTitle,
```
MelReyCG: `computeWidth()` / `computeSectionWidth()` ?
The concept of row is bound to tables / arrays, isn't it?

src/coreComponents/codingUtilities/Section.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+void Section::setMinWidth( integer const & minWidth )
+{
+  m_rowMinWidth = minWidth;
+}
+
+void Section::computeMaxRowSize( string_view title,
+                                 std::vector< string > const & rowsDescription )
+{
+  std::string const footerStr = "End : ";
```
MelReyCG: use string_view when possible.
```suggestion
  static string_view constexpr footerStr = "End : ";
```

src/coreComponents/codingUtilities/Section.hpp
```diff
@@ -0,0 +1,124 @@
...
+                          std::vector< string > const & descriptions );
+
+  /**
+   * @brief Build a description from the name and description values
+   * @param descriptionName The decription name
+   * @param decriptionsValues The description values
+   */
+  void formatAndInsertDescriptions( std::string const & descriptionName,
```
MelReyCG: Can't you use `string_view` parameter here? (There are other locations in your PR)

src/coreComponents/codingUtilities/Section.cpp
```diff
@@ -0,0 +1,131 @@
...
+                                        descriptionLength,
+                                        descriptionValues[idxValue] ) );
+  }
+}
+
+void Section::begin( std::ostream & oss )
+{
+  computeMaxRowSize( m_sectionTitle, m_descriptions );
+
+  string const lineSection =  GEOS_FMT( "{:#>{}}\n", "", m_rowLength );
```
MelReyCG: what do you think of
```suggestion
  string const lineSection =  GEOS_FMT( " {:#>{}} \n", "", m_rowLength-2 );
```
(also for bottom line)

src/coreComponents/codingUtilities/Section.cpp
MelReyCG: Is it possible to have the special character as a customisable `char` constant?

src/coreComponents/codingUtilities/Section.cpp
```diff
@@ -0,0 +1,131 @@
...
+  if( rowsDescription.size() == 0 )
+  {
+    return;
+  }
+
+  auto it = std::max_element( rowsDescription.begin(),
+                              rowsDescription.end(),
+                              []( auto const & a, auto const & b ) {
+    return a.size() < b.size();
+  } );
```
MelReyCG: here you can write it, as you prefer:
```suggestion
  auto it = std::max_element( rowsDescription.begin(), rowsDescription.end(),
                              []( auto const & a, auto const & b ) { return a.size() < b.size(); } );
```

src/coreComponents/codingUtilities/Section.cpp
```diff
@@ -0,0 +1,131 @@
...
+    oss << GEOS_FMT( "##{:<{}}{:<{}}##", " ", m_marginBorder, description, descriptionLength );
+    if( &description != &m_descriptions.back())
+    {
+      oss << '\n';
+    }
+    else
+    {
+      oss << "\n\n";
+    }
+  }
```
MelReyCG: I would avoid breaking the code flow with `return` at the middle:
```suggestion
  if( !m_descriptions.empty())
  {
    for( string const & description : m_descriptions )
    {
      oss << GEOS_FMT( "##{:<{}}{:<{}}##", " ", m_marginBorder, description, descriptionLength );
      if( &description == &m_descriptions.back())
      {
        oss << '\n';
      }
    }
  }
  oss << '\n';
```
Not the only place for this remark. I suggest getting some information about good practices about that.

src/coreComponents/codingUtilities/Section.cpp
```diff
@@ -0,0 +1,131 @@
...
+}
+
+void Section::end( std::ostream & oss ) const
+{
+  std::string footerTitle = "End : " + m_sectionTitle;
```
MelReyCG: I would factorize all formating expressions (I think there is more).

MelReyCG (changes requested):

src/coreComponents/codingUtilities/Section.hpp
arng40: move to constructor

src/coreComponents/codingUtilities/Section.hpp
arng40: setName removed

src/coreComponents/codingUtilities/Section.hpp
arng40: y i think it's better

src/coreComponents/codingUtilities/Section.cpp
arng40: it will change the layout

src/coreComponents/codingUtilities/Section.cpp
arng40: 2nd one better agree

src/coreComponents/codingUtilities/Section.hpp
```diff
@@ -0,0 +1,120 @@
...
+ * Copyright (c) 2019-     GEOSX Contributors
+ * All rights reserved
+ *
+ * See top level LICENSE, COPYRIGHT, CONTRIBUTORS, NOTICE, and ACKNOWLEDGEMENTS files for details.
+ * ------------------------------------------------------------------------------------------------------------
+ */
+
+/**
+ * @file Section.hpp
+ */
```
MelReyCG, line 17: `codingUtilities` is a "package" with only helping functions / types to extend the `std` (they could go in another project).
The `Section` component is a formatter, and by its format, is only dedicated to `GEOS`, so it cannot go in there.
You can start by adding it in `fileIO/Section` like you did with your tables.

src/coreComponents/codingUtilities/Section.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+#include "Section.hpp"
+#include <algorithm>
+
+namespace geos
+{
+
+Section::Section( string_view sectionTitle ):
+  m_sectionTitle( string( sectionTitle ) ),
+  m_rowMinWidth( 20 )
```
MelReyCG: I think this is way too short, as sections should be more visible than the average GEOS content. I would expect something around 100 characters. It may be shorter, but not that much.

Don't forget to update the description if it changes.

src/coreComponents/codingUtilities/Section.cpp
```diff
@@ -0,0 +1,122 @@
...
+      oss << horizontalChars;
+      oss << '\n';
+    }
+  }
+  oss << "\n";
+}
+
+void Section::endSection( std::ostream & oss ) const
+{
+  string const footerTitle = GEOS_FMT( "{} {}", "End :", m_sectionTitle );
```
MelReyCG: `GEOS_FMT( "End : {}", m_sectionTitle );` would be sufficient + the previous comment on this line has not been adressed.

src/coreComponents/codingUtilities/Section.hpp
```diff
@@ -0,0 +1,120 @@
...
+
+private:
+
+  /**
+   * @brief Compute the max string size (m_rowLength) between title and the description(s)
+   * @param m_sectionTitle The table title
+   * @param descriptions The descriptions vector
+   * @return The max row length of the section
+   */
+  void computeMaxWidth( string_view m_sectionTitle,
```
MelReyCG: Actually, the point of this method is to compute `m_rowLength`.
I would not talk about "max" things, because it is related to implementation (not useful to the class user).
I would rename it to `computeWidth()`, and rename the `m_rowLength` to `m_sectionWidth`.

MelReyCG (changes requested):

src/coreComponents/codingUtilities/Section.cpp
MelReyCG: Indeed, I am wondering if this layout is the best.

src/coreComponents/codingUtilities/Section.cpp
arng40: updated

src/coreComponents/events/EventManager.cpp
```diff
@@ -207,6 +208,7 @@ bool EventManager::run( DomainPartition & domain )
       }
     }
 
+    section.endSection();
```
MelReyCG: @arng40 Section end output must be optionally able to include information, in order to support PR https://github.com/GEOS-DEV/GEOS/pull/3303 modifications

MelReyCG (changes requested):

src/coreComponents/fileIO/section/Section.cpp
```diff
@@ -0,0 +1,121 @@
...
+      oss << '\n';
+    }
+  }
+  oss << "\n";
+}
+
+void Section::endSection( std::ostream & oss ) const
+{
+  string const footerTitle = GEOS_FMT( "{}{}", m_footerTitle, m_sectionTitle );
+  string const lineSection = GEOS_FMT( "{:#^{}}\n", "", m_sectionWidth );
```
MelReyCG: `lineSection` -> `bottomLine`?

src/coreComponents/fileIO/section/Section.hpp
```diff
@@ -0,0 +1,123 @@
...
+   * Descriptions values can be be any types and will be aligned
+   */
+  template< typename ... Args >
+  void addDescription( string_view descriptionName, Args const & ... args );
+
+  /**
+   * @brief Add a description to the section
+   * @param description The string value of the description
+   */
+  void addDescription( string_view description );
```
MelReyCG: To support PR #3303, do we need a `addEndDescription()` or something related?

src/coreComponents/functions/TableFunction.hpp
```diff
@@ -277,7 +278,9 @@ class TableFunction : public FunctionBase
    * @return The unit of a coordinate dimension, or units::Unknown if no units has been specified.
    */
   units::Unit getDimUnit( localIndex const dim ) const
-  { return size_t(dim) < m_dimUnits.size() ? m_dimUnits[dim] : units::Unknown; }
+  {
+    return size_t(dim) < m_dimUnits.size() ? m_dimUnits[dim] : units::Unknown;
+  }
```
MelReyCG: Please revert to the original oneliner here.

src/coreComponents/fileIO/section/Section.hpp
```diff
@@ -0,0 +1,123 @@
...
+ * Copyright (c) 2018-2020 TotalEnergies
+ * Copyright (c) 2019-     GEOSX Contributors
+ * All rights reserved
+ *
+ * See top level LICENSE, COPYRIGHT, CONTRIBUTORS, NOTICE, and ACKNOWLEDGEMENTS files for details.
+ * ------------------------------------------------------------------------------------------------------------
+ */
+
+/**
+ * @file Section.hpp
```
MelReyCG: As the package has been relocated, please move `Section.*pp` in `common/format`. You can directly drop your two files in there as it doesn't seem necessary to have a dedicated folder.

MelReyCG (changes requested):

paveltomin:
what is your definition of Section?
since GEOS does not have Section concept, it should be clearly defined, otherwise it's quite arbitrary how one can do the split

arng40:
> what is your definition of Section? since GEOS does not have Section concept, it should be clearly defined, otherwise it's quite arbitrary how one can do the split

it's to visually separate the different log step (GenerateMesh, inizialisation,... ), and to differentiate the timesteps and see the summary information at the end of each timestep.

paveltomin:
> > what is your definition of Section? since GEOS does not have Section concept, it should be clearly defined, otherwise it's quite arbitrary how one can do the split
> 
> it's to visually separate the different log step (GenerateMesh, inizialisation,... ), and to differentiate the timesteps and see the summary information at the end of each timestep.

i understand why it is needed, asking how do you define a section? just common sense?

arng40:
Yes, we define a section object in the section code, for example :
```
Section section( "Mesh generation" );
  section.beginSection();
  generateMesh();
  section.endSection();
```

paveltomin:
> Yes, we define a section object in the section code, for example :
> 
> ```
> Section section( "Mesh generation" );
>   section.beginSection();
>   generateMesh();
>   section.endSection();
> ```

I mean not code definition, but can you define is words what is section?

"Section is ...."

I assume it comes from ECLIPSE etc but there section is clearly defined from the input file, while in GEOS - what is section?

rrsettgast:
@arng40 
There is more here than just adding sections. Does this contain commits from another PR?

rrsettgast:
I don't think we should be calling these headers "Sections" as they do not correspond with the flow of the code. I think "Start/End" would be more flexible...and then perhaps some hierarchy of headers would be nice. So we can give more detailed indications of where we are in the flow of the code.

arng40:
@rrsettgast @paveltomin Myview is that Sections should separate the main initialization & destruction steps of the ProblemManager, and each simulation timesteps. I added the Sections I identified in the PR description.
Initially this PR was based on my adding Table PR, that's why there are lot of commit

paveltomin:
> @rrsettgast @paveltomin Myview is that Sections should separate the main initialization & destruction steps of the ProblemManager, and each simulation timesteps. I added the Sections I identified in the PR description. Initially this PR was based on my adding Table PR, that's why there are lot of commit

ok, that makes sense, thanks for clarifying

src/coreComponents/common/format/table/TableLayout.hpp
```diff
@@ -48,7 +48,7 @@ class TableLayout
   /**
    * @brief Enumeration for table sections.
    */
-  enum Section { header, values };
+  enum LogPart { header, values };
```
MelReyCG: I don't think you really would like to rename this enum (at least, not to LogPart).

src/coreComponents/fileIO/logPart/LogPart.hpp
```diff
@@ -0,0 +1,162 @@
...
+   * @brief Draw the first part of the section. It include the title and optionnaly, the description(s);
+   * @param os An output stream (by default, std::cout)
+   */
+  void begin( std::ostream & os = std::cout );
+
+  /**
+   * @brief Draw the last part of the section. It include the title
+   * @param oss An output stream (by default, std::cout)
+   */
+  void end( std::ostream & oss = std::cout ) const;
```
MelReyCG: Here, we seem to be able to `begin()` on multiple output targets. So, to ensure that the content is the same everywhere, I would expect `begin()` to be const, and internal values like `m_sectionWidth` to be computed when assessors like `addDescription()` are used.

_Side note: addEndDescription() has no effect on `m_sectionWidth`, is it the desired behavior?_

src/coreComponents/fileIO/logPart/LogPart.hpp
```diff
@@ -0,0 +1,162 @@
...
+  void setMinWidth( integer const & minWidth );
+
+  /**
+   * @brief Draw the first part of the section. It include the title and optionnaly, the description(s);
+   * @param os An output stream (by default, std::cout)
+   */
+  void begin( std::ostream & os = std::cout );
+
+  /**
+   * @brief Draw the last part of the section. It include the title
```
MelReyCG:
```suggestion
   * @brief Draw the last part of the section. It include the title and optionnaly, the end description(s).
```

src/coreComponents/fileIO/logPart/LogPart.hpp
```diff
@@ -0,0 +1,162 @@
...
+   * @brief Add a description to the end of the section
+   * @param description The string value of the description
+   */
+  void addEndDescription( string_view description );
+
+  /**
+   * @brief Set the minimal width of a row
+   * @param minWidth The minimal width of the table
+   */
+  void setMinWidth( integer const & minWidth );
```
MelReyCG: a good feature in the future would be to have a `setMaxWidth()` with line wrapping, to ensure that a section cannot be too large.

src/coreComponents/fileIO/logPart/LogPart.hpp
```diff
@@ -0,0 +1,162 @@
...
+  string m_footerTitle;
+  /// section length
+  integer m_sectionWidth;
+  /// min width of section length
+  integer m_rowMinWidth = 70;
+
+  /// description border margin
+  static constexpr integer m_marginBorder = 2;
+  /// numbers of character used as border
+  static constexpr integer m_nbBorderChar = 2;
```
MelReyCG: Please add the `#` character as a centralized constant here.

src/coreComponents/fileIO/logPart/LogPart.cpp
```diff
@@ -0,0 +1,121 @@
...
+#include <algorithm>
+
+namespace geos
+{
+
+LogPart::LogPart( string_view sectionTitle ):
+  m_sectionTitle( string( sectionTitle ) ),
+  m_sectionWidth( sectionTitle.length() )
+{
+  m_footerTitle = GEOS_FMT( "End : {}", m_sectionTitle );
```
MelReyCG: 1. If I was you, I would add a `updateLayout()` private method that takes in parameter every text size like this one, the description lines, the `m_sectionTitle`, and update `m_rowMinWidth` whenever a larger size is found. I would prevent any side case to produce oversized lines.
2. please use the term `part` as a prefix for your variables (if you really need one), `section` is now irrelevant.
3. What do you think of replacing `"End : {}"` by `"End of {}"`

src/coreComponents/fileIO/logPart/LogPart.cpp
```diff
@@ -0,0 +1,121 @@
...
+void LogPart::setMinWidth( integer const & minWidth )
+{
+  m_rowMinWidth = minWidth;
+}
+
+void LogPart::formatAndInsertDescriptions( std::vector< string > & descriptionContainer,
+                                           string_view name,
+                                           std::vector< string > const & values )
+{
+  string const nameFormatted =  GEOS_FMT( "- {}: ", string( name ));
```
MelReyCG: I think that it is the responsability of the class user to add those `"- "` at the start of their texts (imagine if you only have one description).

src/coreComponents/events/EventManager.cpp
```diff
@@ -170,8 +170,9 @@ bool EventManager::run( DomainPartition & domain )
       m_dt = dt_global;
 #endif
     }
-
-    outputTime();
+    LogPart logPart( "TIMESTEP START" );
```
MelReyCG: I would name this part `"TIMESTEP"` or `"Timestep no. {}"`. It is strange to see `"End : TIMESTEP START"` at its end.

MelReyCG (changes requested):

src/coreComponents/common/format/table/TableLayout.hpp
arng40: revet

src/coreComponents/fileIO/logPart/LogPart.cpp
arng40: if we have one description we will not have the `"- "`, it's only for mutiple values in the description

src/coreComponents/fileIO/logPart/LogPart.hpp
arng40: y

src/coreComponents/fileIO/logPart/LogPart.hpp
arng40: I could do this in another PR

src/coreComponents/fileIO/logPart/LogPart.hpp
arng40: after `addDescription()` we call void `formatAndInsertDescriptions()` where we check `m_sectionWidth`

MelReyCG:
Can you share a full log generated after this PR?

paveltomin (approved):
just check what happened with schema/docs changes

src/coreComponents/fileIO/logPart/LogPart.cpp
```diff
@@ -0,0 +1,157 @@
...
+  string const line = string( m_logPartWidth, m_borderCharacter );
+  string const topPart =  GEOS_FMT( "{}\n{}{}\n",
+                                    line,
+                                    buildTitlePart(),
+                                    line );
+  os << GEOS_FMT( "\n{}{}\n", topPart, bottomPart );
+  clear();
+}
+
+void LogPart::end( std::ostream & os )
```
MelReyCG: Please restore the "End of: " output, don't forget to take the full size of the end into account for sizing the width.

src/coreComponents/fileIO/logPart/LogPart.cpp
```diff
@@ -0,0 +1,157 @@
...
+    {
+      m_formattedDescriptions.push_back( m_descriptionNames[idxName] );
+    }
+    else
+    {
+      string_view name = m_descriptionNames[idxName];
+      string_view spaces = std::string( maxNameSize - name.size(), ' ' );
+      string_view nameFormatted = GEOS_FMT( "{}{} ", name, spaces );
+
+      m_formattedDescriptions.push_back( GEOS_FMT( "{}{}", nameFormatted, m_descriptionValues[idxName][0] ));
```
MelReyCG: I think it would less mask what you want to do.
```suggestion
      string_view spaces = std::string( maxNameSize - name.size(), ' ' );
      string_view nameFormatted = GEOS_FMT( "{}{}", name, spaces );

      m_formattedDescriptions.push_back( GEOS_FMT( "{} {}", nameFormatted, m_descriptionValues[idxName][0] ));
```

src/coreComponents/fileIO/logPart/LogPart.hpp
```diff
@@ -0,0 +1,140 @@
...
+   * @brief Construct a new LogPart
+   * @param m_logPartTitle The logPart title
+   */
+  LogPart( string_view m_logPartTitle );
+
+  /**
+   * @brief Add a description to the logPart by concatening a description name and descriptions values.
+   * @param name The description name
+   * @param args Descriptions values to be concatened.
+   * Descriptions values can be be any types and will be aligned
```
MelReyCG:
```suggestion
   * @note Descriptions values can be be any formatted types. Values will be aligned altogether.
```

src/coreComponents/fileIO/logPart/unitTests/testLogPart.cpp
```diff
@@ -0,0 +1,138 @@
...
+  logPart.setMinWidth( 70 );
+  logPart.begin( oss );
+  EXPECT_EQ ( oss.str(),
+              "\n######################################################################\n"
+              "##                          TIMESTEP START                          ##\n"
+              "######################################################################\n"
+              "##  - Time :       00h08m20s out of 2d, 21h26m40s (0% completed)    ##\n"
+              "##                 500 s / 250000 s                                 ##\n"
+              "##  - Delta Time : 00h16m40s (1000 s)                               ##\n"
+              "##  Description test                                                ##\n\n"
```
MelReyCG: shouldn't we have the following?
```suggestion
              "\n####################################################################\n"
              "##                          TIMESTEP START                        ##\n"
              "####################################################################\n"
              "##  - Time :       00h08m20s out of 2d, 21h26m40s (0% completed)  ##\n"
              "##                 500 s / 250000 s                               ##\n"
              "##  - Delta Time : 00h16m40s (1000 s)                             ##\n"
              "##  Description test                                              ##\n\n"
```

src/coreComponents/fileIO/logPart/LogPart.hpp
```diff
@@ -0,0 +1,140 @@
...
+  std::vector< std::vector< string > > m_descriptionValues;
+
+  /// title of logPart
+  string m_logPartTitle;
+  /// Title footer string
+  string m_footerTitle;
+  /// min width of logPart length
+  size_t m_rowMinWidth = 70;
+  /// logPart length
+  size_t m_logPartWidth;
```
MelReyCG:
```suggestion
  struct Description {
    /// Title footer string
    string m_title;

    /// Name of the description, formatted to be : [Name] : [Values1]\n[Values2]
    std::vector< string > m_descriptionNames;
    /// Values in the descrpption the description
    std::vector< std::vector< string > > m_descriptionValues;
    /// Vector containing the descriptions formatted by formatDescriptions()
    std::vector< string > m_formattedDescriptionLines;

    /// logPart length
    size_t m_logPartWidth;
    
    void formatDescriptions();
  };

  /// min width of logPart length
  size_t m_rowMinWidth = 70;

  Description startDesc;
  Description endDesc;
```

And so, you should be able to easily restore the `addEndDescription()` methods, which should be more user friendly from my point of view.

I'm not sure that `void formatDescriptions();` should be relocated here (and that its signature is complete), but if the code is well structured, it should be possible.

MelReyCG (changes requested):

src/coreComponents/fileIO/logPart/unitTests/testLogPart.cpp
arng40: because min width is set to 70

src/coreComponents/fileIO/logPart/LogPart.hpp
arng40: i've done a lot of refac with your proposal 👨‍🍳

src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
```diff
@@ -276,8 +277,10 @@ bool PhysicsSolverBase::execute( real64 const time_n,
   integer const maxSubSteps = m_nonlinearSolverParameters.m_maxSubSteps;
 
   // Keep track of substeps. It is useful to output these.
-  std::vector< real64 > subStepDt( maxSubSteps, 0.0 );
-  integer numOfSubSteps = 0;
+  std::vector< real64 > & subStepDt = getSubStepDt();
+  subStepDt.resize( maxSubSteps, 0.0 );
```
MelReyCG: As the number of substeps is variable over timesteps, the size of this list should not be constant.
And as `subStepDt` is a list of sub-step**s** delta-time, please add an `s` somewhere in the name (here, in the member variable, and in the methods evocating it).

You may have wanted to do a `reserve()` here, and use `push_back()`, this way the number of elements (`size()`) could serve as your "numOfSubSteps", and this one could be removed.

src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
```diff
@@ -288,7 +291,9 @@ bool PhysicsSolverBase::execute( real64 const time_n,
                                           nextDt,
                                           cycleNumber,
                                           domain );
-    numOfSubSteps++;
+    setNumOfSubSteps( subStep + 1 );
+    std::cout << " 1 substep " << subStep << std::endl;
```
MelReyCG: Please remove this debug log line.

MelReyCG (changes requested):

src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
```diff
@@ -288,8 +285,8 @@ bool PhysicsSolverBase::execute( real64 const time_n,
                                           nextDt,
                                           cycleNumber,
                                           domain );
-    numOfSubSteps++;
-    subStepDt[subStep] = dtAccepted;
+
+    getSubStepDts().push_back( dtAccepted );
```
MelReyCG: Where is it cleared to reinit the dts of the current timestep?

MelReyCG (approved):

arng40:
do not merge : display not good when using mpi

src/coreComponents/common/format/LogPart.hpp
```diff
@@ -0,0 +1,217 @@
...
+  LogPart( string_view m_logPartTitle, int rank = -1 );
+
+  /**
+   * @brief Add a description to the opening logPart by concatening a description name and descriptions values.
+   * @param name The description name
+   * @param args Descriptions values to be concatened.
+   * @note Descriptions values can be be any formatted types. Values will be aligned altogether.
+   */
+  template< typename ... Args >
+  void addDescription( string const & name, Args const & ... args );
```
MelReyCG: why not string view?

src/coreComponents/common/format/LogPart.hpp
```diff
@@ -0,0 +1,217 @@
...
+   * @brief Draw the last part of the logPart. It include the title and optionnaly the end description(s).
+   * @param oss An output stream (by default, std::cout)
+   */
+  void end( std::ostream & oss = std::cout );
+
+private:
+
+  struct Description
+  {
+    /// Title footer string
```
MelReyCG: ???

src/coreComponents/common/format/LogPart.hpp
```diff
@@ -0,0 +1,217 @@
...
+
+  /**
+   * @brief Draw the last part of the logPart. It include the title and optionnaly the end description(s).
+   * @param oss An output stream (by default, std::cout)
+   */
+  void end( std::ostream & oss = std::cout );
+
+private:
+
+  struct Description
```
MelReyCG, line 111:
```suggestion
  /**
   * @brief top or lower description
   */
  struct Description
```

src/coreComponents/common/format/LogPart.hpp
```diff
@@ -0,0 +1,217 @@
...
+
+    /// logPart length
+    size_t m_logPartWidth;
+    /// logPart max length
+    size_t m_logPartMaxWidth;
+    /// logPart min length
+    size_t m_logPartMinWidth;
+    /// max length name of a logPart
+    size_t m_logPartMaxNameWidth;
+  };
```
MelReyCG: I suggest the following:
```c++
  struct Description
  {
    /// Name of the description, it can be splited by \\n
    std::vector< std::vector< string > > m_names;
    /// Values in the description, each value of a description is associated with one name
    std::vector< std::vector< string > > m_values;

    /// logPart max length
    size_t m_logPartMaxWidth;
    /// logPart min length
    size_t m_logPartMinWidth;
  };

  struct FormattedDescription
  {
    /// Vector containing the descriptions formatted by formatDescriptions()
    std::vector< string > m_formattedDescriptionLines; // computed
    /// logPart length
    size_t m_logPartWidth; // computed
    /// max length name of a logPart
    size_t m_logPartMaxNameWidth; // computed
  };

  /// Title footer string
  string m_title;
```
And regroup the other variables at top, just bellow this `m_title`.

This way you would have the classic picture: settings -> formatting -> output
```c++
  FormattedDescription formatDescriptions( Description const & description )
  
  //replacements of "buildX" methods
  void outputTitle( std::ostringstream & oss, string_view title, FormattedDescription const & format )
  void outputDescription( std::ostringstream & oss, FormattedDescription const & format )
```

src/coreComponents/common/format/LogPart.hpp
```diff
@@ -0,0 +1,217 @@
...
+    std::vector< std::vector< string > > m_values;
+    /// Vector containing the descriptions formatted by formatDescriptions()
+    std::vector< string > m_formattedDescriptionLines;
+
+    /// logPart length
+    size_t m_logPartWidth;
+    /// logPart max length
+    size_t m_logPartMaxWidth;
+    /// logPart min length
+    size_t m_logPartMinWidth;
```
MelReyCG: Shouldn't they be the same for top and bottom?

src/coreComponents/common/format/LogPart.hpp
```diff
@@ -0,0 +1,227 @@
...
+    size_t m_width;
+    /// max length name (first part of a description) of a logPart
+    size_t m_maxNameWidth;
+  };
+
+  Description m_startDescription = { "", {}, {}, SIZE_MAX, 50};
+  Description m_endDescription  = { "", {}, {}, SIZE_MAX, 50 };
+
+  FormattedDescription m_formattedStartDescription = { {}, 50, 0 };
+  FormattedDescription m_formattedEndDescription  = { {}, 50, 0 };
```
MelReyCG: It is not mandatory to keep those as members variables, could be local in each function.

src/coreComponents/common/format/LogPart.hpp
MelReyCG: I don't understand, did you changed anything?

arng40: yep i've put all above in description struct except for logPartWidth

src/coreComponents/common/format/LogPart.hpp
arng40: I prefer this way, it gives me an overview, I could change it if it's a bad practice

src/coreComponents/common/format/StringUtilities.hpp
```diff
@@ -254,6 +254,15 @@ string addCommaSeparators( T const & num );
...
 std::vector< STRING_T > divideLines( size_t & linesWidth, string_view value );
 
+/**
+ * @brief Divides a string by newline characters and returns a vector of strings containing each line.
+ * @param value The input string to divide into lines
+ * @tparam STRING_T The type of the string (string or string_view)
+ * @return A vector of STRING_T objects, each containing a single line from the input
+ */
+template< typename STRING_T >
+std::vector< STRING_T > divideLines( string_view value );
```
MelReyCG: Is this new signature of `divideLines()` useful? I don't see where it is used & if it really adds to the existing one.

arng40: ups mb, I miss this on merge conflict

MelReyCG (approved):

arng40:
smw I'm getting this kind of error in integrated test :
`Scalar values of types int32 and int32 differ: 4, 6`

corbett5 (approved):

src/coreComponents/events/EventManager.cpp
```diff
@@ -297,4 +308,30 @@ void EventManager::outputTime() const
...
+    {
+      logMessage << ", ";
+    }
+    logMessage << subStepDts[i] << " " << units::getSymbol( units::Unit::Time );
+  }
+
+  if( logMessage.rdbuf()->in_avail() == 0 )
+    logMessage << "/";
+
+  logpart.addEndDescription( "- dt: ", logMessage.str() );
```
MelReyCG:
```suggestion
  logpart.addEndDescription( "- substep dts: ", logMessage.str() );
```

src/coreComponents/dataRepository/ExecutableGroup.hpp
```diff
@@ -119,8 +119,23 @@ class ExecutableGroup : public dataRepository::Group
...
+  /**
+   * @brief Get the track of substeps for each substep
+   * @return A const vector containing the time step size for each substep
+   */
+  std::vector< real64 > & getSubStepDts() { return m_subStepDts; }
+
 private:
 
+  /// Keep track of substeps
+  std::vector< real64 > m_subStepDts = {};
```
MelReyCG: Does this class miss a virtual destructor?

src/coreComponents/events/EventManager.cpp
```diff
@@ -196,6 +199,13 @@ bool EventManager::run( DomainPartition & domain )
       else if( subEvent->isReadyForExec() )
       {
         earlyReturn = subEvent->execute( m_time, m_dt, m_cycle, 0, 0, domain );
+
+        if( subEvent->getEventTarget()->getTimesteppingBehavior() == ExecutableGroup::TimesteppingBehavior::DeterminesTimeStepSize )
```
MelReyCG: In some cases, `getEventTarget()` can return a `nullptr` isn't it?

arng40: indeed

CusiniM:
Here is a full log example. 

[logExample.log](https://github.com/user-attachments/files/20211904/logExample.log)

CusiniM (approved):

CusiniM:
IMO we should remove lines like

`Time: 1.90e+03 s, dt: 100 s, Cycle: 19`

though

CusiniM:
and it would be nice if all sections begin and end markers were the same size and large enough for all log messages to fit within them.

rrsettgast:
l

> IMO we should remove lines like
> 
> `Time: 1.90e+03 s, dt: 100 s, Cycle: 19`
> 
> though

This is the output for an explicit time integrator.


@arng40 I think that these types of things might be better encapsulated in the `PhysicsSolver` rather than the `EventBase` level.

rrsettgast (changes requested):
There needs to be an option for minimal output when using explicit time integration.

arng40:
> @arng40 I think that these types of things might be better encapsulated in the `physicsSolver` rather than the `EventBase` level.

Is the responsability of `physicsSolver` to format substeps information ? Is there any other tasks/events which could have substeps logics ?
If not, I think we can pass the logParts et let the  `physicsSolver` fill it.
Our idea was to format all the timestep at the `events` level.

arng40:
> and it would be nice if all sections begin and end markers were the same size and large enough for all log messages to fit within them.

Actually we have a flexible with max size for the logPart. the log can potentially have very long lines which can not be predicted, if the user's terminal isn't wide enough and word-wrapping is enabled, then it doesn't matter for very long lines, but it's critical for `logParts` formatting. 
```
##############################################
##################################
##                                  TIMESTEP  
                                ##
##############################################
##################################
##  - Time        : 01h23m20s out of 01h31m40s
 (91% completed)                ##
##                  5000 s / 5500 s           
                                ##
##  - Delta Time  : 00h08m20s (500 s)         
                                ##
##  - Cycle       : 10                        
                                ##
```

rrsettgast (approved):
@rrsettgast  make an issue for moving data to a more appropriate location...discussion on where that is if it does need to be moved.

CusiniM:
> > and it would be nice if all sections begin and end markers were the same size and large enough for all log messages to fit within them.
> 
> Actually we have a flexible with max size for the logPart. the log can potentially have very long lines which can not be predicted, if the user's terminal isn't wide enough and word-wrapping is enabled, then it doesn't matter for very long lines, but it's critical for `logParts` formatting.
> 
> ```
> ##############################################
> ##################################
> ##                                  TIMESTEP  
>                                 ##
> ##############################################
> ##################################
> ##  - Time        : 01h23m20s out of 01h31m40s
>  (91% completed)                ##
> ##                  5000 s / 5500 s           
>                                 ##
> ##  - Delta Time  : 00h08m20s (500 s)         
>                                 ##
> ##  - Cycle       : 10                        
>                                 ##
> ```

can we set a reasonable max size and enforce it? We also don't need to do this before merging this PR.

paveltomin:
@arng40 is it good to go back into the merge queue?

arng40:
> @rrsettgast make an issue for moving data to a more appropriate location...discussion on where that is if it does need to be moved.

I put it back the way it was before
Good to go in MQ

arng40:
I also update the PR description with the new logPart default size

PR #3020, opened 2024-03-04
Fixes spikes when importing fractures in parallel

TotoGaz:
2d elements are pointing to information on their neighbor 3d elements.
Before the ghosting process, there could be situations where all the information needed by the 2d elements has not arrived yet.
This PR uses the ghosted information to better fill all the mappings. Especially, some great care is taken about the collocated nodes and edges.

This PR was validated on several cases that previously had issues with the ghosting of fractures.

However, the final code is a bit convoluted and does not make me believe it will be rock solid.
Still, I guess it's an improvement. 🤷


TotoGaz:
@CusiniM I think this is ready for a first review.

TotoGaz:
@paveltomin Can you test my PR please? I've been able to validate cases that were not working previously with it.

paveltomin:
> @paveltomin Can you test my PR please? I've been able to validate cases that were not working previously with it.

Thank you!
Just did quick tests and everything looks good. Let's merge it.

paveltomin:
After your changes, there will be no elements without nodes, right?
No need for trick like this:
image
?

TotoGaz:
> After your changes, there will be no elements without nodes, right? No need for trick like this: ...

I did not focus on this, but no, such a case should not happen anymore 🤞... until next time 🤣

paveltomin:
> > After your changes, there will be no elements without nodes, right? No need for trick like this: ...
> 
> I did not focus on this, but no, such a case should not happen anymore 🤞... until next time 🤣

Good, can you put a check to make

paveltomin (approved):

TotoGaz:
The integrated tests give
```
FAILED:  2 python3(impermeableFault_smoke_01_2_restartcheck), python3(permeableFault_smoke_01_2_restartcheck)
TIMEOUT:  1 geosx(pennyShapedViscosityDominated_smoke_01_1_geos)
```
Which I think is fair.

src/coreComponents/mesh/FaceElementSubRegion.cpp
```diff
@@ -474,30 +474,45 @@ std::map< globalIndex, globalIndex > buildReferenceCollocatedNodes( ArrayOfArray
...
 
 /**
- * @brief Returns a mapping that links any collocated edge to the collocated edge with the lowest index.
- * @param referenceCollocatedNodes The mapping that links any collocated node to its collocated node with the lowest index.
- * @param nl2g The local to glocal mapping for nodes.
- * @param edgeToNodes The edge to nodes mapping.
- * @param elem2dToEdges The 2d elem to edges mapping.
- * @param edgeGhostRanks The ghost rank of the edges.
- * @return The computed map.
+ * @brief Computes the mapping with links a pair of collocated nodes
```
CusiniM:
```suggestion
 * @brief Computes the mapping which links a pair of collocated nodes
```

src/coreComponents/mesh/FaceElementSubRegion.cpp
```diff
@@ -474,30 +474,45 @@ std::map< globalIndex, globalIndex > buildReferenceCollocatedNodes( ArrayOfArray
...
+ * to all the edges having their nodes collocated to this key pair of nodes.
+ * @param referenceCollocatedNodes Mapping that link all the collocated nodes
+ * to the collocated node with the lowest index (considered as a reference).
+ * The reference node can then be used to make connection between geometrical objects
+ * relying on collocated noes (in the case of this function, edges).
```
CusiniM:
```suggestion
 * relying on collocated nodes (in the case of this function, edges).
```

src/coreComponents/mesh/FaceElementSubRegion.cpp
```diff
@@ -474,30 +474,45 @@ std::map< globalIndex, globalIndex > buildReferenceCollocatedNodes( ArrayOfArray
...
+  // It's important to note that the key of `edgeIds` are the global indices of the nodes, in no particular order.
   std::map< std::pair< globalIndex, globalIndex >, localIndex > edgesIds;
-  for( int ei = 0; ei < elem2dToEdges.size(); ++ei )
+  for( localIndex lei = 0; lei < edgeToNodes.size( 0 ); ++lei )
   {
-    for( localIndex const & edi: elem2dToEdges[ei] )
+    auto const & nodes = edgeToNodes[lei];
+    globalIndex const & gni0 = nl2g[nodes[0]];
+    globalIndex const & gni1 = nl2g[nodes[1]];
+    if( hasCollocatedNode( gni0 ) && hasCollocatedNode( gni1 ) )
```
CusiniM, line 516: seems that you used to got through the faceElements (or 2dElements for this) but now you do it relying directly on the edgeToNodes map?

src/coreComponents/mesh/FaceElementSubRegion.cpp
```diff
@@ -568,27 +597,31 @@ SortedArray< localIndex > makeSortedArrayIota( localIndex newSize, localIndex va
 
 /**
  * @brief Builds the 2d face to 2d elements mapping.
- * @param num2dFaces The number of 2d faces.
- * @param num2dElems The number of 2d elements.
  * @param elem2dToEdges The 2d element (geometrical faces in 3d) to edges mapping.
  * @param edgesTo2dFaces The edges to 2d faces (geometrical edges in 3d) mapping.
```
CusiniM, line 604: `edgesTo2dFaces` here the edges are the 3D edges right?

src/coreComponents/mesh/FaceElementSubRegion.cpp
```diff
@@ -598,64 +631,244 @@ ArrayOfArrays< geos::localIndex > build2dFaceTo2dElems( std::size_t num2dFaces,
...
 {
-  arrayView1d< globalIndex const > const nl2g = nodeManager.localToGlobalMap();
+  auto const num2dElems = elem2dToNodes.size();
 
-  // First let's create the reference mappings for both nodes and edges.
-  std::map< globalIndex, globalIndex > const referenceCollocatedNodes = buildReferenceCollocatedNodes( m_2dElemToCollocatedNodesBuckets );
-  std::map< localIndex, localIndex > const referenceCollocatedEdges =
-    buildReferenceCollocatedEdges( referenceCollocatedNodes, nl2g, edgeManager.nodeList(), m_toEdgesRelation.toViewConst(), edgeManager.ghostRank().toViewConst() );
+  // We loop over all the collocated nodes attached to the 2d elements.
+  // If a node is on the rank while not attached to the 2d element, then we add a connection.
```
CusiniM, line 668: so this happens after ghosting? I don't think it's clear in the description.

src/coreComponents/mesh/MeshLevel.cpp
```diff
@@ -348,7 +348,10 @@ void MeshLevel::generateAdjacencyLists( arrayView1d< localIndex const > const &
       }
       for( localIndex const & edi: elem2dToEdges[ei] )
       {
-        edgeAdjacencySet.insert( edi );
+        if( edi > -1 )
```
CusiniM: what's the meaning of -1?

src/coreComponents/mesh/FaceElementSubRegion.cpp
```diff
@@ -598,64 +631,244 @@ ArrayOfArrays< geos::localIndex > build2dFaceTo2dElems( std::size_t num2dFaces,
...
+}
+
+
+/**
+ * @brief Uses the two input mappings to reorder the @p elem2dToNodes mapping.
+ * @param[in] elem2dToFaces The 2d element to faces mapping.
+ * @param[in] facesToNodes The face to nodes mapping.
+ * @param[out] elem2dToNodes The 2d element to nodes that will be overwritten.
+ * @details The @p elem2dToNodes a priori does not respect any specific nodes order.
+ * But the vtk output await the first part of the nodes to make a first face (in the correct order),
```
CusiniM:
```suggestion
 * But the vtk output expects the first half of the nodes to make a first face (in the correct order),
```

CusiniM (approved):

src/coreComponents/mesh/FaceElementSubRegion.cpp
TotoGaz: Yes, every where in the code the quantity is assumed 3d (unless tagged 2d).

src/coreComponents/mesh/FaceElementSubRegion.cpp
TotoGaz: Yes. I have added some comments in the doxygen.

src/coreComponents/mesh/FaceElementSubRegion.cpp
TotoGaz: I've improved the comments.

src/coreComponents/mesh/MeshLevel.cpp
TotoGaz: That was indeed not needed. I've removed that.

castelletto1:
Prioritizing this PR in the merge queue since new baselines are not needed.

PR #3033, opened 2024-03-08
Remove wrong zero diag check

### LINKED ISSUE #3031: CompositionalMultiphaseFVM crash in Debug build
When running simulation with a `CompositionalMultiphaseFVM` with a Debug build of GEOS, this assertion keeps failing : 
_src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseBaseKernels.hpp:789_
```c++
      GEOS_ASSERT_MSG( LvArray::math::abs( stack.localJacobian[ic][ic] ) > minDensForDivision,
                       GEOS_FMT( "Zero diagonal in Jacobian: equation {}, value = {}", ic, stack.localJacobian[ic][ic] ) );
```

I noticed this issue by creating the unit test of [one of my PR](https://github.com/GEOS-DEV/GEOS/pull/2954), I also tried with the xml input in `testComMultiphaseFlow.cpp`.

I also tried with these settings (and it keeps failing) :
- LAI : hypre, trilinos
- `solverType` : gmres, fgmres, direct
- `preconditioner` : iluk, none, jacobi., amg, mgr..
- Fluid types : Ezroki fluid, Philips fluid, with different PVT tables
- varying pressure and composition fraction in the reservoir.

paveltomin:
Remove wrong zero diag check.
It is always triggered for equation number 1 because of equations/unknowns layout:
```
       p     rho_0      rho_1
eq 0   x       x          0
eq 1   x       0          x
               ^ this is always zero
vol    x       x          x
```

Thanks to @MelReyCG for catching this.


src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseBaseKernels.hpp
```diff
@@ -782,13 +782,6 @@ class ElementBasedAssemblyKernel
 
       }
     }
-
-    // check zero diagonal (works only in debug)
-    for( integer ic = 0; ic < numComp; ++ic )
-    {
-      GEOS_ASSERT_MSG( LvArray::math::abs( stack.localJacobian[ic][ic] ) > minDensForDivision,
-                       GEOS_FMT( "Zero diagonal in Jacobian: equation {}, value = {}", ic, stack.localJacobian[ic][ic] ) );
-    }
```
dkachuma, line 791: Would we want to check `stack.localJacobian[ic][ic+1]` instead?

paveltomin: I am not sure...
When I originally added the check, my main concern was the liner solver, but now I don't know.
Maybe @castelletto1 and @victorapm can clarify what should be checked.

MelReyCG (approved):
I'm approving this PR because I know it solves the current problem with `CompositionalMultiphaseFVM` in debug build. I think that adding back a proper issue here should be done in a separated issue & PR.

src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseBaseKernels.hpp
victorapm: Zero diagonals can be an issue for the robustness of ILU0, which is used global relaxation in the level where the densities are eliminated.

Would it be possible to change the order of the equations such that we don't have zero diagonals?

paveltomin: By default GEOS already doing that, if I remember correctly - first equation will be replaced by sum of all equations and all other equations are shifted - old first becomes new second and so on.
After that:
```
         p      rho_0      rho_1
eq 0   x0+x1    x0+x1        x1
eq 1     x0       x0         0
vol      x        x          x
```
So no zero diag anymore.

PR #3039, opened 2024-03-14
ci: Spack docker containers

bmhan12:
This PR attempts to run Github Actions using Docker images with dependencies built with uberenv, spack.

Related to GEOS-DEV/thirdPartyLibs#261
Relates to https://github.com/GEOS-DEV/LvArray/pull/318


untereiner:
Hi @bmhan12, 
Is it possible to replace the TPL_TAG stored in a hidden file inside this repo by a stable tag name inside the CI scripts ? Let's say "latest" ?
Then each container configuration can be updated without the need of a commit into GEOS to update the tag.

It is also possible to tag each version of the containers with the current tags in dockerhub to pull them if needed.

bmhan12:
> Is it possible to replace the TPL_TAG stored in a hidden file inside this repo by a stable tag name inside the CI scripts ? Let's say "latest" ?

It is probably possible, but we would need to be careful in choosing which and when images get to be "latest".

untereiner:
Does it make sens to say that geos develop branch should be in sync with tpl develop branch (latest commit). But when doing a geos release a corresponding release has to be done on the tpl side ?

wrtobin (approved):

scripts/ci_build_and_test_in_container.sh
```diff
@@ -71,19 +77,22 @@ exit 1
...
+args=$(or_die getopt -a -o h --long build-exe-only,cmake-build-type:,code-coverage,data-basename:,enable-hypre:,enable-hypre-device:,enable-trilinos:,exchange-dir:,host-config:,install-dir-basename:,makefile,ninja,no-install-schema,no-run-unit-tests,nproc:,repository:,run-integrated-tests,sccache-credentials:,test-code-style,test-documentation,help -- "$@")
 
 # Variables with default values
 BUILD_EXE_ONLY=false
 BUILD_GENERATOR=""
 GEOS_INSTALL_SCHEMA=true
 HOST_CONFIG="host-configs/environment.cmake"
+ENABLE_HYPRE=OFF
+ENABLE_HYPRE_DEVICE=CPU
+GEOS_LA_INTERFACE=Trilinos
```
CusiniM: I think the default should be Hypre.

scripts/ci_build_and_test_in_container.sh
```diff
@@ -71,19 +77,22 @@ exit 1
...
 RUN_UNIT_TESTS=true
 RUN_INTEGRATED_TESTS=false
 UPLOAD_TEST_BASELINES=false
 TEST_CODE_STYLE=false
 TEST_DOCUMENTATION=false
+ENABLE_TRILINOS=ON
```
CusiniM: I would set this to OFF by default and turn Hypre on.

scripts/ci_build_and_test_in_container.sh
```diff
@@ -233,6 +249,10 @@ or_die python3 scripts/config-build.py \
                ${BUILD_GENERATOR} \
                -DBLT_MPI_COMMAND_APPEND='"--allow-run-as-root;--oversubscribe"' \
                -DGEOS_INSTALL_SCHEMA=${GEOS_INSTALL_SCHEMA} \
+               -DENABLE_HYPRE=${ENABLE_HYPRE} \
+               -DENABLE_HYPRE_DEVICE=${ENABLE_HYPRE_DEVICE} \
+               -DENABLE_TRILINOS=${ENABLE_TRILINOS} \
+               -DGEOS_LA_INTERFACE:PATH=${GEOS_LA_INTERFACE} \
```
CusiniM, line 257: why the need to specify both `ENABLE_TRILINOS` and `GEOS_LA_INTERFACE`?

CusiniM (approved):

scripts/ci_build_and_test_in_container.sh
bmhan12: Done!

scripts/ci_build_and_test_in_container.sh
bmhan12: Done!

scripts/ci_build_and_test_in_container.sh
bmhan12: > why the need to specify both ENABLE_TRILINOS and GEOS_LA_INTERFACE?

The current way the linear algebra packages are configured in the CI jobs is by setting the `ENABLE_HYPRE` and `ENABLE_TRILINOS` variables as environment variables:
https://github.com/GEOS-DEV/GEOS/blob/bd709034a07eb9cb4192cc7bbc420da30e147a24/.github/workflows/build_and_test.yml#L199-L207

Then reading those environment variables into CMake and choosing a `GEOS_LA_INTERFACE` from there:
https://github.com/GEOS-DEV/GEOS/blob/bd709034a07eb9cb4192cc7bbc420da30e147a24/host-configs/environment.cmake#L16-L43

Instead of reading in environment variables from `environment.cmake`, the spack Docker containers each generate a host-config (`spack-generated.cmake`) to use instead, where the `GEOS_LA_INTERFACE` is by default set to Hypre. The snippet added here is for overriding the spack host-config defaults by defining those variables through CMake (passing them to `config-build.py`) instead of reading them through environment variables. This is done for CI jobs that have `GEOS_LA_INTERFACE` set to Trilinos.

src/docs/sphinx/buildGuide/SpackUberenv.rst
```diff
@@ -3,25 +3,98 @@
...
+GEOS is transitioning to a new `Uberenv <https://github.com/LLNL/uberenv>`_ and `Spack <https://github.com/spack/spack/>`_ system for building our dependencies. We refer the reader to the `Uberenv documentation <https://uberenv.readthedocs.io/en/latest/>`_ and `Spack documentation <https://spack.readthedocs.io/en/latest/index.html>`_, in particular the Spack documentation sections worth reading are:
 
-Building the dependencies can be as simple as running
+* `Specs and dependencies <https://spack.readthedocs.io/en/latest/basic_usage.html#specs-dependencies>`_
+* `Virtual dependencies <https://spack.readthedocs.io/en/latest/basic_usage.html#virtual-dependencies>`_
+* `Manual compiler configuration <https://spack.readthedocs.io/en/latest/getting_started.html?highlight=compilers.yaml#manual-compiler-configuration>`_
+* `External packages <https://spack.readthedocs.io/en/latest/packages_yaml.html#external-packages>`_
+
+Building the dependencies can be as simple as running:
```
CusiniM, line 13: I would remove this part about this generic command and start by saying that a `spack-env` file must be specified.

rrsettgast (approved):

PR #3040, opened 2024-03-14
Modifying rst table type names

cssherman:


scripts/SchemaToRSTDocumentation.py
```diff
@@ -191,6 +191,10 @@ def buildTableValues(type_map, link_string='XML', include_defaults=True):
                 if k in type_map[att_name]:
                     table_row[jj] = type_map[att_name][k]
 
+                    # Fix type strings
+                    if ('Type' in k):
+                        table_row[jj] = table_row[jj].replace('_lt_', '<').replace('_gt_', '>').replace('_cm_', ',')
```
CusiniM: I also added a '-' to replace spaces so I think we need to fix those too.

CusiniM (approved):

CusiniM (approved):
Cool, thanks!

CusiniM:
I m putting this in the merge queue but we will have to update all the rst tables

PR #3049, opened 2024-03-20
Unify convergence check for lagrange-contact solvers

paveltomin:
Main thing: unify convergence check for lagrange-contact solvers
- do not assign max force to zero in `SinglePhasePoromechanicsConformingFractures::assembleElementBasedContributions` to ensure the same normalization across the solvers and FIM/SEQ
- do not compute `Rdisplacement` in `SolidMechanicsLagrangeContact::calculateContactResidualNorm` and rely directly on equivalent computed by `SolidMechanicsLagrangianFEM::calculateResidualNorm`
- keep the same (relative) logic for `Rtraction` for now but remove confusing `Rtotal`

Side thing: add derivative
```
dRdT( i, 0 ) = Ja * traction[kfe][i] * dLimitTau_dNormalTraction / vauxNorm;
```
in `SolidMechanicsLagrangeContact::assembleTractionResidualDerivativeWrtDisplacementAndTraction`, it seems missing and once added improves convergence for some cases.


src/coreComponents/physicsSolvers/contact/SolidMechanicsLagrangeContact.cpp
```diff
@@ -1289,10 +1236,9 @@ void SolidMechanicsLagrangeContact::
                   {
                     for( localIndex i = 1; i < 3; ++i )
                     {
-                      elemRHS[i] = +Ja * ( traction[kfe][i] - limitTau * vaux[ i-1 ] / vauxNorm );
-                    }
-                    for( localIndex i = 1; i < 3; ++i )
-                    {
+                      elemRHS[i] = Ja * traction[kfe][i] * ( 1.0 - limitTau / vauxNorm );
+
+                      dRdT( i, 0 ) = Ja * traction[kfe][i] * dLimitTau_dNormalTraction / vauxNorm;
```
paveltomin, line 1240: @castelletto1 could you please check this added derivative?

jhuang2601:
In `SolidMechanicsLagrangianFEM::calculateResidualNorm`, `Rdisplacement` is actually computed. I agree to remove `Rtotal`, which is a bit confusing; but suggest to keep both `Rdisplacement` and `Rtraction` for `SolidMechanicsLagrangeContact::calculateContactResidualNorm` to check the convergence for different primary variables.

paveltomin:
> In `SolidMechanicsLagrangianFEM::calculateResidualNorm`, `Rdisplacement` is actually computed. I agree to remove `Rtotal`, which is a bit confusing; but suggest to keep both `Rdisplacement` and `Rtraction` for `SolidMechanicsLagrangeContact::calculateContactResidualNorm` to check the convergence for different primary variables.

My understanding is that `Rdisplacement` is the same thing as `Rsolid` (up to normalization). Both are checking momentum equation rhs, L2 norm.

src/coreComponents/physicsSolvers/contact/SolidMechanicsLagrangeContact.cpp
```diff
@@ -1314,11 +1258,8 @@ void SolidMechanicsLagrangeContact::
               {
                 for( localIndex i = 0; i < 3; ++i )
```
CusiniM:
```suggestion
                for( int i = 0; i < 3; ++i )
```

src/coreComponents/physicsSolvers/contact/SolidMechanicsLagrangeContact.cpp
```diff
@@ -1301,9 +1247,7 @@ void SolidMechanicsLagrangeContact::
                     for( localIndex i = 1; i < 3; ++i )
```
CusiniM:
```suggestion
                    for( int i = 1; i < 3; ++i )
```

src/coreComponents/physicsSolvers/contact/SolidMechanicsLagrangeContact.cpp
CusiniM: it looks correct to me.

CusiniM (approved):

PR #3050, opened 2024-03-20
Spatially-varying grain bulk modulus within a region.

CusiniM:
In this PR:

- [x] Make grainBulkModulus a property varying cell by cell 

If an heterogeneous distribution of the Biot coefficient is needed it should be easy enough to obtain it by specifying an heterogeneous grain bulk modulus. 

I think there would be more value in modifying the code so that the Biot coefficient can be treated as an actual tensor as it should be.


joshua-white:
Yeah, personally I think we should have the user just input (1) a spatially-varying grain modulus and (2) a spatially-varying solid model.  We can then compute the Biot tensor internally using equations (26) and (31) of the paper referenced above. We could probably just add a small function to solidModel to return it given the grain modulus for the cell.  This would guarantee consistency, and wouldn't require the user to input an "extra" parameter that we would need to check for consistency internally anyway.  It would also handle the case of plasticity, anisotropy, etc. automatically.

Guotong-Ren:
This is my personal opinion, if we have item 1, item 2 is more-or-less redundant. Import of Biot from input is only to complement the usability.

paveltomin (approved):

CusiniM:
> In this PR:
> 
>     * [x]  Make grainBulkModulus a property varying cell by cell
> 
>     * [x]  Remove dependency of biotCoefficient on the grainBulkModulus and force it to be a user-defined parameter.
> 
> 
> Personally, I don't particularly like the second point. Quoting [this work](https://www.sciencedirect.com/science/article/pii/S0045782520304102): "The Biot tensor b is a function of Ce (drained elastic tangent moduli tensor of the solid skeleton) and Ks (bulk modulus of the solid mineral), indicating the portion of elastic deformation taken by the pore space out of the total elastic strain of the solid skeleton, which explains why b is paired to de." Thus, I think that the Biot coefficient (actually the Biot tensor) has a strong link to the choice of the solid model and it should be computed based on that. Thus, specifying it will force us to check for consistency.
> 
> Additionally, if we start introducing dependencies of the mechanical moduli on the state variables then, we will have to also update the Biot tensor accordingly which comes a lot more natural if it is not fixed.
> 
> If an heterogeneous distribution of the Biot coefficient is needed it should be easy enough to obtain it by specifying an heterogeneous grain bulk modulus.
> 
> I think there would be more value in modifying the code so that the Biot coefficient can be treated as an actual tensor as it should be.

I have reverted the changes related to the second point.

.integrated_tests.yaml
```diff
@@ -1,7 +1,7 @@
 ---
 baselines:
   bucket: geosx
-  baseline: integratedTests/baseline_integratedTests-pr3125-5101-7764ffb
+  baseline: integratedTests/baseline_integratedTests-pr3050-5125-d8a56bc
```
```suggestion
  baseline: integratedTests/baseline_integratedTests-pr3125-5101-7764ffb
```
```suggestion
  baseline: integratedTests/baseline_integratedTests-pr3050-5125-d8a56bc
```

src/coreComponents/constitutive/solid/porosity/BiotPorosity.cpp
```diff
@@ -46,8 +48,12 @@ BiotPorosity::BiotPorosity( string const & name, Group * const parent ):
     setDescription( "Flag enabling uniaxial approximation in fixed stress update" );
 
   registerField( fields::porosity::biotCoefficient{}, &m_biotCoefficient ).
-    setApplyDefaultValue( 1.0 ); // this is useful for sequential simulations, for the first flow solve
-  // ultimately, we want to be able to load the biotCoefficient from input directly, and this won't be necessary anymore
+    setApplyDefaultValue( -1.0 ).
```
CusiniM:
```suggestion
    setApplyDefaultValue( 1.0 ).
```

.integrated_tests.yaml
```diff
@@ -1,7 +1,6 @@
----
 baselines:
-  bucket: geosx
-  baseline: integratedTests/baseline_integratedTests-pr3125-5101-7764ffb
+  bucket: https://storage.googleapis.com/geosx/geosx
```
```suggestion
  bucket: https://storage.googleapis.com/geosx
```

.integrated_tests.yaml
```diff
@@ -1,7 +1,6 @@
----
 baselines:
-  bucket: geosx
-  baseline: integratedTests/baseline_integratedTests-pr3125-5101-7764ffb
+  bucket: https://storage.googleapis.com/geosx
```
```suggestion
  bucket: geosx
```

.integrated_tests.yaml
```diff
@@ -1,7 +1,6 @@
...
+  baseline: integratedTests/baseline_integratedTests-pr3125-5101-7764ffb.tar.gz
```
```suggestion
  baseline: integratedTests/baseline_integratedTests-pr3125-5101-7764ffb
```

jhuang2601 (approved):
Thanks for this necessary update.

BASELINE_NOTES.md
```diff
@@ -6,6 +6,11 @@ This file is designed to track changes to the integrated test baselines.
 Any developer who updates the baseline ID in the .integrated_tests.yaml file is expected to create an entry in this file with the pull request number, date, and their justification for rebaselining.
 These notes should be in reverse-chronological order, and use the following time format: (YYYY-MM-DD).
 
+<<<<<<< HEAD
```
CusiniM:
```suggestion
```

.integrated_tests.yaml
```diff
@@ -1,4 +1,3 @@
----
 baselines:
   bucket: geosx
   baseline: integratedTests/baseline_integratedTests-pr3141-5289-995723a
```
```suggestion
  baseline: integratedTests/baseline_integratedTests-pr3050-5325-9f50d94
```

PR #3056, opened 2024-03-25
Freeze porosity when doing poromechanics initialization

paveltomin:
This PR attempts to fix the following issue:
- during poromechanics initialization step the porosity is being updated which affects the mechanical solution trough changes in body force
- once initialization is done, the porosity changes are ignored (which is right thing to do) but that moves mechanical state slightly from equilibrium, visibly enough to affect the results in some cases

The proposed solution is to freeze the porosity during the initialization stage.


paveltomin:
This trick
https://github.com/GEOS-DEV/GEOS/blob/cb20b88c5e2d9afcf0b727bd8abb70f2c5366772/src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp#L2313
and this
https://github.com/GEOS-DEV/GEOS/blob/cb20b88c5e2d9afcf0b727bd8abb70f2c5366772/src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.cpp#L697
can be removed.

jhuang2601:
Even frozen porosity change during initialization, some inevitable update of porosity after initialization would still move the model away from equilibrium, as Biot porosity depends on stress and pressure. To enforce equilibrium, I think we can either keep the porosity changes during initialization or change the reference state for Biot porosity model to the initial conditions (user input).

paveltomin:
> Even frozen porosity change during initialization, some inevitable update of porosity after initialization would still move the model away from equilibrium, as Biot porosity depends on stress and pressure. To enforce equilibrium, I think we can either keep the porosity changes during initialization or change the reference state for Biot porosity model to the initial conditions (user input).

This PR is trying to achieve the second option - "change the reference state for Biot porosity model to the initial conditions"
We did some tests with @Guotong-Ren and it seems to work fine

jhuang2601:
> > Even frozen porosity change during initialization, some inevitable update of porosity after initialization would still move the model away from equilibrium, as Biot porosity depends on stress and pressure. To enforce equilibrium, I think we can either keep the porosity changes during initialization or change the reference state for Biot porosity model to the initial conditions (user input).
> 
> This PR is trying to achieve the second option - "change the reference state for Biot porosity model to the initial conditions" We did some tests with @Guotong-Ren and it seems to work fine

But in the current implementation, reference state corresponds to zero stress and pressure conditions.
image
If forcing the defaultReferencePorosity to be the initial one, reference value for pressure and stress should also be updated accordingly?

jhuang2601:
Double check the code. Reference state is always defined as the one from previous step and incremental formula is used, which makes sense for this PR.

jhuang2601 (approved):

binw1 (approved):

paveltomin:
`FAILED:  7 python3(ThermoPoroElastic_staircase_co2_smoke_01_2_restartcheck), python3(impermeableFault_smoke_01_2_restartcheck), python3(permeableFault_smoke_01_2_restartcheck), python3(PoroElastic_staircase_singlephase_3d_fim_01_2_restartcheck), python3(PoroElastic_staircase_singlephase_3d_sequential_01_2_restartcheck), python3(PoroElastic_staircase_co2_3d_fim_01_2_restartcheck), python3(PoroElastic_staircase_co2_3d_sequential_01_2_restartcheck)`
Since staircase tests use poromechanics initialization, they are expected to have diffs

PR #3059, opened 2024-03-27
Separate stick, slip, open parts of contact residual and try to normalize them properly

paveltomin:
Separately compute contact residuals for elements in stick, slip, and open states and normalize:
- Stick: normalize locally rhs by element area and compute L2-norm of that
- Slip and Open: compute L2-norm of rhs and normalize by Linf norm or rhs (similar to for as continuous mechanics)

 
Bonus:
`lineSearchResidualFactor` - factor used to check residual increase for triggering line search (currently any residual increase will trigger line search which is too conservative)


inputFiles/lagrangianContactMechanics/ContactMechanics_TFrac_smoke.xml
```diff
@@ -49,7 +49,7 @@
       zCoords="{ 0, 1 }"
       nx="{ 5, 52, 5 }"
       ny="{ 5, 52, 5 }"
-      nz="{ 2 }"
+      nz="{ 3 }"
```
paveltomin: this is needed to stabilize the case - otherwise it somehow runs fine locally but fails in CI

src/coreComponents/physicsSolvers/contact/SolidMechanicsEmbeddedFractures.cpp
```diff
@@ -537,83 +537,90 @@ real64 SolidMechanicsEmbeddedFractures::calculateResidualNorm( real64 const & ti
 
   if( !m_useStaticCondensation )
   {
+    real64 const fractureResidualNorm = calculateFractureResidualNorm( domain, dofManager, localRhs );
```
paveltomin, line 540: didn't really check anything here, just introduced `calculateFractureResidualNorm` function

inputFiles/lagrangianContactMechanics/ContactMechanics_TFrac_smoke.xml
CusiniM: weird.

src/coreComponents/physicsSolvers/NonlinearSolverParameters.cpp
```diff
@@ -60,6 +60,11 @@ NonlinearSolverParameters::NonlinearSolverParameters( string const & name,
     setInputFlag( InputFlags::OPTIONAL ).
     setDescription( "Iteration when line search starts." );
 
+  registerWrapper( viewKeysStruct::lineSearchResidualFactorString(), &m_lineSearchResidualFactor ).
+    setApplyDefaultValue( 1.0 ).
+    setInputFlag( InputFlags::OPTIONAL ).
+    setDescription( "Factor to determine residual increase." );
```
CusiniM: so what would reasonable values be?

src/coreComponents/physicsSolvers/contact/SolidMechanicsEmbeddedFractures.cpp
CusiniM: so same code just in a different function, right?

src/coreComponents/physicsSolvers/contact/SolidMechanicsEmbeddedFractures.cpp
```diff
@@ -537,83 +537,90 @@ real64 SolidMechanicsEmbeddedFractures::calculateResidualNorm( real64 const & ti
...
 
-    RAJA::ReduceSum< parallelDeviceReduce, real64 > localSum( 0.0 );
+  RAJA::ReduceSum< parallelDeviceReduce, real64 > localSum( 0.0 );
 
-    // globalResidualNorm[0]: the sum of all the local sum(rhs^2).
-    // globalResidualNorm[1]: max of max force of each rank. Basically max force globally
-    real64 globalResidualNorm[2] = {0, 0};
+  // globalResidualNorm[0]: the sum of all the local sum(rhs^2).
+  // globalResidualNorm[1]: max of max force of each rank. Basically max force globally
+  real64 globalResidualNorm[2] = {0, 0};
```
CusiniM:
```suggestion
  real64 globalResidualNorm[2] = {0.0, 0.0};
```

src/coreComponents/physicsSolvers/contact/SolidMechanicsEmbeddedFractures.cpp
```diff
@@ -537,83 +537,90 @@ real64 SolidMechanicsEmbeddedFractures::calculateResidualNorm( real64 const & ti
...
-        forAll< parallelDevicePolicy<> >( subRegion.size(),
-                                          [localRhs, localSum, dofNumber, rankOffset, ghostRank] GEOS_HOST_DEVICE ( localIndex const k )
+      forAll< parallelDevicePolicy<> >( subRegion.size(),
+                                        [localRhs, localSum, dofNumber, rankOffset, ghostRank] GEOS_HOST_DEVICE ( localIndex const k )
+      {
+        if( ghostRank[k] < 0 )
         {
-          if( ghostRank[k] < 0 )
+          localIndex const localRow = LvArray::integerConversion< localIndex >( dofNumber[k] - rankOffset );
+          for( localIndex i = 0; i < 3; ++i )
```
CusiniM:
```suggestion
          for( int i = 0; i < 3; ++i )
```

src/coreComponents/physicsSolvers/contact/SolidMechanicsLagrangeContact.cpp
```diff
@@ -695,43 +699,77 @@ real64 SolidMechanicsLagrangeContact::calculateContactResidualNorm( DomainPartit
...
           localIndex const localRow = LvArray::integerConversion< localIndex >( dofNumber[k] - rankOffset );
-          for( localIndex dim = 0; dim < 3; ++dim )
+          switch( fractureState[k] )
           {
-            localSum += localRhs[localRow + dim] * localRhs[localRow + dim];
+            case contact::FractureState::Stick:
+              {
+                for( localIndex dim = 0; dim < 3; ++dim )
+                {
+                  real64 const norm = localRhs[localRow + dim] / area[k];
```
CusiniM, line 734: why dividing by an area? the residual has the units of a cubic length so I would expect the normalizer to also have those units so that we keep everything dimensionless. We could use the minimum aperture or something like that. It's not that obvious for the tangential component though. Maybe the max jump at the previous timestep?

CusiniM (approved):
Do you see improvements in the convergence? I am not fully conviced by the stick case normalization. I think we should try to keep dimensionless.

CusiniM:
also, why not merging this into develop directly?

paveltomin:
> also, why not merging this into develop directly?

i was not sure this PR will fly, i can merge it into https://github.com/GEOS-DEV/GEOS/pull/3049 but would prefer you to have a look at #3049 first separately

src/coreComponents/physicsSolvers/contact/SolidMechanicsLagrangeContact.cpp
paveltomin: I didn't find any good "reference aperture" value which is not too small to normalize by
So after discussion with @Guotong-Ren we decided that 1m as reference is not that bad

src/coreComponents/physicsSolvers/contact/SolidMechanicsEmbeddedFractures.cpp
paveltomin: yes, just moved into a function, similar to conforming code

paveltomin:
> Do you see improvements in the convergence? I am not fully conviced by the stick case normalization. I think we should try to keep dimensionless.

Yes, some case from @jhuang2601 started to behave more reasonably (see some discussion here https://fcmaelstrom.slack.com/archives/C06LS3LH51P/p1711552600352519)

src/coreComponents/physicsSolvers/NonlinearSolverParameters.cpp
paveltomin: 1.1, 2.0, 10.0 ?
doing line search every time when residual does not decrease is a bit too much...

CusiniM: okay, then maybe add it to the description and I would also add a check that the value entered is greater than 0.0.

CusiniM:
> > also, why not merging this into develop directly?
> 
> i was not sure this PR will fly, i can merge it into #3049 but would prefer you to have a look at #3049 first separately

I think I would merge these changes directly into develop separately.

paveltomin:
> > > also, why not merging this into develop directly?
> > 
> > 
> > i was not sure this PR will fly, i can merge it into #3049 but would prefer you to have a look at #3049 first separately
> 
> I think I would merge these changes directly into develop separately.

this branch happen to be based on #3049

src/coreComponents/physicsSolvers/NonlinearSolverParameters.cpp
paveltomin: done

PR #3060, opened 2024-03-28
Elastic Vertical Transverse Isotropic (VTI) wave solver implementation

acitrain:
In this PR, we add the Elastic VTI physic for the second order wave solver.

The main difference between this one and the isotropic case (ElasticWaveEquationSEM.cpp) is the elasticity tensor C which is different.


sframba (commented):
Please use the base Elastic classes and have the new VTI classes inherit from the isotropic case. Most of the code can be reused. Override only the methods that need overriding, and reuse the base methods for all the basic setup. This will reduce a lot the added code and improve clarity and maintenability

src/coreComponents/physicsSolvers/wavePropagation/ElasticVTIWaveEquationSEM.cpp
```diff
@@ -0,0 +1,773 @@
...
+#include "ElasticMatricesSEMKernel.hpp"
+#include "PrecomputeSourcesAndReceiversKernel.hpp"
+
+namespace geos
+{
+
+using namespace dataRepository;
+using namespace fields;
+
+ElasticVTIWaveEquationSEM::ElasticVTIWaveEquationSEM( const std::string & name,
```
sframba: Have this class inherit from ElasticWaveEquationSEM. You can avoid:
- almost all the registerWrapper calls in the initializer (just add the new ones)
- most of the registerDataOnMesh method (jsut add register for the new fields)
- essentially everything else in this class.

To do this, you can template the ElasticWaveEquation calss as 
```
template< typename KERNELS = ElasticWaveEquationSEMKernels >
class ElasticWaveEquation :: public WaveSolverBase { bla bla bla }
```
and then
```
class ElasticWaveVTIEquation :: public ElasticWaveEquation<ElasticVTIWaveEquationKernels > { bla bla bla }
```

src/coreComponents/physicsSolvers/wavePropagation/ElasticVTIWaveEquationSEM.hpp
```diff
@@ -0,0 +1,223 @@
...
+#include "WaveSolverBase.hpp"
+#include "mesh/MeshFields.hpp"
+#include "physicsSolvers/SolverBase.hpp"
+#include "ElasticFields.hpp"
+#include "ElasticVTIFields.hpp"
+
+
+namespace geos
+{
+
```
sframba: Define the inheritance here

src/coreComponents/physicsSolvers/wavePropagation/ElasticVTIWaveEquationSEMKernel.hpp
```diff
@@ -0,0 +1,292 @@
...
+
+#include "finiteElement/kernelInterface/KernelBase.hpp"
+#include "WaveSolverUtils.hpp"
+#include "ElasticFields.hpp"
+
+namespace geos
+{
+using namespace fields;
+/// Namespace to contain the elastic wave kernels.
+namespace ElasticVTIWaveEquationSEMKernels
```
sframba: Same here. This class should inherit from the base Elastic class.
you just need to override:
- the StackVariables class
- the setup() method
- the quadraturePointKernel() method

for example, template the base ElasticWaveEquationSEMKernel class:
```
class StackVariables;
template< typename SUBREGION_TYPE,
          typename CONSTITUTIVE_TYPE,
          typename FE_TYPE,
          typename STACK_VARS = StackVariables >
class ExplicitElasticVTISEM : public finiteElement::KernelBase< SUBREGION_TYPE, CONSTITUTIVE_TYPE, FE_TYPE, 1, 1 >

void setup( localIndex const k,
              STACK_VARS & stack ) const

etc...
```
and then in the VTI class
```
class StackVariables;
template< typename SUBREGION_TYPE,
          typename CONSTITUTIVE_TYPE,
          typename FE_TYPE >
class ExplicitElasticVTISEM : public ExplicitElasticSEM< SUBREGION_TYPE, CONSTITUTIVE_TYPE, FE_TYPE, StackVariables >
```

sframba (commented):
This is where the base classes should be used. Also a few (untested) tips on how to implement inheritance. This should reduce the code by roughly 80% and improve maintenability

sframba (approved):

PR #3063, opened 2024-04-02
Refactoring of the ghosting setup

### LINKED ISSUE #3101: Rearchitecture of the ghosting setup
In order to allow the import of
- wells and EDFMs,
- on top of the DFN and the main 3d mesh,
- while managing all their interactions,

we need to reorganize the ghosting setup strategy of `geos`.

Here’s a list of the main actions items

Theoretical background
- [x] Global numbering techniques
- [x] Parallel adjacency matrix
- [x] Regular presentations to the team
- [x] Handover to Ryan
- [ ] Improving the algorithm (how many times do we have to do upward/downward multiplication? Can we merge upwards and downward matrix?)

Implementation
- [x] Global numbering techniques
- [x] Parallel adjacency matrix
- [ ] Plug into current GEOS (find the best way to bring the feature).
  - [x] Make it available for the "general case"
  - [ ] Handle Fracture
  - [ ] Handle wells (if required)
  - [ ] Handle particle methods
  - [ ] Handle high-order methods
  - [ ] Handle the `InternalMeshGenerator`
  - [ ] Handle multiple cell types
  - [ ] Temporary indicator to use new/old ghosting
  - [ ] Hook up to use/create globalIDs
  - [ ] Finish tpetra conversion (hard coded matrix sizes)
- [ ] Trilinos use and possibility of using LAI + LvArray
- [ ] GEOS packing
- [ ] Typed integer use
- [ ] Cleaning implementation
- [ ] Remove obsolete code

Add Documentation

TotoGaz:


src/coreComponents/mesh/generators/NewGhosting.cpp
```diff
@@ -0,0 +1,1173 @@
...
+  MpiWrapper::allReduce( numLocMissingCompound, numGlbMissingCompound, 2, MPI_SUM );
+  std::size_t const & numLocMissingIndices = numLocMissingCompound[0];
+  std::size_t const & numLocMissingNodes = numLocMissingCompound[1];
+  Tpetra::global_size_t const numGlbMissingIndices = numGlbMissingCompound[0];
+  Tpetra::global_size_t const numGlbMissingNodes = numGlbMissingCompound[1];
+
+  std::size_t const numGlbNodes = gis.nodes.get() + 1;  // TODO Use strongly typed integers
+  std::size_t const numOwnedNodes = std::size( ownedNodesIdcs );  // TODO Use strongly typed integers
+
+  auto const missingIndicesMap = make_rcp< TriMap const >( numGlbMissingIndices, numGlbMissingIndices, TriGlbIdx{ 0 }, comm );
```
ryar9534: Should the second argument here be numLocMissingIndices instead of the global ?

TotoGaz: Hmm, yes probably, I would have expected a failure then...
I guess my current test case is too simple.

PR #3065, opened 2024-04-05
Fix inconsistencies between serial and parallel conforming fracs run + add new output fields.

CusiniM:
In this PR:

- [x] add some useful fields to be output on the fracs. The slip is provided as a scalar as the sign can vary depending on the localRef and make interpretation confusing.
- [x] Force normals of faces next to fracture elements to be pointing outward (i.e., toward the fracture). 
- [x] Force the `elems2dToFaceMap` to have the faces ordered based on the global index of the neighboring cells to force consistency.
- [x] Imposed that the sparsity pattern of the jacobian matrix  is set based on the stencil (important for the stabilization matrix). 
- [x] Modified xmls so that meshes have aspect ratios reasonable for MGR to actually work. 

https://github.com/GEOS-DEV/integratedTests/pull/115


CusiniM:
@jhuang2601 You should be able to test this branch.

CusiniM:
@TotoGaz I think we can add in this PR the check that @matteofrigo5 pointed out that is necessary to do on the normal vectors for the external import to work.

CusiniM:
> @TotoGaz I think we can add in this PR the check that @matteofrigo5 pointed out that is necessary to do on the normal vectors for the external import to work.

Done! 

@jhuang2601 you should be able to test again the external mesh cases.

src/coreComponents/physicsSolvers/contact/SolidMechanicsLagrangeContact.cpp
```diff
@@ -1267,13 +1295,15 @@ void SolidMechanicsLagrangeContact::
 
           localIndex const localRow = LvArray::integerConversion< localIndex >( elemDOF[0] - rankOffset );
 
-
           for( localIndex idof = 0; idof < 3; ++idof )
           {
             localRhs[localRow + idof] += elemRHS[idof];
 
+            // std::cout << "global faceElement: " << localToGlobal[kfe] << " idof: " << idof << " nnz: " <<
+            // localMatrix.numNonZeros(localRow + idof) << std::endl;
```
CusiniM:
```suggestion
```

src/coreComponents/physicsSolvers/contact/SolidMechanicsLagrangeContact.cpp
```diff
@@ -1499,7 +1529,7 @@ void SolidMechanicsLagrangeContact::assembleStabilization( MeshLevel const & mes
             // Combine E and nu to obtain a stiffness approximation (like it was an hexahedron)
             for( localIndex j = 0; j < 3; ++j )
             {
-              stiffDiagApprox[ kf ][ i ][ j ] = E / ( ( 1.0 + nu )*( 1.0 - 2.0*nu ) ) * 2.0 / 9.0 * ( 2.0 - 3.0 * nu ) * volume / ( boxSize[j]*boxSize[j] );
+              stiffDiagApprox[ kf ][ i ][ j ] = 1.0e-2 * E / ( ( 1.0 + nu )*( 1.0 - 2.0*nu ) ) * 2.0 / 9.0 * ( 2.0 - 3.0 * nu ) * volume / ( boxSize[j]*boxSize[j] );
```
CusiniM: remove and maybe add a variable that can be set by the user.

src/coreComponents/physicsSolvers/contact/SolidMechanicsLagrangeContact.cpp
```diff
@@ -527,6 +534,12 @@ void SolidMechanicsLagrangeContact::setupDofs( DomainPartition const & domain,
                           contact::traction::key(),
                           DofManager::Connector::Elem,
                           meshTargets );
+
+  NumericalMethodsManager const & numericalMethodManager = domain.getNumericalMethodManager();
+  FiniteVolumeManager const & fvManager = numericalMethodManager.getFiniteVolumeManager();
+  FluxApproximationBase const & stabilizationMethod = fvManager.getFluxApproximation( m_stabilizationName );
+
+  dofManager.addCoupling( contact::traction::key(), stabilizationMethod );
```
CusiniM, line 542: the face coupling could be removed since this was added.

host-configs/apple/macOS_Matteo.cmake
```diff
@@ -1,4 +1,7 @@
 set( HOMEBREW_DIR "/Users/cusini1/local" )
 set( CONFIG_NAME "macOS_Matteo" ) 
 
+set(Python3_ROOT_DIR /Users/cusini1/pythonVenv/geosxEnv CACHE PATH "")
```
TotoGaz: Do you want to merge this?

src/coreComponents/mesh/FaceElementSubRegion.cpp
```diff
@@ -1048,4 +1048,104 @@ std::set< std::set< globalIndex > > FaceElementSubRegion::getCollocatedNodes() c
   return result;
 }
 
+void FaceElementSubRegion::flipFaceMap( FaceManager & faceManager,
+                                        ElementRegionManager const & elemManager )
```
TotoGaz, line 1052: Did you see the `fixNeighborMappingsInconsistency` function in the same file.
Does it conflict? Could the two fixes be done in one unique location?

src/coreComponents/mesh/FaceElementSubRegion.cpp
```diff
@@ -1048,4 +1048,104 @@ std::set< std::set< globalIndex > > FaceElementSubRegion::getCollocatedNodes() c
...
+
+  forAll< parallelHostPolicy >( this->size(), [=]( localIndex const kfe )
+  {
+    if( elems2dToFaces.sizeOfArray( kfe ) != 2 )
+    {
+      return;
+    }
+
+    localIndex const f0 = elems2dToFaces[kfe][0];
+    localIndex const f1 = elems2dToFaces[kfe][1];
```
TotoGaz:
```suggestion
    localIndex & f0 = elems2dToFaces[kfe][0];
    localIndex & f1 = elems2dToFaces[kfe][1];
```
then you can `std::swap( f0, f1 )` down below.
At your convenience.

TotoGaz (approved):

host-configs/apple/macOS_Matteo.cmake
CusiniM: yes. it's a bit silly to have my own hostconfig but it's quite convenient.

TotoGaz: Maybe change it with an `env` variable or a `which python3` or add a `-DPython3_ROOT_DIR=...` when you run your cmake?

src/coreComponents/mesh/FaceElementSubRegion.cpp
CusiniM: Honestly I did look into it but a lot is already happening n that function. I thought it was kind of easier to just add this cleanup step separately.

PR #3066, opened 2024-04-05
Update default hypre parameters

### LINKED ISSUE #2992: SEQ fails on GPUs
SEQ strategy is failing on GPUs as of https://github.com/GEOS-DEV/GEOS/commit/da6e83fc4f0f43a803b3300a3ec955ae0f404662

Example output for SEAM (coarse):

```
------------------- TIMESTEP START -------------------
    - Time:       -3168y, -319d, -4h-1m-4s (-100000000000 s)
    - Delta Time: 3168y, 319d, 04h01m04s (100000000000 s)
    - Cycle:      0
------------------------------------------------------


Time: -1.00e+11 s, dt: 100000000000 s, Cycle: 0

Task `ELASTICITY.PRE.INIT.STEP`: at time -100000000000s, physics solver `SEAM` is set to perform stress initialization during the next time step(s)
Task `ELASTICITY.PRE.INIT.STEP`: at time -100000000000s, physics solver `LINEAR.ELASTICITY` is resetting total displacement and velocity to zero
  Iteration  1: RESERVOIRANDWELL.SOLVER
        ( Rflow ) = ( 0.00e+00 )        ( Rwell ) = ( 0.00e+00 )        ( R ) = ( 0.00e+00 )
        MGR preconditioner: numComponentsPerField = [3, 4]
        Last LinSolve(iter,res) = (   0, 0.00e+00 )
        FLOW.SOLVER: Max pressure change: 0.000 Pa (before scaling)
        FLOW.SOLVER: Max component density change: 0.000 kg/m3 (before scaling)
        FLOW.SOLVER: Min pressure scaling factor: 1
        FLOW.SOLVER: Min component density scaling factor: 1
        RESERVOIRANDWELL.SOLVER: Global solution scaling factor = 1
        FLOW.SOLVER: Max phase volume fraction change = 0.0000
    Attempt:  0, ConfigurationIter:  0, NewtonIter:  1
        ( Rflow ) = ( 0.00e+00 )        ( Rwell ) = ( 0.00e+00 )        ( R ) = ( 0.00e+00 )
  Iteration  1: LINEAR.ELASTICITY
        ( Rsolid ) = ( 2.74e+01 )        ( R ) = ( 2.74e+01 )
***** ERROR
***** LOCATION: ${GEOS_PATH}/src/coreComponents/linearAlgebra/interfaces/hypre/HyprePreconditioner.cpp:445
***** Controlling expression (should be false): ierr != 0
***** Rank 2: Error in call to m_precond->destroy( m_precond->ptr )
Expected ierr == 0
  ierr = 1
  0 = 0
```

victorapm:
This PR updates the default values of the following parameters for GPU runs:

- Krylov restart size: 50
- AMG coarsening type: PMIS
- AMG relaxation type: L1-Jacobi

In addition:

- Add`aggressiveInterpMaxNonZeros` with default value equal to 16 for limiting the AMG complexity when using aggressive coarsening.
- Turn off F-relaxation in MGR when G-relaxation is active.
- Turn off aggressive coarsening in MGR's bottom AMG when using GPUs. 

Non-related change: a few schemas were updated after `make install`.

Closes #2992


jhuang2601 (approved):
Thanks for fixing this. SEAM CO2 cases can be running with this update.

src/coreComponents/linearAlgebra/utilities/LinearSolverParameters.hpp
```diff
@@ -116,7 +116,11 @@ struct LinearSolverParameters
   {
     real64 relTolerance = 1e-6;       ///< Relative convergence tolerance for iterative solvers
     integer maxIterations = 200;      ///< Max iterations before declaring convergence failure
+#if defined(GEOS_USE_HYPRE_CUDA) || defined(GEOS_USE_HYPRE_HIP)
+    integer maxRestart = 50;          ///< Max number of vectors in Krylov basis before restarting
+#else
```
castelletto1: Reducing to `50` the size of the Krylov space prior to restarting was very much needed. Thank you. 

I wonder if keeping such value to `200` makes sense when running on CPU. I suggest we reduce it to `100`. What do you think?

castelletto1 (approved):
Looks good to me, thank you @victorapm

castelletto1:
New baselines are needed since default linear solver parameters have been changed.

src/coreComponents/linearAlgebra/utilities/LinearSolverParameters.hpp
victorapm: Agreed!

paveltomin: Small suggestion is to try to be consistent between CPU and GPU if possible 
Should simplify the cross comparison and analysis

victorapm: That's a good suggestion. The reason for having different parameters is because each backend (CPU or GPU) have different features. `maxRestart`, for example, can be higher for CPUs because we have more memory available. That improves robustness of the solver as it will likely do less restarts

If we want "optimal" parameters, it's better to allow their values to differ according to the platform.
If we want easier comparisons, it's better to have same defaults for CPUs and GPUs

src/cmake/thirdparty/SetupGeosxThirdParty.cmake
```diff
@@ -636,7 +636,7 @@ if(DEFINED HYPRE_DIR AND ENABLE_HYPRE)
         set( HYPRE_DEPENDS ${HYPRE_DEPENDS} superlu_dist )
     endif()
     if( ${ENABLE_HYPRE_DEVICE} STREQUAL "CUDA" )
-        set( EXTRA_LIBS ${CUDA_cusparse_LIBRARY} ${CUDA_cublas_LIBRARY} ${CUDA_curand_LIBRARY} )
+        set( EXTRA_LIBS ${CUDA_cusparse_LIBRARY} ${CUDA_cublas_LIBRARY} ${CUDA_curand_LIBRARY} ${CUDA_cusolver_LIBRARY} )
```
victorapm: @CusiniM This is required to make the CI test pass: https://github.com/GEOS-DEV/GEOS/actions/runs/8767976628/job/24061832742?pr=3087

paveltomin (approved):

PR #3072, opened 2024-04-10
Neumann instead of Dirichlet boundary for 2nd order elastic wave free surface condition

sframba:
The free-surface condition for the elastic wave equation is the Neumann one, i.e., 

$\left.\frac{\partial u_{x,y,z}}{\partial n}\right\|_{\partial\Omega}=0$

Currently, the Dirichlet condition is applied instead:

$\left. u_{x,y,z}\right\|_{\partial\Omega}=0$

leading to an unphysical polarity change in the reflected wave. To correct this, one can simply remove the forced zeroing-out of boundary node displacements.


acitrain (approved):

sframba:
We are still having trouble accessing LC -- can someone please do the rebaseline if this gets to the top of the merge queue? Only the elastic wave solver should be affected. Thanks!

PR #3075, opened 2024-04-11
Introduce configuration tolerance

paveltomin:
Introduce configuration tolerance, the configuration is assumed converged if changed area is below the defined fraction of total area:
`changedArea <= configurationTolerance * totalArea`

Limits for `configurationTolerance`:
- minimum value is 0 (default) when any configuration change makes it unconverged
- values >=1 mean effectively configuration check is disabled

No default behavior change but needs rebaseline because of the new parameter.


src/coreComponents/physicsSolvers/contact/SolidMechanicsLagrangeContact.cpp
```diff
@@ -1941,27 +1944,28 @@ bool SolidMechanicsLagrangeContact::updateConfiguration( DomainPartition & domai
...
-                         MPI_LAND,
-                         MPI_COMM_GEOSX );
+  // Compute global area of changed elements
+  changedArea = MpiWrapper::sum( changedArea );
+  // and total area of fracture elements
+  totalArea = MpiWrapper::sum( totalArea );
 
-  return hasConfigurationConvergedGlobally;
+  // Assume converged if changed area is below certain fraction of total area
+  return changedArea <= m_nonlinearSolverParameters.m_configurationTolerance * totalArea;
```
CusiniM, line 1968: does this mean that if only a small element changes state we do not update the configuration?

paveltomin: My intention was to allow to converge when only a small portion of fracture elements change state
Typically happens when things are flip flopping

paveltomin: Configuration should be updated, this PR should be only about convergence criteria for configuration change - currently it is strict yes or no, here we are trying to allow some tolerance

paveltomin:
please review

matteofrigo5 (approved):
It looks fine to me

jhuang2601 (approved):

Guotong-Ren (approved):

PR #3086, opened 2024-04-19
Add a pressure-dependent permeability model and calculate the transmissibility derivatives

frankfeifan:
In this PR, we do the following things

- [x] Added a pressure-dependent permeability model. For now, it is an exponential relationship, but we can add more model types later. 
- [x] Calculate the derivatives of the transmissibility w.r.t pressure in `CellElementStencileTPFA`
- [x] Added a simple 1D example for this new feature.
- [x] Rebaseline


paveltomin (approved):

src/coreComponents/constitutive/permeability/PressurePermeability.hpp
```diff
@@ -0,0 +1,256 @@
...
+                                          arraySlice1d< real64 > const & dPerm_dPressure ) const
+{
+  for( localIndex i=0; i < permeability.size(); i++ )
+  {
+    if( pressureDependenceConstants[i] < 1e-20 )
+    {
+      permeability[i] = referencePermeability[i]; 
+      dPerm_dPressure[i] = 0.0;
+    }
+    else
```
CusiniM: is this branching really necessary? if you want to avoid values that are too small we can add a check when reading the inputfile.

CusiniM (approved):

src/coreComponents/constitutive/permeability/PressurePermeability.hpp
frankfeifan: > is this branching really necessary? if you want to avoid values that are too small we can add a check when reading the inputfile.

It is not a must for sure. I was thinking the component of `pressureDependenceConstants` can be set to zero, basically means no pressure dependence, but it will give us problem only when calculating the derivative for the hyperbolic type. Of course, I can add a check to avoid zero `pressureDependenceConstants`. So when using this model, one can assign a very small value if there is little pressure dependence in certain direction(s).

CusiniM: yeah, I would just add a check in `postProcessInput`

Issue #3014, opened 2024-02-29
Logging macros & Log Level refactor


MelReyCG:
To base the work of the #2277 EPIC, we need to :
- Refactor logging macros and the logger namespace so we have - at least under the hood - a real class,
- the messages should be all equally aligned with the outputting rank (see below),
- Refactor the `LogLevel`, it should be an `enum` with clear and documented labels (see below),

----

Refactored log output exemple :
```
INFO      : well ‘my well (myxml.xml:123)’ is shut
 RANK 170

===================================================
TIMESTEP  : Time: 0s, dt:864s, Cycle: 0
 RANK     0
===================================================

WARNING   : Time-step required will be increased based on state change.
 RANK 1
```

----

`LogLevel` enum proposal : 
```
/// @brief Enumerate the logging levels from the most importants messages to the one that contain the more details.
enum class LogLevel : int16_t
{
  Silent = -1,    // Useful as a globalLogLevel for silencing the logger. shouldn't be used as a message level.
  Important = 0,  // Application level messages (help, almost blocking warnings, application informations & phases)
  Progress = 1,   // Time step attempts, newton loop progress of Solver objects
  Detailed = 2,   // More detailed info and stats that affect progress, e.g. residual norms.
  Trace = 3,      // This is intended as a user-level debugging tool. detailed info trace what each component is
                  // doing. e.g. a line for every part of the assembly process, every boundary condition that's being
                  // applied by a physics solver, etc.
  Debug = 4,      // Information that are only relevant for a developper in a debugging context.
  DebugTrace = 5, // This level is useful to have deeper debugging information (which can be potencially heavy to log).
};
```

PR #3089, opened 2024-04-23
Adding logLevel documentation

### LINKED ISSUE #3179: Review all log level `Wrapper` documentation (`Wrapper::appendDescription()`),
The user need is to know what effect log levels values can have on the log output.

Proposal: add documentation on all existing `logLevel` wrappers & enforce to add description when enabling the `logLevel` wrapper.

arng40:
This PR aim to document all `logLevel` in GEOS and to invite the contributors to set descriptions when they enable the `logLevel` wrapper:

Currently, we have this output in all `Group` documentation files:
```
========= ========= ======== ================
Name      Type      Default  Description                                                                             
========= ========= ======== ================
[...]
logLevel  integer   0        Log level     
========= ========= ======== ================
```

This PR brings more context to all `logLevels` of GEOS:

I.e. `SolidMechanics_LagrangianFEM.rst` :
```
========= ========= ======== ========================================================================================
Name      Type      Default  Description                                                                             
========= ========= ======== ========================================================================================
[...]
logLevel  integer   0        | Sets the level of information to write in the standard output (the console typically).
                             | A level of 0 outputs minimal information, higher levels require more.                 
                             | logLevel >= 1                                                                         
                             |  - Information on line search                                                         
                             |  - Information on global solution scaling factor                                      
                             |  - Information on the timestep                                                        
                             |  - Print residual norm                                                                
                             | logLevel >= 1 and configuration didn't converge                                       
                             |  - Information about testing new configuration and print the time step                
                             | logLevel >= 1 and incorrect solution                                                  
                             |  - Information about line search failed                                               
                             | logLevel >= 1 and linear system                                                       
                             |  - Information on number of iterations and residual reduction                         
                             | logLevel >= 1 and non linear system                                                   
                             |  - Information on each newton Iteration                                               
                             | logLevel >= 1 and residual norm above the max allowed residual norm                   
                             |  - Indicate allowed residual norm                                                     
                             | logLevel >= 2                                                                         
                             |  - Output to screen the assembled linear system and solutions (matrices and vectors)  
                             |  - Infos on residuals values                                                          
                             | logLevel >= 2 and target set is empty                                                 
                             |  - Warning about boundary conditions                                                  
                             | logLevel >= 3                                                                         
                             |  - Output to file the assembled linear system and solutions (matrices and vectors)    
========= ========= ======== ========================================================================================
```


src/coreComponents/constitutive/relativePermeability/RelpermDriver.cpp
```diff
@@ -52,6 +50,9 @@ RelpermDriver::RelpermDriver( const geos::string & name,
     setInputFlag( InputFlags::OPTIONAL ).
     setApplyDefaultValue( "none" ).
     setDescription( "Baseline file" );
+
+  appendLogLevel( { "logLevel >= 1", "Infos about RelpermDriver" } );
+  appendLogLevel( { "logLevel >= 1", "Indicate if the internal results are consistent" } );
```
MelReyCG: I would have expect either this syntax:
```c++
  appendLogLevel( { "logLevel >= 1", "Infos about RelpermDriver" },
                  { "logLevel >= 1", "Indicate if the internal results are consistent" } );
```
Or this one
```c++
  appendLogLevel( "logLevel >= 1", "Infos about RelpermDriver" );
  appendLogLevel( "logLevel >= 1", "Indicate if the internal results are consistent" );
```

src/coreComponents/dataRepository/Group.cpp
```diff
@@ -79,6 +79,29 @@ void Group::deregisterWrapper( string const & name )
...
+  try
+  {
+    WrapperBase & wrapper = getWrapperBase( logLevelName );
+    appendLogLevelToWrapper( wrapper, log );
+  }catch( std::domain_error const & )
+  {
+    WrapperBase & wrapper = registerWrapper( logLevelName, &m_logLevel );
+    enableLogLevelInput();
+    appendLogLevelToWrapper( wrapper, log );
+  }
```
MelReyCG: In this project, try-catch is for adding more info to errors, not to recover from an erroneous state.
Use a if-else syntax.

src/coreComponents/dataRepository/Group.hpp
```diff
@@ -854,6 +854,12 @@ class Group
   ///@}
   //END_SPHINX_INCLUDE_REGISTER_WRAPPER
 
+  /**
+   * @brief Append a log level/description to the description of the wrapped object
+   * @param log An entry (log level/ log Description) to append
+   */
+  void appendLogLevel( std::pair< const std::string, const std::string > log );
```
MelReyCG: `appendLogLevelDescription( string_view loggingCondition, string_view caseDescription )` ?

src/coreComponents/dataRepository/WrapperBase.hpp
```diff
@@ -695,6 +747,10 @@ class WrapperBase
   /// A string description of the wrapped object
   string m_description;
 
+  /// Map for building the log level string for each wrapper
+  /// key : logLevel, values : description(s) for a level
+  std::map< std::string, std::vector< std::string > > m_logLevelsDescriptions;
```
MelReyCG: Only one `logLevel` per `Group` = only one map per `Group`
Every wrapper should not have `logLevel` functionnalities, `Group` should.
Those method & attributes you added in `Wrapper` are not at the right place.

MelReyCG (changes requested):

MelReyCG (changes requested):
Blocking comments

src/coreComponents/constitutive/relativePermeability/RelpermDriver.cpp
arng40: The second one is more convenient

paveltomin:
How is all that connects with proposal about `logLevel` labels in https://github.com/GEOS-DEV/GEOS/issues/3014 ?

paveltomin:
I like the initiative and direction but I don't really like that logLevel info is so detailed and not systematic.
For example, for linear solver it should look something like that (just an example):
```
• (=0) No messages generated.
• (>=1) A summary of the linear solver control parameters.
• (=2) Convergence history for the iterative solver at each nonlinear iteration.
• (=3) Additional convergence history from the preconditioner.
• (>3) Additional convergence history from deeper levels of the solver hierarchy.
```
It is

arng40:
This PR is related to the #3179 and not the #3014.

src/coreComponents/dataRepository/Group.hpp
```diff
@@ -854,6 +854,13 @@ class Group
   ///@}
   //END_SPHINX_INCLUDE_REGISTER_WRAPPER
 
+  /**
+   * @brief Append a levelCondition and a log description to the description of the wrapped object
+   * @param levelCondition The level condition to append
+   * @param logDescription The log description to append
+   */
+  void appendLogLevelDescription( string_view levelCondition, string_view logDescription );
```
TotoGaz: Is there a way, instead of having now 4 `public` member functions and 2 attributes directly related to logs (and a entry in the `viewKeysStruct`) to gather all the behaviors related to the log into a dedicated `class`?
I personally do not want to add more and more stuff into `Group` which is already big enough (+/- 120 `public` member functions IIRC).

TotoGaz (commented):
Could we discuss the this in a meeting first?

src/coreComponents/dataRepository/Group.hpp
MelReyCG: A bit off-topic comment : I don't know if it would help, but in a long-term effort. I would say that if we use a similar approach with the group's wrappers with a `WrapperContainers`, we could remove around 36 members (including const/non-const), allowing to clarify that methods like "size()" are related to Wrappers.

arng40:
@TotoGaz Thanks for your feedback, in reply to your suggestion :
Would a highly coupled class be an issue? Group would only have 1 attribute & 1 getter for this class.
Or is it better to have a purely not coupled class? Group would have 1 attribute, 1 getter, and 1 method (`appendLogLevelDescription()`).

PR #3091, opened 2024-04-24
feat: Transmissibility output

### LINKED ISSUE #3012: Add a cell / cell transmissibility data table
This issue is part of the #2277 EPIC.

MelReyCG:
This PR is after the #3247.

------------

This PR aims to add a new component: the `StencilOutput`, that allow to add a `CellToCellOutput` task node in the xml to output `ConnectionData` (global id pairs, transmissibilities).

**For now**, by using the `writeCSV="1"` attribute, the `CellToCellOutput` will output this data in a CSV. Here is an exemple:

- XML excerpt
```xml
  <Solvers gravityVector="{0.0, 0.0, 0.0}">
    <SinglePhaseFVM
      name="singlePhaseFlowSolver"
      […]
    </SinglePhaseFVM>
  </Solvers>

  <Tasks>
    <CellToCellOutput
      name="stencilOutput"
      flowSolverName="singlePhaseFlowSolver"
      writeCSV="1"
      logLevel="1" />
  </Tasks>

  <Events maxTime="10.0">
    <SoloEvent
      name="stencilOutputEvent"
      targetCycle="1"
      target="/Tasks/stencilOutput" />
  </Events>
```

- First 10 lines of the generated HDF5
```csv
time [s] , A global id , B global id , AB transmissibility , BA transmissibility
       4 ,           0 ,        2715 , 7.55463e-18                          , -7.55463e-18                        
       4 ,           0 ,       47546 , 8.06167e-18                          , -8.06167e-18                        
       4 ,           0 ,      134459 , 6.63826e-18                          , -6.63826e-18                        
       4 ,           0 ,      165202 , 7.78139e-18                          , -7.78139e-18                        
       4 ,           1 ,        4642 , 1.83325e-17                          , -1.83325e-17                        
       4 ,           1 ,       53907 , 1.76737e-17                          , -1.76737e-17                        
       4 ,           1 ,      184341 , 1.81705e-17                          , -1.81705e-17                        
       4 ,           2 ,        5459 , 1.7414e-17                           , -1.7414e-17                         
       4 ,           2 ,       63907 , 1.86145e-17                          , -1.86145e-17                        
       4 ,           2 ,      177257 , 1.86145e-17                          , -1.86145e-17                        
```
_(aligned here for readability)_


src/coreComponents/physicsSolvers/fluidFlow/StencilOutput.cpp
```diff
@@ -0,0 +1,233 @@
...
+    {
+      NumericalMethodsManager const & numericalMethodManager = domain.getNumericalMethodManager();
+      FiniteVolumeManager const & fvManager = numericalMethodManager.getFiniteVolumeManager();
+      FluxApproximationBase const & fluxApprox = fvManager.getFluxApproximation( m_solver->getDiscretizationName() );
+
+      fluxApprox.forStencils< CellElementStencilTPFA >( mesh, [&]( auto const & stencil )
+      {
+        // gather
+        auto const stencilWrapper = stencil.createKernelWrapper();
+        array1d< KernelConnectionData > const kernelData = gatherTimeStepData( mesh, stencilWrapper );
```
paveltomin: the naming is a bit confusing - why it is time step data?

src/coreComponents/physicsSolvers/fluidFlow/StencilOutput.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+template< typename CONN_DATA_CONT >
+void StencilOutput::writeCsvData( std::ofstream & outputFile, CONN_DATA_CONT const & data, real64 const time ) const
+{
+  for( StencilOutput::ConnectionData const & conn : data )
+  {
+    outputFile << time << ","
+               << conn.globalId[0] << ","
+               << conn.globalId[1] << ","
+               << conn.transmissibility[0] << ","
```
paveltomin: not sure it makes much sense to print the same transmissibility value twice but it's ok, not a big deal

src/coreComponents/physicsSolvers/fluidFlow/StencilOutput.cpp
```diff
@@ -0,0 +1,233 @@
...
+  {
+    m_solver->forDiscretizationOnMeshTargets( domain.getMeshBodies(), [&] ( string const &,
+                                                                            MeshLevel & mesh,
+                                                                            arrayView1d< string const > const & )
+    {
+      NumericalMethodsManager const & numericalMethodManager = domain.getNumericalMethodManager();
+      FiniteVolumeManager const & fvManager = numericalMethodManager.getFiniteVolumeManager();
+      FluxApproximationBase const & fluxApprox = fvManager.getFluxApproximation( m_solver->getDiscretizationName() );
+
+      fluxApprox.forStencils< CellElementStencilTPFA >( mesh, [&]( auto const & stencil )
```
paveltomin: could you please check if all that will work for https://github.com/GEOS-DEV/GEOS/blob/develop/inputFiles/compositionalMultiphaseFlow/co2_hybrid_1d.xml ?
not sure hybrid discretization has the same things as TPFA

src/coreComponents/physicsSolvers/fluidFlow/StencilOutput.cpp
karimifard: Maybe it will be more useful to output two half-transmissibilities.

src/coreComponents/physicsSolvers/fluidFlow/StencilOutput.cpp
```diff
@@ -0,0 +1,233 @@
...
+        connData[iConn].transmissibility[i] = transmissibility[0][i];
+        connData[iConn].regionId[i] =    stencilWrapper.getElementRegionIndices()( iConn, i );
+        connData[iConn].subRegionId[i] = stencilWrapper.getElementSubRegionIndices()( iConn, i );
+        connData[iConn].elementId[i] =   stencilWrapper.getElementIndices()( iConn, i );
+      }
+    } );
+  }
+
+private:
+  StencilOutputKernel()
```
dkachuma: ` = delete`?

src/coreComponents/physicsSolvers/fluidFlow/StencilOutput.cpp
dkachuma: I suppose the key here is that `fvManager.getFluxApproximation( m_solver->getDiscretizationName() );` might return null if the solver doesn't have a flux approximation. So I would add a validator (`postProcessInput()`) to the `solverName` field of this to ensure that the solver provided has the appropriate flux approximation.

src/coreComponents/physicsSolvers/fluidFlow/StencilOutput.hpp
```diff
@@ -0,0 +1,169 @@
...
+      return globalId[0] != other.globalId[0] ? globalId[0] < other.globalId[0] :
+             globalId[1] < other.globalId[1];
+    }
+  };
+
+
+  /**
+   * @brief the list of the current registered output target filepaths.
+   */
+  std::set< string > outputFiles;
```
dkachuma: I was a bit confused by this one. Maybe add the `m_` prefix to make it easier to work out that it's a member variable.

src/coreComponents/physicsSolvers/fluidFlow/StencilOutput.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+
+bool StencilOutput::execute( real64 const time_n,
+                             real64 const dt,
+                             integer const GEOS_UNUSED_PARAM( cycleNumber ),
+                             integer const GEOS_UNUSED_PARAM( eventCounter ),
+                             real64 const GEOS_UNUSED_PARAM( eventProgress ),
+                             DomainPartition & domain )
+{
+  if( m_writeCSV > 0 && MpiWrapper::commRank() == 0 )
```
dkachuma: Does it mean the root process has all the connections?

src/coreComponents/physicsSolvers/fluidFlow/StencilOutput.cpp
```diff
@@ -0,0 +1,233 @@
...
+template< typename CONN_DATA_CONT >
+void StencilOutput::writeCsvData( std::ofstream & outputFile, CONN_DATA_CONT const & data, real64 const time ) const
+{
+  for( StencilOutput::ConnectionData const & conn : data )
+  {
+    outputFile << time << ","
+               << conn.globalId[0] << ","
+               << conn.globalId[1] << ","
+               << conn.transmissibility[0] << ","
+               << conn.transmissibility[1] << std::endl;
```
dkachuma: I would avoid the flush that is implied by `std::endl` and use `\n` - it might speed things up a fraction.

CusiniM (commented):
It seems that we are trying to only write a file for rank 0 but I don't see when you gather all stencil data there? This would only work for small cases.... 
In any case, why not using the timehistory and writing in a parallel format instead? @wrtobin opinions? 

Besides debugging I don't really see why someone would want to write out transmissibilities.

src/coreComponents/physicsSolvers/fluidFlow/StencilOutput.cpp
MelReyCG: Despite my proposal is not totally finished, it now takes into account all connections within all ranks.

MelReyCG:
@CusiniM 
> It seems that we are trying to only write a file for rank 0 but I don't see when you gather all stencil data there? This would only work for small cases.... In any case, why not using the timehistory and writing in a parallel format instead? @wrtobin opinions?
> 
> Besides debugging I don't really see why someone would want to write out transmissibilities.

This proposal is not completely finished, but it now takes into account all connections within all ranks. The `CellToCellDataCollection` now can be used for bigger cases.

To explain why we want to output this data, the goal is to compare the transmissibility data from other simulator(s), in order to understand the differences between the simulations.

paveltomin (approved):

CusiniM (approved):

rrsettgast:
@MelReyCG Can you resolve conflicts?

src/coreComponents/finiteVolume/docs/StencilModel.plantuml
CusiniM, line 1: Can you remove these? Coz they are also written in French.

src/coreComponents/physicsSolvers/fluidFlow/StencilDataCollection.cpp
```diff
@@ -0,0 +1,273 @@
...
+    m_solver->forDiscretizationOnMeshTargets( domain.getMeshBodies(),
+                                              [&] ( string const &,
+                                                    MeshLevel & mesh,
+                                                    arrayView1d< string const > const & )
+    {
+      NumericalMethodsManager const & numericalMethodManager = domain.getNumericalMethodManager();
+      FiniteVolumeManager const & fvManager = numericalMethodManager.getFiniteVolumeManager();
+      FluxApproximationBase const & fluxApprox = fvManager.getFluxApproximation( m_solver->getDiscretizationName() );
+
+      fluxApprox.forStencils< CellElementStencilTPFA >( mesh, [&]( auto const & stencil )
```
CusiniM: Maybe supporting the output for all the other stencils could be useful too. I m thinking of debugging of fractured problems.

rrsettgast:
@MelReyCG Can you merge develop into this?

src/coreComponents/physicsSolvers/fluidFlow/StencilDataCollection.cpp
MelReyCG: Of course, I don't currently have any cases with fractures that need this output, so I may add more supported stencils later, but this component is opened to addition!

src/coreComponents/finiteVolume/docs/StencilModel.plantuml
MelReyCG: Are we keeping this in the repo if I translated them?

dkachuma (approved):

untereiner (approved):

MelReyCG:
Just leaving a note here, no CI step is legitimately failing here. The CUDA builds were working but crash now because of a CI bug since I updated my branch.

CusiniM:
This PR needs one more review.

src/coreComponents/dataRepository/Group.hpp
```diff
@@ -338,8 +338,12 @@ class Group
...
                             << dumpSubGroupsNames(),
                    std::domain_error );
-
-    return dynamicCast< T & >( *child );
+    T * const castedChild = dynamicCast< T * >( child );
+    GEOS_THROW_IF( castedChild == nullptr,
+                   GEOS_FMT( "{} was expected to be a '{}'.",
+                             child->getDataContext(), LvArray::system::demangleType< T >() ),
+                   BadTypeError );
+    return *castedChild;
```
rrsettgast, line 346: There is a `dynamic_cast` function in `RTTypes.hpp` for this.
```c++
template< typename NEW_TYPE, typename EXISTING_TYPE >
NEW_TYPE dynamicCast( EXISTING_TYPE & val )
{
  static_assert( std::is_reference< NEW_TYPE >::value, "NEW_TYPE must be a reference." );

  using POINTER_TO_NEW_TYPE = std::remove_reference_t< NEW_TYPE > *;
  POINTER_TO_NEW_TYPE ptr = dynamicCast< POINTER_TO_NEW_TYPE >( &val );
  GEOS_ERROR_IF( ptr == nullptr, "Cast from " << LvArray::system::demangleType( val ) << " to " <<
                 LvArray::system::demangleType< NEW_TYPE >() << " failed." );

  return *ptr;
}
```

rrsettgast (approved):

PR #3092, opened 2024-04-24
Stabilized Fixed-Stress

ryar9534:
Activates the option for pressure stabilization in fixed-stress, multiphase poromechanics.

Some items in, for example, MultiphasePoromechanics.cpp could be moved to the base poromechanics solver class once single-phase stabilization is implemented, which I propose to do in a follow-up PR (though if it is preferred to do it here, I am happy to do that as well).


src/coreComponents/physicsSolvers/multiphysics/MultiphasePoromechanics.cpp
```diff
@@ -118,6 +118,12 @@ void MultiphasePoromechanics< FLOW_SOLVER >::registerDataOnMesh( Group & meshBod
   if( m_stabilizationType == StabilizationType::Global ||
       m_stabilizationType == StabilizationType::Local )
   {
+
+    if( this->getNonlinearSolverParameters().m_couplingType == NonlinearSolverParameters::CouplingType::Sequential )
+    {
+      this->flowSolver()->enableJumpStabilization();
```
paveltomin, line 124: this enables the stabilization by default for sequential run, right?

ryar9534: It shouldn't since its in the if statements for Global or Local stabilization.

paveltomin (approved):

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoromechanics.cpp
CusiniM: I almost think that we should have it on by default t.b.h.

CusiniM (approved):
This looks good. I would move on and merge it quickly and then focus on adding it to singlephase coz we certainly need it for those benchmarks involving contact.

jhuang2601 (approved):
Thanks for this timely implementation and also expect single-phase stabilization scheme.

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoromechanics.cpp
ryar9534: Because of how it can smear interfaces, I would be hesitant to leave it on by default. But this is certainly open for discussion as more people start to make use of it.

CusiniM: with the permeabilities that people usually pick for the burdens, people will likely see weird pressures but yeah... it's a tough choice.

paveltomin:
now you don't have a test covering the new feature?

jhuang2601:
If pressure stabilization is enabled by default, rebaseline is still needed for certain smoke tests with sequential coupling.

ryar9534:
I will add new tests when doing the single phase stabilization PR, as the current test using stabilization isnt super meaningful right now anyway. Stabilization is not enabled by default, so it shouldnt affect any existing tests. Nicola and I ran the integrated tests and it only failed the one staircase example with sequential where I had added stabilization @paveltomin @jhuang2601

PR #3093, opened 2024-04-24
Small clean up for SinglePhasePoromechanicsEmbeddedFractures

paveltomin:
Removed redundant functions that are already available from the base CoupledSolver


CusiniM (approved):

CusiniM (approved):
Let's run the tests just to be sure.

Issue #3000, opened 2024-02-26
Rename label `ci: ready to be merged`


TotoGaz:
Label `ci: ready to be merged` is used to trigger CUDA builds.
Let's rename it `ci: run CUDA builds` accordingly.

TotoGaz:
@ryar9534 I think that you can remove your branch https://github.com/GEOS-DEV/GEOS/tree/3000-rename-label-ci-ready-to-be-merged

PR #3094, opened 2024-04-24
Make optional builds re-runnable on CI

### LINKED ISSUE #3001: Make `cuda_builds`/`run_integrated_tests` builds re-runnable on the CI
- [ ] Currently, if we forget to tag the PR with the `ci: ready for review` label, then  the CUDA builds are not run.
If we want to trigger them afterwards, the whole PR will be rerun (since the `is_not_draft_pull_request` rerun will enforce all its downstream jobs to be rerun).
- [ ] The same pattern applis to the integrated tests. Let's fix it too.

ryar9534:
Refactors the CI workflows such that the checks for labels `ci: ready to be merged` and `ci: run integrated tests` occur within the corresponding Github Actions jobs, which means that these can be re-run if labels are added after an update to the PR branch, for example. 

Resolves #3001


ryar9534:
I think this would do what we want as of now. However, for a PR the job get_current_labels will always fail. This is so that one may change the tags and execute the label-dependent integrated test and CUDA jobs by simply re-running failed jobs. I am not sure if there is a more clever way which avoids this, but at least I think the get_current_labels job should pass on a push to Develop.

.github/workflows/ci_tests.yml
```diff
@@ -60,6 +54,25 @@ jobs:
       run: |
         echo "DOCKER_IMAGE_TAG=$(jq '.build.args.GEOS_TPL_TAG' -r .devcontainer/devcontainer.json)" >> "$GITHUB_OUTPUT"
 
+  get_current_labels:
```
TotoGaz: Doing so you've tightened together `cuda` and `integrated` tests.
Adding an `integrated tests` label will require `get_current_labels` to be rerun and the trailing `cuda` builds will follow.
We should have a separated way of selecting the builds.

Maybe something like
```mermaid
flowchart TD
    is_draft --> whatever_previous 
    whatever_previous --> is_cuda_required
    whatever_previous --> are_integrated_tests_required
    is_cuda_required --> cuda_builds
    are_integrated_tests_required --> run_integrated_tests
```
We'll do the query twice but we don't care.
In the `*_required` jobs, you won't have to perform any tests since they'll be in their respective parent job.

.github/workflows/ci_tests.yml
```diff
@@ -193,8 +206,9 @@ jobs:
   run_integrated_tests:
     needs:
       - is_not_draft_pull_request
+      - get_current_labels
       - cpu_builds
-    if: "${{ contains( fromJSON( needs.is_not_draft_pull_request.outputs.LABELS ), 'ci: run integrated tests') || github.event_name != 'pull_request' }}"
+    if: "${{ always() && (contains( fromJSON( needs.get_current_labels.outputs.LABELS ), 'ci: run integrated tests') || github.event_name != 'pull_request') }}"
```
TotoGaz: The addition of the `always` here is quite unsettling. What's the purpose of it?

.github/workflows/ci_tests.yml
CusiniM: Can we look into manual triggers options instead of using labels? People usually add the labels and then leave them there forever. I think that, most of the times, one just wants to run the cuda builds once, and if they pass, probably does not need to rerun them any more until the very last second. So, I wonder if having an actual button to trigger  those jobs could be a good solution. I think it would save a lot of useless builds. Considering that we only have 2 runners I think it does matters.

ryar9534: I believe things should now be separated as suggested. Note that now the job that checks for a specific label (ready to be merged or run integrated tests) now passes when that label is found. Thus it doesnt pick up when a label is removed, for example, but Im not sure if that happens often in the workflow.

Manual triggers are also an option - I have no idea how they work and how/if it would work with the all tests passed check (i.e. can you pass this if you only manually trigger one job, but all of the others passed previously)

.github/workflows/ci_tests.yml
```diff
@@ -193,8 +239,9 @@ jobs:
   run_integrated_tests:
     needs:
       - is_not_draft_pull_request
+      - get_integrated_test_labels
       - cpu_builds
-    if: "${{ contains( fromJSON( needs.is_not_draft_pull_request.outputs.LABELS ), 'ci: run integrated tests') || github.event_name != 'pull_request' }}"
+    if: "${{ (contains( fromJSON( needs.get_integrated_test_labels.outputs.LABELS ), 'ci: run integrated tests') || github.event_name != 'pull_request') }}"
```
TotoGaz: Do you then need to check that the labels indeed contain `ci: run integrated tests`?
If you're here, then `get_integrated_test_labels` did not fail, which means that `ci: run integrated tests` was found?
Would it make the job if you rename `get_integrated_test_labels` into `are_integrated_tests_required` and let `are_integrated_tests_required` fail if not? And then you can remove the `outputs` from your guard jobs.

Same for CUDA.

.github/workflows/ci_tests.yml
TotoGaz: @CusiniM Do you know what GHA provides in terms of manual run?

.github/workflows/ci_tests.yml
ryar9534: Good point, I think you are totally right. Should be fixed now.

.github/workflows/ci_tests.yml
```diff
@@ -60,6 +54,50 @@ jobs:
...
 
+  are_cuda_builds_required:
+    needs: [is_not_draft_pull_request]
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Check for updated cuda labels
+        id: update_labels
+        run: |
+          if [[ ${{github.event_name}} != 'pull_request' ]]; then
+            echo "LABELS=['ci: ready to be merged']" >> "$GITHUB_OUTPUT"
```
TotoGaz:
```suggestion
```

.github/workflows/ci_tests.yml
```diff
@@ -60,6 +54,50 @@ jobs:
...
+            exit 0
+          else
+            pr_json=$(curl -H "Accept: application/vnd.github+json" https://api.github.com/repos/${{ github.repository }}/pulls/${{ github.event.number }})
+            LABELS=$(echo ${pr_json} | jq -crM '[.labels[].name]')
+            echo "LABELS=$LABELS" >> "$GITHUB_OUTPUT"
```
TotoGaz:
```suggestion
```

.github/workflows/ci_tests.yml
```diff
@@ -60,6 +54,50 @@ jobs:
...
+
+  are_integrated_tests_required:
+    needs: [is_not_draft_pull_request]
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Check for updated test labels
+        id: update_labels
+        run: |
+          if [[ ${{github.event_name}} != 'pull_request' ]]; then
+            echo "LABELS=['ci: run integrated tests']" >> "$GITHUB_OUTPUT"
```
TotoGaz:
```suggestion
```

.github/workflows/ci_tests.yml
```diff
@@ -60,6 +54,50 @@ jobs:
...
+            exit 0
+          else
+            pr_json=$(curl -H "Accept: application/vnd.github+json" https://api.github.com/repos/${{ github.repository }}/pulls/${{ github.event.number }})
+            LABELS=$(echo ${pr_json} | jq -crM '[.labels[].name]')
+            echo "LABELS=$LABELS" >> "$GITHUB_OUTPUT"
```
TotoGaz:
```suggestion
```

.github/workflows/ci_tests.yml
```diff
@@ -193,8 +231,8 @@ jobs:
   run_integrated_tests:
     needs:
       - is_not_draft_pull_request
+      - are_integrated_tests_required
```
TotoGaz:
```suggestion
      - are_integrated_tests_required
```

.github/workflows/ci_tests.yml
```diff
@@ -231,7 +269,7 @@ jobs:
     name: ${{ matrix.name }}
     needs:
       - is_not_draft_pull_request
```
TotoGaz, line 227:
```suggestion
```

ryar9534:
Randy also had an interesting suggestion to give `build_and_test` an input so it can query the labels as a first step, then decide to continue to run. This could get rid of these extra jobs which just read labels and also remove this issue that if you remove a label it is not currently detected. Im going to take a look at that.

ryar9534:
I implemented Randy's suggestion, the only thing that remains is handling the non-PR case for when the workflow is run.

.github/workflows/build_and_test.yml
```diff
@@ -47,13 +47,31 @@ on:
         required: false
         type: boolean
         default: true
+      IS_OPTIONAL:
+        required: false
+        type: boolean
+        default: false
+      REQUIRED_LABEL:
+        required: false
+        type: string
```
TotoGaz: Do you need `IS_OPTIONAL` or could you rely on `REQUIRED_LABEL` being defined to gather if the job has to be run conditionally?

.github/workflows/build_and_test.yml
```diff
@@ -47,13 +47,31 @@ on:
...
     secrets:
       GOOGLE_CLOUD_GCP:
         required: false
 jobs:
   build_test_deploy:
     runs-on: ${{ inputs.RUNS_ON }}
     steps:
+    - name: Check labels
```
TotoGaz: Can you update the label such that its intent is clearer for a new comer?

TotoGaz (commented):
I would have preferred to have the workflow part in the workflow file, but it's OK.

.github/workflows/build_and_test.yml
ryar9534: Nope, I was just having problems with the syntax. Its fixed now.

.github/workflows/build_and_test.yml
ryar9534: Can I go ahead and change/add the labels in Github? I think Randy and Matteo were onboard with this as well. So changing the `ci: ready to be merged` to something indicating it runs the CUDA builds?

TotoGaz: As you wish, it's this issue https://github.com/GEOS-DEV/GEOS/issues/3000

TotoGaz: But be careful to be in sync. When you merge you'll need to rename the `label`.
Plus I was also referring to the `- name: Check labels` which is not so obvious.

.github/workflows/build_and_test.yml
```diff
@@ -47,13 +47,28 @@ on:
...
+        default: 'NONE'
     secrets:
       GOOGLE_CLOUD_GCP:
         required: false
 jobs:
   build_test_deploy:
     runs-on: ${{ inputs.RUNS_ON }}
     steps:
+    - name: does_pr_have_necessary_labels
+      if: ${{inputs.REQUIRED_LABEL != 'NONE' && github.event_name == 'pull_request'}}
```
TotoGaz:
```suggestion
      if: ${{inputs.REQUIRED_LABEL && github.event_name == 'pull_request'}}
```
Is this working as expected if you do not define any default for `REQUIRED_LABEL`?

ryar9534: You are correct, this works. I have adjusted the code!

.github/workflows/build_and_test.yml
ryar9534: I have made the name of the step more descriptive. Perhaps its easier to change the label name in a separate PR just for clarity.

TotoGaz (approved):
Did you perform enough testing?

ryar9534:
> Did you perform enough testing?

I think it seems to be working

TotoGaz:
> I think it seems to be working

Champagne 🍾 

Can you update the description of the PR with what you've been doing and then we're good to merge I think.
@CusiniM Did you find anything related to making some jobs conditional based on manual (or not) triggers?
I'm not sure GHA provides a lot of flexibility in that regard.

CusiniM:
> > I think it seems to be working
> 
> Champagne 🍾
> 
> Can you update the description of the PR with what you've been doing and then we're good to merge I think. @CusiniM Did you find anything related to making some jobs conditional based on manual (or not) triggers? I'm not sure GHA provides a lot of flexibility in that regard.

yes, there is something called `workflow_dispatch` that might do what we want. 

https://stackoverflow.com/questions/58877569/how-to-trigger-a-step-manually-with-github-actions

it's unclear to me whether this can only apply to a subset of the jobs and not to the entire workflow. 

https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/

ryar9534:
> yes, there is something called `workflow_dispatch` that might do what we want.

Yeah we already do have that in the workflow,

>it's unclear to me whether this can only apply to a subset of the jobs and not to the entire workflow.

but this was also my question. It seems like you'd have to pass in inputs to tell it to only do a subset of the jobs.

PR #3102, opened 2024-04-29
Change `Unsupported` doc tag to `Best effort`

TotoGaz:
Also add a minimal example on how to configure `sccache` in the `Dockerfile`


ryar9534 (approved):
Looks good to me, I think Herve and I will be able to try it with folks shortly

PR #3103, opened 2024-04-30
Additional use of streak2 for CI testing

rrsettgast:
- [x] Enable Cuda builds on `streak2`
- [x] Generalize self hosted runner certificate process for multiple container OS (Ubuntu, CentOS)


.github/workflows/ci_tests.yml
```diff
@@ -237,8 +240,12 @@ jobs:
             ENABLE_HYPRE_DEVICE: CUDA
             ENABLE_HYPRE: ON
             ENABLE_TRILINOS: OFF
-            RUNS_ON: Runner_8core_32GB
-
+            RUNS_ON: streak2
+            NPROC: 16
+            DOCKER_RUN_ARGS: "--cpus=16 --memory=192g --runtime=nvidia --gpus all"
+            DOCKER_CERTS_DIR: "/usr/local/share/ca-certificates"
+            DOCKER_CERTS_UPDATE_COMMAND: "update-ca-certificates"
```
TotoGaz: Is it possible to get the bundle of certificates from `streak*` into the container by using a _read-only_ `mount` `docker` approach?

If we do something like `docker run -v /path/to/llnl.crt.pem:/usr/local/share/ca-certificates/llnl/bundle.crt:ro` (note the trailing `ro`), then simply running `update-ca-certificates` or `update-ca-trust` (we can let the distribution decide what's available) would do the job?

So we could reduce those lines to:
```suggestion
            DOCKER_RUN_ARGS: "--cpus=16 --memory=192g --runtime=nvidia --gpus all -v /path/to/llnl.crt.pem:/usr/local/share/ca-certificates/llnl/bundle.crt:ro"
```
and remove the copy part (and everything related to the `hostname`) from `scripts/ci_build_and_test_in_container.sh`?

.github/workflows/build_and_test.yml
```diff
@@ -108,6 +119,12 @@ jobs:
         script_args=()
 
 
+        DOCKER_CERTS_DIR=${{ inputs.DOCKER_CERTS_DIR }}
+        DOCKER_CERTS_UPDATE_COMMAND=${{ inputs.DOCKER_CERTS_UPDATE_COMMAND }}
+        NPROC=${{ inputs.NPROC }}
```
TotoGaz: If you can, you could define `nproc` as a `script_args` (`--nproc=...`), which would default (in the same way we handle the default arguments) to `nproc` in `ci_build_and_test_in_container.sh`?
That way, it's not an `env` variable.

.github/workflows/ci_tests.yml
rrsettgast: yeah. I forgot to do that...

.github/workflows/build_and_test.yml
rrsettgast: yeah. probably better.

.github/workflows/ci_tests.yml
rrsettgast: There are several certificate files to "copy", and they need their extension changed. So I think the the "copy" is preferable to the volume mounting of specific files.

.github/workflows/build_and_test.yml
rrsettgast: this is done...and it is apparently `--nproc ...` with no equals!

.github/workflows/build_and_test.yml
```diff
@@ -142,11 +166,12 @@ jobs:
...
-          RUNNER_CERTIFICATES_DIR=/etc/pki/ca-trust/source/anchors/
-          mkdir -p ${GITHUB_WORKSPACE}/certificates
-          cp ${RUNNER_CERTIFICATES_DIR}/*.crt* ${GITHUB_WORKSPACE}/certificates
-        fi
+        # if [ ${{ inputs.RUNS_ON }} == 'streak' ] || [ ${{ inputs.RUNS_ON }} == 'streak2' ]; then
+        #   RUNNER_CERTIFICATES_DIR=/etc/pki/ca-trust/source/anchors/
+        #   mkdir -p ${GITHUB_WORKSPACE}/certificates
+        #   cp ${RUNNER_CERTIFICATES_DIR}/*.crt* ${GITHUB_WORKSPACE}/certificates
+        # fi
+
```
rrsettgast:
```suggestion
```

.github/workflows/build_and_test.yml
```diff
@@ -186,6 +211,15 @@ jobs:
...
         # but for this precise call, we want to deal with it more precisely
+
+        echo ${docker_args[@]} \
+        -h=`hostname` \
+        ${{ inputs.DOCKER_REPOSITORY }}:${{ inputs.DOCKER_IMAGE_TAG }} \
+        ${GITHUB_WORKSPACE_MOUNT_POINT}/scripts/ci_build_and_test_in_container.sh \
+        ${script_args[@]}
+
+        echo "running docker command..."
+
```
rrsettgast:
```suggestion
```

scripts/ci_build_and_test_in_container.sh
```diff
@@ -1,21 +1,21 @@
 #!/bin/bash
 set -o pipefail
 
+echo "Running in container"
+
```
rrsettgast:
```suggestion
```

scripts/ci_build_and_test_in_container.sh
```diff
@@ -1,21 +1,21 @@
...
-echo "running nproc"
-nproc
 
 # docs.docker.com/config/containers/resource_constraints
 # Inside the container, tools like free report the host's available swap, not what's available inside the container.
 # Don't rely on the output of free or similar tools to determine whether swap is present.
-echo "running free -m"
-free -m
+# echo "running free -g"
+# free -g
```
rrsettgast:
```suggestion
echo "running free -g"
free -g
```

scripts/ci_build_and_test_in_container.sh
```diff
@@ -156,19 +158,20 @@ EOT
...
-    update-ca-certificates 
-    # gcloud config set core/custom_ca_certs_file cert.pem'
-    
-    NPROC=8
-  else
-    NPROC=$(nproc)
+    ${DOCKER_CERTS_UPDATE_COMMAND}
   fi
-  echo "Using ${NPROC} cores."
+  # gcloud config set core/custom_ca_certs_file cert.pem'
```
rrsettgast:
```suggestion
```

scripts/ci_build_and_test_in_container.sh
```diff
@@ -156,19 +158,20 @@ EOT
...
 
   echo "sccache initial state"
   ${SCCACHE} --show-stats
 fi
 
+if [ -z "${NPROC}" ]; then
+  NPROC=$(nproc)
+  echo "NPROC unset, setting to ${NPROC}..."
+fi
```
rrsettgast, line 173:
```suggestion
```

scripts/ci_build_and_test_in_container.sh
```diff
@@ -214,6 +217,8 @@ or_die python3 scripts/config-build.py \
                ${SCCACHE_CMAKE_ARGS} \
                ${ATS_CMAKE_ARGS}
 
+echo "done with config-build.py"
+
```
rrsettgast:
```suggestion
```

scripts/ci_build_and_test_in_container.sh
```diff
@@ -230,6 +235,7 @@ if [[ "${TEST_DOCUMENTATION}" = true ]]; then
 fi
 
 # Performing the requested build.
+echo "Running Ninja with ${NPROC} cores."
```
rrsettgast:
```suggestion
```

rrsettgast:
@TotoGaz I think this is good

.github/workflows/ci_tests.yml
TotoGaz: From the doc I could see that
```
update-ca-certificates  is a program that updates the directory [/etc/ssl/certs](file:///etc/ssl/certs) to hold SSL
       certificates  and  generates  ca-certificates.crt,  a  concatenated  single-file  list  of
       certificates.
```
So could we take that `ca-certificates.crt` and mount it like `-v /path/to/ca-certificates.crt:/somewhere/llnl.pem:ro` it should be enough? We can mount one file with another name/extension. That would spare us the copy burden?

ryar9534 (approved):

PR #3108, opened 2024-05-03
Fixed issue #2853 with a flag

liyangrock:
This PR addressed issue #2853 by adding a flag to determine if the lagging update should be applied to the fracture stencil weights. 

This was done specifically for the `hydrofracture` solver, as it was observed that issue #2853 was currently causing performance degradation only for this particular solver. Therefore, the flag was exclusively introduced to the `hydrofracture` solver.

For further discussions related to issue #2853, please see #2922 .


src/coreComponents/physicsSolvers/fluidFlow/FlowSolverBase.cpp
```diff
@@ -753,6 +754,8 @@ void FlowSolverBase::prepareStencilWeights( DomainPartition & domain ) const
     FiniteVolumeManager const & fvManager = numericalMethodManager.getFiniteVolumeManager();
     FluxApproximationBase const & fluxApprox = fvManager.getFluxApproximation( getDiscretizationName() );
     ElementRegionManager::ElementViewAccessor< arrayView1d< real64 const > > hydraulicAperture =
+      m_isLaggingFractureStencilWeightsUpdate ?
+      mesh.getElemManager().constructViewAccessor< array1d< real64 >, arrayView1d< real64 const > >( fields::flow::aperture0::key() ) :
```
paveltomin, line 759: @CusiniM is this even correct? should `aperture0` be mapped into hydraulic aperture?

paveltomin: it looks ok, sorry, i just got again confused by the naming

src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp
```diff
@@ -881,13 +891,19 @@ void HydrofractureSolver< POROMECHANICS_SOLVER >::updateState( DomainPartition &
 
   Base::updateState( domain );
 
-  // remove the contribution of the hydraulic aperture from the stencil weights
-  flowSolver()->prepareStencilWeights( domain );
+  if( !m_isLaggingFractureStencilWeightsUpdate )
+  {
+    // remove the contribution of the hydraulic aperture from the stencil weights
+    flowSolver()->prepareStencilWeights( domain );
```
paveltomin, line 864: @karimifard why `prepareStencilWeights` is needed separately from `updateStencilWeights`?
can't they be merged into one function?

Guotong-Ren:
To understand, is the flow area in fracture from last time step? Do you have any thoughts on evaluation of the time step size impact on the solution?

src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp
karimifard: We cannot store additional info in the stencil (CUDA limitation) so we need to remove the contribution of the aperture from the stencil (`prepareStencilWeights`) before updating the aperture and then update the stencil with the new aperture.

liyangrock:
> To understand, is the flow area in fracture from last time step? Do you have any thoughts on evaluation of the time step size impact on the solution?

Indeed, the calculation of fracture stencil weights is performed using the `hydraulicAperture` from the previous time step. I have conducted a comparison between `dt=0.25s` and `dt=5.0s` for viscosity-dominated KGD cases. The resulting net pressure and aperture at the mouth are displayed below. 

The difference between the numerical result (with dt=0.25s) and the analytical result can be attributed to the fact that the numerical scenario employs a low toughness value (i.e., 1e4), whereas the analytical solution does not incorporate any toughness.

kgd_verification

The input files are included here.
[kgdVis.zip](https://github.com/GEOS-DEV/GEOS/files/15206362/kgdVis.zip)

paveltomin (approved):

src/coreComponents/physicsSolvers/fluidFlow/FlowSolverBase.cpp
CusiniM: yeah, we should probably work the naming coz it's confusing.

CusiniM (approved):

paveltomin:
@rrsettgast how to merge this?

rrsettgast:
@paveltomin . Can you create a branch on the main repo that contains these changes? that would be easiest for me...otherwise I have to just merge it and hope for the best.

rrsettgast:
replaced with https://github.com/GEOS-DEV/GEOS/pull/3196

PR #3021, opened 2024-03-05
Preparatory work for fractures + wells

paveltomin:
For now, assuming wells are not directly connected to the fractures.

In this version, only `SinglePhasePoromechanicsConformingFracturesReservoir` solver can run, although I am having hard time to build a well-behaving example. Other versions of solvers with wells are drafted but commented out for now.

Added poromechanics initialization for contact solver - technically runs but needs refinements.

Side thing: add default template types to avoid too long nested definitions

PR #3110, opened 2024-05-06
Clean up catalogName a bit

paveltomin:
catalog name related piece from https://github.com/GEOS-DEV/GEOS/pull/3021/


rrsettgast:
@paveltomin Did you search the whole codebase? There may be more places

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoromechanics.hpp
```diff
@@ -52,7 +53,21 @@ class SinglePhasePoromechanics : public PoromechanicsSolver< FLOW_SOLVER, MECHAN
...
+  {
+    return "SinglePhasePoromechanics";
+  }
+  template< typename _FLOW_SOLVER=FLOW_SOLVER >
+  static
+  typename std::enable_if< std::is_same< _FLOW_SOLVER, SinglePhaseReservoirAndWells< SinglePhaseBase > >::value, string >::type
+  catalogName()
+  {
+    return SinglePhaseReservoirAndWells< SinglePhaseBase >::catalogName() + "Poromechanics";
+  }
```
TotoGaz:
```suggestion
 static string catalogName()
  {
      if constexpr ( std::is_same_v< FLOW_SOLVER, SinglePhaseBase > )
      {
          return "SinglePhasePoromechanics";
      }
      else if constexpr( std::is_same_v< FLOW_SOLVER, SinglePhaseReservoirAndWells< SinglePhaseBase > > )
      {
          return SinglePhaseReservoirAndWells< SinglePhaseBase >::catalogName() + "Poromechanics";
      }
      else
      {
          static_assert(false, "blah blah");
      }
  }
```
Is this working?

paveltomin: i took it from here to be honest: https://github.com/GEOS-DEV/GEOS/blob/ddf3e97a398c1261cae81cf2f864112061e5b259/src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseFVM.hpp#L101
can you your version if it's preferred

TotoGaz: If my version is working and doing the right thing, I think it's more modern and simple.
I think that the `if constexpr` c++17 feature was thought for this use case (at least).

paveltomin: that `static_assert` fails for some reason, not sure why

TotoGaz: My `static_assert` code was bad, I wrote it very fast and surely you need to be something a little smarter!

paveltomin: i put together something, let's see if it can pass the tests

paveltomin:
> @paveltomin Did you search the whole codebase? There may be more places

i looked through yes but would not guarantee i paid full attention
mostly looked into solvers part

src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp
```diff
@@ -35,38 +35,17 @@ using namespace constitutive;
...
 {
-  return HydrofractureSolverCatalogNames< POROMECHANICS_SOLVER >::name();
+  return "Hydrofracture";
 }
-
+//template<>
+//string HydrofractureSolver< MultiphasePoromechanics< CompositionalMultiphaseBase > >::catalogName()
+//{
+//  return "MultiphaseHydrofracture";
+//}
```
TotoGaz: You do not want to spread the pattern for this function as well?

src/coreComponents/physicsSolvers/multiphysics/SinglePhaseReservoirAndWells.hpp
```diff
@@ -26,13 +26,13 @@
 namespace geos
 {
 
-template< typename SINGLEPHASE_RESERVOIR_SOLVER >
-class SinglePhaseReservoirAndWells : public CoupledReservoirAndWellsBase< SINGLEPHASE_RESERVOIR_SOLVER,
+template< typename RESERVOIR_SOLVER >
```
TotoGaz, line 30: Please add a `@tparam` that emphasize that we need a single phase solver (or the constraint is not valid anymore)?

TotoGaz (approved):

src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp
paveltomin: just did, thanks

src/coreComponents/physicsSolvers/multiphysics/SinglePhaseReservoirAndWells.hpp
paveltomin: added, it still has to be single phase yes

rrsettgast (approved):

PR #3120, opened 2024-05-10
Add missing tests into ATS and adjust output naming

paveltomin:
Also disable restarting for cases with EmbeddedSurfaceGenerator, since its postRestartInitialization function is not implemented (documented in https://github.com/GEOS-DEV/GEOS/issues/3126)


jafranc (approved):
Got all expected failure :)
Thanks for this update. :+1:

PR #3125, opened 2024-05-14
Register pressureGradient only for hybrid FVM (single phase)

paveltomin:


CusiniM:
In reality, it would be good if we could find a way to compute this also for tpfa. It's just harder to come up with a way to approximate it.

CusiniM (approved):

paveltomin:
> In reality, it would be good if we could find a way to compute this also for tpfa. It's just harder to come up with a way to approximate it.

something like that?
image
@karimifard can confirm if that makes sense

CusiniM:
> > In reality, it would be good if we could find a way to compute this also for tpfa. It's just harder to come up with a way to approximate it.
> 
> something like that? image @karimifard can confirm if that makes sense

yeah, what's annoying is the number of derivatives that entails... I guess we will just lag the term.

karimifard:
> > > In reality, it would be good if we could find a way to compute this also for tpfa. It's just harder to come up with a way to approximate it.
> > 
> > 
> > something like that? image @karimifard can confirm if that makes sense
> 
> yeah, what's annoying is the number of derivatives that entails... I guess we will just lag the term.

The formula should work for Cartesian grids but it's not clear to me it will work for non Cartesian unstructured grids. If I am not wrong, it is essentially a weighted average of face velocity, with the weights being distance to face times the face area, which is a volume. In the Cartesian case, these weights add up to the volume and the formula works but in general adding all the weights/volumes will be larger than the volume of the cell. So if we want to use something like (17) we need to adjust the cell volume, or maybe just average the face velocities (q/A). Another way to estimate the velocity inside the element is to use a least square approximation. I may have missed something...

CusiniM:
> > > > In reality, it would be good if we could find a way to compute this also for tpfa. It's just harder to come up with a way to approximate it.
> > > 
> > > 
> > > something like that? image @karimifard can confirm if that makes sense
> > 
> > 
> > yeah, what's annoying is the number of derivatives that entails... I guess we will just lag the term.
> 
> The formula should work for Cartesian grids but it's not clear to me it will work for non Cartesian unstructured grids. If I am not wrong, it is essentially a weighted average of face velocity, with the weights being distance to face times the face area, which is a volume. In the Cartesian case, these weights add up to the volume and the formula works but in general adding all the weights/volumes will be larger than the volume of the cell. So if we want to use something like (17) we need to adjust the cell volume, or maybe just average the face velocities (q/A). Another way to estimate the velocity inside the element is to use a least square approximation. I may have missed something...

The least square approximation s what we use for the Mimetic. the annoying thing with tpfa was that we don't store the face velocities so to compute cell center values we would need to do another stencil loop coz we need the neighbors' pressures.

PR #3130, opened 2024-05-17
Augmented Lagrangian (stick mode) - InterfaceKernelBase - Bubble Functions

matteofrigo5:
This PR introduces the following three features:

- Development of `kernelBaseInterface`: This new interface is designed to launch kernels that perform operations on interface elements.
- Implementation of Bubble Element and Face Basis Functions: These functions are used for stabilizing the first-order displacement cell constant pressure formulation in contact mechanics (and the formulation can be easily extended to flow problems).
- Augmented Lagrangian Method for Contact Mechanics: This initial implementation focuses on the stick mode. Subsequent PRs will address the slip and open modes once this PR is merged.


src/coreComponents/finiteElement/elementFormulations/H1_Hexahedron_Lagrange1_GaussLegendre2.hpp
```diff
@@ -75,6 +75,10 @@ class H1_Hexahedron_Lagrange1_GaussLegendre2 final : public FiniteElementBase
 public:
   /// The number of nodes/support points per element.
   constexpr static localIndex numNodes = LagrangeBasis1::TensorProduct3D::numSupportPoints;
+
+  /// The number of nodes/support points per element.
```
CusiniM: ?

src/coreComponents/finiteElement/elementFormulations/H1_Hexahedron_Lagrange1_GaussLegendre2.hpp
```diff
@@ -209,6 +213,42 @@ class H1_Hexahedron_Lagrange1_GaussLegendre2 final : public FiniteElementBase
     return calcN( q, N );
   }
 
+  /**
+   * @brief Calculate shape functions values for each support face at a
```
CusiniM: what's a support face? A face to which we are associating a bubble function?

src/coreComponents/finiteElement/elementFormulations/H1_TriangleFace_Lagrange1_Gauss1.hpp
```diff
@@ -142,6 +142,42 @@ class H1_TriangleFace_Lagrange1_Gauss1 final : public FiniteElementBase
...
                      real64 ( &N )[numNodes] );
 
+  /**
+   * @brief Calculate shape bubble functions values at a given point in the parent space.
+   * @param pointCoord coordinates of the given point.
+   * @param N An array to pass back the shape function values.
+   */
+  GEOS_HOST_DEVICE
+  GEOS_FORCE_INLINE
+  static void calcBubbleN( real64 const (&pointCoord)[2],
```
CusiniM, line 152: it seems that for 3D elements the bubble functions live on the faces whereas for 2D elements there is one per element. This is clearly how we are using them for contact but if we want this to be more generic we will need element wise bubbles for the 3D cells elements, right?

src/coreComponents/mesh/CellElementSubRegion.hpp
```diff
@@ -379,6 +405,12 @@ class CellElementSubRegion : public ElementSubRegionBase
   /// Map from local Cell Elements to Embedded Surfaces
   EmbSurfMapType m_toEmbeddedSurfaces;
 
+  /// List of the elements sharing some face with the interface elements (useful for bubble elements)
+  array1d< localIndex > m_bubbleCells;
+
+  /// List of face ids which are shared with the interface elements (useful for bubble elements)
+  array2d< localIndex > m_toFaceElements;
```
CusiniM: I am not sure I fully understand these 2 lists.  They also don't seem to be initialized nor registered as wrappers. I would do that. Make sure you set the sizeFromParent to false. 

it seems to be that `m_bubbleCells` just contains the indices of all 3D cells that are connected to a 2D fracture cell. I would say: 
"List of the elements adjacent to a faceElement (useful for bubble elements)"

The second list is a bit more confusing. What's the size of that array2d? Are you trying to store the `localIndex` of each face connected to the `faceElement` ?  That map already exists in the `FaceElementSubRegion` so I would use that one.

src/coreComponents/physicsSolvers/contact/SolidMechanicsALMBubbleKernels.hpp
```diff
@@ -0,0 +1,448 @@
...
+   * @copydoc finiteElement::ImplicitKernelBase::StackVariables
+   */
+  struct StackVariables
+  {
+public:
+    /// The number of displacement dofs per element.
+    static constexpr int numUdofs = numNodesPerElem * 3;
+
+    /// The number of jump dofs per element.
+    static constexpr int numBubbleUdofs = numFacesPerElem * 3;
```
CusiniM, line 122: are you currently assuming that all faces get a bubble?

src/coreComponents/physicsSolvers/contact/SolidMechanicsAugmentedLagrangianContact.cpp
```diff
@@ -0,0 +1,1347 @@
...
+                                                        "",
+                                                        kernelFactory );
+
+      GEOS_UNUSED_VAR( maxTraction );
+
+    } );
+
+  } );
+
+  // Loop for assembling contributes of bubble elements (Abb, Abu, Aub)
```
CusiniM, line 340: you may want to break this down into multiple functions.

src/coreComponents/physicsSolvers/contact/SolidMechanicsAugmentedLagrangianContact.cpp
```diff
@@ -0,0 +1,1347 @@
...
+  {
+    ElementRegionManager & elemManager = mesh.getElemManager();
+    SurfaceElementRegion & region = elemManager.getRegion< SurfaceElementRegion >( getUniqueFractureRegionName() );
+    FaceElementSubRegion & subRegion = region.getUniqueSubRegion< FaceElementSubRegion >();
+
+    arrayView2d< real64 const > const dispJump = subRegion.getField< contact::dispJump >();
+    arrayView2d< real64 > const oldDispJump = subRegion.getField< contact::oldDispJump >();
+    arrayView2d< real64 > const deltaDispJump  = subRegion.getField< contact::deltaDispJump >();
+    //arrayView1d< integer const > const fractureState = subRegion.getField< contact::fractureState >();
+    //arrayView1d< integer > const oldFractureState = subRegion.getField< contact::oldFractureState >();
```
CusiniM: clean up

src/coreComponents/physicsSolvers/contact/SolidMechanicsAugmentedLagrangianContact.cpp
```diff
@@ -0,0 +1,1347 @@
...
+          }
+        }
+
+      }
+    }
+  } );
+
+}
+
+// TODO: Is it possible to define this method once? Similar to SolidMechanicsLagrangeContact
```
CusiniM, line 1195: what do you mean?

src/coreComponents/physicsSolvers/contact/SolidMechanicsAugmentedLagrangianContact.cpp
```diff
@@ -0,0 +1,1347 @@
...
+
+      arrayView2d< localIndex const > const elemsToFaces = cellElementSubRegion.faceList().toViewConst();
+
+      RAJA::ReduceSum< ReducePolicy< parallelDevicePolicy<> >, localIndex > nBubElems_r( 0 );
+
+      localIndex const n_max = cellElementSubRegion.size() * elemsToFaces.size( 1 );
+      array1d< localIndex > keys( n_max );
+      array1d< localIndex > perms( n_max );
+      array1d< localIndex > vals( n_max );
+      array1d< localIndex > localFaceIds( n_max );
```
CusiniM, line 855: if you only need these in here and they don't need to go on device I would use `std::vector`

src/coreComponents/physicsSolvers/contact/SolidMechanicsAugmentedLagrangianContact.cpp
```diff
@@ -0,0 +1,1347 @@
...
+
+      solidMechanicsALMKernels::ALMJumpUpdateFactory kernelFactory( dispDofNumber,
+                                                                    bubbleDofNumber,
+                                                                    dofManager.rankOffset(),
+                                                                    voidMatrix.toViewConstSizes(),
+                                                                    voidRhs.toView(),
+                                                                    dt,
+                                                                    faceElementList );
+
+      real64 maxTraction = finiteElement::
```
CusiniM, line 566: you probably can define this not to return a max traction since it's just updating the jump.

CusiniM (approved):

src/coreComponents/physicsSolvers/contact/SolidMechanicsALMJumpUpdateKernels.hpp
```diff
@@ -0,0 +1,275 @@
...
+namespace solidMechanicsALMKernels
+{
+
+/**
+ * @copydoc geos::finiteElement::ImplicitKernelBase
+ */
+template< typename CONSTITUTIVE_TYPE,
+          typename FE_TYPE >
+class ALMJumpUpdate :
+  public ALMKernelsBase< CONSTITUTIVE_TYPE,
```
CusiniM, line 36: I may be wrong but I don't think this Kernel needs to know the constitutive_type. You can write it's own launch if you only need to dispatch on the fe_type.

src/coreComponents/finiteElement/elementFormulations/H1_Hexahedron_Lagrange1_GaussLegendre2.hpp
matteofrigo5: Fixed the comment with: `faces/support point`

src/coreComponents/finiteElement/elementFormulations/H1_Hexahedron_Lagrange1_GaussLegendre2.hpp
matteofrigo5: Yes, a face to which we are associating a bubble function. I have modified  this comment with: `Calculate face bubble functions values for each face at a given point in the parent space.`

src/coreComponents/finiteElement/elementFormulations/H1_TriangleFace_Lagrange1_Gauss1.hpp
matteofrigo5: There are two types of bubble functions both for 3D and 2D elements:
- face bubble functions, which live on the faces/edges, for 3D/2D respectively;
- element bubble functions, which live on the elements.
In the code, when I refer to the first type, I have added `Face` in the function name.
Hence, the answer to your question is yes, in the more generic case we will need element wise bubbles functions for the 3D and also face bubble functions for 2D elements.

src/coreComponents/physicsSolvers/contact/SolidMechanicsALMBubbleKernels.hpp
matteofrigo5: Yes, I'm. But I assemble in the global matrix only the contribute of the face bubble connected to the fracture.

src/coreComponents/physicsSolvers/contact/SolidMechanicsAugmentedLagrangianContact.cpp
matteofrigo5: Is it convenient?

src/coreComponents/physicsSolvers/contact/SolidMechanicsAugmentedLagrangianContact.cpp
matteofrigo5: This function is very similar to the one used in `SolidMechanicsLagrangeContact`.
I was wondering if a new method could be defined in ContactSolverBase, for example. 
Anyway I will have a clearer idea about this function when the slip mode for ALM is also implemented.
In the meantime may I leave this comment as a reminder?

src/coreComponents/physicsSolvers/contact/SolidMechanicsAugmentedLagrangianContact.cpp
matteofrigo5: These variables are necessary to compute the lists of elements and faces required to assemble the bubble contributions.
All the operations to compute these lists are designed to run on GPU, so yes, they need to go on device.

src/coreComponents/mesh/CellElementSubRegion.hpp
matteofrigo5: Yes, you are right. I have to register the two lists as wrappers.

The second list is a 2D array containing the `localIndex` of each face in both the physical and parent element spaces, which are connected to the `faceElement`. I can't use the map in the `FaceElementSubRegion` because I need to access this list during the loop over the cells adjacent to a `faceElement`.

paveltomin:
@matteofrigo5 please check test results and apply rebaseline if everything looks ok

PR #3133, opened 2024-05-20
Force face-based ordering for node list of FaceElements

CusiniM:
Currently the nodes of `FaceElementSubRegions` are ordered differently when the `subRegion` is filled from an external mesh and when it is filled by the `SurfaceGenerator`. 

Specifically, the `SurfaceGenerator` orders the nodes of the `FaceElementSubRegion` as if it was an hexahedra whereas when reading the nodes from an external mesh the local odering il CCW for each face. This inconsistency creates an issue in the vtk output. In this PR we modify the odering of the nodes imposed by these two approaches to be consistent by always storing the nodes related to face 0 CCW, followed by the nodes related to face 1 also stored CCW. 

However, using the 3D element map for plotting could be interesting coz it would allow a 3D view of the facelemetn which has some advantages. To do so, we could probably plot faceElements as general polyhedra though, for which I assume that the CCW ordering by face will work.


src/coreComponents/physicsSolvers/surfaceGeneration/SurfaceGenerator.cpp
```diff
@@ -725,13 +725,13 @@ int SurfaceGenerator::separationDriver( DomainPartition & domain,
         localIndex const numNodesInFace = faceToNodeMap.sizeOfArray( faceMap[ kfe ][ 0 ] );
         for( localIndex a = 0; a < numNodesInFace; ++a )
         {
-          localIndex const aa = a < 2 ? a : numNodesInFace - a + 1;
-          localIndex const bb = aa == 0 ? aa : numNodesInFace - aa;
+          // localIndex const aa = a < 2 ? a : numNodesInFace - a + 1;
+          // localIndex const bb = aa == 0 ? aa : numNodesInFace - aa;
```
TotoGaz: Consider removing (or adding a comment?).

TotoGaz (approved):
So we're now in sync with
```c++
static std::vector< int > getVtkConnectivity( ElementType const elementType, localIndex const numNodes )
{
  switch( elementType )
  {
    ...
    case ElementType::Triangle:      return { 0, 1, 2 };
    case ElementType::Quadrilateral: return { 0, 1, 2, 3 };
    ...
```
?

CusiniM:
> So we're now in sync with
> 
> ```c++
> static std::vector< int > getVtkConnectivity( ElementType const elementType, localIndex const numNodes )
> {
>   switch( elementType )
>   {
>     ...
>     case ElementType::Triangle:      return { 0, 1, 2 };
>     case ElementType::Quadrilateral: return { 0, 1, 2, 3 };
>     ...
> ```
> 
> ?

we should be. HOwever, the better fix would be to use the `faceToNodeMap` for plotting a `faceElement` as a 2D cell instead of using the `elemToNodesMap` or to plot it as a 3D polyhedral element using the 'elemToNodesMap`.

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp
```diff
@@ -361,7 +363,9 @@ getSurface( FaceElementSubRegion const & subRegion,
 
   for( localIndex ei = 0; ei < subRegion.size(); ei++ )
   {
-    auto const & nodes = elemToNodes[ei];
+    // we use the nodes of face 0
+    localIndex const faceIndex = elemToFaces(ei, 0);
+    auto const & nodes = faceToNodes[faceIndex];
```
CusiniM: @TotoGaz I think this is a better fix.

rrsettgast:
Face elements should be plotted as volume elements to allow for inspection in deformed meshes. I.e. when looking at a crack from the side with the displacement amplified, it is nice to see the contents of the crack and have variable coloring available.

src/coreComponents/fileIO/vtk/VTKPolyDataWriterInterface.cpp
TotoGaz: As you wish

CusiniM:
> Face elements should be plotted as volume elements to allow for inspection in deformed meshes. I.e. when looking at a crack from the side with the displacement amplified, it is nice to see the contents of the crack and have variable coloring available.

I have made this optional for now. It's a nice tool for us but I have seen people being already quite confused about how fault elements are modeled that I think it's better if they are not induced to think that they are 3D elements.

matteofrigo5 (approved):

src/coreComponents/fileIO/Outputs/VTKOutput.hpp
```diff
@@ -89,6 +89,7 @@ class VTKOutput : public OutputBase
     static constexpr auto plotFileRoot = "plotFileRoot";
     static constexpr auto writeFEMFaces = "writeFEMFaces";
     static constexpr auto writeGhostCells = "writeGhostCells";
+    static constexpr auto writeFaceElementsAs3D = "writeGhostCells";
```
CusiniM:
```suggestion
    static constexpr auto writeFaceElementsAs3D = "writeFaceElementsAs3D";
```

src/coreComponents/physicsSolvers/surfaceGeneration/SurfaceGenerator.cpp
CusiniM:
```suggestion
```

CusiniM:
> > Face elements should be plotted as volume elements to allow for inspection in deformed meshes. I.e. when looking at a crack from the side with the displacement amplified, it is nice to see the contents of the crack and have variable coloring available.
> 
> I have made this optional for now. It's a nice tool for us but I have seen people being already quite confused about how fault elements are modeled that I think it's better if they are not induced to think that they are 3D elements.

Actually, to make this work properly for all shapes I will need to do a bit of extra work. I ll do it in a follow up PR so that I can merge this one and fix the current bug.

PR #3134, opened 2024-05-21
Logger / LogLevel rework

MelReyCG:
Draft PR


src/coreComponents/common/logger/LoggingObject.hpp
```diff
@@ -0,0 +1,137 @@
...
+
+  void setLogLevel( LogLevel value );
+
+
+  virtual void logRank0( LogMsg::CallParams loc, string_view msg, LogRouter router = LogRouter::m_main ) const = 0;
+
+  // TODO not exactly sure how to implement that (avoid formating if cond is false, keeping condition as metadata).
+  // virtual void logRank0If( bool cond, LogMsg::CallParams loc, string_view msg, LogRouter router = LogRouter::m_main ) const = 0;
+
+  virtual void log( LogMsg::CallParams loc, string_view msg, LogRouter router = LogRouter::m_main ) const = 0;
```
MelReyCG: @rrsettgast As the meeting of today concluded, I will explore a solution where `LoggingObject` would be a "light version" when called from a kernel. If that does not work, I will go for macros.

paveltomin:
@MelReyCG close?

MelReyCG:
@paveltomin No. I will come back to it.

MelReyCG:
There is a lot of work on the branch, I really need to keep it.

PR #3136, opened 2024-05-22
Implement stability test

dkachuma:
Implements stability test for compositional mixtures. Given a fluid mixture at a prescribed pressure and temperature, this tests if the mixture is thermodynamically stable as a single phase.

The solution starts with the 2 compositions determined by the Wilson k-values. For each composition, successive substitution is used to converge to a stationary point. The procedure then outputs the minimum tangent plane distance from the 2 trial compositions. Success is determined when both trial compositions converge to stationary points.

Michelson showed that a mixture will be stable if and only if the tangent-plane-distance (TPD) denoted by $g(y)$ in [CompositionalMultiphaseFluid](https://geosx-geosx--3136.com.readthedocs.build/en/3136/coreComponents/constitutive/docs/CompositionalMultiphaseFluid.html) is non-negative for all trial compositions $y$. For a 2 component system, we can enumerate this since the composition is essentially one dimensional. I have done this for a 20%-80% mix of $CH_4$ and $C_3H_{10}$ in the image below.
tpd
The x-axis is the trial composition expressed by the molar fraction of $CH_4$ in the trial composition. To determine if this is non-negative for all $y$, we need to find the stationary points where $g'(y)=0$ because these will include the minima. If the function is non-negative at all these stationary points, then we conclude that the mixture is stable. The feed composition $z$ (in this case $z_{CH_4}=0.2$) is itself a (trivial) stationary point and the value of $g$ at this point is zero. To get other stationary points we start with the values of $z$ perturbed by the Wilson k-values. For this example we get extreme values with the first having $y_{CH_4}\approx 0$ and the other $y_{CH_4}\approx 1$. Then we do successive substitution to converge to the stationary points. These are shown on the diagram as squares. One of the initial guesses almost invariably converges to the trivial stationary point. For pressures of 10bar and 200bar, we get extra stationary points but the value of $g$ at these points is positive - so we conclude that the mixture is stable as a single phase. For 50bar one of the stationary points gives a negative value of $g$ indicating instability.

This will next be linked to the negative flash. The point is to avoid going into the negative flash if the mixture is stable as a single phase. This will save some flops but more importantly will avoid inconsistent phase labelling that results from relying  on the solution of the flash when the mixture is single phase.


src/coreComponents/constitutive/fluid/multifluid/compositional/functions/StabilityTest.hpp
```diff
@@ -0,0 +1,231 @@
...
+private:
+  /**
+   * @brief Calculate which components are present.
+   * @details Creates a list of indices whose components have non-zero mole fraction.
+   * @param[in] numComps number of components
+   * @param[in] composition the composition of the fluid
+   * @param[out] presentComponents the list of present components
+   * @return the number of present components
+   */
+  template< typename ARRAY >
```
jafranc: Do we need templatization on `ARRAY` type ?

src/coreComponents/constitutive/fluid/multifluid/compositional/functions/StabilityTest.hpp
```diff
@@ -0,0 +1,231 @@
...
+  GEOS_FORCE_INLINE
+  static real64 normalizeComposition( integer const numComps,
+                                      arraySlice1d< real64, USD > const & composition )
+  {
+    real64 totalMoles = 0.0;
+    for( integer ic = 0; ic < numComps; ++ic )
+    {
+      totalMoles += composition[ic];
+    }
+    real64 const oneOverTotalMoles = 1.0 / (totalMoles + MultiFluidConstants::epsilon);
```
jafranc: wouldn't it be safer to throw/return 1 on null `totalMoles` ?

jafranc (approved):

src/coreComponents/constitutive/fluid/multifluid/compositional/functions/StabilityTest.hpp
CusiniM: I think this is good coz it makes it more generic. I would call it `ARRAY_TYPE` though.

CusiniM: I would do the same for the composition though.

src/coreComponents/constitutive/fluid/multifluid/compositional/functions/StabilityTest.hpp
```diff
@@ -0,0 +1,231 @@
...
+    // Check for machine-zero feed values
+    integer presentCount = 0;
+    for( integer ic = 0; ic < numComps; ++ic )
+    {
+      if( MultiFluidConstants::epsilon < composition[ic] )
+      {
+        presentComponents[presentCount++] = ic;
+      }
+    }
+    presentComponents.resize( presentCount );
```
CusiniM, line 196: this is not ideal.

CusiniM (commented):
I do not remember the details of this. I would suggest to write a documentation ( if possible with an image that explains what happens for a binary system) as part of this PR.

dkachuma:
> I do not remember the details of this. I would suggest to write a documentation ( if possible with an image that explains what happens for a binary system) as part of this PR.

I have added a description of the stability test based on Michelson. [CompositionalMultiphaseFluid](https://geosx-geosx--3136.com.readthedocs.build/en/3136/coreComponents/constitutive/docs/CompositionalMultiphaseFluid.html)

src/coreComponents/constitutive/fluid/multifluid/compositional/functions/StabilityTest.hpp
dkachuma: I think it's better to make it explicit that this is a stack array. I have therefore removed the template parameter - it was a copy from another area which is not required here since there's only a single call to this.

src/coreComponents/constitutive/fluid/multifluid/compositional/functions/StabilityTest.hpp
dkachuma: I have added an assert.

src/coreComponents/constitutive/fluid/multifluid/compositional/functions/StabilityTest.hpp
dkachuma: This is a resize on a `stackArray1d` which I believe shouldn't have much overhead.

CusiniM (approved):

src/coreComponents/constitutive/docs/CompositionalMultiphaseFluid.rst
```diff
@@ -8,12 +8,117 @@ Overview
...
+
+After calculating the phase fractions, phase component fractions, phase densities, phase viscosities,
+and their derivatives with respect to pressure, temperature, and component fractions, the
+:ref:`CompositionalMultiphaseFlow` then moves on to assembling the accumulation and flux terms.
+
+Step 1: Computation of the phase fractions and phase component fractions (flash)
+================================================================================
+Stability test
+-------------------------------
+The first step is to determine if provided mixture with total molar fractions :math:`z_c` is stable
```
CusiniM:
```suggestion
The first step is to determine if the provided mixture with total molar fractions :math:`z_c` is stable
```

src/coreComponents/constitutive/docs/CompositionalMultiphaseFluid.rst
```diff
@@ -8,12 +8,117 @@ Overview
...
+as a single phase at the current pressure :math:`p` and temperature :math:`T`. However, this can only
+be confirmed with through stability testing.
```
CusiniM:
```suggestion
be confirmed through stability testing.
```

src/coreComponents/constitutive/docs/CompositionalMultiphaseFluid.rst
```diff
@@ -8,12 +8,117 @@ Overview
...
+With the two starting mixtures, the stationarity condition is solved using successive substitution to
+determine the stationary points. If both initial states converge to a solution which has :math:`g(y)\geq 0`
+then the mixture is deemed to be stable, otherwise it is deemed unstable.
+
+Phase labeling
+----------------------------------------
+Once it is confirmed that the fluid with composition :math:`z` is stable as a single phase at the current
+pressure and temperature, it must be labeled as either 'liquid' or 'vapor'. This is necessary only to apply
+the correct relative permeability function for calculating the phase's flow properties. The properties of the
+fluid (density, viscosity, enthalpy) are unchaged by the assignment of the label.
```
dkachuma:
```suggestion
fluid (density, viscosity) are unchanged by the assignment of the label.
```

PR #3137, opened 2024-05-22
testing branch for develop

rrsettgast:
This is a branch for troubleshooting develop with minor alterations for testing. This branch should never be merged and should be reverted to develop when testing is complete.


scripts/ci_build_and_test_in_container.sh
```diff
@@ -183,15 +184,23 @@ if [[ "${RUN_INTEGRATED_TESTS}" = true ]]; then
 
   python3 -m pip cache purge
 
+  pip config set global.cert $HOME/certs/ca-bundle.crt
```
CusiniM:
```suggestion

```

PR #3138, opened 2024-05-22
Properly sync nonlinear solver params for coupled solver

paveltomin:
Some solvers logic depends on the number of Newton iterations.
`synchronizeNonlinearSolverParameters` was supposed to take care of that but it was not placed correctly.
In particular, the issue is that currently number of iterations for coupled solvers does not propagate down to subsolvers.
This PR tries to fix that.
Also add some debug info output for time step selection.


src/coreComponents/physicsSolvers/NonlinearSolverParameters.hpp
```diff
@@ -97,6 +99,8 @@ class NonlinearSolverParameters : public dataRepository::Group
 
   virtual void postProcessInput() override;
 
+  void print();
```
dkachuma:
```suggestion
  void print() const;
```

src/coreComponents/physicsSolvers/NonlinearSolverParameters.cpp
```diff
@@ -183,41 +183,44 @@ void NonlinearSolverParameters::postProcessInput()
...
-    {
-      GEOS_LOG_RANK_0( GEOS_FMT( "  Sequential convergence criterion = {}", EnumStrings< SequentialConvergenceCriterion >::toString( m_sequentialConvergenceCriterion ) ) );
-      GEOS_LOG_RANK_0( GEOS_FMT( "  Subcycling = {}", m_subcyclingOption ) );
-    }
+    print();
   }
 }
 
-
+void NonlinearSolverParameters::print()
```
dkachuma:
```suggestion
void NonlinearSolverParameters::print() const
```

src/coreComponents/physicsSolvers/SolverBase.cpp
```diff
@@ -350,15 +368,21 @@ real64 SolverBase::setNextDtBasedOnNewtonIter( real64 const & currentDt )
   {
     // Easy convergence, let's increase the time-step.
     nextDt = currentDt * m_nonlinearSolverParameters.timeStepIncreaseFactor();
+    if( m_nonlinearSolverParameters.getLogLevel() > 0 )
+      GEOS_LOG_RANK_0( GEOS_FMT( "{}: number of iterations = {} is less than {}, next time step = {}", getName(), newtonIter, iterIncreaseLimit, nextDt ));
```
dkachuma:
```suggestion
      GEOS_LOG_RANK_0( GEOS_FMT( "{}: number of iterations = {} is less than {}, next time step = {} (increase)", getName(), newtonIter, iterIncreaseLimit, nextDt ));
```

dkachuma: I would want to highlight that this is an increase.

src/coreComponents/physicsSolvers/multiphysics/CoupledSolver.hpp
```diff
@@ -472,15 +468,15 @@ class CoupledSolver : public SolverBase
 
         if( isConverged )
         {
+          // we still want to count current iteration
```
dkachuma, line 507: Not sure we need this. We might exit the loop with `iter==maxIterNewton` which may be reported somewhere else.

dkachuma (commented):
Looks good to me. My only question is: would we want to keep this as a "feature"? That is the user deliberately chooses different non-linear parameters for the different solvers?

paveltomin:
> Looks good to me. My only question is: would we want to keep this as a "feature"? That is the user deliberately chooses different non-linear parameters for the different solvers?

this logic only applies for fully coupled solvers, in this case different params does not make much sense

for sequential it is not going to sync anything and there is flexibility to choose different settings

src/coreComponents/physicsSolvers/multiphysics/CoupledSolver.hpp
paveltomin: the time step cut logic below relies on `isConverged` flag
i think it makes sense to count the iteration that was performed, and in current implementation it will only affect the time step selector (currently selector sees 0 as number of iterations when outer loop converges in one iteration)

src/coreComponents/physicsSolvers/SolverBase.cpp
```diff
@@ -271,11 +271,17 @@ bool SolverBase::execute( real64 const time_n,
...
           nextDt = dtRemaining / 2;
+          if( m_nonlinearSolverParameters.getLogLevel() > 0 )
+            GEOS_LOG_RANK_0( GEOS_FMT( "{}: shortening time step to {} to cover ramaining time {} in two steps", getName(), nextDt, dtRemaining ));
+        }
       }
       else
       {
         nextDt = dtRemaining;
+        if( m_nonlinearSolverParameters.getLogLevel() > 0 )
+          GEOS_LOG_RANK_0( GEOS_FMT( "{}: shortening time step to {} to match ramaining time", getName(), nextDt ));
```
ryar9534:
```suggestion
          GEOS_LOG_RANK_0( GEOS_FMT( "{}: shortening time step to {} to match remaining time", getName(), nextDt ));
```

src/coreComponents/physicsSolvers/SolverBase.cpp
```diff
@@ -271,11 +271,17 @@ bool SolverBase::execute( real64 const time_n,
       {
         // better to do two equal steps than one big and one small (even potentially tiny)
         if( nextDt * 2 > dtRemaining )
+        {
           nextDt = dtRemaining / 2;
+          if( m_nonlinearSolverParameters.getLogLevel() > 0 )
+            GEOS_LOG_RANK_0( GEOS_FMT( "{}: shortening time step to {} to cover ramaining time {} in two steps", getName(), nextDt, dtRemaining ));
```
ryar9534:
```suggestion
            GEOS_LOG_RANK_0( GEOS_FMT( "{}: shortening time step to {} to cover remaining time {} in two steps", getName(), nextDt, dtRemaining ));
```

ryar9534 (approved):

dkachuma (approved):

Issue #2876, opened 2023-12-07
Some checks in testMultiFluid are effectively inactive


paveltomin:
Some logic mismatch was introduced in
https://github.com/GEOS-DEV/GEOS/pull/278
so effectively `testNumericalDerivatives` check that uses `checkRelativeErrorFormat` and condition inside:
`if (delta > absTol && delta > relTol * value)`
is not triggered because default `absTol = std::numeric_limits<real64>::max()`
Current version of GEOS produces huge diffs that are hidden.

paveltomin:
@dkachuma just FYI

paveltomin:
fixed by https://github.com/GEOS-DEV/GEOS/pull/3140

PR #3140, opened 2024-05-23
Break up multi fluid test

dkachuma:
This PR breaks up the `testMultiFluid` unit test. This test does a full fluid update call and tests the obtained values of fluid properties against expected values. It also does numerical derivative checks at selected fluid conditions (pressure, temperature and composition).

This test was getting larger and taking longer to build especially with CUDA. Instead of having all the fluid types in one test, this is now broken up into different tests for the different multiphase fluid types. The common part is factored out into the `MultiFluidTest` test fixture.

The numerical derivative tests in this case were not active because a very large absolute tolerance was used (see #2876). These are now activated for most of the fluid types. As a consequence, a fix on the temperature derivative was required in `EzrokhiBrineDensity.hpp`. The tests are not activated for the `CompositionalMultiphaseFluidPVTPackage` because this already uses finite differences to calculate derivatives and for some of the test compositions, it is actually unable to give reliable values (the package itself needs updating).

The inclusion of `testCompFlowUtils.hpp` in  `constitutiveTestHelpers.hpp` was pulling in too many include files into the compilation of thee tests. The methods required for the multiphase fluid tests have now been moved to a smaller include `testFlowUtils.hpp`.

This will likely require a rebaseline due to the change in the Ezrokhi density derivative.


src/coreComponents/constitutive/fluid/multifluid/CO2Brine/CO2BrineFluid.cpp
```diff
@@ -194,11 +194,13 @@ void CO2BrineFluid< PHASE1, PHASE2, FLASH >::checkTablesParameters( real64 const
 template< typename PHASE1, typename PHASE2, typename FLASH >
 void CO2BrineFluid< PHASE1, PHASE2, FLASH >::initializePreSubGroups()
 {
+#if defined(GEOS_DEVICE_COMPILE)
```
paveltomin, line 197: opposite or I am mssing something?

paveltomin (approved):

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/CO2BrineFluid.cpp
dkachuma: The error is supposed to be issued when it's a GPU run. The model is now activated for CPU only runs.

dkachuma:
The fix to the derivative in `EzrokhiBrineDensity` has changed the response in `class09_pb3_smoke_3d`. The change in pressure is slightly different and quite expected. I would say a rebaseline is OK.
```
Error: /Problem/domain/MeshBodies/mesh1/meshLevels/Level0/ElementRegions/elementRegionsGroup/reservoir/elementSubRegions/1_hexahedra/deltaPressure
	Arrays of types float64 and float64 have 125 values of which 8 fail both the relative and absolute tests.
		Max absolute difference is at index (107,): value = 15333.86215468496, base_value = 15333.859636742622
		Max relative difference is at index (56,): value = 43.7011882327497, base_value = 43.70022329688072
```

PR #3142, opened 2024-05-24
Adding field to output total strain

### LINKED ISSUE #3210: Elastic/Plastic strain outputs in GEOS
The output of mechanical strain in addition to displacements from GEOS has been requested, and I can do/have done some implementation as I have availability. 

Output of total strain has been implemented in develop as of PR #3142 .

Next step is the output of a decomposed elastic/plastic strain (really just need one or the other, and the remaining can be recovered with the total strain).

As the constitutive laws for plastic materials have different implementations (and even different theory), I propose we start with some that are perhaps more widely used in our cases for maximum efficiency.

If you would like to request a specific constitutive model or discuss priorities, we can do so here.

ryar9534:
This PR adds registers total strain as a field within solid mechanics solvers so that it can be output to vtk. An incremental strain is updated when strain increments are computed within the quadrature point kernel, and these are added to the total stran at the end of the time step. We may want to rethink this once the new packing/data interface is in place.

I believe rebaselining will be needed as there is new output, but none of the results should change. 

I also propose we have the TotalEnergies folks propose which constitutive laws they use most frequently and from which they would like elastic/plastic strain output as well, and that can be incrementally added. 

An example output from my staircase runs which shows x displacement and the corresponding XX strain:
Screen Shot 2024-05-26 at 3 47 09 PM
Screen Shot 2024-05-26 at 3 47 27 PM


paveltomin (approved):

src/coreComponents/physicsSolvers/multiphysics/poromechanicsKernels/MultiphasePoromechanics_impl.hpp
```diff
@@ -642,6 +642,15 @@ quadraturePointKernel( localIndex const k,
   LvArray::tensorOps::fill< 6 >( stack.strainIncrement, 0.0 );
   FE_TYPE::symmetricGradient( dNdX, stack.uhat_local, stack.strainIncrement );
 
+  for( int is = 0; is < 6; ++is )
+  {
+    if( q == 0 )
```
CusiniM: why?

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsFields.hpp
```diff
@@ -66,6 +74,22 @@ DECLARE_FIELD( incrementalDisplacement,
...
 
+DECLARE_FIELD( strain,
+               "strain",
+               array2dLayoutStrain,
+               0,
+               LEVEL_0,
+               WRITE_AND_READ,
+               "Average strain in cell" );
+
+DECLARE_FIELD( incrementalStrain,
```
CusiniM: do we really need both of these fields? Isn't the strain enough?

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.cpp
```diff
@@ -933,6 +939,16 @@ void SolidMechanicsLagrangianFEM::implicitStepComplete( real64 const & GEOS_UNUS
       string const & solidMaterialName = subRegion.template getReference< string >( viewKeyStruct::solidMaterialNamesString() );
       SolidBase & constitutiveRelation = getConstitutiveModel< SolidBase >( subRegion, solidMaterialName );
       constitutiveRelation.saveConvergedState();
+
+      solidMechanics::arrayView2dLayoutIncrStrain incStrain = subRegion.getField< solidMechanics::incrementalStrain >();
+      solidMechanics::arrayView2dLayoutStrain strain = subRegion.getField< solidMechanics::strain >();
+      for( localIndex k = 0; k < subRegion.size(); ++k )
```
CusiniM: This should be a raja loop.

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.cpp
```diff
@@ -1316,6 +1332,15 @@ void SolidMechanicsLagrangianFEM::resetStateToBeginningOfStep( DomainPartition &
...
       }
     } );
+
+    ElementRegionManager & elementRegionManager = mesh.getElemManager();
+    elementRegionManager.forElementSubRegions< CellElementSubRegion >( regionNames,
+                                                                       [&]( localIndex const,
+                                                                            CellElementSubRegion & subRegion )
+    {
+      solidMechanics::arrayView2dLayoutIncrStrain incStrain = subRegion.getField< solidMechanics::incrementalStrain >();
+      incStrain.zero();
```
CusiniM:
```suggestion
 subRegion.getField< solidMechanics::incrementalStrain >().zero();
```

CusiniM: if you are setting it to zero here I am not sure I see the point of that `if (q == 0)`. I also think that this means that the field will always be zero when we plot it so I would only store the strain and not the incremental one.

src/coreComponents/physicsSolvers/solidMechanics/kernels/FixedStressThermoPoromechanics_impl.hpp
```diff
@@ -133,6 +135,14 @@ quadraturePointKernel( localIndex const k,
 
   FE_TYPE::symmetricGradient( dNdX, stack.uhat_local, strainInc );
 
+  for( int is = 0; is < 6; ++is )
+  {
+    if( q == 0 )
```
CusiniM: see previous comment.

src/coreComponents/physicsSolvers/solidMechanics/kernels/ImplicitSmallStrainQuasiStatic_impl.hpp
```diff
@@ -128,6 +129,15 @@ void ImplicitSmallStrainQuasiStatic< SUBREGION_TYPE, CONSTITUTIVE_TYPE, FE_TYPE
 
   FE_TYPE::symmetricGradient( dNdX, stack.uhat_local, strainInc );
 
+  for( int is = 0; is < 6; ++is )
+  {
+    if( q == 0 )
+    {
+      m_incStrain[k][is] = 0.0;
+    }
+    m_incStrain[k][is] += strainInc[is]*detJxW/m_elementVolume[k];
```
CusiniM: same code seems to be repeated a few times.

ryar9534: I agree - Is there a place you would recommend placing a function which does the strain tracking? Its seems like a lot of the quadrature point kernel is kind of repeated in the various places it is used.

ryar9534:
Thanks for the review @CusiniM . You picked up on the things I was hoping to ask about. In particular, the need for both incremental strain and total strain, and the zeroing I have to do - It seemed to me that the strainInc used in the quadraturePointKernel is an increment from the previous converged time step to the current state, so I don't think you can just sum all of the increments across newton iterations, if that makes sense. I think if you dont do the zeroing in the quadrature point kernel then the results directly depend on the number of newton iterations, or at least thats what I saw during testing. I think a separate variable for the increment is needed for a similar reason, but perhaps we can think of something more clever.

CusiniM:
> Thanks for the review @CusiniM . You picked up on the things I was hoping to ask about. In particular, the need for both incremental strain and total strain, and the zeroing I have to do - It seemed to me that the strainInc used in the quadraturePointKernel is an increment from the previous converged time step to the current state, so I don't think you can just sum all of the increments across newton iterations, if that makes sense. I think if you dont do the zeroing in the quadrature point kernel then the results directly depend on the number of newton iterations, or at least thats what I saw during testing. I think a separate variable for the increment is needed for a similar reason, but perhaps we can think of something more clever.

Hmm, if instead of using `u_hat`, we use `u` don't we get directly the total strain? At that point there is no need to store the incremental one.

CusiniM:
> > Thanks for the review @CusiniM . You picked up on the things I was hoping to ask about. In particular, the need for both incremental strain and total strain, and the zeroing I have to do - It seemed to me that the strainInc used in the quadraturePointKernel is an increment from the previous converged time step to the current state, so I don't think you can just sum all of the increments across newton iterations, if that makes sense. I think if you dont do the zeroing in the quadrature point kernel then the results directly depend on the number of newton iterations, or at least thats what I saw during testing. I think a separate variable for the increment is needed for a similar reason, but perhaps we can think of something more clever.
> 
> Hmm, if instead of using `u_hat`, we use `u` don't we get directly the total strain? At that point there is no need to store the incremental one.

I am also still not entirely convinced that it's not worth it to launch a simpler kernel just for this computation at the end of the timestep instead of trying to combine it with the assembly. I think we would only need to dispatch on the elem type since all we need is to know how to perform the interpolation and the integration and we don't care about the constitutive type (I think).

ryar9534:
@CusiniM Should be good to go now if youd like to take a look.

PR #3147, opened 2024-05-29
Fix #3146 - Error in stat reduction for Function triggered `PeriodicEvent`

### LINKED ISSUE #3146: Function triggered `PeriodicEvent` does not work properly in parallel
**Describe the bug**
When set-up like this on a central region `c2`
```xml
    <PeriodicEvent
      name="ConditionalDump"
            logLevel="2"
            maxEventDt="0.1"
            function="sg"
            threshold="0.1"
            object="/Problem/domain.../region2/elementSubRegions/c2"
            stat="0" // min
            target="/Outputs/vtkOutput"/>
```
in an adaptation of `buckleyLeverett_base` the _min_ and _average_ seems to be missing in parallel.
The `max` (resp. option `stat="2"`) return 1e10.

There is few adjustment to be made both in `PeriodicEvent` and `FunctionBase` to correct the behavior (see #3147)

jafranc:
Fix #3146 - Error in stat reduction for Function triggered `PeriodicEvent`

Using such a XML block as 
```xml
    <PeriodicEvent
      name="ConditionalDump"
            logLevel="2"
      maxEventDt="0.1"
            function="sg"
            threshold="0.1"
            object="/Problem/.../mesh/.../c2"
            stat="0"
            target="/Outputs/vtkOutput"/>
```

Attempting to use `function` attribute of `Periodic` event and its related attribute `stat` to trigger periodic event if min (0 is for min, 1 is for average, 2 is for max) of the `function` is above `threshold` was not working properly:
- in parallel
- if `stat=2`, i.e. for maximum

indeed, 
- reduction for parallel runs other than max were not implemented.
- a typo in `FunctionBase.cpp` was comparing min to max.

Moreover, for partitions not intersecting (`N=0` in _FunctionBase.cpp_)
with `object` the workflow was returning `1e10` for min and `-1e10` for max in parallel reductions.


paveltomin:
@jafranc can you explain a little bit what was the issue and how you got it fixed?

jafranc:
> @jafranc can you explain a little bit what was the issue and how you got it fixed?

I updated the description above. I am reporting Oleg experience and fix there.

paveltomin (approved):

PR #3152, opened 2024-05-30
Some random things

### LINKED ISSUE #3155: Enhance report of negative time in simulation log
In the log file we get a report of the current time in the form
```
- Time:       131d, 09h36m00s (11352960 s)
```
This is fine as long as the current time is positive. When the current time is negative, this looks like
```
- Time:       -569y, -227d, 00h-25m-12s (-17975520000 s)
```
I find this quite confusing.

Can we simplify this and have a single negative symbol such as:
```
- Time:       - (569y, 222d, 18h00m00s) (-17975520000 s)
```

paveltomin:
1. Print percentage of simulation complete:
```
------------------- TIMESTEP START -------------------
    - Time:       11h31m46s (41506.95883192398 s) out of 12h00m00s (43200 s) (96%)
    - Delta Time: 00h01m40s (100 s)
    - Cycle:      2167
------------------------------------------------------
```
2. Make `minScalingFactor` modifiable from input.
3. Some cosmetic things in output.
4. Changes the format for negative time in `units::TimeFormatInfo` so that this is printed as (see #3155):
```
    - Time:       -(11h31m46s) (-41506.95883192398 s) (96%)
```


src/coreComponents/events/EventManager.cpp
```diff
@@ -227,15 +227,29 @@ bool EventManager::run( DomainPartition & domain )
...
-                             units::TimeFormatInfo::fromSeconds( m_time ),
-                             units::TimeFormatInfo::fromSeconds( m_dt ),
-                             m_cycle ));
+  GEOS_LOG_RANK_0( "\n------------------- TIMESTEP START -------------------" );
+  if( m_maxTime < std::numeric_limits< real64 >::max() )
+  {
+    GEOS_LOG_RANK_0( GEOS_FMT( "    - Time:       {} out of {} ({}%)",
+                               units::TimeFormatInfo::fromSeconds( m_time ),
+                               units::TimeFormatInfo::fromSeconds( m_maxTime ),
+                               fmt::format( "{:.{}f}", 100.0 * m_time / m_maxTime, 0 ) ) );
```
dkachuma: I think this needs to include `m_minTime` for example:

- `m_time` might be negative
- `m_maxTime` might be zero.

dkachuma: Could the text here be extremely long? For example simulation up to 1000 years.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -150,6 +150,12 @@ CompositionalMultiphaseBase::CompositionalMultiphaseBase( const string & name,
     setInputFlag( InputFlags::OPTIONAL ).
     setApplyDefaultValue( 1.0 ).
     setDescription( "Maximum (absolute) component density change in a sequential iteration, used for outer loop convergence check" );
+
+  this->registerWrapper( viewKeyStruct::minScalingFactorString(), &m_minScalingFactor ).
+    setSizedFromParent( 0 ).
+    setInputFlag( InputFlags::OPTIONAL ).
+    setApplyDefaultValue( 0.01 ).
+    setDescription( "Minimum value for solution scaling factor" );
```
dkachuma, line 158: I think we need a validator for this to ensure that the user doesn't provide a negative value or a value greater than 1.

dkachuma (commented):
I think the calculation of the percentage needs to take into account cases where `minTime` is negative. I have been working with these cases on SPE11C and I added issue #3155 which is somehow related. Maybe it's a simple change to `units::TimeFormatInfo::fromSeconds` which can be added to this PR.

src/coreComponents/events/EventManager.cpp
paveltomin: I removed `m_maxTime` output and included `m_minTime`
For https://github.com/GEOS-DEV/GEOS/issues/3155 - I agree it is confusing and I don't really like the mixed format with years months etc going together to be honest
Would just report in years or in months or in days or etc...
@dkachuma feel free to modify this PR directly, I don't mind at all

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
paveltomin: added, thanks

dkachuma (approved):

src/coreComponents/events/EventManager.cpp
dkachuma: Pushed a change to `Units.cpp`. @MelReyCG Can you please take a look.

MelReyCG: Hi there,
The mixed format is really helping depending on the use case:
- the `y, d, h/m/s` time is useful for users to quickly read and understand the simulation time (they are more used to `100y 0d, 00h00m00s` than to `3153600000s`),
- the second format is ideal and for sub-second simulation times and debugging purposes,
- the percentage you added is a good idea to quickly see the simulation progression.

The problem is that sometimes, we need all formats within the same log, and we can have simulations that can take 24h+, so changing the desired format in the `xml` is not an option.

We could have a separated second format:
```
------------------- TIMESTEP START -------------------
    - Time:       11h31m46s out of 12h00m00s  (96%)
                  ( 41506.95883192398 s / 43200 s )
    - Delta Time: 00h01m40s (100 s)
    - Cycle:      2167
------------------------------------------------------
```

... or we could separate the ending time:
```
------------------- TIMESTEP START -------------------
    - Time:       11h31m46s ( 41506.95883192398 s )
                  96% out of 12h00m00s ( 43200 s )
    - Delta Time: 00h01m40s (100 s)
    - Cycle:      2167
------------------------------------------------------
```

I'll get in touch with the users here and potentially get back with more feedback.

src/coreComponents/common/Units.cpp
```diff
@@ -37,6 +37,11 @@ TimeFormatInfo::TimeFormatInfo( double const totalSeconds, int const years, int
 string TimeFormatInfo::toString() const
 {
   std::ostringstream oss;
+  constexpr double negativeTime = -1.0 + LvArray::NumericLimits< double >::epsilon;
+  if( m_totalSeconds < negativeTime )
+  {
+    oss << "-(";
+  }
```
MelReyCG: I would like to propose two alternatives for here:
- a `if( m_totalSeconds < 0.0 )` to let clearly appear that the time is still negative,
- or a `if( m_totalSeconds <= -1.0 )`, which would be more simple & precise, as `-1.0+epsilon` leave room for a sneaky value.

MelReyCG (commented):
A minor comment

src/coreComponents/events/EventManager.cpp
MelReyCG: Hi again @paveltomin,
Here, users have voted for the first alternative I posted, where formats are not mixed within the same line :)

```
------------------- TIMESTEP START -------------------
    - Time:       11h31m46s out of 12h00m00s  (96%)
                  ( 41506.95883192398 s / 43200 s )
    - Delta Time: 00h01m40s (100 s)
    - Cycle:      2167
------------------------------------------------------
```

If you'd like to, I can push what is needed in your PR to set it up.

paveltomin: Yes, please go ahead and do that. Thank you!

src/coreComponents/common/Units.cpp
dkachuma: Initially I had `if( m_totalSeconds < 0.0 )` which was printing something like
```
-(00h00m00s) (-1.11e-07 s)
```
I didn't like the negative sign here. But I agree that it might be it is good to highlight that the time is negative.

src/coreComponents/common/Units.cpp
```diff
@@ -37,6 +37,11 @@ TimeFormatInfo::TimeFormatInfo( double const totalSeconds, int const years, int
 string TimeFormatInfo::toString() const
 {
   std::ostringstream oss;
+  constexpr double negativeTime = -1.0 + LvArray::NumericLimits< double >::epsilon;
```
dkachuma:
```suggestion
```

src/coreComponents/common/Units.cpp
```diff
@@ -37,6 +37,11 @@ TimeFormatInfo::TimeFormatInfo( double const totalSeconds, int const years, int
...
+  if( m_totalSeconds < negativeTime )
```
dkachuma:
```suggestion
  if( m_totalSeconds < 0.0 )
```

src/coreComponents/common/Units.cpp
```diff
@@ -45,8 +50,12 @@ string TimeFormatInfo::toString() const
   {
     oss << m_days << "d, ";
   }
-  oss << GEOS_FMT( "{:0>2}h{:0>2}m{:0>2}s ({} s)",
-                   m_hours, m_minutes, m_seconds, m_totalSeconds );
+  oss << GEOS_FMT( "{:0>2}h{:0>2}m{:0>2}s", m_hours, m_minutes, m_seconds );
+  if( m_totalSeconds < negativeTime )
```
dkachuma:
```suggestion
  if( m_totalSeconds < 0.0 )
```

src/coreComponents/common/unitTests/testUnits.cpp
```diff
@@ -112,6 +112,26 @@ TEST( Units, SystemDurationFormatTest )
       "5500y, 20d, 12h02m25s (173565007345 s)",
       Years( 5500 ) + Days( 20 ) + hours( 12 ) + seconds( 145 ) ),
 
+    DurationCase(
+      "00h00m00s (-1.11e-07 s)",
```
dkachuma:
```suggestion
      "-(00h00m00s) (-1.11e-07 s)",
```

dkachuma (approved):

MelReyCG:
Hi @paveltomin,
Just to notify that I pushed, here is a simulated exemple:
```
------------------------- TIMESTEP START -------------------------
    - Time:       12y, 119d, 09h42m56s out of 15y, 308d, 09h35m20s (78% completed)
                  (389000000 s / 500000000 s)
    - Delta Time: 5d, 18h53m20s (500000 s)
    - Cycle:      792
--------------------------------------------------------------------
```
Don't hesitate to give feedback on this

paveltomin:
> Don't hesitate to give feedback on this

Looks good to me, thanks
@dkachuma what do you think?

MelReyCG:
Just suggestions in order to highlight the current timestep:
- should we use ` / ` instead of ` out of ` (it could be more easily seen)?
- should we jump another line just before the ` out of `?

MelReyCG (approved):
Everything looks good to me anyway. I or Arnaud will come back on the formatting of the timestep when we will add the log-sections.

MelReyCG (approved):
Everything looks good to me anyway. I or Arnaud will come back on the formatting of the timestep when we will add the log-sections, so if you have an concern, don't hesitate to leave them in here.

paveltomin:
> Just suggestions in order to highlight the current timestep:
> 
> * should we use `/` instead of `out of` (it could be more easily seen)?
> * should we jump another line just before the `out of`?

`/` is fine, and I would prefer less lines

PR #3156, opened 2024-06-03
feat: feature/byer3/thermal well

tjb-ltk:
Add thermal well formulation for modeling convection in compositional flow models


src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
```diff
@@ -863,12 +909,23 @@ void CompositionalMultiphaseWell::updatePhaseVolumeFraction( WellElementSubRegio
...
+                                                 subRegion,
+                                                 fluid )
+:    isothermalCompositionalMultiphaseBaseKernels::
+      PhaseVolumeFractionKernelFactory::
+      createAndLaunch< parallelDevicePolicy<> >( m_numComponents,
+                                                 m_numPhases,
+                                                 subRegion,
+                                                 fluid );
+
+  maxDeltaPhaseVolFrac = MpiWrapper::max( maxDeltaPhaseVolFrac );
```
paveltomin: this should be better outside of `subRegion` section
we observed bad parallel degradation for flow solver when similar sync was placed inside subregion loop

src/coreComponents/linearAlgebra/interfaces/hypre/mgrStrategies/ThermalCompositionalMultiphaseReservoirFVM.hpp
```diff
@@ -0,0 +1,122 @@
...
+ *
+ */
+class ThermalCompositionalMultiphaseReservoirFVM : public MGRStrategyBase< 3 >
+{
+public:
+  /**
+   * @brief Constructor.
+   * @param numComponentsPerField array with number of components for each field
+   */
+  explicit ThermalCompositionalMultiphaseReservoirFVM( arrayView1d< int const > const & numComponentsPerField )
```
paveltomin, line 65: CC @castelletto1 and @victorapm and @jafranc

src/coreComponents/physicsSolvers/fluidFlow/ThermalCompositionalMultiphaseBaseKernels.hpp
```diff
@@ -800,24 +817,27 @@ class SolutionCheckKernelFactory
                    integer const allowNegativePressure,
                    CompositionalMultiphaseFVM::ScalingType const scalingType,
                    real64 const scalingFactor,
+                   arrayView1d< real64 const > const pressure,
```
paveltomin, line 827: why you prefer this instead of old way?

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
```diff
@@ -833,20 +854,35 @@ void CompositionalMultiphaseWell::updateVolRatesForConstraint( WellElementSubReg
...
 
   string const & fluidName = subRegion.getReference< string >( viewKeyStruct::fluidNamesString() );
   MultiFluidBase & fluid = subRegion.getConstitutiveModel< MultiFluidBase >( fluidName );
 
+  arrayView3d< real64 const, multifluid::USD_PHASE > m_phaseInternalEnergy=fluid.phaseInternalEnergy();
+  arrayView4d< real64 const, multifluid::USD_PHASE_DC > m_dPhaseInternalEnergy=fluid.dPhaseInternalEnergy();
+  std::cout << subRegion.getName() << std::endl;
+
+  for( integer i=0; i<subRegion.size(); i++ )
+    std::cout << i << " tpupd0 " << temp[i] << " " <<  pres[i]   << " rate " << connRate[i] << " compfrac " << compFrac[i][0] << " " << compFrac[i][1] <<
```
paveltomin: remove

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
```diff
@@ -878,14 +935,47 @@ void CompositionalMultiphaseWell::updateTotalMassDensity( WellElementSubRegion &
...
+                                               fluid )
+  :
   TotalMassDensityKernelFactory::
     createAndLaunch< parallelDevicePolicy<> >( m_numComponents,
                                                m_numPhases,
                                                subRegion,
                                                fluid );
+
 }
+void CompositionalMultiphaseWell::assembleSystem( real64 const time,
```
paveltomin: why this becomes different with WellSolverBase::assembleSystem ?

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
```diff
@@ -931,280 +1019,393 @@ void CompositionalMultiphaseWell::initializeWells( DomainPartition & domain )
...
-
-      // get well secondary variables on well elements
-      string const & fluidName = subRegion.getReference< string >( viewKeyStruct::fluidNamesString() );
-      MultiFluidBase & fluid = getConstitutiveModel< MultiFluidBase >( subRegion, fluidName );
-      arrayView3d< real64 const, multifluid::USD_PHASE > const & wellElemPhaseDens = fluid.phaseDensity();
-      arrayView2d< real64 const, multifluid::USD_FLUID > const & wellElemTotalDens = fluid.totalDensity();
 
-      // 4) Back calculate component densities
-      constitutive::constitutiveUpdatePassThru( fluid, [&] ( auto & castedFluid )
+      if( time_n <= 0.0 ||
```
paveltomin, line 984: why?
what about wells that open later for time>0?

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
```diff
@@ -931,280 +1019,393 @@ void CompositionalMultiphaseWell::initializeWells( DomainPartition & domain )
...
+
 }
 
 
-void CompositionalMultiphaseWell::assembleVolumeBalanceTerms( DomainPartition const & domain,
-                                                              DofManager const & dofManager,
-                                                              CRSMatrixView< real64, globalIndex const > const & localMatrix,
-                                                              arrayView1d< real64 > const & localRhs )
+real64
+CompositionalMultiphaseWell::calculateResidualNorm1( real64 const & time_n,
```
paveltomin: rename?

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
```diff
@@ -1279,33 +1479,134 @@ CompositionalMultiphaseWell::scalingForSystemSolution( DomainPartition & domain,
   string const wellDofKey = dofManager.getKey( wellElementDofName() );
 
   real64 scalingFactor = 1.0;
-  forDiscretizationOnMeshTargets( domain.getMeshBodies(), [&] ( string const &,
-                                                                MeshLevel & mesh,
-                                                                arrayView1d< string const > const & regionNames )
+  if( 0 )
```
paveltomin: why disabled?

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
```diff
@@ -1467,26 +1834,45 @@ CompositionalMultiphaseWell::applySystemSolution( DofManager const & dofManager,
...
+  DofManager::CompMask connRateMask( m_numDofPerWellElement, numFluidComponents()+1, numFluidComponents()+2 );
   GEOS_UNUSED_VAR( dt );
   // update all the fields using the global damping coefficients
   dofManager.addVectorToField( localSolution,
                                wellElementDofName(),
                                fields::well::pressure::key(),
                                scalingFactor,
-                               { m_numDofPerWellElement, 0, 1 } );
+                               pressureMask );
+  //{ m_numDofPerWellElement, 0, 1 } );
```
paveltomin: remove commented

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
tjb-ltk: full check is      if( time_n <= 0.0 ||
          ( !wellControls.isWellOpen( time_n ) && wellControls.isWellOpen( time_n + dt ) )  ) Previously a well was only initialized at time <=0.   the or part allows initialization for a well that is currently closed and next dt its scheduled to be opened.  Models containing wells using rate tables with first nonzero rate at t>0 were failing to converge.

paveltomin: i see, yeah, that part is tricky...

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
tjb-ltk: Previously well solver base assemble system cold drive both single phase and compositional runs (same calling sequence  assembleAccum, assemblevolbal ....).  The composiitonal Jac assemble has been changed... looks a bit more like the reservoir assemlbeSystem , supports thermal, and the accumulation and vol bal kernels have been combined.. Thermal hasn't been added to single phase yet  so single phase and compositional need there own assembleSystem..

paveltomin: i had a second look and the two `assembleSystem` versions for single phase and compositional look the same after your changes
could you please double check that everything is correct?

paveltomin:
@tjb-ltk is it ready to be cleaned up, re-reviewed and merged?

tjb-ltk:
No.. in training ... probably by Tuesday

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
tjb-ltk: removed in latest push

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
tjb-ltk: you are correct.  in latest push they are moved from specializations and wellsolverbase has assemble system

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
tjb-ltk: moved comm to outside subregion loop in latest push

src/coreComponents/physicsSolvers/fluidFlow/wells/PerforationFluxKernels.hpp
```diff
@@ -0,0 +1,890 @@
...
+
+struct NoOpStuct
+{
+  NoOpStuct(){}
+};
+
+namespace isothermalPerforationFluxKernels
+{
+
+using namespace constitutive;
```
CusiniM, line 54: please remove this. We should avoid `using namespace` in headers.

src/coreComponents/physicsSolvers/fluidFlow/wells/PerforationFluxKernels.hpp
```diff
@@ -0,0 +1,890 @@
...
+    /// Views on energy flux
+    arrayView1d< real64 > const & m_energyPerfFlux;
+    arrayView3d< real64 > const & m_dEnergyPerfFlux;
+
+    /// Views on temperature
+    ElementViewConst< arrayView1d< real64 const > > const m_temp;
+
+    /// Views on phase enthalpies
+    ElementViewConst< arrayView3d< real64 const, multifluid::USD_PHASE > > const m_resPhaseEnthalpy;
+    ElementViewConst< arrayView4d< real64 const, multifluid::USD_PHASE_DC > > const m_dResPhaseEnthalpy;
```
CusiniM: The idea of the `StackVariables` class is to contain all stack allocated variables. `ArrayViews` are not stack allocated and they actually cannot be constructed like this in a DEVICE function. I would remove this StackVariables object and make all these views members of your Kernel class. They will then be capture by value in the kernelLaunch lambda. 


if you do this, the CUDA compilation error you are seeing should disappear.

src/coreComponents/physicsSolvers/fluidFlow/wells/PerforationFluxKernels.hpp
CusiniM: please remove this. We should avoid `using namespace` in headers.

src/coreComponents/physicsSolvers/fluidFlow/wells/PerforationFluxKernels.hpp
```diff
@@ -0,0 +1,890 @@
...
+    /**
+     * @brief Constructor for the stack variables
+     */
+
+    GEOS_HOST_DEVICE
+    StackVariables() {}
+
+  };
+
+  template< typename STRUCT = NoOpFunc >
```
CusiniM:
```suggestion
  template< typename FUNC = NoOpFunc >
```
?

src/coreComponents/physicsSolvers/fluidFlow/wells/PerforationFluxKernels.hpp
```diff
@@ -0,0 +1,890 @@
...
+  GEOS_HOST_DEVICE
+  inline
+  void
+  computeFlux( localIndex const iperf, STRUCT && stack= NoOpFunc{} ) const
```
CusiniM:
```suggestion
  computeFlux( localIndex const iperf, StackVariables & stack, FUNC && assemblyKernelOp = NoOpFunc{}) const
```

?
But do you really need that callback? In the compositional flux kernels it is used to add additional operations like those needed to add thermal effects. Here, you seem to be doing it with that `if constexpr` (which I think is a good approach) so I don't think you need the callback at all.

CusiniM (changes requested):

src/coreComponents/physicsSolvers/fluidFlow/wells/PerforationFluxKernels.hpp
tjb-ltk: StackVariables was used to support constexpr for thermal parts in isothermal compute.  Not sure if it can be removed, unless dispatch is used and constexpr is removed.   I like the inlining that constexpr gives you... I'll try initializing the stack every time in the icon loop as in res flux,  originally thought this is overhead so did it on class construction

CusiniM: oh, I see. It can't really be done like that. So, yes you will have to either use that callback design to assemble the two cases. 
If we think that it's nice to have a single `computePerforationFlux` function with an `if constexpr `block to separate the two cases, we could do something like this https://godbolt.org/z/Ehhh5o983 but I would like to hear what @wrtobin and @rrsettgast think about this before sending you down this path.

host-configs/apple/macOS_base.cmake
```diff
@@ -6,24 +6,38 @@ message("CONFIG_NAME = ${CONFIG_NAME}")
...
-set(MPIEXEC "${HOMEBREW_DIR}/bin/mpirun" CACHE PATH "")
+#set(MPI_C_COMPILER "${HOMEBREW_DIR}/bin/mpicc" CACHE PATH "")
+#set(MPI_CXX_COMPILER "${HOMEBREW_DIR}/bin/mpicxx" CACHE PATH "")
+#set(MPI_Fortran_COMPILER "${HOMEBREW_DIR}/bin/mpifort" CACHE PATH "")
+#set(MPIEXEC "${HOMEBREW_DIR}/bin/mpirun" CACHE PATH "")
+
+set(MPI_C_COMPILER "mpicc" CACHE PATH "")
+set(MPI_CXX_COMPILER "mpicxx" CACHE PATH "")
+set(MPI_Fortran_COMPILER "mpifort" CACHE PATH "")
+set(MPIEXEC "mpirun" CACHE PATH "")
```
CusiniM: can you revert these changes? you should be able to just set a different homebrew path if you are not using the default one.

host-configs/apple/macOS_base.cmake
```diff
@@ -6,24 +6,38 @@ message("CONFIG_NAME = ${CONFIG_NAME}")
...
 
-set( BLAS_LIBRARIES ${HOMEBREW_DIR}/opt/lapack/lib/libblas.dylib CACHE PATH "" FORCE )
-set( LAPACK_LIBRARIES ${HOMEBREW_DIR}/opt/lapack/lib/liblapack.dylib CACHE PATH "" FORCE )
+#set( BLAS_LIBRARIES ${HOMEBREW_DIR}/opt/lapack/lib/libblas.dylib CACHE PATH "" FORCE )
+#set( LAPACK_LIBRARIES ${HOMEBREW_DIR}/opt/lapack/lib/liblapack.dylib CACHE PATH "" FORCE )
+
+#set( BLAS_LIBRARIES /Users/byer3/apps/opt/lapack-3.11/libblas.3.dylib CACHE PATH "" FORCE )
+#set( LAPACK_LIBRARIES /Users/byer3/apps/opt/lapack-3.11/liblapack.3.dylib CACHE PATH "" FORCE )
+set( BLAS_LIBRARIES /Users/byer3/apps/opt/lapack-3.11/libblas.dylib CACHE PATH "" FORCE )
+set( LAPACK_LIBRARIES /Users/byer3/apps/opt/lapack-3.11/liblapack.dylib CACHE PATH "" FORCE )
```
CusiniM: same for these since they only work for your machine.

host-configs/apple/macOS_base.cmake
```diff
@@ -40,5 +54,7 @@ endif()
 
 # ATS
 set(ATS_ARGUMENTS "--machine openmpi --ats openmpi_mpirun=${MPIEXEC}"  CACHE PATH "")
-
+#set(GEOSX_TPL_DIR "/Users/byer3/GEOS-DEV/thirdPartyLibs/install-mac-debug" CACHE PATH "" FORCE)
+set(GEOSX_TPL_DIR "/Users/byer3/GEOS-DEV-20240517/thirdPartyLibs/install-mac_rel-release" CACHE PATH "" FORCE)
+set(GEOSX_TPL_DIR "/Users/byer3/gd_20240517/thirdPartyLibs/install-mac_rel-release" CACHE PATH "" FORCE)
```
CusiniM: same here.

src/coreComponents/physicsSolvers/fluidFlow/wells/PerforationFluxKernels.hpp
tjb-ltk: pushed restructured version

host-configs/apple/macOS_base.cmake
tjb-ltk: done

host-configs/apple/macOS_base.cmake
tjb-ltk: done

host-configs/apple/macOS_base.cmake
tjb-ltk: done

src/coreComponents/unitTests/fluidFlowTests/testCompFlowUtils.hpp
```diff
@@ -53,6 +53,20 @@ void fillNumericalJacobian( arrayView1d< real64 const > const & residual,
   } );
 }
 
+inline
+void setNumericalJacobianValue( localIndex const rowIndex,
+                                globalIndex const colIndex,
+                                real64 const val,
+                                CRSMatrixView< real64, globalIndex > const & jacobian )
```
CusiniM:
```suggestion
                                CRSMatrixView< real64, globalIndex const > const & jacobian )
```
?

src/coreComponents/physicsSolvers/fluidFlow/wells/PerforationFluxKernels.hpp
```diff
@@ -0,0 +1,881 @@
...
+                          Deriv::dC );
+
+          for( integer jc = 0; jc < CP_Deriv::nDer; ++jc )
+          {
+            m_dCompPerfRate[iperf][TAG::RES][ic][jc]  += dFlux[TAG::RES][jc] *  m_resPhaseCompFrac[er][esr][ei][0][ip][ic];
+            m_dCompPerfRate[iperf][TAG::RES][ic][jc]  += flux * dCompFrac[jc];
+            m_dCompPerfRate[iperf][TAG::WELL][ic][jc] += dFlux[TAG::WELL][jc] *  m_resPhaseCompFrac[er][esr][ei][0][ip][ic];
+          }
+        }
+        if constexpr ( IS_THERMAL )
```
CusiniM, line 349: you don't need these if `constexpr` nor the IS_THERMAL template parameter I would use the callback to assemble the thermal terms.

src/coreComponents/physicsSolvers/fluidFlow/wells/PerforationFluxKernels.hpp
```diff
@@ -0,0 +1,881 @@
...
+  }
+
+protected:
+
+  /// Views on well element properties
+  /// Element phase fraction
+  arrayView3d< real64 const, multifluid::USD_PHASE > const m_wellElemPhaseFrac;
+  arrayView4d< real64 const, multifluid::USD_PHASE_DC > const m_dPhaseFrac;
+  arrayView3d< real64 const, multifluid::USD_PHASE >  m_wellElemPhaseEnthalpy;
+  arrayView4d< real64 const, multifluid::USD_PHASE_DC >  m_dWellElemPhaseEnthalpy;
```
CusiniM:
```suggestion
  arrayView3d< real64 const, multifluid::USD_PHASE > const m_wellElemPhaseEnthalpy;
  arrayView4d< real64 const, multifluid::USD_PHASE_DC > const m_dWellElemPhaseEnthalpy;
```

src/coreComponents/physicsSolvers/fluidFlow/wells/PerforationFluxKernels.hpp
```diff
@@ -0,0 +1,881 @@
...
+
+  /// Views on energy flux
+  arrayView1d< real64 > const & m_energyPerfFlux;
+  arrayView3d< real64 > const & m_dEnergyPerfFlux;
```
CusiniM:
```suggestion
  arrayView1d< real64 > const m_energyPerfFlux;
  arrayView3d< real64 > const m_dEnergyPerfFlux;
```

src/coreComponents/unitTests/fluidFlowTests/testCompFlowUtils.hpp
```diff
@@ -53,6 +53,21 @@ void fillNumericalJacobian( arrayView1d< real64 const > const & residual,
...
+inline
+void setNumericalJacobianValue( localIndex const rowIndex,
+                                globalIndex const colIndex,
+                                real64 const val,
+                                CRSMatrixView< real64, const globalIndex > const jacobian )
+{
+  forAll< parallelDevicePolicy<> >( 1, [=] GEOS_HOST_DEVICE ( localIndex const k )
+  {
+    GEOS_UNUSED_VAR( k );
+    jacobian.addToRow< parallelDeviceAtomic >( row, &colIndex, &val, 1 );
```
CusiniM:
```suggestion
    jacobian.addToRow< parallelDeviceAtomic >( rowIndex, &colIndex, &val, 1 );
```

src/coreComponents/physicsSolvers/fluidFlow/wells/PerforationFluxKernels.hpp
paveltomin: @rrsettgast ?

tjb-ltk: The constexpr was introduced and discussed early in dev.   A fair bit of code will need to be added/refactored if this isnt going to be allowed.

src/coreComponents/physicsSolvers/fluidFlow/wells/PerforationFluxKernels.hpp
tjb-ltk: Done in merge

src/coreComponents/physicsSolvers/fluidFlow/wells/PerforationFluxKernels.hpp
tjb-ltk: done

src/coreComponents/physicsSolvers/fluidFlow/wells/PerforationFluxKernels.hpp
tjb-ltk: No, the code is more readable this way

.gitmodules
```diff
@@ -11,5 +11,6 @@
 	path = src/coreComponents/fileIO/coupling/hdf5_interface
 	url = ../../GEOS-DEV/hdf5_interface.git
 [submodule "scripts/uberenv"]
-	path = scripts/uberenv
-	url = ../../LLNL/uberenv.git
+        path = scripts/uberenv
+        url = ../../LLNL/uberenv.git
```
CusiniM: ?

src/coreComponents/common/KernelLaunchSelectors.hpp
```diff
@@ -0,0 +1,169 @@
...
+    { GEOS_ERROR( "Unsupported thermal state: " << value ); }
+  }
+}
+
+template< typename T, typename LAMBDA >
+void kernelLaunchSelectorCompThermSwitch( T value, bool const isThermal, LAMBDA && lambda )
+{
+  static_assert( std::is_integral< T >::value, "kernelLaunchSelectorCompSwitch: value type should be integral" );
+
+  //constexpr T a = isThermal ? std::integral_constant< T, 1 >() : std::integral_constant< T, 0 >();
```
CusiniM:
```suggestion
```

.gitmodules
tjb-ltk: removed tab, now same as dev

src/coreComponents/common/KernelLaunchSelectors.hpp
tjb-ltk: commented code now removed

src/coreComponents/constitutive/fluid/singlefluid/SingleFluidBase.hpp
```diff
@@ -276,6 +276,14 @@ class SingleFluidBase : public ConstitutiveBase
   virtual real64 defaultDensity() const = 0;
   virtual real64 defaultViscosity() const = 0;
 
+/**
+ * @brief Get the thermal flag.
+ * @return boolean value indicating whether the model can be used to assemble the energy balance equation or not
+ * @detail if isThermal is true, the constitutive model compute the enthalpy and internal energy of the phase.
+ *         This can be used to check the compatibility of the constitutive model with the solver
+ */
+  virtual bool isThermal() const { return false; }
```
paveltomin, line 285: do you need to add one with `true` in `ThermalCompressibleSinglePhaseFluid `?

src/coreComponents/linearAlgebra/interfaces/hypre/mgrStrategies/ThermalCompositionalMultiphaseReservoirFVM.hpp
paveltomin: @victorapm and @castelletto1 can you guys please confirm that the recipe  looks good?

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -2266,8 +2267,8 @@ void CompositionalMultiphaseBase::computeCFLNumbers( geos::DomainPartition & dom
 
     fluxApprox.forAllStencils( mesh, [&] ( auto & stencil )
     {
-
       typename TYPEOFREF( stencil ) ::KernelWrapper stencilWrapper = stencil.createKernelWrapper();
+      std::cout << "fluxApprox.forAllStencils : mesh " << mesh.getName()  << std::endl;
```
paveltomin: remove debug

src/coreComponents/linearAlgebra/interfaces/hypre/mgrStrategies/ThermalCompositionalMultiphaseReservoirFVM.hpp
victorapm: It looks good! @tjb-ltk can you just update the comments (well info is missing). We should have something like this plus temperature info https://github.com/GEOS-DEV/GEOS/blob/713046244c75a48eb198b30497c49a190433c88d/src/coreComponents/linearAlgebra/interfaces/hypre/mgrStrategies/CompositionalMultiphaseReservoirFVM.hpp#L34-L53

src/coreComponents/common/KernelLaunchSelectors.hpp
```diff
@@ -0,0 +1,167 @@
...
+    { GEOS_ERROR( "Unsupported thermal state: " << value ); }
+  }
+}
+
+template< typename T, typename LAMBDA >
+void kernelLaunchSelectorCompThermSwitch( T value, bool const isThermal, LAMBDA && lambda )
+{
+  static_assert( std::is_integral< T >::value, "kernelLaunchSelectorCompSwitch: value type should be integral" );
+
+  if( isThermal )
```
CusiniM: instead of all these nested `if` and `switch`, I think that you can do:

````c++
auto invokeLambda = [&] (auto val1, bool const isThermal) {
    if (isThermal) {
        lambda(val1, std::integral_constant<int, 1>{});
    } else {
        lambda(val1, std::integral_constant<int, 0>{});
    }
};

switch(value)
{
    case 1:
        invokeLambda(std::integral_constant<T, 1>{}, isThermal);  
        return;
    case 2:
        invokeLambda(std::integral_constant<T, 2>{}, isThermal); 
        return;
    case 3:
        invokeLambda(std::integral_constant<T, 3>{}, isThermal); 
        return;
     /// etc.   
    default:
       { GEOS_ERROR( "Unsupported number of components: " << value ); return; }
}
````
@wrtobin ?

**Review by victorapm — APPROVED** [review]
The linear solver changes look good. Just need to update the header comment for the new MGR strategy

File: src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseBaseKernels.hpp [hunk 27]
```diff
@@ -1463,20 +1467,17 @@ class ScalingForSystemSolutionKernelFactory
...
+                   arrayView2d< real64 const, compflow::USD_COMP > const compDens,
+                   arrayView1d< real64 > pressureScalingFactor,
+                   arrayView1d< real64 > compDensScalingFactor,
                    globalIndex const rankOffset,
                    integer const numComp,
                    string const dofKey,
                    ElementSubRegionBase & subRegion,
                    arrayView1d< real64 const > const localSolution )
   {
-    arrayView1d< real64 const > const pressure =
```
paveltomin, line 1470: why do you prefer this way? passing as arguments to `createAndLaunch` instead of grabbing inside?

src/coreComponents/physicsSolvers/fluidFlow/ThermalCompositionalMultiphaseBaseKernels.hpp
paveltomin: yes, why :)

paveltomin: reuse for wells?

tjb-ltk: yes reuse for wells

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
```diff
@@ -582,31 +588,37 @@ void CompositionalMultiphaseWell::updateBHPForConstraint( WellElementSubRegion &
...
+  arrayView1d< real64 > const & dCurrentBHP =
+    wellControls.getReference< array1d< real64 > >( CompositionalMultiphaseWell::viewKeyStruct::dCurrentBHPString() );
+
+  geos::internal::kernelLaunchSelectorCompThermSwitch( numComp, isThermal, [&] ( auto NC, auto ISTHERMAL )
   {
-    real64 const diffGravCoef = refGravCoef - wellElemGravCoef[iwelemRef];
-    currentBHP = pres[iwelemRef] + totalMassDens[iwelemRef] * diffGravCoef;
-    dCurrentBHP_dPres = 1 + dTotalMassDens_dPres[iwelemRef] * diffGravCoef;
-    for( integer ic = 0; ic < numComp; ++ic )
+    //integer constexpr NUM_COMP = NC();
```
paveltomin: remove commented

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
```diff
@@ -1610,77 +1887,82 @@ void CompositionalMultiphaseWell::assemblePressureRelations( real64 const & time
...
+        arrayView1d< real64 const > const & wellElemTotalMassDens =
+          subRegion.getField< fields::well::totalMassDensity >();
+        arrayView2d< real64 const, compflow::USD_FLUID_DC > const & dWellElemTotalMassDens =
+          subRegion.getField< fields::well::dTotalMassDensity >();
+
+        bool controlHasSwitched = false;
+        isothermalCompositionalMultiphaseBaseKernels::
+          KernelLaunchSelectorCompTherm< PressureRelationKernel >( numFluidComponents(), isThermal,
+                                                                   //isothermalCompositionalMultiphaseBaseKernels::KernelLaunchSelector1<
+                                                                   // PressureRelationKernel >( numFluidComponents(),
```
paveltomin: remove commented

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWellFields.hpp
```diff
@@ -113,7 +113,23 @@ DECLARE_FIELD( phaseVolumeFraction_n,
                NOPLOT,
                WRITE_AND_READ,
                "Phase volume fraction at the previous converged time step" );
+/*
+   DECLARE_FIELD( dCurrentBHP,
```
paveltomin: remove

paveltomin (approved):

src/coreComponents/common/KernelLaunchSelectors.hpp
```diff
@@ -0,0 +1,167 @@
...
+      {
+        lambda( std::integral_constant< T, 1 >(), std::integral_constant< T, 1 >() ); return;
+      }
+      case 2:
+      { lambda( std::integral_constant< T, 2 >(), std::integral_constant< T, 1 >() ); return; }
+      case 3:
+      { lambda( std::integral_constant< T, 3 >(), std::integral_constant< T, 1 >() ); return; }
+      case 4:
+      { lambda( std::integral_constant< T, 4 >(), std::integral_constant< T, 1 >() ); return; }
+      case 5:
```
CusiniM: and I would do something similar here too.

CusiniM:
````c++
auto invokeLambda = [&] (auto numComp, T numPhases) {
    if (numPhases == 1) {
        lambda(numComp, std::integral_constant<T, 1>{});
    } else if (numPhases == 2) {
        lambda(numComp, std::integral_constant<T, 2>{});
    }
    else if (numPhases == 3)
    {
    lambda(numComp, std::integral_constant<T, 3>{});
    }
    else
    { /// error}
};
````

File: src/coreComponents/dataRepository/Group.hpp [hunk 32]
```diff
@@ -879,7 +880,6 @@ class Group
 
     for( auto const & subGroupIter : m_subGroups )
     {
-      std::cout << indent << subGroupIter.second->getName() << std::endl;
```
CusiniM: why this change?

src/coreComponents/unitTests/fluidFlowTests/testCompFlowUtils.hpp
```diff
@@ -53,6 +53,21 @@ void fillNumericalJacobian( arrayView1d< real64 const > const & residual,
...
+                                globalIndex const colIndex,
+                                real64 const val,
+                                CRSMatrixView< real64, const globalIndex > const jacobian )
+{
+  forAll< parallelDevicePolicy<> >( 1, [=] GEOS_HOST_DEVICE ( localIndex const k )
+  {
+    GEOS_UNUSED_VAR( k );
+    jacobian.addToRow< parallelDeviceAtomic >( rowIndex, &colIndex, &val, 1 );
+    // jacobian.removeNonZero( rowIndex, colIndex );
+    // jacobian.insertNonZero( rowIndex, colIndex, val );
```
CusiniM:
```suggestion
    // jacobian.removeNonZero( rowIndex, colIndex );
    // jacobian.insertNonZero( rowIndex, colIndex, val );
```

src/coreComponents/unitTests/wellsTests/testThermalReservoirCompositionalMultiphaseSSWells.cpp
CusiniM, line 1: we really need to start removing tests from this folder instead of adding new ones.

src/coreComponents/physicsSolvers/fluidFlow/wells/PerforationFluxKernels.hpp
CusiniM: but his by default is a no op so that IS_THERMAL is not needed.

src/coreComponents/physicsSolvers/fluidFlow/wells/PerforationFluxKernels.hpp
```diff
@@ -0,0 +1,881 @@
...
+        {
+          m_dCompPerfRate[iperf][TAG::WELL][ic][CP_Deriv::dT] = m_wellElemCompFrac[iwelem][ic] * dFlux[TAG::WELL][CP_Deriv::dT];
+        }
+        for( integer jc = 0; jc < NC; ++jc )
+        {
+          m_dCompPerfRate[iperf][TAG::WELL][ic][CP_Deriv::dC+jc] += m_wellElemCompFrac[iwelem][ic] * dFlux[TAG::WELL][CP_Deriv::dC+jc];
+          m_dCompPerfRate[iperf][TAG::WELL][ic][CP_Deriv::dC+jc] += m_dWellElemCompFrac_dCompDens[iwelem][ic][jc] * flux;
+        }
+      }
+      if constexpr ( IS_THERMAL )
```
CusiniM, line 475: same comment here

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWellFields.hpp
```diff
@@ -113,7 +113,23 @@ DECLARE_FIELD( phaseVolumeFraction_n,
...
+               "Derivative of current BHP with respect to pressure, temperature, and global component density" );
 
+   DECLARE_FIELD( dCurrentPhaseVolRate,
+               "dCurrentPhaseVolRate",
+               array2dLayoutPhase,
+               0,
+               NOPLOT,
+               NO_WRITE,
+               "Derivative of current phase volume rate with respect to pressure, temperature, and global component density" );
+ */
```
CusiniM: remove

src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseBaseKernels.hpp
```diff
@@ -2445,6 +2454,54 @@ void KernelLaunchSelector2( integer const numComp, integer const numPhase, ARGS
   }
 }
 
+template< typename KERNELWRAPPER, typename ... ARGS >
+void KernelLaunchSelector_NC_NP_THERM( integer const numComp, integer const numPhase, integer const isThermal, ARGS && ... args )
```
CusiniM, line 2456: why did you have to add this? didn't we already have multiphase compostional thermal kernels?

src/coreComponents/dataRepository/Group.hpp
tjb-ltk: added back.. thought it was my debug

src/coreComponents/common/KernelLaunchSelectors.hpp
CusiniM, line 1: if this file sits in common it should be a lot more generic IMO.

src/coreComponents/dataRepository/Group.hpp
```diff
@@ -30,6 +30,7 @@
 
 
 #include <iostream>
+#include <typeinfo>
```
CusiniM: why?

src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseBaseKernels.hpp
CusiniM: I am also confused by this.

src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseBaseKernels.hpp
CusiniM: is this ever used? I can't find it...

CusiniM: It seems that we have too many of these dispatches. We should only have 1 dispatch for compNumber, phaseNumber and thermal and use it everywhere.

CusiniM (commented):
Let's start with this.

src/coreComponents/dataRepository/Group.hpp
tjb-ltk: removed

src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseBaseKernels.hpp
tjb-ltk: so the code can be reused for res and wells.   getFields  is scoped with well and reservoir uses flow ...plus IMHO the prop id shouldn't be bundle with its shape

src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseBaseKernels.hpp
tjb-ltk: will delete... yes we have several dispatches..In the spirit of " too many"  1)  derivative offset structs...  since prop derivatives have thermal slot at index 1.   traits classes  templated on thermal are needed in isothermal runs...  where should we put these structs...     . I think these and dispatches should all be in some common location so the developer knows what is available...   2) further single/multiphase should be combined, and 3) specializations should only be for different variable sets or implicitness  and phase comp partitioning arrays introduced ....

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWellFields.hpp
tjb-ltk: done

src/coreComponents/unitTests/wellsTests/testThermalReservoirCompositionalMultiphaseSSWells.cpp
tjb-ltk: ok.... where do they go...   ideally one could have option to run simulator using numerical derivatives and then its clearly out of unit test...

src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseBaseKernels.hpp
CusiniM: what do you mean by bundled with its shape?

CusiniM: this kind of call  `subRegion.getField< fields::flow::globalCompDensity >();`  should be totally valid also in the case of a well.

src/coreComponents/physicsSolvers/fluidFlow/wells/SinglePhaseWell.cpp
```diff
@@ -407,14 +418,16 @@ void SinglePhaseWell::initializeWells( DomainPartition & domain )
...
+                                         real64 const & dt,
+                                         DomainPartition & domain,
                                          DofManager const & dofManager,
                                          CRSMatrixView< real64, globalIndex const > const & localMatrix,
                                          arrayView1d< real64 > const & localRhs )
 {
   GEOS_MARK_FUNCTION;
-
+  GEOS_UNUSED_VAR( time_n );
+  GEOS_UNUSED_VAR( dt );
```
CusiniM, line 434:
```suggestion
  GEOS_UNUSED_VAR( time_n );
  GEOS_UNUSED_VAR( dt );
```

src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseBaseKernels.hpp
CusiniM: oh I see, we have separate traits for the wells. Unless we are doing it so that the dofManager uses different labels we have no reason to do so.

tjb-ltk: yes the could.. but they haven't been....   the DECLARE_FIELD has the array shape in it along with other fields.. Im familiar with having an independent  property GUID system that can be used  in subsystems for allocation    , defining what is serialized , units for property , etc..  but not all bundled together...

CusiniM: well, that macro has everything need to build a trait with all the info needed to register a wrapper on a group. At the end of the day, we need to to know the types of the objects... we could certalize the definition of the keys to a single location, or one per physicsPackage.

tjb-ltk: "we could certalize the definition of the keys to a single location, or one per physicsPackage".   I support that....

src/coreComponents/common/KernelLaunchSelectors.hpp
tjb-ltk: Good suggestion... added something similar

template< typename S, typename T, typename LAMBDA >
void invokePhaseDispatchLambda ( S val, T numPhases, LAMBDA && lambda )
{
  if( numPhases == 1 )
  {
    lambda( val, std::integral_constant< T, 1 >());
    return;
  }
  else if( numPhases == 2 )
  {
    lambda( val, std::integral_constant< T, 2 >());
    return;
  }
  else if( numPhases == 3 )
  {
    lambda( val, std::integral_constant< T, 3 >());
    return;
  }
  else
  {
    GEOS_ERROR( "Unsupported state: " << numPhases );
  }
};

src/coreComponents/common/KernelLaunchSelectors.hpp
tjb-ltk: done

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWellFields.hpp
tjb-ltk: done

src/coreComponents/physicsSolvers/fluidFlow/wells/SinglePhaseWell.cpp
tjb-ltk: ran uncrustify done

paveltomin:
@CusiniM @rrsettgast can we please merge it? bunch of useful changes for wells here

CusiniM:
> @CusiniM @rrsettgast can we please merge it? bunch of useful changes for wells here

I had left some comments and if they were addressed then I am fine with merging it. However, some things in this PR are done in a way that it's quite different from the rest of the code so I wanted someone else to have a close look at it too.

src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseBaseKernels.hpp
CusiniM: why not using the ones you created?

CusiniM (approved):
I am not sure I would have put those dispatch functionalities in common/internal. Otherwise this PR looks good to me now

src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseBaseKernels.hpp
tjb-ltk: In. KernelLaunchSelector2 there is a comment // Ideally this would be inside the dispatch, but it breaks on Summit with GCC 9.1.0 and CUDA 11.0.3.   I thought it would be better to make that change in different PR.

src/coreComponents/common/KernelLaunchSelectors.hpp
tjb-ltk: I could move to fluidflow...

src/coreComponents/constitutive/fluid/singlefluid/SingleFluidBase.hpp
rrsettgast: @tjb-ltk Is this resolved?

src/coreComponents/mesh/WellElementSubRegion.cpp
```diff
@@ -155,7 +155,8 @@ bool visitNeighborElements( MeshLevel const & mesh,
                             SortedArray< globalIndex > & elements,
                             localIndex & erMatched,
                             localIndex & esrMatched,
-                            localIndex & eiMatched )
+                            localIndex & eiMatched,
+                            globalIndex & giMatched )
```
rrsettgast, line 160: `giMatched` not documented

src/coreComponents/physicsSolvers/fluidFlow/ThermalCompositionalMultiphaseBaseKernels.hpp
```diff
@@ -599,7 +601,7 @@ class ScalingForSystemSolutionKernel : public isothermalCompositionalMultiphaseB
     {
       // compute the change in temperature
       real64 const temp = m_temperature[ei];
-      real64 const absTempChange = LvArray::math::abs( m_localSolution[stack.localRow + m_numComp + 1] );
+      real64 const absTempChange = LvArray::math::abs( m_localSolution[stack.localRow + m_temperatureOffset] );
```
rrsettgast, line 602: this change looks strange to me...is this OK?

src/coreComponents/physicsSolvers/multiphysics/CoupledReservoirAndWellKernels.hpp
```diff
@@ -0,0 +1,592 @@
+/*
```
rrsettgast, line 1: The changes here don't seem well related...or thermal related. I missed it? Can you explain?

src/coreComponents/unitTests/fluidFlowTests/testCompFlowUtils.hpp
rrsettgast:
```suggestion
```

rrsettgast (approved):

src/coreComponents/common/KernelLaunchSelectors.hpp
tjb-ltk: moved to physicsSolvers

src/coreComponents/mesh/WellElementSubRegion.cpp
tjb-ltk: documented in latest commit... single phase thermal well support will be different PR

src/coreComponents/physicsSolvers/fluidFlow/ThermalCompositionalMultiphaseBaseKernels.hpp
tjb-ltk: Agree.  the change was needed so that wells could use the scaling kernel implemented for the reservoir.  The location of the temperature variable is different for well and reservoir, hence the offset variable.  It's ok but fragile.  Another PR could see this being replaced with a traits class, but defining/changing lineups with traits classes would probably need more vetting beyond this kernel.

src/coreComponents/physicsSolvers/multiphysics/CoupledReservoirAndWellKernels.hpp
tjb-ltk: Originally the well-res coupling derivatives were not in kernel classes but assembled in CompositionalMultiphaseReservoirAndWells.cpp.  With the addition of thermal ,kernels were added for iso + thermal and are in CoupledReservoirAndWellKernels.hpp

src/coreComponents/unitTests/fluidFlowTests/testCompFlowUtils.hpp
tjb-ltk: removed in lastest commit

.integrated_tests.yaml
```diff
@@ -1,6 +1,6 @@
 baselines:
   bucket: geosx
-  baseline: integratedTests/baseline_integratedTests-pr2878-8188-ed2ded2
+  baseline: integratedTests/baseline_integratedTests-pr3156-8300-32d1d28
```
```suggestion
  baseline: integratedTests/baseline_integratedTests-pr3156-8366-aa71f2a
```

PR #3160, opened 2024-06-04
Two experimental options for compositional flow solver

paveltomin:
Two experimental options:
- `targetRelativeCompDensChange`: use relative change of component density to select time step size
- `maxRelativeCompDensChange`: chop component density Newton update exciding maximum relative change
Both are inactive by default

Should go after https://github.com/GEOS-DEV/GEOS/pull/3138 (merged) and https://github.com/GEOS-DEV/GEOS/pull/3152 (merged)

Was hoping it might help with SPE11a but no, still worth to keep for testing imho.


src/coreComponents/physicsSolvers/NonlinearSolverParameters.cpp
```diff
@@ -187,44 +188,50 @@ void NonlinearSolverParameters::postProcessInput()
                         getWrapperDataContext( viewKeysStruct::lineSearchResidualFactorString() ) << ": should be positive" );
 
   if( getLogLevel() > 0 )
-  {
```
dkachuma, line 190: I was once told that this is part of the geos coding style. i.e. brackets even for a single statement.

dkachuma: I was once told that this is a geos coding style i.e. include brackets even for a single statement.

src/coreComponents/physicsSolvers/NonlinearSolverParameters.cpp
```diff
@@ -187,44 +188,50 @@ void NonlinearSolverParameters::postProcessInput()
...
   if( m_couplingType == CouplingType::Sequential )
   {
-    GEOS_LOG_RANK_0( GEOS_FMT( "  Sequential convergence criterion = {}", EnumStrings< SequentialConvergenceCriterion >::toString( m_sequentialConvergenceCriterion ) ) );
-    GEOS_LOG_RANK_0( GEOS_FMT( "  Subcycling = {}", m_subcyclingOption ) );
+    tableData.addRow( "Sequential convergence criterion", m_sequentialConvergenceCriterion );
+    tableData.addRow( "Subcycling", m_subcyclingOption );
   }
+  TableLayout const tableLayout = TableLayout( {
+      TableLayout::ColumnParam{"Parameter", TableLayout::Alignment::left},
+      TableLayout::ColumnParam{"Value", TableLayout::Alignment::center},
```
dkachuma: My personal preference would be left alignment for both columns.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -2081,44 +2101,61 @@ real64 CompositionalMultiphaseBase::setNextDtBasedOnStateChange( real64 const &
...
       RAJA::ReduceMax< parallelDeviceReduce, real64 > subRegionMaxTempChange( 0.0 );
       RAJA::ReduceMax< parallelDeviceReduce, real64 > subRegionMaxPhaseVolFracChange( 0.0 );
+      RAJA::ReduceMax< parallelDeviceReduce, real64 > subRegionMaxCompDensChange( 0.0 );
 
       forAll< parallelDevicePolicy<> >( subRegion.size(), [=] GEOS_HOST_DEVICE ( localIndex const ei )
       {
         if( ghostRank[ei] < 0 )
         {
-          // switch from relative to absolute when pressure less than 1
+          // switch from relative to absolute when values less than 1
```
dkachuma, line 2132: The value of 1 seems rather arbitrary and feels like we shouldn't use it for all quantities. It's especially not very meaningful for temperature. I would go with
```cpp
constexpr real64 minPressure = 1.01325e5; // 1 atm
constexpr real64 minTemp = 273.15; // 0 degC
```
Not sure what to use for component density - maybe 1 is OK in this case.
In any case, it's not that necessary especially considering it's a change of behaviour.

dkachuma (approved):

dkachuma:
We already have the composition change `maxCompFractionChange`. Doesn't this indirectly cover the functionality provided by `maxRelativeCompDensChange`?

paveltomin:
> We already have the composition change `maxCompFractionChange`. Doesn't this indirectly cover the functionality provided by `maxRelativeCompDensChange`?

If one density value is much smaller than another, `maxCompFractionChange` is not going to capture the changes i think, that's why i tried to introduce that relative scaling

src/coreComponents/physicsSolvers/NonlinearSolverParameters.cpp
paveltomin: reverted that change

src/coreComponents/physicsSolvers/NonlinearSolverParameters.cpp
paveltomin: changed

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
paveltomin: let's keep all that in mind and address when there is a good reason

PR #3163, opened 2024-06-07
refactor: Trying to improve adaptive linear solver tolerance selection

paveltomin:
1. Pull out hard-coded params to be able to modify them from input:
```
- strongestTol = 1e-8 (default value)
- adaptiveGamma = 0.1
- adaptiveExponent = 1.0
```
2. Adjust `altKrylovTol` logic. Not sure that new one is good but old one is really confusing.
3. Add output to see what tolerance is selected and why (triggered by `logLevel>0` in linear solver params).


paveltomin:
I guess this PR is more a discussion point for now.
I found that adaptive tolerance approach is both quite useful for large scale realistic cases and at the same time the algorithm is quite confusing.
Trying to adjust it to make tolerance choice more understandable.
Please let me know what you think.

src/coreComponents/physicsSolvers/SolverBase.cpp
```diff
@@ -672,22 +672,24 @@ bool SolverBase::lineSearchWithParabolicInterpolation( real64 const & time_n,
...
-  real64 const strongestTol = 1e-8;
-  real64 const exponent = 2.0;
-  real64 const gamma = 0.9;
-
-  real64 normRatio = newNewtonNorm / oldNewtonNorm;
-  if( normRatio > 1 )
-    normRatio = 1;
-
-  real64 newKrylovTol = gamma*std::pow( normRatio, exponent );
-  real64 altKrylovTol = gamma*std::pow( oldNewtonNorm, exponent );
```
paveltomin, line 745: this one is confusing, i replaced `oldNewtonNorm` but old linear solver tolerance, but not quite sure about that
`oldNewtonNorm` squared is usually way too tight

victorapm:
Thanks, Pavel! Just checking... The algorithm is described [here](https://epubs.siam.org/doi/abs/10.1137/0917003?journalCode=sjoce3), correct?

paveltomin:
> Thanks, Pavel! Just checking... The algorithm is described [here](https://epubs.siam.org/doi/abs/10.1137/0917003?journalCode=sjoce3), correct?

I think yes, but because I don't have access I can't say for sure.
I was reading this one https://softlib.rice.edu/pub/CRPC-TRs/reports/CRPC-TR94463.pdf which should be the same I guess.

paveltomin:
any thoughts?

victorapm:
I don't have much to add for now but I'll keep thinking about this. @oseikuffuor1 could you chime in?

paveltomin:
ping

paveltomin:
review anyone?
we don't use this feature that often but sometimes it's quite useful for bigger/longer cases

victorapm (approved):
Sorry! I thought I had approved this before

rrsettgast:
@castelletto1 @CusiniM Can you guys review this?

paveltomin:
rebaseline required only because of the new params, no real results change

rrsettgast (approved):

src/coreComponents/linearAlgebra/utilities/LinearSolverParameters.hpp
```diff
@@ -124,6 +124,9 @@ struct LinearSolverParameters
 #endif
     integer useAdaptiveTol = false;   ///< Use Eisenstat-Walker adaptive tolerance
     real64 weakestTol = 1e-3;         ///< Weakest allowed tolerance when using adaptive method
+    real64 strongestTol = 1e-8;       ///< Strongest allowed tolerance when using adaptive method
+    real64 adaptiveGamma = 0.1;       ///< Gamma parameter for adaptive method
+    real64 adaptiveExponent = 1.0;    ///< Exponent parameter for adaptive method
```
castelletto1, line 129: The default values --  `exponent = 2.0` and `gamma = 0.9` -- have changed. I understand that the old exponent is probably too high. How did you choose the new values?

castelletto1 (approved):
Looks good to me, thank you!

src/coreComponents/linearAlgebra/utilities/LinearSolverParameters.hpp
paveltomin: Yes, you are right.
I just found that exponent=2 is too aggressive but at the same time 0.9 is too weak.
Found 1.0 and 0.1 are more balanced but could be case dependent, so I pulled them out into the input.

PR #3165, opened 2024-06-11
Small bug fix

paveltomin:


tjb-ltk:
Good find.. the constructor initiated value to 1 and the wrong key name some how prevented it from being set to zero.......  However in the fix, the setApplyDefaultValue should be 1, consistent with reservoir default .  I would assume that this value should always be the same as reservoir...

paveltomin:
> Good find.. the constructor initiated value to 1 and the wrong key name some how prevented it from being set to zero....... However in the fix, the setApplyDefaultValue should be 1, consistent with reservoir default . I would assume that this value should always be the same as reservoir...

yes, you are right
just fixed, thanks!

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
```diff
@@ -73,8 +73,8 @@ CompositionalMultiphaseWell::CompositionalMultiphaseWell( const string & name,
     setInputFlag( InputFlags::OPTIONAL ).
     setDescription( "Use mass formulation instead of molar" );
 
-  this->registerWrapper( viewKeyStruct::useMassFlagString(), &m_useTotalMassEquation ).
-    setApplyDefaultValue( 0 ).
+  this->registerWrapper( viewKeyStruct::useTotalMassEquationString(), &m_useTotalMassEquation ).
+    setApplyDefaultValue( 1 ).
```
dkachuma, line 77:
```suggestion
    setApplyDefaultValue( m_useTotalMassEquation ).
```

dkachuma (approved):
Great find.

tjb-ltk (approved):

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
tjb-ltk: I think the intent was to use all this  machinery to set the default and avoid initializing value on class construction..

paveltomin:
Rebaseline required due to appearance of `useTotalMassEquation` in well solver params. No real results change.

PR #3100, opened 2024-04-28
Adding pressure stabilization for single phase poromechanics

ryar9534:
Implemented stabilized flux kernels to enable pressure jump stabilization for single-phase poromechanics. Also moves some stabilization-related functions from MultiphasePoromechanics to PoromechanicsSolver, as now that single-phase stabilization exists it could be unified. 

Also adding input files to perform integrated tests with more physical meaning directly related to pressure stability.

PR #3168, opened 2024-06-12
Code cleanup

paveltomin:
should go after https://github.com/GEOS-DEV/GEOS/pull/3100

- trying to remove some duplications in well solvers
- remove some duplication in reservoir and wells solvers

in both cases - just move what possible to the base class


src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.cpp
tjb-ltk, line 1: why return 0

paveltomin: This is just for consistency with the multiphase version, to be able to use them uniformly

src/coreComponents/physicsSolvers/fluidFlow/wells/SinglePhaseWell.cpp
tjb-ltk, line 1: at some point we could rename mixture connectionrate (compositional version) to connectionrate and move to well solver base...  they are the same size...

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.cpp
tjb-ltk: fair enough.. at some point this needs to be rationalized.....the client could get this value other ways or the return a struct  to include other quants

src/coreComponents/physicsSolvers/multiphysics/CoupledReservoirAndWellsBase.hpp
tjb-ltk, line 1: nice cleanup and reorg... clarifies the which solver is being called  and

tjb-ltk (approved):

PR #3170, opened 2024-06-13
Fix tutorial example for thermal debonding wellbore problem

### LINKED ISSUE #3169: Thermal debonding validation example is not valid
I have re-run the thermal debonding example with the link given below. It is no more valid with the present GEOS version. Some recent modifications have made this example invalid.

Link to the example: 

https://geosx-geosx.readthedocs-hosted.com/en/latest/docs/sphinx/advancedExamples/validationStudies/wellboreProblems/casedContactThermoElasticWellbore/Example.html

Invalid results observed from my recent re-run, we expect to obtain a perfect fit between GEOS and reference results as shown in the link above. However, my recent re-run give an invalid comparion as shown in the figure below:

image

When I plot the casing-cement and cement-rock interfaces, it plots something strange as shown in the figure below. The good visualization can be found in the link above.

image

jhuang2601:
This small PR fixes the reported issue (#3169).

It should be merged after #3133.


jhuang2601:
```
=======================
Test failures
=======================
CasedThermoElasticWellbore_ImperfectInterfaces_smoke_01
PoroElastic_hybridHexPrism_co2_fim_3d_01
PoroElastic_hybridHexPrism_co2_sequential_3d_01
validationCase_01
=======================
Overall status: FAILED
=======================
IntegratedTests failed. Rebaseline is required.
```

sytuannguyen (changes requested):
Very good fix, there still a little difference in the last figure (red curves) however this fix sound very good. I am just not ok with the term "no debonding" in the legends of the first figure. Indeed, these results are for the case with debonding as we can see the displacement jump at r approx 0.21 m

sytuannguyen (approved):

PR #3173, opened 2024-06-17
Extend BLAS interface

dkachuma:
Extends the BLAS/LAPACK interface to include linear solves where the solution and rhs are matrices. The addition is a function to solve
$$AX=B$$
where $A$ is an $n{\times}n$ matrix and $X$ and $B$ are $n{\times}m$ matrices.

This also adds an "in-place" version of `solveLinearSystem`. This takes the matrix $A$ and the rhs $B$ only. In using this, the caller is allowing LAPACK to destroy both the $A$ and $B$. On exit $A$ is replaced by an LU factorisation of $A$ and $B$ is replaced by the solution (in true LAPACK style). This has the advantage of not requiring extra allocations of memory within the call.

For the row-major version, I couldn't figure out how to call BLAS/LAPACK. So I was forced to allocate space anyway for the transposition of the solution matrix $X$. Two exceptions to this are 1) $m=1$ the vector case in which case row-major vs col-major is meaningless and 2) $m=n$ in which case $X$ can be transposed in place.


src/coreComponents/denseLinearAlgebra/interfaces/blaslapack/BlasLapackLA.cpp
```diff
@@ -379,6 +379,145 @@ void matrixLeastSquaresSolutionSolve( arraySlice2d< real64, USD > const & A,
...
+    int const INCY = 1;
+    int const K = LvArray::integerConversion< int >( B.size( ) );
+    GEOS_dcopy( &K, B.dataIfContiguous(), &INCX, X.dataIfContiguous(), &INCY );
+  }
+
+  // For row-major form, we need to reorder into col-major form
+  // This might require an extra allocation
+  real64 * solutionData = X.dataIfContiguous();
+  array2d< real64, MatrixLayout::COL_MAJOR_PERM > X0;
+  if constexpr ( USD == MatrixLayout::ROW_MAJOR )
```
victorapm, line 475: Does it make sense to create a helper function for this operations (and also save some lines later when we transpose the data back)?

victorapm (approved):
Looks good! Just some minor comments... Also: should we put a warning somewhere that this is not supported on GPUs (apparently)?

> For the row-major version, I couldn't figure out how to call BLAS/LAPACK. So I was forced to allocate space anyway for the transposition of the solution matrix

It seems that lapacke follows this same approach and there's nothing much we can do, see:

https://github.com/Reference-LAPACK/lapack/blob/8b468db25c0c5a25d8e0020c7e2134e14cfd33d0/LAPACKE/src/lapacke_dgetrf_work.c
https://github.com/Reference-LAPACK/lapack/blob/8b468db25c0c5a25d8e0020c7e2134e14cfd33d0/LAPACKE/utils/lapacke_dge_trans.c

dkachuma:
> It seems that lapacke follows this same approach and there's nothing much we can do, see:
> 
I actually looked at lapacke to get some inspiration and was gravely disappointed to see that they just copy. I also considered in-place transposes but they are so complicated and quite unreadable.

victorapm:
Yes, avoiding the copy would be great... Hmm, if this is to be called inside a FEM kernel or, generally speaking, called multiple times, I guess we could allocate the extra work space only once outside the `solveLinearSystem` function. At least, we get away with constant alloc/dealloc this way. Just sharing the idea...

src/coreComponents/denseLinearAlgebra/interfaces/blaslapack/BlasLapackLA.cpp
dkachuma: Great idea! Added some helpers.

dkachuma:
> Yes, avoiding the copy would be great... Hmm, if this is to be called inside a FEM kernel or, generally speaking, called multiple times, I guess we could allocate the extra work space only once outside the `solveLinearSystem` function. At least, we get away with constant alloc/dealloc this way. Just sharing the idea...

Wasn't sure why kind of threaded code this is called from. Didn't want to introduce something that might complicate that. But I'm sure it can be done. Anyhow the current uses of this are only 1d so we shouldn't be hitting this issue yet.

CusiniM (approved):
Unfortunately, I am afraid that doing that copy is the only solution.

Issue #3073, opened 2024-04-10
XML Inclusions


untereiner:
**Describe the issue**
For now inclusions are handled by an ad hoc xml element. There is need of a special treatment to use included files. 

**Proposed cleanup**
XML Inclusions (XInclude) is a W3C specification that defines a way to include XML fragments from different sources into a single document. We could use this official mechanism instead of an ad hoc one. What do you think ?

**Additional context**
n.a.

cssherman:
This is do-able, but I have a few concerns:

1) Our xml interpreter (pugixml) does not support xi:include natively, so we would have to do this all ourselves
2) The syntax of xi:include isn't as user friendly as our current implementation
3) Validating the included xml files is a bit harder using xinclude, as they aren't required to conform to the parent schema

untereiner:
> This is do-able, but I have a few concerns:
> 
> 1. Our xml interpreter (pugixml) does not support xi:include natively, so we would have to do this all ourselves

I just saw that: https://github.com/zeux/pugixml/issues/338 
It seems weird but yes, we would have to to it

> 2. The syntax of xi:include isn't as user friendly as our current implementation

aside from the `xi:` not sure it is very different
```
<Included>
  <File name="./FieldCaseTutorial3_base.xml" />
</Included>
 ``` 
vs
``` 
<xi:include href="./FieldCaseTutorial3_base.xml"/> 
```

> 3. Validating the included xml files is a bit harder using xinclude, as they aren't required to conform to the parent schema

I don't understand this concern. Do you have an example ?

untereiner:
For me a schema makes a whole. There is not parent or child schemas.
Using xinclude in xml files is just a way to break down the complexity of a large document.

PR #3180, opened 2024-06-20
Draft: Input Parsing Refactor / Generic Input Extension Mechanisms

wrtobin:


untereiner:
Hello @wrtobin , 
I was looking at you draft and I don't understand the objective of the modifications for xml inclusions.
Why are the included xml files more inlaid in the `Group` concept ?
What is the concept of  `Generic Input Extension`

I think this work is also related to this discussion https://github.com/GEOS-DEV/GEOS/issues/3073

I would be in favor of keeping things separated:
- on one side xml files (even splitted files), the entire xml tree can be reconstructed and validated
- on the other side the elements of the tree that can be instantiated as objects by geos.

src/coreComponents/dataRepository/InputProcessing.hpp
```diff
@@ -0,0 +1,420 @@
...
+protected:
+  virtual void processNode( data_node *, document_node & ) = 0;
+  virtual data_node & retrieveOrCreateDataNode( data_node * parent, string const & dataNodeName, string const & docNodeName ) = 0;
+
+  void validateUniqueChildren( data_node & dataParent, document_node & docNode, string const & dataNodeName, std::map< string, std::set< string > > & childNamesMap )
+  {
+    auto & knownChildNames = childNamesMap[ dataParent.getName() ];
+    document_node_pos docNodePos = m_document.getNodePosition( docNode );
+    GEOS_ERROR_IF( std::find( knownChildNames.begin(), knownChildNames.end(), dataNodeName ) != knownChildNames.end(),
+                  GEOS_FMT( "Error: An Input block cannot contain children with duplicated names.\n"
```
MelReyCG, line 163: `An input node` ?

src/coreComponents/dataRepository/InputProcessing.hpp
```diff
@@ -0,0 +1,420 @@
...
+private:
+  document_type const & m_document;
+};
+
+using PreprocessOnly = InputProcessor< xmlWrapper::xmlDocument,
+                                       Group,
+                                       TerseSyntax< xmlWrapper::xmlDocument, Group >,
+                                       Declaration< xmlWrapper::xmlDocument, Group >,
+                                       TerseSyntax< xmlWrapper::xmlDocument, Group > >;
+
```
MelReyCG, line 408: I like the fact that we could be agnostic about the type & process of the input file in `Group`

src/coreComponents/dataRepository/InputProcessing.hpp
```diff
@@ -0,0 +1,420 @@
...
+  void restorePathPrefix( const string & oldPrefix )
+  {
+    Path::setPathPrefix( oldPrefix );
+  }
+
+  document_type & m_document;
+  inputExtension::InputExtender< document_node > m_inputExtender;
+};
+
+template < typename DocumentType, typename DataNode, typename... Phases >
```
MelReyCG:
```suggestion
template < typename DocumentType, typename DataNode, typename... InputPhases >
```

src/coreComponents/dataRepository/InputProcessing.hpp
MelReyCG: if the xml would be versioned, we could add a `LegacyCompatibility` phase that would transform the data model to the last version.

src/coreComponents/dataRepository/xmlWrapper.cpp
```diff
@@ -143,69 +145,8 @@ void addNodeFileInfo( xmlNode targetNode, string const & filePath )
  * @brief Returns true if the addNodeFileInfo() command has been called of the specified node.
  */
 bool xmlDocument::hasNodeFileInfo() const
-{ return !getFirstChild().attribute( filePathString ).empty(); }
+{ return !getFirstChild().attribute( inputParsing::filePathString ).empty(); }
 
-void xmlDocument::addIncludedXML( xmlNode & targetNode, int const level )
```
MelReyCG, line 148: As you guys already said about the moving / deletion of `xmlDocument::addIncludedXML()`, I also think it is up to the file format to define a way to include files.

wrtobin:
Part of this work is attempting to separate input file format dependence / parsing from internal group/attribute instantiation and definition. 

There is still work to do (clearly) but this includes abstracting inclusion to the point that files of arbitrary (but supported) hierarchical input formats can include other files of arbitrary (but supported) hierarchical input formats.

Whether we want to fully allow that is a different concern from designing a system that is componentized in such a way to facilitate trivially supporting that. 

To accomplish that inclusion needs to be independent of document type, where it sits right now ( the current implementation ) is mostly a result of iteratively removing it from the xmlwrapper, rather than a hard decision of how we should actually accomplish this.

untereiner:
The separation of file format / groups is a very good idea
But, file inclusion is a file thing and is more or less supported depending on the chosen format. It should then be handled at the file level IMHO. The input parser should take as input a complete document to parse. 

If follow you remark I'd say it should be possible to include yaml files in json files and theses files in xml files. But that's a bit weird.

And yes the work sits in the xmlwrapper but I am concerned about the handling of these includes.

wrtobin:
The inclusion modification is currently the part of this work I like the least, yeah. 

While I don't necessarily want to support inclusion of files of different formats, my intent is to implement things in a generic enough way that such a thing is possible (though not supported), since really we're just working with a bunch of abstract trees in all cases.

I'm amenable to reworking it and seeing if putting inclusion processing into document abstractions makes things cleaner.

MelReyCG (commented):
A few comments

PR #3181, opened 2024-06-21
Decouple debug matrix output from logLevel

paveltomin:
Small thing - introduce `writeMatrix` flag to separate debug matrix output from `logLevel`
Logic is similar to old one:
`writeMatrix = 1` - print to screen
`writeMatrix >= 2` - print to file


src/coreComponents/physicsSolvers/SolverBase.hpp
```diff
@@ -804,6 +804,9 @@ class SolverBase : public ExecutableGroup
   /// Custom preconditioner for the "native" iterative solver
   std::unique_ptr< PreconditionerBase< LAInterface > > m_precond;
 
+  /// flag for debug output of matrix, rhs, and solution
+  integer m_writeMatrix;
```
victorapm: How about `m_writeLinearSystem` since we do matrix, rhs, and sol?

victorapm (approved):
Looks good! Just adding couple of ideas...

Maybe in the future we could do `writeMatrix > 2` means print systems produced after `writeMatrix - 2` linear system calls.

paveltomin:
> Looks good! Just adding couple of ideas...

thanks, just implemented both

PR #3188, opened 2024-06-26
Link against gRPC

MichaelSekachev:


MichaelSekachev:
Hi Randy. I filed this PR to a branch called msek/ix-geos-coupling instead of the developer branch because I need to work with a one-year old GEOS code, i.e. the specific commit Pierre Samier used for his development. I have a same-name branch in the TPL as well:
https://github.com/GEOS-DEV/thirdPartyLibs/commits/msek/ix-geos-coupling/

MichaelSekachev:
The issue I have is when I run
`$ python scripts/config-build.py -hc host-configs/TOTAL/rdhpc8-CPU.cmake -bt Debug`

I get the following error:
```
CMake Warning at cmake/thirdparty/SetupGeosxThirdParty.cmake:880 (message):
  ENABLE_GRPC is ON but GRPC_DIR isn't defined.
  cmake/CMakeBasics.cmake:49 (include)
  CMakeLists.txt:71 (include)

-- Not using grpc.
```

MichaelSekachev:
Here are also the logs for TPL and GEOS:
[output-tpl-python-rdhpc8-debug.txt](https://github.com/user-attachments/files/15994911/output-tpl-python-rdhpc8-debug.txt)
[output-geos-python-rdhpc8-debug.txt](https://github.com/user-attachments/files/15995191/output-geos-python-rdhpc8-debug.txt)

rrsettgast:
@MichaelSekachev 
you need to define `GRPC_DIR` somewhere. There is a default entry you should add in `host-configs/tpls.cmake`.

host-configs/tpls.cmake
```diff
@@ -52,6 +52,10 @@ if(EXISTS ${GEOSX_TPL_DIR}/fesapi)
   set(FESAPI_DIR ${GEOSX_TPL_DIR}/fesapi CACHE PATH "" FORCE)
 endif()
 
+if(EXISTS ${GEOSX_TPL_DIR}/grpc)
+  set(GRPC_DIR ${GEOSX_TPL_DIR}/grpc CACHE PATH "" FORCE)
+endif()
```
rrsettgast, line 57: @MichaelSekachev
you need to define GRPC_DIR somewhere. There is a default entry that should be added in host-configs/tpls.cmake like I have done here...I actually don't know the correct path, so you should correct it if this is incorrect.

MichaelSekachev: @rrsettgast Thank you. Apparently, I did not see where I was supposed to do it. I was distracted by a variable named the same way in the main CMakeLists.txt file of the TPL itself. Thank you for going the extra mile and pushing the commit. I was able to configure the code just fine now. Next step is to play with cpp files and try to add an include file in one of the cpp files. I just randomly picked `src/coreComponents/functions/CompositeFunction.hpp` and I think I need to modify the `src/coreComponents/functions/CMakeLists.txt` similar to MATHPRESSO.

PR #3193, opened 2024-06-28
feat: Geothermal Gradient for Single Phase flow

npillardou:
The purpose of this PR is to add the geothermal gradient capacities for single-phase flow.

At the moment the temperature VS Elevation is correlated with the composition VS Elevation, maybe this point could be extended to something that is independent of the composition: 

    <HydrostaticEquilibrium
      name="Initialisation"
      datumElevation="-3790"
      datumPressure="4.3e7"
      objectPath="ElementRegions"
      initialPhaseName="water"
      componentNames="{ co2, water }"
      componentFractionVsElevationTableNames="{ CO2VD, H2OVD }"
      temperatureVsElevationTableName="initTempTable" />  

Further work is needed to account for composition in the hydrostatic pressure kernel, as it is done in CompositionalMultiphaseFlow.


src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.cpp
```diff
@@ -505,6 +505,21 @@ void SinglePhaseBase::computeHydrostaticEquilibrium( DomainPartition & domain )
...
+    TableFunction::KernelWrapper tempTableWrapper;
+    bool tempTableGradient = false;
+
+    if ( !fs.getTemperatureVsElevationTableName().empty() )
+    {
+
+      tempTableGradient = true;
+      string const tempTableName = fs.getTemperatureVsElevationTableName();
+      TableFunction const & tempTable = functionManager.getGroup< TableFunction >( tempTableName );
+      tempTableWrapper = tempTable.createKernelWrapper();
```
paveltomin, line 523: this should be used in `HydrostaticPressureKernel`, no? i mean let's say density formally depends on temperature, then temperature should be used there

npillardou: Yes I agree with you, we should modify the kernel into something like: 

HydrostaticPressureKernel::launch( numPointsInTable,
                                           maxNumEquilIterations,
                                           equilTolerance,
                                           gravVector,
                                           minElevation,
                                           elevationIncrement,
                                           datumElevation,
                                           datumPressure,
                                           fluidWrapper,
                                           tempTableWrapper,
                                           elevationValues.toNestedView(),
                                           pressureValues.toView() ); 
                                           
But behind that, it requires further developments in the constitutive also

paveltomin: > But behind that, it requires further developments in the constitutive also

can you elaborate?

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.cpp
```diff
@@ -562,6 +577,7 @@ void SinglePhaseBase::computeHydrostaticEquilibrium( DomainPartition & domain )
     // TODO: this last step should probably be delayed to wait for the creation of FaceElements
     arrayView2d< real64 const > const elemCenter = subRegion.getElementCenter();
     arrayView1d< real64 > const pres = subRegion.getField< fields::flow::pressure >();
+    arrayView1d< real64 > const temp = subRegion.getField< fields::flow::temperature >();
 
     RAJA::ReduceMin< parallelDeviceReduce, real64 > minPressure( LvArray::NumericLimits< real64 >::max );
```
tjb-ltk, line 602: You could add.   kernelLaunchSelectorThermalSwitch and then use constexpr to have compile time inclusion of thermal code instead of if tempTableGradient

paveltomin:
@npillardou are you planning to finish this PR or it can be closed?

herve-gross:
@dkachuma and @jhuang2601 please take a look

npillardou:
> @npillardou are you planning to finish this PR or it can be closed?

Sorry @paveltomin I totally forgot to answer you. I should take some time to investigate a good implementation, maybe with the help of Jian and Dick.

jafranc (commented):
Would it be possible to make it consistent with this:
https://github.com/GEOS-DEV/GEOS/blob/45f272c3d8313d6443df260436bc1a2cf5244b6b/src/coreComponents/physicsSolvers/fluidFlow/kernels/compositional/HydrostaticPressureKernel.hpp#L49-L56

from #3795

jafranc (commented):
@npillardou can you add an inputFile using it ?

jafranc:
Ok looking at it I am a bit puzzled, most likely because I am a little rusted.

IIUC we are trying to create a shorthand to have a single phase non thermally dependent case to initialize a _passive_ temperature field base on a table under the `<HydrostaticEquilibrium>` tag, am I correct @npillardou  ?

jafranc:
If *true*, I would group everything temperature introduced in `SinglePhaseBase::computeHydrostaticEquilibrium()` at the end of the function, to the expense of repeating the loops (IIRC the cost won't be much).

If *false* and there is plan and intention to have the single phase initialization to take into account both pressure and temperature, then as @tjb-ltk we need some `constexpr` dispatch that could elude code generation in the part it is not needed and error thrown if there is thermal simulation launched without thermal init (isn't it ?)

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.cpp
jafranc: I think the dispatch will only be valid if we are actually dealing with thermal dependent solver (as the `m_isThermal` in the class below) and not if the intent is to impose a passive temperature field

jhuang2601:
> If _true_, I would group everything temperature introduced in `SinglePhaseBase::computeHydrostaticEquilibrium()` at the end of the function, to the expense of repeating the loops (IIRC the cost won't be much).
> 
> If _false_ and there is plan and intention to have the single phase initialization to take into account both pressure and temperature, then as @tjb-ltk we need some `constexpr` dispatch that could elude code generation in the part it is not needed and error thrown if there is thermal simulation launched without thermal init (isn't it ?)

@jafranc It should be the latter one. As of now, I just address Pavel's comment to use temperature field in `HydrostaticPressureKernel`. This is still a WIP process to add `constexpr` dispatch.

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.cpp
jhuang2601: > this should be used in `HydrostaticPressureKernel`, no? i mean let's say density formally depends on temperature, then temperature should be used there

Yes, add temperature field in `HydrostaticPressureKernel`, considering temperature dependent fluid density

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.cpp
jhuang2601: Just switch to `m_isThermal`

src/coreComponents/physicsSolvers/fluidFlow/kernels/singlePhase/ThermalHydrostaticPressureKernel.hpp
```diff
@@ -0,0 +1,223 @@
...
+struct HydrostaticPressureKernel
+{
+
+  template< typename FLUID_WRAPPER >
+  static bool
+  computeHydrostaticPressure( integer const maxNumEquilIterations,
+                              real64 const & equilTolerance,
+                              real64 const (&gravVector)[ 3 ],
+                              FLUID_WRAPPER fluidWrapper,
+                              TableFunction::KernelWrapper tempTableWrapper,
```
jafranc, line 44: Shouldn't temperature value be passed as a `real64 const`/`real64` to be consistent with unitarian nature of the intended interface ?
```suggestion
                              real64 const & temp,
```

src/coreComponents/physicsSolvers/fluidFlow/kernels/singlePhase/ThermalHydrostaticPressureKernel.hpp
```diff
@@ -0,0 +1,223 @@
...
+                              real64 const & refPres,
+                              real64 const & refDens,
+                              real64 const & newElevation,
+                              real64 & newPres,
+                              real64 & newDens )
+  {
+    // Step 1: guess the pressure with the refDensity
+
+    real64 const gravCoef = gravVector[2] * ( refElevation - newElevation );
+    real64 const temp = tempTableWrapper.compute( &newElevation );
```
jafranc, line 55:
```suggestion
```

src/coreComponents/physicsSolvers/fluidFlow/kernels/singlePhase/ThermalHydrostaticPressureKernel.hpp
```diff
@@ -0,0 +1,223 @@
...
+    // Step 3: compute the mass density and pressure at the reference elevation
+
+    array1d< real64 > dens( pressureValues.size() );
+
+    bool const hasConvergedRef =
+      computeHydrostaticPressure( maxNumEquilIterations,
+                                  equilTolerance,
+                                  gravVector,
+                                  fluidWrapper,
+                                  tempTableWrapper,
```
jafranc, line 154:
```suggestion
                                 temp[iRef], 
```

src/coreComponents/physicsSolvers/fluidFlow/kernels/singlePhase/ThermalHydrostaticPressureKernel.hpp
```diff
@@ -0,0 +1,223 @@
...
+
+    localIndex const numEntriesAboveRef = size - iRef - 1;
+    forAll< serialPolicy >( numEntriesAboveRef, [=, &hasConverged] ( localIndex const i )
+    {
+      bool const hasConvergedAboveRef =
+        computeHydrostaticPressure( maxNumEquilIterations,
+                                    equilTolerance,
+                                    gravVector,
+                                    fluidWrapper,
+                                    tempTableWrapper,
```
jafranc, line 177:
```suggestion
                                    temp[iRef+i],
```

src/coreComponents/physicsSolvers/fluidFlow/kernels/singlePhase/ThermalHydrostaticPressureKernel.hpp
```diff
@@ -0,0 +1,223 @@
...
+
+    localIndex const numEntriesBelowRef = iRef;
+    forAll< serialPolicy >( numEntriesBelowRef, [=, &hasConverged] ( localIndex const i )
+    {
+      bool const hasConvergedBelowRef =
+        computeHydrostaticPressure( maxNumEquilIterations,
+                                    equilTolerance,
+                                    gravVector,
+                                    fluidWrapper,
+                                    tempTableWrapper,
```
jafranc, line 202:
```suggestion
                                    temp[iRef-i],
```

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.cpp
```diff
@@ -518,28 +536,54 @@ void SinglePhaseBase::computeHydrostaticEquilibrium( DomainPartition & domain )
...
+            HydrostaticPressureKernel::launch( numPointsInTable,
+                                               maxNumEquilIterations,
+                                               equilTolerance,
+                                               gravVector,
+                                               minElevation,
+                                               elevationIncrement,
+                                               datumElevation,
+                                               datumPressure,
+                                               fluidWrapper,
+                                               tempTableWrapper,
```
jafranc, line 552:
```suggestion
                                               tempValues.toView(),
```

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.cpp
```diff
@@ -492,6 +500,16 @@ void SinglePhaseBase::computeHydrostaticEquilibrium( DomainPartition & domain )
     // Step 3.2: retrieve the fluid model to compute densities
     // we end up with the same issue as in applyDirichletBC: there is not a clean way to retrieve the fluid info
```
jafranc, line 515: Why not group that closer to the call-site ?

jafranc (commented):
Except from a small comment on the interface, to pass simple `const/non-const  real64` instead of full table wrapper and group all new addition at the bottom instead of across the file, the implementation looks good to me. Thanks @jhuang2601 !

Maybe we should also touch-base the problem of having to provid obvious compositionVSDepth in single phase case instead of automatically fill them with 1-or-0 ?

paveltomin (approved):

npillardou:
To keep track : 

Based on the previous development, I have just added a 3D test with a geothermal gradient option. 
I included this case in the integrated tests. 

However, I noticed that the 'same' case in isothermal conditions, which was already present in the folder, was no longer converging. Therefore, I modified the lineSearch option to 'None' to resolve this issue. I also added this case to the integrated tests — it might be interesting to have it here too.

jhuang2601 (approved):

jafranc (approved):

jhuang2601:
@MelReyCG @arng40 For the PR, this CI check kept failing on `testFlowStatistics` :https://github.com/GEOS-DEV/GEOS/actions/runs/18139901718/job/51627947541?pr=3193
Can you take a look and advise how to fix it?

MelReyCG:
This does not seem really like a issue really related with the test but a memory corruption.
More precisely, it seems like a corrupted memory detection while freeing the memory of a CellElementSubregion .

First, I would suggest testing that with the problematic system with bounds-checking.
Then testing to remove parts of the PR, to try to see the problem source.

jhuang2601:
Due to change in `CompressibleSinglePhaseUpdate< ExponentApproximationType::Full`, all smoke tests with `CompressibleSinglePhaseFluid` failed as expected and hence rebaseline.

herve-gross (approved):

src/coreComponents/constitutive/fluid/singlefluid/CompressibleSinglePhaseFluid.hpp
```diff
@@ -155,6 +155,7 @@ class CompressibleSinglePhaseFluid : public SingleFluidBase
 
   /// Type of kernel wrapper for in-kernel update (TODO: support multiple EAT, not just linear)
   using KernelWrapper = CompressibleSinglePhaseUpdate< ExponentApproximationType::Linear, ExponentApproximationType::Linear >;
+//  using KernelWrapper = CompressibleSinglePhaseUpdate< ExponentApproximationType::Full, ExponentApproximationType::Linear >;
```
rrsettgast:
```suggestion
```

rrsettgast (approved):

src/coreComponents/constitutive/fluid/singlefluid/CompressibleSinglePhaseFluid.cpp
```diff
@@ -110,9 +110,9 @@ void CompressibleSinglePhaseFluid::postInputInitialization()
   // Due to the way update wrapper is currently implemented, we can only support one model type
   auto const checkModelType = [&]( ExponentApproximationType const value, auto const & attribute )
   {
-    GEOS_THROW_IF_NE_MSG( value, ExponentApproximationType::Linear,
-                          GEOS_FMT( "{}: invalid model type in attribute '{}' (only linear currently supported)", getFullName(), attribute ),
-                          InputError );
+    GEOS_THROW_IF( value != ExponentApproximationType::Linear && value != ExponentApproximationType::Full,
```
paveltomin: linear is not supported anymore?

jhuang2601: Both linear and full are valid. If the value is not one of these, it throws an error.

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.cpp
```diff
@@ -552,17 +599,25 @@ void SinglePhaseBase::computeHydrostaticEquilibrium( DomainPartition & domain )
...
     arrayView2d< real64 const > const elemCenter = subRegion.getElementCenter();
     arrayView1d< real64 > const pres = subRegion.getField< flow::pressure >();
+    arrayView1d< real64 > const temp = subRegion.getField< flow::temperature >();
 
-    RAJA::ReduceMin< parallelDeviceReduce, real64 > minPressure( LvArray::NumericLimits< real64 >::max );
 
-    forAll< parallelDevicePolicy< > >( targetSet.size(), [=] GEOS_HOST_DEVICE ( localIndex const i )
+    RAJA::ReduceMin< parallelHostReduce, real64 > minPressure( LvArray::NumericLimits< real64 >::max );
+
+    forAll< parallelHostPolicy >( targetSet.size(), [=] GEOS_HOST_DEVICE ( localIndex const i )
```
paveltomin, line 603: @rrsettgast @klevzoff 
changing this from `parallelDevicePolicy` to `parallelHostPolicy` seems to resolve that unit test failure

Issue #3104, opened 2024-04-30
Wave solvers: Source on a mesh node/edge results in wrong signal (same for receivers)


Bertbk:
**Describe the bug**
When running a wave propagation solver (eg acoustic), if the source lies on the boundary of 2 or more elements, the amplitude of the source signal can be counted multiple times, or not a all, resulting in a wrong result.
This has been reported on issue #2256 but closed for duplicate reason with #2227. However, the problem persists. Could it be due to parallel concurrency?

**To Reproduce**
Launch numerous time an xml file with the following geometry and solver. The source is at the origin and the mesh is such that the origin is a mesh node. The complete file is provided at the end.

```xml
  <Mesh>
    <InternalMesh
      name="mesh"
      elementTypes="{ C3D8 }"
      xCoords="{ -1500,1500}"
      yCoords="{ -1500,1500 }"
      zCoords="{ -1500,1500 }"
      nx="{ 150 }"
      ny="{ 150 }"
      nz="{ 150 }"
      cellBlockNames="{ cb }"/>
  </Mesh>
  <Solvers>
    <AcousticSEM
	name="acousticSolver"
      cflFactor="0.25"
      discretization="FE1"
      targetRegions="{ Region }"
      sourceCoordinates="{ { 0, 0, 0 } }"
      outputSeismoTrace = "1"
      dtSeismoTrace="0.0013"
      timeSourceFrequency="5"
      receiverCoordinates="{ { 1, 2, 3 }}"/>
  </Solvers>
  <NumericalMethods>
    <FiniteElements>
      <FiniteElementSpace
        name="FE1"
        order="1"
        formulation="SEM"/>
    </FiniteElements>
  </NumericalMethods>
```

**Screenshots**

Here is a plot of different signal obtained through the same piece of code. The amplitude varies from 0 to the double. 99 simulations have been launched. The source is at the origin belonging to 8 elements, the receiver is close to at (1,2,3), inside an element.

![acousticSolver_Sources](https://github.com/GEOS-DEV/GEOS/assets/5602767/91c1031c-408b-4f86-a0d1-ab585f07d7e1)

**Platform (please complete the following information):**
- Machine: pangea3
- Compiler: tested with gcc
- GEOSX Version: develop

**Additional information**

[...]

Issue #2888, opened 2023-12-12
Locate seismic point sources and point receivers in the mesh and calculate constant source and receiver terms


shenchengyi:
**Point sources/receivers shared by several elements**
The current algorithm to find the sources and receivers will not stop looking for them once a point is located. For example, if a source or a receiver is on an internal edge shared by 4 elements, the point will be found on 4 neighbour elements and the constant terms will be calculated 4 times. With GPU this leads to incorrect source/receiver constant values. Note that these incorrect values are not reproducible: they vary from one run to another.

**To Reproduce**
1. Add some printf in AcousticWaveEquationSEMKernel.hpp or ElasticWaveEquationSEMKernel.hpp like: printf("Elem %d source %d sourceIsAccessible %d\n", k, isrc, sourceIsAccessible[isrc]); around L98 and printf("Node=%d Elem=%d sourceCoords=%.3f %.3f %.3f sourceConstants=%.3e \n", a,k,coords[0],coords[1],coords[2],sourceConstants[isrc][a]); inside for( localIndex a = 0; a < numNodesPerElem; ++a ){}
2. Recompile GEOS (with or without pyGEOS)
3. use the xml input below (the mesh file path is available in the xml)
4. run geos or pygeos
5. in the log we will have the following lines (the sourceConstants may not be reproducible) 
Node=0 Elem=8255 sourceCoords=6325.000 3162.500 408.000 sourceConstants=0.000e+00                                                            
Node=0 Elem=8128 sourceCoords=6325.000 3162.500 408.000 sourceConstants=0.000e+00                                                            
Node=0 Elem=8127 sourceCoords=6325.000 3162.500 408.000 sourceConstants=0.000e+00                                                            
Node=0 Elem=8256 sourceCoords=6325.000 3162.500 408.000 sourceConstants=2.000e-01                                                            
Node=1 Elem=8127 sourceCoords=6325.000 3162.500 408.000 sourceConstants=0.000e+00                                                            
Node=1 Elem=8128 sourceCoords=6325.000 3162.500 408.000 sourceConstants=0.000e+00                                                            
Node=1 Elem=8256 sourceCoords=6325.000 3162.500 408.000 sourceConstants=0.000e+00                                                            
Node=2 Elem=8127 sourceCoords=6325.000 3162.500 408.000 sourceConstants=0.000e+00                                                            
[...]

PR #3202, opened 2024-07-02
fix: robust point location for degenerate configurations

sframba:
This PR adds a new method to determine if a point lies inside or outsilde a (convex) polyhedral cell.
Compared to the current method, the new implementation is slower, but handles gracefully the degenerate conditions of points lying on mesh vertices, edges and faces. It does so by:
- Triangulating each face of a cell in a consistent manner, by always choosing the lowest global-id vertex as root;
- Ordering each triangle again using global ids, so that the edges and vertices are encountered in a consistent order across MPI ranks;
- Computing the winding number (an integer) subtended by each triangle;
- If the point lies on a vertex, edge or face, this will be found consistently by all ranks, whatever the cell from which the face/edge/vertex is seen. This is true because the order in which the vertices are traversed (and thus the result of the calculation) is exactly the same no matter the originating cell or MPI rank;
- Finally, if a degenerate condition is found, the {face,edge,vertex}-to-element maps of the host element are used to choose consistently a single representative element. The adjacent element with the lowest global ID is chosen, ensuring general consistency across MPI ranks.

The only ambiguity left is when the point lies on a face/edge/vertex that is a boundary for an MPI rank, but an interior face/edge/vertex for another rank. In this case, both ranks will claim the point. However, at least one of them will consider the point to belong to a ghost element. This is a desirable outcome, since the desired behaviour in this case can be different according to the specific application (e.g., sources vs. receivers for the seismic wave case). This outcome leaves the choice to the caller method on how to handle this case, using ghost rank information.

The new point location algorithm is used for wave propagation kernels, fixing [issue 3104](https://github.com/GEOS-DEV/GEOS/issues/3104) and [issue 2888](https://github.com/GEOS-DEV/GEOS/issues/2888).


acitrain (approved):

src/coreComponents/mesh/utilities/ComputationalGeometry.hpp
```diff
@@ -435,6 +434,375 @@ bool isPointInsidePolyhedron( arrayView2d< real64 const, nodes::REFERENCE_POSITI
...
+ * @param[in] ay y-coordinate of the first vertex
+ * @param[in] az z-coordinate of the first vertex
+ * @param[in] bx x-coordinate of the second vertex
+ * @param[in] by y-coordinate of the second vertex
+ * @param[in] bz z-coordinate of the second vertex
+ * @return 0 if the vertices coincide, +1 if a > b and -1 if b > a
+ */
+template< typename COORD_TYPE, typename POINT_TYPE >
+GEOS_HOST_DEVICE
+int lexicographicalCompareVertex( POINT_TYPE const ax, POINT_TYPE const ay, POINT_TYPE const az,
```
untereiner, line 451: (ax,ay,az) and (bx,by,bz) can be of different concrete types ? (with the assumption that these two types overload comparison operator)

src/coreComponents/mesh/utilities/ComputationalGeometry.hpp
```diff
@@ -435,6 +434,375 @@ bool isPointInsidePolyhedron( arrayView2d< real64 const, nodes::REFERENCE_POSITI
...
+                    ( v3x * v1y - v3y * v1x ) * v2z;
+  if( sign > 0 )
+    return 1;
+  else if( sign < 0 )
+    return -1;
+  return 0;
+}
+/**
+ * @brief Method to find the reference element touching a vertex. The element with the lowest global ID is chosen
+ * list.
```
untereiner: "in chosen list" ?

src/coreComponents/mesh/utilities/ComputationalGeometry.hpp
```diff
@@ -435,6 +434,375 @@ bool isPointInsidePolyhedron( arrayView2d< real64 const, nodes::REFERENCE_POSITI
...
+ * @param[in] nodesToElements map from nodes to elements
+ * @param[in] nodeLocalToGlobal global indices of nodes
+ * @param[in] elementLocalToGlobal global indices of elements
+ * @param[in] elemCenter coordinates of the element centroid
+ * @param[in] point coordinates of the query point
+ * @return whether the point is inside
+ */
+template< typename COORD_TYPE, typename POINT_TYPE >
+GEOS_HOST_DEVICE
+bool isPointInsideConvexPolyhedronRobust( localIndex element,
```
untereiner: is it possible to have two functions:
`integer computeWindingNumber(...)` and 
```
bool isPointInsideConvexPolyhedronRobust() 
{
 return computeWindingNumber() != 0;
}
 ```

untereiner (commented):
Is it possible to add a test into` testComputationalGeometry.cpp` ?

src/coreComponents/mesh/utilities/ComputationalGeometry.hpp
sframba: It was supposed to be `is chosen from the list`. I have corrected the comment

src/coreComponents/mesh/utilities/ComputationalGeometry.hpp
sframba: Yes, for exemple here:
https://github.com/GEOS-DEV/GEOS/blob/9cf5ab5540ea296194363713f1bb3972dc618d53/src/coreComponents/physicsSolvers/wavePropagation/shared/PrecomputeSourcesAndReceiversKernel.hpp#L60
mesh coordinates (`nodeCoords`) are of type `WaveSolverBase::wsCoordType` (whihc is `real32`), while source and receiver coordinates are plain old `real64`. This is done for memory reasons on GPU, where having all mesh coordinates in `real64` can be too expensive

src/coreComponents/mesh/utilities/ComputationalGeometry.hpp
sframba: Done

untereiner (approved):

sframba:
@rrsettgast Is it good to go? If so, could you please add your review?

rrsettgast (approved):

PR #3206, opened 2024-07-02
feat: Use stability test in flash

dkachuma:
This PR adds a stability test before launching the negative two-phase flash. If the mixture is not found to be unstable for the test compositions during the stability test, it is simply labelled as liquid or vapour using the Li-correlation as described in the [documentation](https://geosx-geosx.readthedocs-hosted.com/en/latest/coreComponents/constitutive/docs/CompositionalMultiphaseFluid.html). Only when the mixture is found to be unstable, is the negative two-phase flash launched.

- The Li-correlation requires the critical volumes. This makes the critical volume an input for all the two-phase compositional models (not just the ones that use LBC viscosity).
- The stability test is changed. Once a test composition is found that makes the TPD negative, the iterations are terminated even before reaching the stationary point.
- A description of the flash procedure is added to the `CompositionalMultiphaseFluid` rst.
- Some heuristics for single phase detection are removed from the flash.


paveltomin:
review?

paveltomin (approved):

CusiniM (approved):

corbett5 (approved):

src/coreComponents/constitutive/unitTests/testStabilityTest9Comp.cpp
corbett5, line 1: Where are these values coming from? Why do they need to be updated? It does seems like this is more of a hybrid unit/integrated test since we are comparing with pretty extensive baseline data.

dkachuma: These values are coming from comparison with another package ([darts-flash](https://gitlab.com/open-darts/darts-flash)).

They are being changed here because I changed the stability test to terminate iterations once instability is determined. Due to this, the TPD (updated values) that determines instability is slightly different. The importance of these values is in whether or not they are significantly negative.

I like to have these rather different data sets because they cover more branch combinations in the code.

PR #3052, opened 2024-03-21
Umpire logs refactor - table output + adding percentages

arng40:
#### What we currently have :

```
Umpire          DEVICE sum across ranks: 15293.4 GB
Umpire          DEVICE         rank max:   31.3 GB
Umpire       DEVICE::0 sum across ranks: 15293.4 G
Umpire       DEVICE::0         rank max:   31.3 GB
Umpire            HOST sum across ranks: 7272.2 GB
Umpire            HOST         rank max:   14.7 GB
Umpire HYPRE_DEVICE_POOL sum across ranks: 7178.4 GB
Umpire HYPRE_DEVICE_POOL         rank max:   14.8 GB
Umpire          PINNED sum across ranks:  256.3 GB
Umpire          PINNED         rank max:  589.6 MB
```

#### This is what I propose inspired by the display proposed by @victorapm 
```  
+-----------------------------+---------------------+---------------------+---------------------+--------------------+
|         Umpire Memory Pool  |     Min over ranks  |    Max  over ranks  |    Avg  over ranks  |    Sum over ranks  |
|  (reserved / % over total)  |                     |                     |                     |                    |
+-----------------------------+---------------------+---------------------+---------------------+--------------------+
|                     DEVICE  |    1.4 MB   (0.0%)  |    1.7 MB   (0.0%)  |    1.5 MB   (0.0%)  |   6.1 MB   (0.0%)  |
|                  DEVICE::0  |    1.4 MB   (0.0%)  |    1.7 MB   (0.0%)  |    1.5 MB   (0.0%)  |   6.1 MB   (0.0%)  |
|                       HOST  |    2.3 MB   (0.0%)  |    2.5 MB   (0.0%)  |    2.4 MB   (0.0%)  |   9.8 MB   (0.0%)  |
|              HYPRE_UM_POOL  |  504.7 KB   (0.0%)  |  645.8 KB   (0.0%)  |  542.4 KB   (0.0%)  |   2.1 MB   (0.0%)  |
|                     PINNED  |  488.1 KB   (0.0%)  |  963.0 KB   (0.0%)  |  725.5 KB   (0.0%)  |   2.8 MB   (0.0%)  |
|                         UM  |    4.0 GB  (25.3%)  |    4.0 GB  (25.3%)  |    4.0 GB  (25.3%)  |  16.0 GB  (25.3%)  |
+-----------------------------+---------------------+---------------------+---------------------+--------------------+
```
#### Code side
[...]

PR #3207, opened 2024-07-02
fix: Check for existence of_ SC_AVPHYS_PAGES to fix MacOS compilation. 

CusiniM:
The merge of [PR](https://github.com/GEOS-DEV/GEOS/pull/3052) has broken compilation on my mac laptop. I have patched it like this.


rrsettgast:
Time to use the GitHub M1 runners?

rrsettgast (approved):
Is there another option on the Mac?

CusiniM:
> Is there another option on the Mac?



> Is there another option on the Mac?

I found [this](https://stackoverflow.com/questions/30511579/linux-sc-phys-pages-not-working-on-mac-os-x) but it's kind of old and I have not tried it.

rrsettgast:
> > Is there another option on the Mac?
> 
> > Is there another option on the Mac?
> 
> I found [this](https://stackoverflow.com/questions/30511579/linux-sc-phys-pages-not-working-on-mac-os-x) but it's kind of old and I have not tried it.

@arng40 Can you give it a shot? Do you have a Mac?

arng40:
> > > Is there another option on the Mac?
> > 
> > 
> > > Is there another option on the Mac?
> > 
> > 
> > I found [this](https://stackoverflow.com/questions/30511579/linux-sc-phys-pages-not-working-on-mac-os-x) but it's kind of old and I have not tried it.
> 
> @arng40 Can you give it a shot? Do you have a Mac?

I don't have a MAC so it would be difficult to test alternatives.
A solution could be to integrate a compilation check on macos in CI

src/coreComponents/common/MemoryInfos.cpp
```diff
@@ -27,8 +27,13 @@ MemoryInfos::MemoryInfos( umpire::MemoryResourceTraits::resource_type resourceTy
     case umpire::MemoryResourceTraits::resource_type::pinned:
       #if defined( _SC_PHYS_PAGES ) && defined( _SC_PAGESIZE )
       m_totalMemory = sysconf( _SC_PHYS_PAGES ) * sysconf( _SC_PAGESIZE );
+      #if defined(_SC_AVPHYS_PAGES)
       m_availableMemory = sysconf( _SC_AVPHYS_PAGES ) * sysconf( _SC_PAGESIZE );
       #else
+      GEOS_WARNING( "Unknown device avaialable memory size getter for this system." );
```
MelReyCG, line 33:
```suggestion
      GEOS_WARNING( "Unknown device available memory size getter for this system." );
```

src/coreComponents/common/MemoryInfos.cpp
```diff
@@ -27,8 +27,13 @@ MemoryInfos::MemoryInfos( umpire::MemoryResourceTraits::resource_type resourceTy
     case umpire::MemoryResourceTraits::resource_type::pinned:
       #if defined( _SC_PHYS_PAGES ) && defined( _SC_PAGESIZE )
       m_totalMemory = sysconf( _SC_PHYS_PAGES ) * sysconf( _SC_PAGESIZE );
+      #if defined(_SC_AVPHYS_PAGES)
       m_availableMemory = sysconf( _SC_AVPHYS_PAGES ) * sysconf( _SC_PAGESIZE );
       #else
```
MelReyCG, line 32: Would something like this be a way to find the available memory on MaxOS? Unfortunately, I we don't have any MacOS to test that here.

```suggestion
      #elif /* is compiling for MacOS */
        {
          uint64_t mem;
          size_t len = sizeof( mem );
          sysctlbyname( "hw.memsize", &mem, &len, nullptr, 0 );
          m_availableMemory= mem / sysconf(_SC_PAGE_SIZE);
        }
```

Should we regroup the system-dependant component like this in a namespace ?

MelReyCG (commented):
Hello there, a few comments to contribute on this issue.

CusiniM:
> > > > Is there another option on the Mac?
> > > 
> > > 
> > > > Is there another option on the Mac?
> > > 
> > > 
> > > I found [this](https://stackoverflow.com/questions/30511579/linux-sc-phys-pages-not-working-on-mac-os-x) but it's kind of old and I have not tried it.
> > 
> > 
> > @arng40 Can you give it a shot? Do you have a Mac?
> 
> I don't have a MAC so it would be difficult to test alternatives. A solution could be to integrate a compilation check on macos in CI

is it now available? I thought that we could not find any M1 runner but if they now have one we really should. It's true that we don't really need to support MacOS but a lot of us work on macbooks so it would be convenient to have a CI job to cover it.

PR #3209, opened 2024-07-03
fix: allow for both linear and quadratic dissipation functions with SpectralSplit

CusiniM:
There was an hardcoded preprocessor variiable that forced the computation of the degradation function for the `SpectralSplit` to only be valid for a linear dissipation function. This PR removes that ifdef and makes both options available. 

Because of the way the model is currently used I had to come up with a mixed solution. The selection is done statically for the phase field damage kernels but dynamically for the smallStrainUpdate. It's fairly easy to template the full damage model but I want to have a chat with @frankfeifan about the formulation and what degradation/dissipation functions combinations are actually valid.


src/coreComponents/physicsSolvers/simplePDE/PhaseFieldDamageFEMKernels.hpp
```diff
@@ -229,28 +229,28 @@ class PhaseFieldDamageKernel :
       {
```
CusiniM, line 216: we should make this if become an `if constexpr` by adding the dissipation function type as a template parameter for this kernel.

src/coreComponents/physicsSolvers/simplePDE/PhaseFieldDamageFEMKernels.hpp
```diff
@@ -229,28 +229,28 @@ class PhaseFieldDamageKernel :
...
         stack.localResidual[ a ] -= detJ * ( N[a] * qp_damage
                                              + ( ell * ell * LvArray::tensorOps::AiBi< 3 >( qp_grad_damage, dNdX[a] )
-                                                 + N[a] * (ell*strainEnergyDensity/Gc) * m_constitutiveUpdate.getDegradationDerivative( qp_damage ) ) );
+                                                 + N[a] * (ell*strainEnergyDensity/Gc) * m_constitutiveUpdate.template getDegradationDerivative< 2 >( qp_damage ) ) );
 
       }
       for( localIndex b = 0; b < numNodesPerElem; ++b )
       {
         if( m_localDissipationOption == LocalDissipation::Linear )
```
CusiniM: same here

paveltomin:
@frankfeifan @rrsettgast is this still needed?

paveltomin (approved):

frankfeifan:
> @frankfeifan @rrsettgast is this still needed?

Sorry, I was swamped with several things and just saw this message. IMO, I'd like to keep this PR as a place to start cleaning up some codes of the phase-field. I will find some time to revive this.

PR #3213, opened 2024-07-05
refactor: Add component selector for fluid models

dkachuma:
In [IsothermalCompositionalMultiphaseFVMKernels.hpp:2127-2157](https://github.com/GEOS-DEV/GEOS/blob/94a0e86aad417bff6d52c082fe1c898647cf8ca2/src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseFVMKernels.hpp#L2127-L2157) and [ThermalCompositionalMultiphaseFVMKernels.hpp:1397-1432](https://github.com/GEOS-DEV/GEOS/blob/94a0e86aad417bff6d52c082fe1c898647cf8ca2/src/coreComponents/physicsSolvers/fluidFlow/ThermalCompositionalMultiphaseFVMKernels.hpp#L1397-L1432), we have calls of the form
```cpp
constitutive::constitutiveUpdatePassThru( fluidBase, [&]( auto & fluid )
{
  using FluidType = TYPEOFREF( fluid );
  // ...
  isothermalCompositionalMultiphaseBaseKernels::internal::kernelLaunchSelectorCompSwitch( numComps, [&]( auto NC )
  {
    integer constexpr NUM_COMP = NC();

    // ...
    using kernelType = DirichletFaceBasedAssemblyKernel< NUM_COMP, NUM_DOF, typename FluidType::KernelWrapper >;
    
    // ...
  } );
} );
```
This expands over all available fluid types and within each expansion, we further expand over all possible numbers of components. The possible number of components is currently set to be from 1 to 5. However, not all fluid types support all these component numbers. For example the `CO2BrineFluid` models support only 2 components. As such we end up here with unnecessary expansions. The end result of all these expansions is to make the `CompositionalMultiphaseFVM` class extremely large to compile especially for CUDA builds. This is so bad that some of the fluid models have been commented out in [MultiFluidSelector.hpp:39-58](https://github.com/GEOS-DEV/GEOS/blob/94a0e86aad417bff6d52c082fe1c898647cf8ca2/src/coreComponents/constitutive/fluid/multifluid/MultiFluidSelector.hpp#L39-L58) just to be able to compile on GHA.

This PR tries to reduce the number of combinations by explicitly listing, for each fluid type, the number of components that it allows. The expansion is a combination of the fluid model type and number of components. Furthermore, not all fluid models support thermal simulations. So the thermal kernels are expanded only over those models that support thermal simulations. 

Some other changes
- Each fluid model class is tagged with a static boolean `isThermalType()` method.
- Some validation of the user input has been added to ensure that the number of components listed in `MultiFluidSelector` are honoured.
- Added a couple of integrated tests to include the constant pressure boundary condition.


paveltomin (approved):

src/coreComponents/constitutive/fluid/multifluid/MultiFluidSelector.hpp
```diff
@@ -79,6 +79,111 @@ void constitutiveUpdatePassThru( MultiFluidBase & fluid,
...
+};
+template<>
+struct Components< CompositionalTwoPhaseLohrenzBrayClarkViscosity >
+{
+  using type = camp::int_seq< integer, 2, 3, 4, 5 >;
+};
+template<>
+struct Components< CompositionalTwoPhaseConstantViscosity >
+{
+  using type = camp::int_seq< integer, 2, 3, 4, 5 >;
```
CusiniM: I think that you can do something like this to avoid specifying each number:

````c++
template<integer Begin, integer End, integer... Is>
struct create_integer_sequence_impl
    :  create_integer_sequence_impl<Begin, End - 1, End - 1, Is...> {};

template<integer Begin, Integer... Is>
struct create_integer_sequence_impl<Begin, Begin, Is...>
{
    using type = camp::int_seq<integer, Is...>;
};
````
and then use it like:

````c++
template<>
struct Components< CompositionalTwoPhaseConstantViscosity >
{
using type = typename create_integer_sequence_impl<2, 5>::type;
}
````

and maybe I would define the value in the max and min value of components as constexpr in each fluid model itself and have something like:

````c++
template<typename FLUID_TYPE>
struct Components< >
{
using type = typename create_integer_sequence_impl<FLUID_TYPE::min_n_components, FLUID_TYPE::max_n_components>::type;
}
````
@wrtobin thoughts?

Comment (wrtobin) [reply src/coreComponents/constitutive/fluid/multifluid/MultiFluidSelector.hpp]: IIRC camp has builtin support for range generation... but looking at it it seems like they only support [0...N) generation.
So yeah we should be able to avoid explicitly listing all numbers unless there is a compelling reason to do so.

I think its typically logical to add a third component for the step size which defaults to 1, so we can e.g. count by 2's etc, but that is just as a general mechanism.

Comment (CusiniM) [reply src/coreComponents/constitutive/fluid/multifluid/MultiFluidSelector.hpp]: if we make it even more generic it could probably even go in `common` or in `datatypes`. Honestly, it should probably live directly in `camp`. I also had a look in `camp` and I didn't see exactly what we need. 

In any case I don't see a real reason for explicitly listing all numbers, so I think we should do something like what I wrote above but @dkachuma feel free to also the step size and move it somewhere else.

Comment (dkachuma) [reply src/coreComponents/constitutive/fluid/multifluid/MultiFluidSelector.hpp]: I particularly like the idea of min/max component with each fluid. Let me see how I can fit this in.

File: src/coreComponents/common/TypeDispatch.hpp [hunk 2]
```diff
@@ -153,6 +153,12 @@ using Increment = typename IncrementImpl< T, N >::type;
 
 } // namespace internal
 
+/**
+ * @brief Generate a sequence of integers from @p Begin up to (and including) @p End.
+ */
+template< integer Begin, integer End >
+using IntegerSequence = internal::Increment< camp::make_idx_seq_t< End - Begin + 1 >, Begin >;
```
CusiniM, line 161: nice! We did have something like this already.

src/coreComponents/constitutive/fluid/multifluid/MultiFluidSelector.hpp
```diff
@@ -79,6 +79,71 @@ void constitutiveUpdatePassThru( MultiFluidBase & fluid,
...
 
+namespace detail
+{
+/**
+ * @brief Structure to execute a lambda on a fluid model depending on the number of components
+ * @tparam THERMAL_ACTIVE Determines if the call has been deactivated due to the lambda being thermal
+ *         but the fluid model not being thermal
+ * @tparam COMPONENT_LIST Type listing the components to expand over
+ */
+template< bool THERMAL_ACTIVE, typename COMPONENT_LIST >
```
CusiniM: this is really more like `THERMAL_ACTIVEA_AND_SUPPORTED` right?

src/coreComponents/constitutive/fluid/multifluid/MultiFluidSelector.hpp
```diff
@@ -79,6 +79,71 @@ void constitutiveUpdatePassThru( MultiFluidBase & fluid,
...
+struct ComponentSelector;
+
+/** @brief If the call is deactivated due to the fluid not being thermal
+ */
+template< typename COMPONENT_LIST >
+struct ComponentSelector< false, COMPONENT_LIST >
```
CusiniM: instead of specializing the implementation you can `if constexpr` that boolean template parameter and have a single implementation. I let you decide where to put the `if constexpr`. It could even go directly inside `constitutiveComponentUpdatePassThru` and do something like:

````c++
if constexpr (active)
{
detail::ComponentSelector< active, Components >::execute( numComps, fluid, std::forward< LAMBDA >( lambda ));
} 
else
{
/// throw an error. 
}

CusiniM (approved):

MelReyCG (approved):
I approve the `common/TypeDispatch` part.

Just a side question, what do you think of the name `isThermalType()` to keep the `isX()` naming with the `thermal()` static method?

rrsettgast (approved):

PR #3194, opened 2024-06-28
fix: Use aperture table in poromechanics with conforming fractures

paveltomin:
Main thing:
replace
`
hydraulicAperture[k] = minimumHydraulicAperture[k] + aperture[k];
`
with
`
hydraulicAperture[k] = contactWrapper.computeHydraulicAperture( aperture[k], dHydraulicAperture_dNormalJump );
`
for poromechanics with conforming fractures.
At least this way there will be no negative aperture and no negative volume.

Small things:
- Separately account for new slip in `computeFractureStateStatistics` to see new segments slipped.
- Output normal displacement tolerance, sliding tolerance, normal traction tolerance when `logLevel>=2`.
- For `logLevel>=10` - display fracture state changes with debug information.
- Small fixes for `deltaVolume` logic: assign it to 0 once applied into volume (to avoid double counting), update mass with updated volume. Those are important for fractures after poromechanics initialization that changes the volume of the fractures.

PR #3215, opened 2024-07-06
refactor: fix creation of mass handling.

CusiniM:
This PR simply moves the location of the assignment of the `createdMass` variable that tracks how much mass is created after each propagation step. This allows to set its default values to zero so taht we don't see artifacts in simulations that do not include any propagation and for which that variable is useless. 
Personally, I think that the initialization of new elements and how that `createdMass` is used should be revisited. For example, it should be equal to `mass_n` for a new element and currently it is not. 

However, it is easier do that separately because it will impact the solution of the hydraulic fracturing tests so I propose that we start by merging this simple change and address the actual algorithm later.


src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp
```diff
@@ -1067,13 +1071,15 @@ void HydrofractureSolver< POROMECHANICS_SOLVER >::initializeNewFractureFields( D
...
               initialAperture = std::min( initialAperture, aperture[fractureElementIndex] );
+              initialDensity  = std::min( initialDensity, fluidDensity_n[fractureElementIndex] );
             }
           }
         }
         if( m_newFractureInitializationType == InitializationType::Pressure )
         {
           fluidPressure[newElemIndex] = initialPressure > 1.0e98? 0.0:initialPressure;
           fluidPressure_n[newElemIndex] = fluidPressure[newElemIndex];
+          createdMass[newElemIndex] =  fluidPressure[newElemIndex] * initialAperture * elementArea;
```
CusiniM:
```suggestion
          createdMass[newElemIndex] =  initialDensity[newElemIndex] * initialAperture * elementArea;
```

src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp
```diff
@@ -1067,13 +1071,15 @@ void HydrofractureSolver< POROMECHANICS_SOLVER >::initializeNewFractureFields( D
...
               initialAperture = std::min( initialAperture, aperture[fractureElementIndex] );
+              initialDensity  = std::min( initialDensity, fluidDensity_n[fractureElementIndex] );
             }
           }
         }
         if( m_newFractureInitializationType == InitializationType::Pressure )
         {
           fluidPressure[newElemIndex] = initialPressure > 1.0e98? 0.0:initialPressure;
           fluidPressure_n[newElemIndex] = fluidPressure[newElemIndex];
+          createdMass[newElemIndex] =  initialDensity[newElemIndex] * initialAperture * elementArea;
```
CusiniM:
```suggestion
          createdMass[newElemIndex] =  initialDensity * initialAperture * elementArea;
```

paveltomin (approved):

src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp
```diff
@@ -1067,13 +1071,15 @@ void HydrofractureSolver< POROMECHANICS_SOLVER >::initializeNewFractureFields( D
             {
               initialPressure = std::min( initialPressure, fluidPressure_n[fractureElementIndex] );
```
frankfeifan: I wonder why we use the `fluidPressure_n`of the neighboring fracture element instead of `fluidPressure`? Same question for the `fluidDensity`.

frankfeifan (approved):

src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp
CusiniM: after a propagation stage we should reset all those variables and resolve so I would expect these two to be the same.

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBaseKernels.hpp
```diff
@@ -355,7 +355,7 @@ class SurfaceElementBasedAssemblyKernel : public ElementBasedAssemblyKernel< Sur
                                      CRSMatrixView< real64, globalIndex const > const & localMatrix,
                                      arrayView1d< real64 > const & localRhs )
     : Base( rankOffset, dofKey, subRegion, fluid, solid, localMatrix, localRhs )
-    , m_creationMass( subRegion.getReference< array1d< real64 > >( SurfaceElementSubRegion::viewKeyStruct::creationMassString() ) )
+    , m_creationMass( subRegion.getField< fields::flow::massCreated >(); )
```
CusiniM:
```suggestion
    , m_creationMass( subRegion.getField< fields::flow::massCreated >() )
```

ryar9534:
Does this interact at all with #3194?

rrsettgast:
@CusiniM I remember the reason that we can't get rid of the creationMass is because there isn't anywhere to steal it from in the case of a viscosity dominated case with no pore fluid in the rock. For example, in a KGD when creating a new face (new tip), the old tip has a negative pressure and taking mass from it will mess up the simulation. We really need to implement some concept of partial fracture or saturation at the tip to avoid this.

In the case where there is pore fluid, we should steal it from the rock pores. If there is no pore fluid...then there is no good answer except for a partial fracture/saturation treatment that I can think of

paveltomin:
> Does this interact at all with #3194?

It should not, I reverted everything related to mass creation in that PR

CusiniM:
> @CusiniM I remember the reason that we can't get rid of the creationMass is because there isn't anywhere to steal it from in the case of a viscosity dominated case with no pore fluid in the rock. For example, in a KGD when creating a new face (new tip), the old tip has a negative pressure and taking mass from it will mess up the simulation. We really need to implement some concept of partial fracture or saturation at the tip to avoid this.
> 
> In the case where there is pore fluid, we should steal it from the rock pores. If there is no pore fluid...then there is no good answer except for a partial fracture/saturation treatment that I can think of

Yeah, I have not changed anything of the logic of how it's used. I have only changed where it's initialized. 

I still think that, since we use an implicit scheme, `mass_n` should be `0` for a newly created element. This would make the scheme conservative. I do remember that we struggle with convergence though. The good news is that, in reality, the case with no pore fluid in the rock is only useful to match analytical solutions because all real applications should consider a saturated rock around it.

wrtobin (approved):

src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp
```diff
@@ -1020,8 +1021,15 @@ void HydrofractureSolver< POROMECHANICS_SOLVER >::initializeNewFractureFields( D
 
       arrayView1d< real64 > const fluidPressure_n = subRegion.getField< fields::flow::pressure_n >();
       arrayView1d< real64 > const fluidPressure = subRegion.getField< fields::flow::pressure >();
+      string const & fluidName = subRegion.getReference< string >( FlowSolverBase::viewKeyStruct::fluidNamesString() )
```
CusiniM:
```suggestion
      string const & fluidName = subRegion.getReference< string >( FlowSolverBase::viewKeyStruct::fluidNamesString() );
```

src/coreComponents/mesh/SurfaceElementRegion.cpp
```diff
@@ -101,8 +101,6 @@ localIndex SurfaceElementRegion::addToFractureMesh( real64 const time_np1,
 
   arrayView1d< real64 > const ruptureTime = subRegion.getField< fields::ruptureTime >();
 
-  arrayView1d< real64 > const creationMass = subRegion.getReference< real64_array >( FaceElementSubRegion::viewKeyStruct::creationMassString() );
-
   arrayView2d< real64 const > const faceCenter = faceManager->faceCenter();
   arrayView2d< real64 > const elemCenter = subRegion.getElementCenter();
   arrayView1d< real64 const > const elemArea = subRegion.getElementArea().toViewConst();
```
ryar9534: elemArea seems to be unused now

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.cpp
```diff
@@ -94,13 +94,21 @@ void SinglePhaseBase::registerDataOnMesh( Group & meshBodies )
...
       subRegion.registerField< fields::flow::mass_n >( getName() );
+      
 
       if( m_isThermal )
       {
         subRegion.registerField< dMobility_dTemperature >( getName() );
       }
     } );
 
+    elemManager.forElementSubRegions< FaceElementSubRegion >( regionNames,
```
CusiniM:
```suggestion
    elemManager.forElementSubRegions< SurfaceElementSubRegion >( regionNames,
```

src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp
```diff
@@ -1009,149 +1010,164 @@ void HydrofractureSolver< POROMECHANICS_SOLVER >::initializeNewFractureFields( D
...
-      ArrayOfArraysView< localIndex const > const faceMap = subRegion.faceList().toViewConst();
+        ArrayOfArraysView< localIndex const > const faceMap = subRegion.faceList().toViewConst();
 
-      arrayView1d< real64 > const fluidPressure_n = subRegion.getField< fields::flow::pressure_n >();
-      arrayView1d< real64 > const fluidPressure = subRegion.getField< fields::flow::pressure >();
+        arrayView1d< real64 > const fluidPressure_n = subRegion.getField< fields::flow::pressure_n >();
+        arrayView1d< real64 > const fluidPressure = subRegion.getField< fields::flow::pressure >();
+        string const & fluidName = subRegion.getReference< string >( FlowSolverBase::viewKeyStruct::fluidNamesString() );
+        SingleFluidBase const & fluid = subRegion.getConstitutiveModel< SingleFluidBase >( fluidName );
+        // arrayView2d< real64 const > const fluidDensity_n = fluid.density_n();
```
CusiniM:
```suggestion
```

src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp
```diff
@@ -1009,149 +1010,164 @@ void HydrofractureSolver< POROMECHANICS_SOLVER >::initializeNewFractureFields( D
...
+        real64 const defaultDensity = fluid.defaultDensity();
+        arrayView1d< real64 > const massCreated  = subRegion.getField< fields::flow::massCreated >();
+        // arrayView1d< real64 > const mass_n  = subRegion.getField< fields::flow::mass_n >();
```
CusiniM:
```suggestion
```

src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp
```diff
@@ -1009,149 +1010,164 @@ void HydrofractureSolver< POROMECHANICS_SOLVER >::initializeNewFractureFields( D
...
+        // zero
 
-          auto connIter = edgesToConnectorEdges.find( edgeIndex );
-          if( connIter == edgesToConnectorEdges.end() )
+        forAll< serialPolicy >( newFractureElements.size(), [&] ( localIndex const k )
+        {
+          localIndex const newElemIndex = newFractureElements[k];
+          real64 initialPressure = 1.0e99;
+          real64 initialAperture = 1.0e99;
+          // real64 initialDensity = 1.0e99;
```
CusiniM:
```suggestion
```

src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp
```diff
@@ -1009,149 +1010,164 @@ void HydrofractureSolver< POROMECHANICS_SOLVER >::initializeNewFractureFields( D
...
+            for( localIndex kfe=0; kfe<numElems; ++kfe )
             {
-              zeroDisp = false;
+              localIndex const fractureElementIndex = fractureConnectorsToFaceElements[connectorIndex][kfe];
+
+              if( newFractureElements.count( fractureElementIndex ) == 0 )
+              {
+                initialPressure = std::min( initialPressure, fluidPressure_n[fractureElementIndex] );
+                initialAperture = std::min( initialAperture, aperture[fractureElementIndex] );
+                // initialDensity  = std::min( initialDensity, fluidDensity_n[fractureElementIndex][0] );
```
CusiniM:
```suggestion
```

rrsettgast (approved):

BASELINE_NOTES.md
```diff
@@ -6,10 +6,9 @@ This file is designed to track changes to the integrated test baselines.
 Any developer who updates the baseline ID in the .integrated_tests.yaml file is expected to create an entry in this file with the pull request number, date, and their justification for rebaselining.
 These notes should be in reverse-chronological order, and use the following time format: (YYYY-MM-DD).
 
-PR #3194 (2024-07-10)
+PR #3215 (2024-07-23)
```
CusiniM, line 9:
```suggestion
PR #3215 (2024-07-23)
======================
Changed the default value for massCreation and name of the wrapper.

PR #3194 (2024-07-10)
======================
Use aperture table in poromechanics with conforming fractures. Rebaseline the corresponding cases.

```

BASELINE_NOTES.md
```diff
@@ -6,10 +6,9 @@ This file is designed to track changes to the integrated test baselines.
...
 ======================
-Use aperture table in poromechanics with conforming fractures. Rebaseline the corresponding cases.
-
+Changed the default value for massCreation and name of the wrapper.
```
CusiniM, line 11:
```suggestion
```

BASELINE_NOTES.md
```diff
@@ -6,10 +6,15 @@ This file is designed to track changes to the integrated test baselines.
...
 
+PR #3215 (2024-07-23)
+======================
+
 PR #3194 (2024-07-10)
 ======================
 Use aperture table in poromechanics with conforming fractures. Rebaseline the corresponding cases.
 
+======================
+Changed the default value for massCreation and name of the wrapper.
```
CusiniM:
```suggestion
```

BASELINE_NOTES.md
```diff
@@ -6,10 +6,15 @@ This file is designed to track changes to the integrated test baselines.
 Any developer who updates the baseline ID in the .integrated_tests.yaml file is expected to create an entry in this file with the pull request number, date, and their justification for rebaselining.
 These notes should be in reverse-chronological order, and use the following time format: (YYYY-MM-DD).
 
+PR #3215 (2024-07-23)
+======================
```
CusiniM, line 10:
```suggestion
======================
Changed the default value for massCreation and name of the wrapper.
```

.integrated_tests.yaml
```diff
@@ -2,6 +2,7 @@ baselines:
   bucket: geosx
   baseline: integratedTests/baseline_integratedTests-pr3233-6393-534ac25
```
```suggestion
  baseline: integratedTests/baseline_integratedTests-pr3215-6396-9b43cc8
```

.integrated_tests.yaml
```diff
@@ -2,6 +2,7 @@ baselines:
...
 
+
```
CusiniM, line 4:
```suggestion
```

PR #3224, opened 2024-07-12
feat: Add Taper Layers in wave solvers

acitrain:
This PR add a first version of Taper layers for wave solvers
These layers are set to improve the efficiency of absoption of boundaries.
This is how the version version works: 
- We compute a taper profile which is equal to one when you are outside the taper layers and to : 
$d_x = \exp\left(\displaystyle \frac{3V_{\mathrm{min}}}{2L}\log(R)\left(\frac{L-x}{L}\right)^2\cdot \Delta t\right)$ where $V_{\mathrm{min}}$ is the minimum P-wavespeed, L the length of the taper layer and $\log(R)$ is the log reflectivitity coefficient.
- Then we multiply by the array at time n and n+1 by the taperCoeff

For this first version only second order solver contains the taper functionality.

In a future PR, we can imagine to make the taper profile independant of the timestep and use the frequency instead.


acitrain:
Hi @rrsettgast 

Could you take a look at it and approve if it seems good to you?
It is ready except integrated test but I want to do it after a code owner review 

Thank you ! 

PS: I don't know why the docs build does not pass, it is only saying "We encountered a problem with a command while building your project. To resolve this error, double check your project configuration and installed dependencies are correct and have not changed recently."

acitrain:
@sframba  @rrsettgast @CusiniM can you review this please ? 

Thanks

rmadec-cs (approved):

src/coreComponents/physicsSolvers/wavePropagation/sem/elastic/secondOrderEqn/isotropic/ElasticWaveEquationSEM.cpp
```diff
@@ -457,15 +479,40 @@ void ElasticWaveEquationSEM::initializePostInitialConditionsPreSubGroups()
...
+          m_timeStep = subEvent->getReference< real64 >( EventBase::viewKeyStruct::forceDtString() );
+        }
+      }
+
+    }
+
+    if( m_useTaper==1 )
+    {
+      real32 vMin;
+      vMin = getGlobalMinWavespeed( mesh, regionNames );
```
Victor-M-Gomes: Hello,

The current definition is not optimal. 

For example: When `m_timestepStabilityLimit==1` a correct dt (`m_timeStep/dtOut`) value is calculated, however, by setting `m_timestepStabilityLimit==0` obliges all other calls to `initializePostInitialConditionsPreSubGroups()` to use the user-input dt (either from xml or pygeosx). This scenario is often present when interfacing GEOS through python (pygeosx).

One option would be to never set `m_timestepStabilityLimit==0`, however that implicates a dt calculation at each new call to `initializePostInitialConditionsPreSubGroups()`. This may be useful for certain applications, but one often just wants to calculate the dt once. Besides that, at the current state, if  `initializePostInitialConditionsPreSubGroups()` is called a second time to calculate the dt (meaning that `m_timestepStabilityLimit==1` is always true), the dt calculation does not work, giving `nan` values.

After exploring a bit, the `nan` values are due to a de-synchronization between GPU and CPU copies in the `computeTimeStep` function. By calling `initializePostInitialConditionsPreSubGroups()`, and therefore `computeTimeStep`, a second time, the ux_n, uy_n, and uz_n arrays are not updated by the loop in lines 553-558, thus the division operations in lines 569-574 will give `nan`. 

A simple workaround is to avoid:
```
for( localIndex a = 0; a < sizeNode; ++a )
    {
      ux_n[a] = (real64)rand()/(real64) RAND_MAX;
      uy_n[a] = (real64)rand()/(real64) RAND_MAX;
      uz_n[a] = (real64)rand()/(real64) RAND_MAX;
}
```
simply writing:

```

for( localIndex a = 0; a < sizeNode; ++a )
    {
      ux_n[a] = 1.0;
      uy_n[a] = 1.0;
      uz_n[a] = 1.0;
}
```
and add the following loop to ensure the GPU values are not zero:

```
forAll< EXEC_POLICY >( sizeNode, [=] GEOS_HOST_DEVICE ( localIndex const a )
    {
      ux_n[a] = 1.0;
      uy_n[a] = 1.0;
      uz_n[a] = 1.0;
} );
```
Finally, it is not clear to me why we would need: `return m_timeStep * m_cflFactor;` in line 678.

Notice that these comments are valid to the acoustic case also.

Coming back to the origin of the problem, it may be wise to add a third state `m_timestepStabilityLimit==2` (for example). When `m_timestepStabilityLimit==1` the dt is always calculated, if `m_timestepStabilityLimit==2` the dt is calculated once and then set to state 3 (one that does not exist, implying that no further modification of `m_timeStep` will take place), and if `m_timestepStabilityLimit==0` the user-input dt is used (xml or pygeosx).

src/coreComponents/physicsSolvers/wavePropagation/shared/TaperKernel.hpp
```diff
@@ -0,0 +1,158 @@
...
+      real32 dist = taperCoeff[a];
+
+      if( dist<sizeT )
+      {
+        taperCoeff[a] = LvArray::math::exp((((3*vMin)/(2*sizeT))*log( r )*pow((sizeT-dist)/sizeT, 2 ))*dt );
+      }
+      else
+      {
+        taperCoeff[a] = 1.0;
+      }
```
Victor-M-Gomes: Great job with the taper, choosing `vMin `over `vMax` was a really good choice since it usually leads to better results. This needs to be modified in the description of the PR though:
$d_x = \displaystyle \frac{-3V_{max}}{2L}log(R)(\frac{x}{L})^2$ becomes $d_x = \exp\left( \displaystyle \frac{3V_{min}}{2L}log(R)(\frac{L-x}{L})^2 dt \right)$ (the dt also appears in the definition).

After much testing the border reflections have been successfully attenuated and modelling results are significantly improved. 

For the future (in another PR maybe) I would suggest substituting $V_{min} \times dt$ by some $\lambda_{min} = V_{min}/f_{max}$, where $f_{max}$ is the maximum frequency that one wishes to consider in the modelling. Right now $f_{max} == f_{sampling}$, the maximum frequency that exists in the data. By allowing the user to input $\lambda_{min}$ or $f_{max}$, the equation would have more physical meaning and the use of $dt$ could be avoided. Another option would be to set $f_{max} \simeq f_{Ricker} \times 3$ (or a similar factor) to avoid defining new inputs. The goal of the suggestion is to avoid using dt in the definition of the taper, while also inserting additional physical meaning.  

Finally, it may be important to mention that the choice of L and R are somewhat empirical. For a given R, the greater the L, the better the attenuation, but that also comes with the cost of increasing the model to fit the taper. Thus, in general a compromise between L and R need to be found. For the bounds of R, I would (based on some tests) place them in the interval [ $10^{-3}$ , 1).

Victor-M-Gomes (commented):
This PR brings an important addition to the SEM wavepropagation module of GEOS. With this new absorbing boundary I've checked that one can achieve significantly better modelling results.

I suggest some small modifications to avoid some problems encountered during testing.

acitrain:
@rrsettgast @CusiniM can one of you take a look to this? 
It is ok on our side we just need a code owner review to go to the merge queue and do integrated test when this PR will be at the top 

Thanks !

src/coreComponents/physicsSolvers/wavePropagation/shared/TaperKernel.hpp
sframba: @Victor-M-Gomes I have changed the PR description accordingly

sframba: Re the dt, I don't find it so absurd having a dt in the formula, since if you cut the timestep in two, you want to apply the same factor as you did before for a given time. But I guess the other parameters are also heurisitc... Thoughts?

src/coreComponents/physicsSolvers/wavePropagation/sem/acoustic/secondOrderEqn/isotropic/AcousticWaveEquationSEM.cpp
```diff
@@ -510,8 +578,10 @@ real64 AcousticWaveEquationSEM::computeTimeStep( real64 & dtOut )
 
     stiffnessVector.zero();
     p.zero();
+    //Lien to ensure that the using array stays on GPU (useful when we cal this routine several times)
+    forAll< parallelHostPolicy >( sizeNode, [p] ( localIndex const ){} );
```
sframba, line 568: Should probably be replaced by a "touch" on CPU

src/coreComponents/physicsSolvers/wavePropagation/sem/acoustic/secondOrderEqn/isotropic/AcousticWaveEquationSEM.cpp
```diff
@@ -510,8 +578,10 @@ real64 AcousticWaveEquationSEM::computeTimeStep( real64 & dtOut )
...
   } );
-  return m_timeStep * m_cflFactor;
+  return 0;
```
sframba: Sure about this?

src/coreComponents/physicsSolvers/wavePropagation/sem/elastic/secondOrderEqn/isotropic/ElasticWaveEquationSEM.cpp
```diff
@@ -633,8 +692,11 @@ real64 ElasticWaveEquationSEM::computeTimeStep( real64 & dtOut )
     ux_n.zero();
     uy_n.zero();
     uz_n.zero();
+    //Lien to ensure that the using array stays on GPU (useful when we cal this routine several times)
+    forAll< parallelHostPolicy >( sizeNode, [ux_n, uy_n, uz_n] ( localIndex const ){} );
```
sframba, line 696: Same comments as above

src/coreComponents/physicsSolvers/wavePropagation/shared/WaveSolverBase.hpp
```diff
@@ -468,6 +486,14 @@ DECLARE_FIELD( referencePosition32,
                NOPLOT,
                WRITE_AND_READ,
                "Copy of the referencePosition from NodeManager in 32 bits integer" );
+DECLARE_FIELD( taperCoeff,
+               "taperCoeff",
+               array1d< real32 >,
+               1.0,
+               LEVEL_0,
```
sframba: NOPLOT

src/coreComponents/unitTests/wavePropagationTests/CMakeLists.txt
```diff
@@ -1,7 +1,8 @@
 # Specify list of tests
 set( gtest_geosx_tests
      testWavePropagation.cpp
-     testWavePropagationQ2.cpp
+     testWavePropagationTaper.cpp
+     #testWavePropagationQ2.cpp
```
sframba: Q2 no longer needed?

herve-gross (approved):
Technical approval

src/coreComponents/physicsSolvers/wavePropagation/sem/acoustic/secondOrderEqn/isotropic/AcousticWaveEquationSEM.cpp
sframba: After discussion, let's keep it like this for now, as it guarantees that corner cases involving `pyGEOS` scripts work as expected.

src/coreComponents/physicsSolvers/wavePropagation/sem/elastic/secondOrderEqn/isotropic/ElasticWaveEquationSEM.cpp
sframba: See above comment

sframba (approved):
Review comments have been implemented or otherwise resolved - approved

rrsettgast (approved):

PR #3227, opened 2024-07-17
refactor: Preparatory for well to frac connection

paveltomin:
This PR includes preparatory changes for supporting well to frac connections.

- replace `CellElementSubRegion` with `ElementSubRegionBase` where possible
- where not possible - use templates
- introduce `targetRegion` for perforations, to be able to connect to matrix or fracture (if `targetRegion` is not specified that means perforation will try to connect to all regions)
- adjust `connectPerforationsToMeshElements` logic - since target region is already defined, loop through subregions in `searchLocalElements`, also adjust `visitNeighborElements` and `initializeLocalSearch`


src/coreComponents/mesh/Perforation.cpp
tjb-ltk, line 1: shouldn't there be a default region

src/coreComponents/mesh/MeshForLoopInterface.hpp
tjb-ltk, line 1: For my own better understanding of feature, generalizing to ElementSubRegionBase allows what additional functionality ?

src/coreComponents/mesh/Perforation.cpp
paveltomin: yeah, i guess empty should mean "connect to all"?

src/coreComponents/mesh/MeshForLoopInterface.hpp
paveltomin: just two truncated versions - one loops though specific region, another through specific subregion, nothing special

src/coreComponents/mesh/Perforation.cpp
paveltomin: should be fine now
not specified = connect to all

paveltomin:
need review please

paveltomin:
@CusiniM, @MelReyCG, @castelletto1, @corbett5, @cssherman, @jhuang2601, @matteofrigo5, @rrsettgast, @untereiner, and/or @wrtobin please review

paveltomin:
need review please

paveltomin:
Waiting on code owner review from @CusiniM, @Guotong-Ren, @castelletto1, @frankfeifan, @jhuang2601, @matteofrigo5, @rrsettgast, @ryar9534, @untereiner, @wrtobin, and/or 2 others.

paveltomin:
@CusiniM, @Guotong-Ren, @castelletto1, @frankfeifan, @jhuang2601, @matteofrigo5, @rrsettgast, @ryar9534, @untereiner, @wrtobin please review

Guotong-Ren (approved):

paveltomin:
@CusiniM, @MelReyCG, @castelletto1, @corbett5, @frankfeifan, @jhuang2601, @rrsettgast, @ryar9534, @untereiner, and/or @wrtobin please review, i need to move this forward, it's been a while

src/coreComponents/mesh/generators/WellGeneratorBase.hpp
```diff
@@ -185,6 +185,12 @@ class WellGeneratorBase : public MeshComponentBase
    */
   arrayView1d< real64 const > getPerfSkinFactor() const { return m_perfSkinFactor; };
 
+  /**
+   * @brief Get the target region for a perforation.
+   * @return the target regions for a perforation
+   */
+  arrayView1d< string const > getPerfTargetRegion() const { return m_perfTargetRegion; };
```
MelReyCG, line 192: Just to be sure to understand, it can connect to multiple region if the perforation is at the interface of two or more regions?

paveltomin: i think it can and the name should "regions", not "region", let me fix it

paveltomin: oh no, you are right, each perf only connects to one region, sorry my confusion

MelReyCG: Ok, thanks for the explanation! It's clearer now. The returned list of these methods are for perforations data here, it imply something like `getPerfsX()`.

src/coreComponents/mesh/Perforation.cpp
```diff
@@ -46,6 +46,11 @@ Perforation::Perforation( string const & name, Group * const parent )
     setApplyDefaultValue( 0.0 ).
     setInputFlag( InputFlags::OPTIONAL ).
     setDescription( "Perforation skin factor" );
+
+  registerWrapper( viewKeyStruct::targetRegionString(), &m_targetRegionName ).
+    setRTTypeName( rtTypes::CustomTypes::groupNameRef ).
+    setInputFlag( InputFlags::OPTIONAL ).
+    setDescription( "Target region to connect the perforation" );
```
MelReyCG, line 53: Why can it be customized by the user? Cannot it be determined automatically?

src/coreComponents/mesh/WellElementSubRegion.cpp
```diff
@@ -137,6 +138,33 @@ void collectElementNodes( CellElementSubRegion const & subRegion,
...
 
+template< typename SUBREGION_TYPE >
+bool isPointInsideElement( SUBREGION_TYPE const & GEOS_UNUSED_PARAM( subRegion ),
+                           arrayView2d< real64 const, nodes::REFERENCE_POSITION_USD > const & GEOS_UNUSED_PARAM( referencePosition ),
+                           localIndex const & GEOS_UNUSED_PARAM( eiLocal ),
+                           ArrayOfArraysView< localIndex const > const & GEOS_UNUSED_PARAM( facesToNodes ),
+                           real64 const (&GEOS_UNUSED_PARAM( elemCenter ))[3],
+                           real64 const (&GEOS_UNUSED_PARAM( location ))[3] )
+{
+  // only CellElementSubRegion is currently supported
```
MelReyCG, line 149: I may be wrong, but this code should not run if not supported.
```suggestion
  static_assert( false, "only CellElementSubRegion is currently supported" );
```

src/coreComponents/mesh/generators/WellGeneratorBase.hpp
paveltomin: yes, same pattern as for other functions/perf. data

src/coreComponents/mesh/Perforation.cpp
paveltomin: the idea here is to be able to connect well to a fracture
fractures are geometrically zero volume objects so a perforation that lies on a fracture may numerically also get connected to a cell block next to it
to avoid this situation and to allow the user to control the behavior, `targetRegion` can be optionally specified
when not provided, GEOS will try to connect perforation with all the regions

src/coreComponents/mesh/WellElementSubRegion.cpp
paveltomin: i think it will start crashing if i add that assert
the needed functions are implemented in https://github.com/GEOS-DEV/GEOS/pull/3359
hope we can merge both soon

paveltomin:
@CusiniM, @MelReyCG, @castelletto1, @corbett5, @frankfeifan, @jhuang2601, @rrsettgast, @ryar9534, @untereiner, and/or @wrtobin need review

inputFiles/poromechanicsFractures/singlePhasePoromechanics_FaultModel_well_fim_new_smoke.xml
CusiniM, line 1: Maybe the first examples should be fracture flow with well only.

src/coreComponents/mesh/Perforation.cpp
CusiniM: this sounds reasonable to me.

src/coreComponents/mesh/WellElementSubRegion.cpp
```diff
@@ -243,6 +271,111 @@ bool visitNeighborElements( MeshLevel const & mesh,
...
+      if( !elements.contains( eiGlobal ))
+      {
+        elements.insert( eiGlobal );
+
+        real64 const elemCenter[3] = { elemCenters[eiLocal][0],
+                                       elemCenters[eiLocal][1],
+                                       elemCenters[eiLocal][2] };
+
+        // perform the test to see if the point is in this reservoir element
+        // if the point is in the resevoir element, save the indices and stop the search
```
CusiniM: why the template parameter `SUBREGION_TYPE`? won't this approach only work for 3D elements?

src/coreComponents/mesh/WellElementSubRegion.cpp
```diff
@@ -243,6 +271,111 @@ bool visitNeighborElements( MeshLevel const & mesh,
   return matched;
 }
 
+/**
+ * @brief Search the reservoir elements that can be accessed from the set "nodes".
+          Stop if a reservoir element containing the perforation is found.
+          If not, enlarge the set "nodes"
+ * @param[in] meshLevel the mesh object (single level only)
+ * @param[in] location the location of that we are trying to match with a reservoir element
+ * @param[inout] nodes the nodes that have already been visited
```
CusiniM: I may not know the algorithm very well but why are we using nodes? Are these the nodes of the 1d well mesh?

src/coreComponents/mesh/WellElementSubRegion.cpp
```diff
@@ -345,6 +513,91 @@ bool searchLocalElements( MeshLevel const & mesh,
...
+ * @brief Search for the reservoir element that contains the well element.
+          To do that, loop over the reservoir elements that are in the neighborhood of (erInit,esrInit,eiInit)
+ * @param[in] meshLevel the mesh object (single level only)
+ * @param[in] location the location of that we are trying to match with a reservoir element
+ * @param[in] targetRegionIndex the target region index for the reservoir element
+ * @param[inout] esrMatched the subregion index of the reservoir element that contains "location", if any
+ * @param[inout] eiMatched the element index of the reservoir element that contains "location", if any
+ * @param[inout] giMatched the element global index of the reservoir element that contains "location", if any
+ */
+bool searchLocalElements( MeshLevel const & mesh,
```
CusiniM, line 319: is this a new function? How was this done previously?

src/coreComponents/mesh/WellElementSubRegion.cpp
```diff
@@ -770,59 +1023,66 @@ void WellElementSubRegion::connectPerforationsToMeshElements( MeshLevel & mesh,
...
 
-    localIndex erInit     = -1;
-    localIndex esrInit    = -1;
-    localIndex eiInit     = -1;
-    globalIndex giMatched  = -1;
+    localIndex const targetRegionIndex = elemManager.getRegions().getIndex( perfTargetTegionGlobal[iperfGlobal] );
+    if( targetRegionIndex >= 0 )
+    {
+      erStart = targetRegionIndex;
+      erEnd = erStart + 1;
```
CusiniM, line 842: why `+1 `? There could be more than 2 regions.

src/coreComponents/mesh/WellElementSubRegion.cpp
```diff
@@ -770,59 +1023,66 @@ void WellElementSubRegion::connectPerforationsToMeshElements( MeshLevel & mesh,
...
-    localIndex erMatched  = -1;
-    localIndex esrMatched = -1;
-    localIndex eiMatched  = -1;
+    localIndex erStart = -1, erEnd = -1;
 
-    localIndex erInit     = -1;
-    localIndex esrInit    = -1;
-    localIndex eiInit     = -1;
-    globalIndex giMatched  = -1;
+    localIndex const targetRegionIndex = elemManager.getRegions().getIndex( perfTargetTegionGlobal[iperfGlobal] );
```
CusiniM:
```suggestion
    localIndex const targetRegionIndex = elemManager.getRegions().getIndex( perfTargetRegionGlobal[iperfGlobal] );
```

src/coreComponents/mesh/generators/WellGeneratorBase.hpp
CusiniM: If it's only one region per perforation, I am not sure I understand what's happening in `assignUnownedElementsInReservoir`. There there seems to be a loop over multiple regions.

CusiniM (approved):
I don't love where some of these functions to connect perforations and cells are located. Seems like this should be part of a separate object in the mesh generation.

src/coreComponents/mesh/WellElementSubRegion.cpp
paveltomin: in https://github.com/GEOS-DEV/GEOS/pull/3359 i add a version of `isPointInsideElement` for `SurfaceElementSubRegion`

src/coreComponents/mesh/WellElementSubRegion.cpp
paveltomin: yes, new version, old one is just above, line 466

paveltomin: let me try merge them as much as possible

src/coreComponents/mesh/WellElementSubRegion.cpp
paveltomin: when `targetRegion` is used it assumes there is only one target region for that perforation

src/coreComponents/mesh/WellElementSubRegion.cpp
paveltomin: i removed old functions, hopefully didn't break anything, will check the tests soon

paveltomin:
```
@CusiniM 
If it's only one region per perforation, I am not sure I understand what's happening in assignUnownedElementsInReservoir. There there seems to be a loop over multiple regions.
```
that function is whatever it was, didn't touch it until today
now i changes it to use new versions of other functions, but logic should be unchanged

src/coreComponents/mesh/WellElementSubRegion.cpp
paveltomin: those are reservoir mesh nodes, i am not sure why it was done that way instead of collecting elements but there should be a good reason

paveltomin:
`INFO: No unfiltered differences were found.`

Only diffs like this:

```
Error: /Problem/Mesh/mesh/well_injector1/injector1_perf2
	Group has a child 'targetRegion' in the file to compare but not the baseline file.
```

Should be good to go

src/coreComponents/mesh/generators/WellGeneratorBase.hpp
paveltomin: it does a search through regions, once it finds an element, the loop is over

src/coreComponents/mesh/WellElementSubRegion.cpp
paveltomin: tests are ok and there is no function duplication anymore

PR #3230, opened 2024-07-18
refactor: Log level documentation on SolverBase and children

arng40:
This PR is the first iteration to document all `logLevel` in GEOS and to invite the contributors to set descriptions when they enable the `logLevel` wrapper:
In the first iteration, the work of this PR would be limited to `Solverbase `and its `children`.

Currently, we have this output in all `Group` documentation files:
```
========= ========= ======== ================
Name      Type      Default  Description                                                                             
========= ========= ======== ================
[...]
logLevel  integer   0        Log level     
========= ========= ======== ================
```

This PR allows to add more context to all `logLevels` of GEOS:

I.e. `SolidMechanics_LagrangianFEM.rst` :
```
========= ========= ======== ========================================================================================
Name      Type      Default  Description                                                                             
========= ========= ======== ========================================================================================
[...]
logLevel  integer   0        | Sets the level of information to write in the standard output (the console typically).
                             | A level of 0 outputs minimal information, higher levels require more.                 
                             | 1                                                                         
                             |  - Information on line search                                                         
                             |  - Information on global solution scaling factor                                      
                             |  - Information on the timestep                                                        
                             |  - Print residual norm                                                                                               
                             |  - Information about testing new configuration and print the time step                                             
                             |  - Information about line search failed                                                                                             
                             |  - Information on number of iterations and residual reduction                                                              
                             |  - Information on each newton Iteration                                                               
                             |  - Indicate allowed residual norm                                                     
                             | 2                                                                         
                             |  - Output to screen the assembled linear system and solutions (matrices and vectors)  
                             |  - Infos on residuals values                                                                                                        
                             |  - Warning about boundary conditions                                                  
                             | 3                                                                         
                             |  - Output to file the assembled linear system and solutions (matrices and vectors)    
========= ========= ======== ========================================================================================
```

In this PR, both of the mechanism will exists at the same times, the old one will be marked as `deprecated`. The upcoming PR will remove it and document all other groups.
```  
/// [DEPRECATED] Enable verbosity input for object
void enableLogLevelInput();
```


src/coreComponents/linearAlgebra/interfaces/hypre/HypreMGR.cpp
```diff
@@ -74,7 +74,7 @@ void hypre::mgr::createMGR( LinearSolverParameters const & params,
   {
     GEOS_LOG_RANK_0( GEOS_FMT( "        MGR preconditioner: numComponentsPerField = {}", numComponentsPerField ) );
   }
-  if( params.logLevel >= 1024 )
+  if( params.logLevel >= 1 )
```
victorapm:
```suggestion
  if( params.logLevel >= 1024 )
```
pointmarkers array is supposed to be printed out only for high enough values of logLevel.

src/coreComponents/mesh/DomainPartition.cpp
```diff
@@ -45,6 +45,8 @@ DomainPartition::DomainPartition( string const & name,
 
   registerGroup( groupKeys.meshBodies );
   registerGroup< constitutive::ConstitutiveManager >( groupKeys.constitutiveManager );
+
```
victorapm: Why adding two blank lines?

victorapm (changes requested):

src/coreComponents/linearAlgebra/interfaces/hypre/HypreMGR.cpp
CusiniM: we may want to have a dedicated hypre Logging logic.

victorapm: That might be a good idea, 1:1 with hypre's `print_level` to avoid confusions. On the other hand, we would need yet another variable in the code...

src/coreComponents/fileIO/Outputs/VTKOutput.cpp
```diff
@@ -39,6 +39,7 @@ VTKOutput::VTKOutput( string const & name,
   m_levelNames(),
   m_writer( getOutputDirectory() + '/' + m_plotFileRoot )
 {
+  //deprecated
```
MelReyCG: Please remove the `deprecated` mentions from the implementations. By removing them and the useless space, you can affect much less file with this PR (which is the point).

src/coreComponents/common/Logger.hpp
```diff
@@ -432,20 +432,23 @@
     }                                                                          \
   } while( false )
 
+//[DEPRECATED]
```
MelReyCG: add this kind of tags in the doxygen documentation
```suggestion
 * @deprecated will be replaced by ...
```

src/coreComponents/constitutive/relativePermeability/RelpermDriver.cpp
```diff
@@ -52,6 +50,9 @@ RelpermDriver::RelpermDriver( const geos::string & name,
     setInputFlag( InputFlags::OPTIONAL ).
     setApplyDefaultValue( "none" ).
     setDescription( "Baseline file" );
+
+
+  addLogLevel< logInfo::LineSearch >();
```
MelReyCG: To keep in the scope of this PR, please revert all modifications that are not on SolverBase children.

src/coreComponents/fieldSpecification/SourceFluxBoundaryCondition.cpp
```diff
@@ -31,17 +31,10 @@ SourceFluxBoundaryCondition::SourceFluxBoundaryCondition( string const & name, G
...
-                              FieldSpecificationBase::viewKeyStruct::scaleString(),
-                              CompositionalMultiphaseBase::viewKeyStruct::useMassFlagString() ) );
-
-  getWrapper< real64 >( FieldSpecificationBase::viewKeyStruct::scaleString() ).
-    setDescription( GEOS_FMT( "Multiplier of the {0} value. If no {0} is provided, this value is used directly.",
-                              FieldSpecificationBase::viewKeyStruct::functionNameString() ) );
+
+
+
+
```
MelReyCG: Why did you removed the documentation of `functionName` & `scaleString`?

src/coreComponents/mesh/DomainPartition.cpp
arng40: it was an omission

src/coreComponents/linearAlgebra/interfaces/hypre/HypreMGR.cpp
MelReyCG: @arng40 Please revert to the original value as changing loglevel is not the scope of this PR.

src/coreComponents/physicsSolvers/fluidFlow/ReactiveCompositionalMultiphaseOBL.cpp
```diff
@@ -116,6 +116,9 @@ ReactiveCompositionalMultiphaseOBL::ReactiveCompositionalMultiphaseOBL( const st
     setInputFlag( InputFlags::OPTIONAL ).
     setDescription( "List of fluid phases" );
 
+
+  //addLogLevel( "logLevel >= 1 first nonlinear iteration", "Information on the source flux boundary conditions" );
```
MelReyCG: Is there missing logLevel docs for this class?

src/coreComponents/physicsSolvers/fluidFlow/FlowSolverBase.cpp
```diff
@@ -122,6 +122,8 @@ FlowSolverBase::FlowSolverBase( string const & name,
     setApplyDefaultValue( 0.1 ).
     setDescription( "Maximum (absolute) temperature change in a sequential iteration, used for outer loop convergence check" );
 
+
```
MelReyCG: You may have forgotten this one too.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseHybridFVM.cpp
```diff
@@ -49,6 +49,8 @@ CompositionalMultiphaseHybridFVM::CompositionalMultiphaseHybridFVM( const std::s
   CompositionalMultiphaseBase( name, parent ),
   m_lengthTolerance( 0 )
 {
+
```
MelReyCG: Forgotten?

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.cpp
```diff
@@ -434,6 +443,7 @@ real64 CompositionalMultiphaseFVM::calculateResidualNorm( real64 const & GEOS_UN
     }
     residualNorm = sqrt( globalResidualNorm[0] * globalResidualNorm[0] + globalResidualNorm[1] * globalResidualNorm[1] );
 
+
     if( getLogLevel() >= 1 && logger::internal::rank == 0 )
```
MelReyCG: Why not using the LogLevelInfo here?

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -170,6 +170,13 @@ CompositionalMultiphaseBase::CompositionalMultiphaseBase( const string & name,
     setInputFlag( InputFlags::OPTIONAL ).
     setApplyDefaultValue( 0.01 ).
     setDescription( "Minimum value for solution scaling factor" );
+
+
+
+
```
MelReyCG: forgotten?

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseHybridFVM.cpp
```diff
@@ -47,6 +47,7 @@ SinglePhaseHybridFVM::SinglePhaseHybridFVM( const string & name,
   m_areaRelTol( 1e-8 )
 {
```
MelReyCG: forgotten?

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseFVM.cpp
```diff
@@ -59,7 +59,9 @@ template< typename BASE >
 SinglePhaseFVM< BASE >::SinglePhaseFVM( const string & name,
                                         Group * const parent ):
   BASE( name, parent )
-{}
+{
+  SinglePhaseBase::addLogLevel< logInfo::ResidualValues >();
```
MelReyCG: Why not using the `logInfo::ResidualValues` in this class?

src/coreComponents/physicsSolvers/multiphysics/CompositionalMultiphaseReservoirAndWells.cpp
```diff
@@ -42,7 +43,9 @@ CompositionalMultiphaseReservoirAndWells< RESERVOIR_SOLVER >::
 CompositionalMultiphaseReservoirAndWells( const string & name,
                                           Group * const parent )
   : Base( name, parent )
-{}
+{
+  Base::template addLogLevel< logInfo::CrossflowWarning >();
```
MelReyCG: Not used?

src/coreComponents/physicsSolvers/multiphysics/CompositionalMultiphaseReservoirAndWells.hpp
```diff
@@ -38,6 +38,8 @@ class CompositionalMultiphaseReservoirAndWells : public CoupledReservoirAndWells
   using Base::getLogLevel;
   using Base::m_solvers;
   using Base::m_linearSolverParameters;
+  // template< typename LOG_LEVEL_INFO  > 
+  // using Base::addLogLevel<LOG_LEVEL_INFO>;
```
MelReyCG: please remove if not useful

MelReyCG (changes requested):

src/coreComponents/physicsSolvers/multiphysics/CompositionalMultiphaseReservoirAndWells.cpp
arng40: now added

src/coreComponents/common/LogLevelsInfo.hpp
```diff
@@ -0,0 +1,194 @@
...
+/**
+ * @brief Trait used to check whether a LOG_LEVEL_INFO structure is valid.
+ * @tparam LOG_LEVEL_INFO The log level structure to check.
+ */
+template< typename LOG_LEVEL_INFO >
+static constexpr bool is_log_level_info =
+  std::is_same_v< int, decltype(LOG_LEVEL_INFO::getMinLogLevel()) > &&
+  std::is_same_v< std::string_view, decltype(LOG_LEVEL_INFO::getDescription()) >;
+
+/** This 3 method would replace the ones in Logger.hpp  */
```
MelReyCG:
```suggestion
/** Those 3 macros will replace the ones in Logger.hpp  */
```

src/coreComponents/common/LogLevelsInfo.hpp
```diff
@@ -0,0 +1,194 @@
...
+/**
+ * @brief Output messages based on current Group's log level.
+ * @param[in] logInfo Strut containing log level desscription
+ * @param[in] msg a message to log (any expression that can be stream inserted)
+ */
+#define GEOS_LOG_INFO_LEVEL( logInfo, msg ) GEOS_INFO_IF( isLogLevelActive< logInfo >( this->getLogLevel() ), msg );
```
MelReyCG: Those macros are nice, but why do you have both `INFO_LEVEL` & `LEVEL_INFO` naming?

src/coreComponents/physicsSolvers/SolverBase.cpp
```diff
@@ -320,36 +329,36 @@ real64 SolverBase::setNextDt( real64 const & currentDt,
   {
     if( nextDtNewton > currentDt )
     {
-      GEOS_LOG_LEVEL_RANK_0( 1, GEOS_FMT( "{}: time-step required will be increased based on number of iterations.",
-                                          getName() ) );
+      GEOS_LOG_LEVEL_INFO_RANK_0( logInfo::TimeStep, GEOS_FMT( "{}: time-step required will be increased based on number of iterations.",
```
MelReyCG, line 388: Did you forget the messages line 323 & 326 or is it on purpose?
If it is a mistake, please verify that there is no other ones in your targeted classes.

src/coreComponents/common/LogLevelsInfo.hpp
```diff
@@ -0,0 +1,194 @@
...
+
+/**
+ * @brief Output messages (with one line per rank) based on current Group's log level.
+ * @param[in] logInfo Strut containing log level desscription
+ * @param[in] msg a message to log (any expression that can be stream inserted)
+ */
+#define GEOS_LOG_LEVEL_INFO_BY_RANK( logInfo, msg ) GEOS_INFO_IF( isLogLevelActive< logInfo >( this->getLogLevel() ), msg );
+
+
+/// @cond DO_NOT_DOCUMENT
```
MelReyCG: You can wrap your structs in only one `@cond DO_NOT_DOCUMENT` :D

src/coreComponents/physicsSolvers/SolverBase.cpp
```diff
@@ -812,7 +821,7 @@ real64 SolverBase::nonlinearImplicitStep( real64 const & time_n,
     {
       // cut timestep, go back to beginning of step and restart the Newton loop
       stepDt *= dtCutFactor;
-      GEOS_LOG_LEVEL_RANK_0 ( 1, GEOS_FMT( "New dt = {}", stepDt ) );
+      GEOS_LOG_LEVEL_INFO_RANK_0 ( logInfo::SolverTimeStep, GEOS_FMT( "New dt = {}", stepDt ) );
```
MelReyCG: The log level doesn't match the old one, why?

src/coreComponents/physicsSolvers/SolverBase.cpp
```diff
@@ -922,9 +932,9 @@ bool SolverBase::solveNonlinearSystem( real64 const & time_n,
     if( residualNorm > m_nonlinearSolverParameters.m_maxAllowedResidualNorm )
     {
       string const maxAllowedResidualNormString = NonlinearSolverParameters::viewKeysStruct::maxAllowedResidualNormString();
-      GEOS_LOG_LEVEL_RANK_0( 1, GEOS_FMT( "    The residual norm is above the {} of {}. Newton loop terminated.",
-                                          maxAllowedResidualNormString,
-                                          m_nonlinearSolverParameters.m_maxAllowedResidualNorm ) );
+      GEOS_LOG_LEVEL_INFO_RANK_0( logInfo::SolverTimeStep, GEOS_FMT( "    The residual norm is above the {} of {}. Newton loop terminated.",
```
MelReyCG: Doesn't match

src/coreComponents/physicsSolvers/SolverBase.cpp
```diff
@@ -1179,17 +1186,17 @@ void SolverBase::debugOutputSystem( real64 const & time,
                        nonlinearIteration,
                        getName() + "_mat",
                        "System matrix",
-                       m_writeLinearSystem == 1,
-                       m_writeLinearSystem >= 2 );
+                       isLogLevelActive< logInfo::ScreenLinearSystem >( getLogLevel() ),
+                       isLogLevelActive< logInfo::FileLinearSystem >( getLogLevel() ) );
```
MelReyCG: Doesn't match the old ones too.

src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp
```diff
@@ -920,8 +924,8 @@ real64 HydrofractureSolver< POROMECHANICS_SOLVER >::setNextDt( real64 const & cu
   {
     nextDt = m_surfaceGenerator->getTimestepRequest() < 1e99 ? m_surfaceGenerator->getTimestepRequest() : currentDt;
   }
+  GEOS_LOG_LEVEL_INFO_RANK_0( logInfo::SolverTimeStep, this->getName() << ": nextDt request is "  << nextDt );
```
MelReyCG: I think you might create a `logInfo::NextDT` (lvl=3, desc about the changing DT) that target only this kind of call, and the other one would fallback to `logInfo::TimeStep` (lvl=1)

src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.hpp
```diff
@@ -23,6 +23,7 @@
 #include "physicsSolvers/surfaceGeneration/SurfaceGenerator.hpp"
 #include "physicsSolvers/multiphysics/SinglePhasePoromechanics.hpp"
 #include "physicsSolvers/fluidFlow/SinglePhaseBase.hpp"
+#include "common/LogLevelsInfo.hpp"
```
MelReyCG: I thing the `LogLevelInfo.hpp` include should either go in all `.cpp`s, or in `SolverBase.hpp`. I would prefer that in the `.cpp`s for allowing targeted `LogLevelInfo.hpp` if that get sense later.

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoromechanicsConformingFractures.cpp
```diff
@@ -43,6 +44,8 @@ SinglePhasePoromechanicsConformingFractures< FLOW_SOLVER >::SinglePhasePoromecha
                                                                                                          Group * const parent )
   : Base( name, parent )
 {
+  Base::template addLogLevel< logInfo::Dof >();
```
MelReyCG: You might have forgotten to use this one.

MelReyCG (changes requested):

src/coreComponents/physicsSolvers/SolverBase.cpp
arng40: It would be part of a future PR, it concerning nonlinearSolverParameter

src/coreComponents/physicsSolvers/SolverBase.cpp
arng40: fixed

src/coreComponents/physicsSolvers/SolverBase.cpp
arng40: fixed too

src/coreComponents/common/LogLevelsInfo.hpp
```diff
@@ -0,0 +1,170 @@
...
+#define GEOS_LOG_LEVEL_INFO_RANK_0( logInfo, msg ) GEOS_INFO_IF( isLogLevelActive< logInfo >( this->getLogLevel() ), msg );
+
+/**
+ * @brief Output messages (with one line per rank) based on current Group's log level.
+ * @param[in] logInfo Strut containing log level desscription
+ * @param[in] msg a message to log (any expression that can be stream inserted)
+ */
+#define GEOS_LOG_LEVEL_INFO_BY_RANK( logInfo, msg ) GEOS_INFO_IF( isLogLevelActive< logInfo >( this->getLogLevel() ), msg );
+
+
```
MelReyCG, line 60: I suggest to add a region like:

```
/**
 * @name Common LogLevels info structures. They must comply with the `is_log_level_info` trait.
 */
///@{

///@}
```

src/coreComponents/common/LogLevelsInfo.hpp
```diff
@@ -0,0 +1,171 @@
...
+static constexpr bool is_log_level_info =
+  std::is_same_v< int, decltype(LOG_LEVEL_INFO::getMinLogLevel()) > &&
+  std::is_same_v< std::string_view, decltype(LOG_LEVEL_INFO::getDescription()) >;
+
+/** ThOSE 3 macros would replace the ones in Logger.hpp  */
+/**
+ * @brief Output messages based on current Group's log level.
+ * @param[in] logInfo Strut containing log level desscription
+ * @param[in] msg a message to log (any expression that can be stream inserted)
+ */
```
MelReyCG: Those macro should not be far from `isLogLevelActive()`. Should `isLogLevelActive()` be moved here?

src/coreComponents/common/LogLevelsInfo.hpp
```diff
@@ -0,0 +1,171 @@
...
+ * @brief Output messages (with one line per rank) based on current Group's log level.
+ * @param[in] logInfo Strut containing log level desscription
+ * @param[in] msg a message to log (any expression that can be stream inserted)
+ */
+#define GEOS_LOG_LEVEL_INFO_BY_RANK( logInfo, msg ) GEOS_INFO_IF( isLogLevelActive< logInfo >( this->getLogLevel() ), msg );
+
+
+/// @cond DO_NOT_DOCUMENT
+struct LineSearch
+{
```
MelReyCG: I suggest to add a region like:

```
/**
 * @name Common LogLevels info structures. They must comply with the `is_log_level_info` trait.
 */
///@{

...

///@}
```

src/coreComponents/common/LogLevelsInfo.hpp
```diff
@@ -0,0 +1,171 @@
...
+ * Copyright (c) 2018-2020 TotalEnergies
+ * Copyright (c) 2019-     GEOSX Contributors
+ * All rights reserved
+ *
+ * See top level LICENSE, COPYRIGHT, CONTRIBUTORS, NOTICE, and ACKNOWLEDGEMENTS files for details.
+ * ------------------------------------------------------------------------------------------------------------
+ */
+
+/**
+ * @file LogLevelsInfo.hpp
```
MelReyCG: Please describe the responsability of this file here (i.e. as it is done in DataTypes.hpp).
I would expect this file to be a repository of all common `LogLevelInfo.hpp` + to expose the way to use them.

Also, should this file be located in `dataRepository`, or is there futures potencial uses in `common` or `codingUtilities`?

src/coreComponents/common/LogLevelsInfo.hpp
```diff
@@ -0,0 +1,171 @@
...
+ * @tparam LOG_LEVEL_INFO The log level structure to check.
+ */
+template< typename LOG_LEVEL_INFO >
+static constexpr bool is_log_level_info =
+  std::is_same_v< int, decltype(LOG_LEVEL_INFO::getMinLogLevel()) > &&
+  std::is_same_v< std::string_view, decltype(LOG_LEVEL_INFO::getDescription()) >;
+
+/** ThOSE 3 macros would replace the ones in Logger.hpp  */
+/**
+ * @brief Output messages based on current Group's log level.
```
MelReyCG: I would add a region to wrap the ways to use the LogLevelsInfo (macros, potencially `isLogLevelActive()`).

MelReyCG (changes requested):

src/coreComponents/common/LogLevelsInfo.hpp
arng40: I think `isLogLevelActive`() should be moved in `LogLevelsInfo `and the file `LogLevelsRegistry ` only use is to keep track of log level documention for a group
or maybe create another file ?

src/coreComponents/physicsSolvers/SolverBase.cpp
MelReyCG: Alright, it won't be hard to find

src/coreComponents/common/LogLevelsInfo.hpp
arng40: I think `LogLevelInfo `should be moved in `dataRepository `otherwise it all add a depandance between `common `and `dataRepo `
Because file in dataRpository, for me, is a little unclear, in the future add some subPackages to keep things clear like in common

src/coreComponents/physicsSolvers/SolverBase.cpp
MelReyCG: It still doesn't match the previous behaviour. How is it related to the `logLevel`, I may have missed something?

src/coreComponents/common/LogLevelsInfo.hpp
MelReyCG: It is indeed an opened discussion. We will work on clarifying thoses packages to help their maintainability + we may have to split some if this is relevant.

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoromechanicsConformingFractures.cpp
MelReyCG: Still not adressed.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.cpp
```diff
@@ -435,11 +435,8 @@ real64 CompositionalMultiphaseFVM::calculateResidualNorm( real64 const & GEOS_UN
     }
     residualNorm = sqrt( globalResidualNorm[0] * globalResidualNorm[0] + globalResidualNorm[1] * globalResidualNorm[1] );
 
-    if( getLogLevel() >= 1 && logger::internal::rank == 0 )
-    {
-      std::cout << GEOS_FMT( "        ( R{} ) = ( {:4.2e} )        ( Renergy ) = ( {:4.2e} )",
-                             coupledSolverAttributePrefix(), globalResidualNorm[0], globalResidualNorm[1] );
-    }
+    GEOS_LOG_LEVEL_INFO( logInfo::ResidualNorm, GEOS_FMT( "        ( R{} ) = ( {:4.2e} )        ( Renergy ) = ( {:4.2e} )",
```
MelReyCG: Be careful to not change the amount of ranks that logs messages. Here, there was only the rank 0, now all ranks will spam this message.

src/coreComponents/physicsSolvers/fluidFlow/ReactiveCompositionalMultiphaseOBL.cpp
MelReyCG: previous remark is still not adressed if `ReactiveCompositionalMultiphaseOBL` is in the scope of this PR.

src/coreComponents/dataRepository/LogLevelsRegistry.hpp
```diff
@@ -0,0 +1,72 @@
...
+ * @brief Verify if a log level is active
+ * @tparam LOG_LEVEL_INFO The structure containing log level information.
+ * @param level Log level to be checked.
+ * @return `true` if the log level is active, `false` otherwise.
+ * @pre `LOG_LEVEL_INFO` must satisfy `logInfo::is_log_level_info`.
+ *
+ */
+template< typename LOG_LEVEL_INFO >
+std::enable_if_t< logInfo::is_log_level_info< LOG_LEVEL_INFO >, bool >
+isLogLevelActive( int level )
```
MelReyCG: Instead of directly giving the log level as parameter, would it be possible to have a templated type parameter `LOGGING_OBJECT const & loggingObject` for which we call a `loggingObject.getLogLevel()` method?
This would allow to ask if it has registered the required `LogLevelInfo`, and `GEOS_ERROR()` if it's not the case.

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseFVM.cpp
MelReyCG: Why did you remove the `SinglePhaseBase::addLogLevel< logInfo::ResidualValues >();` instead of using it for this file logs?

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseHybridFVM.cpp
MelReyCG: I think you should check another time all the `SolverBase` children logging calls. I think there are some missed calls.

MelReyCG (changes requested):
I will need a final verification pass for log levels conservation.

src/coreComponents/dataRepository/LogLevelsRegistry.hpp
arng40: - 1st  point : ReactiveCompositionalMultiphaseOBL : i didn't change this file
2st point : TODO

MelReyCG: As an exemple, I would expect a crash in this scenario:
```c++
template< typename BASE >
SinglePhaseFVM< BASE >::SinglePhaseFVM( const string & name,
                                        Group * const parent ):
  BASE( name, parent )
{
  addLogLevelInfo< logInfo::SolverConfig >();
}

template< typename BASE >
void SinglePhaseFVM< BASE >::initializePreSubGroups()
{
  BASE::initializePreSubGroups();

  //[...]

  GEOS_LOG_LEVEL_INFO( logInfo::TimeStep, "very useful TimeStep info" ); // logInfo::TimeStep not registered = crash
}

```

another solution :
- require a `LogLevelRegistry const &` parameter in `isLogLevelActive()`
- modify the macros: `#define GEOS_LOG_LEVEL_INFO( logInfo, msg ) GEOS_INFO_IF( isLogLevelActive< logInfo >( this->getLogLevel(), this->getLogRegistry() ), msg );`
- reverse the include way for `LogLevelsRegistry.hpp` & `LogLevelsInfo.hpp`

also, please verify your headers, `Format.hpp` may be for cpps only.

MelReyCG: In the commit `53e8dff5a66ac4c2786a4e708e150ca30dea4a24`, you made the `LogLevelInfo` registering verification by checking the description. It is hazardous and not optimal. Basing the verification on the `typeid` would be way better. I think we should delay that verification in its own PR.

arng40:
https://github.com/GEOS-DEV/GEOS/pull/3230#discussion_r1725065505 
- For now, the log level refactor is not applied to all children, if the new structure is approved I will extend to all children

src/coreComponents/dataRepository/Group.hpp
```diff
@@ -1712,6 +1730,22 @@ Wrapper< T > & Group::registerWrapper( string const & name,
...
 }
 
+template< typename LOG_LEVEL_INFO >
+std::enable_if_t< logInfo::is_log_level_info< LOG_LEVEL_INFO >, void >
+Group::addLogLevel()
+{
+  Wrapper< integer > * wrapper = getWrapperPointer< integer >( viewKeyStruct::logLevelString() );
+  if( wrapper == nullptr )
+  {
+    wrapper = &registerWrapper( viewKeyStruct::logLevelString(), &m_logLevel );
```
dkachuma, line 1743: I suppose this implicitly adds a `logLevel` wrapper to the `Group`. Would there be a danger of calling this outside a `Group` constructor and changing the schema?

src/coreComponents/dataRepository/LogLevelsRegistry.hpp
```diff
@@ -0,0 +1,59 @@
...
+   * @param level The minimum log level
+   * @param description The description for the log level
+   */
+  void addEntry( integer level, std::string_view description );
+
+  /**
+   * @brief Construct the log level string description for a wrapper
+   * @return The log level string description
+   */
+  string buildLogLevelDescription();
```
dkachuma:
```suggestion
  string buildLogLevelDescription() const;
```

dkachuma (approved):

src/coreComponents/dataRepository/Group.hpp
arng40: Good point, all `m_logLevelsRegistry ` is set  to nullptr  in `postInputInitializationRecursive `, afer the schema generation,  so  I've added a `GEOS_ERROR_IF` in `addLogLevel` to check if `m_logLevelsRegistry `is set to `nullptr`

MelReyCG: @dkachuma 
It indeed adds the `logLevel` wrapper, as `enableLogLevel()` did before this PR.
Calling it during or after the `postInputInitialization` phase will generate an error as the `LogLevelRegistry` are released (`Group.cpp:248`).

Your point applies to all `Wrapper` descriptions, we may generalize this solution for all of them later.
Maybe the descriptions should only be generated when `ProblemManager::generateDocumentation()` would be called?

CusiniM:
I am ready to review this one but can you resolve the conflicts coz now it's hard to review some of the files.

arng40:
> I am ready to review this one but can you resolve the conflicts coz now it's hard to review some of the files.

All good now

src/coreComponents/dataRepository/Group.hpp
```diff
@@ -1712,6 +1731,24 @@ Wrapper< T > & Group::registerWrapper( string const & name,
   return rval;
 }
 
+template< typename LOG_LEVEL_INFO >
+std::enable_if_t< geos::is_log_level_info< LOG_LEVEL_INFO >, void >
+Group::addLogLevel()
+{
+  GEOS_ERROR_IF( m_logLevelsRegistry == nullptr, "You cannot call addLogLevel after schema generation" );
+
+  Wrapper< integer > * wrapper = getWrapperPointer< integer >( viewKeyStruct::logLevelString() );
```
rrsettgast, line 1740:
```suggestion
  WrapperBase * wrapperBase = getWrapperPointer< integer >( viewKeyStruct::logLevelString() );
```

there was a compiler error I ran into about a failed cast in a situation like this. The typed `Wrapper` was not required, so we can just use a `WrapperBase`. We could add a `getWrapperBasePointer`

rrsettgast (approved):
looks ok to me. There are a couple of unresolved comments to address before merge.

rrsettgast:
@arng40 The rst files are no longer stored in the repository. You should remove them from this PR.

src/coreComponents/dataRepository/Group.hpp
arng40: In your 1st remark, i should write smth like this ? : 
```
WrapperBase * wrapper = getWrapperPointer< integer >( viewKeyStruct::logLevelString() );
 Wrapper< integer > & wrapperT = Wrapper< integer >::cast( *wrapper );
```

arng40: I don't understand, with your example how can I use wrapper method (i.e setApplyDefaultValue), I will have to cast to wrapper<integer>, won't I ? or I miss smth ?
Wouldn't we need a compilation error here if the expected type is wrong here?

rrsettgast: no. I didn't see you were calling `setApplyDefaultValue()` for some reason. disregard.

src/coreComponents/physicsSolvers/SolverBase.cpp
```diff
@@ -606,14 +615,14 @@ bool SolverBase::lineSearch( real64 const & time_n,
     applyBoundaryConditions( time_n, dt, domain, dofManager, localMatrix, localRhs );
     rhs.close();
 
-    if( getLogLevel() >= 1 && logger::internal::rank==0 )
+    if( logger::internal::rank==0 )
```
paveltomin: this is redundant, please check for other places

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.cpp
```diff
@@ -435,11 +435,8 @@ real64 CompositionalMultiphaseFVM::calculateResidualNorm( real64 const & GEOS_UN
     }
     residualNorm = sqrt( globalResidualNorm[0] * globalResidualNorm[0] + globalResidualNorm[1] * globalResidualNorm[1] );
 
-    if( getLogLevel() >= 1 && logger::internal::rank == 0 )
-    {
-      std::cout << GEOS_FMT( "        ( R{} ) = ( {:4.2e} )        ( Renergy ) = ( {:4.2e} )",
-                             coupledSolverAttributePrefix(), globalResidualNorm[0], globalResidualNorm[1] );
-    }
+    GEOS_LOG_LEVEL_INFO_RANK_0( logInfo::ResidualNorm, GEOS_FMT( "        ( R{} ) = ( {:4.2e} )        ( Renergy ) = ( {:4.2e} )",
```
paveltomin: `cout` was used here for a reason: all those GEOS_LOG macros add `endl` at the end, but residuals from solvers are all printed in a line, otherwise it will look very long and ugly
please revert or find a way to preserve the current form for residual output

src/coreComponents/dataRepository/LogLevelsInfo.hpp
```diff
@@ -0,0 +1,193 @@
...
+struct LineSearch
+{
+  static constexpr int getMinLogLevel() { return 1; }
+  static constexpr std::string_view getDescription() { return "Information on line search"; }
+};
+
+struct LineSearchFailed
+{
+  static constexpr int getMinLogLevel() { return 1; }
+  static constexpr std::string_view getDescription() { return "On Incorrect solution, Information on failed line search"; }
```
paveltomin: this description does not make much sense

src/coreComponents/dataRepository/LogLevelsInfo.hpp
```diff
@@ -0,0 +1,193 @@
...
+};
+
+struct ScalingFactor
+{
+  static constexpr int getMinLogLevel() { return 1; }
+  static constexpr std::string_view getDescription() { return "Information on global solution scaling factor"; }
```
paveltomin: how about local? different struct?

src/coreComponents/dataRepository/LogLevelsInfo.hpp
```diff
@@ -0,0 +1,193 @@
...
+struct FileLinearSystem
+{
+  static constexpr int getMinLogLevel() { return 2; }
+  static constexpr std::string_view getDescription() { return "Output to file the assembled linear system and solutions (matrices and vectors)"; }
+};
+
+struct SolverConfig
+{
+  static constexpr int getMinLogLevel() { return 1; }
+  static constexpr std::string_view getDescription() { return "On non convergance, Information about testing new configuration and print the time step"; }
```
paveltomin: confusing

src/coreComponents/dataRepository/LogLevelsInfo.hpp
```diff
@@ -0,0 +1,193 @@
...
+struct ResidualValues
+{
+  static constexpr int getMinLogLevel() { return 1; }
+  static constexpr std::string_view getDescription() { return "Print the residual values"; }
+};
+
+struct LinearSystem
+{
+  static constexpr int getMinLogLevel() { return 1; }
+  static constexpr std::string_view getDescription() { return "Information oon linear system"; }
```
paveltomin: typo

src/coreComponents/dataRepository/LogLevelsInfo.hpp
```diff
@@ -0,0 +1,193 @@
...
+};
+
+struct CrossflowWarning
+{
+  static constexpr int getMinLogLevel() { return 1; }
+  static constexpr std::string_view getDescription() { return "If the well is injector and crossflow enabled, display informations about crossflow for injectors"; }
```
paveltomin: I don't understand the choice of what is being defined as struct in this file and what is not - there are more general categories like line search or linear solver, but then this crossflow warning - very specific, why?

src/coreComponents/dataRepository/LogLevelsInfo.hpp
```diff
@@ -0,0 +1,193 @@
...
+struct Dof
+{
+  static constexpr int getMinLogLevel() { return 2; }
+  static constexpr std::string_view getDescription() { return "The summary of declared fields and coupling"; }
+};
+
+struct PoromechanicsPhaseFraction
+{
+  static constexpr int getMinLogLevel() { return 1; }
+  static constexpr std::string_view getDescription() { return "Print phase volume fraction"; }
```
paveltomin: same here - phase volume fraction but why it is called PoromechanicsPhaseFraction???

paveltomin (changes requested):
please address my comments

src/coreComponents/dataRepository/LogLevelsInfo.hpp
arng40: For the moment the log doc is only for `SolverBase ` and it's children so we define all struct at the same place, `LogLelvelsInfo.hpp`, and then following the amount of struct, for each folder, put  the corresponding log structure

src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp
```diff
@@ -70,6 +71,11 @@ HydrofractureSolver< POROMECHANICS_SOLVER >::HydrofractureSolver( const string &
     setApplyDefaultValue( 0 ).
     setInputFlag( InputFlags::OPTIONAL );
 
+  Base::template addLogLevel< logInfo::CrossflowWarning >();
```
paveltomin: why CrossflowWarning is here?

src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp
```diff
@@ -70,6 +71,11 @@ HydrofractureSolver< POROMECHANICS_SOLVER >::HydrofractureSolver( const string &
...
+  Base::template addLogLevel< logInfo::HydraulicAperture >();
+  Base::template addLogLevel< logInfo::SolverTimeStep >();
+  Base::template addLogLevel< logInfo::SolverNextDt >();
```
paveltomin: SolverTimeStep and SolverNextDt should be already covered in SolverBase, no?

src/coreComponents/dataRepository/LogLevelsInfo.hpp
arng40: The aim is to have an overview of what loglevels bring us, which is why I've remained vague about LineSearch, but the description can be rephrase.
I did a first pass on the solverBase descriptions but I agree, some of them could be better written.

src/coreComponents/dataRepository/LogLevelsInfo.hpp
arng40: It's related to Poromechanics but i could rephrase with  "Print Poromechanics phase volume fraction"

src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp
arng40: not required, removed

src/coreComponents/dataRepository/LogLevelsInfo.hpp
arng40: description updated, first of all I want to keep it generic

src/coreComponents/dataRepository/LogLevelsInfo.hpp
arng40: updated

src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp
arng40: I removed SolverTimeStep and I left tSolverNextDt because it's a different logLovel

src/coreComponents/dataRepository/LogLevelsInfo.hpp
paveltomin: then i suggest to group everything related to linesearch under the same struct
same for others, try to group as much as possible

paveltomin:
I tried to address my concerns here https://github.com/GEOS-DEV/GEOS/pull/3378
Please let me know what you think

src/coreComponents/common/logger/Logger.hpp
```diff
@@ -456,20 +456,23 @@
  * @brief Output messages based on current Group's log level.
  * @param[in] minLevel minimum log level
  * @param[in] msg a message to log (any expression that can be stream inserted)
+ * @deprecated Will be replaced by GEOS_LOG_LEVEL_INFO
```
CusiniM, line 474: shall we create an issue to remind ourselves to remove these at some point?

src/coreComponents/dataRepository/Group.hpp
CusiniM: @MelReyCG, yes I think that would be better. A compile time check would be preferable IMO but it's not obvious to me how we could do it.

src/coreComponents/dataRepository/Group.hpp
```diff
@@ -863,6 +865,16 @@ class Group
...
 
+  /**
+   * @brief Append a levelCondition and a log description to the description of the wrapped object given a log info struct.
+   * Must be called in constructor.
+   * @tparam LOG_LEVEL_INFO The log documentation to add.
+   * @return void if the trait is verified.
+   */
+  template< typename LOG_LEVEL_INFO >
+  std::enable_if_t< geos::is_log_level_info< LOG_LEVEL_INFO >, void >
+  addLogLevel();
```
CusiniM, line 876: does it actually add a `LogLevel` or is it rather adding to a `LogLevel`? We can discuss this further and leave it for a follow up PR but I wonder if the possible levels should not be fixed somewhere and be the same for all groups and let this function add logging to one of the predefined options. Although I do know that some times unusual log level numbers are added for Hypre.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.cpp
CusiniM: @arng40 maybe you can add a different macro that does not break the line?

src/coreComponents/physicsSolvers/fluidFlow/ReactiveCompositionalMultiphaseOBL.cpp
```diff
@@ -367,10 +367,7 @@ real64 ReactiveCompositionalMultiphaseOBL::calculateResidualNorm( real64 const &
 
   real64 const residual = m_useDARTSL2Norm ? MpiWrapper::max( localResidualNorm ) : std::sqrt( MpiWrapper::sum( localResidualNorm ) );
 
-  if( getLogLevel() >= 1 && logger::internal::rank==0 )
-  {
-    std::cout << GEOS_FMT( "        ( Rflow ) = ( {:4.2e} )", residual );
-  }
+  GEOS_LOG_LEVEL_INFO_RANK_0( logInfo::ResidualNorm, GEOS_FMT( "        ( Rflow ) = ( {:4.2e} )", residual ) );
```
CusiniM: @paveltomin also applies here.

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseFVM.cpp
```diff
@@ -228,11 +228,8 @@ real64 SinglePhaseFVM< BASE >::calculateResidualNorm( real64 const & GEOS_UNUSED
     }
     residualNorm = sqrt( globalResidualNorm[0] * globalResidualNorm[0] + globalResidualNorm[1] * globalResidualNorm[1] );
 
-    if( getLogLevel() >= 1 && logger::internal::rank == 0 )
-    {
-      std::cout << GEOS_FMT( "        ( R{} ) = ( {:4.2e} )        ( Renergy ) = ( {:4.2e} )",
-                             FlowSolverBase::coupledSolverAttributePrefix(), globalResidualNorm[0], globalResidualNorm[1] );
-    }
+    GEOS_LOG_LEVEL_INFO_RANK_0( logInfo::ResidualNorm, GEOS_FMT( "        ( R{} ) = ( {:4.2e} )        ( Renergy ) = ( {:4.2e} )",
```
CusiniM: and here

src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp
CusiniM: why is it different? I am not sure that's intended.

CusiniM (commented):
This looks good but I think that the comment about not breaking the line after outputting residual norms must be addressed before merging.

arng40:
@CusiniM I just updated my branch with pavel's pr, it should looks good now

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseFVM.cpp
```diff
@@ -246,10 +243,7 @@ real64 SinglePhaseFVM< BASE >::calculateResidualNorm( real64 const & GEOS_UNUSED
       solverBaseKernels::L2ResidualNormHelper::computeGlobalNorm( localResidualNorm[0], localResidualNormalizer[0], residualNorm );
     }
 
-    if( getLogLevel() >= 1 && logger::internal::rank == 0 )
-    {
-      std::cout << GEOS_FMT( "        ( R{} ) = ( {:4.2e} )", FlowSolverBase::coupledSolverAttributePrefix(), residualNorm );
-    }
+    GEOS_LOG_LEVEL_INFO_RANK_0( logInfo::Convergence, GEOS_FMT( "        ( R{} ) = ( {:4.2e} )", FlowSolverBase::coupledSolverAttributePrefix(), residualNorm ));
```
CusiniM: I think this is still an issue @arng40 . @paveltomin ?

src/coreComponents/common/logger/Logger.hpp
MelReyCG: [Done !](https://github.com/GEOS-DEV/GEOS/issues/3385) :)

src/coreComponents/dataRepository/Group.hpp
arng40: It add a `logLevel` wrapper if there is none, and the current `loglevel` description is appended to the wrapper description, all subClasses inherit descriptions from parent; if that answers the 1st part of your question.

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseFVM.cpp
paveltomin: Yes, the line break issue for residuals output still needs to be figured out

src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp
paveltomin: should be addressed now

paveltomin (approved):
please figure out what can be done with line break for residuals output, otherwise it should be good to go

paveltomin:
and there are few places in wave solvers that I didn't really touch but they should be changed for consistency

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseFVM.cpp
MelReyCG: What do you guys think of fully filling the message that will be output as a single line at once?
Here is what I am suggesting at `SolverBase:950`:
```c++
      // get residual norms
      std::ostringstream convergenceState;
      residualNorm = calculateResidualNorm( time_n, stepDt, domain, m_dofManager, m_rhs.values(), convergenceState );
      convergenceState << GEOS_FMT( "        ( R ) = ( {:4.2e} )", residualNorm );
      GEOS_LOG_LEVEL_INFO_RANK_0( logInfo::Convergence, convergenceState );
```
(`convergenceState` would be passed as reference in the `calculateResidualNorm()` calls, and filled just like the `std::cout` was filled before)

This way we definitively prevent this line to be "cut" if a warning appear.

paveltomin: I would just create a macro without line break maybe

src/coreComponents/dataRepository/Group.hpp
MelReyCG: Just an idea: we could do that by adding a general `GeneratedDocsRegistry` that would only be created & used during `ProblemManager::generateDocumentation()`. This way we would not need any compile-time check.
It could have a generic `GenerateDescription< GROUP_T >()` method that could be implemented to generate & store the `Wrapper` documentation for each `Group` types (in their respective CPP files).
This way we could free the documentation generation & storage from the simulation runs, and the documentation would be **stored at only one place rather than in each `Wrapper` instance** of each `Group`. Also, it doesn't rely on adding more virtual methods to `Group` nor `Wrapper`.
From my point of view, it would also clarify a bit where & when the documentation is generated and needed.

src/coreComponents/dataRepository/LogLevelsRegistry.hpp
```diff
@@ -0,0 +1,59 @@
...
+   * @return The log level string description
+   */
+  string buildLogLevelDescription() const;
+
+private:
+
+
+  /// Map for building the log level string for each wrapper
+  /// key : a logLevel condition, values : a set of description for a corresponding loglevel
+  std::map< integer, std::vector< std::string > > m_logLevelsDescriptions;
```
MelReyCG: `///` is for singleline comments, use this here
```suggestion
  /**
   * @brief Map for building the log level string for each wrapper.
   *        key : a logLevel condition, values : a set of description for a corresponding loglevel.
   */
  std::map< integer, std::vector< std::string > > m_logLevelsDescriptions;
```

src/coreComponents/dataRepository/LogLevelsRegistry.cpp
```diff
@@ -0,0 +1,46 @@
...
+string LogLevelsRegistry::buildLogLevelDescription() const
+{
+  std::ostringstream description;
+  description << "Sets the level of information to write in the standard output (the console typically).\n"
+                 "Level 0 outputs no specific information for this solver. Higher levels require more outputs.";
+  for( auto const & [logLevel, logDescriptions] : m_logLevelsDescriptions )
+  {
+    description << GEOS_FMT( "\n{}\n", logLevel );
+    for( size_t idxDesc = 0; idxDesc< logDescriptions.size(); idxDesc++ )
+    {
```
MelReyCG:
```suggestion
    for( size_t idxDesc = 0; idxDesc < logDescriptions.size(); idxDesc++ )
```

src/coreComponents/dataRepository/LogLevelsRegistry.cpp
```diff
@@ -0,0 +1,46 @@
...
+{
+
+void LogLevelsRegistry::addEntry( integer condition, std::string_view description )
+{
+  m_logLevelsDescriptions[condition].push_back( string( description ) );
+}
+
+/**
+ * @brief Construct the log level string description for a wrapper
+ */
```
MelReyCG: No need for documentation here? (->hpp)

src/coreComponents/dataRepository/LogLevelsInfo.hpp
```diff
@@ -0,0 +1,83 @@
...
+namespace geos
+{
+
+/**
+ * @brief Trait used to check whether a LOG_LEVEL_INFO structure is valid.
+ * @tparam LOG_LEVEL_INFO The log level structure to check.
+ */
+template< typename LOG_LEVEL_INFO >
+static constexpr bool is_log_level_info =
+  std::is_same_v< int, decltype(LOG_LEVEL_INFO::getMinLogLevel()) > &&
```
MelReyCG: You want an `integer` type here, below line 47, and in the `LogLevelsInfo.hpp` files.

src/coreComponents/physicsSolvers/contact/ContactSolverBase.cpp
```diff
@@ -42,6 +43,8 @@ ContactSolverBase::ContactSolverBase( const string & name,
 
   this->getWrapper< string >( viewKeyStruct::surfaceGeneratorNameString() ).
     setInputFlag( dataRepository::InputFlags::FALSE );
+
+  addLogLevel< logInfo::Configuration >();
```
MelReyCG: Be careful when adding a `logLevel` to a parent class, all child class must implement it.
`SolidMechanicsAugmentedLagrangianContact`: ok
`SolidMechanicsEmbeddedFractures`: unused
`SolidMechanicsLagrangeContact`: ok
`SolidMechanicsPenaltyContact`: unused

Don't forget to remove the include file.

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
```diff
@@ -1748,7 +1750,7 @@ void CompositionalMultiphaseWell::shutDownWell( real64 const time_n,
                                                     rankOffset,
                                                     localMatrix,
                                                     rhsValue,
-                                                    pres[ei], // freeze the current pressure value
+                                                    pres[ei],         // freeze the current pressure value
```
MelReyCG: Why adding those spaces here, l.1764 and l.1774?

src/coreComponents/dataRepository/LogLevelsInfo.hpp
```diff
@@ -0,0 +1,83 @@
...
+#define GEOS_COMMON_LOGLEVELSINFO_HPP
+
+#include "common/DataTypes.hpp"
+#include "common/format/Format.hpp"
+
+namespace geos
+{
+
+/**
+ * @brief Trait used to check whether a LOG_LEVEL_INFO structure is valid.
```
MelReyCG, line 29: As we no longer have exemples in this file, could you add an `@exemple` of a struct here?

src/coreComponents/physicsSolvers/SolverBase.cpp
```diff
@@ -98,6 +96,17 @@ SolverBase::SolverBase( string const & name,
...
 
+  addLogLevel< logInfo::Fields >();
+  addLogLevel< logInfo::LineSearch >();
+  addLogLevel< logInfo::Solution >();
+  addLogLevel< logInfo::Convergence >();
+  addLogLevel< logInfo::TimeStep >();
+  addLogLevel< logInfo::LinearSolver >();
+  addLogLevel< logInfo::NonlinearSolver >();
+  addLogLevel< logInfo::Timers >();
+  addLogLevel< logInfo::Initialization >();
```
MelReyCG: `logInfo::Initialization` seems to only apply to `SolidMechanicsStateReset` & `PoromechanicsInitialization`.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseStatistics.cpp
```diff
@@ -407,33 +410,45 @@ void CompositionalMultiphaseStatistics::computeRegionStatistics( real64 const ti
...
-                                        getName(), regionNames[i], time, regionStatistics.minTemperature, regionStatistics.averageTemperature, regionStatistics.maxTemperature ) );
-    GEOS_LOG_LEVEL_RANK_0( 1, GEOS_FMT( "{}, {} (time {} s): Total dynamic pore volume: {} rm^3",
-                                        getName(), regionNames[i], time, regionStatistics.totalPoreVolume ) );
-    GEOS_LOG_LEVEL_RANK_0( 1, GEOS_FMT( "{}, {} (time {} s): Phase dynamic pore volume: {} rm^3",
-                                        getName(), regionNames[i], time, regionStatistics.phasePoreVolume ) );
-    GEOS_LOG_LEVEL_RANK_0( 1, GEOS_FMT( "{}, {} (time {} s): Phase mass: {} {}",
-                                        getName(), regionNames[i], time, regionStatistics.phaseMass, massUnit ) );
+    GEOS_LOG_LEVEL_INFO_RANK_0( logInfo::Statistics,
+                                GEOS_FMT( "{}, {} (time {} s): Pressure (min, average, max): {}, {}, {} Pa",
+                                          getName(), regionNames[i], time, regionStatistics.minPressure, regionStatistics.averagePressure, regionStatistics.maxPressure ) );
```
MelReyCG: Can you factorize a bit?
```suggestion
    GEOS_LOG_LEVEL_INFO_RANK_0( logInfo::Statistics,
                                GEOS_FMT( "{}: Pressure (min, average, max): {}, {}, {} Pa",
                                          statPrefix, regionStatistics.minPressure, regionStatistics.averagePressure, regionStatistics.maxPressure ) );
```

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseStatistics.cpp
```diff
@@ -247,16 +248,22 @@ void SinglePhaseStatistics::computeRegionStatistics( real64 const time,
...
-                                        getName(), regionNames[i], time, regionStatistics.minDeltaPressure, regionStatistics.maxDeltaPressure ) );
-    GEOS_LOG_LEVEL_RANK_0( 1, GEOS_FMT( "{}, {} (time {} s): Temperature (min, average, max): {}, {}, {} K",
-                                        getName(), regionNames[i], time, regionStatistics.minTemperature, regionStatistics.averageTemperature, regionStatistics.maxTemperature ) );
-    GEOS_LOG_LEVEL_RANK_0( 1, GEOS_FMT( "{}, {} (time {} s): Total dynamic pore volume: {} rm^3",
-                                        getName(), regionNames[i], time, regionStatistics.totalPoreVolume ) );
-    GEOS_LOG_LEVEL_RANK_0( 1, GEOS_FMT( "{}, {} (time {} s): Total fluid mass: {} kg",
-                                        getName(), regionNames[i], time, regionStatistics.totalMass ) );
+    GEOS_LOG_LEVEL_INFO_RANK_0( logInfo::Statistics,
+                                GEOS_FMT( "{}, {} (time {} s): Pressure (min, average, max): {}, {}, {} Pa",
+                                          getName(), regionNames[i], time, regionStatistics.minPressure, regionStatistics.averagePressure, regionStatistics.maxPressure ) );
```
MelReyCG: Same as `CompositionalMultiphaseStatistics`.

MelReyCG (changes requested):
+ You may have forgotten `SourceFluxStatistics` & `SolverStatistics` (with `logInfo::Statistics`)

src/coreComponents/physicsSolvers/contact/ContactSolverBase.cpp
paveltomin: why "must"? what is your suggestion here?

MelReyCG: If one look at the `SolidMechanicsEmbeddedFractures`  or `SolidMechanicsPenaltyContact`  documentation, he will expect that the coresponding `logLevel` will output configuration information, which is not possible if I didn't miss anything.

MelReyCG: So my suggestion is to add this `addLogLevel()` in the `SolidMechanicsAugmentedLagrangianContact` and `SolidMechanicsLagrangeContact` constructors only.

paveltomin: I see you point, it makes sense, thanks

src/coreComponents/physicsSolvers/SolverBase.cpp
arng40: also for statistics

CusiniM (approved):

paveltomin:
this was around the top of the queue, no?
somehow felt back into "in review" but should be good to merge?

MelReyCG (approved):
@victorapm 's requests seem to be adressed as well as mines, so I think this PR is right to be merged.
There will be a follow-up PR for the other messages.

victorapm (approved):

PR #3234, opened 2024-07-23
refactor: Use matrix-matrix interface in LAPACK solve

dkachuma:
This PR replaces multiple matrix-vector solves with a single matrix-matrix solve following [PR3173](https://github.com/GEOS-DEV/GEOS/pull/3173). This also uses a version of the LAPACK call the destroys the coefficient matrix and rhs matrix to avoid further allocation in `BlasLapackLA`.


paveltomin (approved):

src/coreComponents/constitutive/fluid/multifluid/compositional/functions/NegativeTwoPhaseFlash.hpp
```diff
@@ -198,23 +198,21 @@ struct NegativeTwoPhaseFlash
...
+   * @param[in/out] A the coefficient matrix. Destroyed after call
+   * @param[in/out] X the rhs and solution
    * @return @c true if the problem is well solved @c false otherwise
    */
+  template< int USD >
   GEOS_HOST_DEVICE
-  static bool solveLinearSystem( arraySlice2d< real64 const > const & A,
-                                 arraySlice1d< real64 const > const & b,
-                                 arraySlice1d< real64 > const & x )
+  static bool solveLinearSystem( arraySlice2d< real64, USD > const & A,
```
CusiniM, line 204: I personally preferred having the `rhs` and the `solution` into two separate arrays. I am guessing that this is the lapack interface though... We could keep the separation on the GEOS side though.

CusiniM (approved):

src/coreComponents/constitutive/fluid/multifluid/compositional/functions/NegativeTwoPhaseFlash.hpp
dkachuma: The separate `rhs`/`solution` interface on `BlasLapackLA` still exists.  This change destroys the rhs but avoids having to allocate extra space for `solution` and `rhs` knowing that we are not going to need the rhs or the matrix after the solve.

CusiniM: is the allocation the same? A single `array2d` instead of 2 `array1d`?

dkachuma: My mistake. The allocation was on the matrix. With the previous interface, everytime we call the solve, additional space had to be allocated for the LU factors of the matrix. This way there is no extra space allocated since it's a single solve and the space allocated for the matrix is used for the LU factors (destroying the matrix but once we have solved we don't care anymore).

With the single column by column solve it would have been nice to factorize the matrix and then use the LU factors on each column but we don't have that interface. Instead I added an interface where you can do an in-place matrix-matrix solve.

PR #3197, opened 2024-06-30
refactor: separate friction and hydraulic aperture - stress relation.

CusiniM:
This pull request (PR) introduces a minor refactoring of the contact constitutive models. The changes are intended to facilitate:

1. The implementation of rate-and-state- dependent friction.

2. The incorporation of various normal stress-hydraulic aperture relationships beyond the current table-based approach.

3. The use of a different contact enforcement in the `Hydrofracture` solver.  

Below, I detail each proposed change and the rationale behind them:

1.  Separation of Friction Law and Hydraulic Aperture-Stress Relation:
   - Objective: Decouple the friction law and hydraulic aperture-stress relationship into two distinct objects.
   - Rationale: These elements serve different purposes, particularly evident in scenarios where hydraulic aperture does not influence dry fractures. This separation simplifies the user experience by eliminating unnecessary table definitions for non-flow scenarios and supports more focused class responsibilities.
   - Implementation: 
      - Friction models: derived from a `FrictionBase` class. These would be, for example, `CoulombFriction` and `RateAndStateFriction` . In general I would say that a friction model should have pretty limited responsibilities. The main thing it should know how to do is compute the `shearTraction` given the normal one. 
      - HydraulicAperture - stress models: derived from `HydraulicApertureBase` (I hate this name but could not think of a better one), these laws define the relationship between the `hydraulicAperture` and the contact stress. The current aperture table would, for example, be part of these and we could immediately add `BartonBandis` and that exponential model (in the past, @frankfeifan had to hardcode these in the kernels).

2. Move the penalty parameter from the friction model to the physicsSolvers t
  - Objective: Remove the penalty parameter from the friction model and integrate it into the physics solvers where necessary.
 - Rationale: The penalty parameter is a numerical paramer, not a physical property, and should be treated as such.

3. Introduction of `SolidMechanicsPenaltyContact`:
 - Objective: Create a distinct physics package for the penalty method, separating it from the `SolidMechanicsSolver`.
[...]

PR #3235, opened 2024-07-23
ConstantPermeability model fix for CompressibleSolidConstantPermeability

sohailwaziri:
dPerm_dDispJump was not defined for the ConstantPermeability model. In this PR, the field is added so that ConstantPermeability model can be used for CompressibleSolidConstantPermeability. This fix would enable assignment of constant permeability to fractures.


paveltomin (approved):

CusiniM:
I think we should avoid this if possible coz it will mean allocating another array4d every time we use this model which is quite a lot of extra "useless" data for all matrix cells, even in our flow simulations.  It would be better to finish up https://github.com/GEOS-DEV/GEOS/pull/3197 and, in there, add the possibility of having a constant hydraulic aperture that has no dependence on the jump.

PR #3236, opened 2024-07-23
feat: Immiscible water model

dkachuma:
Completes the water on the immiscible water model. This defines a compositional fluid model in which the water component does not participate in the flash but instead makes up it's own 3rd water/aqueous phase which is immiscible with the other 2 phases.

The phase partition is done via `ImmiscibleWaterFlash`. This removes the water component for the negative two phase flash of the "hydrocarbon" phases and reconstitutes the result into 3 phases.

The water properties (density and viscosity) are by providing reference values at a reference pressure and temperature and providing a compressibility and a thermal expansion coefficient.


src/coreComponents/constitutive/fluid/multifluid/compositional/models/EquationOfState.hpp
```diff
@@ -37,12 +37,14 @@ namespace compositional
 enum class EquationOfStateType : integer
 {
   PengRobinson,
-  SoaveRedlichKwong
+  SoaveRedlichKwong,
+  ImmiscibleWater
```
paveltomin: i am probably missing something
a bit confused here - how will it know what eos is to use for hc?

paveltomin (approved):

src/coreComponents/constitutive/fluid/multifluid/compositional/models/EquationOfState.hpp
dkachuma: This is specified in the xml. The phases are "lined up" with the EOS e.g.
```xml
 <CompositionalThreePhaseFluidLohrenzBrayClark
      name="blackOil"
      phaseNames="{ oil, gas, water }"
      equationsOfState="{ pr, pr, immiscibleWater }" />
```
However, I think since the `immiscibleWater` "EOS" is implied in the element name (`CompositionalThreePhaseFluidLohrenzBrayClark`), there is no need to specify this here. I'm inclined to remove it.

paveltomin:
move to merge queue?

corbett5 (approved):
Tests look fine to me.

paveltomin:
@rrsettgast please review

paveltomin:
Waiting on code owner review from @CusiniM, @jhuang2601, and/or @rrsettgast.

CusiniM (approved):
LGTM

PR #3239, opened 2024-07-23
refactor: Towards modularization of physicsSolver

rrsettgast:
- This moves the addition of header/source files into each subdirectory of the `physicsSolvers` module.  
- This introduces cmake flags to enable each subdir
- This DOES NOT create a module for each subdirectory


cssherman (approved):

wrtobin (approved):

paveltomin (approved):

acitrain (approved):

src/cmake/GeosxOptions.cmake
```diff
@@ -103,6 +103,16 @@ if( GEOS_PARALLEL_LINK_JOBS )
     set( CMAKE_JOB_POOL_LINK link_job_pool )
 endif()
 
+# Physics packages
+option( GEOS_ENABLE_CONTACT "Enables contact physics package" ON )
+option( GEOS_ENABLE_FLUIDFLOW "Enables fluid flow physics package" ON )
+option( GEOS_ENABLE_INDUCEDSEISMICITY "Enables induced seismicity physics package" ON )
+option( GEOS_ENABLE_MULTIPHYSICS "Enables multiphysics physics package" ON )
```
CusiniM, line 110: eventually we will need some logic to ensure that if someone, for example, enables multiphysics the other required packages are also enabled.

src/coreComponents/physicsSolvers/CMakeLists.txt
```diff
@@ -7,233 +7,49 @@ set( physicsSolvers_headers
...
-     wavePropagation/sem/elastic/secondOrderEqn/isotropic/ElasticWaveEquationSEM.cpp
-     wavePropagation/sem/elastic/firstOrderEqn/isotropic/ElasticFirstOrderWaveEquationSEM.cpp
-     wavePropagation/sem/acoustic/firstOrderEqn/isotropic/AcousticFirstOrderWaveEquationSEM.cpp
-     wavePropagation/sem/acoustic/secondOrderEqn/anisotropic/AcousticVTIWaveEquationSEM.cpp
-     wavePropagation/sem/acoustoelastic/secondOrderEqn/isotropic/AcousticElasticWaveEquationSEM.cpp )
+     SolverStatistics.cpp )
 
-  include( solidMechanics/kernels/SolidMechanicsKernels.cmake)
+if( GEOS_ENABLE_CONTACT )
+  add_subdirectory( contact )
```
CusiniM, line 21: I can move contactmechanics kernels to their own directory too.

src/pygeosx/pygeosx.cpp
```diff
@@ -5,7 +5,7 @@
  * Copyright (c) 2016-2024 Lawrence Livermore National Security LLC
  * Copyright (c) 2018-2024 Total, S.A
  * Copyright (c) 2018-2024 The Board of Trustees of the Leland Stanford Junior University
- * Copyright (c) 2018-2024 Chevron
```
CusiniM, line 8: did you run uncrustify on your mac? I think that every time I run it on ruby it changes that whitespace.

CusiniM (approved):

src/pygeosx/pygeosx.cpp
rrsettgast: on ruby...but only these two files....which is strange

src/cmake/GeosxOptions.cmake
rrsettgast: yeah. I am certain this needs to be developed, but I do think that "multiPhysics" isn't a good classification. There should be sub-folders/component under "multiPhysics" with specific coupled capabilities in each component. I think we discussed this at some point. I would like to get the existing circular dependencies worked out prior to full modularization of the physics packages.

src/coreComponents/physicsSolvers/CMakeLists.txt
rrsettgast: did I miss those?

castelletto1 (approved):

PR #3244, opened 2024-07-25
fix: Fix link flags after pr#3219

Algiane:
Fixes issue #3243.

 Also linked to the PR #3159 that is broken since the merge of the develop branch.

PR #3245, opened 2024-07-25
refactor: Change PhysicsSolver to PhysicsPackage

rrsettgast:
Changes:
- physicsSolvers -> physicsPackages
- SolverBase -> PhysicsPackageBase
- PhysicsSolverManager -> PhysicsPackageManager
- ContactSolverBase -> ContactPackageBase
- FlowSolverBase -> FlowPackageBase


.github/CODEOWNERS
```diff
@@ -28,16 +28,16 @@
...
+/src/coreComponents/physicsPackages/fluidFlow          @CusiniM @paveltomin @klevzoff @francoishamon
+/src/coreComponents/physicsPackages/inducedSeismicity  @CusiniM @rrsettgast
+/src/coreComponents/physicsPackages/multiphysics       @CusiniM @paveltomin @castelletto1 @rrsettgast
+/src/coreComponents/physicsPackages/python             @cssherman @corbett5
+/src/coreComponents/physicsPackages/simplePDE          @castelletto1 @rrsettgast @CusiniM 
+/src/coreComponents/physicsPackages/solidMechanics     @rrsettgast @castelletto1 @CusiniM 
+/src/coreComponents/physicsPackages/surfaceGeneration  @rrsettgast @CusiniM @jhuang2601
+/src/coreComponents/physicsPackages/wavePropagation    @sframba @acitrain
 /src/coreComponents/schema
 /src/coreComponents/unitTests                         @rrsettgast @corbett5 @wrtobin
```
rrsettgast:
```suggestion
/src/coreComponents/unitTests                         @rrsettgast @CusiniM 
```

acitrain (changes requested):
Hi Randy,

I tried to compile on Pangea 3 (only GEOS, units tests are currently broken due to a previous merge see PR #3244 ), and I found a remaining SolverBase inside 

src/coreComponents/physicsPackages/python/PySolver.cpp : geos::SolverBase * group; it is on line 43

CusiniM (approved):
It's pretty hard to review but it looks good. I think it would be nice to also collect all kernel files for the flow solvers in a subdir of fluidFlow like we did for the mechanics ones.

acitrain (approved):
Seems good ! 
Thanks Randy!

castelletto1:
Now that we got rid of the concept of `physics solver`, what is the meaning of function `solverStep` in `PhysicsPackageBase.hpp`? Maybe we should rename it.

rrsettgast:
> Now that we got rid of the concept of `physics solver`, what is the meaning of function `solverStep` in `PhysicsPackageBase.hpp`? Maybe we should rename it.

Good point. Options that I have thought about are:

- `physicsPackageStep()`: pretty self explanatory that it is a "step" for the physics package.
- `timeStep()`: From the codes point of view it is a timestep even if there is no significance on `dt` such as in a quasi-static mechanics step, or equilibrium heat flow.
- `physicsPackageIncrement()`: detach from the "timestep" and just call it an increment. The event manager kind of imposes the timestep, but perhaps the physics package shouldn't care. And in relativistic physics, the passing of time is not constant on all clocks.


* I am going with `simulationStep()`

PR #3247, opened 2024-07-26
refactor: Remove circular dependencies from fileIO/common/codingutilities

MelReyCG:
The point of this PR is to solve the circular dependencies of `fileIO`, `common` and `codingutilities` by:
- adding two `basicIO` and `format` in `common`.
  * `format` will contain `Format.*pp`, `fileIO/Table/*`, `fileIO/Section/*` 
    - _=> no more `fileIO <-> common`_
  * `logger`: will contain utilities logging (only one file for now but is going to grow)
    - _=> prevent further circular dependencies_
- lighten the `DataType.*pp` by making it only related to static types:
  * `rtTypes` & `Regex` will be in `codingUtilities`, aside of `StringEnums.hpp`
    - _=> no more `codingUtilities <-> common`_
  * no more MPI constants nor formatting nor logger inclusion in this header
  * no more need for a `DataType.cpp` 
    - _=> this header can now be included anywhere without needed any linking_


src/coreComponents/common/MpiWrapper.cpp
```diff
@@ -32,6 +32,12 @@
 namespace geos
 {
 
+#ifdef GEOS_USE_MPI
+MPI_Comm MPI_COMM_GEOSX;
```
CusiniM: maybe we can change that `MPI_COMM_GEOSX` to `MPI_COMM_GEOS` ?

CusiniM (approved):

src/coreComponents/common/MpiWrapper.cpp
MelReyCG: For sure!

untereiner (approved):

MelReyCG:
@rrsettgast @CusiniM I reverted the branch to the commit before the "MPI_COMM_GEOSX to MPI_COMM_GEOS" because this renaming now has its [own PR](https://github.com/GEOS-DEV/GEOS/pull/3259).

wrtobin (approved):

rrsettgast (approved):

PR #3255, opened 2024-07-30
refactor: Remove "using namespace" in header files

rrsettgast:
We should NOT have `using namespace` in headers. There are some cases where it isn't dangerous, like if in a function scope....but other than that we disallow it.


untereiner (approved):

CusiniM (approved):

castelletto1 (approved):

rrsettgast:
@acitrain @corbett5 @sframba @wrtobin Please review.

dkachuma (approved):

src/coreComponents/physicsSolvers/wavePropagation/sem/acoustic/secondOrderEqn/isotropic/AcousticWaveEquationSEM.cpp
```diff
@@ -790,13 +790,13 @@ real64 AcousticWaveEquationSEM::explicitStepForward( real64 const & time_n,
       if( m_enableLifo )
```
sframba, line 790: There are still two `using namespace` at the beginning of this file:
```
using namespace fields;
```
shouldn't they be removed as well?

rrsettgast: It is only header files we should avoid. Source files are ok

wrtobin (approved):

PR #3260, opened 2024-08-01
ci: split compliance and build step

untereiner:


rrsettgast:
@untereiner the way that these display in the actions tab is not good. There is no differentiation between the jobs on the summary/left bar.

untereiner:
@rrsettgast  In the Pull Request workflow you have each job prefixed by the reusable workflow name:
- Compliance Checks / semantic_pull_request
- etc.
- Build step / Ubuntu (20.04, gcc 9.4.0, open-mpi 4.0.3) / build_test_deploy
- etc.

It is not ideal. I'll try something else

rrsettgast:
replaced with #3266

PR #3266, opened 2024-08-02
ci: split ci into atomic workflows

untereiner:
1. Compliance Tests & CPU triggered on builds on a subset of PR events
2. CPU builds triggered on a push on develop or on a workflow call (cf. 1.)
3. GPU builds triggered on a push on develop or on a label event in a PR against develop
4. integrated tests triggered on a label event in a PR against develop
5. code coverage triggered on a label event in a PR against develop


untereiner:
:(

PR #3269, opened 2024-08-03
fix: testUnits failure on MacOS

### LINKED ISSUE #3263: Failing testUnits
This is macOS 13, apple clang 14, following quick start guide. Any ideas?

```
Start 87: testUnits

87: Test command: /tmp/geos/GEOS/build-macOS_arm-release/tests/testUnits "testUnits"
87: Working Directory: /tmp/geos/GEOS/build-macOS_arm-release/coreComponents/common/unitTests
87: Test timeout computed to be: 1500
87: Running main() from /tmp/geos/GEOS/src/cmake/blt/thirdparty_builtin/googletest/googletest/src/gtest_main.cc
87: [==========] Running 1 test from 1 test suite.
87: [----------] Global test environment set-up.
87: [----------] 1 test from Units
87: [ RUN      ] Units.SystemDurationFormatTest
87: /tmp/geos/GEOS/src/coreComponents/common/unitTests/testUnits.cpp:154: Failure
87: Expected equality of these values:
87:   durationCase.m_expectedString.c_str()
87:     Which is: "00h00m00s (1.11e-07 s)"
87:   TimeFormatInfo::fromDuration( durationCase.m_systemDuration ).toString().c_str()
87:     Which is: "00h00m00s (0 s)"
87: (Max possible duration = 9223372036854 s)
87: 
87: [  FAILED  ] Units.SystemDurationFormatTest (0 ms)
87: [----------] 1 test from Units (0 ms total)
87: 
87: [----------] Global test environment tear-down
87: [==========] 1 test from 1 test suite ran. (0 ms total)
87: [  PASSED  ] 0 tests.
87: [  FAILED  ] 1 test, listed below:
87: [  FAILED  ] Units.SystemDurationFormatTest
87: 
87:  1 FAILED TEST
1/1 Test #87: testUnits ........................***Failed    0.04 sec
```

rrsettgast:
MacOS truncates std::duration to microseconds. This PR changes the `SystemClock` to the `std::chrono::high_resolution_clock` to avoid truncation issues.
resolves #3263


timokoch:
mmh. Not 100% sure whether this test is robust. This here uses the default formatter: https://github.com/GEOS-DEV/GEOS/blob/b2e51fd8699bb2e08754361c3380eda703f8bdb6/src/coreComponents/common/Units.cpp#L67  I think this is implementation-defined (or depends on locale). I also don't think this has necessarily something to do with macOS. Could also be slight differences between fmt/std::format for example. Of course disabling the test for mac will pass on mac ;). Maybe a better and portable solution is to specify the format string in https://github.com/GEOS-DEV/GEOS/blob/b2e51fd8699bb2e08754361c3380eda703f8bdb6/src/coreComponents/common/Units.cpp#L67? Something like `{:.5e}`?

rrsettgast:
> mmh. Not 100% sure whether this test is robust. This here uses the default formatter:
> 
> https://github.com/GEOS-DEV/GEOS/blob/b2e51fd8699bb2e08754361c3380eda703f8bdb6/src/coreComponents/common/Units.cpp#L67
> 
> I think this is implementation-defined (or depends on locale). I also don't think this has necessarily something to do with macOS. Could also be slight differences between fmt/std::format for example. Of course disabling the test for mac will pass on mac ;). Maybe a better and portable solution is to specify the format string in
> https://github.com/GEOS-DEV/GEOS/blob/b2e51fd8699bb2e08754361c3380eda703f8bdb6/src/coreComponents/common/Units.cpp#L67
> 
> ? Something like `{:.5e}`?

Hello @timokoch ,
This problem appears to be due to the use of `std::chrono::system_clock` which truncates to microsecond on MacOS. 
here is the header from my MacOS system:
```c++
// -*- C++ -*-
//===----------------------------------------------------------------------===//
//
// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
// See https://llvm.org/LICENSE.txt for license information.
// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
//
//===----------------------------------------------------------------------===//

#ifndef _LIBCPP___CHRONO_SYSTEM_CLOCK_H
#define _LIBCPP___CHRONO_SYSTEM_CLOCK_H

#include <__chrono/duration.h>
#include <__chrono/time_point.h>
#include <__config>
#include <ctime>

#if !defined(_LIBCPP_HAS_NO_PRAGMA_SYSTEM_HEADER)
#  pragma GCC system_header
#endif

_LIBCPP_BEGIN_NAMESPACE_STD

namespace chrono
{

class _LIBCPP_EXPORTED_FROM_ABI system_clock
{
public:
    typedef microseconds                     duration;
    typedef duration::rep                    rep;
    typedef duration::period                 period;
    typedef chrono::time_point<system_clock> time_point;
    static _LIBCPP_CONSTEXPR_SINCE_CXX14 const bool is_steady = false;

    static time_point now() _NOEXCEPT;
    static time_t     to_time_t  (const time_point& __t) _NOEXCEPT;
    static time_point from_time_t(time_t __t) _NOEXCEPT;
};

#if _LIBCPP_STD_VER >= 20

template <class _Duration>
using sys_time    = time_point<system_clock, _Duration>;
using sys_seconds = sys_time<seconds>;
using sys_days    = sys_time<days>;

#endif

} // namespace chrono

_LIBCPP_END_NAMESPACE_STD

#endif // _LIBCPP___CHRONO_SYSTEM_CLOCK_H

```

timokoch (approved):
Ok seems like `high_resolution_clock` makes sense here and makes the test pass. Just out of interest: what it this test actually testing? Seems it (a) checks if the time formatter gives the expected format and (b) checks that fromSeconds and fromDuration gives the same. But why is the duration for the fromDuration test cast to the system clock duration type? Why not just use the duration type that chrono determines?

rrsettgast:
> Ok seems like `high_resolution_clock` makes sense here and makes the test pass. Just out of interest: what it this test actually testing? Seems it (a) checks if the time formatter gives the expected format and (b) checks that fromSeconds and fromDuration gives the same. But why is the duration for the fromDuration test cast to the system clock duration type? Why not just use the duration type that chrono determines?

I am actually not sure of the reason to do the conversion. It wouldn't be my choice, but I don't think I reviewed this when it went in to `develop`. I will as the originator of the reasons for this.

PR #3273, opened 2024-08-07
build: Corbett/kernel specs

corbett5:
TODO: If the `json` file is modified CMake doesn't run again.

Not a fan of having to duplicate the function declarations again in the `.cpp.template` file. Many of the functions have very complicated declarations and the errors obtained from mistakes are sure to cause headaches. For me this was in `physicsSolvers/fluidFlow/CompositionalMultiphaseHybridFVMKernels_impl.hpp`. One alternative is to define explicit instantiation macros in this header so at least you don't have to update two different files. You'd still have to list out the function arguments twice unless we resorted to some more intrusive macros.


src/coreComponents/common/TypeDispatch.hpp
```diff
@@ -375,17 +375,25 @@ bool dispatch( LIST const combinations,
...
+    // auto typeListPrinter = [typePrinter]( auto tlist ){ return internal::listToString( typename decltype( tlist )::type{}, "\n  ", "", typePrinter ); };
 
     GEOS_ERROR( "Types were not dispatched. The types of the input objects are:\n" <<
-                "( "<<(  ( "\n  " + LvArray::system::demangle( internal::typeIdWrapper( objects ).name() ) ) + ... )<<" \n)\n"<<
-                "and the dispatch options are:\n"<<
-                internal::listToString( combinations, "\n(", "\n)", typeListPrinter ) );
+                "( "<<(  ( "\n  " + LvArray::system::demangle( internal::typeIdWrapper( objects ).name() ) ) + ... )<<" \n)\n"
+                // <<
+                // "and the dispatch options are:\n"<<
+                // internal::listToString( combinations, "\n(", "\n)", typeListPrinter )
```
corbett5: Because `combinations` has so many different types I have run into compiler bugs when trying to print it out. Luckily the stack trace here seems to work find and that contains the types.

src/coreComponents/finiteElement/kernelInterface/KernelBase.hpp
```diff
@@ -466,6 +467,114 @@ real64 regionBasedKernelApplication( MeshLevel & mesh,
...
+  // save the maximum residual contribution for scaling residuals for convergence criteria.
+  real64 maxResidualContribution = 0;
+
+  NodeManager & nodeManager = mesh.getNodeManager();
+  EdgeManager & edgeManager = mesh.getEdgeManager();
+  FaceManager & faceManager = mesh.getFaceManager();
+  ElementRegionManager & elementRegionManager = mesh.getElemManager();
+
+  // using SUBREGION_TYPES = types::Slice< DISPATCH_TYPE_LIST >;
+  using SUBREGION_TYPES = types::TypeList< CellElementSubRegion >;
```
corbett5: This is a hack, but I am not sure how we want to do this. The `types::Slice` that I commented out will return a list of the first element from the list-of-lists `DISPATCH_TYPE_LIST`, but for what we have now this is just a list containing many copies of `CellElementSubRegion`. So many copies in fact that NVCC gives up in `forElementSubRegions`.

src/coreComponents/physicsSolvers/kernels-all.json
```diff
@@ -0,0 +1,512 @@
+{
```
corbett5: @rrsettgast do we need all these combinations?

corbett5:
@rrsettgast @wrtobin I think you guys should give this a look over. I commented a few of the pain points. It passes most of the integrated tests on ruby (except a few FPE runtime errors and restart checks I think are unrelated to this PR) and I have been unable to run them on Lassen.

All in all the compilation is still extremely slow. It takes almost two hours to compile on Lassen, most of which is spent linking.

src/coreComponents/physicsSolvers/kernels-all.json
```diff
@@ -0,0 +1,512 @@
...
+        "Q2_Hexahedron_Lagrange_GaussLobatto",
+        "Q3_Hexahedron_Lagrange_GaussLobatto",
+        "Q4_Hexahedron_Lagrange_GaussLobatto",
+        "Q5_Hexahedron_Lagrange_GaussLobatto",
+        "H1_Hexahedron_VEM_Gauss1",
+        "H1_Wedge_VEM_Gauss1",
+        "H1_Prism11_VEM_Gauss1"
+      ]
+    },
+    "explicit": []
```
rrsettgast: @corbett5 Can you put an example here and I will fill out the valid combinations?

corbett5: @rrsettgast this is not how it's done, but I'll probably try and change it to accept this format.

```
"explicit": [
  ["CellElementSubRegion", "DamageSpectral<ElasticIsotropic>", "H1_Hexahedron_Lagrange1_GaussLegendre2"],
  ["CellElementSubRegion", "DamageSpectral<ElasticIsotropic>", "H1_Wedge_Lagrange1_Gauss6"]
]
```

The order of the values should be the same as the order of the `vars`. In this case this is

```
"vars": [ "SUBREGION_TYPE", "CONSTITUTIVE_TYPE", "FE_TYPE" ]
```

src/coreComponents/finiteElement/kernelInterface/KernelBase.hpp
CusiniM: Can we just extract a list of unique subregion types from that slice? I am not saying I know how to do it but I am sure we can come up with some meta-programming magic to do that.  To be seen if nvcc does not crash while doing that though...

corbett5: I thought about that, I wanted to check in here before I invested the time.

CusiniM: we could also just loop over all `subRegions`  and then have a runtime check on the type inside the dispatch and only launch the kernels if the real type matches the dispatched one. Basically if the subregion cast failsinstead of errorring out we just ignore it.

corbett5: That was what I tried at first. Aside from the fact that I imagine there are cases where we would prefer to error out instead of silently proceeding it has the problem that not all subregions have the `FiniteElementBase` subgroup so the following call will fail

```c++
FiniteElementBase & subRegionFE = elementSubRegion.template getReference< FiniteElementBase >( finiteElementName );
```

If we want to go this way we could do the same thing as we do for the constitutive model and allocate a `NullFiniteElement` class.

CusiniM: I had not though of that... this is certainly not the most elegant path so let's see if we can come up with something cleaner as we add use cases for this.

corbett5: Certainly for now the simplest thing would be to just keep it as is `using SUBREGION_TYPES = types::TypeList< CellElementSubRegion >;` and then re-evaluate once we have a need for other types.

rrsettgast (approved):

CusiniM (approved):
Thanks. 

I would start by merging it as it is now. In the future, I would add the option to specify via command line a json file and store a couple in the repository with minimal configurations (only elasticity and hexahedra). We may also want to break it down into multiple files.

corbett5:
> Thanks.
> 
> I would start by merging it as it is now. In the future, I would add the option to specify via command line a json file and store a couple in the repository with minimal configurations (only elasticity and hexahedra). We may also want to break it down into multiple files.

To me it would make sense if the full specifications for each solver lived in a `json` file adjacent to the solver itself, but we'd certainly want to keep the ability for a user to specify a single override file.

src/coreComponents/physicsSolvers/CMakeLists.txt
```diff
@@ -115,7 +146,35 @@ endif()
...
+foreach( kernelTemplateFile ${kernelTemplateFileList} )
+  get_filename_component( jsonKey ${kernelTemplateFile} NAME_WE )
+  generateKernels( TEMPLATE ${kernelTemplateFile}
+                   JSON kernelSpecs
+                   KEY ${jsonKey}
+                   HEADERS headerFiles
+                   SOURCES sourceFiles )
+
+  list(APPEND physicsSolvers_headers ${headerFiles})
+  list(APPEND physicsSolvers_sources ${sourceFiles})
```
CusiniM: @corbett5, @rrsettgast I had not noticed but in reality the kernels should be added to the sources of each individual package not to the `physicsSolvers` component. We may want to fix this before merging it. I am happy to help though.

rrsettgast: @CusiniM I pushed everything down into the components

PR #3274, opened 2024-08-08
feat: Do not allow linear solver fail feature

jafranc:
Add flags and logic to drop nonlinear loop if a linear step ends up in `NonConverged` state.


src/coreComponents/physicsSolvers/SolverBase.hpp
```diff
@@ -586,6 +586,7 @@ class SolverBase : public ExecutableGroup
     static constexpr char const * targetRegionsString() { return "targetRegions"; }
     static constexpr char const * meshTargetsString() { return "meshTargets"; }
     static constexpr char const * writeLinearSystemString() { return "writeLinearSystem"; }
+    static constexpr char const * noLinearSolveFailString() { return "noLinearSolveFail"; }
```
CusiniM: maybe `allowNonConvergedLinearSolverSolution` ?

jafranc: Will amend asap

paveltomin:
@CusiniM @rrsettgast this should be good to go

jafranc:
@paveltomin I tried pushing this to merge queue adding flags to trigger all required testing. Is it still a `must-rebaseline` as we add to the registered vars or did the baseline logic changed ?

paveltomin:
> @paveltomin I tried pushing this to merge queue adding flags to trigger all required testing. Is it still a `must-rebaseline` as we add to the registered vars or did the baseline logic changed ?

I think you did everything correctly but tests job just randomly failed on the compilation
This PR should require rebaseline, nothing changed for that
And it can not go into merge queue yet since there are no reviews

CusiniM (approved):

PR #3159, opened 2024-06-04
ci: Action with Pangea-3 installation reproduction and ppc64le emulation

Algiane:
New job that: 
  - emulates a ppc64 architecture (using the [docker/setup-qemu-action](https://github.com/docker/setup-qemu-action) that relies on the use of [`qemu`](https://www.qemu.org/) through the [`qemu-user-static`](https://github.com/multiarch/qemu-user-static) image);

  - deploys a AlmaLinux-8 image with prebuilt TPLs and Geos' dependencies installed with respect to the required pangea3 modules:
    - CMake-3.26
    - gcc-9.4.0
    - ompi-4.1.2
    - cuda-11.5.0
    - openblas-0.3.18
    - lsf-10.1 
 
  - adds a `HOST_ARCH` matrix variable to the job matrix to trigger the installation and the use of the emulation layer;

  - builds Geos and the unit tests on `streak-2` self-hosted runner
  
  - associated to the TPLs PR [257](https://github.com/GEOS-DEV/thirdPartyLibs/pull/257). 
  
  **Remark**
Unit tests are not run because, due to the emulation layer, the GPUs cannot be used inside docker (the x86_64 drivers coming from the host are not usable inside the ppc64le image). Using them is theoritically possible but not so easy: if I am not wrong, at least one GPU has to be dedicated to the ppc64le image, the suitable drivers have to be installed and the GPU restarted with this driver.

**Tasks**
- [ ] Wait on #3276 to switch from ninja to cmake
- [ ] Implement reduction in runtime by limiting to wave solver compilation as in #3340 
- [ ] Add dependency on CI flag `ci: run CUDA builds`
- [ ] run on streak2 (8 cores)

PR #3276, opened 2024-08-09
ci: add the possitility to use other build generators than ninja inside the ci

Algiane:
# PR objectives
   1. Provide the possibility to remove the dependency to the `Ninja` tool in the ci (smaller docker images, lighter setting)
   2. Test the Pangea environment (we mainly use `unix makefiles`, not `Ninja` ones) since errors may differ depending on the generator used (see [pangea3 PR](https://github.com/GEOS-DEV/GEOS/pull/3159) where ninja fails while make succeed)

# Implementation
  - Add the possibility to pass one of the `--ninja`, `--makefile`, `--xcode` or `--eclipse` command line arguments to the [`ci_build_and_test_in_container.sh`](https://github.com/GEOS-DEV/GEOS/commit/b41f2059f44109dec80f61708c88aed77f9657eb#diff-0cef284cfdb08fea23f36d5d918fe37db5c5af609b2837dc27253a354dc1e641) script. Variable names have been chosen in agreement with the [`config-build.py`](https://github.com/GEOS-DEV/GEOS/blob/b41f2059f44109dec80f61708c88aed77f9657eb/scripts/config-build.py) script. The `--makefile` option is added to build unix makefiles.
 
   - Add a `BUILD_GENERATOR` input for the [`build_and_test.yml`](https://github.com/GEOS-DEV/GEOS/pull/3276/files#diff-48c0ee97c53013d18d6bbae44648f7fab9af2e0bf5b0dc1ca761e18ec5c478f2) script. `--ninja` option is used by default to enable the Ninja generator (as it was the initial ci behaviour). The default value is used for CPU builds.

   - Use of the `BUILD_GENERATOR` value of the GPU builds as input for the [`build_and_test.yml`](https://github.com/GEOS-DEV/GEOS/pull/3276/files#diff-48c0ee97c53013d18d6bbae44648f7fab9af2e0bf5b0dc1ca761e18ec5c478f2) call (because I will link this PR to another one where one of the GPU build will use the unix makefile generator).

   - Set the `BUILD_GENERATOR` variable to `--ninja` in the matrix of GPU builds. If not explicitely provided in the matrix definition, an empty string (leading to unix makefile generation) is passed as `BUILD_GENERATOR` input to the build_and_test.yml script 
  
# Validation
Commit [`077a1c08bdbf5`](https://github.com/GEOS-DEV/GEOS/commit/077a1c08bdbf5330c332ed7257929396a9874304) has been used to validate that for now all jobs that were using `ninja` still use it and that the PR doesn't modify the ci.

Fixes PR #3159 for which ninja build is broken since commit [4bef600a](https://github.com/GEOS-DEV/GEOS/commit/4bef600a0dfdf8e8a9bf5ad868a131164d84f39b).


CusiniM (approved):

scripts/ci_build_and_test_in_container.sh
```diff
@@ -38,12 +38,18 @@ Usage: $0
       run a code build and test.
   --data-basename output.tar.gz
       If some data needs to be extracted from the build, the argument will define the tarball. Has to be a `tar.gz`.
+  --eclipse
+      Use "Eclipse CDT4 - Unix Makefiles" as build system generator.
```
rrsettgast: Why would we include this?

scripts/ci_build_and_test_in_container.sh
```diff
@@ -58,6 +64,8 @@ Usage: $0
       Basename of the json credentials file to connect to the sccache cloud cache.
   --test-code-style
   --test-documentation
+  --xcode
+      Use "Xcode" as build system generator.
```
rrsettgast: also...why include this?

rrsettgast (approved):
I don't see a use for Xcode and eclipse in this context.

scripts/ci_build_and_test_in_container.sh
Bubusch: removed

scripts/ci_build_and_test_in_container.sh
Bubusch: removed

Algiane: It was for genericity purpose. The script can be executed outside the ci by a local `docker run` call so I was guessing that we may want to use one of the generators supported by the project. 
But I have not tested it so ok to remove it and to focus on the ci only.

PR #3279, opened 2024-08-12
feat: averaging of stress components using quadrature

ryar9534:
Implementation of a new averaged stress output per cell, where the averaging for each component is done using the quadrature rule of the finite element. 

The new field is averageStress (6 components per cell) and is registered/owned by the solid mechanics solver, like the average strain (which I have made some minor changes to in order to unify with this new stress). This average stress is stored in addition to the stress at the quadrature points, but the stress owned by the constitutive law is no longer output (using the arithmetic mean). 

An annoying rebaseline will be required, as this is gonna change all of the mechanics/poromechanics.


src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsFields.hpp
```diff
@@ -46,6 +46,10 @@ using array2dLayoutStrain = array2d< real64, cells::STRAIN_PERM >;
 using arrayView2dLayoutStrain = arrayView2d< real64, cells::STRAIN_USD >;
 using arrayViewConst2dLayoutStrain = arrayView2d< real64 const, cells::STRAIN_USD >;
 
+using array2dLayoutAvgStress = array2d< real64, cells::STRAIN_PERM >;
+using arrayView2dLayoutAvgStress = arrayView2d< real64, cells::STRAIN_USD >;
+using arrayViewConst2dLayoutAvgStress = arrayView2d< real64 const, cells::STRAIN_USD >;
```
CusiniM: can we call `cells::STRAIN_PERM` something like`cells::TENSOR_TWO_PERM`?

CusiniM (approved):
we should make the stress `NO_PLOT`.

ryar9534:
> we should make the stress `NO_PLOT`.

The new average stress or the old stresses?

CusiniM:
> > we should make the stress `NO_PLOT`.
> 
> The new average stress or the old stresses?

the old ones should disappear I guess.

ryar9534:
> > > we should make the stress `NO_PLOT`.
> > 
> > 
> > The new average stress or the old stresses?
> 
> the old ones should disappear I guess.

So are we ok with the overall increased memory footprint? The advantage of the old stresses is that the average is never stored explicitly

CusiniM:
> > > > we should make the stress `NO_PLOT`.
> > > 
> > > 
> > > The new average stress or the old stresses?
> > 
> > 
> > the old ones should disappear I guess.
> 
> So are we ok with the overall increased memory footprint? The advantage of the old stresses is that the average is never stored explicitly

well, yes, but it's also wrong and we certainly don't want the output to depend on FE.

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsFields.hpp
castelletto1: `SYMMETRIC_RANK_TWO_TENSOR_PERM`

src/coreComponents/physicsSolvers/solidMechanics/kernels/StrainHelper.hpp
```diff
@@ -214,6 +214,182 @@ class AverageStrainOverQuadraturePointsKernelFactory
...
+
+
+
+/**
+ * @class AverageStressOverQuadraturePointsKernelFactory
+ * @brief Class to create and launch the kernel
+ */
+class AverageStressOverQuadraturePointsKernelFactory
+{
+public:
```
castelletto1, line 229: Having this class in `StrainHelper.hpp` could be a bit confusing.

src/coreComponents/physicsSolvers/solidMechanics/kernels/StrainHelper.hpp
```diff
@@ -214,6 +214,182 @@ class AverageStrainOverQuadraturePointsKernelFactory
...
+  }
+
+  /**
+   * @brief Increment the average property with the contribution of the property at this quadrature point
+   * @param k The element index
+   * @param q The quadrature point index
+   * @param stack The StackVariables object that hold the stack variables.
+   */
+  GEOS_HOST_DEVICE
+  void quadraturePointKernel( localIndex const k,
```
castelletto1: This file contains two functions `quadraturePointKernel` once for the strain and one for the stress. In my opinion only one function should exist to perform the averaging of a second order symmetric stress tensor -- and probably of any scalar (rank 0), vector (rank 1), ... . The class [`AverageOverQuadraturePointsBase`](https://github.com/GEOS-DEV/GEOS/blob/93b3a72a965fad4cc349674286bafb08761dac86/src/coreComponents/mesh/utilities/AverageOverQuadraturePointsKernel.hpp#L38)  should be responsible for this task, and the quantity to be averaged should either provided directly or with a lambda function that evaluates such property.

src/coreComponents/physicsSolvers/solidMechanics/kernels/StrainHelper.hpp
```diff
@@ -214,6 +214,182 @@ class AverageStrainOverQuadraturePointsKernelFactory
...
+                              StackVariables & stack ) const
+  {
+    //real64 const weight = FE_TYPE::transformedQuadratureWeight( q, stack.xLocal, stack.feStack ) / m_elementVolume[k];
+
+    real64 dNdX[ FE_TYPE::maxSupportPoints ][3];
+    real64 const detJxW = m_finiteElementSpace.template getGradN< FE_TYPE >( k, q, stack.xLocal, stack.feStack, dNdX );
+
+    for( int icomp = 0; icomp < 6; ++icomp )
+    {
+      m_avgStress[k][icomp] += detJxW*m_stress[k][q][icomp]/m_elementVolume[k];
```
castelletto1: Here we divide by the cell element volume (`m_elementVolume[k]`) to compute the increment to the average value. In `AverageOverQuadraturePointsBase` the cell element contribution is directly [included in the weight definition](https://github.com/GEOS-DEV/GEOS/blob/93b3a72a965fad4cc349674286bafb08761dac86/src/coreComponents/mesh/utilities/AverageOverQuadraturePointsKernel.hpp#L208). Both correct, but it would be nice to be consistent

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsFields.hpp
CusiniM: the only issue of making it too long is that every time we want to get a view out then we have to write like:
`arrayView2d< real64 , cells::SYMMETRIC_RANK_TWO_TENSOR_PERM >` but that's indeed the best name. Although I don't think the permutation would change for non symmetric tensor.

src/coreComponents/physicsSolvers/solidMechanics/kernels/StrainHelper.hpp
CusiniM: we could generalize it and come up with a single kernel. We could try to make it accept a list of views instead of a specific one. It would also mean that we do all averaging in a single kernel launch which it is indeed better. We could use something similar to the `elementViewAccessors` and create an `array1d` of all scalars, one for all vectors and one for all tensors to be averaged and have the `averagingKernel` only ask for the keys of the fields that need to be averaged. 
The strain probably remains a bit of a special case though and i don't know that we do any averaging for other quantities.

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsFields.hpp
castelletto1: Yes, you are right `RANK_TWO_TENSOR_PERM` should be enough.

castelletto1: It's long, but clear and general.

src/coreComponents/physicsSolvers/solidMechanics/kernels/StrainHelper.hpp
CusiniM: IMO we should divide only once at the end.

src/coreComponents/physicsSolvers/solidMechanics/kernels/StrainHelper.hpp
castelletto1: In the long term, such computations should be managed by a numerical quadrature library (i.e. --> [Shiva](https://github.com/GEOS-DEV/Shiva))

src/coreComponents/physicsSolvers/solidMechanics/kernels/StrainHelper.hpp
castelletto1: Generally speaking, I would argue that `m_elementVolume[k]` should not even be required and computed directly in the quadrature process. For this specific case:

```c++
real64 cellVolume{0.0}
for( int icomp = 0; icomp < 6; ++icomp )
{
  m_avgStress[k][icomp] += detJxW*m_stress[k][q][icomp];
  cellVolume += detJxW;
}
m_avgStress[k][icomp] /= cellVolume;
```

paveltomin (approved):

CusiniM:
what's the status of this PR?

ryar9534:
> what's the status of this PR?

@CusiniM I just didnt know if this is still something folks want in develop. I originally did it to see if it helped with some stress oscillations in NL but it didnt.

rrsettgast:
@herve-gross @jhuang2601 
Can we figure out if this is needed?

paveltomin:
@herve-gross ping

herve-gross:
We discussed this issue last week. Some of the problem is purely visual (smoothing stresses in post-processing suffices). @castelletto1 and @jhuang2601 are putting together a case to check what is needed inside and outside of GEOS.

jhuang2601 (approved):

herve-gross (approved):
It will be great to have documentation on this too

jhuang2601:
> It will be great to have documentation on this too

@herve-gross I update the Kirsch wellbore problem to output and verify averageStress with analytical solutions.

PR #3283, opened 2024-08-14
refactor: Reuse computeSinglePhaseFlux

paveltomin:


paveltomin:
somehow requires rebaseline due to some numerical diffs, for example:
```
Error: /Problem/domain/MeshBodies/mesh/meshLevels/Level0/ElementRegions/elementRegionsGroup/Domain/elementSubRegions/cb1/deltaPressure
	Arrays of types float64 and float64 have 21 values of which 10 fail both the relative and absolute tests.
		Max absolute difference is at index (np.int64(19),): value = 0.00641802791506052, base_value = 0.006416955031454563
		Max relative difference is at index (np.int64(20),): value = 0.00043659377843141556, base_value = 0.00043603498488664627
	Statistics of the q values greater than 1.0 defined by absolute tolerance: N = 4
		max = 10728.836059570312, mean = 867.0169682729812, std = 2522.583500320097
		max is at index (np.int64(19),), value = 0.00641802791506052, base_value = 0.006416955031454563
	Statistics of the q values greater than 1.0 defined by relative tolerance: N = 6
		max = 1585.5540718947793, mean = 88.88597318065965, std = 336.763511641095
		max is at index (np.int64(18),), value = 0.09559668321162462, base_value = 0.09559607692062855
```

CusiniM (approved):
why was this even different? 
we should add a unitTest for this computeFlux function.

ryar9534 (approved):

ryar9534:
> why was this even different? we should add a unitTest for this computeFlux function.

We should probably note somewhere to add the unit test, as if this gets merged I have a feeling we will forget

jhuang2601 (approved):

dkachuma (approved):

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseFVMKernels.hpp
```diff
@@ -37,12 +37,14 @@
...
 #include "physicsSolvers/fluidFlow/SinglePhaseBaseKernels.hpp"
 #include "physicsSolvers/fluidFlow/StencilAccessors.hpp"
+#include "physicsSolvers/fluidFlow/FluxKernelsHelper.hpp"
 
 namespace geos
 {
 
 namespace singlePhaseFVMKernels
 {
+using namespace fluxKernelsHelper;
```
rrsettgast: please don't add `using namespace` inside of headers.

rrsettgast (changes requested):

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseFVMKernels.hpp
paveltomin: removed it

rrsettgast (approved):

rrsettgast:
@paveltomin @jhuang2601 @CusiniM 
so does this get taken off the queue until a unit test is created?

CusiniM:
> @paveltomin @jhuang2601 @CusiniM so does this get taken off the queue until a unit test is created?
okay let's just add an issue.

paveltomin:
> @paveltomin @jhuang2601 @CusiniM so does this get taken off the queue until a unit test is created?

issue is here https://github.com/GEOS-DEV/GEOS/issues/3297

i will merge it for now

PR #3280, opened 2024-08-12
feat: add rate- and state-dependent friction model & quasi-dynamic solver.

CusiniM:
- [x] Add Rate- and State-dependent friction model
- [x] Create a `QuasiDynamicEQ` solver that drives EQ simulations by:
  - uses a stress solver to update tractions on the fault (It can be any of the fracture solvers). 
  - launches kernels to solve rate-and-state equations on each fault element.
- [x] If no stress solver is specified, add the possibility to add  function to compute tractions on the fault (e.g., the spring-slider shear traction formula)  
- [x] Add rate- and state kernels that, for now, use Euler-backward to solve the R&S equations (for now 1d on the fault).
- [x] Add spring-slider xml example and integratedTest.

For now, everything is 1D on the fault. The 2D case and the coupling to the contact solver with Lagrange multiplier will be tackled in a followup PR.

LvArray:: https://github.com/GEOS-DEV/LvArray/pull/329

PR #3287, opened 2024-08-19
feat: Add kernel callable dense LA solvers for small systems.

CusiniM:


CusiniM:
@dkachuma this should be useful for the PVT module too.

src/coreComponents/denseLinearAlgebra/denseLASolvers.hpp
```diff
@@ -0,0 +1,349 @@
...
+ * @param[in,out] b The right-hand side vector. It is used to compute the solution, but the values are not preserved.
+ * @param[out] x The solution vector. The result of solving the system `Ax = b` using back substitution.
+ */
+template< std::ptrdiff_t N, 
+          typename MATRIX_TYPE,
+          typename RHS_TYPE,
+          typename SOL_TYPE >
+void solveUpperTriangularSystem( MATRIX_TYPE const & A, RHS_TYPE & b, SOL_TYPE && x )
+{
+  for (std::ptrdiff_t i = N - 1; i >= 0; --i)
```
castelletto1:
```suggestion
  for (integer i = N - 1; i >= 0; --i)
```

src/coreComponents/denseLinearAlgebra/denseLASolvers.hpp
```diff
@@ -0,0 +1,349 @@
...
+ * @param[in] b The constant right-hand side vector.
+ * @param[out] x The solution vector. The result of solving the system `Ax = b`.
+ */
+template< std::ptrdiff_t N, 
+          typename MATRIX_TYPE,
+          typename RHS_TYPE,
+          typename SOL_TYPE >
+GEOS_HOST_DEVICE 
+inline
+void solve( MATRIX_TYPE const & A, RHS_TYPE const & b, SOL_TYPE && x )
```
castelletto1: We introduce some code duplication compared with what is available in `BlasLapackLA` -- specifically for `2x2` and `3x3` matrices.

src/coreComponents/denseLinearAlgebra/denseLASolvers.hpp
CusiniM: I was trying to be consistent with all LvArray methods. Honestly these systems should be so small that my guess is that this should not matter though.

src/coreComponents/denseLinearAlgebra/denseLASolvers.hpp
CusiniM: yes, but this can be called in kernel. We could remove the blas lapack ones.

castelletto1: Agreed, the custom implementations in `BlasLapackLA.cpp` should be dropped in favor of those you added here

CusiniM: @dkachuma do you also need a version to handle multiple `rhs`?

src/coreComponents/denseLinearAlgebra/denseLASolvers.hpp
```diff
@@ -0,0 +1,349 @@
...
+  static_assert( N > 0, "N must be greater than 0." );
+  internal::checkSizes< N, N >( matrix );
+  internal::checkSizes< N >( b ); 
+  internal::checkSizes< N >( x );
+  
+
+  // Step 1: Transform  into an upper triangular matrix 
+  
+  // 1.a. Find the pivot
+  for (std::ptrdiff_t i = 0; i < N; ++i)
```
victorapm: Minor detail: how large do we expect `N` to be? For large enough (depends on the GPU, but let's say a conservative 8x8), the current threading strategy will lead to register spilling and suboptimal performance.

victorapm (approved):

src/coreComponents/denseLinearAlgebra/denseLASolvers.hpp
castelletto1: This is very important @victorapm. Once we agree on the value for `N`, we should throw an error if larger systems are used.

CusiniM: that is a good point. I was thinking of adding a `static_assert` to enforce N to be smaller than a certain number (10?). What to do for larger systems is up for discussion. We would have to come up with a mechanism to change the threading strategy depending on the size of the system. Maybe it's possible through `raja`?

src/coreComponents/denseLinearAlgebra/denseLASolvers.hpp
```diff
@@ -0,0 +1,349 @@
...
+  real64[N] b_copy{};
+
+  for(std::ptrdiff_t i=0; i < N; ++j)
+  {
+    b_copy[i] = b[i];
+    for( std::ptrdiff_t j=0; j < N; ++j )
+    {
+      A_copy[i][j] = A[i][j];
+    };
+  };
```
castelletto1: Could we add a flag to optionally copy `A` and `b`, the default implementation overwriting the provided matrix and rhs?

castelletto1: The option would be irrelevant for the `2x2` and `3x3` cases.

src/coreComponents/denseLinearAlgebra/denseLASolvers.hpp
CusiniM: Actually, in the way the current `BlasLapack::solveLinearSystem` is called the system size is not known at compilation time. I guess we can add an interface with a dispatch to cal this method with the proper size?

src/coreComponents/denseLinearAlgebra/denseLASolvers.hpp
CusiniM: Like instead of defining the `const` and `non-const` version of this? The version right below takes non-const A and b and modifies them.

castelletto1: Precisely

CusiniM: I find having a `const` and `non-const `version more elegant than a flag but we can have just the non-const version with a flag o another template parameter.

CusiniM: Actually, doing what you suggested removed some code duplications.

src/coreComponents/denseLinearAlgebra/denseLASolvers.hpp
```diff
@@ -0,0 +1,291 @@
...
+ *
+ * This function solves the linear system `Ax = b`, where `A` is an upper triangular matrix, using
+ * back substitution. The solution `x` is computed and stored in the provided output vector.
+ *
+ * @tparam N The size of the square matrix `A`.
+ * @tparam MATRIX_TYPE The type of the matrix `A`.
+ * @tparam RHS_TYPE The type of the right-hand side vector `b`.
+ * @tparam SOL_TYPE The type of the solution vector `x`.
+ * @param[in] A The upper triangular matrix representing the coefficients of the system.
+ * @param[in,out] b The right-hand side vector. It is used to compute the solution, but the values are not preserved.
```
dkachuma: I don't understand this comment. This method looks like it doesn't change `b`.

src/coreComponents/denseLinearAlgebra/denseLASolvers.hpp
```diff
@@ -0,0 +1,291 @@
...
+          typename MATRIX_TYPE,
+          typename RHS_TYPE,
+          typename SOL_TYPE,
+          bool MODIFY_MATRIX = true >
+GEOS_HOST_DEVICE
+inline
+void solve( MATRIX_TYPE & A, RHS_TYPE & b, SOL_TYPE && x )
+{
+  static_assert( N > 0, "N must be greater than 0." );
+  static_assert( N < 10, "N cannot be larger than 9" );
```
dkachuma, line 277: For the PVT application, we may have a size of 2*NC+1 where NC is number of components. I think locally (with PVT driver) we allow up to 9 components.

dkachuma: Just seen the comments above about registers. Maybe I can "manually" reduce the matrix size to NC+1.

src/coreComponents/denseLinearAlgebra/denseLASolvers.hpp
dkachuma: @CusiniM Yes the system I have has multiple `rhs`. Of course I can do that as several single `rhs` calls if that was deemed more efficient than allocating another matrix.

dkachuma (approved):

corbett5:
I feel like these would fit in better in LvArray. I mean it's already got the functions for solving 2x2 and 3x3 eigenvalue and inverse problems, why shouldn't the linear system solvers live there as well? Where ever they wind up there should be unit tests as well.

src/coreComponents/denseLinearAlgebra/denseLASolvers.hpp
```diff
@@ -0,0 +1,291 @@
...
+inline
+void solveThreeByThreeSystem( MATRIX_TYPE const & A, RHS_TYPE const & b, SOL_TYPE && x )
+{
+  LvArray::tensorOps::internal::checkSizes< 3, 3 >( A );
+  LvArray::tensorOps::internal::checkSizes< 3 >( b );
+  LvArray::tensorOps::internal::checkSizes< 3 >( x );
+
+  real64 const detA = A[0][0] * (A[1][1] * A[2][2] - A[1][2] * A[2][1]) -
+                      A[0][1] * (A[1][0] * A[2][2] - A[1][2] * A[2][0]) +
+                      A[0][2] * (A[1][0] * A[2][1] - A[1][1] * A[2][0]);
```
rrsettgast:
```suggestion
  real64 const invDetA = 1.0 / ( A[0][0] * (A[1][1] * A[2][2] - A[1][2] * A[2][1]) -
                                 A[0][1] * (A[1][0] * A[2][2] - A[1][2] * A[2][0]) +
                                 A[0][2] * (A[1][0] * A[2][1] - A[1][1] * A[2][0]) );
```

src/coreComponents/denseLinearAlgebra/denseLASolvers.hpp
```diff
@@ -0,0 +1,291 @@
...
+                       A[0][1] * (b[0] * A[2][2] - b[2] * A[2][0]) +
+                       A[0][2] * (b[0] * A[1][2] - b[1] * A[1][0]);
+
+  real64 const detX2 = A[0][0] * (A[1][1] * b[2] - A[1][2] * b[1]) -
+                       A[0][1] * (A[1][0] * b[2] - A[1][2] * b[0]) +
+                       A[0][2] * (A[1][0] * b[1] - A[1][1] * b[0]);
+
+  x[0] = detX0 / detA;
+  x[1] = detX1 / detA;
+  x[2] = detX2 / detA;
```
rrsettgast:
```suggestion
  x[0] = detX0 * invDetA;
  x[1] = detX1 * invDetA;
  x[2] = detX2 * invDetA;
```

src/coreComponents/denseLinearAlgebra/denseLASolvers.hpp
```diff
@@ -0,0 +1,291 @@
...
+  LvArray::tensorOps::internal::checkSizes< 2, 2 >( A );
+  LvArray::tensorOps::internal::checkSizes< 2 >( b );
+  LvArray::tensorOps::internal::checkSizes< 2 >( x );
+
+  real64 const detA = A[0][0] * A[1][1] - A[0][1] * A[1][0];
+
+  GEOS_ERROR_IF_LT_MSG( LvArray::math::abs( detA ), LvArray::NumericLimits< real64 >::epsilon; , "Singular system." );
+
+  x[0] = (A[1][1] * b[0] - A[0][1] * b[1] ) / detA;
+  x[1] = (A[0][0] * b[1] - A[1][0] * b[0] ) / detA;
```
rrsettgast:
```suggestion
  real64 const invDetA = 1.0 / ( A[0][0] * A[1][1] - A[0][1] * A[1][0] );

  GEOS_ERROR_IF_LT_MSG( LvArray::math::abs( detA ), LvArray::NumericLimits< real64 >::epsilon; , "Singular system." );

  x[0] = (A[1][1] * b[0] - A[0][1] * b[1] ) * invDetA;
  x[1] = (A[0][0] * b[1] - A[1][0] * b[0] ) * invDetA;
```

rrsettgast (changes requested):
These optimizations are not guaranteed to be done by the compiler.

we should have a plan for a central location for all operations of this type. It was in `LvArray`. If you want to move these out of `LvArray` we should discuss that. I would push for a kernel callable LA package separate from `GEOS` and `LvArray`, but that is just how I roll.

castelletto1:
> I feel like these would fit in better in LvArray. I mean it's already got the functions for solving 2x2 and 3x3 eigenvalue and inverse problems, why shouldn't the linear system solvers live there as well? Where ever they wind up there should be unit tests as well.

Yes, linear algebra operations, including tensor operations, should live in the same place.

> we should have a plan for a central location for all operations of this type. It was in `LvArray`. If you want to move these out of `LvArray` we should discuss that. I would push for a kernel callable LA package separate from `GEOS` and `LvArray`, but that is just how I roll.

I agree with @rrsettgast . Linear algebra and tensor operations should be a package independent of `LvArray` and of `GEOS`.

src/coreComponents/denseLinearAlgebra/denseLASolvers.hpp
castelletto1: I would check that denominator is nonzero before diving.

src/coreComponents/denseLinearAlgebra/denseLASolvers.hpp
castelletto1: We should make sure we don't divide by zero.

rrsettgast: If we divide by zero, then the game is over??....isn't it? I think we just let it fill a `NaN` and preferably throw a floating point exception.

rrsettgast:
> > I feel like these would fit in better in LvArray. I mean it's already got the functions for solving 2x2 and 3x3 eigenvalue and inverse problems, why shouldn't the linear system solvers live there as well? Where ever they wind up there should be unit tests as well.
> 
> Yes, linear algebra operations, including tensor operations, should live in the same place.
> 
> > we should have a plan for a central location for all operations of this type. It was in `LvArray`. If you want to move these out of `LvArray` we should discuss that. I would push for a kernel callable LA package separate from `GEOS` and `LvArray`, but that is just how I roll.
> 
> I agree with @rrsettgast . Linear algebra and tensor operations should be a package independent of `LvArray` and of `GEOS`.

isn't there a kernel callable set of dense linear algebra functions?

castelletto1:
> > > I feel like these would fit in better in LvArray. I mean it's already got the functions for solving 2x2 and 3x3 eigenvalue and inverse problems, why shouldn't the linear system solvers live there as well? Where ever they wind up there should be unit tests as well.
> > 
> > 
> > Yes, linear algebra operations, including tensor operations, should live in the same place.
> > > we should have a plan for a central location for all operations of this type. It was in `LvArray`. If you want to move these out of `LvArray` we should discuss that. I would push for a kernel callable LA package separate from `GEOS` and `LvArray`, but that is just how I roll.
> > 
> > 
> > I agree with @rrsettgast . Linear algebra and tensor operations should be a package independent of `LvArray` and of `GEOS`.
> 
> isn't there a kernel callable set of dense linear algebra functions?

There are many libraries. However, our need is to solve very small problems in each thread. @corbett5 do you have any experience with using Eigen in a device (CUDA) kernel?

src/coreComponents/denseLinearAlgebra/denseLASolvers.hpp
corbett5: There's also already a function for the determinant.
https://github.com/GEOS-DEV/LvArray/blob/69948929ca40acf490143fc020f72cd12d331f37/src/fixedSizeSquareMatrixOps.hpp#L38

corbett5:
Breaking out all the kernel callable tensor operations in LvArray into a new repo sounds fine to me.

@castelletto1 no I have not tried Eigen. I know that they have device compatible compile time fixed size 2x2 and 3x3 eigen decompositions, but I'm not sure for example if they have a general suite of device compatible fixed size routines. If they do I doubt they have the support for a static 2x2 and 3x3 symmetric format. Using Eigen would also lock us into the pattern of always performing operations on local variables, where as for example with `LvArray::tensorOps` you can pass in whatever mix of local variables and `LvArray::Array` objects you want.

castelletto1 (approved):
I understand @CusiniM needs this kernel for #3280. I propose we move forward incrementally, and focus on the refactoring we all seem to agree on in a future PR.

@CusiniM please make sure to add unit tests as required by @corbett5.

src/coreComponents/denseLinearAlgebra/unitTests/testDenseLASolvers.cpp
```diff
@@ -0,0 +1,96 @@
...
+  {
+    std::mt19937 generator( seed ); 
+    std::uniform_real_distribution< real64 > distribution( -10.0, 10.0 );
+    for( ptrdiff_t i=0; i<N; ++i )
+    {
+      solution[i] = distribution( generator );
+      for( ptrdiff_t j=0; j<N; ++j )
+      {
+        matrix[i][j] =  distribution( generator );
+      }
```
castelletto1:
```suggestion
      for( ptrdiff_t j=0; j<N; ++j )
      {
        matrix[i][j] =  distribution( generator );
      }
      matrix[i][i] += 1.0e2 * machinePrecision;
```

Extremely unlikely, but in principle a singular matrix could be generated. I would add a perturbation to the diagonal to ensure it is nonsingular.

src/coreComponents/denseLinearAlgebra/denseLASolvers.hpp
CusiniM: used it!

CusiniM:
Added unitTests. This should be good to go for now.

src/coreComponents/denseLinearAlgebra/denseLASolvers.hpp
```diff
@@ -0,0 +1,309 @@
...
+    }
+  }
+
+  // Step 2: Backward substitution
+  solveUpperTriangularSystem< N >( A, b, std::forward< SOL_TYPE >( x ) );
+
+  return true;
+}
+
+}; // internal namespace
```
CusiniM:
```suggestion
}; // details namespace
```

rrsettgast (approved):

src/coreComponents/denseLinearAlgebra/unitTests/testDenseLASolvers.cpp
```diff
@@ -0,0 +1,249 @@
...
+  void test_singularSystem()
+  {
+    SingularLinearSystem< size > LS( 2024 );
+
+    forAll< parallelDevicePolicy<> >( 1, [=] GEOS_HOST_DEVICE ( int )
+    {
+
+      real64 sol[size];
+      real64 matrix[size][size];
+      real64 rhs[size];
```
CusiniM:
```suggestion
      real64 sol[size]{};
      real64 matrix[size][size]{};
      real64 rhs[size]{};
```

PR #3044, opened 2024-03-15
Update integrated test baseline storage

cssherman:
geosPythonPackages PR: https://github.com/GEOS-DEV/geosPythonPackages/pull/14

PR #3289, opened 2024-08-20
Remove outdated notes on git modules from QuickStart.rst

MakisH:
The step to deinit the `integratedTests` is not needed anymore (and is a bit confusing) since https://github.com/GEOS-DEV/GEOS/pull/3044.

At the moment, a user would run into this error:

```shell
$ git submodule deinit integratedTests
error: pathspec 'integratedTests' did not match any file(s) known to git
```


CusiniM:
I have added these changes to https://github.com/GEOS-DEV/GEOS/pull/3296

PR #3290, opened 2024-08-20
FORGE Hydraulic Fracturing Simulation

frankfeifan:
This is a branch to update features in hydraulic fracturing simulations mainly for Utah FORGE.


CusiniM:
does this contain everything that is in https://github.com/GEOS-DEV/GEOS/pull/3310 ? To me it looks like this could be merged.

frankfeifan:
> does this contain everything that is in #3310 ? To me it looks like this could be merged.

Yes, but in addition to that, this PR also has some changes for THM model, e.g., making the perm model pressure-dependent in PorousSolid. We can merge #3310 I think.

PR #3294, opened 2024-08-21
fix: Bug with elastic wave propagator at higher orders

### LINKED ISSUE #3311: Busted array bounds in wave solvers
When tracking down a FPE in `inputFiles/wavePropagation/elas3D_Q3_abc_smoke.xml`, I ran with debug on. Thus with range checking on. The top of the stack trace is:
```
***** ERROR
***** LOCATION: /usr/WS2/settgast/Codes/geosx/GEOS/src/coreComponents/LvArray/src/indexing.hpp:209
***** Controlling expression (should be false): invalidIndices( dims, indices ... )
Invalid indices. dimensions = { 64, 3 } indices = { 64, 0 }

** StackTrace of 24 frames **
Frame 0: void LvArray::indexing::checkIndices<int, int, int>(int const*, int const, int const) 
Frame 1: int LvArray::ArrayView<double const, 2, 1, int, LvArray::ChaiBuffer>::linearIndex<int, int>(int const, int const) const 
Frame 2: double const& LvArray::ArrayView<double const, 2, 1, int, LvArray::ChaiBuffer>::operator()<int, int>(int, int) const 
Frame 3: geos::PreComputeSourcesAndReceivers::Compute3DSourceAndReceiverConstantsWithDAS<RAJA::PolicyBaseT<(RAJA::Policy)3, (RAJA::Pattern)1, (RAJA::Launch)0, (camp::resources::v1::Platform)1, RAJA::policy::omp::Parallel, RAJA::wrapper<RAJA::policy::omp::omp_for_schedule_exec<RAJA::policy::omp::Auto> > >, geos::finiteElement::Qk_Hexahedron_Lagrange_GaussLobatto<geos::finiteElement::LagrangeBasis3GL> const>(int, LvArray::ArrayOfArraysView<int const, int const, true, LvArray::ChaiBuffer>, LvArray::ArrayView<double const, 2, 1, int, LvArray::ChaiBuffer>, LvArray::ArrayView<long long const, 1, 0, int, LvArray::ChaiBuffer>, LvArray::ArrayView<long long const, 1, 0, int, LvArray::ChaiBuffer>, LvArray::ArrayOfArraysView<int const, int const, true, LvArray::ChaiBuffer>, LvArray::ArrayView<int const, 2, 1, int, LvArray::ChaiBuffer> const&, LvArray::ArrayView<int const, 1, 0, int, LvArray::ChaiBuffer>, LvArray::ArrayView<int const, 2, 1, int, LvArray::ChaiBuffer> const&, LvArray::ArrayView<int const, 2, 1, int, LvArray::ChaiBuffer>, LvArray::ArrayView<double const, 2, 1, int, LvArray::ChaiBuffer> const&, LvArray::ArrayView<double const, 2, 1, int, LvArray::ChaiBuffer>, LvArray::ArrayView<int, 1, 0, int, LvArray::ChaiBuffer>, LvArray::ArrayView<int, 2, 1, int, LvArray::ChaiBuffer>, LvArray::ArrayView<double, 2, 1, int, LvArray::ChaiBuffer>, LvArray::ArrayView<double, 2, 1, int, LvArray::ChaiBuffer>, LvArray::ArrayView<double, 2, 1, int, LvArray::ChaiBuffer>, LvArray::ArrayView<double const, 2, 1, int, LvArray::ChaiBuffer>, LvArray::ArrayView<int, 1, 0, int, LvArray::ChaiBuffer>, LvArray::ArrayView<int, 2, 1, int, LvArray::ChaiBuffer>, LvArray::ArrayView<double, 2, 1, int, LvArray::ChaiBuffer>, LvArray::ArrayView<float, 2, 1, int, LvArray::ChaiBuffer>, double, float, float, int, geos::WaveSolverUtils::DASType, int, LvArray::ArrayView<double const, 2, 1, int, LvArray::ChaiBuffer>, geos::Tensor<double, 3>, geos::Tensor<double, 6>)::{lambda(int)#1}::operator()(int) const 
```

This error occurs on 
https://github.com/GEOS-DEV/GEOS/blob/773fcff1b5a51ace652962c8787d565ee1f570e8/src/coreComponents/physicsSolvers/wavePropagation/shared/PrecomputeSourcesAndReceiversKernel.hpp#L469

the code is:
```c++
          for( localIndex a=0; a< numNodesPerElem; ++a )
          {
            for( localIndex i=0; i<3; ++i )
            {
              xLocal[a][i] = baseNodeCoords( elemsToNodes( k, a ), i );
            }
          }
```

with `a=1, i=3`, the error is on the indexing of `baseNodeCoords`, which has dimensions of `(64,3)`.
The index being returned from `elemsToNodes` is `64`, which is out of bounds.

sframba:
This PR solves a bug introduced by PR #3202 in the case of the elastic solver at higher orders. The standard `calcGradN` method of Gauss-Lobatto elements requires the coordinates of all the nodes in the mesh element. These nodes are no longer required for point location due to #3202, so only the mesh corners are passed to `PrecomputeSourceAndReceiverTerm`. However in the case of the elastic solver at higher orders, this interferes with `calcGradN`.
To correct this, a specialized version of the method, `calcGradNWithCorners`, is introduced and used. This method only requires the mesh corners instead of all the element nodes, fixind the issue.

resolves #3311


src/coreComponents/finiteElement/elementFormulations/Qk_Hexahedron_Lagrange_GaussLobatto.hpp
```diff
@@ -923,6 +968,57 @@ calcGradN( localIndex const q,
...
+GEOS_HOST_DEVICE
+GEOS_FORCE_INLINE
+real64
+Qk_Hexahedron_Lagrange_GaussLobatto< GL_BASIS >::calcGradNWithCorners( real64 const (&coords)[3],
+                                                                       real64 const (&X)[8][3],
+                                                                       real64 (& gradN)[numNodes][3] )
+{
+  real64 J[3][3] = {{0}};
+
+  jacobianTransformation( coords, X, J );
```
acitrain: I think you should use the other version of jacobianTransformation here, the one with qa, qb and qc.
Because if not the compilation crash because the compiler tries to put the value of numNodes to every possible values depending the orders (see the crahs inside the CIà

acitrain (changes requested):

src/coreComponents/finiteElement/elementFormulations/Qk_Hexahedron_Lagrange_GaussLobatto.hpp
sframba: You're right, I'll add a version for the calculation of the jacobian at an arbitrary point that does not require the location of high-order nodes

acitrain (approved):

sframba:
@rrsettgast  @castelletto1 @andrea-borio @CusiniM can any of you please give a quick review? This PR only impacts the wave propagators. Thanks!

rrsettgast (approved):

acitrain:
@rrsettgast I think this PR is also fixing  issue #3311, one of the main change inside this PR is the line you mentionned with the use of an other (and correct) elemToNodes map

Update: I tried to run the test cases mentionned on the issue (which indeed crash on develop) using this branch and it don't throw any error, so I think that this PR indeed fix the issue

PR #2738, opened 2023-10-04
generate code coverage reports

tbeltzun:
This is a WIP for generating code coverage reports for the GEOS code.

Early attempt to fix https://github.com/GEOS-DEV/GEOS/issues/2746.

I don't know if I'll have time to work on this (postdoc time is limited :/), but this might be a start.

Don't hesitate to push directly in this branch to contribute if you are interested in this feature.

https://github.com/GEOS-DEV/thirdPartyLibs/pull/255

https://github.com/GEOS-DEV/LvArray/pull/312

https://github.com/GEOS-DEV/hdf5_interface/pull/12

resolves #2746

PR #3300, opened 2024-08-22
fix: reenable floating point exceptions

### LINKED ISSUE #3281: Improved handling of NaNs

Warning message when NaNs are found in GEOS, removing the ambiguity over the source of the issue (linear solver or GEOS)

The following leads to a crash in hypre, but the problem is on the linear system assembled in GEOS.
```
        ( Rflow ) = ( 2.18e-01 )        ( Rsolid ) = ( -nan )        ( R ) = ( -nan )
```

A warning message like:
```
        ( Rflow ) = ( 2.18e-01 )        ( Rsolid ) = ( -nan )        ( R ) = ( -nan )

*********************************************************************
*                    WARNING! WARNING! WARNING!                     *
*                                                                   *
*  The assembled sparse linear system contains invalid numbers      *
*  (NaNs). This may cause the solver to fail. Please check your     *
*  input values, boundary conditions, or model setup for potential  *
*  issues that might lead to these errors.                          *
*********************************************************************

```

rrsettgast:
This PR:

- [x] reenables floating point exceptions, which were disabled (accidentally??) by @tbeltzun in #2738 
- [x] ....and then fix the FPE's that occur in the code now that we are checking.

https://github.com/GEOS-DEV/LvArray/pull/327
https://github.com/GEOS-DEV/thirdPartyLibs/pull/279

resolves #3281


victorapm (approved):

paveltomin (approved):

CusiniM (approved):
src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/CO2SolubilityDuanSun.cpp L68 also needs a small fix.

host-configs/LLNL/ruby-clang-14.cmake
```diff
@@ -2,5 +2,7 @@ include(${CMAKE_CURRENT_LIST_DIR}/../../src/coreComponents/LvArray/host-configs/
 
 # MPI
 set(MPI_HOME /usr/tce/packages/mvapich2/mvapich2-2.3.7-clang-14.0.6-magic CACHE PATH "")
+#set( CMAKE_CXX_FLAGS "-march=x86-64-v2 -mno-avx512f" CACHE STRING "" FORCE)
```
corbett5, line 6: Do we want to keep these options commented out? If so maybe add a comment saying why they're there.

wrtobin (approved):

src/coreComponents/mainInterface/ProblemManager.cpp
```diff
@@ -139,6 +139,8 @@ ProblemManager::ProblemManager( conduit::Node & root ):
     setRestartFlags( RestartFlags::WRITE ).
     setDescription( "Whether to disallow using pinned memory allocations for MPI communication buffers." );
 
+  chai::ArrayManager::getInstance()->disableCallbacks();
```
wrtobin, line 142: why is this needed?

corbett5, line 142: Won't this globally disable move logging? I thought this was a command line option?

corbett5 (approved):

src/coreComponents/mainInterface/ProblemManager.cpp
rrsettgast: I wanted this off by default for all the unitTests.

host-configs/LLNL/ruby-clang-14.cmake
rrsettgast: I can add a comment. Someone is going to have to come back and investigate.

src/coreComponents/mainInterface/ProblemManager.cpp
rrsettgast: It is...but the command line option is executed after this, but isn't called for any unitTests. Putting this here turns it off for the unit tests, as it is quite annoying for those. I assume it can be enabled after being disabled? Perhaps that is a bad assumption?

host-configs/LLNL/dane-gcc-12.cmake
```diff
@@ -6,4 +6,6 @@ set(MPI_HOME /usr/tce/packages/mvapich2/mvapich2-2.3.7-gcc-12.1.1-magic CACHE PA
 # ATS
 set(ATS_ARGUMENTS "--machine slurm112"  CACHE STRING "")
 
+#set( CMAKE_CXX_FLAGS "-march=x86-64-v2 -mno-avx512f" CACHE STRING "" FORCE)
```
rrsettgast, line 10:
```suggestion
# This is here to note the required flags for using valgrind. These will have to be propagated to the TPL's
#set( CMAKE_CXX_FLAGS "-march=x86-64-v2 -mno-avx512f" CACHE STRING "" FORCE)
```

host-configs/LLNL/ruby-clang-14.cmake
rrsettgast, line 6:
```suggestion
# This is here to note the required flags for using valgrind. These will have to be propagated to the TPL's
#set( CMAKE_CXX_FLAGS "-march=x86-64-v2 -mno-avx512f" CACHE STRING "" FORCE)
```

host-configs/LLNL/ruby-clang-14.cmake
```diff
@@ -2,5 +2,7 @@ include(${CMAKE_CURRENT_LIST_DIR}/../../src/coreComponents/LvArray/host-configs/
...
+#set( CMAKE_CXX_FLAGS "-g -O2 -fno-omit-frame-pointer -fsanitize=address" CACHE STRING "" FORCE)
```
rrsettgast, line 8:
```suggestion
# This is what is required for using address sanitizer. This should be put into GeosxOptions.cmake when it is incorporated into the build options.
#set( CMAKE_CXX_FLAGS "-g -O2 -fno-omit-frame-pointer -fsanitize=address" CACHE STRING "" FORCE)
```

host-configs/LLNL/ruby-clang-14.cmake
rrsettgast: comments added.

src/coreComponents/mainInterface/ProblemManager.cpp
corbett5: I'm sure you're right as you mentioned in the other comment that it can be re-enabled if desired.

paveltomin:
CC @drmichaeltcvx

PR #3302, opened 2024-08-23
fix: Fix restart issue when running HydroFracture cases.

### LINKED ISSUE #3286: Restart issue for HydroFractureSolver
**Describe the bug**
I recently found an issue in restart when running the `HydroFractureSolver`. The following screenshot shows the meshes around the fracture for `pennyShapedToughnessDominated_smoke`. The left shows the non-restart initial run, and the right shows the restart run at the same time. There is apparently an issue in node ordering in the restart run. Here I only show the `CellElementRegion`, and looks like `SurfaceElementRegion` is fine. 
Screenshot 2024-08-19 at 10 01 20 AM

This issue can also be seen in the following screenshots of logs at the first timestep of restart. The left shows the non-restart initial run, and the right shows the restart run at the same time. Apparently the apertures calculated are different between two cases. Also, the convergence is affected in the following solve steps. In some more complicated problems I'm running, we could eventually get non-convergence. 
Screenshot 2024-08-19 at 10 21 58 AM

The above result is for the case with partition of 2 in the crack normal direction (x in this case), but if we change the partition to 1 in the crack normal direction, it seems the issue is gone. See the following screenshot where the left is the restart run for the case with `-x 2 -y 2 -z 2` , while the right is the restart run for the case with `-x 1 -y 2 -z 2`. However, I don't think it permanently solves the issue. 
Screenshot 2024-08-19 at 10 02 42 AM

**Platform:**
- Machine: LLNL/dane & ruby
- Compiler: gcc 12.1.1 & clang 14.0.6

**Additional context**
From the logs as shown above, we can see the logs of the nonlinear solver and the linear solver are identical between the non-restart and the restart case before fracture propagation, but the values of aperture become different. So the issue should happen between `nonlinearImplicitStep` and the field initialize and update for new fractures. After some testing, I think the issue probably happens in `SurfaceGenerator::solverStep`.

frankfeifan:
This PR fixes issue #3286.

We also reduce the size of the following tests and turn on the `restartchecks`.

- [x]  `kgdToughnessDominated_smoke`
- [x]  `kgdViscosityDominated_smoke`
- [x]  `pennyShapedToughnessDominated_smoke`
- [x]  `pennyShapedViscosityDominated_smoke`
- [x]  `pknViscosityDominated_smoke`


src/coreComponents/mesh/ObjectManagerBase.hpp
```diff
@@ -685,6 +685,10 @@ class ObjectManagerBase : public dataRepository::Group
     /// @return String key to the local->global map
     static constexpr char const * localToGlobalMapString() { return "localToGlobalMap"; }
 
+    static constexpr char const * localMaxGlobalIndexString() { return "localMaxGlobalIndex"; }
+
+    static constexpr char const * maxGlobalIndexString() { return "maxGlobalIndex"; }
+
```
rrsettgast:
```suggestion
    /// @return String key for m_localMaxGlobalIndexString
    static constexpr char const * localMaxGlobalIndexString() { return "localMaxGlobalIndex"; }

    /// @return String key for m_maxGlobalIndexString
    static constexpr char const * maxGlobalIndexString() { return "maxGlobalIndex"; }

```

rrsettgast (approved):

CusiniM (approved):

jhuang2601 (approved):

.integrated_tests.yaml
```diff
@@ -1,7 +1,7 @@
 baselines:
   bucket: geosx
-  baseline: integratedTests/baseline_integratedTests-pr3283-6949-4ee437f
+  baseline: integratedTests/baseline_integratedTests-pr3302-7040-8832703
```
```suggestion
  baseline: integratedTests/baseline_integratedTests-pr3302-7064-57c6f3c
```

cssherman (approved):

src/CMakeLists.txt
```diff
@@ -222,8 +222,8 @@ install( FILES ${CMAKE_BINARY_DIR}/schema.xsd
 ################################
 # Add python environment setup
 ################################
-# message(WARNING "Temporarily changing the geosPythonBranch to feature/packaging")
-# set(GEOS_PYTHON_PACKAGES_BRANCH "feature/packaging" CACHE STRING "" FORCE)
+message(WARNING "Temporarily changing the geosPythonBranch to bugfix/sherman/fixNumpyProd")
+set(GEOS_PYTHON_PACKAGES_BRANCH "bugfix/sherman/fixNumpyProd" CACHE STRING "" FORCE)
```
CusiniM:
```suggestion
# message(WARNING "Temporarily changing the geosPythonBranch to bugfix/sherman/fixNumpyProd")
# set(GEOS_PYTHON_PACKAGES_BRANCH "bugfix/sherman/fixNumpyProd" CACHE STRING "" FORCE)
```

src/CMakeLists.txt
```diff
@@ -222,8 +222,8 @@ install( FILES ${CMAKE_BINARY_DIR}/schema.xsd
 ################################
 # Add python environment setup
 ################################
-# message(WARNING "Temporarily changing the geosPythonBranch to feature/packaging")
-# set(GEOS_PYTHON_PACKAGES_BRANCH "feature/packaging" CACHE STRING "" FORCE)
+# message(WARNING "Temporarily changing the geosPythonBranch to bugfix/sherman/fixNumpyProd")
+# set(GEOS_PYTHON_PACKAGES_BRANCH "bugfix/sherman/fixNumpyProd" CACHE STRING "" FORCE)
```
cssherman, line 226: This branch is now merged, so you can comment these out

CusiniM: they are commented.  I think we can leave the commands there for the next time we need them.

PR #3303, opened 2024-08-23
feat: add log at the end of the timestep with substeps info.

CusiniM:
this makes extracting info from the SPE11 logs a little easier. 

It prints something like this:

````
------------------------- TIMESTEP END -------------------------
    - Cycle:      28
    - N substeps: 2
    - dt:  00h00m02s (2.5 s)  00h00m02s (2.5 s)
--------------------------------------------------------------------
````

at the end of each timestep.


paveltomin (approved):

jafranc (approved):

CusiniM:
shall I push this forward so that we can use it for the spe11 runs?

rrsettgast (approved):

dkachuma (approved):

Issue #3305, opened 2024-08-27
Unable to run Geos on Pangea 3 with shared libraries


acitrain:
**Describe the bug**
Since the merge of PR #3282 we are unable to run GEOS on Pangea3 due to issue on a powerpc architecture to run with shared libraries. We get the follopwing error at the execution: 

geos: error while loading shared libraries: libmainInterface.so: cannot open shared object file: No such file or directory

And sometimes we are unable to compile (error at the link edition) 

**Platform (please complete the following information):**
- Machine Pangea 3
- Compiler gcc/8.4.1

Algiane:
Only the `pygeosx.so` shared library is installed in the `$CMAKE_INSTALL_PREFIX/lib` directory so the other shared libraries against which is built pygeosx are not found.
It can be solved:
  - by adding the path toward the other libs to the `LD_LIBRARY_PATH` variable (so the build directory). It is not clean and not usual as it means that the build directory cannot be deleted without breaking the runtime;
   - by installing all the shared libs in the same location (usual way)

rrsettgast:
The install command was accidentally removed. I will create a quick patch.

acitrain:
Thanks Randy!

I'll close my PR linked with this issue when the patch will be operational

PR #3306, opened 2024-08-27
fix: Disable shared library build on Pangea 3

acitrain:
This PR fix issue #3305 

As for now we are unable to run with shared library on Pangea 3 (powerpc architecture) and we don't have a stable solution for that (as it takes time and it is not a main prority for us)  we disable here the building with shared libraries inside the host-config for Pangea 3 

Here is the error we get : 

 geos: error while loading shared libraries: libmainInterface.so: cannot open shared object file: No such file or directory


Algiane (approved):

CusiniM:
I am guessing we can close this since https://github.com/GEOS-DEV/GEOS/pull/3307 fixes the same issue ?

PR #3282, opened 2024-08-14
build: Remove remaining circular link dependencies

rrsettgast:
This PR seeks to remove remaining circular **link** dependencies.

- [x] Move `OutputBase::getOutputDirectory()`  to a lower level component to avoid dependencies on `fileIO`
- [x] Move constitutive drivers to their own component (`constitutiveDrivers`).
- [x] remove reliance on `ProblemManager::getPhysicsSolverManager()` functions and insert hard coded `getGroup( "Solvers" )` function. Not optimal, but gets rid of the circular dependency.
- [x] reintruduce shared library option
- [x] remove shared `geosx_core` created using `blt_combine_static_libraries`
- [x] remove `GEOS_LINK_PREPEND_FLAG` and `GEOS_LINK_POSTPEND_FLAG` in favor of cmake `$<LINK_LIBRARY:WHOLE_ARCHIVE,${lib}>` option.
- [x] update all paths in documentation as a result of creation of `constitutiveDrivers` module.
- [x] cleanup unit test dependencies

Here is the dependency tree. Note that this doesn't resolve circular header dependencies.
dependency

PR #3307, opened 2024-08-27
fix: add install commands to component libraries

### LINKED ISSUE #3305: Unable to run Geos on Pangea 3 with shared libraries

rrsettgast:
In #3282 I failed to install the shared libraries. This PR adds the required install command.

resolves #3305

### DIFF
--- a/src/coreComponents/codingUtilities/CMakeLists.txt
+++ b/src/coreComponents/codingUtilities/CMakeLists.txt
@@ -36,6 +36,8 @@ set_source_files_properties( Parsing.cpp PROPERTIES LANGUAGE CXX )
 
 target_include_directories( codingUtilities PUBLIC ${CMAKE_SOURCE_DIR}/coreComponents )
 
+install( TARGETS codingUtilities LIBRARY DESTINATION ${CMAKE_INSTALL_PREFIX}/lib )
+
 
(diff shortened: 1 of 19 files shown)


victorapm (approved):

paveltomin (approved):

CusiniM (approved):

Algiane:
We still have a blind spot in the ci: to avoid making a similar mistake again, shouldn't we test that the geos installation works somewhere in the ci.
I haven't look at the implementation of the integrated tests:
  -  do they call the `geos` runtime ?
  - if yes, I guess that the good way is to call the runtime located inside the install dir rather than the one of the binary dir?
  - if it is already the case: shouldn't we try to understand why the ci was not failing and the reproduce the setting  that made the non-regression tests of Pangea fail ?
  
See PR #3313

paveltomin:
CC @drmichaeltcvx

PR #3308, opened 2024-08-27
fix: Small fix

paveltomin:


castelletto1 (approved):

CusiniM (approved):

ryar9534 (approved):

Guotong-Ren (approved):

rrsettgast (approved):

Algiane:
Hi,

Looking only at the PR, it's hard to be sure what the PR corrects: maybe a more descriptive PR title or a very short description could help?

Also, the erroneous lines where not covered before the PR and are still not after. To help people in the future to understand what has been done, what it solves and to test it easily in case of doubt, you may:

   1. commit a relevant test that fails because the 0 row of localMatrix is not accessed and has to be;
   2. commit the fix that solves the test.

This way of working (which follows the philosophy of the Test Driven Development method) is very useful if you think that your code will still be used and developped in few years.

PR #3310, opened 2024-08-28
feat: Make the rock toughness scalable with the fracture size and test anisotropic toughness

frankfeifan:
This PR is to add and test features of rock toughness.

### Scalable rock toughness
a new feature that makes the rock toughness scalable with the fracture size if the hydraulic fracture is very large. The underlying mechanism is that the cohesive region of rock fractures may scale up with the fracture size, thereby increasing the critical fracture energy and thus toughness. The specific formulation is

`KIC = KIC_base*(1 + scalingFactor*sqrt(L))` where `L` is the distance to the fracture origin. 

Here are the changes in this PR:
- [x] Change `rockToughness` to `baseRockToughness`, and add `toughnessScalingFactor` and `fractureOrigin` in `SurfaceGenerator`. `fractureOrigin` is manually input now, but can be removed later if we can find the fracture origin from the initial fracture geometry. 
- [ ] Plot `K_IC` by making it a field variable in `fractureSubRegion`,so that one can visualize the toughness in vtk. 
- [x] Add a simple test to demonstrate the feature

Relevant references: [Scholz, 2010](https://www.sciencedirect.com/science/article/pii/S019181411000163X?ref=pdf_download&fr=RR-2&rr=8ba5ad317a042289 ), [McClure, 2023](https://pangea.stanford.edu/ERE/db/GeoConf/papers/SGW/2023/Mcclure.pdf)

### Anisotropic toughness
Existing `hydrofracture` solver has the capability to assign anisotropic toughness, but looks like there is no relevant test. An example is added to test this feature. Also, some changes may be needed to make this feature easier to use for users.
- [x] Add an example to test anisotropic rock toughness


src/coreComponents/physicsSolvers/surfaceGeneration/SurfaceGenerator.hpp
```diff
@@ -521,7 +526,10 @@ class SurfaceGenerator : public SolverBase
 
     //TODO: rock toughness should be a material parameter, and we need to make rock toughness to KIC a constitutive
     // relation.
-    constexpr static char const * rockToughnessString() { return "rockToughness"; }
+    constexpr static char const * baseRockToughnessString() { return "baseRockToughness"; }
```
CusiniM: maybe `initialRockToughness` ?

inputFiles/hydraulicFracturing/pennyShapedAnisotropicToughness_base.xml
```diff
@@ -0,0 +1,228 @@
...
+      setNames="{ all }"
+      objectPath="ElementRegions"
+      fieldName="K_IC_11"
+      scale="6e6"/>
+    
+    <FieldSpecification
+      name="KIC_21"
+      initialCondition="1"
+      setNames="{ all }"
+      objectPath="ElementRegions"
```
Guotong-Ren, line 179: i guess the anisotropic KI will not affect the propagation direction (still perpendicular to the minimum principal stress) but the propagation speed?

CusiniM: yes, correct. The fracture will remain in the same plane. It changes the propagation speed in the two directions though. It's quite useful to match field data.

CusiniM (approved):
It would be nice to encapsulate everything related to the rock toughness into a constitutive model and completely remove the `rockToughness` as a parameter of the `SurfaceGenerator`. However, I am okay with merging this first and then introduce:

- 1 a PR to  create an appropriate constitutive model. I think the KIC computations could be moved into a costitutive model too.
- 2 a PR to replace the current calculations with the new constitutive model.

CusiniM:
>     * [x]  Add an example to test anisotropic rock toughness

Adding a test is good but I would take the simple example we used in the FORGE report (with the idealized stress profile) an make an advanced example in readthedocs from it.

rrsettgast:
@frankfeifan Questions about adding a size based toughness modification:

1. The toughness increases with increase in size, meaning it is harder to propagate the fracture for a homogeneous body as the size of the fracture grows. I am familiar with arguments based on field scale data, but is this behavior observed at lab scale samples for intact rock...or even some other material?
2. Is this factor a way to account for plastic slip in the rock surrounding the fracture? 
3. For non-penny shaped fracture, are there observations that support different scaling factors in different orientations?

frankfeifan:
> @frankfeifan Questions about adding a size based toughness modification:
> 
> 1. The toughness increases with increase in size, meaning it is harder to propagate the fracture for a homogeneous body as the size of the fracture grows. I am familiar with arguments based on field scale data, but is this behavior observed at lab scale samples for intact rock...or even some other material?
> 2. Is this factor a way to account for plastic slip in the rock surrounding the fracture?
> 3. For non-penny shaped fracture, are there observations that support different scaling factors in different orientations?

@rrsettgast Please find my replies below: 
1. I did some literature study. There are some lab observations about this scale dependence using various experimental methods (e.g., Brazilian test), although not in the context of hydraulic fracturing: https://link.springer.com/article/10.1007/s00603-021-02468-1, https://www.sciencedirect.com/science/article/pii/S1674775522002293#bib91 and references therein. 

2. It’s highly possible and the dissipation associated with plastic slip may be scaled with the fracture size. Actually, one explanation in literature is based on the fracture processing zone (FPZ) where micro-cracking and plastic deformation happen. Consider a cohesive zone model for this FPZ, the area bounded by the traction-crack opening curve is the critical fracture energy (Gc). Many field observations indicate that the opening/crack separation is proportional to the fracture size (L). If we consider the tensile strength keeps constant, then Gc is proportional to L, so K_Ic is proportional to sqrt(L). 
 
3. This is a good point. I didn't see relevant evidence unfortunately. In other HF models, they use a unique scaling factor for the entire fracture and calculate it using either the larger or smaller of the fracture length and height. IMO, if this apparent scale (size) dependence is due to FPZ at the front of the crack tip, then maybe it makes more sense to compute the scaling factor on each fracture cell at the tip individually (what we do now)?

rrsettgast (approved):
We should make a note to follow up on different methods for calculating the factor.

PR #3312, opened 2024-08-29
ci: Enable bounds checking in CI runs

rrsettgast:
This PR enables bounds checking in CI testing runs.

https://github.com/GEOS-DEV/LvArray/pull/344


corbett5:
So I'm pretty sure I know why the the `exampleSortedArray` test is failing. These are the failing statements

```c++
  // Attempt to insert unsorted values.
  int const unsortedInsert[ 2 ] = { 4, 0 };
  EXPECT_DEATH_IF_SUPPORTED( set.insert( unsortedInsert, unsortedInsert + 2 ), "" );

  // Attempt to insert nonUnique values.
  int const notUnique[ 2 ] = { 5, 5 };
  EXPECT_DEATH_IF_SUPPORTED( set.insert( notUnique, notUnique + 2 ), "" );

  // Attempt to remove unsorted values.
  int const unsortedRemove[ 2 ] = { 4, 2 };
  EXPECT_DEATH_IF_SUPPORTED( set.remove( unsortedRemove, unsortedRemove + 2 ), "" );
```

But the statement that is intendend to catch these errors is

```c++
LVARRAY_ASSERT( isSortedUnique( first, last ) );
```

which is only enabled in debug builds. I think the fix should be to simply change these statements into `LVARRAY_ERROR_IF` if bounds checking is enabled. Yes it's not as fast as simple bounds checking but it's a small overhead compared to the cost of sorted insert. And it could really mess stuff up if not caught. 

@rrsettgast what do you think?

rrsettgast:
Sounds good to me

wrtobin (approved):

corbett5:
Not sure what to do about the remaining issue with the single build. The error is in `testStackArray` when the compiler warns that a stack array may be used uninitialized. But that's kind of the point, we don't want to be initializing the values and also in this case there is no uninitialized access.

`buf` here is a default constructed `StackBuffer` which means the values in it's c-array are unintialized. But the offending code method doesn't actually look at the values 

```c++
template< typename BUFFER >
inline LVARRAY_HOST_DEVICE CONSTEXPR_WITHOUT_BOUNDS_CHECK
void check( BUFFER const & buf, std::ptrdiff_t const size )
{
#ifdef LVARRAY_BOUNDS_CHECK
  LVARRAY_ERROR_IF_GT( 0, buf.capacity() );
  LVARRAY_ERROR_IF_GT( 0, size );
  LVARRAY_ERROR_IF_GT( size, buf.capacity() );
#else
  LVARRAY_DEBUG_VAR( buf );
  LVARRAY_DEBUG_VAR( size );
#endif
}
```

https://github.com/GEOS-DEV/GEOS/actions/runs/10623472714/job/29449977551?pr=3312#step:8:1276


Easiest fix is to simply use a pragma to disable this warning for this file when that compiler is used. But pretty ugly.

rrsettgast:
@corbett5 Sorry if I don't understand, but the "error: 'array' may be used uninitialized" is a because there is a potential of using the contents of the variable when they have not been initialized? In normal use, we would never want to do this...but in this test case we don't care. Isn't this solved by just initializing the `array` object?

corbett5:
It might solve it, you never know with these problems. But I'm pretty sure the error is coming from the creation of the LvArray::Array itself so logically if we fill in the values later it shouldn't make a difference.

CusiniM (approved):

.github/workflows/ci_tests.yml
```diff
@@ -269,10 +271,10 @@ jobs:
       CMAKE_BUILD_TYPE: Debug
       CODE_COVERAGE: true
       DOCKER_IMAGE_TAG: ${{ needs.is_not_draft_pull_request.outputs.DOCKER_IMAGE_TAG }}
-      DOCKER_REPOSITORY: geosx/ubuntu22.04-gcc11
+      DOCKER_REPOSITORY: geosx/ubuntu22.04-gcc12
```
CusiniM:
```suggestion
      DOCKER_REPOSITORY: geosx/ubuntu22.04-gcc11
```

.github/workflows/ci_tests.yml
```diff
@@ -269,10 +271,10 @@ jobs:
...
       ENABLE_HYPRE: ON
       ENABLE_TRILINOS: OFF
-      GCP_BUCKET: geosx/ubuntu22.04-gcc11
+      GCP_BUCKET: geosx/ubuntu22.04-gcc12
```
CusiniM:
```suggestion
      GCP_BUCKET: geosx/ubuntu22.04-gcc11
```

.github/workflows/ci_tests.yml
```diff
@@ -175,13 +175,14 @@ jobs:
             DOCKER_REPOSITORY: geosx/ubuntu22.04-gcc11
             ENABLE_HYPRE: ON
             ENABLE_TRILINOS: OFF
-            GCP_BUCKET: geosx/ubuntu22.04-gcc11
             BUILD_SHARED_LIBS: ON
+            ENABLE_LVARRAY_BOUNDS_CHECK: OFF
```
CusiniM, line 179:
```suggestion
            ENABLE_LVARRAY_BOUNDS_CHECK: OFF
            GCP_BUCKET: geosx/ubuntu22.04-gcc11
```

PR #3320, opened 2024-08-30
fix: Check for zero permeability in J-function cap pressure

paveltomin:
Otherwise division by zero leads to inf/nan


src/coreComponents/constitutive/capillaryPressure/JFunctionCapillaryPressure.cpp
```diff
@@ -250,6 +250,7 @@ void JFunctionCapillaryPressure::saveConvergedRockState( arrayView2d< real64 con
     {
       permeability = convergedPermeability[ei][0][2];
     }
+    GEOS_ERROR_IF( !isZero( permeability ), GEOS_FMT( "{}: zero permeability in J-function capillary pressure", getFullName() ));
```
dkachuma:
```suggestion
    GEOS_ERROR_IF( permeability < LvArray::NumericLimits<real64>::epsilon, GEOS_FMT( "{}: zero permeability in J-function capillary pressure", getFullName() ));
```
I would completely ban negative values as well.

dkachuma (approved):

src/coreComponents/constitutive/capillaryPressure/JFunctionCapillaryPressure.cpp
```diff
@@ -250,6 +250,7 @@ void JFunctionCapillaryPressure::saveConvergedRockState( arrayView2d< real64 con
     {
       permeability = convergedPermeability[ei][0][2];
     }
+    GEOS_ERROR_IF( permeability < LvArray::NumericLimits< real64 >::epsilon, GEOS_FMT( "{}: zero permeability in J-function capillary pressure", getFullName() ));
```
rrsettgast: @paveltomin We generally cannot have `GEOS_*` macros inside a device lambda.

CusiniM: this is on me. Isn't this calling `LVARRAY_ERROR` behind the scenes ? That one should work fine on device.  not sure about the FMT part though.

CusiniM: To be clear, when compiling on device I thought that `LV_ARRAY_ERROR_IF` had an empty implementation.

CusiniM: Here: https://github.com/GEOS-DEV/LvArray/blob/0a5f594934e428548917376aeb6d3691c3b9206b/src/Macros.hpp#L123

CusiniM: > Here: https://github.com/GEOS-DEV/LvArray/blob/0a5f594934e428548917376aeb6d3691c3b9206b/src/Macros.hpp#L123

so it will use `printf` and it will work but it will have an impact on performance. It's probably okay for this specific kernel.

rrsettgast: Well then...`GEOS_FMT` cannot be called aside a device lambda....or `getFullName` either.

paveltomin: I will pull it out, no problem
Let me move that PR aside for now

rrsettgast: I think you can just use a string literal and it will work.

paveltomin: ok, sounds good, done

CusiniM (approved):

PR #3322, opened 2024-08-30
feat: print out fracture state for contact model

jhuang2601:
This PR aids to output fracture state for contact model, which can be then visualized in paraview during fault stability analysis

image


matteofrigo5 (approved):

herve-gross (approved):

CusiniM (approved):

src/coreComponents/physicsSolvers/contact/ContactFields.hpp
```diff
@@ -134,8 +134,8 @@ DECLARE_FIELD( fractureState,
                "fractureState",
                array1d< integer >,
                FractureState::Stick,
-               NOPLOT,
-               NO_WRITE,
+               LEVEL_0,
+               WRITE_AND_READ,
```
CusiniM, line 138: if you remove this `WRITE_AND_READ` you won't need to rebaseline. However, if we want to use restarts with these fracture cases, I would say that we should keep this so that the fracture state is written out.

paveltomin: I think keeping it is a good idea, may help to better understand the changes

PR #3327, opened 2024-09-03
feat: Nonlinear Max Change Reporting to include cell location

tjb-ltk:
Added global in reporting of newton iter  maxpres/temp/compdens changes.  Global cell element index is reported in vtk output to enable finding areas impacting convergence.


src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.cpp
```diff
@@ -469,9 +469,12 @@ real64 CompositionalMultiphaseFVM::scalingForSystemSolution( DomainPartition & d
 
   string const dofKey = dofManager.getKey( viewKeyStruct::elemDofFieldString() );
   real64 scalingFactor = 1.0;
-  real64 maxDeltaPres = 0.0, maxDeltaCompDens = 0.0, maxDeltaTemp = 0.0;
   real64 minPresScalingFactor = 1.0, minCompDensScalingFactor = 1.0, minTempScalingFactor = 1.0;
 
+  std::vector< std::tuple< real64, globalIndex > > maxRegionDeltaPresLoc;
```
paveltomin: got a bit confused here - is there a real need to store all subregions values?

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.cpp
```diff
@@ -510,33 +514,52 @@ real64 CompositionalMultiphaseFVM::scalingForSystemSolution( DomainPartition & d
...
-  maxDeltaCompDens = MpiWrapper::max( maxDeltaCompDens );
   minPresScalingFactor = MpiWrapper::min( minPresScalingFactor );
   minCompDensScalingFactor = MpiWrapper::min( minCompDensScalingFactor );
 
   string const massUnit = m_useMass ? "kg/m3" : "mol/m3";
-  GEOS_LOG_LEVEL_RANK_0( 1, GEOS_FMT( "        {}: Max pressure change = {} Pa (before scaling)",
-                                      getName(), GEOS_FMT( "{:.{}f}", maxDeltaPres, 3 ) ) );
-  GEOS_LOG_LEVEL_RANK_0( 1, GEOS_FMT( "        {}: Max component density change = {} {} (before scaling)",
-                                      getName(), GEOS_FMT( "{:.{}f}", maxDeltaCompDens, 3 ), massUnit ) );
+  GEOS_LOG_LEVEL_RANK_0( 1, GEOS_FMT( "        {}: Max pressure change = {} Pa (before scaling) at global index {}",
```
paveltomin: might be a bit confusing for a user but up to you - i would use "cell" instead of "global index"

paveltomin (approved):

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.cpp
tjb-ltk: probably not but there is a subregion loop .. so ??

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.cpp
tjb-ltk: agreed

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.cpp
paveltomin: i was thinking something simple like that
```
if (subRegionData.localMaxDeltaPres < maxRegionDeltaPresLoc[1] )
  maxRegionDeltaPresLoc = {subRegionData.localMaxDeltaPres, localToGlobalMap[subRegionData.localMaxDeltaPresLoc};
}
```
?

tjb-ltk: I'll simplify... but thought the max on tuple with lambda function  was pretty cool...

paveltomin:
can we move this to merge queue?

paveltomin:
needs approval from code owner

src/coreComponents/common/MpiWrapper.hpp
```diff
@@ -1115,6 +1125,30 @@ void MpiWrapper::reduce( Span< T const > const src, Span< T > const dst, Reducti
...
+    l_rcv_buffer.resize( numProcs );
+  }
+  MPI_Gather( &tuple, sizeof(tuple_type), MPI_BYTE, &l_rcv_buffer[0], sizeof(tuple_type), MPI_BYTE, 0, comm );
+
+
+  if( myRank == 0 )
+  {
+    tuple= *max_element( begin( l_rcv_buffer ), end( l_rcv_buffer ), []( auto & lhs, auto & rhs ) -> integer {return std::get< 0 >( lhs )  <  std::get< 0 >( rhs ); } );
+  }
+  return tuple;
```
CusiniM: I think you should add `broadcast` step. I understand that you don't need it for this specific task but since you are creating a generic function I would expect it to also perform the broadcast.

src/coreComponents/common/MpiWrapper.hpp
```diff
@@ -1115,6 +1125,30 @@ void MpiWrapper::reduce( Span< T const > const src, Span< T > const dst, Reducti
...
+  // recieve buffer
+  std::vector< tuple_type > l_rcv_buffer;
+
+  int const myRank  = MpiWrapper::commRank( comm );
+  int const numProcs = MpiWrapper::commSize( comm );
+  if( myRank == 0 )
+  {
+    l_rcv_buffer.resize( numProcs );
+  }
+  MPI_Gather( &tuple, sizeof(tuple_type), MPI_BYTE, &l_rcv_buffer[0], sizeof(tuple_type), MPI_BYTE, 0, comm );
```
CusiniM: First, I would suggest only passing the buffer to rank 0 with something like:
```suggestion
  MPI_Gather( &tuple, sizeof(tuple_type), MPI_BYTE, myRank == 0 ? l_rcv_buffer.data() : nullptr, sizeof(tuple_type), MPI_BYTE, 0, comm );
```
second,

I think this is actually kind of problematic. I don't think a tuple is always stored as a contiguous block of memory which is what `MPI_Gather` expects since it's passed as MPI_BYTE. I believe that this can lead to weird behaviors if the data is not contiguous. You could copy it into a buffer with contiguous memory and use that to perform the gather but it is somewhat involved. Do we really need this max over tuples? 

@wrtobin and @rrsettgast can probably comment more on this.

CusiniM (commented):
I am not sure that the MPI max over the tuple is entirely correct as it is. In any case I don't know that we really need it. I feel we only care about the global max value and its location so we don't need to store the max per subregion IMO.

src/coreComponents/common/MpiWrapper.hpp
CusiniM: Also, this really expects a pair not a tuple and I think we should have some checks on the types contained in the tuple. I.e. The first type must be comparable.

rrsettgast: It is not guaranteed by the standard. The tuple should be explicitly packed/unpacked.

src/coreComponents/common/MpiWrapper.hpp
rrsettgast: Either broadcast, or change the name of the function.

rrsettgast (commented):
I wonder if it would be better to create a class for what you are storing in the `tuple` in this PR? That way it would be guaranteed contiguous, and any operations on the data in the `tuple` could be put into type safe functions

tjb-ltk:
I'm changing from tuple to struct...and will broadcast...

src/coreComponents/common/MpiWrapper.hpp
wrtobin: Yeah the types contained within the tuple would all need to be standard layout IIRC for this to be valid. Packing/unpacking is the easier lift than enforcing it through type trait checks.

tjb-ltk: submitted change to use struct for storage and serialization to/from byte string  for mpiallgather

src/coreComponents/common/MpiWrapper.hpp
```diff
@@ -608,6 +608,16 @@ struct MpiWrapper
...
   static void max( Span< T const > src, Span< T > dst, MPI_Comm comm = MPI_COMM_GEOS );
+
+
+  /**
+   * @brief Convenience function for MPI_Gather using a MPI_MAX operation on struct of value and location
+   * @brief Max is performed on value and location (global indice)
+   * @param[in] struct to send into the max gather.
+   * @return struct with max val and location
+   */
+  template< typename valLoc_type > static valLoc_type maxValLoc( valLoc_type l_valLoc, MPI_Comm comm = MPI_COMM_GEOS );
```
CusiniM:
```suggestion
  template< typename CONTAINER_TYPE > static CONTAINER_TYPE maxValLoc( CONTAINER_TYPE l_valLoc, MPI_Comm comm = MPI_COMM_GEOS );
```
might be better?

src/coreComponents/common/MpiWrapper.hpp
```diff
@@ -1115,6 +1125,47 @@ void MpiWrapper::reduce( Span< T const > const src, Span< T > const dst, Reducti
...
+  // Ensure members are not pointers
+  if constexpr ( std::is_pointer_v< decltype(&valLoc_type::value) >) {
+    GEOS_ERROR_IF( false, "Attempting to use pointer for maxValLoc comparison" );
+  }
+  if constexpr ( std::is_pointer_v< decltype(&valLoc_type::location) >) {
+    GEOS_ERROR_IF( false, "Attempting to use pointer for maxValLoc comparison" );
+  }
+  //These generate compile error
+  //GEOS_ERROR_IF(constexpr (std::is_pointer_v<decltype(&valLoc_type::value)>),"Attempting to use pointer for maxValLoc comparison" );
+  //GEOS_ERROR_IF(constexpr (std::is_pointer_v<decltype(&valLoc_type::location)>),"Attempting to use pointer for maxValLoc comparison" );
```
CusiniM:
```suggestion
```

src/coreComponents/common/MpiWrapper.hpp
```diff
@@ -1115,6 +1125,47 @@ void MpiWrapper::reduce( Span< T const > const src, Span< T > const dst, Reducti
...
+{
+  // Force struct to have only 2 data members
+  GEOS_ASSERT_EQ((sizeof(valLoc_type::value)+sizeof(valLoc_type::location)), sizeof(valLoc_type));
+  // Ensure members are not pointers
+  if constexpr ( std::is_pointer_v< decltype(&valLoc_type::value) >) {
+    GEOS_ERROR_IF( false, "Attempting to use pointer for maxValLoc comparison" );
+  }
+  if constexpr ( std::is_pointer_v< decltype(&valLoc_type::location) >) {
+    GEOS_ERROR_IF( false, "Attempting to use pointer for maxValLoc comparison" );
+  }
```
CusiniM:
```suggestion
  if constexpr ( std::is_pointer_v< decltype(&valLoc_type::value) > || std::is_pointer_v< decltype(&valLoc_type::location) > ) {
    GEOS_ERROR( "Attempting to use pointer for maxValLoc comparison" );
  }
```

src/coreComponents/common/MpiWrapper.hpp
```diff
@@ -1115,6 +1125,47 @@ void MpiWrapper::reduce( Span< T const > const src, Span< T > const dst, Reducti
...
+
+  MPI_Allgather( sendbuf, sizeof(valLoc_type), MPI_BYTE, recvbuf, sizeof(valLoc_type), MPI_BYTE, comm );
+
+  // Unpack buffer and find max
+  std::vector< valLoc_type > g_valLoc( numProcs );
+  for( int i = 0; i< numProcs; i++ )
+  {
+    std::memcpy( &g_valLoc[i], &recvbuf[sizeof(valLoc_type)*i], sizeof(valLoc_type));
+  }
+  valLoc_type maxValLoc= *max_element( begin( g_valLoc ), end( g_valLoc ), []( auto & lhs, auto & rhs ) -> integer {return lhs.value  <  rhs.value; } );
```
CusiniM: shouldn't this be:
```suggestion
  valLoc_type maxValLoc= *std::max_element( begin( g_valLoc ), end( g_valLoc ), []( auto & lhs, auto & rhs ) -> bool {return lhs.value  <  rhs.value; } );
```
to me it looks like `std::max_element` expects a bool from the comparison: https://en.cppreference.com/w/cpp/algorithm/max_element

src/coreComponents/common/MpiWrapper.hpp
```diff
@@ -1115,6 +1125,47 @@ void MpiWrapper::reduce( Span< T const > const src, Span< T > const dst, Reducti
...
+
+  delete[] sendbuf;
+  delete[] recvbuf;
```
CusiniM: can we use an `std:vector` of `unsigned char` ? That one would not require us to manually delete the pointers.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.hpp
```diff
@@ -174,6 +174,19 @@ class CompositionalMultiphaseFVM : public CompositionalMultiphaseBase
     Local            ///< Scale the Newton update locally (modifies the Newton direction)
   };
 
+  /**
+   * @brief Storage for value and element location, used to determine global max + location
+   */
+  template< typename value_type, typename index_type >
```
CusiniM:
```suggestion
  template< typename VALUE_TYPE, typename INDEX_TYPE >
```

CusiniM (approved):
Just a few cosmetic changes.

paveltomin:
still needs review/reviews from @MelReyCG, @corbett5, @rrsettgast, and/or @wrtobin

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.hpp
```diff
@@ -174,6 +174,19 @@ class CompositionalMultiphaseFVM : public CompositionalMultiphaseBase
...
+  struct valueLoc
+  {
+    valueLoc(){}
+    valueLoc( value_type val, index_type loc ): value( val ), location( loc ){}
+    value_type value;
+    index_type location;
```
CusiniM:
```suggestion
    VALUE_TYPE value;
    INDEX_TYPE location;
```

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.hpp
```diff
@@ -174,6 +174,19 @@ class CompositionalMultiphaseFVM : public CompositionalMultiphaseBase
     Local            ///< Scale the Newton update locally (modifies the Newton direction)
   };
 
+  /**
+   * @brief Storage for value and element location, used to determine global max + location
+   */
+  template< typename value_type, typename index_type >
+  struct valueLoc
```
CusiniM:
```suggestion
  struct ValueAndLocation
```

src/coreComponents/common/MpiWrapper.hpp
tjb-ltk: already done

src/coreComponents/common/MpiWrapper.hpp
tjb-ltk: already done

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.hpp
tjb-ltk: already done

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.hpp
tjb-ltk: submitted

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.hpp
tjb-ltk: submitted

CusiniM:
Missing a review. @rrsettgast can you approve it?

paveltomin:
Waiting on code owner review from @MelReyCG, @corbett5, @rrsettgast, and/or @wrtobin.

src/coreComponents/common/MpiWrapper.hpp
```diff
@@ -1115,7 +1125,32 @@ void MpiWrapper::reduce( Span< T const > const src, Span< T > const dst, Reducti
   reduce( src.data(), dst.data(), LvArray::integerConversion< int >( src.size() ), getMpiOp( op ), root, comm );
 }
 
+// Mpi helper function to return  struct containing the max value and location across ranks
+template< typename T >
+T MpiWrapper::maxValLoc( T localValueLocation, MPI_Comm comm )
```
corbett5, line 1130: There's certainly a more efficient way to do this, but my guess is this isn't in a performance critical section so it doesn't matter.

corbett5 (approved):

PR #3330, opened 2024-09-05
Design props for directional hysteresis

jafranc:
Design questions raised for implementing directional relative permeabilities and the number of directions


src/coreComponents/constitutive/relativePermeability/TableRelativePermeabilityHysteresis.cpp
```diff
@@ -487,6 +500,8 @@ void TableRelativePermeabilityHysteresis::computeLandCoefficient()
   integer ipNonWetting = 0;
 
   integer const numPhases = m_phaseNames.size();
+  integer const numDir = m_drainageWettingNonWettingRelPermTableNames.size(0);
```
jafranc: in order no to have to repeat every time the capture of `numDir`, maybe make is a member variable of the class and assign and check consistency in the `postInputInitialization()`

src/coreComponents/constitutive/relativePermeability/TableRelativePermeabilityHysteresis.cpp
```diff
@@ -166,7 +166,10 @@ void TableRelativePermeabilityHysteresis::postInputInitialization()
 
   using IPT = TableRelativePermeabilityHysteresis::ImbibitionPhasePairPhaseType;
 
+
```
jafranc, line 170: Maybe assign here member variable `numDir` and check that it is consistent across all tables

src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseFVMKernelUtilities.hpp
```diff
@@ -354,10 +357,23 @@ struct PPUPhaseFlux
     localIndex const esr_up = sesri[k_up];
     localIndex const ei_up = sei[k_up];
 
-    auto const correction_term = LvArray::math::abs( faceNormal[0] + faceNormal[1] + faceNormal[2] + 1e-25 );
-    real64 const mobility = LvArray::math::abs( LvArray::tensorOps::AiBi< 3 >( phaseMob[er_up][esr_up][ei_up][ip], faceNormal ) )
-                            /correction_term;
+    integer const numDir = 3; // phaseMob[er_up][esr_up][ei_up][ip].size();
```
jafranc: this `numDir` could indeed be inferred from `phaseMob` number of elements.

src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseFVMKernelUtilities.hpp
```diff
@@ -503,17 +549,38 @@ struct C1PPUPhaseFlux
...
+
+    else
     {
       real64 const dMob_dP = LvArray::tensorOps::AiBi< 3 >( dPhaseMob[seri[0]][sesri[0]][sei[0]][ip][Deriv::dP], faceNormal );
       dPhaseFlux_dP[0] += Ttrans * potGrad * dMob_dP;
     }
 
     // dC
+
+    if (numDir ==1)
```
jafranc: As we are to that many if-branching, I would suggest investing time in refactoring into two functions and one and only `if (numDir==1)`

src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseFVMKernels.hpp
```diff
@@ -2043,23 +2045,29 @@ class DirichletFaceBasedAssemblyKernel : public FaceBasedAssemblyKernel< NUM_COM
       // *** upwinding ***
       // Step 3.4: upwinding based on the sign of the phase potential gradient
       // It is easier to hard-code the if/else because it is difficult to address elem and face variables in a uniform way
-
+      integer const numDir = 3; //m_phaseMob.size(4);
       if( potDif >= 0 )              // the element is upstream
       {
-
+        if (numDir == 1)
```
jafranc, line 2051: this branch is inactive if l.2048 stays

src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseFVMKernels.hpp
```diff
@@ -147,7 +147,7 @@ class PhaseMobilityKernel : public isothermalCompositionalMultiphaseBaseKernels:
     real64 dVisc_dC[numComp]{};
 
 
-    for( int dir = 0; dir < 3; ++dir )
+    for( int dir = 0; dir < 1; ++dir )
```
jafranc: why is it forced to 1 here ?

jafranc (commented):
Few notes.
On general level, it would be easier to factored out the two cases `if(numDir==1)` and `if(numDir==3)` into functions and let the already defined functions to just do the dispatch.
It would be way easier to read and maintain.

src/coreComponents/constitutive/relativePermeability/TableRelativePermeabilityHysteresis.cpp
Rclarson: Yes, that could work.  Also need to add capability to use whichever input string is present out of optional inputs.

src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseFVMKernelUtilities.hpp
Rclarson: I'll take a look at applying these changes

PR #3331, opened 2024-09-05
chore: increase minimum required cmake version.

CusiniM:
We seem to be getting a CMake error due to a call inside BLT when using a version lower than 3.26 (in reality I have only tried 3.23 and 3.26 so I am not sure about what happens in the middle). 

````
CMake Error at cmake/blt/cmake/BLTPrivateMacros.cmake:214 (target_link_libraries):
  Error evaluating generator expression:
    $<LINK_LIBRARY:WHOLE_ARCHIVE,>
  Expression did not evaluate to a known generator expression
  cmake/blt/cmake/BLTMacros.cmake:332 (blt_setup_target)
  ````

However, I don't see anything in the [BLT release notes](https://github.com/LLNL/blt/releases/tag/v0.6.2) suggesting that CMake 3.26 or higher is required....

### CODE REVIEW

**Comment by rrsettgast** [comment]
This is a cmake error. Not a blt error. The option was added around 3.23

**Comment by CusiniM** [comment]
> This is a cmake error. Not a blt error. The option was added around 3.23

But it's an error trigger by a BLT call if I read that correctly. I can't understand whether the higher cmake version is a strict requirement or not.

**Comment by CusiniM** [comment]
> > This is a cmake error. Not a blt error. The option was added around 3.23
> 
> But it's an error trigger by a BLT call if I read that correctly. I can't understand whether the higher cmake version is a strict requirement or not.

in other words, are we doing something incorrectly or do we just need to require a higher CMake version?

**Comment by rrsettgast** [comment]
https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#genex:LINK_LIBRARY

The `$<LINK_LIBRARY:feature,library-list>` was added in 3.24

**Comment by rrsettgast** [comment]
@wrtobin Do you know version of cmake is/will be available on frontier and tioga/elcap?

File: src/CMakeLists.txt [hunk 1]
```diff
@@ -1,5 +1,5 @@
 
-cmake_minimum_required( VERSION 3.23 )
+cmake_minimum_required( VERSION 3.26 )
```
CusiniM:
```suggestion
cmake_minimum_required( VERSION 3.24 )
```

wrtobin:
> @wrtobin Do you know version of cmake is/will be available on frontier and tioga/elcap?

tioga / the EAS mi300 nodes have up to 3.29.2. I need to renew my olcf frontier access, but they also keep pretty up-to-date on cmake versions and had at least 3.26.X when I was actively working on frontier and crusher.

paveltomin:
CC @drmichaeltcvx

src/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 
-cmake_minimum_required( VERSION 3.23 )
+cmake_minimum_required( VERSION 3.24 )
 
 # At the moment we are manually passing the cuda arch flag.
 if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.23.0")
```
CusiniM:
```suggestion
```

untereiner:
It is reasonable to regularly update min cmake versions !

rrsettgast (approved):

PR #3337, opened 2024-09-06
fix: try this density treatment for gravity

paveltomin:
currently density in gravity term is `0.5(rho_i+rho_j)` where `i` and `j` are the cells
here we are changing it to more commonly used treatment that checks phase presence leading to:
1) `0.5(rho_i+rho_j)` when phase is present in both cells 
2) `rho_i` when only in `i`
3) `rho_j` when only in `j`
probably not a big deal but better to be consistent with mainstream way


src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseFVMKernelUtilities.hpp
```diff
@@ -1238,11 +1273,24 @@ struct computePotentialGravity
...
+      densMean += density;
+      dDensMean_dPres[i] = dDens_dPres;
       for( localIndex jc = 0; jc < numComp; ++jc )
       {
-        dDensMean_dComp[i][jc] = 0.5 * dProp_dComp[jc];
+        dDensMean_dComp[i][jc] = dProp_dComp[jc];
+      }
+      denom++;
+    }
+    if( denom > 1 )
```
dkachuma: Could a phase potentially be absent in both cells? Or is this checked elsewhere?

src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseFVMKernelUtilities.hpp
```diff
@@ -1221,12 +1249,19 @@ struct computePotentialGravity
...
 
     //inner loop to get average density
+    integer denom = 0;
     for( localIndex i = 0; i < numFluxSupportPoints; ++i )
     {
       localIndex const er = seri[i];
       localIndex const esr = sesri[i];
       localIndex const ei = sei[i];
 
+      bool const phaseExists = (phaseVolFrac[er][esr][ei][ip] > 0);
```
dkachuma: In some cases I have seen saturation weighted density in this case. Not sure what the benefit of that is over a simpler mean. Also looks like it might complicate the derivatives.

src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseFVMKernelUtilities.hpp
paveltomin: It can be absent in both yes, there is a check below but it would be better to check even outside of the function
Just don't want to add more complexity for this PR

src/coreComponents/physicsSolvers/fluidFlow/IsothermalCompositionalMultiphaseFVMKernelUtilities.hpp
paveltomin: Yeah, it should be more accurate, but I don't want to complicate the derivatives

paveltomin:
review please?

jafranc (approved):

paveltomin:
@CusiniM, @castelletto1, @cssherman, @jhuang2601, @rrsettgast, and/or @wrtobin could you please review?

CusiniM (approved):
I feel it should be weighted with the saturation but this is better than before.

CusiniM:
> I feel it should be weighted with the saturation but this is better than before.

CusiniM:
> I feel it should be weighted with the saturation but this is better than before.

or maybe using the mass fraction since it's a density.

paveltomin:
need to adjust thermal part too
move out from merge queue for now

PR #3338, opened 2024-09-09
feat: minimum interval for time-step increase after time-step cut

victorapm:
This PR addresses an issue where the simulation oscillates between time-step increases and cuts due to Newton loop stagnation. 

For example, in one simulation, the time-step gradually increases up to a point leading to Newton loop stagnation. Then, the time-step was subsequently reduced, allowing the Newton loop to converge, but it quickly tried to revert back to the original large time-step, causing the same stagnation. This cycle would repeat for many cycles in the simulation.

This PR introduces a parameter to enforce a minimum interval before the time-step can increase again after a reduction occurs, thus preventing the repeated back-and-forth time-step adjustments and avoiding Newton iteration failures in both parts of the simulation.

Side question: we have places where we use "time step" and "time-step". Should we choose one format and use it consistently?


src/coreComponents/physicsSolvers/SolverBase.cpp
```diff
@@ -401,14 +426,14 @@ real64 SolverBase::setNextDtBasedOnNewtonIter( real64 const & currentDt )
   if( newtonIter < iterIncreaseLimit )
   {
     // Easy convergence, let's increase the time-step.
-    nextDt = currentDt * m_nonlinearSolverParameters.timeStepIncreaseFactor();
+    nextDt = round( currentDt * m_nonlinearSolverParameters.timeStepIncreaseFactor() );
```
paveltomin: this is experimental, right?

victorapm: Yes, it's not the main purpose of this PR. If the time-steps are larger than O(10) rounding shouldn't be an issue, right?

src/coreComponents/physicsSolvers/SolverBase.cpp
```diff
@@ -247,6 +248,16 @@ bool SolverBase::execute( real64 const time_n,
...
 {
   GEOS_MARK_FUNCTION;
+
+  /*
+   * Reset counter indicating the number of cycles since the last timestep cut
+   * when the new timestep. "-1" means that no time-step cut has ocurred.
+   * */
+  if( dt < m_nextDt )
+  {
+    m_lastDtCut = -1;
```
paveltomin: so the logic only works with current "execute", no transfer to next cycle?

victorapm: I did this because the current execute might decrease the time-step to meet a exact time target for the event. Then, suppose you come from a recent time-step cut, If I didn't reset `m_lastDtCut`, the code would not be able to increase the time step again (after having decreased not because of a noncoverged Newton, but because it had to meet a target time). Hope this explanation is not too confusing

src/coreComponents/physicsSolvers/SolverBase.cpp
paveltomin: how about < 0.5 ? or i am missing something...

victorapm: Yes, then this doesn't work

src/coreComponents/physicsSolvers/SolverBase.cpp
paveltomin: it's ok, just might be confusing that logic is not applied across cycles
let's say we set `minTimeStepIncreaseInterval=5` and it does a cut inside one cycle, then does let's say 2 succesful subtimesteps and goes into another cycle - then that 5-2 counter is lost?

victorapm: Right, in this scenario, the cut information is lost for the next subEvent and I agree it shouldn't. 

Maybe it's possible to improve the logic here. If you have some ideas, feel free to implement!

victorapm: PS: in your scenario, if the time-step is not decreased to meet a exact target time, then the cut information is not lost in the next subEvent

CusiniM:
> This PR addresses an issue where the simulation oscillates between time-step increases and cuts due to Newton loop stagnation.
> 
> For example, in one simulation, the time-step gradually increases up to a point leading to Newton loop stagnation. Then, the time-step was subsequently reduced, allowing the Newton loop to converge, but it quickly tried to revert back to the original large time-step, causing the same stagnation. This cycle would repeat for many cycles in the simulation.
> 
> This PR introduces a parameter to enforce a minimum interval before the time-step can increase again after a reduction occurs, thus preventing the repeated back-and-forth time-step adjustments and avoiding Newton iteration failures in both parts of the simulation.
> 
> Side question: we have places where we use "time step" and "time-step". Should we choose one format and use it consistently?

I think that if we wanted to follow proper grammatical rules, it should be `time step` whenever it's a noun and `time-step` whenever it's adjectivized (e.g., time-step size, time-step control strategy).

src/coreComponents/physicsSolvers/SolverBase.cpp
CusiniM: yeah, we should not use round here. For some cases we need very tiny time steps.

castelletto1:
> > This PR addresses an issue where the simulation oscillates between time-step increases and cuts due to Newton loop stagnation.
> > 
> > For example, in one simulation, the time-step gradually increases up to a point leading to Newton loop stagnation. Then, the time-step was subsequently reduced, allowing the Newton loop to converge, but it quickly tried to revert back to the original large time-step, causing the same stagnation. This cycle would repeat for many cycles in the simulation.
> > 
> > This PR introduces a parameter to enforce a minimum interval before the time-step can increase again after a reduction occurs, thus preventing the repeated back-and-forth time-step adjustments and avoiding Newton iteration failures in both parts of the simulation.
> > 
> > Side question: we have places where we use "time step" and "time-step". Should we choose one format and use it consistently?
> 
> I think that if we wanted to follow proper grammatical rules, it should be `time step` whenever it's a noun and `time-step` whenever it's adjectivized (e.g., time-step size, time-step control strategy). 


Why not simply `timestep` ([Collins dictionary](https://www.collinsdictionary.com/us/dictionary/english/timestep#:~:text=noun,of%20the%20system%20under%20study))? _The timestep is set to ..._, _The timestep size ..._

src/coreComponents/physicsSolvers/SolverBase.hpp
```diff
@@ -1008,6 +1011,9 @@ class SolverBase : public ExecutableGroup
   /// timestep of the next cycle
   real64 m_nextDt;
 
+  /// Number of cycles since last timestep cut
+  integer m_lastDtCut;
```
CusiniM:
```suggestion
  integer m_numTimestepsSinceLastDtCut;
```

CusiniM:
> > > This PR addresses an issue where the simulation oscillates between time-step increases and cuts due to Newton loop stagnation.
> > > For example, in one simulation, the time-step gradually increases up to a point leading to Newton loop stagnation. Then, the time-step was subsequently reduced, allowing the Newton loop to converge, but it quickly tried to revert back to the original large time-step, causing the same stagnation. This cycle would repeat for many cycles in the simulation.
> > > This PR introduces a parameter to enforce a minimum interval before the time-step can increase again after a reduction occurs, thus preventing the repeated back-and-forth time-step adjustments and avoiding Newton iteration failures in both parts of the simulation.
> > > Side question: we have places where we use "time step" and "time-step". Should we choose one format and use it consistently?
> > 
> > 
> > I think that if we wanted to follow proper grammatical rules, it should be `time step` whenever it's a noun and `time-step` whenever it's adjectivized (e.g., time-step size, time-step control strategy).
> 
> Why not simply `timestep` ([Collins dictionary](https://www.collinsdictionary.com/us/dictionary/english/timestep#:~:text=noun,of%20the%20system%20under%20study))? _The timestep is set to ..._, _The timestep size ..._

The Oxford dictionary does not have the entry as a single word but it would really simplify our life. Spellcheckers don't recognize it though.

CusiniM:
Talking about the actual goal of the PR... The logic is a bit clunky but we certainly need to include some more info about past convergence when selecting the new dt so this could be a start. 

I don't think we can have all those `round` calls though coz we have often test cases (for like lab scale models) in which we use `dt~ 1e-3 s` so that would mess up things. @victorapm if you were tyring to  round up the values to improve the readability of the output we can probably do it using the units utilities https://github.com/GEOS-DEV/GEOS/blob/14a21b5ef217901aaa6de6bf80ba36952b91a6f1/src/coreComponents/common/Units.hpp#L215

victorapm:
Updated the variable name and removed `round` in [2302936](https://github.com/GEOS-DEV/GEOS/pull/3338/commits/23029364c8ff59aa4fff0c61400971753bbd7f84)

Let me know what you guys think!

I suggest voting for deciding wheter to use `timestep`, `time-step`, ...

paveltomin (approved):

CusiniM (approved):

paveltomin:
please also merge https://github.com/GEOS-DEV/SPE11/tree/paludettomag1/spe11c once this PR is done

victorapm:
It seems we'll need to rebaseline given the new logic and default value for `m_numTimestepsSinceLastDtCut`, which I think is good to have. @CusiniM

src/coreComponents/physicsSolvers/NonlinearSolverParameters.cpp
```diff
@@ -112,32 +112,37 @@ NonlinearSolverParameters::NonlinearSolverParameters( string const & name,
...
   registerWrapper( viewKeysStruct::timeStepIncreaseFactorString(), &m_timeStepIncreaseFactor ).
     setApplyDefaultValue( 2.0 ).
     setInputFlag( InputFlags::OPTIONAL ).
-    setDescription( "Factor by which the time step is increased when the number of Newton iterations is small." );
+    setDescription( "Factor by which the time-step is increased when the number of Newton iterations is small." );
+
+  registerWrapper( viewKeysStruct::minTimeStepIncreaseIntervalString(), &m_minTimeStepIncreaseInterval ).
+    setApplyDefaultValue( 10 ).
+    setInputFlag( InputFlags::OPTIONAL ).
+    setDescription( "Value of the minimum interval, since the last time-step cut, for increasing the time-step." );
```
rrsettgast: what does "minimum interval" mean? It is hard to tell from this description. is it "minimum number of time steps"?

src/coreComponents/physicsSolvers/NonlinearSolverParameters.cpp
```diff
@@ -112,32 +112,37 @@ NonlinearSolverParameters::NonlinearSolverParameters( string const & name,
...
+    setDescription( "Fraction of the max Newton iterations below which the solver asks for the time-step to be increased for the next time-step." );
 
   registerWrapper( viewKeysStruct::timeStepDecreaseFactorString(), &m_timeStepDecreaseFactor ).
     setApplyDefaultValue( 0.5 ).
     setInputFlag( InputFlags::OPTIONAL ).
-    setDescription( "Factor by which the time step is decreased when the number of Newton iterations is large." );
+    setDescription( "Factor by which the time-step is decreased when the number of Newton iterations is large." );
 
   registerWrapper( viewKeysStruct::timeStepIncreaseFactorString(), &m_timeStepIncreaseFactor ).
     setApplyDefaultValue( 2.0 ).
```
rrsettgast, line 128: I don't think this default should be `2.0`....should it?

rrsettgast (approved):
When increasing tilmestep, wouldn't it be appropriate to modify the time-step increase size based on the number of newton iterations the current step took to converge? So instead of: "if numNewtonIter < 8 then nextDt = dt * factor" we have the `factor` be dependent on `numNewtonIter`?

src/coreComponents/physicsSolvers/NonlinearSolverParameters.cpp
castelletto1: I agree that the default could be a little more conservative.

victorapm: That makes sense, but this is not the variable I'm adding in this PR (`m_timeStepIncreaseFactor`). Should we make the change here or in another PR?

src/coreComponents/physicsSolvers/NonlinearSolverParameters.cpp
victorapm: Right, it's minimum number of time steps. Let me update the comment

src/coreComponents/physicsSolvers/NonlinearSolverParameters.cpp
castelletto1: Yes, not in this PR

src/coreComponents/physicsSolvers/NonlinearSolverParameters.cpp
castelletto1: I find the comment below very clear:

```c++
GEOS_LOG_LEVEL_RANK_0( 1, GEOS_FMT( "{}: time-step size will be kept the same since it's been {} cycles since last cut.",
                                        getName(), m_numTimestepsSinceLastDtCut ) );
```

What about renaming `m_minNumCyclesFixedTimeStepAfterCut`. It's a bit wordy but a probably clearer?

src/coreComponents/physicsSolvers/SolverBase.cpp
```diff
@@ -337,13 +354,21 @@ void SolverBase::logEndOfCycleInformation( integer const cycleNumber,
 real64 SolverBase::setNextDt( real64 const & currentDt,
                               DomainPartition & domain )
 {
+  integer const minTimeStepIncreaseInterval = m_nonlinearSolverParameters.m_minTimeStepIncreaseInterval;
```
castelletto1:
```suggestion
  integer const minTimeStepIncreaseInterval = m_nonlinearSolverParameters.minTimeStepIncreaseInterval();
```

src/coreComponents/physicsSolvers/NonlinearSolverParameters.cpp
castelletto1: Better `m_minFixedTimeStepNumCyclesAfterCut`

castelletto1 (approved):
Except for the suggestion on rewording (optional), looks good to me. Thank you!

src/coreComponents/physicsSolvers/SolverBase.cpp
victorapm: Thanks for catching!

src/coreComponents/physicsSolvers/NonlinearSolverParameters.cpp
victorapm: I am fine with all options presented

castelletto1:
> I suggest voting for deciding whether to use `timestep`, `time-step`, ...

Let's postpone the renaming to a future PR and consistently apply the change everywhere in GEOS.

With all due respect to the Oxford Dictionary, my preference is to use `timestep`.

PR #3326, opened 2024-09-01
chore: Update quick-start-template.cmake

paveltomin:
also update schema and rst so that make does not complain

PR #3341, opened 2024-09-09
fix: update schema docs from previous PR's that forgot to generate the updates

rrsettgast:


CusiniM (approved):

CusiniM:
not the same diffs I see in https://github.com/GEOS-DEV/GEOS/pull/3326/files

paveltomin (approved):

PR #3349, opened 2024-09-12
feat: Configuration loop acceleration for SolidMechanicsLagrangeContact

sohailwaziri:
image
image


paveltomin:
@castelletto1 @matteofrigo5 @CusiniM this is actually ready to have a look

rrsettgast:
@paveltomin @sohailwaziri 
From the slide image and a quick glance at the code it is unclear exactly what is happening here. Can you explain in words the what is being done. For example, it seems you are trying to "accelerate" the evolution of the shear stress....is that the correct interpretation? What is the "accelerated variable" in this case? Is it the shear traction?...etc. A detailed discussion would be great. Perhaps at the Thursday meeting?

src/coreComponents/physicsSolvers/contact/SolidMechanicsLagrangeContact.cpp
```diff
@@ -2329,6 +2350,90 @@ bool SolidMechanicsLagrangeContact::updateConfiguration( DomainPartition & domai
...
+                                                               real64 const limitTau,
+                                                               real64 const currentTau,
+                                                               integer & fractureState )
+{
+  using namespace fields::contact;
+
+  if( configurationLoopIter == 0 )
+  {
+    m_x0[kfe] = currentTau_unscaled - limitTau;
+  }
```
castelletto1: I tried the example suggested by @matteofrigo5 ([ContactMechanics_SimpleCubes_smoke](https://github.com/GEOS-DEV/GEOS/blob/develop/inputFiles/lagrangianContactMechanics/ContactMechanics_SimpleCubes_smoke.xml)).

It happens that `total_size` computed in `initializeAccelerationVariables` is equal to 0, hence GEOS crashes here because `m_x0` has zero size, as all the other allocated arrays needed for the nonlinear acceleration.


 I suspect it is because in that example the `SurfaceGenerator` is used to split the mesh.
 
 @sohailwaziri: please make sure `initializeAccelerationVariables` is called after the mesh has been split, when the `SurfaceGenerator` is used. Thanks!

paveltomin: > I tried the example suggested by @matteofrigo5 ([ContactMechanics_SimpleCubes_smoke](https://github.com/GEOS-DEV/GEOS/blob/develop/inputFiles/lagrangianContactMechanics/ContactMechanics_SimpleCubes_smoke.xml)).
> 
> It happens that `total_size` computed in `initializeAccelerationVariables` is equal to 0, hence GEOS crashes here because `m_x0` has zero size, as all the other allocated arrays needed for the nonlinear acceleration.
> 
> I suspect it is because in that example the `SurfaceGenerator` is used to split the mesh.
> 
> @sohailwaziri: please make sure `initializeAccelerationVariables` is called after the mesh has been split, when the `SurfaceGenerator` is used. Thanks!

i got it fixed by moving `initializeAccelerationVariables` into `setupSystem`

without acceleration:
```
lagrangiancontact, number of time steps: 11
lagrangiancontact, number of successful outer loop iterations: 13
lagrangiancontact, number of successful nonlinear iterations: 94
lagrangiancontact, number of successful linear iterations: 94
lagrangiancontact, number of time step cuts: 1
lagrangiancontact, number of discarded outer loop iterations: 5
lagrangiancontact, number of discarded nonlinear iterations: 35
lagrangiancontact, number of discarded linear iterations: 35
```

with acceleration:
```
lagrangiancontact, number of time steps: 11
lagrangiancontact, number of successful outer loop iterations: 13
lagrangiancontact, number of successful nonlinear iterations: 95
lagrangiancontact, number of successful linear iterations: 95
lagrangiancontact, number of time step cuts: 1
lagrangiancontact, number of discarded outer loop iterations: 5
lagrangiancontact, number of discarded nonlinear iterations: 35
lagrangiancontact, number of discarded linear iterations: 35
```

so acceleration does not hurt but does not help either
maybe some parameter needs to be tuned as Sohail mentioned yesterday

paveltomin (approved):

src/coreComponents/physicsSolvers/solidMechanics/contact/SolidMechanicsLagrangeContact.cpp
```diff
@@ -2345,6 +2370,90 @@ bool SolidMechanicsLagrangeContact::updateConfiguration( DomainPartition & domai
...
+    else
+    {
+      // apply acceleration
+      m_x2_tilde[kfe] = currentTau_unscaled - limitTau;
+
+      m_omega1[kfe] = -1.0 * m_omega0[kfe] * (m_x1_tilde[kfe] - m_x0[kfe]) / ((m_x1_tilde[kfe] - m_x0[kfe]) - (m_x2_tilde[kfe] - m_x1[kfe]));
+
+      m_x2[kfe] = (1.0 - m_omega1[kfe]) * m_x1[kfe] + m_omega1[kfe] * m_x2_tilde[kfe];
+
+      real64 acc_currentTau_unscaled = m_x2[kfe] + limitTau;
```
CusiniM:
```suggestion
      real64 const acc_currentTau_unscaled = m_x2[kfe] + limitTau;
```

src/coreComponents/physicsSolvers/solidMechanics/contact/SolidMechanicsLagrangeContact.cpp
```diff
@@ -2345,6 +2370,90 @@ bool SolidMechanicsLagrangeContact::updateConfiguration( DomainPartition & domai
...
+      real64 acc_currentTau_scaled = acc_currentTau_unscaled * (1.0 - m_slidingCheckTolerance);
```
CusiniM:
```suggestion
      real64 const acc_currentTau_scaled = acc_currentTau_unscaled * (1.0 - m_slidingCheckTolerance);
```

src/coreComponents/physicsSolvers/solidMechanics/contact/SolidMechanicsLagrangeContact.hpp
```diff
@@ -133,7 +133,8 @@ class SolidMechanicsLagrangeContact : public ContactSolverBase
 
   bool resetConfigurationToDefault( DomainPartition & domain ) const override final;
 
-  bool updateConfiguration( DomainPartition & domain ) override final;
+  bool updateConfiguration( DomainPartition & domain,
+                            integer configurationLoopIter ) override final;
```
CusiniM:
```suggestion
                            integer const configurationLoopIter ) override final;
```

src/coreComponents/physicsSolvers/solidMechanics/contact/SolidMechanicsLagrangeContact.hpp
```diff
@@ -206,8 +207,29 @@ class SolidMechanicsLagrangeContact : public ContactSolverBase
...
+  array1d< real64 > m_x1_tilde; // Unaccelerated variable @ outer iteration v + 1 ( previous iteration )
+  array1d< real64 > m_x2; // Accelerated variable @ outer iteration v + 2 ( current iteration )
+  array1d< real64 > m_x2_tilde; // Unaccelerated variable @ outer iteration v + 1 ( current iteration )
+  array1d< real64 > m_omega0; // Old relaxation factor
+  array1d< real64 > m_omega1; // New relaxation factor
+  integer m_useLocalYieldAcceleration; // flag for applying acceleration to yield
+
+  void initializeAccelerationVariables( DomainPartition & domain );
+
+  void tryLocalYieldAcceleration( integer configurationLoopIter,
```
CusiniM:
```suggestion
  void tryLocalYieldAcceleration( integer const configurationLoopIter,
```

src/coreComponents/physicsSolvers/solidMechanics/contact/SolidMechanicsLagrangeContact.cpp
```diff
@@ -2345,6 +2370,90 @@ bool SolidMechanicsLagrangeContact::updateConfiguration( DomainPartition & domai
...
+  else if( configurationLoopIter == 1 )
+  {
+    m_x1_tilde[kfe] = currentTau_unscaled - limitTau;
+    m_x1[kfe] = m_x1_tilde[kfe];
+    m_omega0[kfe] = 1.0;
+  }
+  else
+  {
+    // only apply acceleration if within a fraction of the limitTau
+    real64 acceleration_buffer = (limitTau - currentTau) / limitTau;
```
CusiniM:
```suggestion
    real64 const acceleration_buffer = (limitTau - currentTau) / limitTau;
```

src/coreComponents/physicsSolvers/solidMechanics/contact/SolidMechanicsLagrangeContact.hpp
```diff
@@ -206,8 +207,29 @@ class SolidMechanicsLagrangeContact : public ContactSolverBase
...
+                                  localIndex kfe,
+                                  real64 currentTau_unscaled,
+                                  real64 limitTau,
+                                  real64 currentTau,
+                                  integer & fractureState );
```
CusiniM, line 228:
```suggestion
                                  integer const & fractureState );
```
?

CusiniM (commented):
So, It seems that the acclerated variables could be registered on the subregions instead of living on the solver? The way these arrays are sized and how they are accessed seems to only work for the case of a single fracture region with a signle fracture subregion. 

If you keep them as they are now, is there a specific reason to use LvArrays? It looks like that, currently, the acceleration is done inside a host loop. I am not sure whether this is by choice or not. I imagine that the host loop is there because of some of the logging.

paveltomin:
> So, It seems that the acclerated variables could be registered on the subregions instead of living on the solver? The way these arrays are sized and how they are accessed seems to only work for the case of a single fracture region with a signle fracture subregion.

Good point, thanks. I converted everything to be registered on the subregions.

> If you keep them as they are now, is there a specific reason to use LvArrays? It looks like that, currently, the acceleration is done inside a host loop. I am not sure whether this is by choice or not. I imagine that the host loop is there because of some of the logging.

I don't think there was a big reason for that, no GPU concerns in mind.

Guotong-Ren (approved):

src/coreComponents/physicsSolvers/solidMechanics/contact/SolidMechanicsLagrangeContact.cpp
```diff
@@ -69,6 +69,11 @@ SolidMechanicsLagrangeContact::SolidMechanicsLagrangeContact( const string & nam
     setApplyDefaultValue( 1.0 ).
     setDescription( "It be used to increase the scale of the stabilization entries. A value < 1.0 results in larger entries in the stabilization matrix." );
 
+  registerWrapper( viewKeyStruct::useLocalYieldAccelerationString(), &m_useLocalYieldAcceleration ).
```
jhuang2601, line 73: Can you add a smoke test to the integratedTest to protect this feature?

src/coreComponents/physicsSolvers/solidMechanics/contact/SolidMechanicsAugmentedLagrangianContact.cpp
```diff
@@ -810,7 +810,8 @@ void SolidMechanicsAugmentedLagrangianContact::updateState( DomainPartition & do
   GEOS_UNUSED_VAR( domain );
 }
 
-bool SolidMechanicsAugmentedLagrangianContact::updateConfiguration( DomainPartition & domain )
+bool SolidMechanicsAugmentedLagrangianContact::updateConfiguration( DomainPartition & domain,
```
jhuang2601, line 812: Can this feature be applied for AugmentedLagrangianContact and EmbeddedFractures?

jhuang2601 (approved):

src/coreComponents/physicsSolvers/solidMechanics/contact/SolidMechanicsAugmentedLagrangianContact.cpp
paveltomin: unfortunately no

src/coreComponents/physicsSolvers/solidMechanics/contact/SolidMechanicsLagrangeContact.cpp
paveltomin: yeah, i will work on that

PR #3353, opened 2024-09-13
feat: Table subs division

arng40:
This PR allows the table to be subdivided, but also simplifies the declaration of the table.
Column declarations can be mixed with string and ColumnParam struct, I.E: 

```
TableLayout const layoutTest( "Rank",
                             TableLayout::ColumnParam{"Nodes", TableLayout::Alignment::center},
                             "Edge",
                             TableLayout::ColumnParam{"Faces", TableLayout::Alignment::center},
                             TableLayout::ColumnParam{"Elems", TableLayout::Alignment::center} );
```
For an example with sub-columns, this will give the following result

 ```
TableLayout const layoutTest( "Cras egestas ipsum a nisl. Vivamus variu dolor utsisicdis parturient montes, nascetur ridiculus mus. Duis nascetur ridiculus mus ",
                              { "Rank",
                                 TableLayout::ColumnParam{"Nodes", TableLayout::Alignment::center, true, {"local", "ghost"}},
                                 "Edge",
                                 TableLayout::ColumnParam{"Faces", {"local", "ghost"}},
                                 TableLayout::ColumnParam{"Elems", TableLayout::Alignment::center, true, {"local", "ghost"}} 
				      });

[data decleration]
```
Output : 
```
               "--------------------------------------------------------------------------------------------------------------------------------------"
               "|  Cras egestas ipsum a nisl. Vivamus variu dolor utsisicdis parturient montes, nascetur ridiculus mus. Duis nascetur ridiculus mus  |"
               "--------------------------------------------------------------------------------------------------------------------------------------"
               "|                         Rank  |           Nodes           |      Edge      |           Faces           |           Elems           |"
               "--------------------------------------------------------------------------------------------------------------------------------------"
               "|                               |    local    |    ghost    |                |    local    |    ghost    |    local    |    ghost    |"
               "--------------------------------------------------------------------------------------------------------------------------------------"
               "|             min(local/total)  |      1      |      2      |       3        |      4      |      5      |      6      |      7      |"
               "|             min(local/total)  |      1      |      2      |       3        |      4      |      5      |      6      |      7      |"
               "--------------------------------------------------------------------------------------------------------------------------------------"
```

-------

Also, column titles are now centered by default, regardless on the content alignement.

### CODE REVIEW

**Comment by paveltomin** [comment]
i seems that this is not based on `develop` but some other branch because there are much more changes than in the description

**Comment by arng40** [comment]
> i seems that this is not based on `develop` but some other branch because there are much more changes than in the description

Indeed, it was based on the output PVT PR `https://github.com/GEOS-DEV/GEOS/pull/3149#pullrequestreview-2118808721` which is still waiting to be merged

**Comment by paveltomin** [comment]
> > i seems that this is not based on `develop` but some other branch because there are much more changes than in the description
> 
> Indeed, it was based on the output PVT PR `https://github.com/GEOS-DEV/GEOS/pull/3149#pullrequestreview-2118808721` which is still waiting to be merged

can you switch the base branch here?

**Comment by arng40** [comment]
> > > i seems that this is not based on `develop` but some other branch because there are much more changes than in the description
> > 
> > 
> > Indeed, it was based on the output PVT PR `https://github.com/GEOS-DEV/GEOS/pull/3149#pullrequestreview-2118808721` which is still waiting to be merged
> 
> can you switch the base branch here?

You mean this PR would be based on develop ? If so, PVT and table subcolumns PR are bound together.

**Comment by paveltomin** [comment]
> > > > i seems that this is not based on `develop` but some other branch because there are much more changes than in the description
> > > 
> > > 
> > > Indeed, it was based on the output PVT PR `https://github.com/GEOS-DEV/GEOS/pull/3149#pullrequestreview-2118808721` which is still waiting to be merged
> > 
> > 
> > can you switch the base branch here?
> 
> You mean this PR would be based on develop ? If so, PVT and table subcolumns PR are bound together.

I mean make this PR based on that another branch, so it is easier to review

**Comment by arng40** [comment]
@paveltomin Would it be possible to review [PVT PR ](https://github.com/GEOS-DEV/GEOS/pull/3149) as this is the oldest one and the priority PR for our users. As a said earlier, this PR is continuity with the PVT PR.

**Comment by paveltomin** [comment]
> @paveltomin Would it be possible to review [PVT PR ](https://github.com/GEOS-DEV/GEOS/pull/3149) as this is the oldest one and the priority PR for our users. As a said earlier, this PR is continuity with the PVT PR. 

I already approved that one

File: src/coreComponents/common/format/table/TableFormatter.hpp [hunk 1]
```diff
@@ -138,7 +137,7 @@ class TableTextFormatter : public TableFormatter
   /**
    * @return A TableLayout converted into a formatted representation.
```
MelReyCG: Try to **explain** more than just describing the code:
```suggestion
   * @return the table with no data, so only the top part is returned. 
```
(something like that, suggestion may not be perfect)

src/coreComponents/common/format/table/unitTests/testTable.cpp
```diff
@@ -289,41 +247,84 @@ TEST( testTable, layoutTable )
...
+    tableData.addRow( "max", "360", "390,1", " YES", 383213213, 712, 48.0, 47.0, 72, 2 );
+
+    TableTextFormatter tableLog( tableLayout );
+
+    EXPECT_EQ( tableLog.toString( tableData ),
+               "\n--------------------------------------------------------------------------------------------------------\n"
+               "|       |  Column1  |                            Nodes   |  Column3  |           Column4   |  Column5  |\n"
+               "--------------------------------------------------------------------------------------------------------\n"
+               "|       |           |   Locales  |  Ghost  |     Active  |           |  Locales  |  Ghost  |           |\n"
+               "--------------------------------------------------------------------------------------------------------\n"
```
MelReyCG: Header texts should be centered or justified at the left side, but cannot be justified on the right side (general standard).
I have a preference to centering it since the title is centered by default. Can you follow the same behaviour the Title has?

src/coreComponents/common/format/table/TableLayout.hpp
```diff
@@ -58,65 +58,145 @@ class TableLayout
     /// Name for a column
     string columnName;
     /// Alignment for a column. By default aligned to the right side
-    Alignment alignment = Alignment::right;
+    Alignment alignment;
```
MelReyCG: As you removed ` = Alignment::right`, `alignment` is no longer initialized if we call 2/4 constructors.

src/coreComponents/common/format/table/unitTests/testTable.cpp
```diff
@@ -27,13 +27,13 @@ using namespace geos;
...
-                                  "Next\nelement"},
-                                 "InternalWellGenerator well_injector1"
-                                 );
+  TableLayout tableLayout( {"Well\nelement no.\nPV weighted\nbar",
+                            "CordX",
+                            "CoordZ",
+                            "Prev\nelement",
+                            "Next\nelement"} );
+  string const title = "InternalWellGenerator well_injector1";
+  tableLayout.setTitle( title );
```
MelReyCG: Please keep on using the `const` whenever possible.

src/coreComponents/common/format/table/TableLayout.hpp
```diff
@@ -51,106 +52,313 @@ class TableLayout
...
   /**
-   * @brief Construct a new Table object, all values in the table are centered by default
-   * @param columnNames The names of the columns
-   * @param title The table name
+   * @brief Construct a new Table Layout object
+   * @param title The table title
+   * @param columns A vector containing all column initialized
    */
-  TableLayout( std::vector< string > const & columnNames, string const & title = "" );
+  TableLayout( string_view title,
```
MelReyCG, line 194: I like the fact that the title is now first in the constructors (more intuitive) :)

src/coreComponents/common/format/table/unitTests/testTable.cpp
```diff
@@ -87,16 +89,17 @@ TEST( testTable, tableClassic )
...
-    TableLayout::ColumnParam{{"Prev\nelement"}, TableLayout::Alignment::right},
-    TableLayout::ColumnParam{{"Next\nelement"}, TableLayout::Alignment::right}
-  }, "InternalWellGenerator well_injector1" );
+  TableLayout tableLayout( {TableLayout::Column{{"Cras egestas"}, TableLayout::Alignment::center},
+                            TableLayout::Column{{"CoordX"}, TableLayout::Alignment::left},
+                            TableLayout::Column{{"C"}, TableLayout::Alignment::left},
+                            TableLayout::Column{{"CoordZ"}, TableLayout::Alignment::left},
+                            TableLayout::Column{{"Prev\nelement"}, TableLayout::Alignment::right},
+                            TableLayout::Column{{"Next\nelement"}, TableLayout::Alignment::right}} );
+  tableLayout.setValuesAlignment( TableLayout::Alignment::right );
```
MelReyCG, line 98: Is the goal of this test to consolidate the `setValuesAlignment()` code?
If so, 
- keep `tableLayout` non-const,
- use your builder pattern here,
- no need to test that more that one time, it makes test heavier for no purposes.

src/coreComponents/common/format/table/unitTests/testTable.cpp
```diff
@@ -289,41 +248,85 @@ TEST( testTable, layoutTable )
...
-// "| value 1 |long value 2|   100.45|4.0129877|      1|      2|\n"
-// "------------------------------------------------------------\n\n"
-//              );
-// }
+  TableLayout tableLayout( {"Cras egestas", "CoordX", "C", "CoordZ", "Prev\nelement", "Next\nelement"} );
+  tableLayout.setTitle( "title" ).setMargin( TableLayout::MarginValue::tiny ).disableLineWrap();
+
+  TableData tableData;
+  tableData.addRow( "1", "2", "3.0", 3.0129877, 2.0f, 1 );
+  tableData.addRow( "1", "2", "3.0", 3.0129877, 2.0f, 1 );
```
MelReyCG: Good thing that we can use a line break in titles :)
- `testLineBreak` would be more accurate as a name (_line wrapping_ term is more dedicated for automatic line breaks).
- I really think it should support line breaks _within content lines_:
```
  tableData.addRow( "1", "2", "one line text", 3.0129877, 2.0f, 1 );
  tableData.addRow( "1", "2", "three\nlines\ntext", 3.0129877, 2.0f, 1 );
  tableData.addRow( "1", "2", "one line text", 3.0129877, 2.0f, 1 );
```

MelReyCG (changes requested):
First pass on test & general functionnalities

MelReyCG (changes requested):
First pass on test & general functionnalities

src/coreComponents/common/format/table/TableFormatter.cpp
```diff
@@ -79,13 +82,47 @@ string TableCSVFormatter::toString< TableData >( TableData const & tableData ) c
 ///////////////////////////////////////////////////////////////////////
 
 /**
- * @brief Build  cell given an alignment, a value and spaces
+ * @brief distributes spaces equally from a string vector
+ * @param vec The vector where we adding spaces
+ * @param totalSpaces Total spaces to be distributed
```
MelReyCG: I have issues for really understand the point of this method, would it be a good rewrite?
```suggestion
 * @param totalSpaces Total spaces to distribute over all strings
```
If not, please fix it so it is as much accurate as possible.

src/coreComponents/common/format/table/TableFormatter.cpp
```diff
@@ -79,13 +82,47 @@ string TableCSVFormatter::toString< TableData >( TableData const & tableData ) c
...
+  int extraSpaces = totalSpaces % numElements;
+
+  for( int i = 0; i < numElements; ++i )
+  {
+    vec[i] += string( baseSpaces, ' ' );
+
+    if( i < extraSpaces )
+    {
+      vec[i] += ' ';
+    }
```
MelReyCG: Nice

src/coreComponents/common/format/table/TableFormatter.cpp
```diff
@@ -79,13 +82,47 @@ string TableCSVFormatter::toString< TableData >( TableData const & tableData ) c
...
- * @brief Build  cell given an alignment, a value and spaces
+ * @brief distributes spaces equally from a string vector
+ * @param vec The vector where we adding spaces
+ * @param totalSpaces Total spaces to be distributed
+ */
+void distributeSpaces( std::vector< string > & vec, int totalSpaces )
+{
+  int numElements = vec.size();
+  int baseSpaces = totalSpaces / numElements;
+  int extraSpaces = totalSpaces % numElements;
```
MelReyCG: A standard method exists to group the calls of divide + modulo.

src/coreComponents/common/format/table/TableLayout.hpp
```diff
@@ -51,106 +52,313 @@ class TableLayout
...
     {}
   };
 
   /**
-   * @brief Struct for a column.
+   * @brief Struct for a TableColumnData.
+   * Each column contains its own parameters (such as name, alignment, etc.).
    */
-  struct Column
+  struct TableColumnData
```
MelReyCG: Renaming to `ColumnData` should suffice as you are already in a `Table` related class, what do you think?

src/coreComponents/common/format/table/TableFormatter.hpp
```diff
@@ -136,9 +135,10 @@ class TableTextFormatter : public TableFormatter
   virtual ~TableTextFormatter() = default;
 
   /**
-   * @return A TableLayout converted into a formatted representation.
+   * @return A TableLayout string representation,
+   * The TableTextFormatter receives hasn't receive any data, so only the top part is returned.
```
MelReyCG:
```suggestion
   * The TableTextFormatter receives hasn't receive any data, so only the header part is returned.
```

src/coreComponents/common/format/table/TableFormatter.cpp
```diff
@@ -79,13 +82,47 @@ string TableCSVFormatter::toString< TableData >( TableData const & tableData ) c
...
+    if( i < extraSpaces )
+    {
+      vec[i] += ' ';
+    }
+  }
+}
+
+void transpose( std::vector< std::vector< string > > & dest,
+                std::vector< std::vector< string > > const & source )
+{
```
MelReyCG, line 87: What happens if `dest` does not following your size expectations? (spoiler: crash)
If you don't want to verify that within this method, you **must** specify your requirements in the method documentation.

src/coreComponents/common/format/table/unitTests/testTable.cpp
```diff
@@ -289,41 +248,85 @@ TEST( testTable, layoutTable )
...
+
+    TableData tableData;
+    tableData.addRow( "min", "125", "375,0001", " YES", 2354654, 562, 43.0, 43.0, 562, 5 );
+    tableData.addRow( "max", "360", "390,1", " YES", 383213213, 712, 48.0, 47.0, 72, 2 );
+
+    TableTextFormatter tableText( tableLayout );
+
+    EXPECT_EQ( tableText.toString( tableData ),
+               "\n--------------------------------------------------------------------------------------------------------\n"
+               "|       |  Column1  |                            Nodes   |  Column3  |           Column4   |  Column5  |\n"
```
MelReyCG: Does it passed `ctest`??

src/coreComponents/common/format/table/unitTests/testTable.cpp
arng40: `setValuesAlignment ` is for cellValues and alignement `TableLayout::Column{}` is for header alignment

src/coreComponents/common/format/table/unitTests/testTable.cpp
```diff
@@ -289,41 +242,85 @@ TEST( testTable, layoutTable )
...
+                                    TableLayout::Column{"Nodes ", TableLayout::Alignment::right, true, {"Locales", "Ghost", "Active"}},
+                                    "Column3",
+                                    TableLayout::Column{"Column4 ", TableLayout::Alignment::right, true, {"Locales", "Ghost"}},
+                                    "Column5"} );
+
+    TableData tableData;
+    tableData.addRow( "min", "125", "375,0001", " YES", 2354654, 562, 43.0, 43.0, 562, 5 );
+    tableData.addRow( "max", "360", "390,1", " YES", 383213213, 712, 48.0, 47.0, 72, 2 );
+
+    TableTextFormatter tableText( tableLayout );
```
MelReyCG, line 259: Would it be hard to refactor to something like this?
```suggestion
    TableLayout tableLayout( {" ",
                              "Column1",
                              TableLayout::Column{"Locales", TableLayout::Alignment::right, true, "Nodes"},
                              TableLayout::Column{"Ghost", TableLayout::Alignment::right, true, "Nodes"},
                              TableLayout::Column{"Active", TableLayout::Alignment::right, true, "Nodes"},
                              "Column3",
                              TableLayout::Column{"Locales", TableLayout::Alignment::right, true, "Column4"},
                              TableLayout::Column{"Ghost", TableLayout::Alignment::right, true, "Column4"},
                              "Column5"} );
```
We would have a "parent column" last parameter, it would not be closed to further sub-sub-column management (through something like "level1 Parent Column/level2 Parent Column")

src/coreComponents/common/format/table/unitTests/testTable.cpp
arng40: I can make line break for cell values if needed, not sure if right this is mandatory

src/coreComponents/common/format/table/TableLayout.hpp
```diff
@@ -51,106 +52,313 @@ class TableLayout
...
+   * @param title The table title
+   * @return The tableLayout reference
+   */
+  TableLayout & setTitle( string_view title );
+
+  /**
+   * @brief Remove the last return line a the end of the table
+   * @return The tableLayout reference
+   */
+  TableLayout & disableLineWrap();
```
MelReyCG: Perfer some signature that allow to enable/disable (by a bool parameter).
Also, prefer the _line break_ term.

src/coreComponents/common/format/table/TableLayout.hpp
```diff
@@ -51,106 +52,313 @@ class TableLayout
...
+    TableColumnData ( Column col )
+      : column( col )
+    {}
+
+    /**
+     * @brief Constructs a TableColumnData with the given parameters.
+     * @param col The parameters for the column.
+     * @param subColumnInit The subcolumns contained in the colum
+     */
+    TableColumnData ( Column col, std::vector< TableColumnData > subColumnInit )
```
MelReyCG: Please verify if **all** `std::vector` passed as _in-parameters_ in your header should better be `const &`.
Also please double check the ordering `X const &` (not `const X &`), at least one is misordered.

src/coreComponents/common/format/table/TableLayout.hpp
```diff
@@ -51,106 +52,313 @@ class TableLayout
...
-    ColumnParam m_parameter;
-    /// A vector containing all column values
-    std::vector< string > m_columnValues;
-    /// The largest string in the column
-    string m_maxStringSize;
+    Column column;
+    /// A vector containing all the values of a column
+    std::vector< string > columnValues;
+    /// The largest string(s) in the column
+    std::vector< string > maxStringSize;
```
MelReyCG: From the name and docs, I do not understand the point of this parameter:
For me, a `maxStringSize` parameter would refer to a number which is the maximum string size, but I wouldn't expect it to be stored as a string. Also, why a list ?

src/coreComponents/common/format/table/TableLayout.hpp
```diff
@@ -51,106 +52,313 @@ class TableLayout
...
-    Alignment alignment = Alignment::right;
+    /// Alignment for  column name. By default aligned to center
+    ColumnAlignment alignmentSettings;
     /// A boolean to display a colummn
     bool enabled = true;
+    /// Vector containing sub columns name
+    std::vector< string > subColumnNames;
     /// Vector containing substring column name delimited by "\n"
-    std::vector< string > splitColumnNameLines;
+    std::vector< string > splitColumnNames = {""};
```
MelReyCG, line 79: Can a user mess with this parameter? It seems like it should be private.

src/coreComponents/common/format/table/TableFormatter.cpp
```diff
@@ -127,284 +164,466 @@ TableTextFormatter::TableTextFormatter( TableLayout const & tableLayout ):
...
+  if( m_tableLayout.isLineBreakEnabled())
+  {
+    tableOutput << '\n';
+  }
+}
+
+void TableTextFormatter::populateColumnsFromTableData( std::vector< TableLayout::TableColumnData > & tableColumnsData,
+                                                       std::vector< std::vector< string > > const & tableDataRows,
+                                                       bool isSubColumn ) const
+{
```
MelReyCG: In my humble opinion, this method should be rewritten to be clearly understandable. Some tips:
- Try to create two clear version rather than this `isSubColumn` flag,
- Try to move the code at the best place possible,
- Avoid having a variable which sense can change because of a condition, it may have a specialized name for each case that may help a lot,
- work on naming, maybe you can add named blocks, add comments if you really think that we need extra info to understand it.

src/coreComponents/common/format/table/TableFormatter.cpp
```diff
@@ -127,284 +164,466 @@ TableTextFormatter::TableTextFormatter( TableLayout const & tableLayout ):
...
+    }
+  }
 }
 
-void TableTextFormatter::splitAndSetColumnNames( std::vector< TableLayout::Column > & columns,
-                                                 size_t & nbHeaderRows,
-                                                 std::vector< std::vector< string > > & splitHeaders ) const
+void TableTextFormatter::splitAndMergeColumnHeaders( std::vector< TableLayout::TableColumnData > & tableColumnsData,
+                                                     size_t & nbHeaderRows,
+                                                     std::vector< std::vector< string > > & splitHeaders ) const
```
MelReyCG: - Do you really use all those out-parameters?
- Why `nbHeaderRows` does not take into account sub-columns? Can it cause issues? (is it tested?)

src/coreComponents/common/format/table/TableFormatter.cpp
```diff
@@ -127,284 +164,466 @@ TableTextFormatter::TableTextFormatter( TableLayout const & tableLayout ):
...
+    std::for_each( tableColumnData.maxStringSize.begin(), tableColumnData.maxStringSize.end(),
+                   [&] ( string maxString ) {
+      maxStringsPerColumn.push_back( string( maxString.length(), m_horizontalLine ) );
+    } );
+  }
+
   integer const columnMargin = m_tableLayout.getColumnMargin();
   integer const borderMargin = m_tableLayout.getBorderMargin();
-  string const tableTitle = string( m_tableLayout.getTitle() );
+  string const patternBetweenColumns = GEOS_FMT( "{:-^{}}", m_horizontalLine, columnMargin );
```
MelReyCG: Looks like an emoticon :-^{}

src/coreComponents/common/format/table/TableFormatter.cpp
```diff
@@ -127,284 +164,466 @@ TableTextFormatter::TableTextFormatter( TableLayout const & tableLayout ):
...
+  {
+    tableOutput << GEOS_FMT( "{}\n", sectionSeparatingLine );
   }
 }
 
-void TableTextFormatter::outputTopRows( std::ostringstream & tableOutput,
-                                        std::vector< string > const & msg,
-                                        string_view topSeparator,
-                                        TableLayout::Alignment alignment ) const
+void TableTextFormatter::outputTitleRow( std::ostringstream & tableOutput,
```
MelReyCG, line 416: Would you prefer a chronological order for those methods to have something like title->header->section->subsection?

MelReyCG (changes requested):
2nd pass, format/table classes

MelReyCG:
I must say that it is a good thing that your worked on renaming/refactoring some parts, it will helps a lot in understanding.

src/coreComponents/common/format/table/TableFormatter.cpp
arng40: the new description fit better for understanding

src/coreComponents/common/format/table/TableFormatter.cpp
arng40: I removed nbHeaderRows parameters in all signature, no issue for we add '\n' to subcolumn bc it's not implemented yet, it will just take the first part of the string

src/coreComponents/common/format/table/TableLayout.hpp
arng40: I think `ColumnStructure ` is better

src/coreComponents/common/format/table/TableLayout.hpp
arng40: this parameter is assigned in tableFormatter function

src/coreComponents/common/format/table/unitTests/testTable.cpp
arng40: ofc

src/coreComponents/common/format/table/TableLayout.hpp
arng40: by default it's always enabled, I don't see when a user is going to reactivate it after having disabling it.

src/coreComponents/common/format/table/unitTests/testTable.cpp
arng40: If needed i could go for this but rn it will take some time to impl this

src/coreComponents/common/format/table/TableLayout.hpp
arng40: I will modify it but a column can have many subColumns and we need to keep track of the largest string for each subColumn, this way the parent know the largest string among the subColumn and itself

src/coreComponents/common/format/table/TableLayout.hpp
MelReyCG: I think it should be annotated as an "internal computed value", as it is not dedicated to be modified (but public, in a set-up structure).

src/coreComponents/common/format/table/unitTests/testTable.cpp
MelReyCG: Ok I see, this is not what we discussed at first, but it is good to be able to modify both setting by this way.
So two requests:
- Keep in mind that the value alignement is more likely to be customised than the header alignement. So please modify the `valueAlignment` in the `Column` constructors.
- Test the `setValuesAlignment()` only one time.
- use your builder pattern here, and declare `tableLayout` constant.

src/coreComponents/common/format/table/TableFormatter.cpp
MelReyCG: Way better now!

src/coreComponents/common/format/table/TableLayout.hpp
MelReyCG: It breaks the usability of the builder pattern if using a variable for setting it.

src/coreComponents/common/format/table/TableLayout.hpp
MelReyCG: again ':D

src/coreComponents/common/format/table/TableLayout.hpp
MelReyCG: I think only an integer would suffice to store the maxStringSize :
```
            Column4   < 7 + 4 spaces = 11 chars
---------------------
  Locales  |  Ghost   < 7 + 5 + 8 spaces + 1 separator = 21 chars
```
I don't think you need to store more

src/coreComponents/common/format/table/TableFormatter.cpp
MelReyCG: I think it needs more work to either fully implement what those lines tries to do:
```c++
    if( !tableColumnData.subColumn.empty())
    {
      std::vector< std::vector< string > > splitSubColHeaders;
      size_t nbHeaderSubColRows = 0;
      splitAndMergeColumnHeaders( tableColumnData.subColumn, nbHeaderSubColRows, splitSubColHeaders );
    }
```

src/coreComponents/common/format/table/TableFormatter.hpp
MelReyCG: Please correct your correction.

src/coreComponents/common/format/table/unitTests/testTable.cpp
MelReyCG: I encountered a scenario that needed this, but I forgot it :(
We can get back on this later.

MelReyCG: Ok, what do you think of this?
```
-------------------------------------
|  stat name  | component |  value  |
-------------------------------------
|  pressure   |    all    | 123.456 |
| temperature |    all    | 123.456 |
-------------------------------------
|             |    oil    | 123.456 |
|    mass     |    gaz    | 123.456 |
|             |   water   | 123.456 |
-------------------------------------
```

I think it would need to be able to have a `TableData::addSeparatorRow()`, and we may need multiline for doing a `tableData.addRow("mass","oil\ngaz\nwater","123.456\n123.456\n123.456")`

src/coreComponents/common/format/table/unitTests/testTable.cpp
MelReyCG: I messed up what the formatting would do with those setting because of the `TableLayout::Alignment` targetting the headers and not the content. As I said in another comment, please update the value alignement in priority.

MelReyCG (changes requested):
I added some requests.

src/coreComponents/common/format/table/TableLayout.hpp
MelReyCG: Please, either set it private or anotate it as an "internal computed value".

src/coreComponents/common/format/table/TableFormatter.cpp
MelReyCG: You seem to now correctly store the results of `splitAndMergeColumnHeaders()` in `splitColumnNames`.
So what is the use of `allDividedHeaderParts` out-parameter?

src/coreComponents/common/format/table/TableFormatter.hpp
MelReyCG: Please fix "The TableTextFormatter receives hasn't receive any data" sentence

MelReyCG:
Work merged in #3406

PR #3356, opened 2024-09-13
feat: Enable reservoir+wells+contact mechanics

paveltomin:
- [x] fim
- [x] sequential
- [x] test cases


jhuang2601 (approved):

paveltomin:
@CusiniM, @MelReyCG, @castelletto1, @corbett5, @rrsettgast, @untereiner, and/or @wrtobin could you please review?

src/coreComponents/common/format/table/TableFormatter.cpp
```diff
@@ -169,7 +169,6 @@ string TableTextFormatter::toString( TableData const & tableData ) const
 
   outputLayout( tableOutput, columns, msgTableError, sectionSeparatingLine );
   outputSectionRows( columns, sectionSeparatingLine, tableOutput, nbValuesRows, TableLayout::Section::values );
-  tableOutput << '\n';
```
CusiniM: won't this affect other tables as well?

paveltomin: it will and I hate that additional line break

CusiniM (approved):

src/coreComponents/common/format/table/TableFormatter.cpp
CusiniM: okay, maybe a separate small PR for this would be better. Unless it's a known issue. What tables are these we are writing to the log? 

Also coz if you move it to its own PR you get rid of some codeowners and we can merge this.

paveltomin: ok let me revert

paveltomin:
rebaseline needed because of `allowNegativePressure` flag added for wells

PR #3358, opened 2024-09-16
fix: Remove reference to fem space in kernelBase.

CusiniM:
I think that because of the way the kernel object is first created on host [here](https://github.com/GEOS-DEV/GEOS/blob/3b5b5a6f2f0a835d8323bab6f1c0eaca0e6f5924/src/coreComponents/finiteElement/kernelInterface/KernelBase.hpp#L442) and then captured by value in [this](https://github.com/GEOS-DEV/GEOS/blob/3b5b5a6f2f0a835d8323bab6f1c0eaca0e6f5924/src/coreComponents/finiteElement/kernelInterface/KernelBase.hpp#L238) `forAll` , it should not be a reference but a copy of the FEM space.

If we use a reference, the capture by value would result in copying a host address that leads to a run time failure when running on GPU systems. I think that this has gone unnoticed because, apart from the VEM ones, all femSpace classes only have static methods and no actual data. 

I don't know whether @castelletto1 or @andrea-borio have tried to run the VEM solver on GPU but I would expect it to fail (unless it's run on lassen with `ats` enabled).


rrsettgast (approved):

CusiniM:
I was getting a `_host_ function called by a _host_device_ function` error because of the destructor. I have removed the virtual destructors and forced the use of the default one and this seems to fix the compilation error. However, these objects do have some virtual methods though so theoretically I think having a virtual destructor would be better. Also, these objects are used on device so I don't think they really need virtual methods. Can one of you guys (@rrsettgast @corbett5 @wrtobin @klevzoff) confirm that this is fine?

rrsettgast:
I do t think these classes are ever using dynamic polymorphism or being used as a base pointer. Perhaps we just get make the functions non virtual

CusiniM:
> I do t think these classes are ever using dynamic polymorphism or being used as a base pointer. Perhaps we just get make the functions non virtual

I guess we do use a base object reference somewhere https://github.com/GEOS-DEV/GEOS/actions/runs/11033518585/job/30644910171?pr=3358

wrtobin:
> I was getting a `_host_ function called by a _host_device_ function` error because of the destructor. I have removed the virtual destructors and forced the use of the default one and this seems to fix the compilation error. However, these objects do have some virtual methods though so theoretically I think having a virtual destructor would be better. However, these objects are used on device so I don't think they really need virtual methods. Can one of you guys (@rrsettgast @corbett5 @wrtobin @klevzoff) confirm that this is fine?

Specifically for this issue, I've sometimes (more on AMD) had to implement a host-device decorated non-default virtual destructor to avoid that specific compilation failure due to the `=default` apparently sometimes not accommodating decoration appropriately (this is definitely a compiler bug on either the host or device compiler, but nonetheless something I've encountered -- thought nvcc had it long fixed though).

That said I agree that we should probably be able to get rid of polymorphism here anyway.

src/coreComponents/finiteElement/elementFormulations/H1_Hexahedron_Lagrange1_GaussLegendre2.hpp
```diff
@@ -92,8 +92,9 @@ class H1_Hexahedron_Lagrange1_GaussLegendre2 final : public FiniteElementBase
   /** @cond Doxygen_Suppress */
   USING_FINITEELEMENTBASE
   /** @endcond Doxygen_Suppress */
-
-  virtual ~H1_Hexahedron_Lagrange1_GaussLegendre2() override
+  
+  GEOS_HOST_DEVICE 
+  virtual ~H1_Hexahedron_Lagrange1_GaussLegendre2() overridePartitionNumbers
```
CusiniM:
```suggestion
  virtual ~H1_Hexahedron_Lagrange1_GaussLegendre2() override
```

PR #3359, opened 2024-09-19
feat: Add functions to connect well perforation to surface elements

paveltomin:
~should go after https://github.com/GEOS-DEV/GEOS/pull/3227~

- move `calculateElementCenters` for subregions earlier into `ProblemManager::generateMeshLevel`, so cell centers for surface elements are available when well perforation is processed
- remove `registerWrapper` for `elementCenter` and `elementVolume` in `EmbeddedSurfaceSubRegion`, they already registered in the base class
- modify `getReservoirElementDimensions`/`getBoundingBox` to support `SurfaceElementSubRegion`
- add `SurfaceElementSubRegion` version of `isPointInsideElement` and helper functions `isPointInPolygon2d/3d`

(not super refined implementation but something to start working with)


paveltomin:
need review please

paveltomin:
need review please

src/coreComponents/mesh/WellElementSubRegion.cpp
```diff
@@ -154,17 +154,72 @@ bool isPointInsideElement( CellElementSubRegion const & subRegion,
...
+    for( integer j = 0; j < 3; ++j )
+    {
+      polygon[i][j] = referencePosition[nodeList( eiLocal, i )][j];
+    }
+  }
+
+  // remove duplicates
+  std::unordered_set< Point3d, PointHash< Point3d >, PointsEqual< Point3d > >
+  unique_points( polygon.begin(), polygon.end());
+  polygon.clear();
```
Guotong-Ren: is degenerated polygon possible here?

paveltomin:
@CusiniM, @castelletto1, @jhuang2601, @rrsettgast please review

src/coreComponents/mesh/FaceManager.cpp
```diff
@@ -235,6 +236,11 @@ void FaceManager::computeGeometry( NodeManager const & nodeManager )
                                                                          m_faceNormal[ faceIndex ] );
 
   } );
+
+  elemManager.forElementSubRegions< CellElementSubRegion, FaceElementSubRegion >( [&] ( auto const & subRegion )
+  {
+    subRegion.calculateElementCenters( X );
+  } );
```
CusiniM: I find a bit weird that a function in `FaceManager` called `computeGeometry` takes care of computed element centers of the subregions. Also, do the fracture elements even exist at this point? I am assuming this occurs way before the event loop in which the surfacegenerator creates the fractures.

src/coreComponents/mesh/PerforationData.cpp
```diff
@@ -233,23 +233,40 @@ void PerforationData::getReservoirElementDimensions( MeshLevel const & mesh,
...
+                                           subRegion.nodeList(),
+                                           nodeManager.referencePosition(),
+                                           boxDims );
+
+    // dx and dz from bounding box
+    dx = boxDims[ 0 ];
+    dy = boxDims[ 1 ];
+    dz = boxDims[ 2 ];
+
+    if( dx < 1e-10 )
```
CusiniM: what's the meaning of this value?

src/coreComponents/mesh/utilities/ComputationalGeometry.hpp
```diff
@@ -433,6 +433,124 @@ bool isPointInsidePolyhedron( arrayView2d< real64 const, nodes::REFERENCE_POSITI
...
+  {
+    auto const & p1 = polygon[i];
+    auto const & p2 = polygon[(i + 1) % n];
+
+    if((point[1] > std::min( p1[1], p2[1] )) &&
+       (point[1] <= std::max( p1[1], p2[1] )) &&
+       (point[0] <= std::max( p1[0], p2[0] )))
+    {
+      real64 const xIntersect = (point[1] - p1[1]) * (p2[0] - p1[0]) / (p2[1] - p1[1]) + p1[0];
+      if( std::abs( p1[0] - p2[0] ) < 1e-10 || point[0] <= xIntersect )
```
CusiniM: the tolerance seems a bit arbitrary, doesn't it?

src/coreComponents/mesh/WellElementSubRegion.cpp
```diff
@@ -154,17 +154,72 @@ bool isPointInsideElement( CellElementSubRegion const & subRegion,
...
+  for( integer i = 0; i < nV; ++i )
+  {
+    for( integer j = 0; j < 3; ++j )
+    {
+      polygon[i][j] = referencePosition[nodeList( eiLocal, i )][j];
+    }
+  }
+
+  // remove duplicates
+  std::unordered_set< Point3d, PointHash< Point3d >, PointsEqual< Point3d > >
```
CusiniM: why would there be duplicates? is it because of the duplicated nodes?

herve-gross:
@jhuang2601 can you check if this can be merged? Thanks

src/coreComponents/mesh/PerforationData.cpp
jhuang2601: It was replaced with `if( isZero( dx ) )`

src/coreComponents/mesh/utilities/ComputationalGeometry.hpp
jhuang2601: Yes, it is a hard coded tol, which is sufficient for checking geometric overlapping.

src/coreComponents/mesh/WellElementSubRegion.cpp
jhuang2601: and also the intersection of multiple surfaces.

jhuang2601 (approved):

jhuang2601:
@rrsettgast @castelletto1 @Guotong-Ren Can you help to review this PR, which is needed for a use case?

Guotong-Ren (approved):

jhuang2601:
Hello @rrsettgast and @castelletto1, could you please take a final review of this PR, which is expected for InSalah case @herve-gross

rrsettgast (approved):

rrsettgast:
> Hello @rrsettgast and @castelletto1, could you please take a final review of this PR, which is expected for InSalah case @herve-gross

@jhuang2601 
I didn't review, but I approved it. I don't have coverage for my time on any GEOS tasks any longer.

src/coreComponents/mesh/WellElementSubRegion.cpp
```diff
@@ -154,15 +155,79 @@ bool isPointInsideElement( CellElementSubRegion const & subRegion,
...
+  unique_points( 0, PointHash< Point3d >(), PointsEqual< Point3d >( geomTol ));
```
bd713: In practice, I'm not sure the `unordered_set` will handle this sort of tolerance-based equality.
Two close coordinates might be seen equal within the tolerance, but still, I would expect both to generate two different hashes, meaning both will be seen as distinct, and kept in the list.

src/coreComponents/mesh/utilities/ComputationalGeometry.hpp
```diff
@@ -433,6 +433,120 @@ bool isPointInsidePolyhedron( arrayView2d< real64 const, nodes::REFERENCE_POSITI
...
+  for( integer i = 0; i < n; i++ )
+  {
+    auto const & p1 = polygon[i];
+    auto const & p2 = polygon[(i + 1) % n];
+
+    if((point[1] > std::min( p1[1], p2[1] )) &&
+       (point[1] <= std::max( p1[1], p2[1] )) &&
+       (point[0] <= std::max( p1[0], p2[0] )))
+    {
+      real64 const xIntersect = (point[1] - p1[1]) * (p2[0] - p1[0]) / (p2[1] - p1[1]) + p1[0];
```
bd713: Could you confirm the expected behavior/convention for a point on the boundary?
I have the impression that a point on a vertical left edge will be considered "outside", while a point on a right edge will be "inside". Same with a horizontal top edge (inside) vs. bottom (outside)

src/coreComponents/mesh/ElementSubRegionBase.hpp
```diff
@@ -279,14 +279,24 @@ class ElementSubRegionBase : public ObjectManagerBase
...
+
+      std::set< std::array< real64, 3 > > uniqueNodes;
+      uniqueNodes.insert( {X( n0, 0 ), X( n0, 1 ), X( n0, 2 )} );
+
       localIndex const numNodes = this->numNodesPerElement( k );
       for( localIndex a = 1; a < numNodes; ++a )
       {
-        LvArray::tensorOps::add< 3 >( elementCenters[k], X[e2n( k, a )] );
+        localIndex const nA = e2n( k, a );
+        auto const [it, inserted] = uniqueNodes.insert( {X[nA][0], X[nA][1], X[nA][2]} );
```
bd713: Should we add a small tolerance to better detect duplicated nodes?

PR #3361, opened 2024-09-20
feat: WaveEquation : new acoustic gradient formulation

rmadec-cs:
New Acoustic gradient formulation with respect to parameters Kappa=1/rho*c2 and Buoyancy=1/rho replacing old gradient formulation with respect to velocity c


sframba:
@acitrain @rrsettgast @CusiniM @wrtobin @corbett5 does any of you have time for a review of this PR? Thanks!

acitrain (approved):

acitrain:
Hi @rrsettgast,

Can you please final check this PR? 
Everything is good rebaseline included

Thank you !

rrsettgast (approved):

rrsettgast:
@rmadec-cs @acitrain 
baselines got updated under this. Gonna have to put baselines back to develop...run integratedTests again...then set the baseline to the new hash....but wait until you are first in line in the queue...otherwise you may have to do it again!

paveltomin (approved):

paveltomin:
> @rmadec-cs @acitrain baselines got updated under this. Gonna have to put baselines back to develop...run integratedTests again...then set the baseline to the new hash....but wait until you are first in line in the queue...otherwise you may have to do it again!

done

src/coreComponents/unitTests/wavePropagationTests/testWavePropagationAdjoint1.cpp
```diff
@@ -0,0 +1,404 @@
...
+
+  for( int i = 0; i < 3; i++ )
+  {
+    real64 tmp_double;
+    tmp_double=rCoord[0][i];
+    rCoord[0][i]=sCoord[0][i];
+    sCoord[0][i]=tmp_double;
+  }
+
+  sCoord.registerTouch( hostMemorySpace );
```
corbett5, line 279: Typically we don't want or need to explicitly call `registerTouch`. If you want to make sure they're on the CPU why not call `move`?

src/coreComponents/unitTests/wavePropagationTests/testWavePropagationAdjoint1.cpp
```diff
@@ -0,0 +1,404 @@
...
+  {}
+
+protected:
+
+  void SetUp() override
+  {
+    setupProblemFromXML( state.getProblemManager(), xmlInput );
+  }
+
+  static real64 constexpr time = 0.0;
```
corbett5, line 188: Probably would be easier to just make these global constexpr variables

corbett5 (approved):

src/coreComponents/unitTests/wavePropagationTests/testWavePropagationAdjoint1.cpp
CusiniM, line 1: It would be better to avoid adding any more unitTest like this one. All the tests in `coreComponents/unitTests` tend to need a lot of components and are not really unitTests but rather integratedTests implemented as if they were unitTests. All of these should slowly be removed and replaced by either a proper unitTest that lives in a specific component or by integratedTests using (or expanding) the curveCheck  capability. 

To me it looks like that even this test could be done either with a curvecheck or by adding a tests through the python interface.

rrsettgast:
@paveltomin I took this out of the queue due to @CusiniM request.

src/coreComponents/unitTests/wavePropagationTests/testWavePropagationAdjoint1.cpp
sframba: @CusiniM the integrated test structure today is quite limiting, as it does not allow to perform anything more than a standard GEOS run. Would it be better perhaps to extract all unit tests that fit the description you give above (e.g., those that contain a full XML input file at the top and go through the whole GEOS initialization chain) and transfer them to integrated tests, while retaining the google test structure? Kinda like "basic" or "preliminary" integrated tests.
This would also allow to increase the size of the tests to more than a handful of mesh elements, and thus the significance of the result. What do you guys think?

CusiniM: I think that it would be better not to have to maintain 2 different `integratedTests` workflows and run all of them through `ats`.  
Some of the `unitTests` in that folder build a full geos simulation to test derivatives. Ideally we would modularize more some of the functions that compute derivatives and come up with a system to mock some of the geos objects so that those can be converted into real unitTests.

What's missing from the current `integratedTests` to convert those waive propagation tests into `integratedTests`? We can probably add what's needed. It would also be good to start adding integratedTests that rely on the python interface + pygeos workflow since it's also what you guys mostly use.

PR #3382, opened 2024-10-02
fix: two hard crash fixes

paveltomin:
one is when all component concentrations are zeros - division by zero
and another is when maxTime is not specified for events - there is overflow when huge double is caster into int


src/coreComponents/constitutive/fluid/multifluid/MultiFluidBase.hpp
```diff
@@ -783,12 +783,14 @@ MultiFluidBase::KernelWrapper::
   real64 totalMolality = 0.0;
   for( integer ic = 0; ic < numComps; ++ic )
   {
+    GEOS_ERROR_IF( m_componentMolarWeight[ic] < LvArray::NumericLimits< real64 >::epsilon, "Zero component molecular weight" );
```
dkachuma: `m_componentMolarWeight` is user input so we should validate it in `postInputInitialization` [MultiFluidBase.cpp:184](https://github.com/GEOS-DEV/GEOS/blob/develop/src/coreComponents/constitutive/fluid/multifluid/MultiFluidBase.cpp#L184) instead of in this kernel function.

dkachuma (approved):

src/coreComponents/constitutive/fluid/multifluid/MultiFluidBase.hpp
paveltomin: moved the check there, thanks

paveltomin:
@corbett5, @cssherman, and/or @rrsettgast could you please review?

src/coreComponents/events/EventManager.cpp
```diff
@@ -54,7 +54,7 @@ EventManager::EventManager( string const & name,
     setDescription( "Start simulation time for the global event loop." );
 
   registerWrapper( viewKeyStruct::maxTimeString(), &m_maxTime ).
-    setApplyDefaultValue( std::numeric_limits< real64 >::max() ).
+    setApplyDefaultValue( std::numeric_limits< integer >::max() * units::YearSeconds ).
```
CusiniM: could even use a smaller number tbh.

src/coreComponents/constitutive/fluid/multifluid/MultiFluidBase.hpp
```diff
@@ -783,12 +783,14 @@ MultiFluidBase::KernelWrapper::
...
   for( integer ic = 0; ic < numComps; ++ic )
   {
+    // component weight can not be zero, checked in MultiFluidBase::postInputInitialization
     real64 const mwInv = 1.0 / m_componentMolarWeight[ic];
     compMoleFrac[ic] = composition[ic] * mwInv; // this is molality (units of mole/mass)
     dCompMoleFrac_dCompMassFrac[ic][ic] = mwInv;
     totalMolality += compMoleFrac[ic];
   }
 
+  GEOS_ERROR_IF( totalMolality < LvArray::NumericLimits< real64 >::epsilon, "Zero total molality, concentrations are zeros for all components" );
```
CusiniM:
```suggestion
  GEOS_ERROR_IF( totalMolality < LvArray::NumericLimits< real64 >::epsilon, "Zero total molality, all component concentrations are equal to zero." );
```

CusiniM (approved):

paveltomin:
@corbett5, @cssherman, and/or @rrsettgast please review, the change is minor and the issue appeared after FTE were enabled

src/coreComponents/constitutive/fluid/multifluid/MultiFluidBase.hpp
```diff
@@ -783,12 +783,14 @@ MultiFluidBase::KernelWrapper::
...
   for( integer ic = 0; ic < numComps; ++ic )
   {
+    // component weight can not be zero, checked in MultiFluidBase::postInputInitialization
     real64 const mwInv = 1.0 / m_componentMolarWeight[ic];
     compMoleFrac[ic] = composition[ic] * mwInv; // this is molality (units of mole/mass)
     dCompMoleFrac_dCompMassFrac[ic][ic] = mwInv;
     totalMolality += compMoleFrac[ic];
   }
 
+  GEOS_ERROR_IF( totalMolality < LvArray::NumericLimits< real64 >::epsilon, "Zero total molality, all component concentrations are equal to zero." );
```
rrsettgast, line 793: This may go bad prior to `totalMolality < LvArray::NumericLimits< real64 >::epsilon`?? If so perhaps a higher number here? Also can we have a graceful recovery somehow? For example if `totalMolality ~= 0` then that means that each `compMoleFrac[ic] ~= 0`, which means that `composition[ic] ~=0`? Should we put a limit on the minimum value of the primary variables? We are talking about void at this point?

src/coreComponents/events/EventManager.cpp
rrsettgast: Perhaps we shouldn't use a number based on `std::numeric_limits< T >::max()`. While the actual value of `std::numeric_limits< real64 >::max()` in seconds is around `5.7e300 years`, this seems like a lot given the age of the universe is `13e9 years`, and the age of the earth is `4e9` years. Perhaps a nice number would be `1e6 years = 3.1536e+13  seconds`?

rrsettgast (approved):

src/coreComponents/constitutive/fluid/multifluid/MultiFluidBase.hpp
paveltomin: all elements `composition` can not be zero, the issue originates somewhere in the initialization (from user input) where the check is missing

paveltomin:
@rrsettgast 
`Perhaps we shouldn't use a number based on std::numeric_limits< T >::max(). While the actual value of std::numeric_limits< real64 >::max() in seconds is around 5.7e300 years, this seems like a lot given the age of the universe is 13e9 years, and the age of the earth is 4e9 years. Perhaps a nice number would be 1e6 years = 3.1536e+13  seconds?`

the issue was that that maxTime get's converted into number of years as integer number and then overflows, that's why i put that fix
i am fine with any number there, just let me know which (CC @CusiniM)

CusiniM:
> @rrsettgast `Perhaps we shouldn't use a number based on std::numeric_limits< T >::max(). While the actual value of std::numeric_limits< real64 >::max() in seconds is around 5.7e300 years, this seems like a lot given the age of the universe is 13e9 years, and the age of the earth is 4e9 years. Perhaps a nice number would be 1e6 years = 3.1536e+13 seconds?`
> 
> the issue was that that maxTime get's converted into number of years as integer number and then overflows, that's why i put that fix i am fine with any number there, just let me know which (CC @CusiniM)

Maybe 1000 years?

paveltomin:
> > @rrsettgast `Perhaps we shouldn't use a number based on std::numeric_limits< T >::max(). While the actual value of std::numeric_limits< real64 >::max() in seconds is around 5.7e300 years, this seems like a lot given the age of the universe is 13e9 years, and the age of the earth is 4e9 years. Perhaps a nice number would be 1e6 years = 3.1536e+13 seconds?`
> > the issue was that that maxTime get's converted into number of years as integer number and then overflows, that's why i put that fix i am fine with any number there, just let me know which (CC @CusiniM)
> 
> Maybe 1000 years?

ok, 1000 years it is

src/coreComponents/events/EventManager.cpp
```diff
@@ -54,7 +54,7 @@ EventManager::EventManager( string const & name,
     setDescription( "Start simulation time for the global event loop." );
 
   registerWrapper( viewKeyStruct::maxTimeString(), &m_maxTime ).
-    setApplyDefaultValue( std::numeric_limits< real64 >::max() ).
+    setApplyDefaultValue( 1000 * units::YearSeconds ). // 1000 years
```
CusiniM: Let's do 10000 just in case.
```suggestion
    setApplyDefaultValue( 10000 * units::YearSeconds ). // 10000 years
```

PR #3383, opened 2024-10-02
feat: keep well variables constant during poromechanics initizaliation

paveltomin:
Useful for sequential, removes the need to create well rate tables with zero rate for t < 0


src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.cpp
tjb-ltk, line 1: In the event section if there are gaps in time a solver isn't called and if you setup a coupled solver but only define the reservoir solver in the events section. coupled solver isn't called.. I could be wrong on this... but if not, why isn't the problem decoupled by setting up different solvers and triggers via events..

paveltomin: one can not run pure solid mechanics solver/kernels for initialization because they don't have pressure contribution in them

paveltomin:
need review please

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -1450,13 +1450,13 @@ void CompositionalMultiphaseBase::applyBoundaryConditions( real64 const time_n,
...
 
-  if( m_keepFlowVariablesConstantDuringInitStep )
+  if( m_keepVariablesConstantDuringInitStep )
   {
     // this function is going to force the current flow state to be constant during the time step
     // this is used when the poromechanics solver is performing the stress initialization
     // TODO: in the future, a dedicated poromechanics kernel should eliminate the flow vars to construct a reduced system
     //       which will remove the need for this brittle passing aroung of flag
-    keepFlowVariablesConstantDuringInitStep( time_n, dt, dofManager, domain, localMatrix.toViewConstSizes(), localRhs.toView() );
+    keepVariablesConstantDuringInitStep( time_n, dt, dofManager, domain, localMatrix.toViewConstSizes(), localRhs.toView() );
```
CusiniM, line 1460: why the change of name? The goal is still to keep flow variables constant while updating displacement, isn't it?

src/coreComponents/physicsSolvers/fluidFlow/wells/WellSolverBase.cpp
```diff
@@ -198,20 +198,23 @@ void WellSolverBase::assembleSystem( real64 const time,
...
 
-  // then assemble the volume balance equations
-  assembleVolumeBalanceTerms( domain, dofManager, localMatrix, localRhs );
+    // then assemble the volume balance equations
+    assembleVolumeBalanceTerms( domain, dofManager, localMatrix, localRhs );
 
-  // then assemble the pressure relations between well elements
-  assemblePressureRelations( time, dt, domain, dofManager, localMatrix, localRhs );
+    // then assemble the pressure relations between well elements
+    assemblePressureRelations( time, dt, domain, dofManager, localMatrix, localRhs );
```
CusiniM: some weird whitespace. Uncrustify should fix it.

CusiniM (approved):

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
paveltomin: just for some symmetry with wells, prefer functions to be called the same so it is easier to find

src/coreComponents/physicsSolvers/fluidFlow/wells/WellSolverBase.cpp
paveltomin: i did run Uncrustify and the code style check is good
not sure what git is showing here

jhuang2601 (approved):

PR #3384, opened 2024-10-03
feat: plastic strain output

### LINKED ISSUE #3210: Elastic/Plastic strain outputs in GEOS

ryar9534:
Introducing a plasticStrain output field in addition to the current total strain. This is done by subtracting the elastic strain (obtained via the getElasticStrain method of the constitutive laws) from the total strain and averaging over quadrature points.

Rebaselining will be required because of the new field.

Ideally, plastic strain would only be output when a plasticity model is being used, but Im not sure how exactly we will do that...

I will run the Drucker-Prager and Modified Cam Clay wellbore problem to test


src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.cpp
```diff
@@ -160,7 +160,9 @@ void SolidMechanicsLagrangianFEM::registerDataOnMesh( Group & meshBodies )
     {
       setConstitutiveNamesCallSuper( subRegion );
 
+      subRegion.registerField< solidMechanics::strain >( getName() ).reference().resizeDimension< 1 >( 6 );
```
rrsettgast:
```suggestion
      subRegion.registerField< solidMechanics::plasticStrain >( getName() ).setDimLabels( 1, voightLabels ).reference().resizeDimension< 1 >( 6 );
```

rrsettgast (approved):
Do we want full tensorial plastic strain, or eps?

paveltomin (approved):

ryar9534:
We must consider the behavior when the following are present:
- [ ] Nonzero initial stress (either prescribed or solved for)
- [ ] Nonzero initial displacements
- [ ] Boundary condition effects

rrsettgast:
@ryar9534 Can you elaborate? Why do each of these things change the plastic strain calc?

> We must consider the behavior when the following are present:
> 
> * [ ]  Nonzero initial stress (either prescribed or solved for)

Is this concern because you have an initial stress in the plastic zone and you don't want to have that plastic strain as part of the reported value? I would think it still needs to be reported as it is the proper current state?

> * [ ]  Nonzero initial displacements

Is this concern that the initial state/strain and the displacement field are incompatible? 

> * [ ]  Boundary condition effects

Could we just interpret this as "your BC's are a problem, please use better BC's or move the boundary."?

ryar9534:
> @ryar9534 Can you elaborate? Why do each of these things change the plastic strain calc?
> 
> > We must consider the behavior when the following are present:
> > 
> > * [ ]   Nonzero initial stress (either prescribed or solved for)
> 
> Is this concern because you have an initial stress in the plastic zone and you don't want to have that plastic strain as part of the reported value? I would think it still needs to be reported as it is the proper current state?
> 
> > * [ ]   Nonzero initial displacements
> 
> Is this concern that the initial state/strain and the displacement field are incompatible?
> 
> > * [ ]   Boundary condition effects
> 
> Could we just interpret this as "your BC's are a problem, please use better BC's or move the boundary."?

These were mostly notes to myself to make sure various cases worked as I am figuring things out. Basically the complexity is that the total strain we are outputting is referenced  from the initial displacement, but the functions I am using to compute the elastic strain only know the stress. So I was seeing weird results where the two strains are sort of not compatible with one another, and thus you cant subtract them to get the plastic strain.

rrsettgast:
> These were mostly notes to myself to make sure various cases worked as I am figuring things out. Basically the complexity is that the total strain we are outputting is referenced from the initial displacement, but the functions I am using to compute the elastic strain only know the stress. So I was seeing weird results where the two strains are sort of not compatible with one another, and thus you cant subtract them to get the plastic strain.

Oh. Duh. Clearly that is a problem. So you would have to save the initial total strain/initial elastic strain or the initial stress. Nice thing is you could do this one quantity per element rather than per quadrature point.

ryar9534:
I think this is ready to re-discuss. I think this would work, the only weird case is when the initial state has a plastic strain component., but Im not sure how often that occurs.

src/coreComponents/constitutive/solid/ElasticIsotropicPressureDependent.hpp
```diff
@@ -223,6 +228,67 @@ void ElasticIsotropicPressureDependentUpdates::getElasticStrain( localIndex cons
...
+void ElasticIsotropicPressureDependentUpdates::getElasticStrainInc( localIndex const k,
+                                                                    localIndex const q,
+                                                                    real64 ( & elasticStrainInc)[6] ) const
+{
+  real64 const mu     = m_shearModulus[k];
+  real64 const p0     = m_refPressure;
+  real64 const eps_v0 = m_refStrainVol;
+  real64 const Cr     = m_recompressionIndex[k];
+  real64 deviator[6];
+  real64 stress[6];
```
CusiniM:
```suggestion
  real64 deviator[6]{};
  real64 stress[6]{};
```

src/coreComponents/constitutive/solid/ElasticIsotropicPressureDependent.hpp
```diff
@@ -223,6 +228,67 @@ void ElasticIsotropicPressureDependentUpdates::getElasticStrain( localIndex cons
...
+
+  elasticStrainVol = std::log( P/p0 ) * Cr * (-1.0) + eps_v0;
+  elasticStrainDev = Q/3./mu;
+
+  twoInvariant::strainRecomposition( elasticStrainVol,
+                                     elasticStrainDev,
+                                     deviator,
+                                     elasticStrainInc );
+
+  real64 oldStrain[6];
```
CusiniM:
```suggestion
  real64 oldStrain[6]{};
```

src/coreComponents/constitutive/solid/ElasticIsotropicPressureDependent.hpp
```diff
@@ -223,6 +228,67 @@ void ElasticIsotropicPressureDependentUpdates::getElasticStrain( localIndex cons
...
+  {
+    stress[i] = m_newStress[k][q][i];
+  }
+
+  twoInvariant::stressDecomposition( stress,
+                                     P,
+                                     Q,
+                                     deviator );
+
+  elasticStrainVol = std::log( P/p0 ) * Cr * (-1.0) + eps_v0;
```
CusiniM:
```suggestion
  real64 const elasticStrainVol = LvArray::math::log( P/p0 ) * Cr * (-1.0) + eps_v0;
```

src/coreComponents/constitutive/solid/ElasticIsotropicPressureDependent.hpp
```diff
@@ -223,6 +228,67 @@ void ElasticIsotropicPressureDependentUpdates::getElasticStrain( localIndex cons
...
+{
+  real64 const mu     = m_shearModulus[k];
+  real64 const p0     = m_refPressure;
+  real64 const eps_v0 = m_refStrainVol;
+  real64 const Cr     = m_recompressionIndex[k];
+  real64 deviator[6];
+  real64 stress[6];
+  real64 P;
+  real64 Q;
+  real64 elasticStrainVol;
```
CusiniM:
```suggestion
```

src/coreComponents/constitutive/solid/ElasticIsotropicPressureDependent.hpp
```diff
@@ -223,6 +228,67 @@ void ElasticIsotropicPressureDependentUpdates::getElasticStrain( localIndex cons
...
+    stress[i] = m_newStress[k][q][i];
+  }
+
+  twoInvariant::stressDecomposition( stress,
+                                     P,
+                                     Q,
+                                     deviator );
+
+  elasticStrainVol = std::log( P/p0 ) * Cr * (-1.0) + eps_v0;
+  elasticStrainDev = Q/3./mu;
```
CusiniM:
```suggestion
  real64 const elasticStrainDev = Q/3./mu;
```

src/coreComponents/constitutive/solid/ElasticIsotropicPressureDependent.hpp
```diff
@@ -223,6 +228,67 @@ void ElasticIsotropicPressureDependentUpdates::getElasticStrain( localIndex cons
...
+  real64 const mu     = m_shearModulus[k];
+  real64 const p0     = m_refPressure;
+  real64 const eps_v0 = m_refStrainVol;
+  real64 const Cr     = m_recompressionIndex[k];
+  real64 deviator[6];
+  real64 stress[6];
+  real64 P;
+  real64 Q;
+  real64 elasticStrainVol;
+  real64 elasticStrainDev;
```
CusiniM:
```suggestion
```

src/coreComponents/constitutive/solid/ElasticIsotropicPressureDependent.hpp
```diff
@@ -223,6 +228,67 @@ void ElasticIsotropicPressureDependentUpdates::getElasticStrain( localIndex cons
...
+  twoInvariant::strainRecomposition( elasticStrainVol,
+                                     elasticStrainDev,
+                                     deviator,
+                                     elasticStrainInc );
+
+  real64 oldStrain[6];
+  for( localIndex i=0; i<6; ++i )
+  {
+    stress[i] = m_oldStress[k][q][i];
+  }
```
CusiniM:
```suggestion
  LvArray::tensorOps::copy<6>( stress, m_oldStress[k][q] );
```
?

src/coreComponents/constitutive/solid/ElasticIsotropicPressureDependent.hpp
```diff
@@ -223,6 +228,67 @@ void ElasticIsotropicPressureDependentUpdates::getElasticStrain( localIndex cons
...
+
+  twoInvariant::stressDecomposition( stress,
+                                     P,
+                                     Q,
+                                     deviator );
+
+  elasticStrainVol = std::log( P/p0 ) * Cr * (-1.0) + eps_v0;
```
CusiniM:
```suggestion
  real64 const elasticStrainVol = LvArray::math::log( P/p0 ) * Cr * (-1.0) + eps_v0;
```

src/coreComponents/constitutive/solid/ElasticIsotropicPressureDependent.hpp
```diff
@@ -223,6 +228,67 @@ void ElasticIsotropicPressureDependentUpdates::getElasticStrain( localIndex cons
...
+  elasticStrainDev = Q/3./mu;
```
CusiniM, line 271:
```suggestion
  real64 const elasticStrainDev = Q/3./mu;
```

src/coreComponents/constitutive/solid/SolidBase.hpp
```diff
@@ -337,6 +337,24 @@ class SolidBaseUpdates
...
+   */
+  GEOS_HOST_DEVICE
+  virtual void getElasticStrainInc( localIndex const k,
+                                    localIndex const q,
+                                    real64 ( & elasticStrainInc )[6] ) const
+  {
+    GEOS_UNUSED_VAR( k );
+    GEOS_UNUSED_VAR( q );
+    GEOS_UNUSED_VAR( elasticStrainInc );
+    GEOS_ERROR( "getElasticStrainInc() not implemented for this model (called when computing plastic strains)" );
```
CusiniM:
```suggestion
    GEOS_ERROR( "getElasticStrainInc() of SolidBase was called." );
```

CusiniM (approved):

src/coreComponents/constitutive/solid/ElasticIsotropicPressureDependent.hpp
ryar9534: should I change the log in the existing getElasticStrain as well?

src/coreComponents/constitutive/solid/ElasticIsotropicPressureDependent.hpp
ryar9534: should I change the log in the existing getElasticStrain as well?

sytuannguyen (approved):

FredericgB:
Hello,
I cannot login to my github account , the two authentication steps failed every time.
Can we set up a meeting for Friday late afternoon to discuss?
Thanks
Frederic


***@***.***
Frederic Bourgeois
OneTech
OT/R&D/SU/CCS/STO
Geomechanics Manager
TotalEnergies CSTJF
Avenue Larribau, 64 018 - Pau cedex - France

src/coreComponents/constitutive/solid/ElasticIsotropic.hpp
```diff
@@ -228,6 +233,23 @@ void ElasticIsotropicUpdates::getElasticStrain( localIndex const k,
...
+                                                   real64 ( & elasticStrainInc)[6] ) const
+{
+  real64 const E = conversions::bulkModAndShearMod::toYoungMod( m_bulkModulus[k], m_shearModulus[k] );
+  real64 const nu = conversions::bulkModAndShearMod::toPoissonRatio( m_bulkModulus[k], m_shearModulus[k] );
+
+  elasticStrainInc[0] = (    (m_newStress[k][q][0] - m_oldStress[k][q][0]) - nu*(m_newStress[k][q][1] - m_oldStress[k][q][1]) - nu*(m_newStress[k][q][2] - m_oldStress[k][q][2]))/E;
+  elasticStrainInc[1] = (-nu*(m_newStress[k][q][0] - m_oldStress[k][q][0]) +    (m_newStress[k][q][1] - m_oldStress[k][q][1]) - nu*(m_newStress[k][q][2] - m_oldStress[k][q][2]))/E;
+  elasticStrainInc[2] = (-nu*(m_newStress[k][q][0] - m_oldStress[k][q][0]) - nu*(m_newStress[k][q][1] - m_oldStress[k][q][1]) +    (m_newStress[k][q][2] - m_oldStress[k][q][2]))/E;
+
+  elasticStrainInc[3] = (m_newStress[k][q][3] - m_oldStress[k][q][3]) / m_shearModulus[k];
```
jhuang2601: This is the formula for engineering shear strain, which should be twice as that of tensorial shear strain.

src/coreComponents/constitutive/solid/ElasticIsotropic.hpp
```diff
@@ -228,6 +233,23 @@ void ElasticIsotropicUpdates::getElasticStrain( localIndex const k,
...
+  elasticStrainInc[4] = (m_newStress[k][q][4] - m_oldStress[k][q][4]) / m_shearModulus[k];
```
jhuang2601: see above

src/coreComponents/constitutive/solid/ElasticIsotropic.hpp
```diff
@@ -228,6 +233,23 @@ void ElasticIsotropicUpdates::getElasticStrain( localIndex const k,
...
+  elasticStrainInc[5] = (m_newStress[k][q][5] - m_oldStress[k][q][5]) / m_shearModulus[k];
```
jhuang2601: see above

BASELINE_NOTES.md
```diff
@@ -6,6 +6,10 @@ This file is designed to track changes to the integrated test baselines.
 Any developer who updates the baseline ID in the .integrated_tests.yaml file is expected to create an entry in this file with the pull request number, date, and their justification for rebaselining.
 These notes should be in reverse-chronological order, and use the following time format: (YYYY-MM-DD).
 
+PR #3384 (2024-12-11)
+=====================
+Added plastic strain output.
+
```
FredericgB, line 12: this is a test

src/coreComponents/constitutive/solid/ElasticIsotropic.hpp
CusiniM: yes, but I think that's what it should be because we are subtracting it from the `totalStrain` which is in engineering notation. What we may want to do is divide the shear components of the `plasticStrian` by 2 so that it is the actual strain. I don't know if `Paraview` expects it in engineering notation though.

src/coreComponents/constitutive/solid/ElasticIsotropic.hpp
```diff
@@ -135,6 +135,11 @@ class ElasticIsotropicUpdates : public SolidBaseUpdates
                                  localIndex const q,
                                  real64 ( &elasticStrain )[6] ) const override final;
 
+  GEOS_HOST_DEVICE
+  virtual void getElasticStrainInc( localIndex const k,
```
rrsettgast, line 139:
```suggestion
  virtual void calculateElasticStrainInc( localIndex const k,
```

Since this isn't really a "getter"

src/coreComponents/constitutive/solid/ElasticIsotropic.hpp
```diff
@@ -228,6 +233,23 @@ void ElasticIsotropicUpdates::getElasticStrain( localIndex const k,
   elasticStrain[5] = m_newStress[k][q][5] / m_shearModulus[k];
 }
 
+GEOS_HOST_DEVICE
+inline
+void ElasticIsotropicUpdates::getElasticStrainInc( localIndex const k,
```
rrsettgast, line 260:
```suggestion
void ElasticIsotropicUpdates::calculateElasticStrainInc( localIndex const k,
```

src/coreComponents/constitutive/solid/ElasticIsotropic.hpp
rrsettgast: was this question resolved? ChatGPT says paraview expects tensorial strain...but I think it came to that conclusion because it calculates tensorial strain.

rrsettgast (commented):
Just thinking that the total and incremental calculations are the same except for the stress inputs. These could be put into a private function so that the calculation is only done in one place.

ryar9534:
> was this question resolved? ChatGPT says paraview expects tensorial strain...but I think it came to that conclusion because it calculates tensorial strain.

To me, it seems like it would be weird if paraview did operations to a tensor field, as IDK how it would know to do it for strain and not everything else. Thoughts @CusiniM ? I thought we were just gonna merge it

paveltomin:
can someone check if new baselines are ok?
i see things like that:
```
Error: /Problem/domain/MeshBodies/mesh1/meshLevels/Level0/ElementRegions/elementRegionsGroup/Domain/elementSubRegions/cb1/strain
	Arrays of types float64 and float64 have 378 values of which 63 fail both the relative and absolute tests.
		Max absolute difference is at index (np.int64(30), np.int64(5)): value = 7.644588483567542e-05, base_value = 0.00015289176967135085
		Max relative difference is at index (np.int64(30), np.int64(5)): value = 7.644588483567542e-05, base_value = 0.00015289176967135085
	Statistics of the q values greater than 1.0 defined by absolute tolerance: N = 63
		max = 7644.588483567542, mean = 124.862153686749, std = 711.8661761259043
		max is at index (np.int64(30), np.int64(5)), value = 7.644588483567542e-05, base_value = 0.00015289176967135085
	Statistics of the q values greater than 1.0 defined by relative tolerance: N = 0
```
is it because plastic part is separate now?

jhuang2601:
> can someone check if new baselines are ok? i see things like that:
> 
> ```
> Error: /Problem/domain/MeshBodies/mesh1/meshLevels/Level0/ElementRegions/elementRegionsGroup/Domain/elementSubRegions/cb1/strain
> 	Arrays of types float64 and float64 have 378 values of which 63 fail both the relative and absolute tests.
> 		Max absolute difference is at index (np.int64(30), np.int64(5)): value = 7.644588483567542e-05, base_value = 0.00015289176967135085
> 		Max relative difference is at index (np.int64(30), np.int64(5)): value = 7.644588483567542e-05, base_value = 0.00015289176967135085
> 	Statistics of the q values greater than 1.0 defined by absolute tolerance: N = 63
> 		max = 7644.588483567542, mean = 124.862153686749, std = 711.8661761259043
> 		max is at index (np.int64(30), np.int64(5)), value = 7.644588483567542e-05, base_value = 0.00015289176967135085
> 	Statistics of the q values greater than 1.0 defined by relative tolerance: N = 0
> ```
> 
> is it because plastic part is separate now?

Which case are you referring to? Plastic strain is registered as a new field (root cause for the failure) and the total strain is not touched in this PR.

paveltomin:
> > is it because plastic part is separate now?
> 
> Which case are you referring to? Plastic strain is registered as a new field (root cause for the failure) and the total strain is not touched in this PR.

for example this
image

ryar9534:
I did touch the total strain in this pr. We decided to divide the shear by 2 following discussions about engineering vs tensor strain. Hence the new baselines.

joshua-white:
> I did touch the total strain in this pr. We decided to divide the shear by 2 following discussions about engineering vs tensor strain. Hence the new baselines.

A large number of functions in the plasticity routine were written assuming engineering strains.  I would definitely double check none of the physical results have changed before rebaselining anything.  Would be very easy to make a mistake here.  The error posted above has a factor of 2x in it, though, so that's encouraging.

ryar9534:
Yes, all the computations seem to be done in engineering strain. @CusiniM and I decided to convert to tensor strain just for output.

jhuang2601:
> Yes, all the computations seem to be done in engineering strain. @CusiniM and I decided to convert to tensor strain just for output.

That sounds good. Let me rerun the tutorial plastic example with strain hardening to confirm it.

jhuang2601:
> > I did touch the total strain in this pr. We decided to divide the shear by 2 following discussions about engineering vs tensor strain. Hence the new baselines.
> 
> A large number of functions in the plasticity routine were written assuming engineering strains. I would definitely double check none of the physical results have changed before rebaselining anything. Would be very easy to make a mistake here. The error posted above has a factor of 2x in it, though, so that's encouraging.

I can reproduce [this tutorial example](https://geosx-geosx.readthedocs-hosted.com/en/latest/docs/sphinx/advancedExamples/validationStudies/wellboreProblems/edpWellbore/Example.html?_sm_au_=iMV6tT998TLJ4JfNQ0WpHK6H8sjL6) with the PR, which confirms that correct engineering strains are utilized by the Extended Drucker-Prager Model.

jhuang2601 (approved):

Issue #3355, opened 2024-09-13
bug in testLifoStorage


rrsettgast:
**Describe the bug**
Occasionally there is an error in the CI on `testLifoStorage`. It is not reproducible, and rerunning will pass most of the time. However, this isn't a good thing to have laying around. It can be seen here:

https://github.com/GEOS-DEV/GEOS/actions/runs/10854274728/job/30124368903?pr=3340

The output for the failed `testLifoStorage` is here:
```
 90/211 Test  #90: testLifoStorage ......................................Subprocess aborted***Exception:   0.71 sec
[==========] Running 7 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 7 tests from LifoStorageTest
[ RUN      ] LifoStorageTest.LifoStorageBufferOnCUDA
Allocated    40.0 B to the HOST  : LvArray::Array<float, 1, camp::int_seq<long, 0l>, int, LvArray::ChaiBuffer>  Free memory on device: 10.9 GB
Allocated    40.0 B to the DEVICE: LvArray::Array<float, 1, camp::int_seq<long, 0l>, int, LvArray::ChaiBuffer>  Free memory on device: 10.9 GB
Moved    40.0 B to the DEVICE: LvArray::Array<float, 1, camp::int_seq<long, 0l>, int, LvArray::ChaiBuffer> 
 LIFO : maximum size 10 buffers 
 LIFO : buffer size 3.8147e-05MB
 LIFO : allocating 3 buffers on host
 LIFO : allocating 2 buffers on device
Allocated   120.0 B to the HOST  : LvArray::Array<float, 2, camp::int_seq<long, 0l, 1l>, long, LvArray::ChaiBuffer>  Free memory on device: 10.9 GB
Allocated    80.0 B to the DEVICE: LvArray::Array<float, 2, camp::int_seq<long, 0l, 1l>, long, LvArray::ChaiBuffer>  Free memory on device: 10.9 GB
Freed    80.0 B to the DEVICE: LvArray::Array<float, 2, camp::int_seq<long, 0l, 1l>, long, LvArray::ChaiBuffer>  Free memory on device: 10.9 GB
Freed   120.0 B to the HOST  : LvArray::Array<float, 2, camp::int_seq<long, 0l, 1l>, long, LvArray::ChaiBuffer>  Free memory on device: 10.9 GB
[...]

PR #3387, opened 2024-10-04
fix: testLifoStorage CI failures

sframba:
The `testLifoStorage` unit tests still occasionally fail (issue #3355).
After #3362, fails seem to only happen in the `LifoStorageBufferOnCUDANoDeviceBuffer` test. This is a corner case as it uses a zero-size device buffer which might be buggy. 
To help smooth the CI process, I propose disabling this specific test for now, pending further review.

### DIFF
--- a/src/coreComponents/common/unitTests/testLifoStorage.cpp
+++ b/src/coreComponents/common/unitTests/testLifoStorage.cpp
@@ -200,7 +200,7 @@ TEST( LifoStorageTest, LifoStorageBufferOnCUDAlargeAutoSizeBoth )
 // that affects the current CI (see issue https://github.com/GEOS-DEV/GEOS/issues/3355).
 // The error appears randomly on some configurations, and is probably related to the
 // size-0 device buffer case that this test covers. This case should not be frequent
-// in practise, but the issue should be resolved and the test reactivated as soon as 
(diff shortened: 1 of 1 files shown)


jiemeng-total (approved):

rrsettgast (approved):
I suggest just commenting out the test, including a comment indicating the problem, and the situations that this tests is providing coverage for.

rrsettgast:
@wrtobin Do you have any comment on this? I think it still needs to be fixed, but this functionality is for the Makutu team, so it won't affect the rest of the code.

wrtobin:
I agree that commenting out and documenting the test (with a link back to this MR or an open issue) would be fine here. 

We still want it to be fixed of course, but since the FIFO storage stuff is only used for the Makutu team it should be fine. One would hope that in use the FIFO storage wouldn't use size-0 buffers on device (since that seems counter to the intent of the container), though having that case not be catastrophic would certainly be preferable. Further, if the intent in practical use is actually never to have a size 0 buffer on device with the LIFO container, we could add a runtime check and error out if it actually happens.

sframba:
I'm going to merge this since it will help gain some time in handling the merge queue (it will avoid some CI re-runs, hopefully)

PR #3388, opened 2024-10-04
feat: Enable fully coupled mode for CompositionalMultiphaseReservoirPoromechanics and SinglePhaseReservoirPoromechanics

paveltomin:
should go after https://github.com/GEOS-DEV/GEOS/pull/3422

- works equivalent to `MultiphasePoromechanicsReservoir` and `SinglePhasePoromechanicsReservoir`, just more convenient to switch between FIM and SEQ within the same solver
- can use the same MGR recipe (order of block is actually the same: mechanics, flow, wells)


paveltomin:
need review please

CusiniM:
I am lost. What are these? What s the difference between these and  `MultiphasePoromechanicsReservoir` and `SinglePhasePoromechanicsReservoir` ?

paveltomin:
> I am lost. What are these? What s the difference between these and `MultiphasePoromechanicsReservoir` and `SinglePhasePoromechanicsReservoir` ?

`CompositionalMultiphaseReservoirPoromechanics` and `SinglePhaseReservoirPoromechanics` were created for sequential since `MultiphasePoromechanicsReservoir` and `SinglePhasePoromechanicsReservoir` are designed differently: they are flow/mech + wells so when one runs them in sequential they will do flow/mech solve and wells solve iteratively.
`CompositionalMultiphaseReservoirPoromechanics` and `SinglePhaseReservoirPoromechanics` were not be able to run in FIM mode which is addressed in this PR.
I know this is confusing and ultimately we should clean this up.

jhuang2601 (approved):

src/coreComponents/physicsSolvers/multiphysics/MultiphasePoromechanics.cpp
```diff
@@ -296,6 +335,33 @@ void MultiphasePoromechanics<>::setMGRStrategy()
...
+  }
+  else
+  {
+    // add Reservoir
+    linearSolverParameters.mgr.strategy = LinearSolverParameters::MGR::StrategyType::multiphasePoromechanicsReservoirFVM;
+  }
+  GEOS_LOG_LEVEL_RANK_0( 1, GEOS_FMT( "{}: MGR strategy set to {}", getName(),
+                                      EnumStrings< LinearSolverParameters::MGR::StrategyType >::toString( linearSolverParameters.mgr.strategy )));
+  // TODO check if needed
+  //linearSolverParameters.mgr.separateComponents = true;
```
victorapm: This should be needed, but next line can be removed

src/coreComponents/physicsSolvers/multiphysics/SinglePhasePoromechanics.cpp
```diff
@@ -139,6 +132,33 @@ void SinglePhasePoromechanics<>::setMGRStrategy()
...
+  }
+  else
+  {
+    // add Reservoir
+    linearSolverParameters.mgr.strategy = LinearSolverParameters::MGR::StrategyType::singlePhasePoromechanicsReservoirFVM;
+  }
+  GEOS_LOG_LEVEL_RANK_0( 1, GEOS_FMT( "{}: MGR strategy set to {}", this->getName(),
+                                      EnumStrings< LinearSolverParameters::MGR::StrategyType >::toString( linearSolverParameters.mgr.strategy )));
+  // TODO check if needed
+  //linearSolverParameters.mgr.separateComponents = true;
```
victorapm: Same comment as before

victorapm (approved):
Thanks!

CusiniM (commented):
Shouldn't we remove the old solvers though? I would only keep this one with the two options in the same solver.

CusiniM (approved):

CusiniM (changes requested):
Actually I would strongly prefer getting rid of `MultiphasePoromechanicsReservoir` and `SinglePhasePoromechanicsReservoir`

paveltomin:
> Actually I would strongly prefer getting rid of `MultiphasePoromechanicsReservoir` and `SinglePhasePoromechanicsReservoir`

One reason they can be useful is to perform an isolated well solve to precondition the nonlinear system, that's why I am hesitant to remove them

CusiniM (approved):

PR #3394, opened 2024-10-11
feat: Add new field specification application prior to SolverBase::solverStep in SolverBase::execute

rrsettgast:
This PR adds the application of `PhysicsSolver` specific `FieldSpecification` prior to the `solverStep`

- [ ] add new `FieldSpecification` prior to the start of the `SolverBase::solverStep`
- [ ] Deprecate `isInitialCondition` flag and replace with a list of initial conditions....somewhere to be determined. The list can be filled by the `isInitialCondition` flag for now, prior to its removal.


CusiniM (approved):
instead of specifying the list of names can't we have the solver apply all field specifications on certain fields like we do for boundary conditions?

rrsettgast:
> instead of specifying the list of names can't we have the solver apply all field specifications on certain fields like we do for boundary conditions?

No. Then we can't target specifications for the solver. We should actually go the other way and have specific lists of BC's to be applied by the physics packages. For example, if we have a displacement field specificaiton, the solid mechanics solver will always apply it. So we should list Dirichlet, Neumann, etc bc's to apply. So we can think of it as the physics package can specify what `FieldSpecification` objects to apply as a BC. 

The application in this PR is simply "overwrite data before the step begins"

herve-gross (approved):

PR #3395, opened 2024-10-11
feat: MGR Strategy for ALM - Face Bubble Functions for Wedge Elements - Multipoint Integration Rules for Tetrahedra and Triangles

matteofrigo5:
This PR introduces the following enhancements:

- [x] -     Two solving strategies for ALM: one utilizing MGR and the other employing AMG on the full system.
- [x] -     Implementation of face bubble functions for wedge elements.
- [x] -     An optional feature to utilize Multipoint Integration Rules for both tetrahedra and triangles.


src/coreComponents/constitutive/contact/CoulombFriction.hpp
```diff
@@ -380,9 +377,9 @@ inline void CoulombFrictionUpdates::updateTraction( arraySlice1d< real64 const >
 
   // Compute the trial traction
   real64 tractionTrial[ 3 ];
-  tractionTrial[ 0 ] = traction[0] + penalty[0] * dispJump[0] * faceArea;
-  tractionTrial[ 1 ] = traction[1] + penalty[1] * (dispJump[1] - oldDispJump[1]) * faceArea;
-  tractionTrial[ 2 ] = traction[2] + penalty[1] * (dispJump[2] - oldDispJump[2]) * faceArea;
+  tractionTrial[ 0 ] = traction[0] + penalty[0] * dispJump[0];
```
rrsettgast, line 380: so this "traction" is now a "traction"?

src/coreComponents/constitutive/contact/FrictionBase.hpp
```diff
@@ -134,7 +133,7 @@ class FrictionBaseUpdates
                                real64 ( & tractionNew )[3],
                                integer & fractureState ) const
   {
-    GEOS_UNUSED_VAR( oldDispJump, dispJump, penalty, traction, faceArea, symmetric, fixedLimitTau,
+    GEOS_UNUSED_VAR( oldDispJump, dispJump, penalty, traction, symmetric, fixedLimitTau,
```
rrsettgast, line 136: why is this function defined?

src/coreComponents/constitutive/contact/FrictionBase.hpp
```diff
@@ -153,9 +152,8 @@ class FrictionBaseUpdates
                                    arraySlice1d< real64 const > const & deltaDispJump,
                                    arraySlice1d< real64 const > const & penalty,
                                    arraySlice1d< real64 const > const & traction,
-                                   real64 const faceArea,
                                    arraySlice1d< real64 > const & tractionNew ) const
-  { GEOS_UNUSED_VAR( dispJump, deltaDispJump, penalty, traction, faceArea, tractionNew ); }
+  { GEOS_UNUSED_VAR( dispJump, deltaDispJump, penalty, traction, tractionNew ); }
```
rrsettgast, line 156: or this one? why not a pure virtual?

src/coreComponents/finiteElement/elementFormulations/H1_Tetrahedron_Lagrange1_Gauss1.hpp
```diff
@@ -63,7 +65,13 @@ class H1_Tetrahedron_Lagrange1_Gauss1 final : public FiniteElementBase
   constexpr static localIndex maxSupportPoints = numNodes;
 
   /// The number of quadrature points per element.
+#if TETRAHEDRON_QUADRATURE_POINTS == 14
+  constexpr static localIndex numQuadraturePoints = 14;
+#elif TETRAHEDRON_QUADRATURE_POINTS == 5
+  constexpr static localIndex numQuadraturePoints = 5;
+#elif TETRAHEDRON_QUADRATURE_POINTS == 1
```
rrsettgast: This is pretty clunky. There should be separate class for the different integration rules.

src/coreComponents/finiteElement/elementFormulations/H1_TriangleFace_Lagrange1_Gauss1.hpp
```diff
@@ -29,6 +29,8 @@ namespace geos
 namespace finiteElement
 {
 
+#define TRIANGLE_QUADRATURE_POINTS 4
```
rrsettgast: same comment here about a separate class.

src/coreComponents/physicsSolvers/contact/ContactFields.hpp
```diff
@@ -87,8 +87,16 @@ DECLARE_FIELD( slip,
                array1d< real64 >,
                0,
                LEVEL_0,
-               NO_WRITE,
-               "Slip." );
+               WRITE_AND_READ,
+               "Slip" );
+
+DECLARE_FIELD( tangentialTraction,
```
rrsettgast, line 93: Are we storing a "traction" and a "tangentialTraction"?

src/coreComponents/physicsSolvers/contact/SolidMechanicsAugmentedLagrangianContact.hpp
```diff
@@ -239,18 +239,50 @@ class SolidMechanicsAugmentedLagrangianContact : public ContactSolverBase
...
+
+    constexpr static char const * simultaneousString() { return "simultaneous"; }
+
+    constexpr static char const * symmetricString() { return "symmetric"; }
+
+    constexpr static char const * iterativePenaltyNFacString() { return "iterPenaltyN"; }
+
+    constexpr static char const * iterativePenaltyTFacString() { return "iterPenaltyT"; }
+
+    constexpr static char const * tolJumpDispNFacString() { return "tolJumpN"; }
```
rrsettgast, line 251: should these be dependent on the contact parameters?

src/coreComponents/physicsSolvers/contact/kernels/SolidMechanicsALMSimultaneousKernels.hpp
```diff
@@ -233,18 +233,19 @@ class ALMSimultaneous :
...
+    stack.localPenalty[2][2] = -m_penalty( k, 3 )*fac;
+    stack.localPenalty[1][2] = -m_penalty( k, 4 )*fac;
+    stack.localPenalty[2][1] = -m_penalty( k, 4 )*fac;
 
     for( int i=0; i<numTdofs; ++i )
     {
       stack.tLocal[i] = m_traction( k, i );
-      stack.dispJumpLocal[i] = m_dispJump( k, i );
-      stack.oldDispJumpLocal[i] = m_oldDispJump( k, i );
+      stack.dispJumpLocal[i] = m_dispJump( k, i )*m_faceArea[k];
```
rrsettgast: this seems odd to me that you would need to multiply by an area here.

src/coreComponents/physicsSolvers/contact/kernels/SolidMechanicsALMSimultaneousKernels.hpp
```diff
@@ -276,9 +277,9 @@ class ALMSimultaneous :
 
     // Compute the trial traction
     real64 dispJump[ 3 ];
-    dispJump[0] = stack.dispJumpLocal[0] * m_faceArea[k];
-    dispJump[1] = ( stack.dispJumpLocal[1] - stack.oldDispJumpLocal[1] ) * m_faceArea[k];
-    dispJump[2] = ( stack.dispJumpLocal[2] - stack.oldDispJumpLocal[2] ) * m_faceArea[k];
+    dispJump[0] = stack.dispJumpLocal[0];
```
rrsettgast, line 279: and here

src/coreComponents/physicsSolvers/contact/kernels/SolidMechanicsALMSimultaneousKernels.hpp
```diff
@@ -423,17 +424,16 @@ struct ComputeTractionSimultaneousKernel
...
           arrayView2d< real64 const > const & dispJump,
           arrayView2d< real64 const > const & deltaDispJump,
-          arrayView1d< real64 const > const & faceElementArea,
           arrayView2d< real64 > const & tractionNew )
   {
 
     forAll< POLICY >( size, [=] GEOS_HOST_DEVICE ( localIndex const kfe )
     {
-      tractionNew[kfe][0] = traction[kfe][0] + penalty[kfe][0] * dispJump[kfe][0] * faceElementArea[kfe];
+      tractionNew[kfe][0] = traction[kfe][0] + penalty[kfe][0] * dispJump[kfe][0];
```
rrsettgast, line 436: I liked the old way here. This is less intuitive...but maybe I don't understand the reason you don't want face area?

rrsettgast (commented):
just a few questions

src/coreComponents/constitutive/contact/FrictionBase.hpp
CusiniM: because these are device functions. I am pretty sure that nvcc throws an error if we use a pure virtual. We do this a lot in our `kernelWrapper` objects (probably with the idea of getting nicer compilation errors) but, tbh, I start thinking that it would be better not to use `virtual` at all for these.

src/coreComponents/physicsSolvers/contact/kernels/SolidMechanicsALMSimultaneousKernels.hpp
CusiniM: I am also a bit confused by these changes which I am guessing are there because of some of the changes I had made in the past but I am now a bit lost. Ideally things should always be what they are called so a traction should be a traction and not its surface integral and the `dispJump` should have the units of a `length` and not of  `length^3`.  Now, here it obviously depends on the units of that penalty but I am also confused by why above there is a jump defined as the jump times the area.

matteofrigo5: I agree that the current implementation may not be as intuitive as desired. Even if it is technically correct, I believe we can achieve a cleaner solution. I want to clarify why we need to scale the penalty matrix. 

We have the following relationship for computing the displacement jump:
`dispJump * faceArea = averageDispJump = Atu * displacement`
In the balance equation, the additional contribution from the Augmented Lagrangian formulation is given by:
`Aut * (tractionOld + penalty * dispJump) = Aut * penalty * Atu * (1/faceArea) * u + Aut tractionOld`

It’s important to note the presence of the term `(1/faceArea)` in this equation. I believe it would be cleaner to apply the scaling only to this specific term.

If I scaled the penalty matrix during setup, it would necessitate adjusting the `dispJump` by the `faceArea` to ensure consistency in the traction update, resulting in:
`traction = tractionOld + scaledPenalty * dispJump * faceArea`

By scaling only the additional term `Aut * penalty * Atu`, we can simplify the implementation and avoid additional adjustments elsewhere.

src/coreComponents/constitutive/contact/CoulombFriction.hpp
matteofrigo5: Yes, it is . Actually, it has always been a "traction". In the previous implementation, `dispJump` was multiplied by `faceArea`. Now that `dispJump` is as it should be, it is used in its original form.

src/coreComponents/finiteElement/elementFormulations/H1_Tetrahedron_Lagrange1_Gauss1.hpp
matteofrigo5: I agree. It was just a temporary implementation before discussing the creation of a separate class. Would you like to implement it in this PR, or should we open a specific one for this purpose?

src/coreComponents/physicsSolvers/contact/ContactFields.hpp
matteofrigo5: It was just for output purposes. Having the absolute value of tangential traction is convenient without needing two separate tangential components.

src/coreComponents/physicsSolvers/contact/SolidMechanicsAugmentedLagrangianContact.hpp
matteofrigo5: As far as the default values are concerned, I would say yes. However, it is convenient to have these tolerances to fine-tune the physics solver for the specific problem.

src/coreComponents/physicsSolvers/contact/kernels/SolidMechanicsALMSimultaneousKernels.hpp
matteofrigo5: See comment below.

src/coreComponents/finiteElement/elementFormulations/H1_Tetrahedron_Lagrange1_Gauss1.hpp
rrsettgast: We should redefine this class 
```c++
template< int NUM_Q_POINTS >
class H1_Tetrahedron_Lagrange1_Gauss final : public FiniteElementBase
{
...
};


using H1_Tetrahedron_Lagrange1_Gauss1 = H1_Tetrahedron_Lagrange1_Gauss<1>;
...
using H1_Tetrahedron_Lagrange1_Gauss14 = H1_Tetrahedron_Lagrange1_Gauss<14>;

```

src/coreComponents/finiteElement/elementFormulations/H1_Tetrahedron_Lagrange1_Gauss.hpp
```diff
@@ -38,15 +38,16 @@ namespace finiteElement
...
+ *           | \   \_           t            1      1   0   0
  *           |  +__  \_         |   s        2      0   1   0
  *           | /2  \__ \_       |  /         3      0   0   1
  *           |/       \__\      | /          =====  ==  ==  ==
  *           +------------+     *------r
  *          0              1
  *
  */
-class H1_Tetrahedron_Lagrange1_Gauss1 final : public FiniteElementBase
+template< int NUM_Q_POINTS >
```
CusiniM:
```suggestion
template< typename NUM_Q_POINTS >
```
and then `NUM_Q_POINTS::value` to use it and`H1_Tetrahedron_Lagrange1_Gauss< std::integral_constant< int, 1 > >`

src/coreComponents/finiteElement/elementFormulations/H1_Tetrahedron_Lagrange1_Gauss.hpp
```diff
@@ -322,7 +323,199 @@ class H1_Tetrahedron_Lagrange1_Gauss1 final : public FiniteElementBase
...
+                                              0.042546020777081466438,
+                                              0.042546020777081466438,
+                                              0.042546020777081466438,
+                                              0.042546020777081466438 };
+      return w[q];
+    }
+    else
+    {
+      GEOS_UNUSED_VAR( q );
+      GEOS_ERROR( "NUM_Q_POINTS not available for this element type" );
```
CusiniM, line 369: maybe `static_assert(true, "NUM_Q_POINTS not available for this element type");` ?

CusiniM: I don't know if we need this. I guess ideally we would like it not to compile at all.

CusiniM: Maybe we can add at the top of the class: 

````++
static_assert(NUM_Q_POINTS::value == 1 || NUM_Q_POINTS::value == 5 |NUM_Q_POINTS::value == 14,
"NUM_Q_POINTS::value must be 1, 5, or 14!");

File: src/coreComponents/finiteElement/FiniteElementDiscretization.cpp [hunk 13]
```diff
@@ -49,6 +49,11 @@ FiniteElementDiscretization::FiniteElementDiscretization( string const & name, G
     setInputFlag( InputFlags::OPTIONAL ).
     setApplyDefaultValue( 0 ).
     setDescription( "Specifier to indicate whether to force the use of VEM" );
+
+  registerWrapper( viewKeyStruct::useHighOrderQuadratureRuleString(), &m_useHighOrderQuadratureRule ).
+    setInputFlag( InputFlags::OPTIONAL ).
+    setApplyDefaultValue( 0 ).
+    setDescription( "Specifier to indicate whether to use a high order quadrature rule" );
```
CusiniM, line 56:
```suggestion
    setDescription( "Flag to indicate whether to use a high order quadrature rule or not." );
```

CusiniM (approved):

PR #3396, opened 2024-10-12
feat: make possible to use solid mechanics solver to perform poromechanics initialization

### LINKED ISSUE #2698: Stress initialization does not properly work in sequential mode when mechanics-only solver is used to perform initialization
Stress initialization does not properly work in sequential mode when mechanics-only solver is used to perform initialization.
The reason is zero bulk density.
Initialization works ok if coupled (flow+mechanics) is applied to perform initialization.

paveltomin:
and call it with poromechanics solver pointer to perform initialization independently from coupled solver


inputFiles/poromechanics/PoroElastic_staircase_co2_3d_fim_new.xml
```diff
@@ -0,0 +1,152 @@
...
+      solidSolverName="linearElasticity"
+      logLevel="1"/>
+    <CompositionalMultiphaseStatistics
+      name="twoPhaseFlowStatistics"
+      flowSolverName="twoPhaseFlow"
+      logLevel="1"
+      computeCFLNumbers="1"
+      computeRegionStatistics="1"/>
+
+    <MechanicsInitialization
```
paveltomin: this is how initialization can be defined now

src/coreComponents/physicsSolvers/multiphysics/PoromechanicsSolver.hpp
```diff
@@ -118,6 +114,10 @@ class PoromechanicsSolver : public CoupledSolver< FLOW_SOLVER, MECHANICS_SOLVER
                    GEOS_FMT( "{} {}: The attribute `{}` of the flow solver `{}` must be set to 1 since the poromechanics solver is thermal",
                              this->getCatalogName(), this->getName(), FlowSolverBase::viewKeyStruct::isThermalString(), this->flowSolver()->getName() ),
                    InputError );
+
+    DomainPartition & domain = this->template getGroupByPath< DomainPartition >( "/Problem/domain" );
+    flowSolver()->initialize( domain );
```
paveltomin: need to call this first to initialize porosity and densities

src/coreComponents/physicsSolvers/multiphysics/PoromechanicsSolver.hpp
```diff
@@ -219,7 +219,7 @@ class PoromechanicsSolver : public CoupledSolver< FLOW_SOLVER, MECHANICS_SOLVER
           setRestartFlags( dataRepository::RestartFlags::NO_WRITE ).
           setSizedFromParent( 0 );
 
-        if( this->getNonlinearSolverParameters().m_couplingType == NonlinearSolverParameters::CouplingType::Sequential )
+        //if( this->getNonlinearSolverParameters().m_couplingType == NonlinearSolverParameters::CouplingType::Sequential )
```
paveltomin: not sure what to do here, please let me know

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.cpp
```diff
@@ -1055,7 +1056,7 @@ void SolidMechanicsLagrangianFEM::assembleSystem( real64 const GEOS_UNUSED_PARAM
                                                                 MeshLevel & mesh,
                                                                 arrayView1d< string const > const & regionNames )
   {
-    if( m_isFixedStressPoromechanicsUpdate )
+    if( m_isFixedStressPoromechanicsUpdate || m_performStressInitialization )
```
rrsettgast, line 1059: Not that this PR is the introduction of this, but the single physics solver shouldn't know about the coupled physics solver.

rrsettgast:
@paveltomin Why can the mechanics solver not be called directly from the poromechanics initialization? I am missing something.

paveltomin:
> @paveltomin Why can the mechanics solver not be called directly from the poromechanics initialization? I am missing something.

I am not sure I understood you correctly, sorry.
When mechanics solver is called directly it should know that the whole problem is a poromechanics problem, not pure mechanics.

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.cpp
tjb-ltk: In the case of coupled reservoir and well solver, where the event section only activates the reservoir solver, doesn't the well solver need to know that reservoir solver is assuming that well variables are constant.  the well solver knows its coupled to the reservoir via perforations.

src/coreComponents/physicsSolvers/multiphysics/PoromechanicsSolver.hpp
tjb-ltk: I'm surprised that initialization of these quantities can be called from a constructor.... versus in some post process routine...    it's good to have explicit naming for initialization... I'm assuming this is capillary gravity equilibrium for the reservoir solver ?

paveltomin: it's a bunch of things including hydrostatic equilibrium
multiphase equilibration is not yet supported, see https://github.com/GEOS-DEV/GEOS/blob/e601fad71502e0af6b0117af2f8c212ee11ca720/src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp#L1153

inputFiles/poromechanics/PoroElastic_staircase_co2_3d_fim_new.xml
tjb-ltk: I'm a way out of my domain with mechanics and the paradigm for what gets tagged as an event versus a bool field on object is not entirely clear ... but generalizing a bit... would there be a ReservoirInitialization Task.... ??  Whenever/if multiphase initialization is implemented there will be a whole suite of initialization options. What leads the user to think I need a MechanicsInit Task and not a ReservoirInit task. ...

paveltomin:
new baselines are needed but no real results changes

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.cpp
```diff
@@ -1021,25 +1024,6 @@ void SolidMechanicsLagrangianFEM::setupSystem( DomainPartition & domain,
     arrayView1d< globalIndex const > const
     dofNumber = nodeManager.getReference< globalIndex_array >( dofManager.getKey( solidMechanics::totalDisplacement::key() ) );
 
-    if( m_contactRelationName != viewKeyStruct::noContactRelationNameString() )
```
paveltomin, line 1024: @CusiniM this seems not needed and when called - it does not work, no FE discretization is registered for face elements

tjb-ltk (approved):

Guotong-Ren (approved):

jhuang2601 (approved):

PR #3398, opened 2024-10-14
feat: Add attenuation for acoustic wave solvers

sframba:
This PR adds attenuation for the acoustic isotropic and VTI wave solvers, analogously as done for the elastic solvers by #3080 .
The method is similar: memory variables are added, one per standard linear solid (SLS). In the acoustic case, one variable per SLS (plus one working variable) is needed for the isotripic case, and two variables per SLS (plus two working variables) for the anisotropic case. In total $N+1$ fields are added in the isotropic case and $2N+2$ in the anisotropic case if $N$ SLSs are activated (cf. the elastic case, which needs $3N+3$ variables for both the isotrpic and anisotropic cases). 
All the other implementation details are the same.
Attenuation is not implemented if PML boundary conditions are active.


src/coreComponents/physicsSolvers/wavePropagation/sem/acoustic/secondOrderEqn/anisotropic/AcousticVTIWaveEquationSEMKernel.hpp
```diff
@@ -487,6 +489,7 @@ class ExplicitAcousticVTISEM : public finiteElement::KernelBase< SUBREGION_TYPE,
         stack.xLocal[ a ][ i ] = m_nodeCoords[ nodeIndex ][ i ];
       }
     }
+    stack.factor = 1.0;
```
acitrain, line 493: This is not used anymore, it can be removed

acitrain: Ho I saw that it is used later on a derived class maybe a comment will be useful

sframba: Done

acitrain (approved):

MelReyCG (approved):
(approving the unit-tests part)

PR #3364, opened 2024-09-20
feat: Separate mass and volume residuals for output in compositional flow solver

paveltomin:

PR #3399, opened 2024-10-15
fix: Bug fix for residual computation

paveltomin:
introduced in https://github.com/GEOS-DEV/GEOS/pull/3364


paveltomin:
@CusiniM can you approve please?

CusiniM (approved):

PR #3401, opened 2024-10-16
fix: Add HU simplified 2-phase version, refactor IHU implementation and fix bugs where possible, change phase/component flux logic

paveltomin:
1.  Another implementation of for hybrid upwinding, limited to 2 phases, tries to do total mobility treatment closer to published schemes
2. Refactored IHU implementation, added few helper functions to minimize code duplication, fixed/attempted to fix the following bugs:

- `densMean * dGravD_dP` should be outside of `for( localIndex j = 0; j < numFluxSupportPoints; ++j )` subloop (not very important since `dTrans_dPres` is often zero), similar fix in capillary flux
- `totMob += phaseMob[seri[ke]][sesri[ke]][sei[ke]][jp]` and derivatives should not be in the `for( localIndex ke = 0; ke < numFluxSupportPoints; ++ke )` subloop - but i could not change this one without getting bad convergence for barrier test case - results with wrong treatment are definitely off but i am not sure what to do with that, @jafranc any thoughts?
- derivatives for total mobility here: `dFractionalFlow_dP[ke] -= fractionalFlow * dTotMob_dP[k_up_ppu] / LvArray::math::max( totMob, minTotMob )` should be taken from `ke` not `k_up_ppu` (same in other similar places)

3. Change phase/component flux logic - assemble full phase flux first and then do _x_cp_ upwinding once to compute component flux. The danger of old treatment is that `phaseFlux` being used in subsequent computations and it might be tricky to separate parts of phase flux to compute component flux


src/coreComponents/physicsSolvers/fluidFlow/kernels/compositional/IHUPhaseFlux.hpp
```diff
@@ -366,11 +364,11 @@ computeFractionalFlowViscous( localIndex const numPhase,
 
     for( localIndex ke = 0; ke < numFluxSupportPoints; ++ke )
     {
-      dFractionalFlow_dP[ke] -= fractionalFlow * dTotMob_dP[k_up_ppu] / LvArray::math::max( totMob, minTotMob );
+      dFractionalFlow_dP[ke] -= fractionalFlow * dTotMob_dP[ke] / LvArray::math::max( totMob, minTotMob );
```
jafranc: I am not sure it has an impact or not, but ''dTotMob_dP[ke]'' can be null (e.g. cocurrent flow 2 phases) and hence this part of the derivative too, while ''dTotMob_dP[k_up_ppu]'' is not.

src/coreComponents/physicsSolvers/fluidFlow/kernels/compositional/IHUPhaseFlux.hpp
```diff
@@ -289,7 +288,6 @@ computeFractionalFlowViscous( localIndex const numPhase,
                               ElementViewConst< arrayView3d< real64 const, constitutive::cappres::USD_CAPPRES > > const & phaseCapPressure,
                               ElementViewConst< arrayView4d< real64 const, constitutive::cappres::USD_CAPPRES_DS > > const & dPhaseCapPressure_dPhaseVolFrac,
                               integer const capPressureFlag,
-                              localIndex & k_up_main,
                               real64 & fractionalFlow,
```
jafranc, line 371: agreed, this was convenient for debug purpose to out every upwind taken in the process but sure heavier on interfaces

src/coreComponents/physicsSolvers/fluidFlow/kernels/compositional/HU2PhaseFlux.hpp
```diff
@@ -0,0 +1,813 @@
...
+    {
+      dViscousPhaseFlux_dP[ke] += fractionalFlow * dTotFlux_dP[ke];
+      for( localIndex jc = 0; jc < numComp; ++jc )
+      {
+        dViscousPhaseFlux_dC[ke][jc] += fractionalFlow * dTotFlux_dC[ke][jc];
+      }
+    }
+
+    // accumulate in the flux and its derivatives (need to be very careful doing that)
+    phaseFlux += viscousPhaseFlux;
```
jafranc: as it will be always same distribution can be refactored out

src/coreComponents/physicsSolvers/fluidFlow/kernels/compositional/HU2PhaseFlux.hpp
```diff
@@ -0,0 +1,813 @@
...
+        for( localIndex jc = 0; jc < numComp; ++jc )
+        {
+          dGravPhaseFlux_dC[k_up_j][jc] += mob_i * dMob_j_dC[jc] * mobTotInv * potDiff;
+        }
+
+        // mobTot derivatives
+        real64 const mobTotInv2 = mobTotInv * mobTotInv;
+        for( localIndex ke = 0; ke < numFluxSupportPoints; ++ke )
+        {
+          dGravPhaseFlux_dP[ke] -= mob_i * mob_j * dMobTot_dP[ke] * mobTotInv2 * potDiff;
```
jafranc: Isn't it there same as product of the gravitational fractional flows ?

src/coreComponents/physicsSolvers/fluidFlow/kernels/compositional/HU2PhaseFlux.hpp
```diff
@@ -0,0 +1,813 @@
...
+      dPhaseFlux_dP[ke] += dGravPhaseFlux_dP[ke];
+      for( localIndex ic = 0; ic < numComp; ++ic )
+        dPhaseFlux_dC[ke][ic] += dGravPhaseFlux_dC[ke][ic];
+    }
+  }
+
+   template< localIndex numComp, localIndex numFluxSupportPoints >
+   GEOS_HOST_DEVICE
+   static void
+   computeCapillaryFlux( const integer & ip, const integer & numPhase,
```
jafranc: apart from viscous term, as every other contribution is treated the same, templating on ''FORCES'' was the initial idea, first discarded as CUDA<12 limitation, though we might want to bring it back

jafranc (approved):
I think it is a good way of clearing out any doubt on the first implementation and refactoring heavy interfaces. 
I need to think a bit more about the derivatives of total mobility to be sure of what should be there.

src/coreComponents/physicsSolvers/fluidFlow/kernels/compositional/HU2PhaseFlux.hpp
paveltomin: not sure i got it, can you elaborate?

src/coreComponents/physicsSolvers/fluidFlow/kernels/compositional/HU2PhaseFlux.hpp
paveltomin: i am confused, sorry
it is jsut taking derivatives piece by piece, did i miss something?

src/coreComponents/physicsSolvers/fluidFlow/kernels/compositional/HU2PhaseFlux.hpp
paveltomin: yeah, i could see that idea but for this implementation would prefer to keep it simpler

src/coreComponents/physicsSolvers/fluidFlow/kernels/compositional/IHUPhaseFlux.hpp
paveltomin: `dTotMob_dP` can have contributions from both cells, right?
so here we are collecting those contributions into `dFractionalFlow_dP`
using `k_up_ppu` does not make much sense, no?

src/coreComponents/physicsSolvers/fluidFlow/kernels/compositional/IHUPhaseFlux.hpp
```diff
@@ -345,32 +331,24 @@ computeFractionalFlowViscous( localIndex const numPhase,
...
 
     for( localIndex ke = 0; ke < numFluxSupportPoints; ++ke )
     {
-      dFractionalFlow_dP[ke] -= fractionalFlow * dTotMob_dP[k_up_ppu] / LvArray::math::max( totMob, minTotMob );
+      dFractionalFlow_dP[ke] -= fractionalFlow * dTotMob_dP[ke] / LvArray::math::max( totMob, minTotMob );
 
       for( localIndex jc = 0; jc < numComp; ++jc )
       {
-        dFractionalFlow_dC[ke][jc] -= fractionalFlow * dTotMob_dC[k_up_ppu][jc] / LvArray::math::max( totMob, minTotMob );
+        dFractionalFlow_dC[ke][jc] -= fractionalFlow * dTotMob_dC[ke][jc] / LvArray::math::max( totMob, minTotMob );
```
paveltomin: same logic here

src/coreComponents/physicsSolvers/fluidFlow/kernels/compositional/HU2PhaseFlux.hpp
jafranc: I mean for all contribution `viscous`, `capillary` and `gravity`, it will be added to the `phaseFlux` the same way. To avoid copy-paste and possible changes, better refactor everything in a helper, isn't it ?

paveltomin: oh ok, sure, will do

src/coreComponents/physicsSolvers/fluidFlow/kernels/compositional/HU2PhaseFlux.hpp
jafranc: No worries, just noticing that as you are force to go _fractional flow_ formulation, it might be simpler to pre-compute fractional flow derivatives correctly upwinded so ``d(mob_beta * f_alpha) = d(mob_beta) f_alpha + d(f_alpha) mob_beta`` rather than having fractional flow all explicited. You'll have total mobility correctly upwinded hidden there though. (which is great when everything is working, less then there is a bug to track)
Another good point is that it would retain which phase is considered and which is anonymous iterative index.

src/coreComponents/physicsSolvers/fluidFlow/kernels/compositional/IHUPhaseFlux.hpp
jafranc: I think unless you'd been careful when building `dTotMob_dP` both are not quite right isn't it ?

Then as the formula is:

```
d(f)_L = ... - f_HU d(mobT_PPU)d(XL)/mobT_PPU
d(f)_R = ... - f_HU d(mobT_PPU)d(XR)/mobT_PPU
```
so if counter current flow occurs,
```
mobT_PPU = mob_a_L + mob_b_R
```
so you don't want 
```
d(mobT)d(XL) = d(mob_a_L + mob_b_L)d(XL)
```
and rather 
```
d(mobT_PPU) = d(mob_a_L )d(XL)
```
no ?

paveltomin: thanks for the explanation
the logic is like that:
`dTotMob_dP` is an array of 2 elements
when being assembled it will write into `k_up`, can be different for each phase
after that `dTotMob_dP[0]` contains derivatives of totMob wrt pressure in cell 0 and `dTotMob_dP[1]` contains derivatives of totMob wrt pressure in cell 1
we are taking `dFractionalFlow_dP` derivative wrt pressure in cell 0/1 by including contribution from `dTotMob_dP` wrt to pressure in cell 0/1
there will be no `d(mob_b_L)d(XL)` contribution in your example, it should effectively become
```
d(mobT)d(XL) = d(mob_a_L)d(XL)
d(mobT)d(XR) = d(mob_b_R)d(XR)
```
please let me know if something still does not match

src/coreComponents/physicsSolvers/fluidFlow/kernels/compositional/HU2PhaseFlux.hpp
paveltomin: i got it now, yeah, this is some matter of personal preference i guess
not sure i prefer to introduce another frac flow containers

src/coreComponents/physicsSolvers/fluidFlow/kernels/compositional/IHUPhaseFlux.hpp
jafranc: No,I think it matches then. Devils in details with those interpolation scheme :+1:

src/coreComponents/physicsSolvers/fluidFlow/kernels/compositional/IHUPhaseFlux.hpp
```diff
@@ -42,7 +42,132 @@ using Deriv = constitutive::multifluid::DerivativeOffset;
...
+  value = 0;
+  for( localIndex ke = 0; ke < numFluxSupportPoints; ++ke )
+  {
+    assignToZero( deriv_dP[ke], deriv_dC[ke] );
+  }
+}
+
+template< localIndex numComp >
+GEOS_HOST_DEVICE
+static void addDerivativesScaled( real64 ( &deriv_dP ), real64 ( & deriv_dC )[numComp],
```
jafranc: Great helpers :+1:  addition though I would def stick to the convention of passing the non-const ref last

paveltomin: sure, let me flip those

paveltomin:
@CusiniM @castelletto1 this is ready for review
our plan is to give it a try for SPE11 and see if HU can help with convergence

paveltomin:
Waiting on code owner review from @CusiniM, @castelletto1, @rrsettgast, and/or @ryar9534.

paveltomin:
@CusiniM, @castelletto1, @rrsettgast, and/or @ryar9534 please review

paveltomin:
@CusiniM, @castelletto1, @rrsettgast, and/or @ryar9534 need review

tjb-ltk (approved):
Nice refactoring to cleanup the phase and  component flux calcs...  clarity and avoiding redundant code certainly helps the next developer

dkachuma (approved):

paveltomin:
@CusiniM, @castelletto1, @frankfeifan, @jhuang2601, @rrsettgast please review

Guotong-Ren (approved):

castelletto1 (approved):

PR #3418, opened 2024-10-30
fix: Standardisation of incorrect sub-group creation errors

MelReyCG:
This PR aim to standardise the behaviour & log message when user attempt to create in a given node:
1. Non-node content, like raw text,
2. Unexpected XML sub-node (thus, unexpected `Group` type name request in `Group::createChild()`)

------
1. Non-node content

For instance, before this PR, entering raw text in a `<ElementRegions>` resulted in a misleading message:
```
***** Rank 0: KeyName () not found in ObjectManager::Catalog
```

We would now have:
```
***** Rank 0: Error in node named "ElementRegions" (staircase_co2_wells_3d.xml, l.124): GEOS XML nodes cannot contain text data nor anything but XML nodes.
Erroneous content: "lorem ipsum"
```
If a `Group` type would need this kind of sub-content later, we would have to manage this data, and silence this error for the case it would be allowed.

------
2. Unexpected sub-node

For instance, before this PR, adding a wrongly spelled `<Perforation>` , or any unexpected name, in a `<InternalWell>` a resulted in:
```
***** LOCATION: /data/pau901/SIM_CS/users/MelvinRey/WorkEnv/GEOSX_repos/GEOS/src/coreComponents/mesh/generators/WellGeneratorBase.cpp:90
***** Rank 0: Unrecognized node: Prerforatio
```

We would now have:
```
***** ERROR
***** LOCATION: /data/pau901/SIM_CS/users/MelvinRey/WorkEnv/GEOSX_repos/GEOS/src/coreComponents/mesh/generators/WellGeneratorBase.cpp:81
***** Controlling expression (should be false): childKey != viewKeyStruct::perforationString()
***** Rank 0: The tag "Prerforatio" is invalid within well_producer1 (staircase_co2_wells_3d.xml, l.74). Please verify the keywords spelling and that input file parameters have not changed.
All available tags are: Perforation
```
Previously, error of this kind were not always managed properly. Also, various `Group` types messages were not standardized nor indicating available sub-type names.

------

This PR also improve how wrong character are underlined:
```
***** Rank 0: Input string validation failed at:
  "well  Region1"
       ^^
  Expected format: Input value must be a string that cannot be empty and contains only upper/lower letters, digits, and the characters  . - _
```
It was broken for some cases by the PR #3403, so I fixed that and also improved it by underlining the (first) whole sequence of wrong characters.


rrsettgast (approved):

wrtobin:
looks like there are some changes from `geos` internal `string` to specifically `std::string` on some interfaces, is there a reason this is needed?

MelReyCG:
@wrtobin half mistake, half in place pratices in `ObjectCatalog.hpp` (this last one has a lot of old `std::string`).
We have a lot of `std::string` in `coreComponents`, but that is a subject for another PR.

wrtobin (approved):

paveltomin (approved):

PR #3426, opened 2024-11-05
fix: reset accumulation in fracture when time step cut occurs in hydrofrac solver

Guotong-Ren:
When time step cut occurs, primary variables are reset to the beginning of newton, while the accumulation in fracture is not updated accordingly to pair with the initial guess.


CusiniM (approved):

paveltomin (approved):

src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp
```diff
@@ -209,10 +209,10 @@ real64 HydrofractureSolver< POROMECHANICS_SOLVER >::fullyCoupledSolverStep( real
     }
 
     // currently the only method is implicit time integration
-    dtReturn = nonlinearImplicitStep( time_n, dt, cycleNumber, domain );
+    dtReturn = nonlinearImplicitStep( time_n, dtReturn, cycleNumber, domain );
```
binw1: Simply change dt to dtReturn here may still have issues, as dt used in different fracture propagation iterations will still be different if timestep cut happens in some of the fracture propagation iteration.

src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp
```diff
@@ -209,10 +209,10 @@ real64 HydrofractureSolver< POROMECHANICS_SOLVER >::fullyCoupledSolverStep( real
...
 
 
-    if( !this->m_performStressInitialization && m_surfaceGenerator->solverStep( time_n, dt, cycleNumber, domain ) > 0 )
+    if( !this->m_performStressInitialization && m_surfaceGenerator->solverStep( time_n, dtReturn, cycleNumber, domain ) > 0 )
```
binw1: Simply change dt to dtReturn here may still have issues, as dt used in different fracture propagation iterations will still be different if timestep cut happens in some of the fracture propagation iteration.

binw1 (approved):

src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp
Guotong-Ren: yeah, for the time being, i reverted it back to original formulation. Thanks.

binw1: Yeah, I then saw the revert in the second commit

Guotong-Ren:
@jhuang2601  can you please review this PR? Thanks!

jhuang2601 (approved):

jhuang2601:
@Guotong-Ren You need to run the command in the the build folder: `make uncrustify_style` to fix the reported issue of CI failure on Check code style

PR #3427, opened 2024-11-06
feat: well time step selector based on rates/bhp tables and clarify well rates logic

paveltomin:
Two things in this PR:

1. Implement `setNextDt` in well solver which looks at all rates/bhp/status tables and makes sure time step is selected so that we honor table intervals. Can be enabled by `timeStepFromTables` flag in well solver.
2. Change logic for well rates - currently inputted and outputted from `t_n+dt` (end of time step) - switch to `t_n` (beginning of time step) - this should be more consistent with `lower` interpolation method that is used and prevent grabbing "future" rate from tables and avoid weird behavior when time step cut can lead to shutting of wells.


src/coreComponents/physicsSolvers/fluidFlow/wells/WellSolverBase.cpp
```diff
@@ -377,4 +377,73 @@ WellControls const & WellSolverBase::getWellControls( WellElementSubRegion const
   return this->getGroup< WellControls >( subRegion.getWellControlsName());
 }
 
+real64 WellSolverBase::setNextDt( real64 const & currentTime, const real64 & lastDt, geos::DomainPartition & domain )
```
paveltomin: main thing is in this function

paveltomin:
@tjb-ltk the second part of condition
```
      if( time_n <= 0.0 ||
          ( !wellControls.isWellOpen( time_n ) && wellControls.isWellOpen( time_n + dt ) )  )
```
in `CompositionalMultiphaseWell::initializeWells` is broken now, not exactly sure what to do with it, any hint?

src/coreComponents/functions/TableFunction.cpp
```diff
@@ -196,6 +196,11 @@ real64 TableFunction::evaluate( real64 const * const input ) const
   return m_kernelWrapper.compute( input );
 }
 
+std::vector< real64 > TableFunction::getCoord( real64 const * const input, InterpolationType interpolationMethod ) const
```
paveltomin: @CusiniM i know std should not be used here, but not sure what should be the right container?

dkachuma: Isn't it simply `array1d`?

src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
```diff
@@ -363,50 +363,51 @@ void PhysicsSolverBase::logEndOfCycleInformation( integer const cycleNumber,
...
+  if( m_nonlinearSolverParameters.getLogLevel() > 0 && nextDtNewton < LvArray::NumericLimits< real64 >::max )
+    GEOS_LOG_RANK_0( GEOS_FMT( "{}: next time step based on number of iterations = {}", getName(), nextDtNewton ));
+  real64 const nextDtStateChange = setNextDtBasedOnStateChange( lastDt, domain );
+  if( m_nonlinearSolverParameters.getLogLevel() > 0 && nextDtStateChange < LvArray::NumericLimits< real64 >::max )
     GEOS_LOG_RANK_0( GEOS_FMT( "{}: next time step based on state change = {}", getName(), nextDtStateChange ));
 
   if( ( m_numTimestepsSinceLastDtCut >= 0 ) && ( m_numTimestepsSinceLastDtCut < minTimeStepIncreaseInterval ) )
   {
-    GEOS_LOG_LEVEL_INFO_RANK_0( logInfo::TimeStep, GEOS_FMT( "{}: time-step size will be kept the same since it's been {} cycles since last cut.",
+    GEOS_LOG_LEVEL_INFO_RANK_0( logInfo::TimeStep, GEOS_FMT( "{}: time-step size should be kept the same since it's been {} cycles since last cut",
```
dkachuma: I think "will" is still more clear here.

src/coreComponents/physicsSolvers/contact/SolidMechanicsLagrangeContact.cpp
```diff
@@ -2356,13 +2356,6 @@ bool SolidMechanicsLagrangeContact::isFractureAllInStickCondition( DomainPartiti
   return ( ( numNewSlip + numSlip + numOpen ) == 0 );
 }
 
-real64 SolidMechanicsLagrangeContact::setNextDt( real64 const & currentDt,
```
dkachuma: Is this intentional? Do we really want this solver to inherit the behaviour of the base class?

src/coreComponents/physicsSolvers/fluidFlow/wells/WellSolverBase.cpp
```diff
@@ -315,4 +315,44 @@ WellControls const & WellSolverBase::getWellControls( WellElementSubRegion const
   return this->getGroup< WellControls >( subRegion.getWellControlsName());
 }
 
+real64 WellSolverBase::setNextDt( real64 const & time, const real64 & lastDt, geos::DomainPartition & domain )
```
dkachuma: I think we should include an option to disable this. I might very finely discretise my rate function without wanting geos to take such small time steps.

src/coreComponents/functions/TableFunction.hpp
```diff
@@ -240,6 +250,14 @@ class TableFunction : public FunctionBase
    */
   virtual real64 evaluate( real64 const * const input ) const override final;
 
+  /**
+   * @brief Method to get coordinates
+   * @param input a scalar input
+   * @param interpolationMethod the interpolation method
+   * @return the coordinates
+   */
+  std::vector< real64 > getCoord( real64 const * const input, InterpolationType interpolationMethod ) const;
```
dkachuma:
```suggestion
  real64 getCoord( real64 const * const input, InterpolationType interpolationMethod, integer const dim = 0 ) const;
```
I would add the dimension to the input and simply return 1 value. It looks like that's what the application needs anyway.

src/coreComponents/physicsSolvers/fluidFlow/wells/WellSolverBase.cpp
```diff
@@ -315,4 +315,44 @@ WellControls const & WellSolverBase::getWellControls( WellElementSubRegion const
...
+  real64 nextDt = PhysicsSolverBase::setNextDt( time, lastDt, domain );
+
+  forDiscretizationOnMeshTargets( domain.getMeshBodies(), [&]( string const &,
+                                                               MeshLevel & mesh,
+                                                               arrayView1d< string const > const & regionNames )
+  {
+    mesh.getElemManager().forElementSubRegions< WellElementSubRegion >( regionNames, [&]( localIndex const,
+                                                                                          WellElementSubRegion & subRegion )
+    {
+      WellControls & wellControls = getWellControls( subRegion );
```
dkachuma: Instead of having `WellControls` expose all its tables, can we simply do `wellControls.setNextDtFromTable( time, nextDt );`

dkachuma (approved):
It's quite a change (switching the well from t+dt to t). But I don't see anything wrong - in any case it seems more correct. I don't think the cases we have with wells would be so greatly affected.

src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
paveltomin: my confusion is: it says it "will be" but then many other things can affect the time step choice so nothing is guaranteed :)

src/coreComponents/physicsSolvers/contact/SolidMechanicsLagrangeContact.cpp
paveltomin: intentional - yes but I forgot about that change and didn't mention it in the description, sorry
it is very confusing what `SolidMechanicsLagrangeContact::setNextDt` does right now by returning `currentDt` and next time step, this basically blocks time step increase? (CC @CusiniM @matteofrigo5 @castelletto1)
could go in a small simple PR with just that fix

src/coreComponents/physicsSolvers/fluidFlow/wells/WellSolverBase.cpp
paveltomin: ok sounds good

src/coreComponents/functions/TableFunction.hpp
paveltomin: will check that

src/coreComponents/physicsSolvers/fluidFlow/wells/WellSolverBase.cpp
tjb-ltk: I noticed in test case that the writeCSV report is a little different ... Before ,first report was end of first DT, so no rates report at t=0, but there is report at last time,  Now report starts at T=0, which is good, but no report at last timestep.  So one needs to add dt to get the full picture, along with manually adding the line if in excel.  Maybe on destruction or if component got end of sim call, the file record could be written..

tjb-ltk: Also how does this affect other output reporting...

src/coreComponents/functions/TableFunction.cpp
paveltomin: followed Dick's suggestion and added `dim` as argument, so return is simply `real64`

src/coreComponents/functions/TableFunction.hpp
paveltomin: done

src/coreComponents/physicsSolvers/fluidFlow/wells/WellSolverBase.cpp
paveltomin: Added `timeStepFromTables` flag to enable new behavior, disabled by default

src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
paveltomin: reverted

src/coreComponents/physicsSolvers/fluidFlow/wells/WellSolverBase.cpp
paveltomin: done

paveltomin:
Waiting on code owner review from @CusiniM, @Guotong-Ren, @VidarStiernstrom, @castelletto1, @cssherman, @frankfeifan, @jhuang2601, @matteofrigo5, @rrsettgast, @ryar9534, and/or 1 others.

Guotong-Ren (approved):

paveltomin:
> @tjb-ltk the second part of condition
> 
> ```
>       if( time_n <= 0.0 ||
>           ( !wellControls.isWellOpen( time_n ) && wellControls.isWellOpen( time_n + dt ) )  )
> ```
> 
> in `CompositionalMultiphaseWell::initializeWells` is broken now, not exactly sure what to do with it, any hint?

@tjb-ltk any thought on this?

paveltomin:
@CusiniM, @VidarStiernstrom, @castelletto1, @cssherman, @frankfeifan, @jhuang2601, @matteofrigo5, @rrsettgast, @ryar9534, and/or @wrtobin please review

paveltomin:
@CusiniM, @VidarStiernstrom, @castelletto1, @cssherman, @frankfeifan, @jhuang2601, @matteofrigo5, @rrsettgast, @ryar9534, @wrtobin please review

src/coreComponents/physicsSolvers/contact/SolidMechanicsLagrangeContact.cpp
rrsettgast: @paveltomin Put this in its own PR.

paveltomin: done https://github.com/GEOS-DEV/GEOS/pull/3490

paveltomin:
@CusiniM, @VidarStiernstrom, @castelletto1, @cssherman, @frankfeifan, @jhuang2601, @matteofrigo5, @rrsettgast, @ryar9534, @wrtobin need review

src/coreComponents/functions/TableFunction.hpp
```diff
@@ -554,6 +575,57 @@ TableFunction::KernelWrapper::interpolateRound( IN_ARRAY const & input ) const
...
+    // Coordinate is within the table axis
+    // Note: find() will return the index of the upper table vertex
+    auto const lower = LvArray::sortedArrayManipulation::find( coords.begin(), coords.size(), input[dim] );
+    subIndex = LvArray::integerConversion< localIndex >( lower );
+
+    // Interpolation types:
+    //   - Nearest returns the value of the closest table vertex
+    //   - Upper returns the value of the next table vertex
+    //   - Lower returns the value of the previous table vertex
+    if( interpolationMethod == TableFunction::InterpolationType::Nearest )
```
tjb-ltk, line 609: Looks like there are 4 interpolation types...  Linear, Nearest, Upper, Lower...   Using the find method defaults to Upper ?    then overridden if method is nearest or lower... Is there any chance of method being called with Linear

src/coreComponents/physicsSolvers/fluidFlow/wells/WellControls.cpp
```diff
@@ -483,4 +483,27 @@ bool WellControls::isWellOpen( real64 const & currentTime ) const
...
+    // small epsilon to make sure we land on the other side of table interval and pick up the right rate
+    real64 const eps = 1e-6;
+    real64 const dtLimit = (table->getCoord( &currentTime, 0, TableFunction::InterpolationType::Upper ) - currentTime) * ( 1.0 + eps );
+    if( dtLimit > eps && dtLimit < nextDt )
+    {
+      nextDt = dtLimit;
+    }
+  }
+}
+
```
tjb-ltk, line 508: Nice addition....   this paves the way for including table of time vs dt to allow users to change max dt based on precieved need...

src/coreComponents/physicsSolvers/PhysicsSolverBase.hpp
tjb-ltk, line 1: Changing BasedOnNewton tier changed to just  IterNumber looks to support workflows beyond newton solvers?  for better understanding. what is an example

tjb-ltk (approved):

src/coreComponents/functions/TableFunction.hpp
paveltomin: calling for `Linear` would not make much sense i think, should i add a check and throw an error for `Linear`?

paveltomin:
@CusiniM, @VidarStiernstrom, @castelletto1, @cssherman, @frankfeifan, @jhuang2601, @matteofrigo5, @rrsettgast, @wrtobin please review

src/coreComponents/physicsSolvers/fluidFlow/wells/WellSolverBase.cpp
```diff
@@ -61,6 +61,11 @@ WellSolverBase::WellSolverBase( string const & name,
     setInputFlag( dataRepository::InputFlags::OPTIONAL ).
     setDescription( "Write rates into a CSV file" );
 
+  this->registerWrapper( viewKeyStruct::timeStepFromTablesFlagString(), &m_timeStepFromTables ).
+    setApplyDefaultValue( 0 ).
+    setInputFlag( dataRepository::InputFlags::OPTIONAL ).
+    setDescription( "Choose time step to honor rates/bhp tables time intervals" );
```
CusiniM, line 67: shouldn't this always be true? I am guessing we always want the timestep to respect the intervals, don't we?

CusiniM (approved):

src/coreComponents/physicsSolvers/fluidFlow/wells/WellSolverBase.cpp
paveltomin: If I remember correctly, @dkachuma requested this behavior to be an option and I was hesitant to make it default

src/coreComponents/functions/TableFunction.hpp
```diff
@@ -173,6 +173,18 @@ class TableFunction : public FunctionBase
     real64
     interpolateRound( IN_ARRAY const & input ) const;
 
+    /**
+     * @brief ...
+     * @param[in] input vector of input value
+     * @param[in] dim table dimension
+     * @param[in] interpolationMethod interpolation method
+     * @return coordinate value
```
MelReyCG: Maybe unify the documentation with the other `getCoord` signature?

src/coreComponents/functions/TableFunction.hpp
```diff
@@ -240,6 +252,15 @@ class TableFunction : public FunctionBase
...
   virtual real64 evaluate( real64 const * const input ) const override final;
 
+  /**
+   * @brief Method to get coordinates
+   * @param input a scalar input
+   * @param dim the table dimension
+   * @param interpolationMethod the interpolation method
+   * @return the coordinate
+   */
+  real64 getCoord( real64 const * const input, localIndex dim, InterpolationType interpolationMethod ) const;
```
MelReyCG, line 262: I think you should annotate the docs that this method is assuming `createKernelWrapper()` has been called (because it is based on the `m_kernelWrapper` data).
Maybe add an error in it?

MelReyCG (approved):
(Minor remarks)
Nice things in this PR!

src/coreComponents/physicsSolvers/fluidFlow/wells/WellSolverBase.cpp
paveltomin: i added time step size in the output, i think that should be enough to represent full picture
no effect on other output reporting is expected

src/coreComponents/physicsSolvers/PhysicsSolverBase.hpp
paveltomin: simply the coupled sequential solver, it's not Newton iterations but the solver uses same logic for time step selection

src/coreComponents/functions/TableFunction.hpp
paveltomin: added
```
  GEOS_ASSERT_MSG( interpolationMethod != InterpolationType::Linear,
                   GEOS_FMT( "{}: TableFunction::getCoord should not be called with {} interpolation method",
                             getDataContext(), EnumStrings< InterpolationType >::toString( interpolationMethod )));
```

src/coreComponents/functions/TableFunction.hpp
paveltomin: added
`GEOS_ASSERT( dim >= 0 && dim < m_coordinates.size() );`

src/coreComponents/functions/TableFunction.hpp
paveltomin: done

Issue #3430, opened 2024-11-08
Obsolete darwin-clang.cmake host configuration file


OmarDuran:
**Describe the issue**
The file the darwin-clang.cmake is obsolete.

**Proposed cleanup**
Remove obsolete file.

rrsettgast:
fixed in https://github.com/GEOS-DEV/GEOS/pull/3451

PR #3431, opened 2024-11-08
fix: Removing obsolete darwin-clang.cmake host configuration file

OmarDuran:
Fixes issue #3430.


rrsettgast:
@OmarDuran 
This file should be deleted. It was replaced by the `macOS_base/arm/x86.cmake` files, but I forgot to remove this file after we were confident in the replacement. Can you use this pr to delete this file?

PR #3374, opened 2024-09-29
fix: Compositional flow kernels - split into separate files, bugfix for gravity treatment in flux for thermal

paveltomin:
depends on https://github.com/GEOS-DEV/GEOS/pull/3372

PR #3433, opened 2024-11-08
fix: small bug fix

paveltomin:
happened here https://github.com/GEOS-DEV/GEOS/pull/3337 (sorry)


paveltomin:
already included in https://github.com/GEOS-DEV/GEOS/pull/3374

PR #3434, opened 2024-11-09
fix: Restart files contain incorrect types for ArrayOfArrays

rrsettgast:
This PR adds proper conduit IO for ArrayOfArrays. Previously the IO routines output ArrayOfArrays as 8-bit integers/char. So diffs were incomparable....example:

```
Error: /Problem/domain/MeshBodies/mesh1/meshLevels/Level0/ElementRegions/elementRegionsGroup/Fracture/elementSubRegions/FractureSubRegion/nodeList/__values__
	Arrays of types int8 and int8 have 244 values of which 90 have differing values.
	Statistics of the differences greater than 0:
		max_index = (np.int64(216),), max = 111, mean = 28.666666666666668, std = 36.53856410242015
	Differing valid characters (substrings reordering detected):
	          !  !  ,  ,  7  7  B  B  
	          !  !  ,  ,  7  B  7  B  
	                         ^  ^     
```


paveltomin (approved):

wrtobin (approved):

src/coreComponents/dataRepository/wrapperHelpers.hpp
```diff
@@ -490,6 +490,100 @@ pullDataFromConduitNode( Array< T, NDIM, PERMUTATION > & var,
...
+
+template< typename T, typename INDEX_TYPE >
+std::enable_if_t< bufferOps::can_memcpy< T > >
+pushDataToConduitNode( ArrayOfArrays< T, INDEX_TYPE > const & var2,
+                       conduit::Node & node )
+{
+  ArrayOfArraysView< T const, INDEX_TYPE > const & var = var2.toViewConst();
+  internal::logOutputType( LvArray::system::demangleType( var ), "Output array via external pointer: " );
+
+  // ArrayOfArray::m_numArrays
```
CusiniM:
```suggestion
  // ArrayOfArray::m_numArrays
```

CusiniM: or maybe you want it as a comment?

src/coreComponents/dataRepository/wrapperHelpers.hpp
```diff
@@ -490,6 +490,100 @@ pullDataFromConduitNode( Array< T, NDIM, PERMUTATION > & var,
...
+
+  // ArrayOfArray::m_sizes
+  INDEX_TYPE const * const sizes = var.getSizes();
+  conduit::DataType const sizesType( conduitTypeInfo< INDEX_TYPE >::id, numArrays );
+  node[ "__sizes__" ].set_external( sizesType, const_cast< void * >( static_cast< void const * >( sizes ) ) );
+
+  // Push the data into conduit
+  constexpr int conduitTypeID = conduitTypeInfo< T >::id;
+  constexpr int sizeofConduitType = conduitTypeInfo< T >::sizeOfConduitType;
+  conduit::DataType const dtype( conduitTypeID, offsets[numArrays] * sizeof( T ) / sizeofConduitType );
```
CusiniM, line 539: I don't fully understand the size here. What' s in `offsets[numArrays]`? 

Shouldn't this be the size of each array * `numOfArrays` * `sizeof(T)` ?

src/coreComponents/dataRepository/wrapperHelpers.hpp
```diff
@@ -490,6 +490,100 @@ pullDataFromConduitNode( Array< T, NDIM, PERMUTATION > & var,
...
+  GEOS_ERROR_IF_NE( varOffsets[0], offsets[0] );
+  for( INDEX_TYPE i = 0; i<*numArrays; ++i )
+  {
+    GEOS_ERROR_IF_NE( varOffsets[i+1], offsets[i+1] );
+    GEOS_ERROR_IF_NE( varSizes[i], sizes[i] );
+  }
+
+
+  localIndex numBytesFromArray =  allocatedSize * sizeof( T );
+  // GEOS_ERROR_IF_NE( numBytesFromArray, valuesNode.dtype().strided_bytes() );
```
CusiniM:
```suggestion
```

src/coreComponents/dataRepository/wrapperHelpers.hpp
rrsettgast: yes

src/coreComponents/dataRepository/wrapperHelpers.hpp
rrsettgast: `offsets` is the location in the buffer that a subarray starts. The next `offset` is where the next one starts. So the last value `offsets[numArrays]` is the end of the buffers as there are `numArrays+1` values in `offsets`. So `offsets[numArrays] is the length of the buffer if I am not mistaken. @corbett5 @wrtobin ??

CusiniM: oh I see. I would write this exact comment in the code.

rrsettgast: From `ArrayOfArraysView.hpp`:

```c++
  /// The number of arrays contained.
  INDEX_TYPE_NC m_numArrays = 0;

  /// Holds the offset of each array, of length m_numArrays + 1. Array i begins at
  /// m_offsets[ i ] and has capacity m_offsets[i+1] - m_offsets[ i ].
  BUFFER_TYPE< INDEX_TYPE > m_offsets;

  /// Holds the size of each array.
  BUFFER_TYPE< SIZE_TYPE > m_sizes;

  /// Holds the values of each array. Values in the range [m_offsets[ i ], m_offsets[ i ] + m_sizes[ i ])
  /// are valid. All other entries contain uninitialized values.
  BUFFER_TYPE< T > m_values;
```

src/coreComponents/dataRepository/wrapperHelpers.hpp
rrsettgast: I uncommented this check

src/coreComponents/dataRepository/wrapperHelpers.hpp
corbett5: I agree with Randy. The last offset is one measure of the total length of the underlying memory. Other measures are the capacity from the underlying buffer, or offsets(numArrays - 1) + sizeOfArray(numArrays - 1) which would be the end of the valid values. For io you should only need to write up to this last value.

corbett5 (approved):
I'll give it a look over this evening if it hasn't been merged by then

src/coreComponents/dataRepository/wrapperHelpers.hpp
rrsettgast: @corbett5 I did notice it would be nice for a bit different nomenclature for this object....for example:

-  there is no accessor function that returns the actual size of the data buffer (offset[m_numArrays])
- Perhaps language of calling the "arrays" is confusing. I think "subArrays" would be better...however if we want to do an generic `NestedArrays` arrays object for an arbitrary nesting, it would be nice to come up with something extendible.
- The accessors for `m_offset` and `m_sizes` protected in `ArrayOfArraysView`. So that means that `ArrayOfArrays` can't provide public access. I am sure this is by design, but there needs to be more data helper functions to avoid needing direct access to these internal data arrays IMO.

corbett5: ### 1
If you're accessing the underlying data (ie through `m_offsets` and `m_sizes`) you could be accessing uninitialized data. Which isn't a problem if you have an `ArrayOfArrays< double >`, but is very much a problem (or at least you need to know what you're doing) if you have an `ArrayOfArrays< string >`. So yeah I didn't make this easy on purpose.

There is a way to get the total underlying buffer capacity
```c++
  /**
   * @return Return the total number values that can be stored before reallocation.
   */
  LVARRAY_HOST_DEVICE constexpr inline
  INDEX_TYPE_NC valueCapacity() const
  { return m_values.capacity(); }
```

One way to do IO for the whole buffer would be

```c++
std::copy(arrayOfArryays[1].begin(),  arrayOfArrays[arrayOfArrays.size() - 1].end(), dst_iter);
```

### 2
Certainly open for a name change. I thought I tried to call them inner arrays in the docs, but it's certainly confusing.

### 3
I agree it's strange that you can `getValues()` becomes accessible once you convert an `ArrayOfArrays` to an `ArrayOfArraysView`. I don't think there would be any harm in making these methods accessible directly from `ArrayOfArrays`.

corbett5: Certainly open for a name change. I thought I tried to call them inner arrays in the docs, but it's certainly confusing.

src/coreComponents/dataRepository/wrapperHelpers.hpp
```diff
@@ -490,6 +490,114 @@ pullDataFromConduitNode( Array< T, NDIM, PERMUTATION > & var,
...
+
+  // correctly set the sizes and capacities of each sub-array
+  localIndex allocatedSize = 0;
+  for( INDEX_TYPE i = 0; i < *numArrays; ++i )
+  {
+    INDEX_TYPE const arrayAllocation = offsets[i+1] - offsets[i];
+    var.setCapacityOfArray( i, arrayAllocation );
+    var.resizeArray( i, sizes[ i ] );
+    allocatedSize += arrayAllocation;
+  }
```
rrsettgast, line 589: @corbett5
I have been trying a few iterations with this section of code. I can't seem to get it right. Can you have a close look at the attempt to pre-allocate values and the subsequent capacity/resize of each array to make sure it makes sense?

rrsettgast:
@corbett5 I can't seem to pass integrated tests after baselines are set. There are diffs in the `ArrayOfArraysView::m_values` in the excess capacity of each array. The values appear to be uninitialized. Here an example:

```
dataset: </Problem/domain/MeshBodies/mesh1/meshLevels/Level0/ElementRegions/elementRegionsGroup/Domain/elementSubRegions/cb1/ToEmbeddedSurfaces/__values__> and </Problem/domain/MeshBodies/mesh1/meshLevels/Level0/ElementRegions/elementRegionsGroup/Domain/elementSubRegions/cb1/ToEmbeddedSurfaces/__values__>
size:           [10]           [10]
position        __values__      __values__      difference          
------------------------------------------------------------
[ 1 ]          22053           22034           19             
[ 5 ]          22053           22034           19             
[ 7 ]          22048           22039           9              
[ 9 ]          22048           22039           9              
4 differences found
```

I would think that we don't care about initialized values...but then we can't just dump the values out and read them back in. We either have to avoid writing them out, or set them to some value. 

Is there a way to set the initial values of uninitialized array entries?

src/coreComponents/dataRepository/wrapperHelpers.hpp
corbett5: For this whole function I think I would do this

```c++
var.resizeFromOffsets( *numArrays, offsets );

for (INDEX_TYPE i = 0; i < numArrays; ++i )
{
  T const * begin = valuesNode.data_ptr() + offsets[i];
  T const * end = begin + sizes[ i ];
  var.appendToArray( i, begin, end );
}
```

corbett5:
> @corbett5 I can't seem to pass integrated tests after baselines are set. There are diffs in the `ArrayOfArraysView::m_values` in the excess capacity of each array. The values appear to be uninitialized. Here an example:
> 
> ```
> dataset: </Problem/domain/MeshBodies/mesh1/meshLevels/Level0/ElementRegions/elementRegionsGroup/Domain/elementSubRegions/cb1/ToEmbeddedSurfaces/__values__> and </Problem/domain/MeshBodies/mesh1/meshLevels/Level0/ElementRegions/elementRegionsGroup/Domain/elementSubRegions/cb1/ToEmbeddedSurfaces/__values__>
> size:           [10]           [10]
> position        __values__      __values__      difference          
> ------------------------------------------------------------
> [ 1 ]          22053           22034           19             
> [ 5 ]          22053           22034           19             
> [ 7 ]          22048           22039           9              
> [ 9 ]          22048           22039           9              
> 4 differences found
> ```
> 
> I would think that we don't care about initialized values...but then we can't just dump the values out and read them back in. We either have to avoid writing them out, or set them to some value.
> 
> Is there a way to set the initial values of uninitialized array entries?

I think the best solution would be to modify the python diff code to realize when it's comparing an ArrayOfArrays. The easiest thing would be to set the uninitialized values in the ArrayOfArrays before writing out.


```C++
T * const values = const_cast< T * >( var.getValues() );
INDEX_TYPE const * const offsets = var.getOffsets();
for( INDEX_TYPE i = 0; i < numArrays; ++i)
{
  INDEX_TYPE const curOffset = offsets[ i ];
  INDEX_TYPE const nextOffset = offsets[ i + 1 ];
  
   for ( j = curOffset + var.sizeOfArray( i ); j < nextOffset; ++j )
   {
      values[ j ] = T() // Probably should be a NaN or something.
   }
}
```

src/coreComponents/dataRepository/wrapperHelpers.hpp
```diff
@@ -515,11 +515,32 @@ pushDataToConduitNode( ArrayOfArrays< T, INDEX_TYPE > const & var2,
...
+  // **** WARNING: alters the uninitialized values in the ArrayOfArrays ****
+  T const * const values = var.getValues();
+  std::vector< T > valuesCopy( values, values + offsets[numArrays] );
+  for( INDEX_TYPE i = 0; i < numArrays; ++i )
+  {
+    INDEX_TYPE const curOffset = offsets[ i ];
+    INDEX_TYPE const nextOffset = offsets[ i + 1 ];
+    for( INDEX_TYPE j = curOffset + var.sizeOfArray( i ); j < nextOffset; ++j )
+    {
+      if constexpr ( std::is_arithmetic< T >::value )
```
corbett5, line 526: Isn't this redundant since `T() = 0` if `std::is_arithmetic< T >::value == true`? Also, why do you choose to make a copy instead of changing the values in the array? If we're going to make a copy anyways I'd suggest compressing out the uninitialized values. The only "problem" with this is that when we read the array back in it is compressed as well.

rrsettgast: > Isn't this redundant since `T() = 0` if `std::is_arithmetic< T >::value == true`? Also, why do you choose to make a copy instead of changing the values in the array? If we're going to make a copy anyways I'd suggest compressing out the uninitialized values. The only "problem" with this is that when we read the array back in it is compressed as well.

- I wasn't sure that `T() = 0` was actually true. If we have a variable declared as `int a;`, is `a` guaranteed to be equal to zero?
- I choose to make a copy because I didn't want to break the `const` promise on the write function. It is of course a double edge sword, because you are not reading in what was passed to the write function...but then changing the `const` object that was passed to the write function seems somewhat illegal?...nevermind...can't make a local copy. Have to break constness.
- I did not want to compress it because I wanted to make sure that the behavior of a restart run is identical to the original run. Compressing the `ArrayOfArrays` would result in behavioral changes as adding new values would require different operations after restart....which of course should be OK...unless there is some sort of allocation error that occurs when you try to insert new values in the `ArrayOfArrays`...etc. This is of course ignoring the fact that the array you read back in will not be exactly identical to the one you wrote out in the "unallocated" values....but those shouldn't matter ever or we have bigger problems.

src/coreComponents/dataRepository/wrapperHelpers.hpp
rrsettgast: @corbett5 Is the name `resizeFromOffsets` accurate? Looking that the function should it be `reserveFromOffsets`?

src/coreComponents/dataRepository/wrapperHelpers.hpp
corbett5: - `int a;` is different from `int a = int()`. The first is uninitialized and can be anything, the second calls the default constructor which for numeric types always returns zero.
- Yeah I think breaking correctness is fine here.
- I get that, if you want as similar behavior as possible then setting the uninitialized values is definitely the way to go.

rrsettgast: @corbett5  
I should know that. thanks!

PR #3348, opened 2024-09-12
fix: plastic/elastic slip fix for EFEM

sohailwaziri:
update/store elastic slip only once at the end of the time step

PR #3439, opened 2024-11-11
fix: EDFM bugfixes, part of #3348

paveltomin:
part of https://github.com/GEOS-DEV/GEOS/pull/3348/
1. derivatives sign fix
2. frac/cell element volume
3. replace FaceElementSubRegion by SurfaceElementSubRegion in SinglePhaseBase
4. fix inconsistency between default aperture and hydraulic aperture table in EDFM test cases


Guotong-Ren (approved):

src/coreComponents/constitutive/contact/CoulombFriction.hpp
```diff
@@ -265,6 +265,8 @@ inline void CoulombFrictionUpdates::computeShearTraction( localIndex const k,
       real64 dLimitTau_dNormalTraction;
       real64 const limitTau = computeLimitTangentialTractionNorm( tractionVector[0],
                                                                   dLimitTau_dNormalTraction );
+      // dLimitTau_dNormalTraction from the function above has a wrong sign, flip it
+      dLimitTau_dNormalTraction = -dLimitTau_dNormalTraction;
```
CusiniM: why not fixing it inside the function?

CusiniM (approved):

src/coreComponents/constitutive/contact/CoulombFriction.hpp
paveltomin: the functions is being used also by `SolidMechanicsLagrangeContact`, just limiting changes to EDFM here

CusiniM: but the sign should not change. I mean I would expect that derivative to be the same

CusiniM (approved):

paveltomin:
checked the tests diffs and everything looks good - only edfm cases are different as expected, improvements in nonlinear convergence as expected

PR #3416, opened 2024-10-28
feat: QD rate and state friction coupled with Mechanics and Poromechanics solvers.

CusiniM:
In this PR:

- [x] Refactoring of QD solvers
   - [x] introduced a base class for implicit and explicit integrators
   - [x] introduced dedicated variables for shear and normal tractions and for background stresses.
   - [x] added coupling with any geos solver (provided that it knows how to compute tractions)
- [x] Introduced a `OneWayCoupled` solver to solve for flow in the fracture and mechanics around it separately. It could probably be done through a Poromechanics solver but I don't think we really want this option to persist. It is just needed to run one of the SCEC benchmarks.
- [x] Add SCEC-BP6 benchmark input files.

PR #3440, opened 2024-11-13
feat: Add Lagrange contact solver with bubble stabilization (stick only)

CusiniM:
I have stripped just the contact solver part out of https://github.com/GEOS-DEV/GEOS/pull/3416

- [x] Move contact assembly kernels into a subfolder
- [x] Adds Lagrange contact kernel with bubble function stabilization (for stick only)
- [x] Adds a field to store the jump that one wants to impose on the fault. 
- [x] Add an integratedTest.


CusiniM:
This should be ready to be reviewed and I would like to merge it quickly so that we can continue with the other PR with the full coupling with rate and state.

inputFiles/lagrangianContactMechanics/LagrangeContactBubbleStab_singleFracCompression_base.xml
```diff
@@ -0,0 +1,134 @@
+<?xml version="1.0" ?>
+
+<Problem>
+  <Solvers
+    gravityVector="{0.0, 0.0, 0.0}">
+
+    <SolidMechanicsLagrangeContactBubbleStab
+      name="AugmentedLagrangian"
```
ryar9534: This is dumb, but should we rename this solver so if future people use it they dont get confused? Because SolidMechanicsLagrangeContactBubbleStab im assuming does not define an augmented lagrange solver

ryar9534 (approved):
Seems like at some point we will want to unify the implementations of the lagrange contact solvers ("jump" stabilized vs bubbles), but thats a ways away

inputFiles/lagrangianContactMechanics/LagrangeContactBubbleStab_singleFracCompression_base.xml
CusiniM: yeah, this is not augemented lagrange it's just the Lagrange multiplier plus bubble that solves directly the saddle point system.

CusiniM:
> Seems like at some point we will want to unify the implementations of the lagrange contact solvers ("jump" stabilized vs bubbles), but thats a ways away

yeah, I plan to completely get rid of it.

Guotong-Ren (approved):
look forward to the see the bubble stabilization

inputFiles/lagrangianContactMechanics/LagrangeContactBubbleStab_singleFracCompression_base.xml
ryar9534: so I think in the input file we should just change the name field so that its not AugmentedLagrangian

CusiniM: yeah, I ll change it.

CusiniM:
```suggestion
      name="contactSolver"
```

inputFiles/lagrangianContactMechanics/LagrangeContactBubbleStab_singleFracCompression_smoke.xml
```diff
@@ -0,0 +1,60 @@
...
+    maxTime="1.0">
+    <SoloEvent
+      name="preFracture"
+      target="/Solvers/SurfaceGen"/>
+
+    <PeriodicEvent
+      name="solverApplications"
+      beginTime="0.0"
+      forceDt="1.0"
+      target="/Solvers/AugmentedLagrangian"/>
```
CusiniM:
```suggestion
      target="/Solvers/contactSolver"/>
```

src/coreComponents/physicsSolvers/contact/ContactSolverBase.cpp
```diff
@@ -116,6 +119,43 @@ void ContactSolverBase::setFractureRegions( dataRepository::Group const & meshBo
                  InputError );
 }
 
+real64 ContactSolverBase::solverStep( real64 const & time_n,
+                                      real64 const & dt,
+                                      const integer cycleNumber,
+                                      DomainPartition & domain )
+{
+  if( cycleNumber == 0 )
```
paveltomin: image

src/coreComponents/physicsSolvers/contact/SolidMechanicsLagrangeContactBubbleStab.cpp
```diff
@@ -0,0 +1,1150 @@
...
+                                               domain,
+                                               dofManager,
+                                               localMatrix,
+                                               localRhs );
+
+  assembleStabilization( dt, domain, dofManager, localMatrix, localRhs );
+
+  assembleContact( dt, domain, dofManager, localMatrix, localRhs );
+
+  // ParallelMatrix parallel_matrix;
```
paveltomin: debug

src/coreComponents/physicsSolvers/contact/SolidMechanicsLagrangeContactBubbleStab.cpp
```diff
@@ -0,0 +1,1150 @@
...
+}
+
+void SolidMechanicsLagrangeContactBubbleStab::implicitStepSetup( real64 const & time_n,
+                                                                 real64 const & dt,
+                                                                 DomainPartition & domain )
+{
+
+  SolidMechanicsLagrangianFEM::implicitStepSetup( time_n, dt, domain );
+
+  forDiscretizationOnMeshTargets( domain.getMeshBodies(), [&] ( string const &,
```
paveltomin, line 256: why doing this every time step?

src/coreComponents/physicsSolvers/contact/SolidMechanicsLagrangeContactBubbleStab.cpp
```diff
@@ -0,0 +1,1150 @@
...
+                                     { getUniqueFractureRegionName() } );
+
+    CommunicationTools::getInstance().synchronizeFields( fieldsToBeSync,
+                                                         mesh,
+                                                         domain.getNeighbors(),
+                                                         true );
+  } );
+}
+
+void SolidMechanicsLagrangeContactBubbleStab::updateState( DomainPartition & domain )
```
paveltomin: remove?

src/coreComponents/physicsSolvers/contact/SolidMechanicsLagrangeContactBubbleStab.cpp
```diff
@@ -0,0 +1,1150 @@
...
+{
+  GEOS_UNUSED_VAR( domain );
+  GEOS_MARK_FUNCTION;
+}
+
+real64 SolidMechanicsLagrangeContactBubbleStab::setNextDt( real64 const & currentDt,
```
paveltomin: remove?

src/coreComponents/physicsSolvers/contact/SolidMechanicsLagrangeContactBubbleStab.cpp
```diff
@@ -0,0 +1,1150 @@
...
+          }
+        }
+
+      }
+    }
+  } );
+
+}
+
+void SolidMechanicsLagrangeContactBubbleStab::updateStickSlipList( DomainPartition const & domain )
```
paveltomin, line 883: who is that StickSlipList ?

paveltomin (approved):

src/coreComponents/physicsSolvers/contact/ContactSolverBase.cpp
CusiniM: we really need to work on that additional initialization stage.

CusiniM: at least with this implementation it really only applies the `initialConditions` to the fractures.

src/coreComponents/physicsSolvers/contact/SolidMechanicsLagrangeContactBubbleStab.cpp
CusiniM: yeah, if theory we don't need to. I mean... we only need to do it for newly created elements. I guess I could move this somewhere.

src/coreComponents/physicsSolvers/contact/SolidMechanicsLagrangeContactBubbleStab.cpp
CusiniM:
```suggestion
```

src/coreComponents/physicsSolvers/contact/SolidMechanicsLagrangeContactBubbleStab.cpp
matteofrigo5: We’ll need this when extending the kernel to slip mode. Specifically, we need to know which elements are in stick and slip states to apply the appropriate kernels to each list.

matteofrigo5 (approved):

.integrated_tests.yaml
```diff
@@ -1,6 +1,6 @@
 baselines:
   bucket: geosx
-  baseline: integratedTests/baseline_integratedTests-pr3339-8707-7c55c70
+  baseline: integratedTests/baseline_integratedTests-pr3440-8820-344280f
```
```suggestion
  baseline: integratedTests/baseline_integratedTests-pr3440-8830-c296adc
```

src/coreComponents/physicsSolvers/contact/SolidMechanicsLagrangeContactBubbleStab.cpp
```diff
@@ -0,0 +1,1180 @@
...
+      //                                            [&]( FieldSpecificationBase const & bc,
+      //                                                 string const &,
+      //                                                 SortedArrayView< localIndex const > const & targetSet,
+      //                                                 SurfaceElementSubRegion & targetGroup,
+      //                                                 string const fieldName )
+      //       {
+      //         bc.applyFieldValue< FieldSpecificationEqual >( targetSet, 0.0, targetGroup, fieldName );
+      //       } );
+      //     }
+      //   } );
```
rrsettgast: I don't understand the explanation here.

src/coreComponents/physicsSolvers/contact/SolidMechanicsLagrangeContactBubbleStab.cpp
rrsettgast: perhaps there can be a specific function that updates newly created elements.

rrsettgast (approved):

src/coreComponents/physicsSolvers/contact/SolidMechanicsLagrangeContactBubbleStab.cpp
CusiniM: the way we loop over meshObjects is a bit different from some of the other loops.

CusiniM: [Here](https://github.com/GEOS-DEV/GEOS/blob/d9bdfe715b6eb89674298a50bc43d8b1a3021d86/src/coreComponents/mesh/MeshObjectPath.hpp#L316) it looks like that we check if the object is a ElementSubRegionBase but not whether the type matches the `OBJECT_TYPE` specified. I could add it but:
1. I wonder if it's a design choice so I want to do it in a separate PR.
2. I think that if we do what we discussed to specify what field specifications are applied, all these small issues should go away.

rrsettgast: got it. I think that part of code is unfinished.

src/coreComponents/physicsSolvers/contact/ContactSolverBase.cpp
paveltomin: @CusiniM would that work when you have mechanics initialization step running as `cycleNumber == 0`?

CusiniM: Certainly not but we need to first figure out how to initialize problems with faults and then abstract the way initialization is performed. 

In any case this solver is only used for Rate and state atm so I am not too worried about it. We could move it to SurfaceGenerator::execute though and then I don't think it would ever be an issue.

PR #3443, opened 2024-11-14
feat: Overall Composition (Z) Formulation

### LINKED ISSUE #2691: Saturation values above 1
GEOS formulation allows saturation values to go above 1 (not sure about below zero).
The computed saturation values are used as is to compute rel perms that might be not really physical thing to do and also may mislead the solver blocking some flow or allowing too much flow.
More reasonable way would be to normalize saturation values used for rel perm computes, see, e.g., https://drive.google.com/open?id=10dHD58pO8dy5hSz4RjVo8m0Mjs6C0a69

rasimHZ:
This feature adds a new molar formulation for compositional flow, whereby the primary unknowns are pressure and overall compositions ($z_c$) instead of pressure and component densities ($\rho_c$). The volume balance is then honored strictly - total density $\rho_T$ is found as function of pressure and compositions.


src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -368,27 +400,48 @@ void CompositionalMultiphaseBase::registerDataOnMesh( Group & meshBodies )
...
+        subRegion.registerField< globalCompFraction >( getName() ).
+          setDimLabels( 1, fluid.componentNames() ).
+          reference().resizeDimension< 1 >( m_numComponents );
+        subRegion.registerField< dGlobalCompFraction_dGlobalCompDensity >( getName() ).
+          reference().resizeDimension< 1, 2 >( m_numComponents, m_numComponents );  
       }
-
-      subRegion.registerField< globalCompFraction >( getName() ).
+      // for Dirichlet boundary condition
+      subRegion.registerField< fields::flow::bcglobalCompFraction >( getName() ).
```
paveltomin: maybe with upper case `G` here: `bcGlobalCompFraction`

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -669,6 +722,27 @@ real64 CompositionalMultiphaseBase::updatePhaseVolumeFraction( ObjectManagerBase
...
 
+real64 CompositionalMultiphaseBase::updatePhaseVolumeFractionZFormulation( ObjectManagerBase & dataGroup ) const
+{
+  GEOS_MARK_FUNCTION;
+
+  string const & fluidName = dataGroup.getReference< string >( viewKeyStruct::fluidNamesString() );
+  MultiFluidBase const & fluid = getConstitutiveModel< MultiFluidBase >( dataGroup, fluidName );
+
+  // For now: isothermal only
+  GEOS_ERROR_IF( m_isThermal, GEOS_FMT( "CompositionalMultiphaseBase {}: Z Formulation is currently not available for thermal simulations", getDataContext() ) );
```
paveltomin: no need for `CompositionalMultiphaseBase`, I think `getDataContext()` should provide enough info

paveltomin: and this check should be in `postInputInitialization`

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -817,16 +918,35 @@ real64 CompositionalMultiphaseBase::updateFluidState( ElementSubRegionBase & sub
...
   GEOS_MARK_FUNCTION;
 
-  updateGlobalComponentFraction( subRegion );
-  updateFluidModel( subRegion );
-  updateCompAmount( subRegion );
-  real64 const maxDeltaPhaseVolFrac = updatePhaseVolumeFraction( subRegion );
-  updateRelPermModel( subRegion );
-  updatePhaseMobility( subRegion );
-  updateCapPressureModel( subRegion );
+  real64 maxdS;
```
paveltomin: just use `maxDeltaPhaseVolFrac`

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -2104,24 +2375,14 @@ void CompositionalMultiphaseBase::chopNegativeDensities( DomainPartition & domai
...
-  {
-    return LvArray::NumericLimits< real64 >::max;
-  }
+  GEOS_MARK_FUNCTION;
 
-  real64 maxRelativePresChange = 0.0;
-  real64 maxRelativeTempChange = 0.0;
-  real64 maxAbsolutePhaseVolFracChange = 0.0;
-  real64 maxRelativeCompDensChange = 0.0;
+  using namespace isothermalCompositionalMultiphaseBaseKernels;
```
paveltomin, line 2235: probably not needed here

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.cpp
```diff
@@ -289,6 +293,53 @@ void CompositionalMultiphaseFVM::assembleFluxTerms( real64 const dt,
...
+    FiniteVolumeManager const & fvManager = numericalMethodManager.getFiniteVolumeManager();
+    FluxApproximationBase const & fluxApprox = fvManager.getFluxApproximation( m_discretizationName );
+
+    string const & elemDofKey = dofManager.getKey( viewKeyStruct::elemDofFieldString() );
+
+    fluxApprox.forAllStencils( mesh, [&] ( auto & stencil )
+    {
+      typename TYPEOFREF( stencil ) ::KernelWrapper stencilWrapper = stencil.createKernelWrapper();
+
+      GEOS_ERROR_IF( m_isThermal, GEOS_FMT(
```
paveltomin: move checks somewhere earlier

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.cpp
```diff
@@ -503,73 +671,52 @@ real64 CompositionalMultiphaseFVM::scalingForSystemSolution( DomainPartition & d
...
-                                                     pressureScalingFactor,
-                                                     compDensScalingFactor,
-                                                     dofManager.rankOffset(),
-                                                     m_numComponents,
-                                                     dofKey,
-                                                     subRegion,
-                                                     localSolution );
+      arrayView1d< real64 > compFracScalingFactor = subRegion.getField< fields::flow::globalCompFractionScalingFactor >();
+
+      GEOS_ERROR_IF( m_isThermal, GEOS_FMT( "CompositionalMultiphaseBase {}: Z Formulation is currently not available for thermal simulations", getDataContext() ) );
```
paveltomin: remove

src/coreComponents/physicsSolvers/fluidFlow/kernels/compositional/zFormulation/AccumulationZFormulationKernel.hpp
```diff
@@ -0,0 +1,417 @@
...
+  /**
+   * @brief Compute the local volume balance contributions to the residual and Jacobian
+   * @tparam FUNC the type of the function that can be used to customize the kernel
+   * @param[in] ei the element index
+   * @param[inout] stack the stack variables
+   * @param[in] phaseVolFractionSumKernelOp the function used to customize the kernel
+   */
+  template< typename FUNC = NoOpFunc >
+  GEOS_HOST_DEVICE
+  void computeVolumeBalance( localIndex const ei,
```
paveltomin: maybe rename to avoid confusion

src/coreComponents/physicsSolvers/fluidFlow/kernels/compositional/zFormulation/PotGradZFormulation.hpp
```diff
@@ -0,0 +1,178 @@
...
+      localIndex const er  = seri[i];
+      localIndex const esr = sesri[i];
+      localIndex const ei  = sei[i];
+
+      // density
+      real64 const density  = phaseMassDens[er][esr][ei][0][ip];
+      real64 const dDens_dP = dPhaseMassDens[er][esr][ei][0][ip][Deriv::dP];
+
+      // average density and derivatives
+      densMean += 0.5 * density;
```
paveltomin: this needs to be synced with current treatment

rasimHZ: Thanks for pointing this out. Synced.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
rasimHZ: fixed

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
rasimHZ: fixed

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.cpp
rasimHZ: done

src/coreComponents/physicsSolvers/fluidFlow/kernels/compositional/zFormulation/AccumulationZFormulationKernel.hpp
rasimHZ: changed to computeCompFracSum

paveltomin:
rebaseline is needed because of
```
Error: /Problem/Solvers/compflow
	Group has a child 'targetCompFracChangeInTimeStep' in the file to compare but not the baseline file.
Error: /Problem/Solvers/compflow
	Group has a child 'useZFormulation' in the file to compare but not the baseline file.
Error: /Problem/Solvers/compflow
	Group has a child 'minCompFrac' in the file to compare but not the baseline file.
```

paveltomin (approved):

paveltomin:
@dkachuma @tjb-ltk @CusiniM this is ready to have a look

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -221,11 +242,45 @@ void CompositionalMultiphaseBase::postInputInitialization()
...
+    if( m_isJumpStabilized ) // useZFormulation is not yet compatible with pressure stabilization
+    {
+      GEOS_ERROR( GEOS_FMT( "{}: pressure stabilization is not yet supported by {}",
+                            getDataContext(), viewKeyStruct::useZFormulationFlagString() ) );
+    }
+
+    DomainPartition & domain = this->getGroupByPath< DomainPartition >( "/Problem/domain" );
+    NumericalMethodsManager const & numericalMethodManager = domain.getNumericalMethodManager();
+    FiniteVolumeManager const & fvManager = numericalMethodManager.getFiniteVolumeManager();
+    FluxApproximationBase const & fluxApprox = fvManager.getFluxApproximation( m_discretizationName );
```
dkachuma: Not unique to this PR but I think `m_discretizationName` has no validation check. So if the user has a typo for example in the name, this might return null for `fluxApprox`.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -286,8 +341,21 @@ void CompositionalMultiphaseBase::registerDataOnMesh( Group & meshBodies )
...
   }
 
-  // n_c components + one pressure ( + one temperature if needed )
-  m_numDofPerCell = m_isThermal ? m_numComponents + 2 : m_numComponents + 1;
+
+  if( m_useZFormulation )
+  {
+    // Z formulation - component densities and pressure are primary unknowns
+    // (n_c-1) overall compositions + one pressure
+    // Testing: let's have sum_c zc = 1 as explicit equation for now
```
dkachuma: I suppose there's a plan to reduce this to `nc` later?

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -2069,100 +2260,276 @@ real64 CompositionalMultiphaseBase::setNextDtBasedOnStateChange( real64 const &
...
         }
       } );
+    } );
+  } );
+}
+
+real64 CompositionalMultiphaseBase::setNextDtBasedOnStateChange( real64 const & currentDt,
+                                                                 DomainPartition & domain )
+{
+  // TODO: put in a seprate function or put flags to avoid duplication of code
```
dkachuma: Agree

src/coreComponents/physicsSolvers/fluidFlow/kernels/compositional/zFormulation/DirichletFluxComputeZFormulationKernel.hpp
```diff
@@ -0,0 +1,539 @@
...
+      {
+        dDensMean_dC[jc] = 0.5 * m_dPhaseMassDens[er][esr][ei][0][ip][Deriv::dC+jc];
+      }
+
+
+      // Step 3.2: compute the (TPFA) potential difference at the face
+
+      real64 const gravTimesDz = m_gravCoef[er][esr][ei] - m_faceGravCoef[kf];
+      real64 const potDif = m_pres[er][esr][ei] - m_facePres[kf] - densMean * gravTimesDz;
+      real64 const f = stack.transmissibility * potDif;
```
dkachuma, line 299: This was a bit difficult to follow below. Can we use a more descriptive name than `f`.

dkachuma (commented):
I can see that most of the kernels are redone. Is there quite some code duplication here. Is there a way we can reuse the same implementations?

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
paveltomin: this https://github.com/GEOS-DEV/GEOS/blob/212eb6ba24976c8c543b70d19995a8328dee2d93/src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.cpp#L171 does not work?

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
rasimHZ: Yes. Currently, the individual derivatives of w.r.t each z_c are computed with flash. When reduced, these derivatives have to be adjusted taking into account constaint sum_c zc = 1. It is possible to recompute the reduced derivatives after the flash using the current output of flash, but I think that might be redundant. I think we should consider a better way.

src/coreComponents/physicsSolvers/fluidFlow/kernels/compositional/zFormulation/DirichletFluxComputeZFormulationKernel.hpp
paveltomin: i think it uses same notations as the original code, if rename - rename both places then

inputFiles/compositionalMultiphaseFlow/overallCompositionFormulation/co2_2d_plume/co2_2d_plume_Z.xml
```diff
@@ -0,0 +1,233 @@
...
+    <!-- Injection rate: 0.002 kg/s -->
+    <SourceFlux
+      name="sourceTerm"
+      objectPath="ElementRegions/region/cellBlock"
+      component="0"
+      scale="-0.002"
+      setNames="{ source }"/>
+
+    <!-- Boundary conditions at right end  -->
+    <FieldSpecification
```
dkachuma, line 172: These boundary conditions are applied on the faces. Is there a different code path if we impose a Dirichlet bc on actual elements instead of faces?

dkachuma (approved):
Looks good to me.

PR #3444, opened 2024-11-15
feat: This PR makes the timestep summary information logging introduced in #3303 contingent on the log level of the solver

sframba:
If the timesteps are very quick and in a large number (for exemple, for many wave proagation problems), the summary timestep information introduced in #3303 produces a lot of logging (tens of thousands of lines), slowing the solver down and producing unreadable log files.
This PR makes the output dependent on the user's choice for the log level. The same log-level-dependent call is used as for all other timestep logging information in the `PhysicsSolverBase` class.

### DIFF
--- a/src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
+++ b/src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
@@ -350,17 +350,17 @@ void PhysicsSolverBase::logEndOfCycleInformation( integer const cycleNumber,
                                                   std::vector< real64 > const & subStepDt ) const
 {
   // The formating here is a work in progress.
-  GEOS_LOG_RANK_0( "\n------------------------- TIMESTEP END -------------------------" );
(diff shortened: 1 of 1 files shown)


paveltomin (approved):

rrsettgast (approved):
@sframba Can you confirm that this works turning on/off? And can we set the `logLevel` values differently based on the time integration?

MelReyCG (approved):

sframba:
> @sframba Can you confirm that this works turning on/off? And can we set the `logLevel` values differently based on the time integration?

I revise my comment: yes it seems to be working. It looks like you need a loglevel >=1 to have the output.
As for the log level, I think there is only one struct today for all TimeStep logging info : [`src/coreComponents/physicsSolvers/LogLevelsInfo.hpp`](https://github.com/GEOS-DEV/GEOS/blob/e9492b5354039ac641c27996b65b502d006e6883/src/coreComponents/physicsSolvers/LogLevelsInfo.hpp#L62). But I know that @MelReyCG is working on harmonizing these log level.
Anyhow, the PR seems to be working, so OK from me for the merge

PR #3445, opened 2024-11-15
fix: bugfix about norm vector size

paveltomin:
many thanks to @tjb-ltk for finding it


paveltomin:
@CusiniM and/or @ryar9534 let's merge this please

ryar9534:
Was the error just that `globalResidualNorm` was scoped incorrectly?

ryar9534 (approved):

paveltomin:
> Was the error just that `globalResidualNorm` was scoped incorrectly?

yes, local and global had different sizes but MPI function was not aware of that leading to wrong memory access

rrsettgast (approved):

CusiniM (approved):

CusiniM:
I have https://github.com/GEOS-DEV/GEOS/pull/3446 that should prevent similar errors from happening.

PR #3446, opened 2024-11-15
feat: MpiWrapper::allReduce overload for arrays.

CusiniM:
We had a bug because two arrays of different size were exchanged by grabbing the buffers directly. This overload will prevent it from happening. 

I had to add a small helper coz `LvArray` has `ValueType` instead of `value_type` like `std` objects.


src/coreComponents/common/TypesHelpers.hpp
```diff
@@ -0,0 +1,66 @@
...
+ * Copyright (c) 2023-2024 Chevron
+ * Copyright (c) 2019-     GEOS/GEOSX Contributors
+ * All rights reserved
+ *
+ * See top level LICENSE, COPYRIGHT, CONTRIBUTORS, NOTICE, and ACKNOWLEDGEMENTS files for details.
+ * ------------------------------------------------------------------------------------------------------------
+ */
+
+/**
+ * @file TypeHelpers.hpp
```
CusiniM:
```suggestion
 * @file TypesHelpers.hpp
```

src/coreComponents/common/TypesHelpers.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+template< typename T >
+struct has_ValueType< T, std::void_t< typename T::ValueType > > : std::true_type {};
+
+}   // namespace internal
+
+template< typename T, typename Enable = void >
+struct get_value_type
+{
+  static_assert( sizeof(T) == 0, "CONTAINER_TYPE must define either value_type or ValueType." );
```
CusiniM:
```suggestion
  static_assert( sizeof(T) == 0, "T must define either value_type or ValueType." );
```

CusiniM:
@rrsettgast , @corbett5 , @wrtobin this is ready for review.

src/coreComponents/common/TypesHelpers.hpp
```diff
@@ -0,0 +1,66 @@
...
+};
+
+template< typename T >
+struct get_value_type< T, std::enable_if_t< !internal::has_value_type< T >::value && internal::has_ValueType< T >::value > >
+{
+  using type = typename T::ValueType;
+};
+
+} // namespace geos
+
```
rrsettgast: doxygen

rrsettgast (approved):

MelReyCG (commented):
At this point, can we drop or at least set to private the following signature?
```
template< typename T >
  static int allReduce( T const * sendbuf, T * recvbuf, int count, MPI_Op op, MPI_Comm comm = MPI_COMM_GEOS );
```

All external calls seem to be able to be translated to `Span<T[]>`.

CusiniM:
> At this point, can we drop or at least set to private the following signature?
> 
> ```
> template< typename T >
>   static int allReduce( T const * sendbuf, T * recvbuf, int count, MPI_Op op, MPI_Comm comm = MPI_COMM_GEOS );
> ```
> 
> All external calls seem to be able to be translated to `Span<T[]>`.

yes, you are right, let me try to do this.

src/coreComponents/common/TypesHelpers.hpp
CusiniM: done

CusiniM:
> At this point, can we drop or at least set to private the following signature?
> 
> ```
> template< typename T >
>   static int allReduce( T const * sendbuf, T * recvbuf, int count, MPI_Op op, MPI_Comm comm = MPI_COMM_GEOS );
> ```
> 
> All external calls seem to be able to be translated to `Span<T[]>`.

done.

paveltomin (approved):

matteofrigo5 (approved):

CusiniM:
@sframba or @acitrain can you please have a quick look at the changes to the wave solvers files and approve this?

src/coreComponents/physicsSolvers/wavePropagation/sem/elastic/secondOrderEqn/isotropic/ElasticWaveEquationSEM.cpp
```diff
@@ -1013,10 +1013,9 @@ void ElasticWaveEquationSEM::cleanup( real64 const time_n,
       computeAllSeismoTraces( time_n, 0.0, uy_np1, uy_n, dasReceivers, m_linearDASVectorY.toView(), true );
       computeAllSeismoTraces( time_n, 0.0, uz_np1, uz_n, dasReceivers, m_linearDASVectorZ.toView(), true );
       // sum contributions from all MPI ranks, since some receivers might be split among multiple ranks
-      MpiWrapper::allReduce( dasReceivers.data(),
-                             dasReceivers.data(),
-                             m_linearDASGeometry.size( 0 ),
-                             MpiWrapper::getMpiOp( MpiWrapper::Reduction::Sum ),
+      MpiWrapper::allReduce( dasReceivers,
+                             dasReceivers,
+                             MpiWrapper::Reduction::Sum,
```
CusiniM: @sframba @acitrain  this change triggers these diffs:

````
NFO: Total number of log files processed: 1537

WARNING: Found unfiltered diff in: /Users/cusini1/Downloads/integratedTests/TestResults/test_data/wavePropagation/elas3D_DAS_smoke_08/1497.python3_elas3D_DAS_smoke_08_2_restartcheck_.log
INFO: Details of diffs:   ********************************************************************************

  Error: /Problem/Solvers/elasticSolver/dasSignalNp1AtReceivers

  	Arrays of types float32 and float32 have 402 values of which 200 fail both the relative and absolute tests.

  		Max absolute difference is at index (np.int64(200), np.int64(1)): value = 0.2, base_value = 0.0

  		Max relative difference is at index (np.int64(200), np.int64(1)): value = 0.2, base_value = 0.0

  	Statistics of the q values greater than 1.0 defined by absolute tolerance: N = 200

  		max = 2000.0001, mean = 500.0, std = 646.787

  Error: /Problem/Solvers/elasticSolver/dasSignalNp1AtReceivers

  	Arrays of types float32 and float32 have 402 values of which 200 fail both the relative and absolute tests.

  		Max absolute difference is at index (np.int64(200), np.int64(1)): value = 0.2, base_value = 0.0

  		Max relative difference is at index (np.int64(200), np.int64(1)): value = 0.2, base_value = 0.0

  	Statistics of the q values greater than 1.0 defined by absolute tolerance: N = 200

  		max = 2000.0001, mean = 500.0, std = 646.787

  Error: /Problem/Solvers/elasticSolver/dasSignalNp1AtReceivers

  	Arrays of types float32 and float32 have 402 values of which 200 fail both the relative and absolute tests.

  		Max absolute difference is at index (np.int64(200), np.int64(1)): value = 0.2, base_value = 0.0

  		Max relative difference is at index (np.int64(200), np.int64(1)): value = 0.2, base_value = 0.0

  	Statistics of the q values greater than 1.0 defined by absolute tolerance: N = 200

  		max = 2000.0001, mean = 500.0, std = 646.787

  Error: /Problem/Solvers/elasticSolver/dasSignalNp1AtReceivers

  	Arrays of types float32 and float32 have 402 values of which 200 fail both the relative and absolute tests.

  		Max absolute difference is at index (np.int64(200), np.int64(1)): value = 0.2, base_value = 0.0

  		Max relative difference is at index (np.int64(200), np.int64(1)): value = 0.2, base_value = 0.0

  	Statistics of the q values greater than 1.0 defined by absolute tolerance: N = 200

  		max = 2000.0001, mean = 500.0, std = 646.787

  Error: /Problem/Solvers/elasticSolver/dasSignalNp1AtReceivers

  	Arrays of types float32 and float32 have 402 values of which 200 fail both the relative and absolute tests.

  		Max absolute difference is at index (np.int64(200), np.int64(1)): value = 0.2, base_value = 0.0

  		Max relative difference is at index (np.int64(200), np.int64(1)): value = 0.2, base_value = 0.0

  	Statistics of the q values greater than 1.0 defined by absolute tolerance: N = 200

  		max = 2000.0001, mean = 500.0, std = 646.787

  Error: /Problem/Solvers/elasticSolver/dasSignalNp1AtReceivers

  	Arrays of types float32 and float32 have 402 values of which 200 fail both the relative and absolute tests.

  		Max absolute difference is at index (np.int64(200), np.int64(1)): value = 0.2, base_value = 0.0

  		Max relative difference is at index (np.int64(200), np.int64(1)): value = 0.2, base_value = 0.0

  	Statistics of the q values greater than 1.0 defined by absolute tolerance: N = 200

  		max = 2000.0001, mean = 500.0, std = 646.787

  Error: /Problem/Solvers/elasticSolver/dasSignalNp1AtReceivers

  	Arrays of types float32 and float32 have 402 values of which 200 fail both the relative and absolute tests.

  		Max absolute difference is at index (np.int64(200), np.int64(1)): value = 0.2, base_value = 0.0

  		Max relative difference is at index (np.int64(200), np.int64(1)): value = 0.2, base_value = 0.0

  	Statistics of the q values greater than 1.0 defined by absolute tolerance: N = 200

  		max = 2000.0001, mean = 500.0, std = 646.787
````

The only explanation I can find is that, in the previous code, m_linearDASGeometry.size( 0 ) was not the size of the array. The fact that `arrayView2d< real32 > const dasReceivers  = m_dasSignalNp1AtReceivers.toView();` makes me wonder why  `m_linearDASGeometry.size( 0 )` was passed as the size. I am not sure how these two objects are related...

sframba: Hi @CusiniM , I think that is exactly the reason: we have an extra receiver that simply tracks time: 

https://github.com/GEOS-DEV/GEOS/blob/fd36d2adcb737e6eef6c7dfcdb61b060bd98d7ef/src/coreComponents/physicsSolvers/wavePropagation/sem/elastic/secondOrderEqn/isotropic/ElasticWaveEquationSEM.cpp#L221

This receiver does not need to be summed across ranks. Therefore only the first `dasReceivers .size() - 1` (that is, `m_linearDASGeometry.size( 0 )`) arrays need to be summed, not the last one. Does this break your new structure or can it be accomodated?

CusiniM: Yeah, it does break it in the sense that the overload I put in place for `arrays` was written with the idea of summing the full array across ranks. Basically I wanted to ensure that one does not provide the wrong size (which caused a couple of bugs in the past). We could add an overload to allow for a the reduce operation to occur on a specified `size <= array.size()`.

rrsettgast: @sframba Is there a reason that this all has to be in one array? It seems that the array is holding different quantity types? If so, can we just create a separate object to hold the last value that is of different quantity type?

CusiniM:
@sframba please have a look now and approve.

src/coreComponents/physicsSolvers/wavePropagation/sem/elastic/secondOrderEqn/isotropic/ElasticWaveEquationSEM.cpp
sframba: @rrsettgast as discussed together, no there is no reason why this should be all one array. We could (should) separate the extra array so that they become of homogeneous dimension. However, this requires some changes in our python code as well. Could we maybe :
- Implement the overload suggested by @CusiniM for now, so we unlock this PR, then
- When the "strong unit typing" work starts officially, add an item about wave solvers' receiver uniformity and assign it to us, so we can take the time to tidy up this problem properly?
What do you guys think?

CusiniM: I have already implemented it here: https://github.com/GEOS-DEV/GEOS/blob/c18b6313d71759026645b7b353a5aecf1db5f893/src/coreComponents/common/MpiWrapper.hpp#L1152

CusiniM: Tests are now passing so this just needs @sframba 's approval now.

PR #2580, opened 2023-07-15
Parallel import of conformal fractures

TotoGaz:

PR #3381, opened 2024-10-02
refactor: Clean up ghost unpacking

rrsettgast:
- [x] unpack ghost objects and their data separately in `CommunicationTools::setupGhosts`
- [x] unpack ghost objects and their data objects in `ParallelTopologyChange`

Issue #2821, opened 2023-11-15
VTK mesh redistribution with default partitionRefinement (segmentatation fault)


castelletto1:
**Describe the bug**
Running GEOS in serial providing the mesh as VTK produces segmentation fault if the number of partitioning refinement iterations (`partitionRefinement`) is not `0` (default value `1`).

**To Reproduce**
See incompressible single-phase flow example below.

@francoishamon, @klevzoff, @untereiner : have you observed this behavior before?

```xml
<?xml version="1.0" ?>

<Problem>
  <Solvers
    gravityVector="{ 0.0, 0.0, 0.0 }">
    <SinglePhaseFVM
      name="SinglePhaseFlow"
      logLevel="1"
      discretization="singlePhaseTPFA"
      targetRegions="{ Domain }">
      <NonlinearSolverParameters
        newtonTol="1.0e-6"
        newtonMaxIter="8"/>
      <LinearSolverParameters
        directParallel="0"/>
    </SinglePhaseFVM>
  </Solvers>

  <Mesh>
    <VTKMesh
      name="mesh"
      partitionRefinement="1"
      file="mesh.vtk"/>
  </Mesh>

  <Events
    maxTime="1.0">
    <PeriodicEvent
      name="outputs"
      timeFrequency="1.0"
      target="/Outputs/vtkOutput"/>

    <PeriodicEvent
      name="solverApplications"
      forceDt="1.0"
      target="/Solvers/SinglePhaseFlow"/>

  </Events>

  <NumericalMethods>
    <FiniteVolume>
      <TwoPointFluxApproximation
        name="singlePhaseTPFA"/>
    </FiniteVolume>
  </NumericalMethods>

  <ElementRegions>
    <CellElementRegion
      name="Domain"
      cellBlocks="{ 0_hexahedra, 1_hexahedra, 2_hexahedra }"
      materialList="{ water, rock }"/>
  </ElementRegions>

  <Constitutive>
    <CompressibleSinglePhaseFluid
      name="water"
      defaultDensity="1000"
      defaultViscosity="0.001"
      referencePressure="0.0"
      compressibility="0.0"
      viscosibility="0.0"/>

    <CompressibleSolidConstantPermeability
      name="rock"
      solidModelName="nullSolid"
      porosityModelName="rockPorosity"
      permeabilityModelName="rockPerm"/>

    <NullModel
      name="nullSolid"/>

    <PressurePorosity
      name="rockPorosity"
[...]

PR #3448, opened 2024-11-16
fix: change FaceElementSubRegion maps to volume elements back to fixed size relations

rrsettgast:
The `FaceElementSubRegion::m_toFacesRelation` and `SurfaceElementSubRegion::m_2dElemToElems` were changed to an `ArrayOfArrays<localIndex>` in #2580 . This change is detrimental to the internal fracture capabilities. This PR changes them back to a `FixedSized` relation.

- Also. this PR fixes a bug in the fracture-matrix flow treatment. Originally the `FaceElementSubRegion:: m_toFacesRelation` was always guaranteed to be a 2d array of size `(numFaceElements, 2)` since a `faceElement` is always bounded by two `faces`. Furthermore, the first value was alway the original face, and the second value was always the new face. When a new fracture is created, the fracture disconnects two matrix elements that were formerly connected, thus the stencil has to be modified. This is done by setting the stencil weight of `FaceElementSubRegion:: m_toFacesRelation(faceElement,0)` equal to zero. In current `develop` the assumption of the of the original face in the first entry is not true. In fact the  `FaceElementSubRegion:: m_toFacesRelation` ordering is inconsistent between ranks. Thus the matrix cells were not correctly disconnected.


paveltomin:
should it resolve any of the issues from https://github.com/GEOS-DEV/GEOS/issues/2821 ?
i tried this case https://github.com/GEOS-DEV/GEOS/issues/2821#issuecomment-2032926093 and still got the same issue

rrsettgast:
> should it resolve any of the issues from #2821 ? i tried this case [#2821 (comment)](https://github.com/GEOS-DEV/GEOS/issues/2821#issuecomment-2032926093) and still got the same issue

I don't think so. I am just changing this back to the original types because:
1. they should have never been changed to `ArrayOfArrays` in the first place. 
2. the diffs in #3381 are too hard to deal with when they are `ArrayOfArrays`. As it stands, this PR has some problems with diffs because the order of the mappings seem to have been flipped from their original order by #2580.

src/coreComponents/dataRepository/BufferOps_inline.hpp
```diff
@@ -1762,7 +1762,11 @@ Pack( buffer_unit_type * & buffer,
       arraySlice1d< globalIndex const > const & relatedObjectLocalToGlobalMap )
 {
   localIndex sizeOfPackedChars = 0;
-  array1d< globalIndex > junk;
+  array1d< globalIndex > junk( var.size( 1 ) );
+  for( localIndex a=0; a<var.size( 1 ); ++a )
+  {
+    junk[a] = -1;
```
CusiniM: since this value now kind of has a meaning maybe we should not call it junk.

src/coreComponents/dataRepository/wrapperHelpers.hpp
```diff
@@ -188,18 +189,36 @@ resize( T & GEOS_UNUSED_PARAM( value ),
         localIndex const GEOS_UNUSED_PARAM( newSize ) )
 {}
 
-
-template< typename T, int NDIM, typename PERMUTATION >
-inline std::enable_if_t< DefaultValue< Array< T, NDIM, PERMUTATION > >::has_default_value >
-resizeDefault( Array< T, NDIM, PERMUTATION > & value,
+template< typename T >
+inline std::enable_if_t< traits::HasMemberFunction_resizeDefault< T > &&
```
CusiniM, line 193: why this change? for the `InterObjectRelation` ?

src/coreComponents/mesh/BufferOps.cpp
```diff
@@ -362,6 +362,77 @@ localIndex Unpack( buffer_unit_type const * & buffer,
 
   return sizeOfUnpackedChars;
 }
+#else
+
+localIndex Unpack( buffer_unit_type const * & buffer,
```
CusiniM: what's this code? is it still useful?

src/coreComponents/mesh/FaceManager.cpp
```diff
@@ -147,6 +150,8 @@ void FaceManager::setGeometricalRelations( CellBlockManagerABC const & cellBlock
 {
   GEOS_MARK_FUNCTION;
 
+  std::cout<<"FaceManager::setGeometricalRelations"<<std::endl;
```
CusiniM:
```suggestion
``` ?

File: src/coreComponents/mesh/FaceManager.cpp [hunk 5]
```diff
@@ -174,27 +179,46 @@ void FaceManager::setGeometricalRelations( CellBlockManagerABC const & cellBlock
...
+    //   if( rank == MpiWrapper::commRank() )
+    //   {
+    //     std::cout<<"FaceManager::setGeometricalRelations rank="<<rank<<std::endl;
+    //     for( int i=0; i<m_toElements.m_toElementRegion.size(0); ++i )
+    //     {
+    //       std::cout<<"m_toElements.m_toElementRegion["<<i<<"] = ( "<<m_toElements.m_toElementRegion(i,0)<<", "<<m_toElements.m_toElementSubRegion(i,0)<<", "<<m_toElements.m_toElementIndex(i,0)<<"), ("
+    //                                                                <<m_toElements.m_toElementRegion(i,1)<<", "<<m_toElements.m_toElementSubRegion(i,1)<<", "<<m_toElements.m_toElementIndex(i,1)<<" )"<<std::endl;
+    //     }
+    //   }
+    // }
```
CusiniM: remove?

src/coreComponents/mesh/FaceManager.cpp
```diff
@@ -174,27 +179,46 @@ void FaceManager::setGeometricalRelations( CellBlockManagerABC const & cellBlock
...
+    arrayView2d< localIndex const > const & elem2dToFaces = subRegion.faceList().toViewConst();
+    for( localIndex ei = 0; ei < elem2dToFaces.size(0); ++ei )
     {
-      for( localIndex const & face: elem2dToFaces[ei] )
+      for( localIndex const & faceIndex: elem2dToFaces[ei] )
       {
-        GEOS_ERROR_IF_EQ_MSG( m_toElements.m_toElementRegion( face, 0 ), -1, GEOS_FMT( err, face ) );
-        GEOS_ERROR_IF_EQ_MSG( m_toElements.m_toElementSubRegion( face, 0 ), -1, GEOS_FMT( err, face ) );
-        GEOS_ERROR_IF_EQ_MSG( m_toElements.m_toElementIndex( face, 0 ), -1, GEOS_FMT( err, face ) );
+        // std::cout<<"FaceManager::setGeometricalRelations ei="<<ei<<", faceIndex="<<faceIndex<<std::endl;
```
CusiniM:
```suggestion
```

src/coreComponents/mesh/ObjectManagerBase.hpp
```diff
@@ -1001,7 +1001,7 @@ void ObjectManagerBase::fixUpDownMaps( TYPE_RELATION & relation,
           allValuesMapped = false;
         }
       }
-      GEOS_ERROR_IF( relation[li][a] == unmappedLocalIndexValue, "Index not set" );
+      //GEOS_ERROR_IF( relation[li][a] == unmappedLocalIndexValue, "Index not set" );
```
CusiniM, line 1007:
```suggestion
```

rrsettgast:
@CusiniM Here is the restart diff of  `FaceElementSubRegion::m_toFacesRelation` using `h5diff` on `rank1` (`rank0` there are no diffs):
```
dataset: </Problem/domain/MeshBodies/mesh1/meshLevels/Level0/ElementRegions/elementRegionsGroup/Fracture/elementSubRegions/FractureSubRegion/faceList/__values__> and </Problem/domain/MeshBodies/mesh1/meshLevels/Level0/ElementRegions/elementRegionsGroup/Fracture/elementSubRegions/FractureSubRegion/faceList/__values__>
size:           [20]           [20]
position        __values__      __values__      difference          
------------------------------------------------------------
[ 0 ]          256             2               254            
[ 1 ]          2               256             254            
[ 2 ]          257             18              239            
[ 3 ]          18              257             239            
[ 4 ]          258             34              224            
[ 5 ]          34              258             224            
[ 6 ]          259             50              209            
[ 7 ]          50              259             209            
[ 8 ]          260             66              194            
[ 9 ]          66              260             194            
[ 10 ]          261             82              179            
[ 11 ]          82              261             179            
[ 12 ]          262             98              164            
[ 13 ]          98              262             164            
[ 14 ]          263             114             149            
[ 15 ]          114             263             149            
[ 16 ]          264             130             134            
[ 17 ]          130             264             134            
[ 18 ]          265             146             119            
[ 19 ]          146             265             119            
20 differences found
```
There are no diffs in the `FaceManager::m_localToGlobalMap`...given that fact, this is the output of the `FaceElementSubRegion::m_toFacesRelation` in terms of local and global indices for the changes in this PR:
```
Rank 0
    m_toFacesRelation [  0](  15, 256 ) -global-> [ 15](  15, 430 )
    m_toFacesRelation [  1](  31, 257 ) -global-> [ 31](  31, 431 )
    m_toFacesRelation [  2](  47, 258 ) -global-> [ 47](  47, 432 )
    m_toFacesRelation [  3](  63, 259 ) -global-> [ 63](  63, 433 )
    m_toFacesRelation [  4](  79, 260 ) -global-> [ 79](  79, 434 )
    m_toFacesRelation [  5](  95, 261 ) -global-> [ 95](  95, 435 )
    m_toFacesRelation [  6]( 111, 262 ) -global-> [111]( 111, 436 )
    m_toFacesRelation [  7]( 127, 263 ) -global-> [127]( 127, 437 )
    m_toFacesRelation [  8]( 143, 264 ) -global-> [143]( 143, 438 )
    m_toFacesRelation [  9]( 159, 265 ) -global-> [159]( 159, 439 )
Rank 1
    m_toFacesRelation [  0](   2, 256 ) -global-> [ 15](  15, 430 )
    m_toFacesRelation [  1](  18, 257 ) -global-> [ 31](  31, 431 )
    m_toFacesRelation [  2](  34, 258 ) -global-> [ 47](  47, 432 )
    m_toFacesRelation [  3](  50, 259 ) -global-> [ 63](  63, 433 )
    m_toFacesRelation [  4](  66, 260 ) -global-> [ 79](  79, 434 )
    m_toFacesRelation [  5](  82, 261 ) -global-> [ 95](  95, 435 )
    m_toFacesRelation [  6](  98, 262 ) -global-> [111]( 111, 436 )
    m_toFacesRelation [  7]( 114, 263 ) -global-> [127]( 127, 437 )
    m_toFacesRelation [  8]( 130, 264 ) -global-> [143]( 143, 438 )
    m_toFacesRelation [  9]( 146, 265 ) -global-> [159]( 159, 439 )
```

So if there are no changes in the `FaceManager::m_localToGlobalMap`, then the ranks in inconsistent in `develop`.

CusiniM:
Yeah they simply seem flipped here on rank 1. But isn't there a stage in develop where they get ordered by global index?

src/coreComponents/mesh/FaceElementSubRegion.cpp
```diff
@@ -42,6 +43,7 @@ FaceElementSubRegion::FaceElementSubRegion( string const & name,
   registerWrapper( viewKeyStruct::detJString(), &m_detJ ).setSizedFromParent( 1 ).reference();
 
   registerWrapper( viewKeyStruct::faceListString(), &m_toFacesRelation ).
+    setApplyDefaultValue( -1 ).
```
CusiniM, line 46: so this works now ?

CusiniM (approved):

src/coreComponents/dataRepository/wrapperHelpers.hpp
rrsettgast: yes. they can have a default value now.

src/coreComponents/mesh/BufferOps.cpp
rrsettgast: This is what I wanted to do to the function, as you can see the function that is used has a no-op branch in the conditional tree. I consider the original function incomplete. I was keeping this around until I fix it....which I can do now that the rest of the branch seems to work.

src/coreComponents/dataRepository/BufferOps_inline.hpp
rrsettgast: I can change it...it is a junk array for cases where we don't have a valid `localIndex`....but something like `invalidGlobalIndex` would be better.

src/coreComponents/mesh/FaceElementSubRegion.cpp
rrsettgast: yes. the change in the wrapperHelpers/Wrapper/traits files allow this.

src/coreComponents/mesh/ObjectManagerBase.hpp
rrsettgast: we want this to be here...but in a subsequent PR we are going to get rid of this concept.

src/coreComponents/dataRepository/wrapperHelpers.hpp
```diff
@@ -188,18 +189,37 @@ resize( T & GEOS_UNUSED_PARAM( value ),
...
+  //                            "  wrapperHelpers::resizeDefault<{}>() called, but the SFINAE filter failed:\n"
+  //                            "    traits::HasMemberFunction_resizeDefault< {} > = {}\n "
+  //                            "    DefaultValue< {} >::has_default_value = {}",
+  //                            LvArray::system::demangleType< T >(),
+  //                            name,
+  //                            LvArray::system::demangleType< T >(),
+  //                            LvArray::system::demangleType< T >(),
+  //                            traits::HasMemberFunction_resizeDefault< T >,
+  //                            LvArray::system::demangleType< T >(),
+  //                            DefaultValue< T >::has_default_value ) );
```
wrtobin: do we wanna remove this entirely or, if we want to retain it for future debugging, add an `ifdef` instead of just commenting it out?

src/coreComponents/mesh/EmbeddedSurfaceSubRegion.cpp
```diff
@@ -63,10 +64,25 @@ EmbeddedSurfaceSubRegion::EmbeddedSurfaceSubRegion( string const & name,
...
+  registerWrapper( viewKeyStruct::surfaceElementsToCellIndexString(), &m_2dElemToElems.m_toElementIndex ).
+    setPlotLevel( PlotLevel::NOPLOT ).
+    setDescription( "A map of face element local indices to the cell local indices" );
+
   m_normalVector.resizeDimension< 1 >( 3 );
   m_tangentVector1.resizeDimension< 1 >( 3 );
   m_tangentVector2.resizeDimension< 1 >( 3 );
   m_2dElemToElems.resize( 0, 1 );
+
+  m_2dElemToElems.setElementRegionManager( dynamicCast< ElementRegionManager & >( getParent().getParent().getParent().getParent() ) );
```
wrtobin, line 84: This is one of those situations I'm unsure whether we want absolute or relative addresses or e.g. a member function on a "static" Group/Node to setup relationships between Group/Nodes in the DataRepo/Graph, I feel like we're inconsistent with this and we should probably have some guidance/best practices related to this, not important for this PR but might be something to discuss at an infra meeting @rrsettgast

src/coreComponents/mesh/FaceElementSubRegion.cpp
```diff
@@ -156,22 +177,26 @@ void FaceElementSubRegion::copyFromCellBlock( FaceBlockABC const & faceBlock )
...
   // It will later be transformed into a sub regions mapping.
   // Last, we fill the regions mapping with dummy -1 values that should all be replaced eventually.
-  auto const elem2dToElems = faceBlock.get2dElemToElems();
-  m_2dElemToElems.resize( num2dElements, 2 );
-  for( int i = 0; i < num2dElements; ++i )
+  auto const & elem2dToElems = faceBlock.get2dElemToElems();
+  for( int kfe = 0; kfe < num2dElements; ++kfe )
   {
-    for( localIndex const & j: elem2dToElems.toCellIndex[i] )
+    for( localIndex k=0; k<elem2dToElems.toBlockIndex.sizeOfArray( kfe ); ++k )
```
wrtobin, line 183: is this one of the arrays we're swapping to known size (e.g. are the face elements in the map of the same topological type)? I would think we could know this at compile-time if so?

src/coreComponents/mesh/FaceElementSubRegion.cpp
```diff
@@ -299,7 +324,7 @@ localIndex FaceElementSubRegion::packUpDownMapsImpl( buffer_unit_type * & buffer
 
 localIndex FaceElementSubRegion::unpackUpDownMaps( buffer_unit_type const * & buffer,
                                                    localIndex_array & packList,
-                                                   bool const overwriteUpMaps,
+                                                   bool const,
```
wrtobin: `GEOS_UNUSED_PARAM`?

 or if its something we want to obsolete/remove either do so or maybe introduce a new macro to suppress warnings like: `GEOS_TODO_REMOVE_PARAM`.

src/coreComponents/mesh/ObjectManagerBase.hpp
wrtobin: wait so can't it be present in the follow-on PR and not in the merge?

wrtobin: wait why? for potential debugging of the develop branch or working branches after merge? if so I would prefer an `ifdef` to make it clear its intended to be present.

wrtobin (approved):
couple comments we might want to resolve, otherwise looks solid.

src/coreComponents/mesh/FaceElementSubRegion.cpp
wrtobin: I think I'm wrong on this, but I'll leave the comment just in case.

src/coreComponents/mesh/ObjectManagerBase.hpp
rrsettgast: @wrtobin What I mean is that there is more to clean up here. Specifically, with the impending ghost unpacking PR (https://github.com/GEOS-DEV/GEOS/pull/3381) that triggered this PR, we won't have a "fixup" step any longer. I wanted to just leave this commented out to point out that this check happened....and may be required to debug a subsequent PR which deletes the entire function.

src/coreComponents/mesh/FaceElementSubRegion.cpp
rrsettgast: Yeah...it is confusing, but this `elem2dToElems` is from the `FaceBlock`, which I didn't change.

src/coreComponents/dataRepository/wrapperHelpers.hpp
rrsettgast: I wrapped this in an `#if !defined(NDEBUG)`

src/coreComponents/mesh/EmbeddedSurfaceSubRegion.cpp
rrsettgast: We can just move this somewhere that has access to `MeshLevel`...that would be better.

src/coreComponents/mesh/FaceElementSubRegion.cpp
rrsettgast: I wrapped this in `GEOS_UNUSED_PARAM`

PR #3449, opened 2024-11-17
fix: bug in GEOS build

victorapm:
Fixes build issues:

1. Double explicit template instantiation when `localIndex` is equal to `globalIndex`
```bash
${GEOS_DIR}/src/coreComponents/common/format/StringUtilities.cpp:97:17: error: duplicate explicit instantiation of 'addCommaSeparators<int>'
   | template string addCommaSeparators( globalIndex const & num );
   |                 ^
${GEOS_DIR}/src/coreComponents/common/format/StringUtilities.cpp:96:17: note: previous explicit instantiation is here
   | template string addCommaSeparators( localIndex const & num );
   |                 ^
```

2. Typo in cmake keyword: `append` -> `list`


src/coreComponents/common/format/StringUtilities.cpp
```diff
@@ -93,7 +93,9 @@ string addCommaSeparators( T const & num )
   return result;
 }
 
+#if GEOS_LOCALINDEX_TYPE_FLAG != GEOS_GLOBALINDEX_TYPE_FLAG
 template string addCommaSeparators( localIndex const & num );
+#endif
 template string addCommaSeparators( globalIndex const & num );
```
rrsettgast: I think this should be better

```suggestion
template string addCommaSeparators( int const & num );
template string addCommaSeparators( long int const & num );
template string addCommaSeparators( long long int const & num );
```

rrsettgast (approved):
@victorapm I approved but please try this change. 

what system is this on? It doesn't show up on CI?

PR #3453, opened 2024-11-18
fix: delta volume fix plus some other stuff

paveltomin:
- when `defaultAperture` is inconsistent with hydraulic aperture table - make hydraulic aperture table override what what specified in `defaultAperture`
- use `poromechanicsFracturesKernels::StateUpdateKernel` in place of `poromechanicsEFEMKernels::StateUpdateKernel` (they are the same), remove the second one as redundant


src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.cpp
```diff
@@ -691,8 +691,7 @@ void SinglePhaseBase::implicitStepSetup( real64 const & GEOS_UNUSED_PARAM( time_
     {
       saveConvergedState( subRegion );
 
-      arrayView1d< real64 > const & dVol = subRegion.template getField< fields::flow::deltaVolume >();
-      dVol.zero();
+      applyDeltaVolume( subRegion );
```
paveltomin, line 671: the main fix is here

src/coreComponents/physicsSolvers/multiphysics/poromechanicsKernels/SinglePhasePoromechanicsEFEM.hpp
```diff
@@ -307,70 +305,6 @@ using SinglePhaseKernelFactory = finiteElement::KernelFactory< SinglePhasePorome
                                                                real64 const (&)[3],
                                                                string const >;
 
-/**
```
paveltomin, line 305: removed since it is basically a copy of the same from DFM

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.hpp
```diff
@@ -387,6 +387,8 @@ class SinglePhaseBase : public FlowSolverBase
    */
   virtual void saveConvergedState( ElementSubRegionBase & subRegion ) const override;
 
+  void applyDeltaVolume( ElementSubRegionBase & subRegion );
```
CusiniM:
```suggestion
  void applyDeltaVolume( ElementSubRegionBase & subRegion ) const;
```

src/coreComponents/physicsSolvers/multiphysics/poromechanicsKernels/SinglePhasePoromechanicsEFEM.hpp
CusiniM: yes they should be the same now that we don't use totalTraction anymore.

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.cpp
paveltomin: the fix is not needed when defaultAperture is consistent with hydraulic aperture table
still it provides some safeguard when it is not consitent and make hydraulic aperture table override what what specified in defaultAperture
@CusiniM, @Guotong-Ren, @rrsettgast please let me know if it makes sense to merge it?

Guotong-Ren: make sense to me. one question, when new element is generated, is it also getting the correct value?

Guotong-Ren (approved):

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.cpp
paveltomin: it should yes

dkachuma (approved):

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.cpp
Guotong-Ren: is there any scenario when only default aperture is used?

CusiniM: it may be what's done for flow only simulations. so I think it may be a little bit of work but we can maybe move it to `hydraulicAperture` Model.

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.cpp
tjb-ltk, line 1: Is the SinglePhaseBase::updateEnergy method  called before implicitStepSetup ?? 
where the vol used is volume+deltaVolume.

paveltomin: called inside updateEnergy after deltaVolume is applied to volume and deltaVolume is assigned to zero
then updateEnergy is called after each Newton

PR #3457, opened 2024-11-21
fix: Add missing flux term assembly in HydrofractureSolver for non-poroelastic case

liyangrock:
When matrix poroelasticity is disabled (m_isMatrixPoroelastic = 0), the HydrofractureSolver was missing the flux term assembly step in its system assembly process. This PR adds the required `flowSolver()->assembleFluxTerms()` call to properly handle fluid flow calculations in the non-poroelastic matrix case.


Guotong-Ren:
the flux term is actually embedded inside assembleHydrofracFluxTerms function call.

CusiniM:
@Guotong-Ren is correct, this is not needed.

liyangrock:
Thank you for pointing that out @Guotong-Ren @CusiniM

PR #3464, opened 2024-11-25
fix: Little bug fix for automatic timestep computation for elastic wave solver

acitrain:
This PR fix littles typos which generates a bug inside the automatic timestep computation routine in the second order elastic wave solver


src/coreComponents/unitTests/wavePropagationTests/testWavePropagationDAS.cpp
```diff
@@ -188,7 +188,7 @@ TEST_F( ElasticWaveEquationSEMTest, SeismoTrace )
 
   //Assert on time-step computed with the automatci time-step routine
   real64 const dtOut = propagator->getReference< real64 >( ElasticWaveEquationSEM::viewKeyStruct::timeStepString() );
-  ASSERT_TRUE( dtOut < 0.04 );
+  ASSERT_TRUE( dtOut < 0.05 );
```
rrsettgast, line 191: should this be a hard coded number? Perhaps you need something scaled by the wave speed and length scale?

rrsettgast (approved):

sframba (approved):

src/coreComponents/unitTests/wavePropagationTests/testWavePropagationDAS.cpp
acitrain: Hi @rrsettgast,

You are right, there are analytical formula for this simple cases, I put it inside the acoustic unit test directly, but for the elastic, It needs a computation of eigenvalues (the one of the elasticity tensor) and I didn't want do do this on an unit test
So I did it outside GEOS and put the value inside the unit test (or at least the closest "simple" value)

rrsettgast: @acitrain @sframba ??

acitrain: Yes @rrsettgast ?

rrsettgast: We posted that the last two messages at the same time. You answered the question.

PR #3465, opened 2024-11-26
fix: format statement for reporting location of max compdens change

tjb-ltk:
Format statement accepted input (value, conv unit, location) but ignored cell location.  Nested format was needed.

### DIFF
--- a/src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.cpp
+++ b/src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.cpp
@@ -588,11 +588,10 @@ real64 CompositionalMultiphaseFVM::scalingForSystemSolution( DomainPartition & d
                                         globalDeltaPresMax.value,
                                         globalDeltaPresMax.location ) );
   GEOS_LOG_LEVEL_INFO_RANK_0( logInfo::Solution,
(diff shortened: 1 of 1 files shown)


src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.cpp
```diff
@@ -588,11 +588,10 @@ real64 CompositionalMultiphaseFVM::scalingForSystemSolution( DomainPartition & d
                                         globalDeltaPresMax.value,
                                         globalDeltaPresMax.location ) );
   GEOS_LOG_LEVEL_INFO_RANK_0( logInfo::Solution,
```
paveltomin, line 591: can we do this instead of nested GEOS_FMT:
```
  GEOS_LOG_LEVEL_INFO_RANK_0( logInfo::Solution,
                              GEOS_FMT( "        {}: Max component density change = {:.3f} {} (before scaling) at cell {}",
                                        getName(),
                                        globalDeltaCompDensMax.value,
                                        massUnit,
                                        globalDeltaCompDensMax.location ) );
```

tjb-ltk: I pushed what you suggested... I need a primer in GEOS_FMT.    ... oh I just realized massUnit is a string... thought it was a conversion factor!!!

paveltomin (approved):

PR #3285, opened 2024-08-15
test: add hf verification cases for leak-off 

Guotong-Ren:
compare to analytical solutions of kgd with leakoff (Bunger et al, 2005). both storage and leakoff toughness dominated regimes  are added.

PR #3466, opened 2024-11-26
fix: an issue of established f-m connections being duplicated in the run time.

Guotong-Ren:
This happens when the initial fracture is created without clear of the two vectors for stencil generation.


paveltomin (approved):

paveltomin:
@CusiniM, @jhuang2601, and/or @rrsettgast please review

src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp
```diff
@@ -176,7 +176,8 @@ real64 HydrofractureSolver< POROMECHANICS_SOLVER >::fullyCoupledSolverStep( real
                                                                             int const cycleNumber,
                                                                             DomainPartition & domain )
 {
-  if( cycleNumber == 0 && time_n <= 0 )
+  // for initial fracture initialization
```
Guotong-Ren: initialNewFractureFields must take care of initial fracture created from execute() in SurfaceGenerator class. Otherwise, the duplication of f-m connections can happen afterwards.    However, when there is a geomechanics initialization step, the previous "IF" will rule out stepping inside due to cycleNumber == 0.

Guotong-Ren: @CusiniM @paveltomin

src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp
```diff
@@ -176,7 +176,8 @@ real64 HydrofractureSolver< POROMECHANICS_SOLVER >::fullyCoupledSolverStep( real
...
+  if( time_n <= 0 )
```
rrsettgast, line 180: @Guotong-Ren Can you explain what the effect of this change is?

rrsettgast (commented):
@Guotong-Ren 
It isn't clear to me what exactly the problem is being addressed, or how the proposed change fixes the problem. Is there an example problem that you can point out to describe the issue and test the "fix"?

src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp
rrsettgast: I think we need a more intricate initialization/initialcondition procedure to handle this cleanly.

Guotong-Ren:
> @Guotong-Ren It isn't clear to me what exactly the problem is being addressed, or how the proposed change fixes the problem. Is there an example problem that you can point out to describe the issue and test the "fix"?

yes, if there is an initialization added before time = 0, the problem will emerge. For reference, you can take a look at the cases in this PR (https://github.com/GEOS-DEV/GEOS/pull/3285). When leak-off is dominant, fracture does not propagate and then mf connection will duplicate. @rrsettgast

paveltomin:
@rrsettgast @CusiniM are you ok to merge this?

rrsettgast:
Not sure this is a solid fix for the issue

paveltomin:
> Not sure this is a solid fix for the issue

Can the if statement be removed completely and just call that initialization unconditionally?

Guotong-Ren:
> > Not sure this is a solid fix for the issue
> 
> Can the if statement be removed completely and just call that initialization unconditionally?

since the field initialization only takes effects when there are new fracture cells, i believe so.

CusiniM:
> > > Not sure this is a solid fix for the issue
> > 
> > 
> > Can the if statement be removed completely and just call that initialization unconditionally?
> 
> since the field initialization only takes effects when there are new fracture cells, i believe so.

Yeah, this should work. The lists should be empty so I would expect that function to do nothing.

CusiniM (approved):

paveltomin:
> > > > Not sure this is a solid fix for the issue
> > > 
> > > 
> > > Can the if statement be removed completely and just call that initialization unconditionally?
> > 
> > 
> > since the field initialization only takes effects when there are new fracture cells, i believe so.
> 
> Yeah, this should work. The lists should be empty so I would expect that function to do nothing.

thanks @CusiniM
@rrsettgast ?

paveltomin:
@rrsettgast is it good for now or we still should be looking for a more fundamental fix?

rrsettgast (approved):
yeah. fine

PR #3467, opened 2024-11-27
fix: make new gravity treatment from #3337 an option

paveltomin:
disabled by default since SPE11 case degradations were observed


victorapm (approved):
This PR matches with the reference results for spe11b. Thanks!

spe11b-times

paveltomin:
@CusiniM and @ryar9534 let's please move this fix quickly into develop, critical for spe11

CusiniM (approved):

CusiniM:
I don' t know that we really need 2 options for the treatment of gravity. In general, I think that what @paveltomin had introduced made more sense... I am not too sure why it would have such a large impact on convergence.

rrsettgast:
@paveltomin The term `newGravity` is quite uninformative. Can you create descriptive options for the "gravity treatment" so that developers/users know what the treatments are?

paveltomin:
> @paveltomin The term `newGravity` is quite uninformative. Can you create descriptive options for the "gravity treatment" so that developers/users know what the treatments are?

Ok will do

paveltomin:
> @paveltomin The term `newGravity` is quite uninformative. Can you create descriptive options for the "gravity treatment" so that developers/users know what the treatments are?

is this https://github.com/GEOS-DEV/GEOS/pull/3486 better?

PR #3471, opened 2024-12-02
feat: add way for field specification to scale input data

ryar9534:
This pr adds a way for fieldSpecifications in the xml to multiplicatively scale material data, rather than only set the value. The use case would be when someone wants to run multiple realizations of a problem, which has been requested by Total users. The current implementation works, but is certainly up for discussion.


src/coreComponents/fieldSpecification/FieldSpecificationBase.cpp
```diff
@@ -83,6 +83,11 @@ FieldSpecificationBase::FieldSpecificationBase( string const & name, Group * par
     setInputFlag( InputFlags::OPTIONAL ).
     setDescription( "Time at which the boundary condition will stop being applied." );
 
+  registerWrapper( viewKeyStruct::isScalingString(), &m_isScaling ).
+    setApplyDefaultValue( 0 ).
+    setInputFlag( InputFlags::OPTIONAL ).
+    setDescription( "Boundary condition is a multiplicative scaling of the values already present." );
+
```
rrsettgast, line 90: could we specify an "operation type" here? So instead of `m_isScaling` specify something like `m_operation`. Then we could have some enum operations. Things that come to mind are equals, addition, scaling....also...it just occurred to me that this stuff could be achieved using a function? I would think this is the better place to embed operations on fields?

rrsettgast (commented):
philosophical question about whether we want to embed the operations in `FieldSpecification` or rely on a function.

src/coreComponents/fieldSpecification/FieldSpecificationBase.cpp
ryar9534: yeah, when @CusiniM and I were talking we saw that there were a few ways to do it. This just made sense to me as a starting point, so we could all hopefully look at something that works before making a final decision of how to do it. (Of course, I need to make it work first)

rrsettgast: Yes, I do recall the discussion in Pau on this. I think we were a little bit premature about the approach of just adding a BC multiply option. Using a function to define the operation will be quite nice as we could then use a table function and apply a table based variable scalar multiplier. We could also define a random function for generating a varying scalar that follows some distribution....of course....this might be more than we want to do.

ryar9534: I agree that the table function way is nice, and thats what  I recommended. I just dont know how users usually paint properties onto the mesh, and if that is easy to turn into a table. I figured I would give this a shot if it was easy.

rrsettgast: We can use the "functions" and add a "ScalarFunction" so the users don't have to be bothered by specifying a table when they don't need that. What do you think? @CusiniM what do you think?

paveltomin:
@herve-gross ping

herve-gross:
@dkachuma can you take a look at this?

dkachuma:
The `FieldSpecification` element already has a `functionName` field. I noticed that this is applied to the entire group. So I think we can already achieve this in certain cases by creating a function with the property that we desire as the input variable.

Some example xml. Here `porosityScaleFunction` is a function that multiplies the reference porosity (`rockPorosity_referencePorosity`) by 1/2. This is assigned to back to `rockPorosity_referencePorosity` in `FieldSpecification` `poro`. Similary we can define the horizontal permeability using a porosity-permeability correlation `permeabilityXCorrelationFunction` or even a symbolic function `permeabilityZCorrelationFunction`.

```xml

  <Functions>
    <!-- Multiply by 1/2 -->
    <TableFunction
      name="porosityScaleFunction"
      inputVarNames="{ rockPorosity_referencePorosity }"
      coordinates="{ 0.0, 1.0 }"
      values="{ 0.0, 0.5 }"
      interpolation="linear" />
    <!-- Discretised form of log(K) = a*\phi + b in milli-Darcy with [a, b] = [9.94, 3.42] -->
    <TableFunction
      name="permeabilityXCorrelationFunction"
      inputVarNames="{ rockPorosity_referencePorosity }"
      coordinates="{ 0.05, 0.10, 0.15, 0.20, 0.25, 0.30, 0.35 }"
      values="{ 50.25, 82.60, 135.78, 223.18, 366.87, 603.05, 991.28 }"
      interpolation="linear" />
    <!-- Symbolic function (tested only on x86_64 CPU) -->
    <SymbolicFunction
      name="permeabilityZCorrelationFunction"
      inputVarNames="{ rockPorosity_referencePorosity }"
      variableNames="{ phi }"
      expression="0.1*exp(9.94*phi+3.42)" />
  </Functions>

  <FieldSpecifications>
    <FieldSpecification
      name="poro"
      initialCondition="1"
      setNames="{ all }"
      objectPath="ElementRegions"
      fieldName="rockPorosity_referencePorosity"
      functionName="porosityScaleFunction"
      scale="1.0" />      

    <FieldSpecification
      name="permx"
      initialCondition="1"
      setNames="{ all }"
      objectPath="ElementRegions"
      functionName="permeabilityXCorrelationFunction"
      component="0"
      scale="9.869233e-16" />   <!-- Convert from mD to m2 -->   
      
    <FieldSpecification
      name="permz"
      initialCondition="1"
      setNames="{ all }"
      objectPath="ElementRegions"
      functionName="permeabilityZCorrelationFunction"
      component="2"
      scale="9.869233e-16" />   <!-- Convert from mD to m2 -->   
  </FieldSpecifications>
```

Complete test case [testCase.zip](https://github.com/user-attachments/files/21585055/testCase.zip)

I think this would cover most of the user cases. Maybe the syntax is a bit clunky. Not sure how to improve that. Maybe we can put the `inputVarNames` field on `FieldSpecification` as short-hand for applying the `scale`.

@rrsettgast @herve-gross What do you think?

PR #3474, opened 2024-12-04
fix: update `m_weights` indexing in SurfaceElementStencil.hpp

liyangrock:
The PR fixes incorrect array indexing when accessing `m_weights` for intersected fractures.


rrsettgast:
@liyangrock Do you have a small example problem you can add to the integrated tests that exercising intersecting fractures?

Guotong-Ren:
if it is TPFA, k array just equals 0, 1 which is identical to the original formulation.  but i think it is good to be consistent

rrsettgast:
@Guotong-Ren Even in TPFA, if there are 3 fracture volumes connected to a "connector" the for loop:
```c++
  for( k[0]=0; k[0]<numPointsInFlux( iconn ); ++k[0] )
  {
    for( k[1]=k[0]+1; k[1]<numPointsInFlux( iconn ); ++k[1] )
    {
```
results in the permutations of `k[2]` of `(0,1), (0,2), (1,2)`. Do I have this incorrect?

Guotong-Ren:
> @Guotong-Ren Even in TPFA, if there are 3 fracture volumes connected to a "connector" the for loop:
> 
> ```c++
>   for( k[0]=0; k[0]<numPointsInFlux( iconn ); ++k[0] )
>   {
>     for( k[1]=k[0]+1; k[1]<numPointsInFlux( iconn ); ++k[1] )
>     {
> ```
> 
> results in the permutations of `k[2]` of `(0,1), (0,2), (1,2)`. Do I have this incorrect?

numPointsInFlux(iconn) = 2 for TPFA. (0,2) and (1,2) should be ruled out in the loop ending criterion?

CusiniM:
> > @Guotong-Ren Even in TPFA, if there are 3 fracture volumes connected to a "connector" the for loop:
> > ```c++
> >   for( k[0]=0; k[0]<numPointsInFlux( iconn ); ++k[0] )
> >   {
> >     for( k[1]=k[0]+1; k[1]<numPointsInFlux( iconn ); ++k[1] )
> >     {
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > results in the permutations of `k[2]` of `(0,1), (0,2), (1,2)`. Do I have this incorrect?
> 
> numPointsInFlux(iconn) = 2 for TPFA. (0,2) and (1,2) should be ruled out in the loop ending criterion?

This is not the case for fractures. Each connection contains the `halfWeight` of each fracture element connected to that connection. In the case of two intersecting fractures, with 4 faceElements connected through a single edge, `m_weights` will have size 4 and then, on the fly, we compute transmissibilities (weights) pair by pair. So, I think this fix is actually needed.

CusiniM:
> @liyangrock Do you have a small example problem you can add to the integrated tests that exercising intersecting fractures?

I think we have one in there. Note that if all elements have same geometry differences won't be noticeable.

rrsettgast:
> > @liyangrock Do you have a small example problem you can add to the integrated tests that exercising intersecting fractures?
> 
> I think we have one in there. Note that if all elements have same geometry differences won't be noticeable.

We should modify it s.t. incorrect calculations will be obvious....this is an easy statement to make...perhaps harder to implement.

liyangrock:
> @liyangrock Do you have a small example problem you can add to the integrated tests that exercising intersecting fractures?

Currently, I don’t have a small example problem related to intersecting fractures that can be added to the integrated tests. I’m not familiar with the integrated tests yet, but I’m open to exploring them.

paveltomin (approved):

paveltomin:
copied to https://github.com/GEOS-DEV/GEOS/pull/3768 to avoid fork issue

PR #3477, opened 2024-12-05
fix: fix issue #2615

### LINKED ISSUE #2615: unintialized error when building NeighborCommunicator.cpp
**Describe the bug**
An uninitialized error when building GEOS.  Perhaps when building NeighborCommunicator.cpp.  See screenshot

**To Reproduce**
Build TPLs and then GEOS on latest ubuntu with gcc 11.4.0 and cmake 3.24.1

**Screenshots**
image

- Ubuntu 22.04.2 LTS
- Compiler: gcc 11.4.0
- GEOSX Version - latest version... cloned 8/3/23

**Additional context**
Built it in the same manner which was successful for us on 4/5/23

paveltomin:
should close https://github.com/GEOS-DEV/GEOS/issues/2615


rrsettgast:
@paveltomin 
the underlying call is this:
```
  void getLocation( string const & key,
                    FieldLocation & location ) const
  {
    if( key.find( m_locationKeys.nodesKey() ) != string::npos )
    {
      location = FieldLocation::Node;
    }
    else if( key.find( m_locationKeys.edgesKey()) != string::npos )
    {
      location = FieldLocation::Edge;
    }
    else if( key.find( m_locationKeys.facesKey()) != string::npos )
    {
      location = FieldLocation::Face;
    }
    else if( key.find( m_locationKeys.elemsKey()) != string::npos )
    {
      location = FieldLocation::Elem;
    }
    else
    {
      GEOS_ERROR( GEOS_FMT( "Invalid key, {}, was provided. Location cannot be retrieved.", key ) );
    }
  }
```
which is properly guarded. The compiler is missing it for the reporting user. We can change function to return the value of `location` and set it explicitly in the call location. That would make more sense.

paveltomin:
> @paveltomin the underlying call is this:
> 
> ```
>   void getLocation( string const & key,
>                     FieldLocation & location ) const
>   {
>     if( key.find( m_locationKeys.nodesKey() ) != string::npos )
>     {
>       location = FieldLocation::Node;
>     }
>     else if( key.find( m_locationKeys.edgesKey()) != string::npos )
>     {
>       location = FieldLocation::Edge;
>     }
>     else if( key.find( m_locationKeys.facesKey()) != string::npos )
>     {
>       location = FieldLocation::Face;
>     }
>     else if( key.find( m_locationKeys.elemsKey()) != string::npos )
>     {
>       location = FieldLocation::Elem;
>     }
>     else
>     {
>       GEOS_ERROR( GEOS_FMT( "Invalid key, {}, was provided. Location cannot be retrieved.", key ) );
>     }
>   }
> ```
> 
> which is properly guarded. The compiler is missing it for the reporting user. We can change function to return the value of `location` and set it explicitly in the call location. That would make more sense.

please check the updated version

paveltomin:
@rrsettgast is it good now?

rrsettgast (approved):

PR #3483, opened 2024-12-06
refactor: remove rel perm from wells

paveltomin:
not used anyway


paveltomin:
@tjb-ltk @dkachuma are you ok with the change?

dkachuma (commented):
This is not used but the code was supposedly forward looking. We might need this in the well later. I'll let @tjb-ltk confirm.

tjb-ltk:
Forward looking for sure... access to the relperm and pvt models would be needed for something like pseudo pressure modeling for condensate dropout, an alternative to fine grids around wells..  but even in this workflow a well estimator is needed, which then requires further development of the parallel framework.  I would confer with the GEOS executive committee about their short/med/long term strategic objectives and plans.

tjb-ltk (approved):
Until a use case comes along its probably better to remove these fields

paveltomin:
@CusiniM, @castelletto1, @jhuang2601 please review

CusiniM (approved):

PR #3460, opened 2024-11-23
refactor: Single phase property container

tjb-ltk:
Refactor single phase constitutive containers to use same underpinnings for storage and access of property and derivative quantities .     Ultimately this will allow single phase thermal well development to have the same look and feel as compositional version.   
TODO in another PR sort out compositional aspects of proppant formulation

PR #3485, opened 2024-12-08
fix: use mass and energy consistently for single phase solvers

paveltomin:
should go after https://github.com/GEOS-DEV/GEOS/pull/3460

unified where possible
for poromechanics - hard to do because porosity is being updated inside the assembly kernel, so left it as is
fixed glitch with double porosity update in poromechanics when it was updated both in flow solver and in poromechanics kernel using different equations leading to potential inconsistencies


src/coreComponents/physicsSolvers/fluidFlow/FlowSolverBaseFields.hpp
```diff
@@ -241,46 +241,38 @@ DECLARE_FIELD( temperatureScalingFactor,
                NO_WRITE,
                "Scaling factors for temperature" );
 
-DECLARE_FIELD( mass,
```
paveltomin, line 244: `mass` went to single phase file since it is used there, not for compositional

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.cpp
```diff
@@ -594,8 +634,6 @@ void SinglePhaseBase::initializeFluidState( MeshLevel & mesh, arrayView1d< strin
 
     // 2. save the initial density (for use in the single-phase poromechanics solver to compute the deltaBodyForce)
     fluid.initializeState();
-
-    updateMass( subRegion );
```
paveltomin, line 567: this call is redundant, `updateMass` already called in updateFluidState few lines above

src/coreComponents/physicsSolvers/fluidFlow/kernels/singlePhase/AccumulationKernels.hpp
```diff
@@ -291,13 +260,11 @@ class SurfaceElementAccumulationKernel : public AccumulationKernel< SurfaceEleme
   void computeAccumulation( localIndex const ei,
                             Base::StackVariables & stack ) const
   {
-    Base::computeAccumulation( ei, stack, [&] ()
+    Base::computeAccumulation( ei, stack );
```
paveltomin, line 270: keep it simple

src/coreComponents/physicsSolvers/fluidFlow/kernels/singlePhase/ThermalAccumulationKernels.hpp
```diff
@@ -68,80 +62,22 @@ class AccumulationKernel : public singlePhaseBaseKernels::AccumulationKernel< SU
...
-    real64 const solidVolume = ( m_volume[ei] + m_deltaVolume[ei] ) * ( 1.0 - m_porosity[ei][0] );
-    real64 const dSolidVolume_dPres = -( m_volume[ei] + m_deltaVolume[ei] ) * m_dPoro_dPres[ei][0];
-    real64 const dSolidVolume_dTemp = -( m_volume[ei] + m_deltaVolume[ei] ) * m_dPoro_dTemp[ei][0];
-
-    // initialize the solid internal energy
-    stack.solidEnergy = solidVolume * m_rockInternalEnergy[ei][0];
-    stack.dSolidEnergy_dPres = dSolidVolume_dPres * m_rockInternalEnergy[ei][0];
-    stack.dSolidEnergy_dTemp = solidVolume * m_dRockInternalEnergy_dTemp[ei][0] + dSolidVolume_dTemp * m_rockInternalEnergy[ei][0];
-  }
+  {};
```
paveltomin, line 82: i could not remove this, somehow compiler gets always confused

CusiniM:
you may want to keep an eye on this: https://github.com/GEOS-DEV/GEOS/pull/3460

paveltomin:
> you may want to keep an eye on this: #3460

Yes thanks

CusiniM (approved):
LGTM

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.cpp
```diff
@@ -257,53 +263,93 @@ void SinglePhaseBase::updateMass( ElementSubRegionBase & subRegion ) const
 {
   GEOS_MARK_FUNCTION;
 
+  using DerivOffset = constitutive::singlefluid::DerivativeOffsetC< 1 >; // TODO check if this is correct
```
tjb-ltk: The assumption with this is that someone doesn't make pressure second index in isothermal run... to be safe I would use
    geos::internal::kernelLaunchSelectorThermalSwitch( isThermal, [&] ( auto ISTHERMAL ) {
      integer constexpr IS_THERMAL = ISTHERMAL(); 
     using DerivOffset = constitutive::singlefluid::DerivativeOffsetC< IS_THERMAL >;
     if constexpr(IS_THERMAL)   // for the thermal bits
    {
   }
    }

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.cpp
```diff
@@ -257,53 +263,93 @@ void SinglePhaseBase::updateMass( ElementSubRegionBase & subRegion ) const
...
+    } );
+  }
 }
 
 void SinglePhaseBase::updateEnergy( ElementSubRegionBase & subRegion ) const
 {
   GEOS_MARK_FUNCTION;
 
+  using DerivOffset = constitutive::singlefluid::DerivativeOffsetC< 1 >;
+
```
tjb-ltk, line 326: by definition this is thermal routine so this is fine

src/coreComponents/physicsSolvers/fluidFlow/kernels/singlePhase/AccumulationKernels.hpp
tjb-ltk: nice... introducing  mass cleaned up accum and flux kernels a fair bit and probably reduced computation at the same time ...    same for introducing energy variable..

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.cpp
paveltomin: added, thanks!

PR #3491, opened 2024-12-13
fix: fix remaining copyright errors

rrsettgast:
There are a couple of copyright errors that persisted after the last PR to fix them. This PR seeks to fix the remaining errors.


src/coreComponents/physicsSolvers/inducedSeismicity/QuasiDynamicEQRK32.cpp
CusiniM, line 1: This solver is renamed in https://github.com/GEOS-DEV/GEOS/pull/3416 which I ll try to merge soon and where I have fixed the copyright so it's probably easier to leave it as is for now.

CusiniM (approved):

untereiner:
Oh I forget to mention that with spdx spec it is easier to match lines and enforce licence / copyright in the CI

PR #3495, opened 2024-12-17
fix: Feature/byer3/mass inj constraint fix

tjb-ltk:
Add missing logic to support switching from fixed mass rate injection rate constraint to max injection pressure.  The code will crash if running a model where this condition is encountered.   No issues with mass injection constraint in scenarios where pressure associated with injection rate is less than maximum injection pressure.   2 test cases added to demonstrate constraint switching using mass injection rate constraint and total injection rate tables.  The cases should produce nearly identical results.

### DIFF
--- a/src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
+++ b/src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
@@ -454,6 +454,7 @@ void CompositionalMultiphaseWell::validateWellConstraints( real64 const & time_n
   WellControls::Control const currentControl = wellControls.getControl();
   real64 const & targetTotalRate = wellControls.getTargetTotalRate( time_n + dt );
   real64 const & targetPhaseRate = wellControls.getTargetPhaseRate( time_n + dt );
+  real64 const & targetMassRate = wellControls.getTargetMassRate( time_n + dt );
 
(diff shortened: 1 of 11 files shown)


src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
```diff
@@ -470,6 +473,18 @@ void CompositionalMultiphaseWell::validateWellConstraints( real64 const & time_n
...
+                 "WellControls " << wellControls.getDataContext() <<
+                 ": Target total rate cannot be used for producers",
+                 InputError );
+  GEOS_THROW_IF( wellControls.isProducer() && !isZero( targetMassRate ),
+                 "WellControls " << wellControls.getDataContext() <<
+                 ": Target mass rate cannot be used for producers",
+                 InputError );
+  GEOS_THROW_IF( !m_useMass && !isZero( targetMassRate ),
+                 "WellControls " << wellControls.getDataContext() <<
+                 ": Target mass rate cannot with m_useMass=0",
```
paveltomin:
```suggestion
                 ": Target mass rate cannot with useMass=0",
```

paveltomin (approved):

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
tjb-ltk: changed

CusiniM (approved):

PR #3498, opened 2024-12-19
feat: Advance HIP support

victorapm:
- New tioga and tuo host configs using `cce-18` and `rocm-6.2.1`
- Compilation warning fixes
- Remove limitation on ParMETIS
- A few workarounds to get the compilation done

https://github.com/GEOS-DEV/thirdPartyLibs/pull/289
https://github.com/GEOS-DEV/LvArray/pull/340


src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
```diff
@@ -760,6 +760,29 @@ real64 PhysicsSolverBase::eisenstatWalker( real64 const newNewtonNorm,
   return krylovTol;
 }
 
+
+void MemoryUsageOutput( std::string const & message )
```
CusiniM, line 803: was this for debugging? Or do you need it? In any case I don't think it should be in `PhysicsSolverBase`

CusiniM:
We should add an image with rocm... something like https://hub.docker.com/r/rocm/dev-ubuntu-24.04

src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
victorapm: This is for debugging, I'll update you guys soon

victorapm:
Replaced by #3831

PR #3500, opened 2024-12-19
fix: issue 3499

### LINKED ISSUE #3499: Nonsensical low permeability check in JFunctionCapillaryPressure.cpp
**Describe the bug**
Line 253 in JFunctionCapillaryPressure.cpp 

GEOS_ERROR_IF( permeability < LvArray::NumericLimits< real64 >::epsilon, "Zero permeability in J-function capillary pressure" );

checks whether permeability is below 64-bit machine precision (~1e-16), but there can be valid permeability values in SI units that may be lower than that. E.g., 0.1 mDarcy = 9.86923e-17 m^2 in Si units. Also, the permeability is not used directly in this function. Instead, sqrt(permeability) is used in the JFunction multiplier. So there is no overflow and this comparison is not necessary.

**To Reproduce**
1. Run a case where some permeability values are of order 0.1 mDarcy = 9.86923e-17 m^2. 
2. This will trigger an error "Zero permeability in J-function capillary pressure". 

**Expected behavior**
Permeabilities like those on the order of 0.1 mDarcy are valid and should not trigger an error.

- Machine: Ubuntu 24.04.1 LTS
- Compiler: gcc 14.2.0
- GEOS Version: 1.2.0

paveltomin:
should fix https://github.com/GEOS-DEV/GEOS/issues/3499


src/coreComponents/constitutive/capillaryPressure/JFunctionCapillaryPressure.cpp
```diff
@@ -250,7 +250,8 @@ void JFunctionCapillaryPressure::saveConvergedRockState( arrayView2d< real64 con
     {
       permeability = convergedPermeability[ei][0][2];
     }
-    GEOS_ERROR_IF( permeability < LvArray::NumericLimits< real64 >::epsilon, "Zero permeability in J-function capillary pressure" );
+    // 9.869233×10−13 is Darcy to sq m factor
+    GEOS_ERROR_IF( permeability < LvArray::NumericLimits< real64 >::epsilon * 9.869233e-13, "Zero permeability in J-function capillary pressure" );
```
CusiniM: why is this error check even necessary? A zero permeability is a valid number and all underflows are flushed to zero.

paveltomin: there is a division down here https://github.com/GEOS-DEV/GEOS/blob/6ebf57643a6fc6136f9687cfbc902bc662fc39a2/src/coreComponents/constitutive/capillaryPressure/JFunctionCapillaryPressure.cpp#L267

paveltomin: @CusiniM ?

dkachuma: Can we put this constant in `Constants.hpp` or `Units.hpp`.

dkachuma (approved):

CusiniM (approved):

src/coreComponents/common/Units.hpp
```diff
@@ -30,6 +30,8 @@ namespace geos
 namespace units
 {
 
+/// Darcy to m^2 conversion factor
+static constexpr double DarcyToSqM = 9.869233e-13;
```
MelReyCG, line 34: The PR looks good to me. Just a small thing, you could move this constant into `PhysicsConstant.hpp` if you think it fits better.

MelReyCG (approved):

src/coreComponents/common/Units.hpp
paveltomin: > The PR looks good to me. Just a small thing, you could move this constant into `PhysicsConstant.hpp` if you think it fits better.

thank
will keep it in units

PR #3486, opened 2024-12-09
refactor: clarify new gravity, clean up kernel flags usage, move CFL computations to FVM solver

paveltomin:
- clarify new gravity as requested here https://github.com/GEOS-DEV/GEOS/pull/3467#issuecomment-2525454349
- clean up kernel flags usage in flow solver to avoid bugs: the basic idea is to set the flags outside of kernels create and launch to avoid mismatch between different version (isothermal/thermal etc) when one flag is set here but is missing there
- move CFL computations to FVM solver where it belongs

PR #3501, opened 2024-12-19
refactor: clean up kernel flags usage in flow solver to avoid bugs

paveltomin:
the basic idea is to set the flags outside of kernels create and launch to avoid mismatch between different version (isothermal/thermal etc) when one flag is set here but is missing there


paveltomin:
merged into https://github.com/GEOS-DEV/GEOS/pull/3486 to avoid dealing with conflicts

PR #3503, opened 2024-12-23
ci: remove code owners from .integrated_tests.yaml

paveltomin:
to avoid new review requirements popping up when baselines are updated and PR going from approved state into review required


CusiniM (approved):
okay, forget my comment in the other PR then. I am fine with this.

rrsettgast (approved):

jhuang2601 (approved):

PR #3508, opened 2025-01-06
refactor: FieldSpecification context when targeting wrong fieldName

### LINKED ISSUE #2320: [EPIC] Improve feedback to user on XML errors

arng40:
Here's the typical error encountered when setting the wrong target for a FieldSpecification :

```
***** LOCATION: /path/to/GEOS/src/coreComponents/fieldSpecification/FieldSpecificationManager.cpp:208
***** Rank 17:
initialPressure_channel/fieldName (staircase_3d.xml, l.133): there is no fieldName named `pressure_t` under the objectPath `ElementRegions/Channel`.
 
 
** StackTrace of 5 frames **
[...]
=====
```
Here we wanted to define initial conditions in FieldSpecifications to impose the value of a property (here: “rock_initialPorosity”).
However, this property doesn't exist because we've given it the wrong name. It expects the name of the Porosity model for stress (i.e. “rockPorosity_initialPorosity” here). 

With more indications we would have (in progress)
```
***** LOCATION: /path/to/GEOS/src/coreComponents/fieldSpecification/FieldSpecificationManager.cpp:228
***** Rank 0: 
initialPressure_channel/fieldName (staircase_3d.xml, l.133): there is no fieldName named `pressure_t` under the objectPath `ElementRegions/Channel`.
Available fields in ElementRegions/Channel are:
bcPressure, bcTemperature, dMass, dMobility, deltaPressure, deltaVolume, gravityCoefficient, initialPressure, initialTemperature, mass, mass_n, mobility, netToGross, pressure, pressure_n, temperature, temperature_n

** StackTrace of 5 frames **
[...]
```


src/coreComponents/fieldSpecification/FieldSpecificationManager.cpp
```diff
@@ -193,20 +193,40 @@ void FieldSpecificationManager::validateBoundaryConditions( MeshLevel & mesh ) c
...
-        GEOS_FMT( fieldNameNotFoundMessage,
+      std::ostringstream fieldNameNotFoundMessage;
+      std::string fieldNamePath =
+        GEOS_FMT( "\n{}: there is no {} named `{}` under the {} `{}`.\n",
                   fs.getWrapperDataContext( FieldSpecificationBase::viewKeyStruct::fieldNameString() ),
                   FieldSpecificationBase::viewKeyStruct::fieldNameString(),
-                  fs.getFieldName(), FieldSpecificationBase::viewKeyStruct::objectPathString(), fs.getObjectPath() );
+                  fs.getFieldName(), FieldSpecificationBase::viewKeyStruct::objectPathString(), fs.getObjectPath());
+
+      fieldNameNotFoundMessage << fieldNamePath;
```
MelReyCG, line 204:
```c++
      fieldNameNotFoundMessage <<
        GEOS_FMT( "\n{}: there is no {} named `{}` under the {} `{}`.\n",
                  fs.getWrapperDataContext( FieldSpecificationBase::viewKeyStruct::fieldNameString() ),
                  FieldSpecificationBase::viewKeyStruct::fieldNameString(),
                  fs.getFieldName(), FieldSpecificationBase::viewKeyStruct::objectPathString(), fs.getObjectPath());
```

src/coreComponents/fieldSpecification/FieldSpecificationManager.cpp
```diff
@@ -193,20 +193,40 @@ void FieldSpecificationManager::validateBoundaryConditions( MeshLevel & mesh ) c
...
+          if( stopIteration )
+            return;
+          auto const & subCellRegion = elemRegion.getSubRegion< CellElementSubRegion >( 0 );
+          if( !subCellRegion.hasWrapper( fs.getFieldName() ) )
+          {
+            stopIteration = true;
+            for( auto const & wrapperIter : subCellRegion.wrappers() )
+            {
+              auto const & wrapper = wrapperIter.second;
+              fieldNameNotFoundMessage << wrapper->getName() <<  " ";
```
MelReyCG: I would add a comma here 
```suggestion
              fieldNameNotFoundMessage << wrapper->getName() <<  ", ";
```

src/coreComponents/fieldSpecification/FieldSpecificationManager.cpp
```diff
@@ -193,20 +193,40 @@ void FieldSpecificationManager::validateBoundaryConditions( MeshLevel & mesh ) c
...
       else
       {
-        GEOS_THROW( errorMsg, InputError );
+        fieldNameNotFoundMessage << GEOS_FMT( "Available fields in {} are:\n", fs.getObjectPath());
+        bool stopIteration = false;
+        ElementRegionManager const & elemRegionGroup = mesh.getElemManager();
+        elemRegionGroup.forElementRegions( [&]( ElementRegionBase const & elemRegion )
+        {
+          if( stopIteration )
+            return;
```
MelReyCG: Are you sure that, by exiting the iteration process, you gather all possible field name?
I'm not sure that all region contains the same set of fields.

Maybe it would be safer to gather them all in a `std::set<string>`, then output them all (it would also have the benefit to sort them).

src/coreComponents/fieldSpecification/FieldSpecificationManager.cpp
MelReyCG: Here you are assuming that all sub-region wrappers are wrappers of fields, but they are not all! `localToGlobalMap globalToLocalMap isExternal ghostRank`, those are attributes not related to mesh fields.
Here is an exemple of fields declaration: `src/coreComponents/physicsSolvers/fluidFlow/FlowSolverBaseFields.hpp`
The goal is to output only fields in this error output.

MelReyCG (changes requested):

src/coreComponents/fieldSpecification/FieldSpecificationManager.cpp
```diff
@@ -194,20 +194,62 @@ void FieldSpecificationManager::validateBoundaryConditions( MeshLevel & mesh ) c
...
+                                              [&]( FieldSpecificationBase const &,
+                                                   string const & setName,
+                                                   SortedArrayView< localIndex const > const & targetSet,
+                                                   Group & targetGroup,
+                                                   string const fieldName )
+          {
+            GEOS_UNUSED_VAR( setName );
+            GEOS_UNUSED_VAR( isTargetSetCreated );
+            GEOS_UNUSED_VAR( targetSet );
+
```
MelReyCG: You can shorter this by using unnamed parameters, which mark these parameters as _unused_.
```suggestion
          fs2.apply< dataRepository::Group >( mesh,
                                              [&]( FieldSpecificationBase const &,
                                                   string const &,
                                                   SortedArrayView< localIndex const > const &,
                                                   Group & targetGroup,
                                                   string const fieldName )
          {
```

Also, `isTargetSetCreated` seems to be used actually, isn't it?

src/coreComponents/fieldSpecification/FieldSpecificationManager.cpp
```diff
@@ -194,20 +194,62 @@ void FieldSpecificationManager::validateBoundaryConditions( MeshLevel & mesh ) c
...
+              WrapperBase const & targetField = targetGroup.getWrapperBase( fieldName );
+              string const solverName = *(targetField.getRegisteringObjects().begin());
+
+              for( auto const & view : targetGroup.wrappers() )
+              {
+                if( *(view.second->getRegisteringObjects().begin()) == solverName )
+                {
+                  fieldNameAvail.insert( view.second->getName()  );
+                }
+              }
```
MelReyCG: I can confirm that the last method of extracting a list of fields works, even if I find it a little weak:
1. I don't think there's a clear distinction between what is a field and what isn't, with `getRegisteringObjects().contains(solverName)`, you assume that solvers only add (and **will ever add**) field Wrappers to the mesh, and this assertion is based on principles I find too unconstrained,
2. You assume that Wrappers that have been registered by another object are necessarily registered by solvers, without type checking with something like `std::is_convertible_v<T,PhysicsSolverBase>()` (GEOS has RTTI).

In absolute terms, I'd rather see a list of fields registered in `ObjectManagerBase` (or deeper), and a check that these fields have indeed been added by a solver (as mentioned in point 2.).

The list could be filled in during `ObjectManagerBase::registerField()`, and we could have a `std::set<string> const & ObjectManagerBase::getFieldNames()`.

MelReyCG (changes requested):

src/coreComponents/fieldSpecification/FieldSpecificationManager.cpp
arng40: I have added a set of registred fields in objectManager and check this set, I cannot verify the point 2 because fieldSpecification doesn't have any dependencies with physicsSolver.

src/coreComponents/fieldSpecification/FieldSpecificationManager.cpp
arng40: yep removed

src/coreComponents/fieldSpecification/FieldSpecificationManager.cpp
```diff
@@ -194,20 +194,63 @@ void FieldSpecificationManager::validateBoundaryConditions( MeshLevel & mesh ) c
...
+                WrapperBase const & targetField = targetObject.getWrapperBase( fieldName );
+                string const solverName = *(targetField.getRegisteringObjects().begin());
+
+                for( auto const & view : targetObject.wrappers() )
+                {
+                  if( *(view.second->getRegisteringObjects().begin()) == solverName )
+                  {
+                    fieldNameAvail.insert( view.second->getName()  );
+                  }
+                }
```
MelReyCG: This loop could completely be flatten.
Also, does the call to getRegisteringObjects() changes the result of the list?

arng40: loop can be flatten, afterwards this is a bit confusing  😅

MelReyCG (approved):
Nice!

MelReyCG (approved):

CusiniM (approved):

rrsettgast (approved):

PR #3512, opened 2025-01-13
Branch with Chevron's host-configs/CVX

drmichaeltcvx:
This "miket/GPU-build-fix" branch was locally created to investigate the "class size mismatch" issues CUDA encounters when building on Chevron's Azure environment.

It also contains host-config .cmake file Chevron is using to build CPU and GPU GEOS in Azure.


drmichaeltcvx:
This branch currently only contains Chevron's host-configs and at this stage I am only interested in having this branch available at the `origin`.

paveltomin:
more recent version https://github.com/GEOS-DEV/GEOS/pull/3738

PR #3514, opened 2025-01-14
refactor: unify flow solver discretization name check

paveltomin:


src/coreComponents/physicsSolvers/fluidFlow/FlowSolverBase.cpp
```diff
@@ -368,6 +368,18 @@ void FlowSolverBase::initializePreSubGroups()
   }
 }
 
+void FlowSolverBase::checkDiscretizationName()
+{
+  DomainPartition & domain = this->getGroupByPath< DomainPartition >( "/Problem/domain" );
```
CusiniM: it would be nice to avoid this IMO but it was not your code.

src/coreComponents/physicsSolvers/fluidFlow/FlowSolverBase.hpp
```diff
@@ -230,6 +230,8 @@ class FlowSolverBase : public PhysicsSolverBase
 
   virtual void initializePreSubGroups() override;
 
+  void checkDiscretizationName();
```
CusiniM:
```suggestion
  void checkDiscretizationName() const;
```

CusiniM (approved):

PR #3517, opened 2025-01-17
refactor: fields and constitutives

paveltomin:
clean up fields and constitutive "registration"


paveltomin:
rebaseline needed:
```
Error: /Problem/domain/MeshBodies/mesh/meshLevels/Level0/ElementRegions/elementRegionsGroup/Channel/elementSubRegions/cb-1_1_2
	Group has a child 'fluidNames' in the baseline file but not the file to compare.
```
```
Error: /Problem/domain/MeshBodies/mesh1/meshLevels/Level0/ElementRegions/elementRegionsGroup/Fracture/elementSubRegions/FractureSubRegion
	Group has a child 'porousMaterialNames' in the file to compare but not the baseline file.
Error: /Problem/domain/MeshBodies/mesh1/meshLevels/Level0/ElementRegions/elementRegionsGroup/Fracture/elementSubRegions/FractureSubRegion
	Group has a child 'porosityModelName' in the file to compare but not the baseline file.
```
no real results change:
```
INFO: Total number of log files processed: 1503

INFO: No unfiltered differences were found.

```

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -250,22 +249,19 @@ void CompositionalMultiphaseBase::registerDataOnMesh( Group & meshBodies )
       }
 
       // If at least one region has a capillary pressure model, consider it enabled for all
-      string const capPresName = getConstitutiveName< CapillaryPressureBase >( subRegion );
-      if( !capPresName.empty() )
+      if( !getConstitutiveName< CapillaryPressureBase >( subRegion ).empty() )
```
dkachuma:
```suggestion
      m_hasCapPressure = !getConstitutiveName< CapillaryPressureBase >( subRegion ).empty();
```

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -228,8 +229,6 @@ void CompositionalMultiphaseBase::postInputInitialization()
 
 void CompositionalMultiphaseBase::registerDataOnMesh( Group & meshBodies )
 {
-  using namespace fields::flow;
```
dkachuma, line 278: Just curious as to why we are removing this. Is this part of the coding standard?

dkachuma (approved):

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
paveltomin: `using namespace fields` already defined in the very beginning of the file, so no need to duplicate it here

paveltomin: i am not sure if there is a coding standard about that, just trying to make things consistent across the solvers

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
paveltomin: this would change the logic, i committed first but now will revert

src/coreComponents/physicsSolvers/fluidFlow/wells/WellSolverBase.cpp
```diff
@@ -155,15 +149,6 @@ void WellSolverBase::initializePostSubGroups()
...
 
-void WellSolverBase::setConstitutiveNamesCallSuper( ElementSubRegionBase & subRegion ) const
-{
-  PhysicsSolverBase::setConstitutiveNamesCallSuper( subRegion );
-  subRegion.registerWrapper< string >( viewKeyStruct::fluidNamesString() ).
-    setPlotLevel( PlotLevel::NOPLOT ).
-    setRestartFlags( RestartFlags::NO_WRITE ).
-    setSizedFromParent( 0 );
-}
-
```
tjb-ltk: Out of curiosity, what   is the need for PhysicsSolverBase::setConstitutiveNamesCallSuper...
why was this removed

paveltomin: i am not exactly sure about the logic behind, i think it originates from this big PR https://github.com/GEOS-DEV/GEOS/pull/1682
`fluidNames` registration moved somewhere else (next to its assignement), so this function becomes redundant

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
MelReyCG: What about `m_hasCapPressure |= !getConstitutiveName< CapillaryPressureBase >( subRegion ).empty();`?

src/coreComponents/physicsSolvers/fluidFlow/proppantTransport/ProppantTransport.cpp
```diff
@@ -90,73 +91,45 @@ void ProppantTransport::postInputInitialization()
 
 void ProppantTransport::registerDataOnMesh( Group & meshBodies )
 {
-  using namespace fields::proppant;
-
   FlowSolverBase::registerDataOnMesh( meshBodies );
```
MelReyCG, line 95: I tend to use local `using namespace` in fonctions where there are a lot of repetition while not affecting the full cpp file, and sometimes it can help reduce line length a lot and/or improve lisibility (I'm not saying this is the case at this precise place).
Is the point to remove local `using namespace`? I'm not sure we should have a strong rule about that.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
paveltomin: ok...

paveltomin: @MelReyCG done, could you please approve?

src/coreComponents/physicsSolvers/PhysicsSolverBase.hpp
```diff
@@ -1113,6 +1148,18 @@ string PhysicsSolverBase::getConstitutiveName( ParticleSubRegionBase const & sub
...
+void PhysicsSolverBase::setConstitutiveName( ElementSubRegionBase & subRegion, string const & name ) const
+{
+  string & constitutiveName = subRegion.registerWrapper< string >( name ).
+                                setPlotLevel( dataRepository::PlotLevel::NOPLOT ).
+                                setRestartFlags( dataRepository::RestartFlags::NO_WRITE ).
+                                setSizedFromParent( 0 ).
+                                reference();
+  constitutiveName = getConstitutiveName< CONSTITUTIVE_BASE_TYPE >( subRegion );
+  GEOS_ERROR_IF( constitutiveName.empty(), GEOS_FMT( "{}: {} not found on subregion {}",
+                                                     getDataContext(), typeid(CONSTITUTIVE_BASE_TYPE).name(), subRegion.getDataContext() ) );
```
MelReyCG: Thanks for uniformizing that :)

I think we should output a user-friendly name to the user here (`Fluid`, `Proppant`...), the `typeid(T)` names can be disturbing sometimes for them as they don't have visibility on it (IIUC, `geos::constitutive::SlurryFluidBase`, `geos::constitutive::ParticleFluidBase`, `geos::constitutive::MultiFluidBase`...).

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
```diff
@@ -184,54 +186,52 @@ void CompositionalMultiphaseWell::registerDataOnMesh( Group & meshBodies )
                                                                    WellElementSubRegion & subRegion )
     {
       string const & fluidName = getConstitutiveName< MultiFluidBase >( subRegion );
-      GEOS_ERROR_IF( fluidName.empty(), GEOS_FMT( "{}: Fluid model not found on subregion {}",
-                                                  getDataContext(), subRegion.getName() ) );
-
```
MelReyCG, line 195: I'm not sure, wasn't the error on the empty `fluidName` useful here? Or is the point to rely on the error in the `getConstitutiveModel()`->`getGroup()`?

src/coreComponents/physicsSolvers/PhysicsSolverBase.hpp
```diff
@@ -1113,6 +1148,18 @@ string PhysicsSolverBase::getConstitutiveName( ParticleSubRegionBase const & sub
...
 }
 
+template< typename CONSTITUTIVE_BASE_TYPE >
+void PhysicsSolverBase::setConstitutiveName( ElementSubRegionBase & subRegion, string const & name ) const
+{
+  string & constitutiveName = subRegion.registerWrapper< string >( name ).
+                                setPlotLevel( dataRepository::PlotLevel::NOPLOT ).
+                                setRestartFlags( dataRepository::RestartFlags::NO_WRITE ).
+                                setSizedFromParent( 0 ).
+                                reference();
```
MelReyCG: Don't we loose the `setDescription()` content of some wrapper in the process?
- Maybe the method could be named `createConstitutive()` and return the `Wrapper<string>`, so we can use the builder pattern to configure it (`setDescription()` calls...)
- or we could give a user-friendly as parameter (`fluid model`, `relative permeability constitutive model`....) and use that in a generic description (and also in the error just below, to replace the `typeid(T).name()`).

src/coreComponents/physicsSolvers/fluidFlow/proppantTransport/ProppantTransport.cpp
paveltomin: here i was just trying to unify `using namespace` for solvers
basic idea is to add `using namespace fields` consistently but use sub namespaces (like `proppant` here) more consistently to avoid confusions between them

src/coreComponents/physicsSolvers/PhysicsSolverBase.hpp
paveltomin: i am not sure description was very useful here, those names are purely internal GEOS think not exposed to users/not written to restart etc, so can be lost?

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
paveltomin: my logic is that this should be already checked once and for all in `setConstitutiveName`, no need to double check again

src/coreComponents/physicsSolvers/PhysicsSolverBase.hpp
paveltomin: agree but again, this error should not really happen for users or i am missing something?

MelReyCG: Oh, I see! I'm pretty sure a user reported one of those errors to me, but anyway, your point is that it's too late to do user-reporting (which I agree, but did not understood at first).

src/coreComponents/physicsSolvers/PhysicsSolverBase.hpp
MelReyCG: Some of them appear in the documentation (like the `SolidInternalEnergy` or the `SinglePhaseThermalConductivityBase`). Even if very straight-forward for now, these documentation entries will be the place where we will be able to insert more helping information later. Also, I'm not sure we want some totally empty documentation entries.

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
MelReyCG: Ok, so the error in `setConstitutiveName()` may be displayed to users.

MelReyCG (approved):

MelReyCG (approved):

castelletto1 (approved):

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.cpp
tjb-ltk, line 1: what is the reason for removing setConstitutiveNamesCallSuper( subRegion );  it was also removed in other files in this PR..

paveltomin: they kind of redundant, see https://github.com/GEOS-DEV/GEOS/pull/3517/files#r1953015229

CusiniM (approved):

src/coreComponents/physicsSolvers/fluidFlow/wells/SinglePhaseWell.hpp
tjb-ltk, line 1: shouldn't  the function setConstitutiveNames have a brief ??  or is there a convention about where to put virtual functions description... I think one of pushes failed because of no function / parameter specification

tjb-ltk (approved):
Nice cleanup and standardization on use of fields...

rrsettgast (approved):
It is a bit hard to differentiate the namespace changes from functional changes.

I am a little wary of adding a bunch of "using namespace" statements, but I guess I am not going to put up a fight over that. I just hope we don't get issues with ambiguity at some point.

paveltomin:
rebaseline due to:
```
Error: /Problem/domain/MeshBodies/mesh1/meshLevels/Level0/ElementRegions/elementRegionsGroup/Fracture/elementSubRegions/FractureSubRegion
	Group has a child 'porosityModelName' in the file to compare but not the baseline file.
Error: /Problem/domain/MeshBodies/mesh1/meshLevels/Level0/ElementRegions/elementRegionsGroup/Fracture/elementSubRegions/FractureSubRegion
	Group has a child 'porousMaterialNames' in the file to compare but not the baseline file.
```
(for cases with fractures)
and
```
Error: /Problem/domain/MeshBodies/mesh1/meshLevels/Level0/faceManager
	Group has a child 'facePressure' in the file to compare but not the baseline file.
```
(for immiscible 2-phase tests)

no real results changes

PR #3523, opened 2025-01-28
feat: SCEC benchmarks BP6 and BP7

CusiniM:
Collection of changes and scripts needed for the solution of the SCEC benchmarks. 

- [ ] Add support for backslip to quasi-dynamic earthquake solver
- [ ] Fix initialization of the system, ensuring that the rate-and-state equations are compatible with stresses
- [ ] Ensure effective compressive normal stress doesnt become negative (tensile).


src/coreComponents/physicsSolvers/contact/SolidMechanicsLagrangeContactBubbleStab.cpp
```diff
@@ -48,7 +48,7 @@ SolidMechanicsLagrangeContactBubbleStab::SolidMechanicsLagrangeContactBubbleStab
 
   LinearSolverParameters & linSolParams = m_linearSolverParameters.get();
   linSolParams.mgr.strategy = LinearSolverParameters::MGR::StrategyType::lagrangianContactMechanicsBubbleStab;
-  linSolParams.mgr.separateComponents = true;
+  linSolParams.mgr.separateComponents = false;
```
CusiniM: @victorapm for this physics solver, which uses [this](https://github.com/GEOS-DEV/GEOS/blob/develop/src/coreComponents/linearAlgebra/interfaces/hypre/mgrStrategies/LagrangianContactMechanicsBubbleStabilization.hpp) MGR strategy, the boomer AMG setup crashes in parallel during setup. Without the filter everything seems to work fine.

The case that I have been using to debug it is: https://github.com/GEOS-DEV/GEOS/blob/develop/inputFiles/lagrangianContactMechanics/LagrangeContactBubbleStab_FixedSlip_smoke.xml and you can reduce the size to :
nx = 4 and ny = 2 and still reproduce the issue. A partition of `-x 2 ` is sufficient to trigger the crash.

CusiniM: this is the stack trace of the crash:

````
Number of element for each fracture state: stick:          380 | new slip:            0 | slip:             0 | open:             0
        ( Rsolid ) = ( 0.00e+00 )        ( Rt  ) = (    1.549432e+01  )        ( R ) = ( 1.55e+01 )
        MGR preconditioner: numComponentsPerField = [3, 3, 3]
Received signal 11: Segmentation fault: 11

** StackTrace of 0 frames **
Frame 0: _sigtramp 
=====

Received signal 15: Terminated: 15

** StackTrace of 24 frames **
Frame 0: _sigtramp 
Frame 1: opal_progress 
Frame 2: mca_pml_ob1_iprobe 
Frame 3: MPI_Iprobe 
Frame 4: hypre_MPI_Iprobe 
Frame 5: hypre_DataExchangeList 
Frame 6: hypre_ParCSRCommPkgCreateApart_core 
Frame 7: hypre_ParCSRCommPkgCreateApart 
Frame 8: hypre_MatvecCommPkgCreate 
Frame 9: hypre_BoomerAMGBuildModExtPEInterpHost 
Frame 10: hypre_BoomerAMGBuildModExtPEInterp 
Frame 11: hypre_BoomerAMGSetup 
Frame 12: hypre_MGRSetup 
Frame 13: geos::HyprePreconditioner::setup(geos::HypreMatrix const&) 
Frame 14: geos::HypreSolver::setup(geos::HypreMatrix const&) 
Frame 15: geos::PhysicsSolverBase::solveLinearSystem(geos::DofManager const&, geos::HypreMatrix&, geos::HypreVector&, geos::HypreVector&) 
Frame 16: geos::PhysicsSolverBase::solveNonlinearSystem(double const&, double const&, int, geos::DomainPartition&) 
Frame 17: geos::PhysicsSolverBase::nonlinearImplicitStep(double const&, double const&, int, geos::DomainPartition&) 
--------------------------------------------------------------------------
prterun noticed that process rank 1 with PID 5091 on node onix exited on
signal 6 (Abort trap: 6).
````

Comment (victorapm) [reply src/coreComponents/physicsSolvers/contact/SolidMechanicsLagrangeContactBubbleStab.cpp]: Thanks for reporting. Let me take a look

File: src/coreComponents/physicsSolvers/inducedSeismicity/QuasiDynamicEarthQuake.cpp [hunk 2]
```diff
@@ -124,8 +131,8 @@ void QuasiDynamicEarthQuake< RSSOLVER_TYPE >::setTargetDispJump( DomainPartition
       forAll< parallelDevicePolicy<> >( subRegion.size(), [=] GEOS_HOST_DEVICE ( localIndex const k )
       {
         targetDispJump( k, 0 ) = 0.0;
-        targetDispJump( k, 1 ) = deltaSlip( k, 0 );
-        targetDispJump( k, 2 ) = deltaSlip( k, 1 );
+        targetDispJump( k, 1 ) = deltaSlip( k, 0 ) + backSlipRate( k, 0 ) * dt;
```
CusiniM: @VidarStiernstrom this is how I simply added the backSlip.

VidarStiernstrom: Yeah, that should be the general ide. We have to be careful with signs tho, or at least be clear of how we think of  backslipRate. The backslip should correspond to rigid body translation of the two solids, where we want to solve the problem such that our solution is u_sol = u - u_rigid. Then we can specify zero homogenous displacement boundary conditions u_sol = 0. If we have a rigid body plate rate Vp I think we should have targetDispJump = deltaSlip - Vp * dt, but I'm not sure. Could be a good idea to check with Eric.

TLDR; We need to double check the sign of backSlipRate.

CusiniM: okay, we can also play with the sign of the backSlipRate and see what gives us the correct solution.

PR #3529, opened 2025-01-29
fix: SegFault when a MPI rank has no data of a sub-region

### LINKED ISSUE #3528: SegFault when a MPI rank has no data of a sub-region
**Describe the bug**
When running a closed case of TE with the lastest versions, GEOS crashes (SegFault).

**To Reproduce**
I cannot join a case, but I'm posting a PR that will resolve this crash.
It happens because, if I understood correctly, in `CompositionalMultiphaseFVM::scalingForSystemSolution`, we assume that all MPI ranks will contains some information (pressure, density, temperature) for all the sub-regions, but in our case an MPI rank only has a non-simulated region.
The function `scalingForSystemSolution()` has an unsafe derefencing: we should not do `*std::max_element()` without checking if the list has any element at all.

**Expected behavior**
The function `scalingForSystemSolution()` should not dereference without checking if there is any information in the pressure/delta-density/temperature sets.

- Machine: Pangea4
- Compiler: gcc 12.2.1, OpenMPI
- Version: `develop` since the PR #3327 was merged.

**Additional info**
I'm posting a quickfix PR that will expose in `MpiWrapper` methods to safely reduce pairs or pair sets over ranks. I'm also adding a unit test to check the consistency of the solution, to be sure that the reduction will remain stable.

MelReyCG:
Closing #3528 

My goal here is to solve the crash, and to propose a way to reduce (Min/Max) pairs lexicographically over MPI ranks. That can typically be useful to get the min/max value (pressure, temperature...) along with its `globalIndex` in the mesh (while ensuring that the `globalIndex` will remain stable).
It could be generalized to tuples, let me know if that could be useful.


corbett5:
I'm sure there's a reason I haven't thought about, but why not use `std::pair`? And those `mpiPairType::get` functions can be more succinctly written as templated variables.

Example: https://godbolt.org/z/3sxTeGrzs

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.cpp
```diff
@@ -550,10 +550,10 @@ real64 CompositionalMultiphaseFVM::scalingForSystemSolution( DomainPartition & d
...
   real64 minPresScalingFactor = 1.0, minCompDensScalingFactor = 1.0, minTempScalingFactor = 1.0;
 
-
-  std::vector< valueAndLocationType > regionDeltaPresMaxLoc;
-  std::vector< valueAndLocationType > regionDeltaTempMaxLoc;
-  std::vector< valueAndLocationType > regionDeltaCompDensMaxLoc;
+  std::vector< MpiWrapper::PairType< real64, globalIndex > > regionDeltaPresMaxLoc;
+  std::vector< MpiWrapper::PairType< real64, globalIndex > > regionDeltaCompDensMaxLoc;
+  std::vector< MpiWrapper::PairType< real64, globalIndex > > regionDeltaTempMaxLoc;
+  int64_t pushCount=0;
```
CusiniM:
```suggestion
 integer pushCount=0;
```

CusiniM: I am guessing that you can also get rid of https://github.com/GEOS-DEV/GEOS/blob/ac45ee2a112dca04a2ecddacfc8252d2a486c29e/src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.hpp#L219

MelReyCG: Sorry, this PR is still in development.
That `pushCount` variable was for testing, I remove it.
And yes, I was about to remove this struct.

MelReyCG:
Hi @corbett5!

1. Yes, I indeed started with `std::pair`, but I did not continue with it because is not trivially copiable. As it seem for the least that it could be used in performance critical places, it seems like a reasonable requirement.
I wonder if this is why the original code here did not go for `std::pair`.
[https://godbolt.org/z/KE7TEMx7d](https://godbolt.org/z/KE7TEMx7d)
[https://stackoverflow.com/questions/64978278/what-is-the-performance-problem-with-stdpair-assignment-operator](https://stackoverflow.com/questions/64978278/what-is-the-performance-problem-with-stdpair-assignment-operator)

2. IIUC, you suggest the following?
```c++
/* no default get() implementation, please add a template specialization and add it in the "testMpiWrapper" unit test. */
template< typename FIRST, typename SECOND >
MPI_Datatype const mpiPairType;

template<> MPI_Datatype const mpiPairType< float, int > = MPI_FLOAT_INT;
template<> MPI_Datatype const mpiPairType< double, int > = MPI_DOUBLE_INT;
template<> MPI_Datatype const mpiPairType< int, int > = MPI_2INT;
template<> MPI_Datatype const mpiPairType< long int, int > = MPI_LONG_INT;
template<> MPI_Datatype const mpiPairType< long int, long int > = getMpiCustomPairType< long int, long int >();
template<> MPI_Datatype const mpiPairType< long long int, long long int > = getMpiCustomPairType< long long int, long long int >();
template<> MPI_Datatype const mpiPairType< double, long int > = getMpiCustomPairType< double, long int >();
template<> MPI_Datatype const mpiPairType< double, long long int > = getMpiCustomPairType< double, long long int >();
template<> MPI_Datatype const mpiPairType< double, double > = getMpiCustomPairType< double, double >();
```

corbett5:
@MelReyCG 

1. Gotcha, yeah I think using `std::pair` would just work, but I think using the custom struct is not a bad idea.
2. I forgot about the `getMpiCustomPairType` function, on second thought would it be cleaner if we eliminated the `mpiPairType` variable completely and just put all the logic in `getMpiCustomPairType`? You can't template specialize functions, but it could be done with `if constexpr` checks of the types I believe.

MelReyCG:
@corbett5 I added a new proposal in the "[proposal #1](https://github.com/GEOS-DEV/GEOS/pull/3529/commits/cfb41e3badf4af93ad1fc33039aa60a7ad4cb2f6)" commit.
For now, I did not go for the `if constexpr` solution, but it would easily be feasible if you guys think it would be cleaner than this proposal.

CusiniM:
> @MelReyCG
> 
>     1. Gotcha, yeah I think using `std::pair` would just work, but I think using the custom struct is not a bad idea.
> 

I remember an old discussion about the std::pair layout not being guaranteed to be contiguous in memory which I think is the reason why it was not used in the first place. I don't recall the details of how it was being used though. 

>     2. I forgot about the `getMpiCustomPairType` function, on second thought would it be cleaner if we eliminated the `mpiPairType` variable completely and just put all the logic in `getMpiCustomPairType`? You can't template specialize functions, but it could be done with `if constexpr` checks of the types I believe.

Personally I don't have a strong preference between using `if constexpr` in a single function or using template specializations as @MelReyCG  has done. They way it's written now seems pretty tidy and easy to read but something equally good could probably be obtained using` if constexpr`. Maybe this second option is a bit more modern?

corbett5:
> 
> Personally I don't have a strong preference between using `if constexpr` in a single function or using template specializations as @MelReyCG has done. They way it's written now seems pretty tidy and easy to read but something equally good could probably be obtained using` if constexpr`. Maybe this second option is a bit more modern?

I think the single function is cleaner, but this is in my opinion a small improvement to a minor part of this PR and I think the current implementation is fine.

MelReyCG:
@CusiniM 
> I remember an old discussion about the std::pair layout not being guaranteed to be contiguous in memory which I think is the reason why it was not used in the first place. I don't recall the details of how it was being used though.

This is exactly why I chose the `struct`, it seems that there is no guarantee that the compilator will be allowed to optimize `std::pair<>` copies as much as structures. Another source: [https://www.reddit.com/r/cpp/comments/ar4ghs/stdpair_disappointing_performance/?rdt=48069](https://www.reddit.com/r/cpp/comments/ar4ghs/stdpair_disappointing_performance/?rdt=48069)

About the `if constexpr`, you can see that I used it in `getMpiPairReductionOp()`, I tend to use it more when it concerns implementation details, or when it can simplify overcomplicated `std::enable_if`.
For such "switch on type T" usage, I prefer the good old template. I think that's a matter of style / tastes. If the code readability is not affected, I think we can keep template specializations.

MelReyCG:
I finished to write the last comments I think, we may be good to go?

CusiniM (approved):
LGTM. Thanks @MelReyCG

src/coreComponents/common/MpiWrapper.hpp
```diff
@@ -706,6 +783,58 @@ MPI_Datatype getMpiType()
...
+            inout = in;
+        }
+      }
+    };
+
+    MPI_Op mpiOp;
+    GEOS_ERROR_IF_NE( MPI_Op_create( customOpFunc, 1, &mpiOp ), MPI_SUCCESS );
+    return mpiOp;
+  };
+  static MPI_Op mpiOp{ createOpHolder() };
```
corbett5, line 862: Why does this need to be static? `mpiOp` is returned by value, so it shouldn't matter if it gets destroyed after the function returns.

src/coreComponents/common/MpiWrapper.hpp
```diff
@@ -706,6 +783,58 @@ MPI_Datatype getMpiType()
...
+template<> MPI_Datatype getMpiPairType< double, int >();
+template<> MPI_Datatype getMpiPairType< double, long int >();
+template<> MPI_Datatype getMpiPairType< double, long long int >();
+template<> MPI_Datatype getMpiPairType< double, double >();
+
+// It is advised to always use this custom operator for pairs as MPI_MAXLOC is not a true lexicographical comparator.
+template< typename FIRST, typename SECOND, MpiWrapper::PairReduction OP >
+MPI_Op getMpiPairReductionOp()
+{
+  static auto const createOpHolder = [] () {
```
corbett5, line 832: What is the point of this `createOpHolder` lambda?

MelReyCG: This lambda helps to scope & encapsulate the `mpiOp` construction. It will be called only once, at the first `getMpiPairReduction<FIRST,SECOND>()` call for the given pair of templated types.

src/coreComponents/common/MpiWrapper.hpp
MelReyCG: The static storage here does matter as `MPI_Op` (or MPI_Datatype for custom types) represents a persistent MPI resource that must:
- be initialized only once to avoid mem-leaks (`MPI_X_create()` functions allocates internal resources),
- remain valid for the entire MPI lifetime (`MPI_Init`->`MPI_Finalize`),
- be shared across all calls to this function.

The static construction I used here is a variation of the _"Meyer's Singleton"_, which lazily constructs a unique instance of an object when the resource is first requested (by calling the lambda in this case).

At first I did not mind calling `MPI_Op_free()` since we already call `MPI_Finalize()` but I found the following in the [MPI documentation](https://www.mpi-forum.org/docs/mpi-4.1/mpi41-report/node268.htm): 
> The call to MPI_FINALIZE does not free objects created by MPI calls; these objects are freed using MPI_ XXX_FREE, MPI_COMM_DISCONNECT, or MPI_FILE_CLOSE calls.

To completely manage these resources lifetime I will work a bit more on this PR.

corbett5: Oh god, yeah I get it. If performance isn't a concern (and I doubt it is) I would simply createa new op every time and free it every time as well. You could do this pretty easily with a RAII object that had a move constructor and deleted copy constructor.

MelReyCG: @corbett5 Indeed, a RAII object would do the job.
Yesterday I finished with another approach in which I kept using the MPI operation / data-type caching and stored their trace to release them on `MPI_Finalize()`. Let me know if my approach is right ([commit](https://github.com/GEOS-DEV/GEOS/pull/3529/commits/d49b850b6e9fb4820239ebc427faa7eae9c5afd2)).

corbett5: I prefer my approach (in which these objects are create and destroyed on the fly) since it would eliminate the need for static variables, but this is still quite the improvement and I'm not sure it's worth your time to rewrite it.

MelReyCG: I was thinking that statics were appropriate here, as the lifetime of those are application-level (same as MPI), because the number of MPI operations/datatypes is really low and the same ones will be really called for each timesteps.

src/coreComponents/common/MpiWrapper.hpp
```diff
@@ -133,6 +135,28 @@ struct MpiWrapper
...
+    Min, //!< Min pair first value
+  };
+
+  /**
+   * @struct PairType
+   * Represents a pair of values. `first` (typically, a given measure) is the primary value for comparison,
+   * `second` (typically an index) will be compared in case of `first` equality.
+   */
+  template< typename FIRST, typename SECOND >
+  struct PairType
```
dkachuma, line 154: Can't we simply use `std::pair`?

dkachuma (approved):

src/coreComponents/common/MpiWrapper.hpp
paveltomin: https://github.com/GEOS-DEV/GEOS/pull/3327#pullrequestreview-2370096798

paveltomin (approved):
looks very complex for just a fix but as long as it works and you guys can agree on the best way to do it...

src/coreComponents/common/MpiWrapper.hpp
MelReyCG: And more details here: https://github.com/GEOS-DEV/GEOS/pull/3529#issuecomment-2631507315

MelReyCG:
@paveltomin Indeed, I put more work than necessary in this PR, but as I was removing the publicly accessible `maxValLoc()` & `valueAndLocation` struct, I would like to propose a stable way of reducing pairs that may be used somewhere else without troubles.

tjb-ltk:
Good find, I can't comment on the details of the fix, but looks general purpose  and nice to see sufficient unit tests so others can find examples of usage..... 
 I assumed all ranks would have elements. What is the use case where a simulation model has cores with no elements,  seems like an expensive way to go...

MelReyCG:
Just to notify, we agreed with @paveltomin to merge his PR #3545 into this one.

rrsettgast (approved):

PR #3531, opened 2025-01-29
refactor: move contact into solidMechanics

paveltomin:


rrsettgast (approved):
@paveltomin I don't see any changes that aren't just path corrections and scope corrections. Is this accurate?

paveltomin:
> @paveltomin I don't see any changes that aren't just path corrections and scope corrections. Is this accurate?

yes correct

CusiniM (approved):

PR #3541, opened 2025-02-07
fix:  sync well constraint

tjb-ltk:
fix: All cores evaluate the well constraint switching logic and assumes that all cores have the current well control id, however depending on the control type, global comms are performed, or not.     A bug was is exposed in a scenario where the well status changed from shut to open, the well control also changed state at well open time,  timestep cuts then closed the well because it wasn't time to open, and after the dt became large enough to open the well again. The core that has the well segment element with assigned constraint is the owner and only the owner changes the control id.  Hence the well current control id wasn't synced with all cores.  
This PR adds comms to ensure all cores have the  control id.


src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
```diff
@@ -1916,9 +1916,17 @@ void CompositionalMultiphaseWell::assemblePressureRelations( real64 const & time
...
+      // Only subregion owner evaluates well control and control changes need to be broadcast to all ranks
+      if ( subRegion.isLocallyOwned() )
+      {
+          owner = MpiWrapper::commRank( MPI_COMM_GEOS );
+      }
+      owner = MpiWrapper::max( owner );
 
       WellControls & wellControls = getWellControls( subRegion );
-
+       GEOS_LOG_RANK("constraint " << wellControls.getName() << " time_n " << time_n << " dt " << dt << " open " << wellControls.isWellOpen( time_n + dt )  );
```
paveltomin: just remove this debug and it should be good to go

paveltomin (approved):
our problematic case can run now, thanks

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp

PR #3543, opened 2025-02-10
fix: quickfix to re-enable well solvers parameters

MelReyCG:
I unintentionally broke in a previous PR (#3418) the ability to configure the `NonlinearSolverParameters` or `LinearSolverParameters` of well solvers.
We needed it recently while experimenting on a case, I'm thinking it is surprising that it did not broke anything in the first place. _Is there a reason for never configuring well solvers?_

For now, I'm not sure that this is the cleanest way to fix it. Should I better create a `struct groupKeyStruct : PhysicsSolverBase::groupKeyStruct` within `WellSolverBase` (and include `WellControls.hpp` in `WellSolverBase.hpp`).


paveltomin (approved):

paveltomin:
thanks, we saw the issue but were too busy with something else to report, sorry

paveltomin:
what is wrong with delegating everything except wellControls to the solver base?

tjb-ltk:
I agree with paveltomins comment about delegating to solver base.  Presently the well doesn't need a linear solver or nonlinear solver.  In the nonlinear one could imagine setting max changes on  well variables updates, but these could added to CompositionalMultiphaseReservoir's NonlinearSolverParamenters instead of  CompositionalMultiphaseWell's  NonlinearSolverParamenters, which I don't think is used.   What prompted putting it back,  a model that had NonlinearSolverParamenters in CompositionalMultiphaseWell ?  If this was only reason then I would change the model

MelReyCG:
1. Sorry, I'm not sure to understand what `SinglePhaseWell` & `CompositionalMultiphaseWell` really are. Are they solvers which rely (non)linear-solvers to define the produced/injected flow in the reservoir?
2. On a stritcly programmatic standpoint, does the `PhysicsSolverBase::execute()` is designed to ever be called on well solvers? If so, `m_nonlinearSolverParameters` would be used extensively in it. If not, I did see an usage of `m_nonlinearSolverParameters.m_minNormalizer` in `SinglePhaseWell` & `CompositionalMultiphaseWell`.
3. Regarding your planned works, do you think we should uncouple `WellSolverBase` from `PhysicsSolverBase`?

PS: @paveltomin No problem, it is me who broke something!

tjb-ltk:
1) no, they just setup equations for the coupled reservoir and well solve
2) if there was ever an well estimator or sequential solve strategy involving isolated well solves, separate nonlinear/linear solver parameters would definitely be needed.  
3) not in my immediate planned works , but as mentioned in 2) that would make sense but requires a discussion broader discussion
  @CusiniM 
  should the params be put back in to avoid breaking other xmls even though params are not used ?

CusiniM:
> I unintentionally broke in a previous PR (#3418) the ability to configure the `NonlinearSolverParameters` or `LinearSolverParameters` of well solvers. We needed it recently while experimenting on a case, I'm thinking it is surprising that it did not broke anything in the first place. _Is there a reason for never configuring well solvers?_
> 

I actually think that this is a good thing. Currently, it makes no sense to try to set `NonlinearSolverParameters`  and `LinearSolverParameters` for the well solvers since they are not really used. There is only one parameter used by the well solvers and it's the `minNormalizer` and I am not even sure it's something we really want users to fiddle with. We could compute an appropriate value for it. So, it may actually be better not to expose these to the users for now since it may be confusing. We could improve the error message by saying that specifying these parameters for a well solver is useless.

CusiniM:
>     3. Regarding your planned works, do you think we should uncouple `WellSolverBase` from `PhysicsSolverBase`?

what I think should happen is separating the nonlinear solver strategy from PhysicsSolverBase and let physics solvers only assemble equations.

src/coreComponents/schema/schema.xsd
```diff
@@ -222,10 +222,18 @@
 				</xsd:unique>
 			</xsd:element>
 			<xsd:element name="Functions" type="FunctionsType" maxOccurs="1">
+				<xsd:unique name="FunctionsCompositeFunctionUniqueName">
```
dkachuma: Some of these schema changes appear unrelated to the code changes. Are these perhaps part of a different PR?

dkachuma (approved):

paveltomin:
@MelReyCG what is the status of this PR?

paveltomin:
@MelReyCG still needed?

PR #3545, opened 2025-02-12
fix: resolve crash when location was not found (-1)

paveltomin:
should go after https://github.com/GEOS-DEV/GEOS/pull/3529

small fix to avoid crash when no max change location was found (-1)


tjb-ltk (approved):

paveltomin:
> small fix to avoid crash when no max change location was found (-1)

MelReyCG (approved):

PR #3546, opened 2025-02-12
ci: fix 1d edfm case and add it to ats

paveltomin:
"fix" = disable line search


tjb-ltk (approved):
disabling ineSearchActive,  is this to make it run, or a case of functionality not available for edfm

paveltomin:
> disabling ineSearchActive, is this to make it run, or a case of functionality not available for edfm

it removes time step cuts and allows it to finish yes
linesearch implementation is very general, not constraint for physics/features

jhuang2601 (approved):

CusiniM (approved):

PR #3549, opened 2025-02-13
fix: re-enable the use of `useCartesianOuterBoundary`

### LINKED ISSUE #3548: Bug in InternalWellboreGenerator for enabling  `useCartesianOuterBoundary`
**Describe the bug**
Instead of using box shaped geometry (Cartesian aligned outer boundary) for wellbore problems, radial shaped outer boundary can be defined previously using `useCartesianOuterBoundary` in InternalWellboreGenerator, which is not working anymore.

**To Reproduce**
For this benchmark example:
https://github.com/GEOS-DEV/GEOS/blob/6769693855ab7cbd185985c1f99c296e395ebc76/inputFiles/wellbore/ThermoPoroElasticWellbore_benchmark.xml#L10-L22
adding `useCartesianOuterBoundary="1"` to define radial shaped outer boundary
and run it with develop repo, following error message is shown:
```
***** ERROR
***** LOCATION: /shared/data1/Users/j0551570/Compilation/Develop_250130/GEOS/src/coreComponents/mesh/generators/InternalWellboreGenerator.cpp:275
***** Controlling expression (should be false): m_cartesianOuterBoundary >= m_vertices[0].size()-1
***** Rank 0: useCartesianOuterBoundary must be strictly smaller than the number of radial blocks (equal to 1 here)
```

**Expected behavior**
Enable the use of `useCartesianOuterBoundary` to define a cylindrical shape of the far-field boundary for axisymmetric wellbore problems.

jhuang2601:
With this fix, following mesh for axisymmetric wellbore problems can be generated with `InternalWellboreGenerator` 
image


sytuannguyen (approved):

rrsettgast:
@jhuang2601 Can you explain the bug in more detail. I don't understand how you were unable to generate that mesh prior to the fix.

paveltomin:
@jhuang2601 any update?

jhuang2601:
> @jhuang2601 Can you explain the bug in more detail. I don't understand how you were unable to generate that mesh prior to the fix.

@rrsettgast Here is the description for the bug: https://github.com/GEOS-DEV/GEOS/issues/3548

jhuang2601:
> @jhuang2601 any update?

@paveltomin Waiting for the review from code owners.

herve-gross:
@rrsettgast and @wrtobin as code owners, can you take a look?

rrsettgast (approved):

PR #3552, opened 2025-02-14
fix: segFault in ThermoPoromechanics EFEM case

CusiniM:


paveltomin (approved):

tjb-ltk (approved):
Merged with single phase prop refactor branch and  ThermoPoroElastic_efem-edfm_verticalFrac_smoke.xml  results matches GEOS-DEV results

PR #3558, opened 2025-02-19
feat: improve bounds checking

rrsettgast:
This PR adds further bounds checking in GEOS.

- [x] std::vector wrapper that has bounds checking for `::operator[]( index )`
~~- [ ] `std::map/unordered_map` wrapper that has bounds checking for `::operator[]( index )`~~
~~- [ ] disallow use of `map/unordered_map::operator[]` to perform insertion.~~

the new wrappers are preprocessor guarded with aliases s.t. they alias to the std::container in the non-bounds checking case.


rrsettgast:
@OmarDuran I think this would solve most issues with bounds checking of `std::vector`

OmarDuran:
> @OmarDuran I think this would solve most issues with bounds checking of `std::vector`

@rrsettgast  Thanks for fixing this, hopefully the CI will be testing bounds soon. I noticed another issue in [VTKUtilities](https://github.com/GEOS-DEV/GEOS/blob/4aa2c9fd542b18cbbb493580a461f53cb33f2e42/src/coreComponents/mesh/generators/VTKUtilities.cpp#L1438) where nodeTri0 has size 3. This clearly fails in Debug mode, so it’ll need to be addressed as well.

arng40:
@rrsettgast What would you like us to do?

wrtobin (approved):
looks good.
- do we want to remove the commented-out map wrappers? 
- im not the biggest fan of `stdVector` as the name, but don't really have a preferred alternative, so if we wanna go with that for now its fine

rrsettgast:
@MelReyCG @arng40 

Can you guys look into the 2nd and third task:

- [ ] std::map/unordered_map wrapper that has bounds checking for ::operator[]( index )
- [ ] disallow use of map/unordered_map::operator[] to perform insertion.

src/coreComponents/common/StdContainerWrappers.hpp
```diff
@@ -0,0 +1,161 @@
...
+  using Base::Base;
+
+  /// inherit std::vector::push_back()
+  using Base::push_back;
+
+  /// inherit std::vector::size()
+  using Base::size;
+
+  /// inherit std::vector::empty()
+  using Base::empty;
```
MelReyCG: Why do these usings are needed? As you are inheriting publicly from `std::vector`, aren't they already visible?

MelReyCG:
The started work here really is a good thing, `GEOS_ENABLE_BOUNDS_CHECK` not having effect on `std` container was problematic.
We can imagine adding some exceptions with more information in the future, in case of out-of-bound or not-existing key.

MelReyCG:
@arng40 and / or me will work on those 2 last tasks. After that, we will also test `GEOS_ENABLE_BOUNDS_CHECK` on the last crashing cases we got.

src/coreComponents/common/StdContainerWrappers.hpp
rrsettgast: yeah...they certainly should be...I don't recall why I put these in....perhaps copilot just started to do it and I didn't think about it.

rrsettgast: I removed them.

PR #3565, opened 2025-02-24
Add Thermal Stress Fix

npillardou:
This PR corrects the thermal stress calculations based on initial Temperature.


npillardou:
I think there is a mistake in the files modified, the good ones are not attached, I don't know why

PR #3566, opened 2025-02-24
refactor: Refactoring dump input option 

arng40:
This PR aim to refactor, with the tableFormatter, the dumping output fonction.
When we have unused input option to an XML element node, we get this kind of input :
```
***** LOCATION: /data/pau901/VDG_GSI/ARNAUDD/geosRepos/codes/GEOS/src/coreComponents/dataRepository/Group.cpp:247
***** Controlling expression (should be false): processedAttributes.count( attributeName ) == 0
***** Rank 0: Error in compflow (2ph_cap_1d_ihu.xml, l.5): XML Node at '/Problem/Solvers/CompositionalMultiphaseFVM' contains unused attribute 'targeftRegions'.
Valid attributes are:
  |         name         |  opt/req  | Description 
  |----------------------|-----------|-----------------------------------------
  | cflFactor            | OPTIONAL  | Factor to apply to the `CFL condition <http://en.wikipedia.org/wiki/Courant-Friedrichs-Lewy_condition>`_ when calculating the maximum allowable time step. Values should be in the interval (0,1]  
  | discretization       | REQUIRED  | Name of discretization object (defined in the :ref:`NumericalMethodsManager`) to use for this solver. For instance, if this is a Finite Element Solver, the name of a :ref:`FiniteElement` should be specified. If this is a Finite Volume Method, the name of a :ref:`FiniteVolume` discretization should be specified. 
  | targetRegions        | REQUIRED  | Allowable regions that the solver may be applied to. Note that this does not indicate that the solver will be applied to these regions, only that allocation will occur such that the solver may be applied to these regions. The decision about what regions this solver will beapplied to rests in the EventManager. 
  | initialDt            | OPTIONAL  | Initial time-step value required by the solver to the event manager. 
  | writeLinearSystem    | OPTIONAL  | Write matrix, rhs, solution to screen ( = 1) or file ( = 2). 
  | logLevel             | OPTIONAL  | Sets the level of information to write in the standard output (the console typically).
Information output from lower logLevels is added with the desired log level 
  | isThermal            | OPTIONAL  | Flag indicating whether the problem is thermal or not. 
  | allowNegativePressure | OPTIONAL  | Flag indicating if negative pressure is allowed 
  | maxAbsolutePressureChange | OPTIONAL  | Maximum (absolute) pressure change in a Newton iteration 
  | maxSequentialPressureChange | OPTIONAL  | Maximum (absolute) pressure change in a sequential iteration, used for outer loop convergence check 
  ```
 With the table, we get instead :
 
 ```
 -------------------------------------------------------------------------------------------------------------------------------------
|               name               |    flag    |                                    Description                                    |
-------------------------------------------------------------------------------------------------------------------------------------
|  logLevel                        |  OPTIONAL  |  Log level                                                                        |
|  lineSearchAction                |  OPTIONAL  |  How the line search is to be used. Options are:                                  |
|                                  |            |  * None    - Do not use line search.                                              |
|                                  |            |  * Attempt - Use line search. Allow exit from line search without achieving       |
|                                  |            |  smaller residual than starting residual.                                         |
|                                  |            |  * Require - Use line search. If smaller residual than starting resdual is not    |
|                                  |            |  achieved, cut time-step.                                                         |
|  lineSearchInterpolationType     |  OPTIONAL  |  Strategy to cut the solution update during the line search. Options are:         |
|                                  |            |  * Linear                                                                         |
|                                  |            |  * Parabolic                                                                      |
|  lineSearchMaxCuts               |  OPTIONAL  |  Maximum number of line search cuts.                                              |
|  lineSearchCutFactor             |  OPTIONAL  |  Line search cut factor. For instance, a value of 0.5 will result in the          |
|                                  |            |  effective application of the last solution by a factor of (0.5, 0.25, 0.125,     |
|                                  |            |  ...)                                                                             |
|  lineSearchStartingIteration     |  OPTIONAL  |  Iteration when line search starts.                                               |
|  lineSearchResidualFactor        |  OPTIONAL  |  Factor to determine residual increase (recommended values: 1.1 (conservative),   |
|                                  |            |  2.0 (relaxed), 10.0 (aggressive)).                                               |
|  normType                        |  OPTIONAL  |  Norm used by the flow solver to check nonlinear convergence. Valid options:      |
|                                  |            |  * Linfinity                                                                      |
|                                  |            |  * L2                                                                             |
|  minNormalizer                   |  OPTIONAL  |  Value used to make sure that residual normalizers are not too small when         |
|                                  |            |  computing residual norm.                                                         |
|  newtonTol                       |  OPTIONAL  |  The required tolerance in order to exit the Newton iteration loop.               |
|  newtonMaxIter                   |  OPTIONAL  |  Maximum number of iterations that are allowed in a Newton loop.                  |
|  newtonMinIter                   |  OPTIONAL  |  Minimum number of iterations that are required before exiting the Newton loop.   |
|  maxAllowedResidualNorm          |  OPTIONAL  |  Maximum value of residual norm that is allowed in a Newton loop                  |
|  allowNonConverged               |  OPTIONAL  |  Allow non-converged solution to be accepted. (i.e. exit from the Newton loop     |
|                                  |            |  without achieving the desired tolerance)                                         |
|  timeStepDecreaseIterLimit       |  OPTIONAL  |  Fraction of the max Newton iterations above which the solver asks for the        |
|                                  |            |  time-step to be decreased for the next time-step.                                |
|  timeStepIncreaseIterLimit       |  OPTIONAL  |  Fraction of the max Newton iterations below which the solver asks for the        |
|                                  |            |  time-step to be increased for the next time-step.                                |
|  timeStepDecreaseFactor          |  OPTIONAL  |  Factor by which the time-step is decreased when the number of Newton iterations  |
|                                  |            |  is large.                                                                        |
|  timeStepIncreaseFactor          |  OPTIONAL  |  Factor by which the time-step is increased when the number of Newton iterations  |
|                                  |            |  is small.                                                                        |
|  subcycling                      |  OPTIONAL  |  Flag to decide whether to iterate between sequentially coupled solvers or not.   |
|  nonlinearAccelerationType       |  OPTIONAL  |  Nonlinear acceleration type for sequential solver.                               |
-------------------------------------------------------------------------------------------------------------------------------------
```


rrsettgast (approved):

src/coreComponents/common/format/table/TableFormatter.cpp
```diff
@@ -241,6 +241,68 @@ void TableTextFormatter::setLinks( std::vector< TableLayout::Column > & columns
   }
 }
 
+std::string_view ltrim( std::string_view s )
```
MelReyCG: Could be constexpr.

src/coreComponents/common/format/table/TableFormatter.cpp
```diff
@@ -241,6 +241,68 @@ void TableTextFormatter::setLinks( std::vector< TableLayout::Column > & columns
...
+{
+  std::size_t const first = s.find_first_not_of( " " );
+  if( first != std::string::npos )
+  {
+    return s.substr( first, ( s.size() - first ) );
+  }
+  return {};
+}
+
+bool formatLinesWithMaxSize( std::vector< std::string > & cellLines, size_t maxLength )
```
MelReyCG: those two functions can be helpful! You could generalize a bit (erase table/cell namings) and add them to `StringUtilities.xpp` :)

Also, this function produces incorrect results for `{ "Les shorts c'est pratique, ça laisse les jambes bien au frais.", 8}` (obviously, also for `maxLength=6`).

src/coreComponents/common/format/table/TableLayout.hpp
```diff
@@ -552,9 +572,14 @@ class TableLayout
...
-  bool m_wrapLine = true;
-
+  // Indicate if we have a line break a the beginning of the table
+  bool m_lineBreakAtBegin = true;
+  // Contain the table tible
   string m_tableTitle;
+  // Contain the max width for each column
+  size_t m_tableColumnMaxWidth = std::numeric_limits< size_t >::max();
+  // Indicate if a max column width has been set
+  bool m_hasMaxColumnWidthSet = false;
```
MelReyCG: you could only keep one attribute here:
```c++
using NoColumnMaxWidth = std::numeric_limits< size_t >::max();
size_t m_tableColumnMaxWidth = NoColumnMaxWidth;
```

And in `isMaxColumnWidthSet()` : `return m_tableColumnMaxWidth == NoColumnMaxWidth;`

src/coreComponents/dataRepository/Group.cpp
```diff
@@ -302,13 +305,29 @@ string Group::dumpInputOptions() const
 {
   string rval;
 
-  bool writeHeader = true;
+  TableLayout logLayout( "", {TableLayout::Column()
```
MelReyCG:
```suggestion
  TableLayout const logLayout( "", {TableLayout::Column()
```

src/coreComponents/dataRepository/Wrapper.hpp
```diff
@@ -727,11 +727,11 @@ class Wrapper final : public WrapperBase
                                                                        targetNode,
                                                                        inputFlag == InputFlags::REQUIRED );
           GEOS_THROW_IF( !m_successfulReadFromInput,
```
MelReyCG, line 729: Can you show a comprehensive output of the error?

MelReyCG (changes requested):

src/coreComponents/dataRepository/Wrapper.hpp
arng40:
```***** XML parsing error at node named CompositionalMultiphaseFVM, attribute discretization (2ph_cap_1d_ihu.xml, l.5).
***** Input value: ''
***** LOCATION: /data/pau901/VDG_GSI/ARNAUDD/geosRepos/codes/GEOS4/src/coreComponents/dataRepository/Wrapper.hpp:729
***** Controlling expression (should be false): !m_successfulReadFromInput
***** Rank 3: XML Node CompositionalMultiphaseFVM (2ph_cap_1d_ihu.xml, l.5) with name=compflow is missing required attribute 'discretization'.
For more details, please refer to documentation at:
http://geosx-geosx.readthedocs-hosted.com/en/latest/docs/sphinx/userGuide/Index.html
```

paveltomin (approved):

src/coreComponents/dataRepository/Wrapper.hpp
MelReyCG: Why did you remove the available options?

src/coreComponents/dataRepository/Group.cpp
```diff
@@ -302,13 +305,29 @@ string Group::dumpInputOptions() const
 {
   string rval;
 
-  bool writeHeader = true;
+  TableLayout const logLayout = TableLayout( "", {TableLayout::Column()
+                                                    .setName( "name" )
+                                                    .setValuesAlignment( TableLayout::Alignment::left ),
+                                                  "flag",
```
MelReyCG:
```suggestion
                                                  "Requirement",
```

src/coreComponents/dataRepository/Wrapper.hpp
arng40: I did not remove the available options, in fact there was each time one option displayed, so I cleaned this part to remove the headers that wasn't needed.
Also I cannot use my table because it means adding others dependencies in Wrapper.hpp

MelReyCG (approved):

PR #3568, opened 2025-02-25
feat: Fix Thermal Stress Updates

npillardou:
This PR tries to correct the thermal stress calculations based on the difference from initialTemperature in the formula.
This way, thermal stress is ONLY introduced by a temperature change from the initial stage.


jhuang2601 (approved):

paveltomin:
Few relatively small comments:

- probably no need to introduce `smallStrainUpdateThermoPoromechanics` and make it different from `smallStrainUpdatePoromechanics`, only one updated version should be enough to avoid confusion and potential inconsistencies
- same for `computeTotalStressWithInitialTemperature` vs `computeTotalStress`

the rest seems fine to me, thanks for the efforts.

Between `initialTemperature` or `referenceTemperature` i prefer the second one, but that would require introducing it for coupled solid and making some logic around that, probably can be done later.

paveltomin (approved):

npillardou:
> Few relatively small comments:
> 
> * probably no need to introduce `smallStrainUpdateThermoPoromechanics` and make it different from `smallStrainUpdatePoromechanics`, only one updated version should be enough to avoid confusion and potential inconsistencies
> * same for `computeTotalStressWithInitialTemperature` vs `computeTotalStress`
> 
> the rest seems fine to me, thanks for the efforts.
> 
> Between `initialTemperature` or `referenceTemperature` i prefer the second one, but that would require introducing it for coupled solid and making some logic around that, probably can be done later.


@paveltomin @jhuang2601 I tried to refactor the implementation of the fix thanks to your comments. 
Please can you review these modifications? 
Thank you

paveltomin:
> @paveltomin @jhuang2601 I tried to refactor the implementation of the fix thanks to your comments. Please can you review these modifications? Thank you

looks good to me, thanks

src/coreComponents/constitutive/solid/PorousSolid.hpp
```diff
@@ -139,6 +139,7 @@ class PorousSolidUpdates : public CoupledSolidUpdates< SOLID_TYPE, BiotPorosity,
                                                   real64 const & pressure_n,
                                                   real64 const & temperature,
                                                   real64 const & temperature_n,
+                                                  real64 const & referenceTemperature,
```
Guotong-Ren, line 142: what is the difference between referenceTemperature and temperature_n here?

src/coreComponents/constitutive/solid/PorousSolid.hpp
```diff
@@ -294,7 +296,7 @@ class PorousSolidUpdates : public CoupledSolidUpdates< SOLID_TYPE, BiotPorosity,
     real64 const bulkModulus = m_solidUpdate.getBulkModulus( k );
     real64 const thermalExpansionCoefficientTimesBulkModulus = thermalExpansionCoefficient * bulkModulus;
 
-    LvArray::tensorOps::symAddIdentity< 3 >( totalStress, -biotCoefficient * pressure - 3 * thermalExpansionCoefficientTimesBulkModulus * temperature );
+    LvArray::tensorOps::symAddIdentity< 3 >( totalStress, -biotCoefficient * pressure - 3 * thermalExpansionCoefficientTimesBulkModulus * deltaTemperature );
```
Guotong-Ren, line 299: it looks to me that pressure should take delta here as well?

src/coreComponents/constitutive/solid/PorousSolid.hpp
npillardou: The temperature_n represents the temperature at the previous converged step, while the referenceTemperature is the initial temperature.

Guotong-Ren: from previous conversations in this stream, i thought initial temperature is going to phase out?

src/coreComponents/constitutive/solid/PorousSolid.hpp
npillardou: The formula for the total stress involving thermal effects is given by :

Capture d'écran 2025-04-23 160443

So I think we should have the pressure and the delta temperature.

src/coreComponents/constitutive/solid/PorousSolid.hpp
npillardou: Not phase out, just replaced, they would prefer a "reference" state for temperature instead of using the term "initial" from what I have understood

src/coreComponents/constitutive/solid/PorousSolid.hpp
Guotong-Ren: you are right. My default interpretation for formulation in geos is the last time step as the reference state. 
- I think a better approach is to create a field to store the old-time thermal induced stresses like the current treatment for strain. so that initial temperature will disappear at this level. 
- You probably noticed deltaPressure always assumes P^n+1 - P^n.  Can you maybe rephrase deltaTemperature to something hinting T^n+1 - T^init? (TBH, i prefer initial temperature instead of reference temperature. but it is not a big deal. @paveltomin )

jhuang2601:
@rrsettgast @CusiniM @castelletto1 @Guotong-Ren This PR is waiting for your review.

jhuang2601:
All comments have been resolved in this PR, which is pending for use for a while.
@rrsettgast @CusiniM @castelletto1 @frankfeifan @Guotong-Ren @paveltomin @herve-gross Could you please take a final look and get it merged?

dkachuma (approved):

jhuang2601:
Expected failures by setting reference state for temperature
```
=======================
Test failures
=======================
CasedThermoElasticWellbore_ImperfectInterfaces_smoke_01
CasedThermoElasticWellbore_smoke_01
CasedThermoElasticWellbore_smoke_04
ThermoPoroDruckerPrager_consolidation_smoke_01
ThermoPoroDruckerPrager_consolidation_smoke_02
ThermoPoroElastic_consolidation_smoke_fim_01
ThermoPoroElastic_consolidation_smoke_fim_02
ThermoPoroElastic_consolidation_smoke_sequential_01
ThermoPoroElastic_consolidation_smoke_sequential_02
ThermoPoroElastic_SinglePhase_ThermalDradient_smoke_fim_01
ThermoPoroElastic_SinglePhase_ThermalDradient_smoke_fim_02
ThermoPoroElastic_SinglePhase_ThermalDradient_smoke_sequential_01
ThermoPoroElastic_SinglePhase_ThermalDradient_smoke_sequential_02
ThermoPoroElastic_staircase_co2_smoke_01
ThermoPoroElastic_staircase_co2_smoke_04
ThermoPoroExtendedDruckerPrager_consolidation_smoke_01
ThermoPoroExtendedDruckerPrager_consolidation_smoke_02
ThermoPoroMCC_consolidation_smoke_01
ThermoPoroMCC_consolidation_smoke_02
```

PR #3570, opened 2025-02-25
OMP Packing Optimization

wrtobin:
It would probably be desirable to determine the block size using some architecture-based heuristic, but I'm uncertain the development effort would achieve worthwhile performance differentiation over some reasonable default.


untereiner:
Hi @wrtobin 
Is this PR somehow related to #2369 ?

CusiniM (commented):
The changes look good to me but have we concluded that we always want to use static scheduling for all packing operations on host?  Maybe we should have someone test this on a few different systems.

PR #3571, opened 2025-02-25
fix: do not allow negative pressure by default, except for hydrofrac/singlephase

paveltomin:


paveltomin:
diffs are because of default change:
```

  Error: /Problem/Solvers/flowSolver/allowNegativePressure/__values__

  	Scalar values of types int32 and int32 differ: 0, 1.

```

CusiniM (commented):
I don't see how you are allowing the hydrofrac solver to have negative pressures though.

paveltomin:
> I don't see how you are allowing the hydrofrac solver to have negative pressures though.

here https://github.com/GEOS-DEV/GEOS/blob/a9e7207a909bfced0a9d6219de79c1baeb93d24e/src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp#L181

CusiniM (approved):

paveltomin:
> > I don't see how you are allowing the hydrofrac solver to have negative pressures though.
> 
> here
> 
> https://github.com/GEOS-DEV/GEOS/blob/a9e7207a909bfced0a9d6219de79c1baeb93d24e/src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp#L181

added it here https://github.com/GEOS-DEV/GEOS/blob/842b173a4a7160791e16b74f7bc9fc943654a9aa/src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.cpp#L400
for all singlephase (needed for integrated tests to pass)

jhuang2601 (approved):
So negative pressure is allowed for single phase flow, including hydroFrac solver.

PR #3559, opened 2025-02-19
refactor: Refactoring raw logs to tables

arng40:
Some log refactoring with more clearer table :

- VTKUtilities.cpp : Load balancing
- CO2Solubility : Negative solubility values encountered
- DofManager.cpp : Fields

Also : 
- minor improvements to the some existing table
- fix duplication table on mesh partitioning #3572
- Improve some log on SpatialPartition.cpp

PR Description from #3572 :

```
The mesh partioning over ranks table is duplicated with the same values when he have more than one discretization in NumericalMethods .
For example :

<NumericalMethods>
    <FiniteElements>
      <FiniteElementSpace
        name="FE1"
        order="1"/>
    </FiniteElements>
    <FiniteVolume>
      <TwoPointFluxApproximation
        name="fluidTPFA"/>
    </FiniteVolume>
  </NumericalMethods>

```
# 

Load balancing
----
```
Number of nodes: 1930023
  Number of elems: 4177966
          Polygon:  676999
             C3D4: 1578378
             C3D5:  457793
             C3D8: 1464796
Load balancing:     min     avg     max
(element/rank): 1017053 1044491 1057642
```

```
-----------------------------------------------------------------------------
|                      Load balancing, element / rank                       |
-----------------------------------------------------------------------------
|   Element type   |  Total over ranks  |  minimum  |  average  |  maximum  |
-----------------------------------------------------------------------------
|  Polygon         |            676999  |   161830  |   175185  |   169249  |
|  C3D4            |           1578378  |   400263  |   645718  |   563844  |
|  C3D5            |            457793  |   482173  |   772067  |   678292  |
|  C3D8            |           1464796  |  1017053  |  1057642  |  1044491  |
|  total elements  |          11718889  |  3911210  |  4535842  |  4385899  |
-----------------------------------------------------------------------------
```
Negative solubility values encountered
----
```
  Pressure (Pa) Temperature (C) CO2 solubility (mol/kg) H2O solubility (mol/kg) 
         100000             120             -0.00995285                       0 
         100000             122              -0.0111554                       0 
         200000             122             -0.00122161                       0 
         100000             116              -0.0076936                       0 
         100000             118             -0.00879985                       0
   ```
   
```
-----------------------------------------------------------------------------------------------
|  Pressure (Pa)  |  Temperature (C)  |  CO2 solubility (mol/kg)  |  H2O solubility (mol/kg)  |
-----------------------------------------------------------------------------------------------
[...]

PR #3572, opened 2025-02-26
fix: mesh partioning table duplication

arng40:
The `mesh partioning over ranks` table is duplicated with the same values when he have more than one discretization in `NumericalMethods` .
For example :
```
<NumericalMethods>
    <FiniteElements>
      <FiniteElementSpace
        name="FE1"
        order="1"/>
    </FiniteElements>
    <FiniteVolume>
      <TwoPointFluxApproximation
        name="fluidTPFA"/>
    </FiniteVolume>
  </NumericalMethods>
  ```


arng40:
@rrsettgast I've changed the if condition but I don't know if it's consistent with what you wrote

rrsettgast:
Only show the highest order FiniteElement Discretization.

arng40:
Adding fix in https://github.com/GEOS-DEV/GEOS/pull/3559

PR #3576, opened 2025-02-27
feat: add an option to skip density and viscosity computes when phase is not present for co2 fluid update

paveltomin:


src/coreComponents/unitTests/constitutiveTests/testPVT_CO2.txt
```diff
@@ -5,53 +5,53 @@
 # columns 5-6 = phase fractions
 # columns 7-8 = phase densities
 # columns 9-10 = phase viscosities
-0.0000e+00 1.0000e+06 3.5000e+02 1.5581e+01 1.0000e+00 0.0000e+00 1.5581e+01 1.0022e+03 1.7476e-05 4.1330e-04
```
dkachuma: I expected the viscosity to be zeroed as well. Is this passing simply because the values are too small and within the check tolerance?

dkachuma (approved):
Probably requires a rebaseline.

src/coreComponents/unitTests/constitutiveTests/testPVT_CO2.txt
paveltomin: maybe, i don't know much about those tests to be honest

paveltomin:
> Probably requires a rebaseline.

yes, rebaseline:
```
Error: /Problem/domain/Constitutive/fluid
	Group has a child 'checkPhasePresence' in the file to compare but not the baseline file.
```

PR #3581, opened 2025-03-04
refactor: Reorganise constitutive tests

dkachuma:
The directory `src/coreComponents/unitTests/constitutiveTests` is getting clattered with constitutive driver based tests. This PR moves all the driver based test xmls into a sub-directory `driverTests` and all their baselines to `baselines`. The idea is to keep `src/coreComponents/unitTests/constitutiveTests` for the actual c++ tests.


CusiniM:
I think it would be better to make the drivers standalone executables and  transform these tests into integratedTests.

jhuang2601 (approved):

dkachuma:
> I think it would be better to make the drivers standalone executables and transform these tests into integratedTests.

These tests are complete geos input files that can be run directly with the `geosx` executable. However I am having trouble adding them as integratedTests because there is no restart or time history - these don't really make sense for these tests. @CusiniM Can you please advise how I can add an integrated test without a restart check or a curve check.

paveltomin (approved):

inputFiles/constitutiveDrivers/scripts/pvtDriverPreprocess.py
```diff
@@ -0,0 +1,176 @@
...
+    the different properties.
+
+    Args:
+        xmlfile (str): Path to the input xml file
+        fileName (str): Path to the output hdf5 file. This will be overwritten
+
+    Returns:
+        geos.hdf5_wrapper: The hdf5 wrapper for the data
+    """
+
```
CusiniM, line 138: are you doing this so that you can then use a `curve_check` ? This is good for now. In the future we may want to simply output an hdf5 directly from the constitutiveDriver.

inputFiles/constitutiveDrivers/testPVT_docExample.xml
```diff
@@ -53,24 +53,4 @@
   </Events>
 
   <!-- SPHINX_PVTDRIVER_EVENTS_END -->
-  <!-- Mesh is not used, but GEOSX throws an error without one.  Will resolve soon-->
-  <Mesh>
```
CusiniM, line 57: were you able to resolver the need for the mesh? I don't see it in this PR but if that's the case, it's great!

herve-gross:
@dkachuma can you take a look at the conflicts?

PR #3582, opened 2025-03-06
feat: composite functions wo math presso

CusiniM:
- [x] Removes mathPresso from CompositeFunction
- [x] Adds simple operations between functions.

This is not yet fully portable but it would be fairly easy to make it evolve so that the operation can occur in a device loop. 

To be clear, MathPresso allowed for more sophisticated expressions but this should be enough for most of our cases and allows us to remove the dependency on mathPresso.


src/coreComponents/functions/unitTests/testFunctions.cpp
```diff
@@ -540,6 +541,155 @@ TEST( FunctionTests, 4DTable_derivatives )
...
+  {
+    set.insert( ii );
+  }
+
+  // Setup the random number generator
+  std::default_random_engine generator;
+  std::uniform_real_distribution< double > distribution( -0.99, 1.99 );
+  for( localIndex tt=0; tt<Ntimes; ++tt )
+  {
+    real64 const t = 1.0; //distribution( generator );
```
CusiniM:
```suggestion
    real64 const t = distribution( generator );
```

paveltomin:
@rrsettgast is this something useful?

PR #3584, opened 2025-03-06
fix: CI style and docs checks

bmhan12:
This PR:
- Get the style and documentation jobs actually running again.
  - These jobs were missing the `HOST_CONFIG` environment variable in Github Actions, so the CMake build could not find its dependencies.
- Fixes doxygen and uncrustify errors. 

- [x] Determine why these jobs were passing, but actually failing ([example](https://github.com/GEOS-DEV/GEOS/actions/runs/13702972819/job/38321240062))
  - Reason: `config-build.py` will always return an exit code of 0 on executing the CMake configuration line, regardless of if configuration is successful or not:  https://github.com/GEOS-DEV/GEOS/blob/5fe125be64c9d2978fdd2b6bdc471050fd85acf5/scripts/config-build.py#L233
  - Solution is to capture the return code and pass it up to `sys.exit(code)`.


CusiniM (approved):
Thanks

rrsettgast (approved):

PR #3587, opened 2025-03-07
feat: Perforation Status  

tjb-ltk:
1) Assign perforations an open/closed state from user specified time vs status table
2) Dt selection includes perforation table time
2) Calculations for inactive perforations will be skipped, but linear solver Awr and Arw arrays are still allocated. 
     - this could be fixed in different PR
3) Well element state introduced to allow closing/opening of segments
     - an element is open if it contains an open perforation, or an upstream element is open
     - Aww = I  for closed element entries
4) 4 integrated tests introduced to test vertical well where perforations are
     - plugged from bottom to top and top to bottom
     - opened from bottom to top and top to bottom


src/coreComponents/mesh/generators/WellGeneratorBase.cpp
```diff
@@ -482,6 +484,11 @@ void WellGeneratorBase::checkPerforationLocationsValidity()
 void WellGeneratorBase::mergePerforations( array1d< array1d< localIndex > > const & elemToPerfMap )
 {
 
+
+  for( globalIndex iwelem = 0; iwelem < m_numElems; ++iwelem )
+  {
+    GEOS_LOG_RANK( "BEFMERGE :" <<getName()<< ": " << iwelem << ": " << elemToPerfMap[iwelem] );
```
paveltomin: remove debug

src/coreComponents/mesh/generators/WellGeneratorBase.cpp
```diff
@@ -520,6 +527,10 @@ void WellGeneratorBase::mergePerforations( array1d< array1d< localIndex > > cons
       }
     }
   }
+  for( globalIndex iwelem = 0; iwelem < m_numElems; ++iwelem )
+  {
+    GEOS_LOG_RANK( "AFTMERGE :" <<getName()<< ": " << iwelem << ": " << elemToPerfMap[iwelem] );
```
paveltomin: remove debug

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
```diff
@@ -1210,6 +1215,38 @@ void CompositionalMultiphaseWell::assembleAccumulationTerms( real64 const & time
                                                        localMatrix,
                                                        localRhs );
         }
+        //wellControls.setWellOpen(false);
```
paveltomin: remove

paveltomin (approved):
don't forget to run code style fix

src/coreComponents/mesh/generators/WellGeneratorBase.cpp
tjb-ltk: done

src/coreComponents/mesh/generators/WellGeneratorBase.cpp
tjb-ltk: done

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp

src/coreComponents/mesh/Perforation.cpp
```diff
@@ -51,6 +52,13 @@ Perforation::Perforation( string const & name, Group * const parent )
     setRTTypeName( rtTypes::CustomTypes::groupNameRef ).
     setInputFlag( InputFlags::OPTIONAL ).
     setDescription( "Target region to connect the perforation" );
+
+  registerWrapper( viewKeyStruct::perfStatusTableString(), &m_perfStatusTable ).
```
dkachuma, line 65: Can you please add the change to `schema.xsd` that corresponds to this.

src/coreComponents/unitTests/wellsTests/testOpenClosePerf.cpp
```diff
@@ -0,0 +1,583 @@
...
+  {
+    WellSolverBase * wellSolverBase = solver->wellSolver();
+    wellSolverBase->setPerforationStatus( time, domain );
+  } );
+}
+
+int main( int argc, char * * argv )
+{
+  ::testing::InitGoogleTest( &argc, argv );
+  g_commandLineOptions = *geos::basicSetup( argc, argv );
```
dkachuma:
```suggestion
  geos::basicSetup( argc, argv );
```

dkachuma (approved):
Looks good to me.

I have a more paradigm question: We are asking the user to provide the perforation status as a 2D array and then we're using this to internally generate a table. Wouldn't it be simpler to just ask the user to provide the table? This would facilitate:
- Very long and complex open/shut sequences can be relegated to a text file instead of clattering the xml.
- If the user wants to control multiple perforations in the same way they can define this in one table.

src/coreComponents/mesh/Perforation.cpp
tjb-ltk: done

src/coreComponents/unitTests/wellsTests/testOpenClosePerf.cpp
tjb-ltk: done

tjb-ltk:
> Looks good to me.
> 
> I have a more paradigm question: We are asking the user to provide the perforation status as a 2D array and then we're using this to internally generate a table. Wouldn't it be simpler to just ask the user to provide the table? This would facilitate:
> 
> * Very long and complex open/shut sequences can be relegated to a text file instead of clattering the xml.
> * If the user wants to control multiple perforations in the same way they can define this in one table.

I'm not aware of the workflows used by the environment developers for GEOS... I assumed
1) when the perforation info for xml is being generated the perf status is also available.
2)  inlining the status info in the Perf object would avoid one level of indirection (the well model generates the table function, not user).   
3) the timing of open/close is/could be different for perfs. so one coordinate for all perfs isn't ideal and each perf needs its own table
 
Are you referring to something like below,  I can see it has advantages to allow using a separate file ,   which is fairly straightforward to make the changes...    

<InternalWell 
 name = "well1"
<Perforation
   name="perf1"
   statusTable="well1_perf1_table"
..
<Perforation
   name="perf2"
   statusTable="well1_perf2_table"
InternalWell/>

<Functions>
<TableFunction
      name="well1_perf1_table"
      coordinates= ...
      values=...   
<TableFunction
      name="well1_perf2_table"
      coordinates= ...
      values=...

src/coreComponents/mesh/Perforation.cpp
```diff
@@ -51,6 +52,13 @@ Perforation::Perforation( string const & name, Group * const parent )
...
+    setInputFlag( InputFlags::OPTIONAL ).
+    setSizedFromParent( 0 ).
+    setDescription( "Table defining perforation status as a function of time. If enterned in Functions section, the name must be the same as the perforation name" );
```
CusiniM: which name must be the same? the name of the function must match the name of the perforation? We should have an attribute to define the function associated to each perf so that multiple perfs can use the same function.

src/coreComponents/mesh/Perforation.cpp
```diff
@@ -59,8 +67,42 @@ void Perforation::postInputInitialization()
   GEOS_ERROR_IF( m_distanceFromHead <= 0,
                  getWrapperDataContext( viewKeyStruct::distanceFromHeadString() ) <<
                  ": distance from well head to perforation cannot be negative." );
-}
 
+  // Setup perforation status function
+  FunctionManager & functionManager = FunctionManager::getInstance();
+
+  if( !functionManager.hasGroup< TableFunction >( getName() ) )
```
CusiniM: here you could pass the name of the function instead of using the perf name and requiring the function name to be the same.

src/coreComponents/mesh/PerforationData.hpp
```diff
@@ -301,6 +337,12 @@ class PerforationData : public ObjectManagerBase
   /// Well skin factor at the perforations
   array1d< real64 > m_wellSkinFactor;
 
+  /// Name of perforations
+  string_array m_perfName;
+
+  /// Local perforation's status
+  array1d< localIndex > m_localPerfStatus;
```
CusiniM, line 361: what's the local perf status?

src/coreComponents/mesh/WellElementRegion.cpp
```diff
@@ -91,11 +91,11 @@ void WellElementRegion::generateWell( MeshLevel & mesh,
 
     if( perforationData->globalToLocalMap().count( iperfGlobal ) > 0 )
     {
-      elemStatusGlobal[iwelemGlobal] |= WellElementSubRegion::WellElemStatus::LOCAL;
+      elemStatusGlobal[iwelemGlobal] |= WellElementSubRegion::WellElemParallelStatus::LOCAL;
     }
     else
     {
-      elemStatusGlobal[iwelemGlobal] |= WellElementSubRegion::WellElemStatus::REMOTE;
+      elemStatusGlobal[iwelemGlobal] |= WellElementSubRegion::WellElemParallelStatus::REMOTE;
```
CusiniM, line 98: can you add some comments? I am not sure I follow what's happening? what's the meaning of remote and local in this context?

src/coreComponents/physicsSolvers/multiphysics/SinglePhaseReservoirAndWells.cpp
```diff
@@ -298,52 +301,55 @@ assembleCouplingTerms( real64 const time_n,
...
-                                                                              &localPerfJacobian[0][0] + 2 * i,
-                                                                              2 );
-            RAJA::atomicAdd( parallelDeviceAtomic{}, &localRhs[eqnRowIndices[i]], localPerf[i] );
+            if( eqnRowIndices[i] >= 0 && eqnRowIndices[i] < localMatrix.numRows() )
+            {
+              localMatrix.addToRowBinarySearchUnsorted< parallelDeviceAtomic >( eqnRowIndices[i],
+                                                                                &dofColIndices[0],
+                                                                                &localPerfJacobian[0][0] + 2 * i,
+                                                                                2 );
+              RAJA::atomicAdd( parallelDeviceAtomic{}, &localRhs[eqnRowIndices[i]], localPerf[i] );
```
CusiniM: Looking at this I would say that if the perf is closed there will be a zero on the diagonal but clearly I am missing where the diagonal is being set.

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
```diff
@@ -1212,10 +1217,40 @@ void CompositionalMultiphaseWell::assembleAccumulationTerms( real64 const & time
...
+        {
+          if( wellElemGhostRank[ei] < 0 )
+          {
+            if( elemStatus[ei]==WellElementSubRegion::WellElemStatus::CLOSED )
+            {
+              mixConnRate[ei] = 0.0;
+              globalIndex const dofIndex = wellElemDofNumber[ei];
+              localIndex const localRow = dofIndex - rank_offset;
+
+              real64 unity = 1.0;
```
CusiniM:
```suggestion
              real64 const unity = 1.0;
```

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
```diff
@@ -1212,10 +1217,40 @@ void CompositionalMultiphaseWell::assembleAccumulationTerms( real64 const & time
...
                                                        localRhs );
         }
+        // get the degrees of freedom and ghosting info
+        arrayView1d< globalIndex const > const & wellElemDofNumber =
+          subRegion.getReference< array1d< globalIndex > >( wellDofKey );
+        arrayView1d< integer const > const & wellElemGhostRank = subRegion.ghostRank();
+        arrayView1d< integer const > const elemStatus =subRegion.getLocalWellElementStatus();
+        array1d< real64 > & mixConnRate =  subRegion.getField< fields::well::mixtureConnectionRate >();
+        localIndex rank_offset = dofManager.rankOffset();
+        forAll< serialPolicy >( subRegion.size(), [=] GEOS_HOST_DEVICE ( localIndex const ei )
```
CusiniM: why `serialPolicy`? this forces it to happen on host.

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
```diff
@@ -1212,10 +1217,40 @@ void CompositionalMultiphaseWell::assembleAccumulationTerms( real64 const & time
...
+              real64 unity = 1.0;
+              for( integer i=0; i < m_numDofPerWellElement; i++ )
+              {
+                globalIndex const rindex =  localRow+i;
+                globalIndex const cindex =dofIndex + i;
+                localMatrix.template addToRow< serialAtomic >( rindex,
+                                                               &cindex,
+                                                               &unity,
+                                                               1 );
+                localRhs[rindex] = 0.0;
```
CusiniM, line 1294: @castelletto1 , @victorapm will this cause issues to the LS? I am guessing that ideally we would like to have a diagonal entry reasonably scaled with the rest of the diagonal entries in the system.

CusiniM (commented):
A few small questions and I think I think that the the way these global status arrays are used deserves some broader discussion to at least generalize this patter because it's different from anything else in geos. 
@rrsettgast should have a look at this before it goes in.

src/coreComponents/mesh/PerforationData.hpp
tjb-ltk: local means the mesh partition contains that perforation, status means open or closed.  Better comment added

src/coreComponents/mesh/WellElementRegion.cpp
tjb-ltk: These are defined in header where the struct is defined.  Several well elements data structures are mirrored on all cores.    Remote/local is used to identify if the element is spatially located on a different ranks mesh partition.

src/coreComponents/physicsSolvers/multiphysics/SinglePhaseReservoirAndWells.cpp
tjb-ltk: In the accumulation kernel

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
tjb-ltk: good eye... changed to parallel

src/coreComponents/mesh/Perforation.cpp
tjb-ltk: submitted changes to allow user defined perf status table names

src/coreComponents/mesh/Perforation.cpp
tjb-ltk: submitted changes to allow user defined perf status table names

src/coreComponents/mesh/Perforation.cpp
```diff
@@ -59,8 +73,59 @@ void Perforation::postInputInitialization()
...
                  ": distance from well head to perforation cannot be negative." );
-}
 
+  // Setup perforation status function
+  FunctionManager & functionManager = FunctionManager::getInstance();
+
+  if( m_perfStatusTableName.empty() )
+  {
+    // No table function provide, defaults name to perforation name
+    m_perfStatusTableName=getName();
```
MelReyCG: What do you think of a name that would be the well name + `_` + the perf name (to ensure unicity).
That would affect the doc line 60.

src/coreComponents/unitTests/wellsTests/testOpenClosePerf.cpp
```diff
@@ -0,0 +1,573 @@
...
+template< typename LAMBDA >
+void testOpenTopDownPerfCheck( CompositionalMultiphaseReservoirAndWells<> & solver,
+                               DomainPartition & domain,
+
+                               LAMBDA && perfFunction )
+{
+  CompositionalMultiphaseWell & wellSolver = *solver.wellSolver();
+
+  typedef std::map< real64, std::vector< int > > map_type;
+  map_type refVal;
```
MelReyCG: To make it clearer, what do you think of:
```suggestion
  map_type refPerfTable;
```

src/coreComponents/mesh/Perforation.cpp
```diff
@@ -51,6 +52,19 @@ Perforation::Perforation( string const & name, Group * const parent )
     setRTTypeName( rtTypes::CustomTypes::groupNameRef ).
     setInputFlag( InputFlags::OPTIONAL ).
     setDescription( "Target region to connect the perforation" );
+
+  registerWrapper( viewKeyStruct::perfStatusTableNameString(), &m_perfStatusTableName ).
+    setRTTypeName( rtTypes::CustomTypes::groupNameRef ).
+    setInputFlag( InputFlags::OPTIONAL ).
+    setDescription(
+    "Name of table function defining perforation status as a function of time. If none specified, a table function with be created internally and assigned the same name as the perforation." );
```
MelReyCG: Maybe add precisions / example:
```suggestion
    "Name of table function defining perforation status as a function of time:\n"
    "- The first row sets the perforation time coordinates,\n"
    "- The following row is a multiplier of the perforation rate (typically, 0.0 to close, 1.0 to open) for each time coordinate.\n"
    "If none specified, a table function with be created internally and assigned the same name as the perforation.\n"
    "A table that opens the perforation at t=100, and closes it at t=200 could be `{{ 0, 100, 200 },{ 0.0, 1.0, 0.0 }}`" );
```

src/coreComponents/mesh/Perforation.cpp
```diff
@@ -59,8 +73,59 @@ void Perforation::postInputInitialization()
...
+      real64 alwaysOpen = 1.0;
+      timeCoord[0].emplace_back( 0 );
+      values.emplace_back( alwaysOpen );
+    }
+    else
+    {
+      // User supplied table in Perforation section
+
+      GEOS_THROW_IF( m_perfStatusTable[0].size() != m_perfStatusTable[1].size(),
+                     GEOS_FMT( "Perforation status table missing time or status : {}", getName() ),
```
MelReyCG: To add the XML file line in the error:
```suggestion
                     GEOS_FMT( "{}: Perforation status table missing time or status.", getDataContext() ),
```

src/coreComponents/mesh/Perforation.cpp
```diff
@@ -59,8 +73,59 @@ void Perforation::postInputInitialization()
...
+  if( m_perfStatusTableName.empty() )
+  {
+    // No table function provide, defaults name to perforation name
+    m_perfStatusTableName=getName();
+  }
+  else
+  {
+    // Table name provide as input , check that it exists
+    GEOS_THROW_IF( !functionManager.hasGroup< TableFunction >( m_perfStatusTableName ),
+                   GEOS_FMT( "Perforation status table function missing  : {}", m_perfStatusTableName ),
```
MelReyCG: To add the XML file line in the error:
```suggestion
                   GEOS_FMT( "{}: Perforation status table function missing  : {}",
                             getDataContext(), m_perfStatusTableName ),
```

src/coreComponents/mesh/Perforation.cpp
tjb-ltk: Good idea ... done

src/coreComponents/mesh/Perforation.cpp
tjb-ltk: Thanks ... good suggestion I've added the description with a slight modification. The intent of perf status is to be an on/off indicator, with potentially other statuses as needed for well actions.   The fractional value I think you are referring to would be a PI multiplier field.

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
tjb-ltk: This is already the current behavior and not introduced by this PR.   If this needs to be changed it can be addressed in new PR  @CusiniM, @castelletto1 , @victorapm

herve-gross (approved):

src/coreComponents/mesh/PerforationData.hpp
paveltomin: maybe "local" is redundant here? we can't access not owned perfs anyway, right?

herve-gross:
@MelReyCG @jhuang2601 can you take a look?

inputFiles/compositionalMultiphaseWell/staged_close_perf_1.xml
```diff
@@ -0,0 +1,250 @@
...
+        type="producer"
+        control="BHP"
+        referenceElevation="0.5"
+        targetBHP="4e6"
+        targetPhaseRate="1e-3"
+        targetPhaseName="water"/>
+    </CompositionalMultiphaseWell>
+  </Solvers>
+
+  <Mesh>
```
jhuang2601, line 9: For all four tests, the only variation lies in the mesh section (perfStatusTable). I recommend creating a base XML file containing all common sections, and defining the smoke tests by simply modifying the mesh section. This approach aligns with the current setup for both smoke and benchmark tests.

src/coreComponents/mesh/Perforation.cpp
```diff
@@ -51,6 +52,22 @@ Perforation::Perforation( string const & name, Group * const parent )
...
+
+  registerWrapper( viewKeyStruct::perfStatusTableNameString(), &m_perfStatusTableName ).
+    setRTTypeName( rtTypes::CustomTypes::groupNameRef ).
+    setInputFlag( InputFlags::OPTIONAL ).
+    setDescription(
+    "Name of table function defining perforation status as a function of time:\n"
+    "- The first row sets the perforation time coordinates,\n"
+    "- The following row is the perforation status ( 0.0 for closed, 1.0 for opened) for each time coordinate.\n"
+    "If none specified, a table function with be created internally and assigned the name formed by well and perforation names, separated by '_'.\n"
+    "A table that opens the perforation at t=100, and closes it at t=200 could be `{{ 0, 100, 200 },{ 0.0, 1.0, 0.0 }}`" );
```
jhuang2601, line 64: Let's keep this for now. However, it would be better to create a basic tutorial case within the online documentation (ReadTheDocs) and include guidance there on how to define and use the perf table function.

jhuang2601: Only the `lower` Interpolation Methods is enforced, right?

src/coreComponents/schema/schema.xsd.other
```diff
@@ -1494,7 +1494,11 @@
 			<xsd:element name="MeshBodies" type="MeshBodiesType" />
 		</xsd:choice>
 		<!--Neighbors => (no description available)-->
+<<<<<<< HEAD
+		<xsd:attribute name="Neighbors" type="std_vector_lt_geos_NeighborCommunicator_cm_-std_allocator_lt_geos_NeighborCommunicator_gt_-_gt_" />
+=======
 		<xsd:attribute name="Neighbors" type="geos_internal_StdVectorWrapper_lt_geos_NeighborCommunicator_cm_-std_allocator_lt_geos_NeighborCommunicator_gt__cm_-false_gt_" />
+>>>>>>> origin/develop
```
jhuang2601: Please update this part after fixing conflict.

jhuang2601 (approved):

src/coreComponents/schema/schema.xsd.other
paveltomin: fixed

src/coreComponents/mesh/Perforation.cpp
tjb-ltk: Good point... the tables generated when the status arrays are defined with in the Perforation are enforced as lower,  but when the user adds them via a TableFunction, there was no validation , validation has been added.

inputFiles/compositionalMultiphaseWell/staged_close_perf_1.xml
tjb-ltk: done

PR #3596, opened 2025-03-12
fix: PVT Temperature Table Range Error Reporting

tjb-ltk:
1) Added unit conversion to error reporting
       2) Increase max allowed in temperature range


src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/PVTFunctionHelpers.hpp
```diff
@@ -204,13 +204,13 @@ initializePropertyTable( string_array const & inputParameters,
 
     real64 const TStart = units::convertKToC( stod( inputParameters[5] ) );
     real64 const TEnd = units::convertKToC( stod( inputParameters[6] ) );
-    real64 const dT = stod( inputParameters[7] );
+    real64 const dT = stod( units::convertKToC( inputParameters[7] ));
```
dkachuma:
```suggestion
    real64 const dT = stod( inputParameters[7] );
```
We don't need to convert the `dT` since it's a temperature difference which should be the same in C or K.

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/PVTFunctionHelpers.hpp
```diff
@@ -204,13 +204,13 @@ initializePropertyTable( string_array const & inputParameters,
...
 
     real64 const minT = 10;
-    real64 const maxT = 350;
-    GEOS_THROW_IF( TStart < minT, "Temperature must be in Kelvin and must be larger than " << minT << " K",
+    real64 const maxT = 77;
```
dkachuma: This value seems too small. Phillips goes up to 350C, Duan & Sun 260C, Span & Wagner 1100K (>800C), Garcia 200C. The limiting one is Spycher et al which goes only to 100C. I would still allow this to go to 200C.

dkachuma (changes requested):
No need to convert `dT`

dkachuma (approved):

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/PVTFunctionHelpers.hpp
tjb-ltk: done

paveltomin (approved):

PR #3597, opened 2025-03-13
feat: Soreide-Whitson equation of state

dkachuma:
Implements the Soreide-Whitson (SW) equation of state.
- This is a set of kernel callable functions to perform the calculation of the compressibility factor and the fugacity coefficients for the SW EoS with an interface similar to the cubic EoS.
- The correlations in the SW EoS depend on the component type. So a component type property is added to the component properties. The components are classified based on the name of the component with unidentified components being tagged as hydrocarbons.
- Implementations are extracted to an `_Impl.hpp` file.


src/coreComponents/constitutive/fluid/multifluid/compositional/functions/SoreideWhitsonEOSModel.hpp
OmarDuran, line 1: Class Naming: Based on the parent class structure, could you please rename SoreideWhitsonEOSModel to SoreideWhitsonEOSPhaseModel? Alternatively, consider renaming the parent class from CubicEOSPhaseModel.hpp to CubicEOSModel.hpp.

src/coreComponents/constitutive/fluid/multifluid/compositional/functions/SoreideWhitsonEOSModelImpl.hpp
OmarDuran, line 1: File Suffix Consistency: I noticed that there are 13 occurrences of implementation files with the suffix `_impl.hpp` compared to only 2 with the suffix `Impl.hpp`. Could you please rename SoreideWhitsonEOSModelImpl.hpp to SoreideWhitsonEOSModel_impl.hpp? If possible, also change CompositionalPropertiesImpl.hpp to CompositionalProperties_impl.hpp.

src/coreComponents/constitutive/fluid/multifluid/compositional/models/ComponentProperties.hpp
```diff
@@ -37,12 +37,19 @@ namespace compositional
...
-  {}
+  enum class ComponentType : integer
+  {
+    HydroCarbon,
+    Water,
+    Hydrogen,
+    HydrogenSulphide,
+    CarbonDioxide,
+    Nitrogen
+  };
```
OmarDuran: Enum File Organization: It's generally recommended in C++ to have separate header files for specific enum types. Could you please add two tests for valid and invalid component types to handle exceptions?

Suggestion for ComponentType.hpp

```
#ifndef COMPONENT_TYPE_HPP
#define COMPONENT_TYPE_HPP

enum class ComponentType : integer {
    HydroCarbon,
    Water,
    Hydrogen,
    HydrogenSulphide,
    CarbonDioxide,
    Nitrogen
};

/**
 * @brief Check if a component has a specific type
 * @param[in] type The type of the component
 * @param[in] targetType The target type to check
 * @return returns true if the type matches the target type
 */
GEOS_HOST_DEVICE
GEOS_FORCE_INLINE
static bool isType(integer const type, ComponentType const targetType) {
    return (type == static_cast<int>(targetType));
}

#endif
```

OmarDuran (changes requested):
Nice work! @dkachuma. I’ve added some comments and suggestions.

src/coreComponents/constitutive/fluid/multifluid/compositional/functions/SoreideWhitsonEOSModel.hpp
dkachuma: Changed to `SoreideWhitsonEOSPhaseModel` to be consistent with the cubic model.

src/coreComponents/constitutive/fluid/multifluid/compositional/models/ComponentProperties.hpp
dkachuma: Added a separate header file with its own unit test

src/coreComponents/constitutive/fluid/multifluid/compositional/models/ComponentType.hpp
```diff
@@ -0,0 +1,93 @@
...
+  HydroCarbon,
+  Water,
+  Hydrogen,
+  HydrogenSulphide,
+  CarbonDioxide,
+  Nitrogen
+};
+
+ENUM_STRINGS( ComponentType,
+              "",
```
paveltomin: how about "hc" here?

paveltomin (approved):
would be nice to add a test case (xml version)

dkachuma:
> would be nice to add a test case (xml version)

This is only part of the functionality. Not complete enough for an xml input file.

OmarDuran (approved):

src/coreComponents/constitutive/fluid/multifluid/compositional/models/ComponentType.hpp
dkachuma: Done

PR #3598, opened 2025-03-13
feat: Add a compositional Phillips brine density model

dkachuma:
Adds a multicomponent density and viscosity model based on the Phillips correlation.
- This is an extension of the 2-component model for the CO2-Brine fluid to multiple components.
- The idea is to use the density from the selected equation of state. However, the water component has a volume shift applied. This volume shift is tabulated as a function of pressure and temperature. It is precalculated in an initialisation step at selected (P,T) points.
- The viscosity is simply tabulated at the same (P,T) points from the Phillips correlation. The salinity in this case (unlike the CO2-Brine implementation) is folded into the table and the compute is simply a table lookup.

Currently this is not complete to add complete xml. The idea is that this will link with the Soreide-Whitson EoS and be added to a fluid model type which will be a compositional fluid model. This will be done in a subsequent PR. Something along the lines
```xml
<CompositionalTwoPhaseFluidPhillipsBrine
      name="FLUID"
      phaseNames="{ water, gas }"
      equationsOfState="{ sw, pr }"
      componentNames="{ CO2, N2, C1, C4, H2O }"
      componentMolarWeight="{ 44.0098e-3, 28.0135e-3, 16.0428e-3, 82.4191e-3, 18.01e-3 }"
      componentCriticalPressure="{ 73.773e5, 33.958e5, 45.992e5, 33.2071e5, 220.64e5  }"
      componentCriticalTemperature="{ 304.128, 126.192, 190.564, 504.216, 647.096 }"
      componentAcentricFactor="{ 0.223, 0.0372, 0.0104, 0.250274, 0.344 }"
      componentMolarWeight="{ 44.0098e-3, 28.0135e-3, 16.0428e-3, 82.4191e-3, 18.01e-3 }"
      waterCompressibility="3.5E-10"
      salinity="1.7"
      pressureCoordinates="{1.0e5, 1.86e5, 3.46e5, 6.45e5, 1.2e6, 2.2e6, 4.16e6, 7.74e6, 1.44e7, 2.68e7, 5.0e7}"
      temperatureCoordinates="{288.15, 293.15, 298.15, 303.15, 308.15, 313.15, 323.15, 333.15, 343.15, 353.15}" />
```
The `pressureCoordinates` and `temperatureCoordinates` will explicitly list the values at which the user wants the tables evaluated. In the current CO2-Brine model, these values are declared in external text files and are always uniformly spaced. For the pressure however, you usually want more values at the lower end because they are log-like functions.


src/coreComponents/constitutive/fluid/multifluid/compositional/models/PressureTemperatureCoordinates.hpp
OmarDuran, line 1: Renaming for Precision: The method IsIncreasing should be renamed to IsStrictlyIncreasing to accurately reflect its behavior, since the condition does not allow for epsilon-repeated values in the sequence.

src/coreComponents/constitutive/fluid/multifluid/compositional/models/PressureTemperatureCoordinates.cpp
```diff
@@ -0,0 +1,123 @@
...
+}
+
+bool PressureTemperatureCoordinates::isIncreasing( arraySlice1d< real64 const > const & array )
+{
+  localIndex const size = array.size();
+  GEOS_ASSERT( 1 < size );
+  real64 constexpr epsilon = MultiFluidConstants::epsilon;
+  for( localIndex i = 1; i < size; ++i )
+  {
+    if( array[i] - array[i-1] < -epsilon )
```
OmarDuran: Inequality Expression: I suggest change the inequality from `array[i] - array[i-1] < -epsilon` to `array[i-1] - array[i] < epsilon`, It aligns better with the positivity of epsilon. While it's a matter of taste, it makes the condition easier to understand.

paveltomin (approved):
same, would be nice to have an xml test case

OmarDuran (approved):

src/coreComponents/constitutive/fluid/multifluid/compositional/models/PressureTemperatureCoordinates.cpp
dkachuma: Actually I think the minus sign here is stray. It should simply be `array[i] - array[i-1] < epsilon`. This ensures that values are strictly increasing.

OmarDuran: Since I tried `array[i-1] - array[i] < epsilon` and the test passed, it would be nice to add a unit test for `bool PressureTemperatureCoordinates::isIncreasing`

MelReyCG (approved):
(only approved for the `Units.hpp` part, as you guys reviewed the rest of the PR twice)

src/coreComponents/constitutive/fluid/multifluid/compositional/models/PressureTemperatureCoordinates.cpp
dkachuma: I want to keep this method protected so I have added unit tests for the parameters. This will test the function indirectly. See [testPressureTemperatureCoordinates.cpp:120-127](https://github.com/GEOS-DEV/GEOS/blob/642048c72e3c4a008c7f51c60984ba57bc51557b/src/coreComponents/constitutive/unitTests/testPressureTemperatureCoordinates.cpp#L120-L127)

PR #3601, opened 2025-03-14
feat: SinglePhaseReactiveTransport implementation

frankfeifan:
This branch aims to implement the single-phase reactive transport model into GEOS. The following tasks are listed for completion:

- [x] A `SinglePhaseReactiveTransport` solver (derived from `SinglePhaseBase`) that contains all needed functions to solve the problem and `SinglePhaseReactiveTransportFields.hpp` to store all necessary fields 
- [x] `AccumulationKernel` and `FluxComputeKernel` to assemble the local rhs and matrix
- [x] Make everything work for both isothermal and thermal
- [ ] Plug in the chemistry module from this [repo](https://github.com/GEOS-DEV/HPCReact) through a `ReactiveSingleFluid` constitutive model 
- [ ] A porosity model that updates the solid porosity with chemical reaction


frankfeifan:
I closed this PR as everything should now be integrated into #3634

PR #3606, opened 2025-03-20
feat: Create Soreide-Whitson fluid model

dkachuma:
Adds the Soreide-Whitson fluid model
- Creates a FlashData object to hold parameters required for the flash. The point of this is that as we have different EoS and different flashes, they may need different parameters (e.g. salinity) and this object encapsulates all the parameters.
- Adds the Soreide-Whitson equation of state to the enumeration. Adds it to `FugacityCalculator` and `CompositionalDensityUpdate`.
- Fixes a typo in the CO2-H2O SW binary interaction coefficient.
- Creates a two phase fluid model (`CompositionalTwoPhasePhillipsBrine`) with a Phillips brine as a liquid phase and a compositional fluid as a gas phase.


src/coreComponents/constitutive/fluid/multifluid/compositional/parameters/EquationOfState.hpp
```diff
@@ -37,12 +37,14 @@ namespace compositional
...
-  SoaveRedlichKwong
+  SoaveRedlichKwong,
+  SoreideWhitson
 };
 
 ENUM_STRINGS( EquationOfStateType,
               "pr",
-              "srk" );
+              "srk",
+              "sw" );
```
paveltomin: maybe full names here? closer to general GEOS style

src/coreComponents/constitutive/fluid/multifluid/compositional/models/CompositionalDensity.hpp
```diff
@@ -194,6 +198,18 @@ void CompositionalDensityUpdate::computeCompressibilityFactor( integer const num
                                   compressibilityFactor,
                                   compressibilityFactorDerivs );
   }
+  else if( equationOfState == EquationOfStateType::SoreideWhitson )
+  {
+    SoreideWhitsonEOSPhaseModel< PengRobinsonEOS >::
```
paveltomin, line 203: no typo here?

src/coreComponents/constitutive/fluid/multifluid/compositional/functions/FugacityCalculator.hpp
```diff
@@ -108,6 +115,17 @@ void FugacityCalculator::computeLogFugacity( integer const numComps,
                                     componentProperties,
                                     logFugacity );
   }
+  else if( equationOfState == EquationOfStateType::SoreideWhitson )
+  {
+    SoreideWhitsonEOSPhaseModel< PengRobinsonEOS >::
```
paveltomin, line 120: and here

paveltomin:
when `getBinaryInteractionCoefficient` is called - can the logic use the symmetry of the matrix?

paveltomin (approved):

src/coreComponents/constitutive/fluid/multifluid/compositional/functions/SoreideWhitsonEOSPhaseModel_impl.hpp
```diff
@@ -243,12 +280,15 @@ computeMixtureCoefficients( integer const numComps,
...
   {
-    for( integer ic = 0; ic < numComps; ++ic )
+    real64 const aj = sqrt( aPureCoefficient[jc] );
+    aMixtureCoefficient += composition[jc] * composition[jc] * aj * aj;
+    for( integer ic = jc+1; ic < numComps; ++ic )
     {
+      real64 const ai = sqrt( aPureCoefficient[ic] );
       real64 kij = 0.0;
       real64 dkij_dT = 0.0;
       getBinaryInteractionCiefficient( pressure, temperature, componentProperties, salinity, ic, jc, kij, dkij_dT );
```
paveltomin: getBinaryInteractionCoefficient

src/coreComponents/constitutive/fluid/multifluid/compositional/models/CompositionalDensity.hpp
```diff
@@ -194,6 +198,18 @@ void CompositionalDensityUpdate::computeCompressibilityFactor( integer const num
...
+    computeCompressibilityFactor( numComps,
+                                  pressure,
+                                  temperature,
+                                  composition,
+                                  componentProperties,
+                                  salinity,
+                                  compressibilityFactor,
+                                  compressibilityFactorDerivs );
```
paveltomin: would it make sense to rename functions that also deal with derivs into `...andDerivs`?
like here `computeCompressibilityFactorAndDerivs`

src/coreComponents/constitutive/fluid/multifluid/compositional/models/NegativeTwoPhaseFlashModel.hpp
```diff
@@ -78,7 +80,8 @@ class NegativeTwoPhaseFlashModelUpdate final : public FunctionBaseUpdate
                                                          temperature,
                                                          compFraction,
                                                          componentProperties,
-                                                         m_liquidEos,
+                                                         m_flashData.liquidEos,
+                                                         m_flashData,
                                                          tangentPlaneDistance,
                                                          kValues[0] );
     GEOS_ERROR_IF( !stabilityStatus,
```
paveltomin: error or warning?

paveltomin: also - worth to output composition too

src/coreComponents/constitutive/fluid/multifluid/compositional/models/CompositionalDensity.hpp
dkachuma: The SW model is based on PR. But the implementation is more generic where you can also base it on SRK if required. But the SRK template instantiation isn't used.

src/coreComponents/constitutive/fluid/multifluid/compositional/models/NegativeTwoPhaseFlashModel.hpp
dkachuma: This is in a kernel callable function. Issuing a warning here might cause all sorts of issues one of which is too many warnings in the log. (Is there a way to limit a certain warning to only so many times?). But we can carry on with the flash even if the stability test fails to converge. I will change it to that i.e. if stability test fails to converge, assume the mixture is unstable and proceed to flash.

MelReyCG (approved):
A quick review to help, as I am code owner of one folder.
Just a remark, about the error you discussed before with @paveltomin: as I think we do not want errors (`printf()`) in kernels for performance reasons, should we output the `stabilityStatus` and do some output from the host when an error happens?

jhuang2601 (approved):

PR #3588, opened 2025-03-08
ci: oduran/xml_validation

OmarDuran:
Simplified Validation Bash Script

Proposed Changes:
This PR simplifies the validation bash script along with some minor improvements on documentation for its usage.

Considerations:
- Calls to Git for inferring a list of XML files for validation have been removed.
- The Git repository directory is now explicitly passed as input to the validation script.
- The script specifically validates files located in the 'examples' and 'inputfiles' directories.
- Python scripting is discarded as an option, as relying on a specific Python version for the script is not ideal.

Outcome:

- A simpler bash script that is easier to maintain and use.
- The process will no longer fail silently; if a file doesn't validate, the error will be clearly reported.

PR #3608, opened 2025-03-21
fix: Keep XML validation log in build directory

dkachuma:
The xml validation log ends up with the sources instead of being in the build directory.

Follow up to #3588 with the change at [validateXMLFiles.bash:49](https://github.com/GEOS-DEV/GEOS/blob/c9886616ac7772eb43fdfdf98b81565d5a6e7bc3/scripts/validateXMLFiles.bash#L49)

### DIFF
--- a/scripts/validateXMLFiles.bash
+++ b/scripts/validateXMLFiles.bash
@@ -7,7 +7,7 @@ if [ -z "$1" ]; then
 fi
 
 SCHEMA=$1; shift
-LOGFILE=xml_validation_results.log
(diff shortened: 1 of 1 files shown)


OmarDuran:
nice catch @dkachuma.

OmarDuran (approved):

CusiniM (approved):

CusiniM:
this pr can only be merged with a review by @rrsettgast

rrsettgast (approved):

PR #3609, opened 2025-03-25
refactor: Refactor/cleanup partitions

rrsettgast:
This PR seeks to remove a redundant definition of `DomainPartition::m_neighbors` and use the one that exists in `PartitionBase`


src/coreComponents/mesh/DomainPartition.cpp
```diff
@@ -104,57 +130,57 @@ void DomainPartition::setupBaseLevelMeshGlobalInfo()
...
-  for( NeighborCommunicator const & neighbor : m_neighbors )
+  for( NeighborCommunicator const & neighbor : neighbors )
   {
     secondNeighborRanks.erase( neighbor.neighborRank() );
   }
 
   for( integer const neighborRank : secondNeighborRanks )
   {
-    m_neighbors.emplace_back( neighborRank );
+    neighbors.emplace_back( neighborRank );
```
bd713, line 183: We used to have two versions of `m_neighbors`: one in `DomainPartition`, one in `SpatialPartition`.
Despite sharing the same name, I have the impression that:
* the `SpatialPartition` version used to be a "strict" list of neighbors.
* the `DomainPartition` version also appended the neighbors of neighbors.

After this refactoring, I understand only one version will remain. It will be stored in `SpatialPartition`, and will now include the neighbors of neighbors.

If that's correct, this might conflict notably with the Particles machinery (e.g. `SpatialPartition::repartitionMasterParticles`) which was relying so far on the "strict" definition of neighbors.

Best case scenario: the MPM will simply iterate over more neighbors for nothing.

NB: the CI tests for MPM may still run fine as they seem to use a 1x1x1 or 2x2x1 partition... meaning they won't generate neighbors of neighbors anyway.

PR #3610, opened 2025-03-26
fix: error in MappedVector::insert 

rrsettgast:
MappedVector::insert incorrectly delete object if it is a duplicate entry and the overwrite flag isn't true. This means that the insert is a no-op, and the object that was supposed to be inserted was deleted. Bad news bears.

- [x] fix incorrect deletion of objects referred to by input pointer
- [x] removed and disallowed redundant registrations, but provided a registration option to permit existence. 
- [x] removed the redundant `name` parameter from certain `registerGroup` and `registerWrapper` methods.

Related to: https://github.com/GEOS-DEV/LvArray/pull/350


rrsettgast:
@wrtobin I made the redundant registration attempts an error if the registration doesn't have the same exact pointer address as the existing entry. This is the only case that is OK imo. There are a lot of errors. Things seem to be registered in the mesh generation/ mesh level creation in a quite inconsistent way. No big surprise. Can you have a go at resolving some the errors?

src/coreComponents/mesh/MeshLevel.cpp
```diff
@@ -515,7 +515,8 @@ void MeshLevel::generateSets()
     {
       arrayView1d< bool const > const nodeInCurSet = nodeInSet[setName];
 
-      SortedArray< localIndex > & targetSet = elementSets.registerWrapper< SortedArray< localIndex > >( setName ).reference();
+//      SortedArray< localIndex > & targetSet = elementSets.registerWrapper< SortedArray< localIndex > >( setName ).reference();
```
CusiniM:
```suggestion
```

src/coreComponents/mesh/MeshLevel.cpp
```diff
@@ -515,7 +515,8 @@ void MeshLevel::generateSets()
...
+      SortedArray< localIndex > & targetSet = elementSets.getReference< SortedArray< localIndex > >( setName );
```
CusiniM, line 505: don't we also have a `getWrapper`?

src/coreComponents/mesh/MeshManager.cpp
```diff
@@ -64,8 +64,8 @@ void MeshManager::generateMeshes( DomainPartition & domain )
 {
   forSubGroups< MeshGeneratorBase >( [&]( MeshGeneratorBase & meshGen )
   {
-    MeshBody & meshBody = domain.getMeshBodies().registerGroup< MeshBody >( meshGen.getName() );
-    meshBody.createMeshLevel( 0 );
+    MeshBody & meshBody = domain.getMeshBodies().getGroup< MeshBody >( meshGen.getName() );
+//    meshBody.createMeshLevel( 0 );
```
CusiniM: ?

src/coreComponents/mesh/ObjectManagerBase.cpp
```diff
@@ -67,7 +67,7 @@ ObjectManagerBase::ObjectManagerBase( string const & name,
     setRestartFlags( RestartFlags::WRITE_AND_READ ).
     setPlotLevel( PlotLevel::NOPLOT );
 
-  m_sets.registerWrapper< SortedArray< localIndex > >( this->m_ObjectManagerBaseViewKeys.externalSet );
+//  m_sets.registerWrapper< SortedArray< localIndex > >( this->m_ObjectManagerBaseViewKeys.externalSet );
```
CusiniM:
```suggestion
```

CusiniM (approved):

src/coreComponents/dataRepository/MappedVector.hpp
```diff
@@ -489,42 +490,27 @@ T * MappedVector< T, T_PTR, KEY_TYPE, INDEX_TYPE >::insert( KEY_TYPE const & key
...
   {
     index = iterKeyLookup->second;
 
-    if( takeOwnership )
-    {
-      m_ownsValues[index] = true;
-    }
-
     // if value is empty, then move source into value slot
     if( m_values[index].second==nullptr )
```
corbett5: I'm curious, when do we use this functionality of inserting a `nullptr`? It seems like the kind of thing that could reasonably be disallowed.

corbett5 (approved):

src/coreComponents/dataRepository/MappedVector.hpp
rrsettgast: ....err....good point.

paveltomin:
@rrsettgast is this still active PR?

rrsettgast:
@wrtobin you were supposed to pick this up a while back.

corbett5:
I'm gonna try and get this working this week.

inputFiles/lagrangianContactMechanics/TFrac_smoke.xml
```diff
@@ -17,14 +17,6 @@
 <!-- SPHINX_SOLVER_END -->
 </Solvers>
 
-  <NumericalMethods>
```
corbett5, line 20: This is already present in the included base file.

inputFiles/wavePropagation/acous3D_Q3_abc_fs_smoke.xml
```diff
@@ -25,13 +25,4 @@
       xMin="{ -0.01, -0.01, 100.99 }"
       xMax="{ 101.01, 101.01, 101.01 }"/>
   </Geometry>
-
-    <!-- The free surface condition the domain -->
```
corbett5, line 29: This is already present in the included base file.

src/coreComponents/constitutive/ConstitutiveBase.cpp
```diff
@@ -53,7 +53,7 @@ void ConstitutiveBase::allocateConstitutiveData( Group & parent, localIndex cons
       if( wrapper.second->sizedFromParent() )
       {
         string const wrapperName = makeFieldName( this->getName(), wrapper.first );
-        parent.registerWrapper( wrapper.second->clone( wrapperName, parent ) ).
+        parent.registerWrapper( wrapper.second->clone( wrapperName, parent ), true ).
```
corbett5, line 56: These sorts of changes `registerWrapper(..., true)` and `registerGroup(..., true)` permit an object with the given name to already exist, and return the existing object.

src/coreComponents/constitutive/ConstitutiveManager.cpp
```diff
@@ -41,12 +41,14 @@ ConstitutiveManager::~ConstitutiveManager()
 {}
 
 
-Group * ConstitutiveManager::createChild( string const & childKey, string const & childName )
+Group * ConstitutiveManager::createChild( string const & childKey,
+                                          string const & childName,
+                                          bool const allowExistence )
```
corbett5, line 46: There were some places where `createChild` needed to change, since this is a virtual function in Group, changing the signature there required changes everywhere.

src/coreComponents/constitutive/ConstitutiveManager.cpp
```diff
@@ -41,12 +41,14 @@ ConstitutiveManager::~ConstitutiveManager()
...
 {
   GEOS_LOG_RANK_0( GEOS_FMT( "{}: adding {} {}", getName(), childKey, childName ) );
   std::unique_ptr< ConstitutiveBase > material =
     ConstitutiveBase::CatalogInterface::factory( childKey, getDataContext(), childName, this );
-  return &registerGroup< ConstitutiveBase >( childName, std::move( material ) );
+  return &registerGroup< ConstitutiveBase >( std::move( material ), allowExistence );
```
corbett5, line 51: I also made it so that methods such as these, which register an existing `Group` or `Wrapper` no longer require an additional name (`childName` in this example). In every instance in the code the provided name is the same as the name of the `Group` (or `Wrapper` itself, and it seems this is something we should enforce.

src/coreComponents/constitutive/fluid/multifluid/MultiFluidBase.cpp
```diff
@@ -37,16 +37,16 @@ MultiFluidBase::MultiFluidBase( string const & name, Group * const parent )
...
   registerWrapper( viewKeyStruct::componentMolarWeightString(), &m_componentMolarWeight ).
     setInputFlag( InputFlags::OPTIONAL ).
-    setDescription( "Component molar weights" );
+    setDescription( "Molar weights of components" );
 
   registerWrapper( viewKeyStruct::phaseNamesString(), &m_phaseNames ).
     setRTTypeName( rtTypes::CustomTypes::groupNameRefArray ).
     setInputFlag( InputFlags::OPTIONAL ).
-    setDescription( "List of fluid phases" );
+    setDescription( "List of fluid phases (e.g. gas, water, oil)" );
```
corbett5, line 49: These obviously weren't required, but I copied the slightly more detailed description from the duplicate registration in `InvariantImmiscibleFluid`.

src/coreComponents/dataRepository/MappedVector.hpp
```diff
@@ -36,33 +37,29 @@ namespace geos
...
 template< typename T,
-          typename T_PTR=T *,
           typename KEY_TYPE=string,
           typename INDEX_TYPE = int >
 class MappedVector
 {
 public:
-  static_assert( std::is_same< T_PTR, T * >::value || std::is_same< T_PTR, std::unique_ptr< T > >::value,
-                 "invalid second template argument for MappedVector<T,T_PTR,KEY_TYPE,INDEX_TYPE>. Allowable types are T * and std::unique_ptr<T>." );
-
```
corbett5, line 54: We were only using `MappedVector` with `T_PTR = T *`, never `unique_ptr< T >`, so I just removed that option. The functionality is equivalent anyways.

src/coreComponents/integrationTests/fluidFlowTests/testCompFlowUtils.hpp
```diff
@@ -80,25 +80,10 @@ void setupProblemFromXML( ProblemManager & problemManager, char const * const xm
...
-  constitutive::ConstitutiveManager & constitutiveManager = domain.getConstitutiveManager();
-  xmlWrapper::xmlNode topLevelNode = xmlProblemNode.child( constitutiveManager.getName().c_str());
-  constitutiveManager.processInputFileRecursive( xmlDocument, topLevelNode );
-
-  MeshManager & meshManager = problemManager.getGroup< MeshManager >( problemManager.groupKeys.meshManager );
-  meshManager.generateMeshLevels( domain );
-
-  ElementRegionManager & elementManager = domain.getMeshBody( 0 ).getBaseDiscretization().getElemManager();
-  topLevelNode = xmlProblemNode.child( elementManager.getName().c_str());
-  elementManager.processInputFileRecursive( xmlDocument, topLevelNode );
```
corbett5, line 100: This was just duplicating the contents of  `problemManager.parseXMLDocument( xmlDocument );`.

src/coreComponents/integrationTests/linearAlgebraTests/CMakeLists.txt
```diff
@@ -25,7 +25,8 @@ foreach(test ${LAI_tests})
 if ( ENABLE_MPI )
   geos_add_test( NAME ${test_name}
                  COMMAND ${test_name} -x ${nranks}
-                 NUM_MPI_TASKS ${nranks} )
+                 NUM_MPI_TASKS ${nranks}
+                 NUM_OMP_THREADS 1 )
```
corbett5, line 29: On Dane `testDofManager` and `testKrylovSolvers` both ran over 1500s and were killed by ctest, almost certainly due to threading issues with the algebra libraries and perhaps BLAS. With this change they run in seconds.

src/coreComponents/integrationTests/xmlTests/testXMLFile.cpp
```diff
@@ -31,9 +31,13 @@ using namespace geos;
 using namespace geos::dataRepository;
 using namespace xmlWrapper;
 
+std::unique_ptr< CommandLineOptions > globalCommandLine = nullptr;
+
 // Tests if the xml file parsing works with one file, one file with nested includes, and multiple files
 TEST( testXML, testXMLFile )
 {
+  GeosxState state( std::make_unique< CommandLineOptions >( *globalCommandLine ) );
```
corbett5, line 39: Having a `GeosxState` for each tests prevents duplicate registration from repeatedly reading in XML files. This is probably what we want to test anyways.

src/coreComponents/mainInterface/ProblemManager.cpp
```diff
@@ -351,69 +353,69 @@ void ProblemManager::setSchemaDeviations( xmlWrapper::xmlNode schemaRoot,
...
+  Group & includedFile = IncludedList.registerGroup< Group >( xmlWrapper::includedFileTag, true );
   includedFile.setInputFlags( InputFlags::OPTIONAL_NONUNIQUE );
   // the name of includedFile is actually a Path.
-  includedFile.registerWrapper< string >( "name" ).
+  includedFile.registerWrapper< string >( "name", nullptr, true ).
     setInputFlag( InputFlags::REQUIRED ).
     setRTTypeName( rtTypes::getTypeName( typeid( Path ) ) ).
     setDescription( "The relative file path." );
 
   schemaUtilities::SchemaConstruction( IncludedList, schemaRoot, targetChoiceNode, documentationType );
```
corbett5, line 367: Many duplicate registrations occurred during schema generation.

src/coreComponents/mesh/ElementRegionManager.cpp
```diff
@@ -268,7 +269,8 @@ void ElementRegionManager::buildSets( NodeManager const & nodeManager )
     {
       arrayView1d< bool const > const nodeInCurSet = nodeInSet[setName];
 
-      SortedArray< localIndex > & targetSet = elementSets.registerWrapper< SortedArray< localIndex > >( setName ).reference();
+      SortedArray< localIndex > & targetSet = elementSets.registerWrapper< SortedArray< localIndex > >( setName, true ).reference();
+      targetSet.clear();
```
corbett5: This `targetSet.clear()` is different from the previous behavior, but seems likely to be what we want.

src/coreComponents/mesh/MeshBody.cpp
```diff
@@ -35,21 +35,20 @@ MeshBody::MeshBody( string const & name,
 
 MeshLevel & MeshBody::createMeshLevel( localIndex const newLevel )
 {
-  return m_meshLevels.registerGroup< MeshLevel >( intToMeshLevelString( newLevel ) );
+  return m_meshLevels.registerGroup< MeshLevel >( intToMeshLevelString( newLevel ), true );
```
corbett5, line 38: Many of the duplicate registrations involved the mesh.

src/coreComponents/mesh/ObjectManagerBase.cpp
```diff
@@ -89,7 +89,9 @@ ObjectManagerBase::CatalogInterface::CatalogType & ObjectManagerBase::getCatalog
 
 SortedArray< localIndex > & ObjectManagerBase::createSet( const string & newSetName )
 {
-  return m_sets.registerWrapper< SortedArray< localIndex > >( newSetName ).reference();
+  auto & set = m_sets.registerWrapper< SortedArray< localIndex > >( newSetName, true ).reference();
+  set.clear();
+  return set;
```
corbett5, line 93: this `set.clear()` behavior is new.

src/coreComponents/mesh/ObjectManagerBase.hpp
```diff
@@ -574,7 +574,7 @@ class ObjectManagerBase : public dataRepository::Group
 
     m_registeredField.insert( FIELD_TRAIT::key());
 
-    return this->registerWrapper< typename FIELD_TRAIT::type >( FIELD_TRAIT::key() ).
+    return this->registerWrapper< typename FIELD_TRAIT::type >( FIELD_TRAIT::key(), nullptr, true ).
```
corbett5, line 577: The `nullptr` here is to satisfy the following signature without having to introduce a similar function without `rkey`

```c++
template< typename T, typename TBASE=T >
Wrapper< TBASE > & registerWrapper( string const & name,
                                    wrapperMap::KeyIndex::index_type * const rkey = nullptr,
                                    bool const allowExistence=false );
```

src/coreComponents/physicsSolvers/wavePropagation/sem/elastic/secondOrderEqn/isotropic/ElasticWaveEquationSEM.cpp
```diff
@@ -47,17 +47,17 @@ ElasticWaveEquationSEM::ElasticWaveEquationSEM( const std::string & name,
   WaveSolverBase( name,
                   parent )
 {
-  registerWrapper( viewKeyStruct::sourceConstantsString(), &m_sourceConstantsx ).
+  registerWrapper( viewKeyStruct::sourceConstantsXString(), &m_sourceConstantsx ).
```
corbett5, line 50: This is one of the true mistakes I found. The few failing integrated tests are due to this and will require a rebaseline.

src/coreComponents/physicsSolvers/multiphysics/PoromechanicsSolver.hpp
```diff
@@ -207,16 +207,16 @@ class PoromechanicsSolver : public CoupledSolver< FLOW_SOLVER, MECHANICS_SOLVER
                                                                 [&]( localIndex const,
                                                                      ElementSubRegionBase & subRegion )
       {
-        subRegion.registerWrapper< string >( viewKeyStruct::porousMaterialNamesString() ).
-          setPlotLevel( dataRepository::PlotLevel::NOPLOT ).
-          setRestartFlags( dataRepository::RestartFlags::NO_WRITE ).
-          setSizedFromParent( 0 );
```
corbett5, line 212: These must have been registered elsewhere, but I couldn't find where.

src/coreComponents/mesh/SurfaceElementRegion.cpp
```diff
@@ -220,7 +220,8 @@ localIndex SurfaceElementRegion::addToFractureMesh( real64 const time_np1,
   for( auto const & setIter : faceManager->sets().wrappers() )
   {
     SortedArrayView< localIndex const > const & faceSet = faceManager->sets().getReference< SortedArray< localIndex > >( setIter.first );
-    SortedArray< localIndex > & faceElementSet = subRegion.sets().registerWrapper< SortedArray< localIndex > >( setIter.first ).reference();
+    SortedArray< localIndex > & faceElementSet = subRegion.sets().getReference< SortedArray< localIndex > >( setIter.first );
+    // faceElementSet.clear();
```
corbett5: @rrsettgast For all the set building I initially cleared the set because I mistakenly assumed the old behavior of `registerWrapper` would replace the previously held data (effectively calling `clear`). This worked fine everywhere else, in fact I still kept that behavior. But here this presented a problem with some integrated tests having diffs in the the size of `FractureElementSubregion/sets/externalSet`, but only in parallel.

My question is: do we actually want to clear the sets every time?

corbett5:
I updated the error messages and conditions as requested. Here are two examples
```
***** ERROR
***** LOCATION: /usr/WS2/corbett5/GEOS/develop/src/coreComponents/dataRepository/Group.cpp:774
***** Rank 0: Tried registering a wrapper "pressure_n" of type real64_array with "/domain/MeshBodies/mesh/meshLevels/Level0/nodeManager"
A wrapper of this name already exists with type real32_array. The existing wrapper was registered by the following:
        acousticSolver

***** ERROR
***** LOCATION: /usr/WS2/corbett5/GEOS/develop/src/coreComponents/dataRepository/Group.cpp:801
***** Controlling expression (should be false): !allowExistence || (existingTypeId != typeId)
***** Rank 0: Tried registering a Group "acousticSolver" of type geos::AcousticWaveEquationSEM with "/Solvers"
A Group of this name already exists with type geos::dataRepository::Group
```

paveltomin (approved):

src/coreComponents/mesh/SurfaceElementRegion.cpp
rrsettgast: I don't know why this would be a registration in the first place. I would not think that we should clear the set unless it had been rendered invalid, which I don't think is the case here.

corbett5: Why it was originally a registration I don't know.

Alright, I'll remove the comment here and calls to `clear` elsewhere. My view though is that in these calls we are basically rebuilding the set from scratch, for consistencies sake it makes more sense to me that we first empty the set. Otherwise, as as evidenced by the error in `FractureElementSubregion/sets/externalSet` there is a disagreement in what is external between `FaceManager` and `FractureElementSubregion`.

src/coreComponents/dataRepository/MappedVector.hpp
```diff
@@ -295,15 +285,36 @@ class MappedVector
...
+   *   if @p keyName already existed.
    */
   T * insert( KEY_TYPE const & keyName,
-              T_PTR source,
-              bool takeOwnership,
-              bool overwrite = false );
+              T * const source,
+              bool const takeOwnership )
+  {
+    GEOS_ASSERT_MSG( source, "Trying to insert a nullptr with name: " << keyName );
```
jafranc, line 295: Does this survive RELEASE ?

src/coreComponents/dataRepository/MappedVector.hpp
```diff
@@ -295,15 +285,36 @@ class MappedVector
...
-   * @return pointer to the object that is held in the MappedVector
+   * @return pointer to the object that is held in the MappedVector, or a nullptr
+   *   if @p keyName already existed.
    */
   T * insert( KEY_TYPE const & keyName,
-              T_PTR source,
-              bool takeOwnership,
-              bool overwrite = false );
+              T * const source,
+              bool const takeOwnership )
```
jafranc, line 293: A bit worried to settle ownership via bool flag in a public interface.

src/coreComponents/dataRepository/MappedVector.hpp
MelReyCG: Seen the work done in this method, I would use `GEOS_ERROR_IF` here.

src/coreComponents/dataRepository/Group.cpp
```diff
@@ -749,5 +740,67 @@ stdVector< string > Group::getWrappersNames() const
...
+      }
+
+      GEOS_ERROR( "Tried registering a wrapper \"" << name << "\" of type " <<  typeName <<
+                  " with \"" << getPath() << "\"\n" <<
+                  "A wrapper of this name already exists with type " << existingWrapper.getRTTypeName() << registrationString );
+    }
+  }
+}
+
+void Group::insertGroup( Group * const newObject, bool const takeOwnership, bool const allowExistence )
```
MelReyCG, line 781: same remark

src/coreComponents/dataRepository/MappedVector.hpp
MelReyCG, line 293: What about providing two version of this method as public?
- one would take a moving reference (`T&&`) and would take ownership,
- the other one would take a pointer, and would never take ownership,
- internally, a private version of `insert()` can have the `takeOwnership` flag to avoid rewriting the same code, it would be called by the two public versions.

PR #3537, opened 2025-02-06
feat: Adding possibility to output all TableFunction in log/CSV + Solving a bug on PVT CSV export

MelReyCG:
This PR aims to:

1. Add the possibility to write `TableFunction`s in the log or in a CSV file through `logLevel="1"` or `writeCSV="1"` (this feature has been requested by TE users),

2. Allow to create tables without layout specification (added a test to illustrate, the goal is to allow to create table without headers, with only data rows)

3. Slightly refactor the `TableFunction` code to remove any PVT concept from its class,

4. Slightly rewrite the `writeCSV` documentation to make it clearer,

5. Add the table coordinates/values min, max and unit metadata at the top of the table,
image

6. For 2D tables, used standard variables to produce narrower output lines (and ensured that the first coordinate is on the X axis),
image

7. Unify the CSV generation when a table is too big and move the warning message inside the table, while keep showing its metadata.
image

----

This PR also solves bugs:
1. on saving a PVT, `SinglePhaseWell` and `CompositionalMultiphaseWell` table to a CSV file when an output folder has not been created yet (lacking `makeDirsForPath()` call),

2. A crash on table generation

3. wrong behaviour of merged cell when they share the same column, and are wider than them:
```
----------------------------------------------------------------------------------------------------------------
|                                                 Lorem Ipsum                                                  |
----------------------------------------------------------------------------------------------------------------
|      A       |      B       |      C       |  X  |      A       |      B       |      C       |      C       |
----------------------------------------------------------------------------------------------------------------
|  0123456789  |  0123456789  |  0123456789  |  X  |  0123456789  |  0123456789  |  0123456789  |  0123456789  |
|  0123456789  |  012345678901234567890123456789  |  X  |  0123456789  |  0123456789  |  0123456789  |  0123456789  |
|  012345678901234567890123456789  |  0123456789  |  X  |  0123456789  |  0123456789  |  0123456789  |  0123456789  |
|  0123456789  |  0123456789  |  0123456789  |  X  |  0123456789  |  0123456789  |  0123456789  |  0123456789  |
|  0123456789  |  0123456789  |  0123456789  |  X  |  01234567890123456789012345678901234567890123456789  |  0123456789  |
|  0123456789  |  0123456789  |  0123456789  |  X  |  0123456789  |  01234567890123456789012345678901234567890123456789  |
|  0123456789  |  0123456789  |  0123456789  |  X  |  0123456789  |  0123456789  |  0123456789  |  0123456789  |
[...]

Issue #3223, opened 2024-07-12
[EPIC] Move forward to C++20


untereiner:
**What is the requested feature?**
I want to open discussions to move forward to c++20

**Is your request related to a specific problem?**
Below is a list of desirable features that are standardized in C++20. 
I also put the min (compiler/stdlib) versions where they are implemented

 |                                 | clang | gcc |
 |:----------:|:-------------:|:------:|
 |  standardized text format  (in replacement of fmt) | 17 | 13 |
 |  range library | 15 | 10 |
 |  concepts library | 13 | 10 |
 |  span | 7 | 10 |
 | litterals as template params | 12 | 9 |

I would like to discuss the feasibility of this change on the different architectures that geos is compiled on..

**Describe the solution you'd like**
n.a.

**Describe alternatives you've considered**
n.a.

**Additional context**
n.a.

CusiniM:
We are mostly ready for this. It's part of the reason why we started compiling with cuda12. The only oustanding issues/questions are:
- do all target platforms have cuda12 and compilers that are new enough for novel c++20 features? I am not sure how to get this answer.
   - TotalEnergies @herve-gross ?
   - LLNL yes
   - Chevron @paveltomin ?
- do the target AMD systems support c++20? I think the answer is yes but @wrtobin can probably confirm this. 
- the geos cuda12 builds on rockylinux have a compilation error that we have not been able to solve (see https://github.com/GEOS-DEV/GEOS/blob/a7ddc53638bc37ac1927904e0c1d4e4dad754cda/.github/workflows/ci_tests.yml#L297). This should be resolved before fully moving to c++20.
- we should upate our CI images to make sure they all have adequate compilers and cuda versions (this is the easiest part).

untereiner:
Thanks @CusiniM !
The question raised today in the team while implementing with string literals and templates.

For TTE maybe @Bubusch can also answer ?

paveltomin:
> * do all target platforms have cuda12 and compilers that are new enough for novel c++20 features?
>   * Chevron @paveltomin ?

CC @drmichaeltcvx
I think we do, I've been compiling with gcc 13 recently and it works fine but there are tons of warnings which would be nice to fix

Bubusch:
@CusiniM @untereiner

Not yet on Pangea 3, I just asked to install cuda12 and gcc12 so we can make the switch.
Pangea 4 is OK.

FYI @sframba @herve-gross

wrtobin:
[...]

PR #3614, opened 2025-03-28
feat: Add the table error managment

arng40:
**This PR goes after** [#3537](https://github.com/GEOS-DEV/GEOS/pull/3537)
 
The main goal of this PR is to not stop the simulation when an error appear during the lifetime of the table.
- Working also for the csv generation with the table.
- All the errors will be append to the table and will give an indication of the table's consistency.

Exemple with a table from SinglePhaseStatistics.cpp
```
----------------------------------------------------------------------------------------------------------
|                           SinglePhaseFlowStatistics, Channel (time 20000 s):                           |
|--------------------------------------------------------------------------------------------------------|
|             statistics             |         min          |       average        |         max         |
|------------------------------------|----------------------|----------------------|---------------------|
|                      Pressure[Pa]  |                 inf  |  512792.08446348715  |  6003671.560233745  |
|               Delta pressure [Pa]  |  -4367397.539716403  |                   /  |  6003671.560233745  |
|                   Temperature [K]  |                   0  |                   0  |                  0  |
|------------------------------------|-------------------------------------------------------------------|
|                        statistics  |                                                            value  |
|------------------------------------|-------------------------------------------------------------------|
|  Total dynamic pore volume [rm^3]  |                                                                   |
|------------------------------------|-------------------------------------------------------------------|
|             Total fluid mass [kg]  |                                                50038.50017705067  |
|--------------------------------------------------------------------------------------------------------|
|  Error : One or more data lines are not equal to the number of headers                                 |
|  Data can be missing/misaligned                                                                        |
|  Warning : Invalid values detected (nan/inf).                                                          |
----------------------------------------------------------------------------------------------------------
```


victorapm (approved):

paveltomin (approved):

src/coreComponents/common/format/table/TableFormatter.cpp
```diff
@@ -120,6 +119,12 @@ string TableCSVFormatter::dataToString( TableData const & tableData ) const
 template<>
 string TableCSVFormatter::toString< TableData >( TableData const & tableData ) const
 {
+  if( tableData.getErrorsList().hasErrors() )
+  {
+    std::vector< string > const & errors = tableData.getErrorsList().errorText;
+    string const consoleWarning = std::accumulate( errors.begin(), errors.end(), std::string( "" ));
+    GEOS_WARNING( consoleWarning );
+  }
```
MelReyCG: 1. I don't think it catches all errors that happens within `headerToString()` & `dataToString()`. Did you test those warnings? Maybe it is too soon to show those messages, I suggest to do that in the destructor (because of my 3. point).
2. You should test in `testTable` that the expected errors are there.
3. In the special case in `TableFunction.cpp:376` where there is a summary table for a `TableCSVFormatter`, I think it would be better to add the error list to the `TableTextFormatter` and clear the `TableCSVFormatter`'s one to only show it once in the log.

MelReyCG (changes requested):

src/coreComponents/common/format/table/unitTests/testTable.cpp
```diff
@@ -670,6 +668,174 @@ TEST( testTable, testFreeLayout )
...
+      .setName( "CoordZ" )
+      .setHeaderAlignment( TableLayout::Alignment::left ),
+    TableLayout::Column()
+      .setName( "Prev\nelement" )
+      .setHeaderAlignment( TableLayout::Alignment::left ),
+    TableLayout::Column()
+      .setName( "Next\nelement" )
+      .setHeaderAlignment( TableLayout::Alignment::center )} );
+
+  LvArray::NumericLimits< float > const realLimit;
```
MelReyCG: I would use `real64` here (=`double`, which fits more to GEOS)

src/coreComponents/common/format/table/TableTypes.hpp
```diff
@@ -38,6 +38,44 @@ enum class CellType : integer
   Hidden
 };
 
+class ErrorListing
```
MelReyCG: As you are in the `geos` namespace, what do you think of naming that `TableErrorListing`?

MelReyCG (changes requested):

src/coreComponents/common/format/table/unitTests/testTable.cpp
```diff
@@ -670,6 +668,174 @@ TEST( testTable, testFreeLayout )
              );
 }
 
+TEST( testTable, table2DMismatchingCoordValues )
+{
+  TableData2D tableData2D;
+
+  // 1. Prepare values
```
Melcx, line 675: Avoid numbering comments.

src/coreComponents/common/format/table/unitTests/testTable.cpp
```diff
@@ -670,6 +668,174 @@ TEST( testTable, testFreeLayout )
...
+               "\n"
+               "-----------------------------------------------------------------------------------------\n"
+               "|         T         |  Pressure = 0  |  Pressure = 1  |  Pressure = 2  |  Pressure = 3  |\n"
+               "|-------------------|----------------|----------------|----------------|----------------|\n"
+               "|  Temperature = 0  |             0  |             0  |             0  |             0  |\n"
+               "|  Temperature = 1  |            11  |             7  |             8  |             9  |\n"
+               "|  Temperature = 2  |            17  |            13  |            14  |            15  |\n"
+               "|  Temperature = 3  |            23  |            19  |            20  |            21  |\n"
+               "|---------------------------------------------------------------------------------------|\n"
+               "|  Warning : The number of values is insufficient for the number of columns * rows:     |\n"
```
Melcx, line 743: Why the message here is not *"Too much data for ..."*?

src/coreComponents/common/format/table/TableData.cpp
```diff
@@ -59,13 +101,31 @@ void TableData2D::collectTableValues( arraySlice1d< real64 const > dim0AxisCoord
...
+      m_errors->addError( GEOS_FMT( "Warning : The number of values is insufficient for the number of columns * rows:\n"
+                                    "Expected {} values, Found {} values", nRow * nCol, values.size() ) );
+      wellConstructValues.resizeDefault( nRow * nCol, 0 );
+    }
+    else
+    {
+      m_errors->addError( GEOS_FMT( "Warning : Too much data for the number of columns * rows:\n"
+                                    "Expected {} values, Found {} values\n"
+                                    "Data may be missaligned", nRow * nCol, values.size() ) );
+    }
```
Melcx, line 121: Suggestion of message here:
```suggestion
    if( values.size() < nRow * nCol )
    {
      m_errors->addError( GEOS_FMT( "Warning: Not enough for the number of columns & rows:\n"
                                    "  - Expected {} values ({} columns x {} rows),\n  - Found {} values",
                                    nRow * nCol, nCol, nRow, values.size() ) );
      wellConstructValues.resizeDefault( nRow * nCol, 0 );
    }
    else
    {
      m_errors->addError( GEOS_FMT( "Warning: Too much data for the number of columns & rows:\n"
                                    "  - Expected {} values ({} columns x {} rows),\n  - Found {} values",
                                    "Data may be missaligned",
                                    nRow * nCol, nCol, nRow, values.size() ) );
    }
```

Melcx (changes requested):
to be continued

src/coreComponents/common/format/table/unitTests/testTable.cpp
MelReyCG: Why the message here is not *"Too much data for ..."*?

src/coreComponents/common/format/table/unitTests/testTable.cpp
MelReyCG: Avoid numbering comments.

src/coreComponents/common/format/table/TableData.cpp
MelReyCG: Suggestion of message here:
```suggestion
    if( values.size() < nRow * nCol )
    {
      m_errors->addError( GEOS_FMT( "Warning: Not enough for the number of columns & rows:\n"
                                    "  - Expected {} values ({} columns x {} rows),\n  - Found {} values",
                                    nRow * nCol, nCol, nRow, values.size() ) );
      wellConstructValues.resizeDefault( nRow * nCol, 0 );
    }
    else
    {
      m_errors->addError( GEOS_FMT( "Warning: Too much data for the number of columns & rows:\n"
                                    "  - Expected {} values ({} columns x {} rows),\n  - Found {} values",
                                    "Data may be missaligned",
                                    nRow * nCol, nCol, nRow, values.size() ) );
    }
```

src/coreComponents/common/format/table/TableData.cpp
```diff
@@ -23,6 +23,45 @@
...
+  {
+    m_rows = other.m_rows;
+    *m_errors = *other.m_errors;
+  }
+  return *this;
+}
+
+bool TableData::operator<(TableData const & other) const
+{
+  return m_rows < other.m_rows;
```
MelReyCG, line 62: 😎

src/coreComponents/common/format/table/TableData.cpp
```diff
@@ -59,13 +101,31 @@ void TableData2D::collectTableValues( arraySlice1d< real64 const > dim0AxisCoord
   arraySlice1d< real64 const > columAxisCoordinates = columnMajorInputValues ? dim0AxisCoordinates : dim1AxisCoordinates;
   integer const nCol = columAxisCoordinates.size();
   integer const nRow = rowAxisCoordinates.size();
-  // TODO: 1. restore the table non-blocking error system. 2. add this assert 3. add any other error to it.
-  GEOS_ASSERT( nRow * nCol == values.size() );
+
+  array1d< real64 > wellConstructValues( values.size() );
```
MelReyCG:
```suggestion
  array1d< real64 > wellConstructedValues( values.size() );
```
or
```suggestion
  array1d< real64 > wellFormedValues( values.size() );
```

src/coreComponents/common/format/table/TableData.hpp
```diff
@@ -45,15 +59,17 @@ class TableData
     /// The cell value
     string value;
 
-    /**
-     * @brief Comparison operator for cell value
-     * @param other The cell data value to compare
-     * @return The comparison result
-     */
+    /// @cond DO_NOT_DOCUMENTreturn The comparison result
```
MelReyCG: is this line valid?

src/coreComponents/common/format/table/TableData.hpp
```diff
@@ -112,9 +128,26 @@ class TableData
     return getCellsData() == comparingTable.getCellsData();
   }
 
+  /**
+   * @brief Get all error messages
+   * @return The vector of error messages
```
MelReyCG: Could be something else than a vector (a `std::set`, a `std::map`, an LVArray...)
```suggestion
   * @return The list of error messages
```

src/coreComponents/common/format/table/TableTypes.hpp
```diff
@@ -38,6 +38,43 @@ enum class CellType : integer
...
+class TableErrorListing
+{
+
+public:
+  /// View on cell error grouping the cell information to display at the end of the table
+  /// Contain all the errors
+  std::vector< string > errorText;
+
+  /**
+   * @brief Add an error that will be display at the end of the table
```
MelReyCG, line 49:
```suggestion
   * @brief Add an error that will be displayed at the end of the table
```

src/coreComponents/common/format/table/TableTypes.hpp
```diff
@@ -38,6 +38,43 @@ enum class CellType : integer
   Hidden
 };
 
+class TableErrorListing
+{
+
+public:
+  /// View on cell error grouping the cell information to display at the end of the table
+  /// Contain all the errors
+  std::vector< string > errorText;
```
MelReyCG: - Why do you qualify it of a view? It seems more like the "real" list of the error messages (exemples of views: `string_view`, pointer, list of `string_view`)
- I think names like `errorMessages`/`errorsList` would emphasize the possibility to have more than one error stored (it is not a single message with multiple lines). The comment also means that only one error can be stored.
- If the `TableErrorListing` is responsible of the list managment, why is this `std::vector` public?
To keep the possibility of exposing the messages, you could have.
```suggestion
public:
  using Iterator = std::vector< string >::const_iterator;

  Iterator begin()
  { return errorText.begin(); }

  Iterator end()
  { return errorText.end(); }
private:
  std::vector< string > errorText;
```

src/coreComponents/common/format/table/TableFormatter.cpp
```diff
@@ -381,24 +487,57 @@ void TableTextFormatter::populateDataCellsLayout( PreparedTableLayout const & ta
...
+    errorCellsLayout.push_back(
+      {
+        std::vector< TableLayout::CellLayout >( nbCells,
+                                                TableLayout::CellLayout( CellType::MergeNext ) ),
+        1
+      } );
+
+    errorCellsLayout[errorCellsLayout.size() - 1].cells[nbCells - 1].setWidth( 0 );
+    errorCellsLayout[errorCellsLayout.size() - 1].cells[nbCells - 1].m_cellType = CellType::Value;
+    errorCellsLayout[errorCellsLayout.size() - 1].cells[nbCells - 1].getLines().emplace_back( error );
```
MelReyCG: Is there a reason for not using the `prepareLayout()` method? This code add the need for the non-const `CellLayout::getLines()`, which is dangerous.
Also, you can use `std::vector<T>::last()` if you want to (I saw you using `front()`).

src/coreComponents/common/format/table/TableFormatter.cpp
```diff
@@ -360,14 +465,15 @@ void TableTextFormatter::populateDataCellsLayout( PreparedTableLayout const & ta
...
 {
-  size_t const nbColumns = !inputDataValues.empty() ? inputDataValues[0].size() : 0;
+  size_t const nbColumns = !inputDataValues.empty() ? inputDataValues.front().size() : 0;
   dataCellsLayout = std::vector< CellLayoutRow >{
     inputDataValues.size(),
     {
-      std::vector< TableLayout::CellLayout >( nbColumns, TableLayout::CellLayout() ), // cells
-      0 // sublinesCount
+      std::vector< TableLayout::CellLayout >( nbColumns, TableLayout::CellLayout() ),   // cells
+      1   // sublinesCount
```
MelReyCG: I don't understand why you set `1` here.

src/coreComponents/common/format/table/TableLayout.cpp
```diff
@@ -311,15 +312,19 @@ void PreparedTableLayout::prepareLayoutRecusive( std::vector< TableLayout::Colum
...
+
+    if( !column.hasChild() )
+    {
+      ++m_totalLowermostColumnCount;
+      if( cellType!= CellType::Hidden )
       {
-        ++m_lowermostColumnCount;
+        ++m_visibleLowermostColumnCount;
       }
     }
```
MelReyCG: Is it a bugfix? Is it linked to the `0`->`1` change? (I don't see a commit message that talk about a bug or something, but the logic seems quite different)

MelReyCG (changes requested):
Review of the bottom of changes.

src/coreComponents/common/format/table/unitTests/testTable.cpp
arng40: removed

src/coreComponents/common/format/table/TableTypes.hpp
arng40: errorText should be private indeed, I will fix that

src/coreComponents/common/format/table/TableData.cpp
arng40: wellFormedValues

src/coreComponents/common/format/table/TableData.hpp
arng40: ups

src/coreComponents/common/format/table/unitTests/testTable.cpp
arng40: changed

src/coreComponents/common/format/table/TableData.hpp
arng40: changed

src/coreComponents/common/format/table/TableFormatter.cpp
arng40: changed

src/coreComponents/common/format/table/TableFormatter.cpp
arng40: reset to 0

src/coreComponents/common/format/table/TableLayout.cpp
arng40: update and simplify

arng40: I revert it for znd populateData & error ( the 0->1) , the refacto was a bit too much :)

arng40:
I have a bug related to arrayofarray.
@wrtobin has already resolved this bug in this pending PR:
https://github.com/GEOS-DEV/GEOS/issues/3223#issuecomment-2389019364
@rrsettgast @CusiniM @wrtobin @corbett5 
Could you apply this fix in LvArray

src/coreComponents/common/format/table/TableData.cpp
MelReyCG: So.. You don't want it ?

src/coreComponents/common/format/table/TableTypes.hpp
```diff
@@ -38,6 +38,71 @@ enum class CellType : integer
...
+
+  /**
+   * @return A const reference to the errors vector.
+   */
+  std::vector< string > const & getErrors() const
+  { return errorText; }
+
+private:
+  /// Contain all the errors  to display at the end of the table
+  std::vector< string > errorText;
```
MelReyCG:
```suggestion
  std::vector< string > m_errorList;
```

src/coreComponents/common/format/table/TableData.cpp
```diff
@@ -96,6 +158,11 @@ TableData2D::TableDataHolder TableData2D::buildTableData( string_view targetUnit
     tableData1D.headerNames.push_back( GEOS_FMT( columnFmt, columnValue ) );
   }
 
+  for( auto errorIt = m_errors->begin(); errorIt != m_errors->end(); errorIt++ )
```
MelReyCG: you should be able to keep a "for each"
```
  for( auto const & error : *m_errors )
```

src/coreComponents/common/format/table/TableFormatter.cpp
```diff
@@ -120,6 +201,12 @@ string TableCSVFormatter::dataToString( TableData const & tableData ) const
 template<>
 string TableCSVFormatter::toString< TableData >( TableData const & tableData ) const
 {
+  if( tableData.getErrorsList().hasErrors() )
+  {
+    std::vector< string > cpyErrors  = tableData.getErrorsList().getErrors();
+    getErrorsList().appendErrors( cpyErrors );
```
MelReyCG, line 204: - I don't see how you managed the point 3. of my previous message (https://github.com/GEOS-DEV/GEOS/pull/3614#discussion_r2031541167)

src/coreComponents/common/format/table/TableFormatter.cpp
```diff
@@ -479,7 +612,7 @@ void TableTextFormatter::stretchColumnsByMergedCellsWidth( stdVector< size_t > &
         bool const isToMergeWithLeft = ( columnId > 0 ) ? ( row.cells[columnId - 1].m_cellType == CellType::MergeNext ) : false;
         bool const isFirstToMerge = isToMergeWithRight && !isToMergeWithLeft;
         if( isFirstToMerge )
-        { // we detected a set of cells to merge, let's compute its total width
+        {   // we detected a set of cells to merge, let's compute its total width
```
MelReyCG: why did you added those spaces here and over other blocks?

src/coreComponents/common/format/table/TableFormatter.hpp
```diff
@@ -185,14 +263,36 @@ class TableTextFormatter : public TableFormatter
...
 
+  /**
+   * @brief Outputs the formatted table lines to the output stream.
+   * @param tableLayout The layout of the table
+   * @param cellsLayout A collection of rows, each containing a layout of cells to be processed and formatted.
+   * @param tableOutput A reference to an `std::ostringstream` where the formatted table will be written.
+   */
+  void outputLines( PreparedTableLayout const & tableLayout,
+                    CellLayoutRows const & cellsLayout,
+                    std::ostringstream & tableOutput ) const;
```
MelReyCG: I think you may want a non-specialized `std::ostream` here and everyone else in the methods parameters

MelReyCG (changes requested):

src/coreComponents/common/format/table/TableFormatter.cpp
arng40: 1 & 2 & 3 done
 Add a function `showErrors` in CSVFormatter

MelReyCG: nice

src/coreComponents/common/format/table/TableFormatter.hpp
```diff
@@ -104,13 +132,52 @@ class TableCSVFormatter : public TableFormatter
...
+
+  /**
+   * @brief Indicate if we print a warning if the table contains any errors By default, show any erros
+   * @param cond The boolean to turn on/off log errors
+   */
+  void showErrors( bool cond )
+  { m_showErrors = cond; }
+
+private:
+  /// Boolean indicating if we show any erros
```
MelReyCG:
```suggestion
  /// Boolean indicating if we show errors on destruction
```

src/coreComponents/common/format/table/TableFormatter.hpp
```diff
@@ -104,13 +132,52 @@ class TableCSVFormatter : public TableFormatter
...
+   */
+  template< typename DATASOURCE >
+  void toStream( std::ostream & outputStream, DATASOURCE const & tableData ) const
+  { toStreamImpl( outputStream, toString( tableData ) ); }
+
+  /**
+   * @brief Indicate if we print a warning if the table contains any errors By default, show any erros
+   * @param cond The boolean to turn on/off log errors
+   */
+  void showErrors( bool cond )
```
MelReyCG:
```suggestion
  /**
   * @brief Indicate if we print the encountered errors on destruction. Enabled by default.
   * @param cond The boolean to turn on/off log errors
   */
  void showErrors( bool cond )
```

src/coreComponents/functions/TableFunction.cpp
```diff
@@ -355,47 +355,49 @@ void TableFunction::outputTableData( OutputOptions const outputOpts ) const
...
-      tableData.addSeparator();
+    {
       tableData.addRow( GEOS_FMT( "CSV Generated to:\n{}/{}.csv", getOutputDirectory(), getName() ) );
       GEOS_LOG( tableLog.toString( tableData ) );
     }
+
+    std::ofstream logStream( joinPath( FunctionBase::getOutputDirectory(), getName() + ".csv" ) );
+    TableCSVFormatter csvFormatter;
+    csvFormatter.showErrors( false );
+    logStream << csvFormatter.toString( *this );
```
MelReyCG: I think the `TableCSVFormatter` should be output first:
- before the `TableTextFormatter`
- the `csvFormatter` object have a wrong visibility scope, the `TableTextFormatter` part could miss the possible CSV output error we could add later.

I know it can be helpful because I need that in a PR that verifies if a CSV has been successfuly opened. If you want you can leave a TODO here to reverse that CSV/Text part.

src/coreComponents/functions/TableFunction.cpp
```diff
@@ -355,48 +355,48 @@ void TableFunction::outputTableData( OutputOptions const outputOpts ) const
...
+  bool const logOutputFailed = isTableTooLargeForLog( *this );
+
+  TableLayout tableLayout( { TableLayout::Column().
+                               setName( getName() ).
+                               setHeaderAlignment( TableLayout::Alignment::center ).
+                               setValuesAlignment( TableLayout::Alignment::left ) } );
+  TableTextFormatter const tableLog( tableLayout );
+  TableData tableData;
+
+  if( outputOpts.writeCSV )
```
MelReyCG: So... you tell the user that too much data leads to saving it in a CSV, but do not generate the csv file?

src/coreComponents/functions/TableFunction.cpp
```diff
@@ -355,48 +355,48 @@ void TableFunction::outputTableData( OutputOptions const outputOpts ) const
   if( MpiWrapper::commRank() != 0 )
     return;
 
-  bool const logOutputFailed = outputOpts.writeInLog && isTableTooLargeForLog( *this );
-  if( outputOpts.writeInLog && !logOutputFailed )
-  { // log output
-    TableLayout tableLayout( getName(), {} );
-    if( outputOpts.writeCSV )
+  bool const logOutputFailed = isTableTooLargeForLog( *this );
```
MelReyCG: with your change, we now show an error about showing the table in the log even if the user only asked for the CSV output

MelReyCG (changes requested):

src/coreComponents/functions/TableFunction.cpp
MelReyCG: I think we can safely keep 
```suggestion
  bool const logOutputFailed = outputOpts.writeInLog && isTableTooLargeForLog( *this );
```

MelReyCG (approved):

paveltomin:
this still requires more reviews and there is a change request pending from Mr. @Melcx

Melcx (approved):

paveltomin:
@arng40 is `feature/rey/umpire-csv-stats` base branch for this PR?

arng40:
> Member

Some features in tableFormatter

paveltomin:
@rrsettgast, @untereiner, @wrtobin can you please review?

rrsettgast (approved):

PR #3615, opened 2025-03-31
fix: Log levels generation on documentation generation

arng40:
Fix a bug on LogLevelsRegistry where log levels were never added


src/coreComponents/dataRepository/LogLevelsRegistry.cpp
```diff
@@ -14,20 +14,23 @@
...
+
+  if( uniqueDescriptions.insert( description ).second )
   {
-    std::vector< string > const & targetValues = m_logLevelsDescriptions[condition];
-    if( std::find( targetValues.begin(), targetValues.end(), description ) != targetValues.end())
-    {
-      m_logLevelsDescriptions[condition].push_back( string( description ) );
-    }
+    targetValues.emplace_back( description );
   }
```
MelReyCG: - What happens if `insert()` is done more than one time?
 - What happens if `second` is false?

arng40: - I checked, and for multiple insert with the same description, I always get unique description
- uniqueDescriptions.insert return pair<iterator, bool> and if false : no insertion

MelReyCG: Ok, see next comment.

src/coreComponents/dataRepository/LogLevelsRegistry.cpp
```diff
@@ -14,20 +14,23 @@
...
 namespace geos
 {
 
 void LogLevelsRegistry::addEntry( integer condition, std::string_view description )
 {
-  if( m_logLevelsDescriptions.count( condition ) > 0 )
+
+  auto & targetValues = m_logLevelsDescriptions[condition];
+
+  std::unordered_set< string_view > uniqueDescriptions( targetValues.begin(), targetValues.end());
```
MelReyCG: Why creating a full copy of a `m_logLevelsDescriptions` entry?
I think a simple & more effective way exists to add an entry. Why do you need that branch?

arng40: right no need to use a set, i think I jsut could use std::find

MelReyCG: Indeed :)

MelReyCG:
To prevent this to happen again and to add some matter in this PR, could you add some unit test that verify the documentations of a given `Group` (that could be a dummy Group in within the unit test).

src/coreComponents/dataRepository/unitTests/testLogLevels.cpp
```diff
@@ -0,0 +1,108 @@
...
+// TPL includes
+#include <gtest/gtest.h>
+#include <gtest/gtest-spi.h>
+
+
+using namespace geos;
+using namespace dataRepository;
+
+
+TEST( testLogLevels, testLogLevel )
```
MelReyCG: Just a suggestion: I think you can just add your test to `testWrapper.cpp`

MelReyCG (approved):
Nice!

src/coreComponents/dataRepository/unitTests/testLogLevels.cpp
arng40: Renamed to TestDocumentationGeneration and it will be more scalable

MelReyCG (approved):

rrsettgast (approved):

PR #3617, opened 2025-03-31
fix: Fix runtime cuda error.

CusiniM:
Capture of member variable in device kernel.


jhuang2601 (approved):
Thanks for this fix.
With this branch, cases with iterative solver (`SneddonProblem_Contact` and `TFrac`) can be running in both serial and parallel on Maple and Pangea3 GPU now.

CusiniM:
@corbett5 @rrsettgast @MelReyCG @cssherman this one is ready and just needs a review.

src/coreComponents/fieldSpecification/TractionBoundaryCondition.cpp
```diff
@@ -244,6 +247,8 @@ void TractionBoundaryCondition::reinitScaleSet( FaceManager const & faceManager,
 
   m_scaleSet.resize( faceSize );
 
+  auto scaleSet = m_scaleSet.toView();
```
corbett5: Would be better if this was `auto const`, but it's not important.

corbett5 (approved):

src/coreComponents/fieldSpecification/TractionBoundaryCondition.cpp
CusiniM:
```suggestion
  auto const scaleSet = m_scaleSet.toView();
```

rrsettgast (approved):

PR #3620, opened 2025-04-03
feat: plug in HPCReact and create proper fluid model..

CusiniM:
- [x] add HPCReact as a submodule. 
 - [x] Remove old reactive fluid, reactions and driver. Everything in there  should be covered by HPCReact, including the testing that was done with the driver.
 - [ ] create a fluid model that interfaces with HPCReact.


paveltomin:
looks like a duplicate of https://github.com/GEOS-DEV/GEOS/pull/3634
(cc @rrsettgast @frankfeifan)

PR #3621, opened 2025-04-07
feat: reservoir volume well constraint

tjb-ltk:
If useSurfaceCondtions=”1” is set, user defined pressure and temperature are used to compute densities/volume needed to setup pressure/rate boundary conditions.  If this flag is set to 0, GEOS uses the top segment pressure.   The common expectation of a reservoir constraint is to control the well based on reservoir conditions.  
GEOS has been modified to allow a reservoir region name to be associated with the well control when modeling reservoir volume constraints.  The user should also note that the frequency of the region average pressure calculation is defined in Single/CompositionalMultiPhaseStatistics definition, and is not updated every Newton iteration, resulting a lagged fluid properties.  Setting cycleFrequency='1' will update the pressure every timestep, but other frequencies will result in larger lags.
If no region name is specified, the calculation defaults to original behavior and uses the top well element pressure


CusiniM:
If it's the standard so be it but I have a hard time understanding why the average region pressure and temperature would be better than using the well segment ones. As a matter of fact I would argue that one should always use upwinded. It may be that I don't fully understand "what reservoir volume constraints" mean though...

tjb-ltk:
In some reservoirs an objective is to maintain reservoir pressure by injecting the some fraction of the reservoir volumes produced.   This control gives an estimate of the reservoir volume produced.  Likewise to inject a certain reservoir volume a similar calculation is performed..  Using the top well element pressure is not representative of the pressure in the reservoir cells drained during production...

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
```diff
@@ -435,13 +436,30 @@ void CompositionalMultiphaseWell::validateInjectionStreams( WellElementSubRegion
...
+    bool useSeg = wellControls.referenceReservoirRegion()=="";
+    GEOS_WARNING_IF( useSeg,
+                     "WellControls referenceReservoirRegion not set and well constraint fluid property calculations will use top segement pressure and temp " );
+    if( useSeg )
+    {
+      wellControls.setRegionAveragePressure( -1 );
+    }
+    else
+    {
+      ElementRegionBase const & region = elemManager.getRegion( "wellControls.referenceReservoirRegion()" );
```
paveltomin: "wellControls.referenceReservoirRegion()" -> wellControls.referenceReservoirRegion() ?

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
```diff
@@ -435,13 +436,30 @@ void CompositionalMultiphaseWell::validateInjectionStreams( WellElementSubRegion
...
+      CompositionalMultiphaseStatistics::RegionStatistics const & stats = region.getReference< CompositionalMultiphaseStatistics::RegionStatistics >( "regionStatistics" );
+      wellControls.setRegionAveragePressure( stats.averagePressure );
```
paveltomin, line 465: using statistics might be tricky
statistics is computed with different frequency than solver, pretty much independent

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
```diff
@@ -735,15 +773,16 @@ void CompositionalMultiphaseWell::updateVolRatesForConstraint( WellElementSubReg
 
         //    We need to evaluate the density as follows:
         //      - Surface conditions: using the surface pressure provided by the user
-        //      - Reservoir conditions: using the pressure in the top element
+        //      - Segment conditions: using the pressure in the top element
+        //      - Reservoir conditions: using the average region pressure
```
paveltomin, line 786: should not it use pressure in perforated cell here when "Reservoir conditions" is used?

tjb-ltk: If the cell is large and source of most of produced volumes then probably ok, but would have to be weighted pressure for multi perfs.    If the cells are small and the drainage radius is further into the reservoir then an region average might be better, IMHO...       Using element pressure was probably for testing.   Currently that is the default unless a region name is entered (many integrated tests would fail).  I could add another option for using a weight perfed cell pressures...   but was going to wait for user to request

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
tjb-ltk: agreed, to mitigate this is stated in the keyword description for referenceReservoirRegion...  Not sure how to make the change so the regionstat code can be set when this option is activated...

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
tjb-ltk: I know...  the user case was single phase....not sure what planet I was on.... its fixed

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
paveltomin: yeah, also `"regionStatistics"` as name is hardcoded here

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
```diff
@@ -672,9 +690,29 @@ void CompositionalMultiphaseWell::updateVolRatesForConstraint( WellElementSubReg
...
+  {
+    // use surface conditions
+    flashPressure = wellControls.getSurfacePressure();
+    flashTemperature = wellControls.getSurfaceTemperature();
+  }
+  else
+  {
+    // use region conditions
+    flashPressure = wellControls.getRegionAveragePressure();
+    if( flashPressure < 0.0 )
```
paveltomin, line 713: this one is for when no statistics is available yet?

src/coreComponents/physicsSolvers/fluidFlow/wells/SinglePhaseWell.cpp
```diff
@@ -130,10 +131,28 @@ string SinglePhaseWell::resElementDofName() const
...
-                                               WellElementSubRegion const & subRegion )
+                                               WellElementSubRegion const & subRegion,
+                                               ElementRegionManager const & elemManager )
 {
-  WellControls const & wellControls = getWellControls( subRegion );
-  WellControls::Control const currentControl = wellControls.getControl();
+  WellControls & wellControls = getWellControls( subRegion );
+  if( !wellControls.useSurfaceConditions() )
+  {
+    bool useSeg = wellControls.referenceReservoirRegion()=="";
```
paveltomin: .empty() ?

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
```diff
@@ -435,13 +436,30 @@ void CompositionalMultiphaseWell::validateInjectionStreams( WellElementSubRegion
...
-                                                           WellElementSubRegion const & subRegion )
+                                                           WellElementSubRegion const & subRegion,
+                                                           ElementRegionManager const & elemManager )
 {
+  WellControls & wellControls = getWellControls( subRegion );
+  if( !wellControls.useSurfaceConditions() )
+  {
+    bool useSeg = wellControls.referenceReservoirRegion()=="";
+    GEOS_WARNING_IF( useSeg,
+                     "WellControls referenceReservoirRegion not set and well constraint fluid property calculations will use top segement pressure and temp " );
```
paveltomin: maybe add the name for WellControls and use `WellControls::viewKeyStruct::referenceReservoirRegionString` instead of "referenceReservoirRegion"

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
```diff
@@ -435,13 +436,30 @@ void CompositionalMultiphaseWell::validateInjectionStreams( WellElementSubRegion
...
 void CompositionalMultiphaseWell::validateWellConstraints( real64 const & time_n,
                                                            real64 const & GEOS_UNUSED_PARAM( dt ),
-                                                           WellElementSubRegion const & subRegion )
+                                                           WellElementSubRegion const & subRegion,
+                                                           ElementRegionManager const & elemManager )
 {
+  WellControls & wellControls = getWellControls( subRegion );
+  if( !wellControls.useSurfaceConditions() )
+  {
+    bool useSeg = wellControls.referenceReservoirRegion()=="";
```
paveltomin: const

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
tjb-ltk: I updated the comment,  the default region ave pressure is set to -1,  if a region name is defined then it will have a pressure > 0.0, hopefully,   if -1 then the top set pressure is used (current GEOS behavior)

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp

paveltomin (approved):
just fix that hardcoded `regionStatistics`

src/coreComponents/physicsSolvers/fluidFlow/wells/WellControls.cpp
```diff
@@ -117,7 +118,16 @@ WellControls::WellControls( string const & name, Group * const parent )
     setDefaultValue( 0 ).
     setInputFlag( InputFlags::OPTIONAL ).
     setDescription( "Flag to specify whether rates are checked at surface or reservoir conditions.\n"
-                    "Equal to 1 for surface conditions, and to 0 for reservoir conditions" );
+                    "Equal to 1 for surface conditions, and to 0 for reservoir conditions.\n"
+                    "See note on referenceReservoirRegion for reservoir condition options" );
+
+  registerWrapper( viewKeyStruct::referenceReservoirRegionString(), &m_referenceReservoirRegion ).
```
dkachuma, line 124: Looks like a schema change. There should be a corresponding change to the `schema.xsd` file.

src/coreComponents/physicsSolvers/fluidFlow/wells/WellControls.cpp
```diff
@@ -49,7 +49,8 @@ WellControls::WellControls( string const & name, Group * const parent )
   m_targetTotalRateTable( nullptr ),
   m_targetPhaseRateTable( nullptr ),
   m_targetBHPTable( nullptr ),
-  m_statusTable( nullptr )
+  m_statusTable( nullptr ),
+  m_regionAveragePressure( -1 )
```
dkachuma, line 53:
```suggestion
  m_regionAveragePressure( -1 ),
  m_regionAverageTemperature( -1 )
```

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
```diff
@@ -406,11 +407,11 @@ void CompositionalMultiphaseWell::validateInjectionStreams( WellElementSubRegion
 
     integer const streamSize = injectionStream.size();
     GEOS_THROW_IF( ( streamSize == 0 ),
-                   "WellControls '" << wellControls.getName() << "'" <<
+                   "WellControls " << wellControls.getName() << "'" <<
```
dkachuma: nitpick: Either lose both quotes or keep both.

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
```diff
@@ -435,13 +436,32 @@ void CompositionalMultiphaseWell::validateInjectionStreams( WellElementSubRegion
...
+    GEOS_WARNING_IF( useSeg,
+                     "WellControls " <<WellControls::viewKeyStruct::referenceReservoirRegionString() <<
+                     " not set and well constraint fluid property calculations will use top segement pressure and temp " );
+    if( useSeg )
+    {
+      wellControls.setRegionAveragePressure( -1 );
+    }
+    else
+    {
+      ElementRegionBase const & region = elemManager.getRegion( wellControls.referenceReservoirRegion());
```
dkachuma, line 456: `getRegion()` here might return a null reference. We need to validate the user's input for `referenceReservoirRegion` even if it's not empty. They may have a typo.

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
dkachuma: Do we know that this has been calculated at all? Could the average pressure be gibberish like 0 or inf?

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
```diff
@@ -672,9 +692,30 @@ void CompositionalMultiphaseWell::updateVolRatesForConstraint( WellElementSubReg
...
+    if( flashPressure < 0.0 )
+    {
+      // region name not set, use segment conditions
+      flashPressure   = pres[iwelemRef];
+      flashTemperature = temp[iwelemRef];
+    }
+    else
+    {
+      // use reservoir region averages
+      flashTemperature = wellControls.getRegionAveragePressure();
```
dkachuma:
```suggestion
      flashTemperature = wellControls.getRegionAverageTemperature();
```

dkachuma: I would treat the temperature separately i.e. also perform a check on the output of `wellControls.getRegionAverageTemperature()` and only assign the element temperature if this is negative.

dkachuma (commented):
I think we need to validate the user's input for `referenceReservoirRegion`.

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp

src/coreComponents/physicsSolvers/fluidFlow/wells/WellControls.cpp
tjb-ltk: done

src/coreComponents/physicsSolvers/fluidFlow/wells/WellControls.cpp
tjb-ltk: done

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
tjb-ltk: Good point..   validation checks have been added to ensure that reservoir region referenced in well control exists..

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
tjb-ltk: done, and validation added, see above comments..

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
tjb-ltk: validation has been added to ensure regionStatistics are calculated

src/coreComponents/physicsSolvers/fluidFlow/wells/WellControls.cpp
tjb-ltk: validation has been added to ensure referenceReservoirRegion exists..

dkachuma (approved):

PR #3624, opened 2025-04-10
fix: time step logic for sequential

paveltomin:
fix time step logic for sequential
some parts of the logic introduced in https://github.com/GEOS-DEV/GEOS/pull/3338 happen in `PhysicsSolverBase::execute` which is not executed for subsolvers of a coupled solver, while time step cut actually happens in a specific subsolver
this PR tries to centralize the logic within `PhysicsSolverBase::setNextDt` that should fix the issue when solver can not increase time step after a cut

### DIFF
--- a/.integrated_tests.yaml
+++ b/.integrated_tests.yaml
@@ -1,6 +1,6 @@
 baselines:
   bucket: geosx
-  baseline: integratedTests/baseline_integratedTests-pr3537-10973-fe0f2c7
+  baseline: integratedTests/baseline_integratedTests-pr3624-11053-ae011c7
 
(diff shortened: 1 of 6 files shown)


victorapm (approved):
I was just looking at this. Thanks for fixing it faster than me!

@jhuang2601 this should solve your convergence issue with the SEAM cases

jhuang2601 (approved):
Thanks for fixing this issue. 
Now the sequential cases of SEAM can be running properly on Pine.

castelletto1 (approved):

CusiniM (approved):

PR #3625, opened 2025-04-16
fix: remove incorrect use of template keyword in function calls with no template argument list

rrsettgast:
remove cases where template keyword is used with not template arg list in function call.

This warning shows up on apple-clang17:

```
/Users/settgast1/Codes/geos/GEOS/src/coreComponents/physicsSolvers/multiphysics/PoromechanicsSolver.hpp:148:20: error: a template argument list is expected after a name prefixed by the template keyword [-Wmissing-template-arg-list-after-template-kw]
  148 |     this->template forDiscretizationOnMeshTargets( domain.getMeshBodies(), [&] ( string const &,
```

### DIFF
--- a/src/coreComponents/physicsSolvers/wavePropagation/sem/elastic/secondOrderEqn/isotropic/ElasticWaveEquationSEMKernel.hpp
+++ b/src/coreComponents/physicsSolvers/wavePropagation/sem/elastic/secondOrderEqn/isotropic/ElasticWaveEquationSEMKernel.hpp
@@ -199,7 +199,7 @@ public:
                               StackVariables & stack ) const
   {
 
-    m_finiteElementSpace.template computeFirstOrderStiffnessTerm( q, stack.xLocal, [&] ( int i, int j, real64 val, real64 J[3][3], int p, int r )
+    m_finiteElementSpace.computeFirstOrderStiffnessTerm( q, stack.xLocal, [&] ( int i, int j, real64 val, real64 J[3][3], int p, int r )
     {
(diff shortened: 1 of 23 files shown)


src/coreComponents/physicsSolvers/wavePropagation/sem/elastic/secondOrderEqn/isotropic/ElasticWaveEquationSEMKernel.hpp
```diff
@@ -199,7 +199,7 @@ class ExplicitElasticSEMBase : public finiteElement::KernelBase< SUBREGION_TYPE,
                               StackVariables & stack ) const
   {
 
-    m_finiteElementSpace.template computeFirstOrderStiffnessTerm( q, stack.xLocal, [&] ( int i, int j, real64 val, real64 J[3][3], int p, int r )
+    m_finiteElementSpace.computeFirstOrderStiffnessTerm( q, stack.xLocal, [&] ( int i, int j, real64 val, real64 J[3][3], int p, int r )
```
CusiniM: those parameters can also be `const` I think. And I would pass `J` by reference.

rrsettgast: ...and perhaps the `J[3][3]` needs to be passed as a reference.

paveltomin (approved):

wrtobin (approved):

Issue #3627, opened 2025-04-22
Unsafe access to std::vector without bounds check


OmarDuran:
**Describe the bug**

In the [L478](https://github.com/GEOS-DEV/GEOS/blob/4aa2c9fd542b18cbbb493580a461f53cb33f2e42/src/coreComponents/common/format/table/TableFormatter.cpp#L478) of  file src/coreComponents/common/format/table/TableFormatter.cpp, the code accesses row.cells[columnId - 1] without verifying that columnId > 0, which lead to undefined behavior if columnId == 0. This should be guarded to ensure safe access within vector bounds.

**To Reproduce**
1. Compile testTable in Debug
2. ctest -R testTable

arng40:
This PR solves this issue https://github.com/GEOS-DEV/GEOS/pull/3614#pullrequestreview-2726188435

PR #3628, opened 2025-04-22
fix: Unsafe access to std::vector without bounds check

OmarDuran:
Ensure safe access in the [L478](https://github.com/GEOS-DEV/GEOS/blob/4aa2c9fd542b18cbbb493580a461f53cb33f2e42/src/coreComponents/common/format/table/TableFormatter.cpp#L478) of file src/coreComponents/common/format/table/TableFormatter.cpp.

fix issue #3627.

### DIFF
--- a/src/coreComponents/common/format/table/TableFormatter.cpp
+++ b/src/coreComponents/common/format/table/TableFormatter.cpp
@@ -475,9 +475,8 @@ void TableTextFormatter::stretchColumnsByMergedCellsWidth( std::vector< size_t >
       {
         CellLayoutRow & row = tableGrid[rowId];
         TableLayout::CellLayout const & cell = row.cells[columnId];
-        TableLayout::CellLayout const & leftCell = row.cells[columnId - 1];
         bool const isToMergeWithRight = cell.m_cellType == CellType::MergeNext;
-        bool const isToMergeWithLeft = columnId > 0 && leftCell.m_cellType == CellType::MergeNext;
+        bool const isToMergeWithLeft = columnId > 0 && row.cells[columnId - 1].m_cellType == CellType::MergeNext;
         bool const isFirstToMerge = isToMergeWithRight && !isToMergeWithLeft;
         if( isFirstToMerge )
         { // we detected a set of cells to merge, let's compute its total width
--- a/src/coreComponents/common/format/table/TableFormatter.cpp
+++ b/src/coreComponents/common/format/table/TableFormatter.cpp
@@ -476,7 +476,7 @@ void TableTextFormatter::stretchColumnsByMergedCellsWidth( stdVector< size_t > &
         CellLayoutRow & row = tableGrid[rowId];
         TableLayout::CellLayout const & cell = row.cells[columnId];
         bool const isToMergeWithRight = cell.m_cellType == CellType::MergeNext;
-bool const isToMergeWithLeft = ( columnId > 0 ) ? ( row.cells[columnId - 1].m_cellType == CellType::MergeNext ) : false; 
+        bool const isToMergeWithLeft = ( columnId > 0 ) ? ( row.cells[columnId - 1].m_cellType == CellType::MergeNext ) : false; 
         bool const isFirstToMerge = isToMergeWithRight && !isToMergeWithLeft;
         if( isFirstToMerge )
         { // we detected a set of cells to merge, let's compute its total width
--- a/src/coreComponents/common/format/table/TableFormatter.cpp
+++ b/src/coreComponents/common/format/table/TableFormatter.cpp
@@ -476,7 +476,7 @@ void TableTextFormatter::stretchColumnsByMergedCellsWidth( stdVector< size_t > &
         CellLayoutRow & row = tableGrid[rowId];
         TableLayout::CellLayout const & cell = row.cells[columnId];
         bool const isToMergeWithRight = cell.m_cellType == CellType::MergeNext;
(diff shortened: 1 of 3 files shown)


src/coreComponents/common/format/table/TableFormatter.cpp
```diff
@@ -475,9 +475,8 @@ void TableTextFormatter::stretchColumnsByMergedCellsWidth( std::vector< size_t >
       {
         CellLayoutRow & row = tableGrid[rowId];
         TableLayout::CellLayout const & cell = row.cells[columnId];
-        TableLayout::CellLayout const & leftCell = row.cells[columnId - 1];
         bool const isToMergeWithRight = cell.m_cellType == CellType::MergeNext;
-        bool const isToMergeWithLeft = columnId > 0 && leftCell.m_cellType == CellType::MergeNext;
+        bool const isToMergeWithLeft = columnId > 0 && row.cells[columnId - 1].m_cellType == CellType::MergeNext;
```
CusiniM, line 479: why?

src/coreComponents/mesh/generators/VTKUtilities.cpp
```diff
@@ -1435,17 +1435,17 @@ std::vector< localIndex > getWedgeNodeOrderingFromPolyhedron( vtkCell * const ce
   {
     auto edgeNode0 = cell->GetEdge( iEdge )->GetPointId( 0 );
     auto edgeNode1 = cell->GetEdge( iEdge )->GetPointId( 1 );
-    auto it0 = std::find( &nodeTri0[0], &nodeTri0[3], edgeNode0 );
-    auto it1 = std::find( &nodeTri0[0], &nodeTri0[3], edgeNode1 );
+    auto it0 = std::find( nodeTri0.begin(), nodeTri0.end(), edgeNode0 );
```
CusiniM, line 1438: I do prefer this version but if we are sure that we are dealing with triangles I do not understand how it's different from the original one.

CusiniM (approved):

src/coreComponents/mesh/generators/VTKUtilities.cpp
OmarDuran: Because bound checks for std::vector::[], the original one breaks when the test is compiled with apple Clang 16.0 in Debug mode.

src/coreComponents/common/format/table/TableFormatter.cpp
```diff
@@ -475,9 +475,8 @@ void TableTextFormatter::stretchColumnsByMergedCellsWidth( stdVector< size_t > &
...
```
rrsettgast: your solution is fine, but this would be better imo....just because we aren't relying on the early exit after the lhs of `&&` fails. 
```suggestion
bool const isToMergeWithLeft = ( columnId > 0 ) ? ( row.cells[columnId - 1].m_cellType == CellType::MergeNext ) : false; 
```

src/coreComponents/mesh/mpiCommunications/CommunicationTools.cpp
```diff
@@ -1030,7 +1030,7 @@ bool CommunicationTools::asyncUnpack( MeshLevel & mesh,
   MpiWrapper::testSome( icomm.size(),
                         icomm.mpiRecvBufferRequest(),
                         &recvCount,
-                        &neighborIndices[0],
+                        &*neighborIndices.begin(),
```
rrsettgast: std::vector<>::begin() is undefined behavior if the vector is 0 length.

```suggestion
                        neighborIndices.data(),
```

rrsettgast (changes requested):

src/coreComponents/common/format/table/TableFormatter.cpp
OmarDuran: thanks @rrsettgast I agree with this

src/coreComponents/mesh/mpiCommunications/CommunicationTools.cpp
OmarDuran: That works better. Once we move to C++20, we can ideally use things like `std::to_address(neighborIndices.begin())`.

src/coreComponents/common/format/table/TableFormatter.cpp
```diff
@@ -475,9 +475,8 @@ void TableTextFormatter::stretchColumnsByMergedCellsWidth( stdVector< size_t > &
       {
         CellLayoutRow & row = tableGrid[rowId];
         TableLayout::CellLayout const & cell = row.cells[columnId];
-        TableLayout::CellLayout const & leftCell = row.cells[columnId - 1];
         bool const isToMergeWithRight = cell.m_cellType == CellType::MergeNext;
-        bool const isToMergeWithLeft = columnId > 0 && leftCell.m_cellType == CellType::MergeNext;
+bool const isToMergeWithLeft = ( columnId > 0 ) ? ( row.cells[columnId - 1].m_cellType == CellType::MergeNext ) : false;
```
rrsettgast:
```suggestion
        bool const isToMergeWithLeft = ( columnId > 0 ) ? ( row.cells[columnId - 1].m_cellType == CellType::MergeNext ) : false; 
```

rrsettgast: @OmarDuran I didn't indent my suggestion properly

rrsettgast (approved):

PR #3629, opened 2025-04-22
feat: Output iteration & convergence information to CSV

arng40:
Summary
---

This PR aims to regroup all solver statistics, including iteration and convergence information, under the existing SolverStatistics class. The statistics can now be output in CSV format

CSV File output
---

When `writeSolvingConvergence` and/or `writeSolverIterations` are set to 1, CSV files will be generated under `convergence/ ` directory and by presented as

**Iterations statistics**  : 
```numTimeSteps, newton Iteration, numTimeStepCuts, setup time, solve time, Successful outer loop,Successful nonlinear,Successful linear,Discarded outer loop,Discarded nonlinear,Discarded linear```

**Convergence statistics** : 
 For the moment only residuals values are output

```Cycle number,time_n (s), dt (s), [Residuals values], [...]```

Logging Improvements
---

Solver iterations  statistics has been refactored  to enhance clarity, by displaying iteration accumulations.
It's triggered by setting a `logLevel >= 1`.

```
-----------------------------------------
|      SolverStatistics iterations      |
|---------------------------------------|
|         Components         |  Iter    |
|----------------------------|----------|
|     Number of Time-steps:  |      32  |
|  Number of Time steps cut  |       0  |
|      Successful nonlinear  |     114  |
|         Successful linear  |     114  |
|      Discarded outer loop  |       0  |
|       Discarded nonlinear  |       0  |
|          Discarded linear  |       0  |
-----------------------------------------
```

### CODE REVIEW

**Comment by paveltomin** [comment]
@arng40 seems to be based on https://github.com/GEOS-DEV/GEOS/pull/3019, could you please change the base branch to simplify the review?
@rrsettgast @CusiniM can we merge https://github.com/GEOS-DEV/GEOS/pull/3019, it's been hanging there for > 1 year

**Comment by arng40** [comment]
@paveltomin In the LogPart PR  I changed the way we retrieved some information about the final time step (`Event Base/Manager and PhysicsSolverBase`) and I needed it for this PR that's why I base this PR on LogPart.

File: src/coreComponents/common/format/table/TableLayout.hpp [hunk 1]
```diff
@@ -723,6 +730,9 @@ class TableLayout
   /// Columns settings hierarchy
   ColumnsList m_tableColumns;
 
+  /// Columns registry ensuring having unique column name
+  std::map< string, integer > m_columnsRegistry;
```
MelReyCG: Why is this constraint needed?
- Currently, it is buggy (i.e. a column name can be modified, and `m_columnsRegistry` is not impacted),
- It is hard to maintain,
- Strangely, it only applies to the first columns level (not sub-columns..?),
- Maybe someone will want to have the same name twice or more for specific reason,
- Any double add is silently not adding the new column.

src/coreComponents/events/EventBase.hpp
```diff
@@ -167,6 +167,19 @@ class EventBase : public ExecutableGroup
...
+   * @brief Get the track of substeps
+   * @return A vector containing the the track of substeps
+   */
+  std::vector< real64 > const & getSubStepDts() const {  return m_target->getSubStepDts(); }
+
+  /**
+   * @brief Get number of time sub-steps allowed for the solver
+   * @return The number of substeps taken
+   */
+  size_t getNumOfSubSteps() const {  return m_target->getSubStepDts().size(); }
```
MelReyCG: - An event does not have substeps. Just do a `getEventTarget().getSubStepDts()`
- pointer might not be non-null, you have to control its value.

src/coreComponents/physicsSolvers/SolverStatistics.hpp
```diff
@@ -76,123 +140,183 @@ class SolverStatistics : public dataRepository::Group
...
-  integer getNumTimeStepCuts() const
-  { return m_numTimeStepCuts; }
+  void setFilename( string_view filename )
+  { m_IterationsFileName = filename; }
+
+private:
+  /// Table containing statistics relative to non linear parameter
+  TableData m_nonLinearData;
+  /// Filename for the iterations CSV.
+  string m_IterationsFileName;
```
MelReyCG:
```suggestion
  string m_iterationsFileName;
```

src/coreComponents/physicsSolvers/fluidFlow/proppantTransport/ProppantTransport.cpp
```diff
@@ -154,6 +154,8 @@ void ProppantTransport::initializePreSubGroups()
 {
   FlowSolverBase::initializePreSubGroups();
 
+  m_solverStatistics.setOutputFilesName( getName() );
```
MelReyCG: Can it be done only once in the base class `FlowSolverBase` or `PhysicsSolverBase`?

src/coreComponents/physicsSolvers/SolverStatistics.cpp
```diff
@@ -101,59 +134,154 @@ void SolverStatistics::logTimeStepCut()
...
-      logStat( "discarded linear iterations", m_numDiscardedLinearIterations );
+      residualsNormCells.emplace_back( TableData::CellData( {CellType::Value,
+                                                             GEOS_FMT( "{}", residual.value )} ));
     }
   }
+
+  m_nonLinearNormsData.addRow( residualsNormCells );
 }
+
+void ConvergenceStatistics::outputResidualNorm( bool writeCSV ) const
```
MelReyCG: Just to standardize a bit?
```suggestion
void ConvergenceStatistics::outputStatistics( bool writeCSV ) const
```

src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
```diff
@@ -1420,7 +1439,8 @@ void PhysicsSolverBase::cleanup( real64 const GEOS_UNUSED_PARAM( time_n ),
                                  real64 const GEOS_UNUSED_PARAM( eventProgress ),
                                  DomainPartition & GEOS_UNUSED_PARAM( domain ) )
 {
-  m_solverStatistics.outputStatistics();
+  m_solverStatistics.m_iterationsStats.outputStatistics( m_writeSolverIterationsCSV );
+  m_solverStatistics.m_convergenceStats.outputResidualNorm( m_writeSolvingConvergenceCSV );
```
MelReyCG: IIUC, you accumulate all statistics, of all convergence iterations, of all simulation steps:
- the memory footprint is growing forever,
- if GEOS crashes, all these informations are lost.

Why not streaming that each time you got a new row?

MelReyCG (changes requested):

src/coreComponents/common/format/table/TableLayout.hpp
arng40: Indeed  I'll remove this

src/coreComponents/events/EventBase.hpp
arng40: It was a squash from an old section commit, removed

MelReyCG (changes requested):
Users will need:
- both cumulated & current timestep solver stats,
- the cycle number in the iteration & convergence stats.

src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
```diff
@@ -93,6 +93,19 @@ PhysicsSolverBase::PhysicsSolverBase( string const & name,
     setRestartFlags( RestartFlags::WRITE_AND_READ ).
     setDescription( "Cut time step if linear solution fail without going until max nonlinear iterations." );
 
+  registerWrapper( viewKeyStruct::writeSolverIterationsCSVFlagString(), &m_writeSolverIterationsCSV ).
+    setApplyDefaultValue( 0 ).
+    setInputFlag( InputFlags::OPTIONAL ).
+    setRestartFlags( RestartFlags::NO_WRITE ).
+    setDescription( "When set to 1, output interations information to a csv" );
+
+  registerWrapper( viewKeyStruct::writeSolvingConvergenceCSVFlagString(), &m_writeSolvingConvergenceCSV ).
```
paveltomin: how about a single flag and 1 will be iterations only, 2 will be iterations and residuals (convergence) ?

src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
```diff
@@ -285,7 +306,8 @@ bool PhysicsSolverBase::execute( real64 const time_n,
     subStepDts[subStep] = dtAccepted;
 
     // increment the cumulative number of nonlinear and linear iterations
-    m_solverStatistics.saveTimeStepStatistics();
+    getIterationStats().iterateTimeStepStatistics();
```
paveltomin, line 313: iterate? collect?

src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
```diff
@@ -647,6 +670,8 @@ bool PhysicsSolverBase::lineSearch( real64 const & time_n,
     GEOS_LOG_LEVEL_RANK_0( logInfo::LineSearch,
                            GEOS_FMT( "        ( R ) = ( {:4.2e} )", residualNorm ) );
 
+    getConvergenceStats().m_residualNormT = residualNorm;
```
paveltomin: why "T" ?

src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
```diff
@@ -744,9 +769,10 @@ bool PhysicsSolverBase::lineSearchWithParabolicInterpolation( real64 const & tim
 
     // get residual norm
     residualNormT = calculateResidualNorm( time_n, dt, domain, dofManager, rhs.values() );
-    GEOS_LOG_LEVEL_RANK_0( logInfo::LineSearch,
+    GEOS_LOG_LEVEL_RANK_0( logInfo::ResidualNorm,
                            GEOS_FMT( "        ( R ) = ( {:4.2e} )", residualNormT ) );
 
+    getConvergenceStats().m_residualNormT = residualNormT;
```
paveltomin: i am not sure how line search should be treated, did you test it? what will be the output?

src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
```diff
@@ -1122,6 +1150,13 @@ bool PhysicsSolverBase::solveNonlinearSystem( real64 const & time_n,
 
     lastResidual = residualNorm;
   }
+  if( m_writeSolverIterationsCSV )
```
paveltomin: so here it is a single flag?

src/coreComponents/physicsSolvers/SolverStatistics.cpp
```diff
@@ -38,122 +55,270 @@ SolverStatistics::SolverStatistics( string const & name, Group * const parent )
     setApplyDefaultValue( 0 ).
     setDescription( "Number of time step cuts" );
 
-  registerWrapper( viewKeyStruct::numSuccessfulOuterLoopIterationsString(), &m_numSuccessfulOuterLoopIterations ).
+  registerWrapper( viewKeyStruct::numSuccessfulOuterLoopIterationsString(),
+                   &m_numSuccessfulOuterLoopIterations ).
     setApplyDefaultValue( 0 ).
     setDescription( "Cumulative number of successful outer loop iterations" );
```
paveltomin: why is it called outer loop here?

src/coreComponents/physicsSolvers/SolverStatistics.cpp
```diff
@@ -38,122 +55,270 @@ SolverStatistics::SolverStatistics( string const & name, Group * const parent )
...
 
-  registerWrapper( viewKeyStruct::numDiscardedLinearIterationsString(), &m_numDiscardedLinearIterations ).
+  registerWrapper( viewKeyStruct::numDiscardedLinearIterationsString(),
+                   &m_numDiscardedLinearIterations ).
     setApplyDefaultValue( 0 ).
     setDescription( "Cumulative number of discarded linear iterations" );
+
+  m_iterationCSVLayout = std::make_unique< TableLayout >();
+  m_iterationCSVLayout->setTitle( GEOS_FMT( "{} iterations", getParent().getName()));
+  m_iterationCSVLayout->addColumns( { "numTimeSteps", "newton Iteration", "numTimeStepCuts",
```
paveltomin: the namings are inconsistent, please clean up (lower-upper, camel case???)

src/coreComponents/physicsSolvers/SolverStatistics.cpp
```diff
@@ -38,122 +55,270 @@ SolverStatistics::SolverStatistics( string const & name, Group * const parent )
...
+
+  struct ResidualInfo
   {
-    GEOS_LOG_RANK_0( GEOS_FMT( "{}, number of {}: {}",
-                               getParent().getName(), name, value ) );
+    const real64 & value;
+    string_view columnName;
+  };
+
+  ResidualInfo residuals[] = {
```
paveltomin: why do you need this? can not it be somewhat general without reference to specific solver?

src/coreComponents/physicsSolvers/SolverStatistics.cpp
```diff
@@ -38,122 +55,270 @@ SolverStatistics::SolverStatistics( string const & name, Group * const parent )
...
+                                                         GEOS_FMT( "{}", m_dt )} ));
 
-  if( m_numTimeSteps > 0 )
+  for( auto const & residual : residuals )
   {
-    logStat( "time steps", m_numTimeSteps );
+    residualsNormCells.emplace_back( TableData::CellData(
+      {
+        CellType::Value,
+        !std::isnan( residual.value ) ? GEOS_FMT( "{}", residual.value ) : "0"
```
paveltomin: why hiding nan?

src/coreComponents/physicsSolvers/fluidFlow/ReactiveCompositionalMultiphaseOBL.cpp
```diff
@@ -182,6 +182,8 @@ void ReactiveCompositionalMultiphaseOBL::implicitStepComplete( real64 const & ti
       porousMaterial.saveConvergedState();
     } );
   } );
+
+  writeStatisticsToTable();
```
paveltomin: can this be somewhere in the solver base?

src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
arng40: I found this for the line search residual : 
PhysicsSolver.hpp 319
```residualNormT the residual norm at the end of the line search```

File: src/coreComponents/physicsSolvers/fluidFlow/ReactiveCompositionalMultiphaseOBL.cpp [hunk 17]
```diff
@@ -602,10 +606,12 @@ void ReactiveCompositionalMultiphaseOBL::initializePostInitialConditionsPreSubGr
...
 
 void
-ReactiveCompositionalMultiphaseOBL::implicitStepSetup( real64 const & GEOS_UNUSED_PARAM( time_n ),
-                                                       real64 const & GEOS_UNUSED_PARAM( dt ),
+ReactiveCompositionalMultiphaseOBL::implicitStepSetup( real64 const & time_n,
+                                                       real64 const & dt,
+                                                       integer const cycleNumber,
                                                        DomainPartition & domain )
 {
+  updateSolverStatistics( time_n, dt, cycleNumber );
```
paveltomin: same, can it go somewhere in the base solver class?

src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
arng40: Forgot to split, I'll ask to some user what they prefer

src/coreComponents/physicsSolvers/SolverStatistics.cpp
arng40: This is the original name but I can rename it.

src/coreComponents/physicsSolvers/fluidFlow/ReactiveCompositionalMultiphaseOBL.cpp
arng40: I couldn't find any appropriate function in the base class where every solver and sub-solver are being called for each solverstep 
implicitStepComplete is the best function I could find

src/coreComponents/physicsSolvers/fluidFlow/ReactiveCompositionalMultiphaseOBL.cpp
arng40: same I couldn't find any appropriate function in the base class

src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
arng40: yes this option is better

src/coreComponents/physicsSolvers/SolverStatistics.cpp
arng40: it's much simpler to set up and I don't think it makes much difference to parse the csv.

src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
MelReyCG: We can go with that

src/coreComponents/physicsSolvers/SolverStatistics.cpp
MelReyCG: I think his goal was to prevent any formatting crash, is that the case @arng40 ?  If so, why not displaying "NaN"?

src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
```diff
@@ -247,7 +269,7 @@ real64 PhysicsSolverBase::solverStep( real64 const & time_n,
   // final step for completion of timestep. typically secondary variable updates and cleanup.
   {
     Timer timer( m_timers["step complete"] );
-    implicitStepComplete( time_n, dt_return, domain );
+    implicitStepComplete( time_n, dt, domain );
```
dkachuma: Is this an intended change in the logic? I'm guessing `dt_return` is different from `dt`.

src/coreComponents/physicsSolvers/PhysicsSolverBase.hpp
```diff
@@ -693,6 +708,12 @@ class PhysicsSolverBase : public ExecutableGroup
 
     /// @return string for the allowNonConvergedLinearSolverSolution wrapper
     static constexpr char const * allowNonConvergedLinearSolverSolutionString() { return "allowNonConvergedLinearSolverSolution"; }
+
+    /// @return string for the writeLinearSystem wrapper
+    static constexpr char const * writeSolvingConvergenceCSVFlagString() { return "writeSolvingConvergenceCSV"; }
```
dkachuma:
```suggestion
    static constexpr char const * writeSolvingConvergenceCSVFlagString() { return "writeConvergenceCSV"; }
```
I wouldn't include "solver" as it is implied by where the field is defined.

src/coreComponents/physicsSolvers/SolverStatistics.hpp
```diff
@@ -22,176 +22,385 @@
...
+    static constexpr char const * IterationsStatisticsString() { return "IterationsStatistics"; }
+  };
+
+  /**
+   * @brief Create a convergence directory if we enable the csv for iteration or convergence statistics
+   * @param writeConvergence Boolean for convergencce CSV output
+   * @param writeIteration Boolean for iteration CSV output
+   */
+  void makeDir( bool writeConvergence, bool writeIteration )
+  {if( writeConvergence || writeIteration ) makeDirsForPath( m_outputDir ); std::cout << "pliplop "<<writeConvergence<<writeIteration<<  m_outputDir <<std::endl; }
```
dkachuma:
```suggestion
  {if( writeConvergence || writeIteration ) makeDirsForPath( m_outputDir ); }
```

src/coreComponents/physicsSolvers/SolverStatistics.hpp
```diff
@@ -22,176 +22,385 @@
...
+  std::unique_ptr< TableCSVFormatter > m_convergenceFormatter;
+  /// Filename for the solver CSV convergence file.
+  string m_convergenceFilename;
+};
+
+/**
+ * @class SolverStatistics
+ * @brief This class records solver statistics for each time step.
+ */
+class SolverStatistics : public dataRepository::Group
```
dkachuma, line 348: Does making this a `Group` mean it can be put in the xml? How does that look - do you have an example?

src/coreComponents/schema/docs/ReactiveBrineThermal.rst
dkachuma, line 1: I think we are not supposed to upload these rst files anymore.

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
```diff
@@ -1358,13 +1358,16 @@ CompositionalMultiphaseWell::calculateResidualNorm( real64 const & time_n,
     GEOS_LOG_LEVEL_RANK_0( logInfo::ResidualNorm, GEOS_FMT( "        ( R{} ) = ( {:4.2e} )        ( Renergy ) = ( {:4.2e} )",
                                                             coupledSolverAttributePrefix(), globalResidualNorm[0], globalResidualNorm[1] ));
 
+    getConvergenceStats().m_residualWell = globalResidualNorm[0];
```
dkachuma: This is probably a `residualMass` so since the well solver will have it's own stats object, I would remove the `residualWell` member variable and use `residualMass`.

src/coreComponents/physicsSolvers/SolverStatistics.cpp
```diff
@@ -38,122 +55,270 @@ SolverStatistics::SolverStatistics( string const & name, Group * const parent )
...
-  m_numSuccessfulLinearIterations += m_currentNumLinearIterations;
-  m_numTimeSteps++;
+  m_convergenceLayout = std::make_unique< TableLayout >();
+
+  m_convergenceLayout->addColumns( {"Cycle number", "time_n (s)", "dt (s)",
+                                    "RMass", "RVol", "REnergy",
+                                    "RFlow", "RBubbleDisp", "RFrac",
+                                    "Rstick", "Rslip", "Ropen",
+                                    "RSolid", "RContact", "RProppant",
+                                    "RWell", "RDamage", "RTotal", "R"} );
```
dkachuma: Some of these columns will have meaning for specific solvers/models. For example `REnergy` is only valid for thermal models; `RProppant` will mean nothing for a lot of models. Is there a way of trimming these based on the model.

dkachuma (commented):
Now that #3019 has been merged, can we see the relevant changes.

dkachuma (commented):
This looks good to me. My only issue is that we should report only those convergence measures that are defined in the model.

src/coreComponents/physicsSolvers/SolverStatistics.cpp
arng40: For this PR, I keep it simple, I will update the mecanism in the next PR

src/coreComponents/physicsSolvers/SolverStatistics.hpp
MelReyCG: That does not seem possible since the `createChild()` method of the parent group `PhysicsSolverBase` does not have this capability.
`SolverStatistics` is a `Group` in order to be - someday - output in HDF5. For now it cannot because "1D" data still cannot be output.

src/coreComponents/physicsSolvers/SolverStatistics.cpp
arng40: I think for the users 0 is is more understandable than NAN

src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
arng40: I think iterate is more suitable

src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
arng40: Do you have a deck to test the behavior ? I can't find one :/

dkachuma (approved):

src/coreComponents/physicsSolvers/PhysicsSolverBase.hpp
paveltomin: @arng40 can you rename `writeSolvingConvergenceCSVFlagString` consistently

paveltomin:
check what happened with `src/coreComponents/LvArray`

src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
```diff
@@ -745,12 +765,13 @@ bool PhysicsSolverBase::lineSearchWithParabolicInterpolation( real64 const & tim
       GEOS_LOG_LEVEL_RANK_0( logInfo::LineSearch,
                              GEOS_FMT( "        Line search @ {:0.3f}:      ", cumulativeScale ) );
     }
-
+    std::cout << " R tested "<< std::endl;
```
paveltomin: remove

src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
```diff
@@ -851,7 +872,7 @@ real64 PhysicsSolverBase::nonlinearImplicitStep( real64 const & time_n,
         else
         {
           // increment the solver statistics for reporting purposes
-          m_solverStatistics.logOuterLoopIteration();
+          getIterationStats().incrementNonlinearIteration();
```
paveltomin: how about `m_convergenceStats` as reference in the solver?

src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
paveltomin: i don't think you need to output this one, the solver should already fill the one it really owns

src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
```diff
@@ -1126,6 +1149,13 @@ bool PhysicsSolverBase::solveNonlinearSystem( real64 const & time_n,
 
     lastResidual = residualNorm;
   }
+  if( getWrapper< integer >( viewKeyStruct::writeSolverString()).reference() >= 1 )
```
paveltomin: i don't like getWrapper blablabla here, please simplify

src/coreComponents/physicsSolvers/PhysicsSolverBase.hpp
```diff
@@ -694,6 +709,9 @@ class PhysicsSolverBase : public ExecutableGroup
     /// @return string for the allowNonConvergedLinearSolverSolution wrapper
     static constexpr char const * allowNonConvergedLinearSolverSolutionString() { return "allowNonConvergedLinearSolverSolution"; }
 
+    /// @return string for the writeLinearSystem wrapper
+    static constexpr char const * writeSolverString() { return "writeSolver"; }
```
paveltomin: `writeSolver` is very ambiguous, also function description seems copied from somewhere else

src/coreComponents/physicsSolvers/SolverStatistics.cpp
```diff
@@ -38,122 +55,268 @@ SolverStatistics::SolverStatistics( string const & name, Group * const parent )
...
     setApplyDefaultValue( 0 ).
     setDescription( "Cumulative number of discarded linear iterations" );
+
+  m_iterationCSVLayout = std::make_unique< TableLayout >();
+  m_iterationCSVLayout->setTitle( GEOS_FMT( "{} iterations", getParent().getName()));
+  m_iterationCSVLayout->addColumns( { "Number of time steps", "Newton iteration", "Number of time step cuts",
+                                      "Setup time", "Solve time",
+                                      "Successful outer loop", "Successful nonlinear", "Successful linear",
+                                      "Discarded outer loop", "Discarded nonlinear", "Discarded linear"} );
+  //m_iterationCSVLayout->addColumn( "Iter" );
```
paveltomin: remove

src/coreComponents/physicsSolvers/SolverStatistics.cpp
paveltomin: yeah, but one has to always sync this statics code to the solvers which is going to be a pain

src/coreComponents/physicsSolvers/SolverStatistics.hpp
```diff
@@ -22,176 +22,388 @@
...
+  /**
+   * @brief Set the Residual Norms filename
+   * @param solverName The solverName as a string_view.
+   */
+  void setOutputFilesName( string_view solverName );
+
+  /**
+   * @return Return the string directory where all CSV are generated
+   */
+  string_view getDirectory() const
```
paveltomin: this is never used

src/coreComponents/physicsSolvers/fluidFlow/wells/SinglePhaseWell.cpp
```diff
@@ -76,6 +76,8 @@ void SinglePhaseWell::registerDataOnMesh( Group & meshBodies )
 {
   WellSolverBase::registerDataOnMesh( meshBodies );
 
+  m_solverStatistics.setOutputFilesName( getName());
```
paveltomin: probably not needed

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseFVM.cpp
```diff
@@ -71,6 +71,8 @@ void SinglePhaseFVM< BASE >::initializePreSubGroups()
 {
   BASE::initializePreSubGroups();
 
+  BASE::m_solverStatistics.setOutputFilesName( BASE::getName() );
```
paveltomin: probably not needed

src/coreComponents/physicsSolvers/SolverStatistics.cpp
```diff
@@ -38,122 +55,268 @@ SolverStatistics::SolverStatistics( string const & name, Group * const parent )
...
+{
+  m_time_n = time_n;
+  m_dt = dt;
+  m_cycleNumber = cycleNumber;
 }
+void ConvergenceStatistics::resetResidualsValue()
+{
+  m_residualSolid = 0;
+  m_totalResidual=0;
+  m_residualNormT=0;
```
paveltomin: why it is only resetting those three?

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseHybridFVM.cpp
```diff
@@ -729,6 +731,8 @@ real64 CompositionalMultiphaseHybridFVM::calculateResidualNorm( real64 const & G
 
   GEOS_LOG_LEVEL_RANK_0( logInfo::ResidualNorm, GEOS_FMT( "        ( R{} ) = ( {:4.2e} )", coupledSolverAttributePrefix(), residualNorm ));
 
+  getConvergenceStats().m_residualProppant = residualNorm;
```
paveltomin: "Proppant" is wrong here

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseFVM.cpp
```diff
@@ -215,6 +217,9 @@ real64 SinglePhaseFVM< BASE >::calculateResidualNorm( real64 const & GEOS_UNUSED
 
     GEOS_LOG_LEVEL_RANK_0_NLR( logInfo::Convergence, GEOS_FMT( "        ( R{} ) = ( {:4.2e} )        ( Renergy ) = ( {:4.2e} )",
                                                                FlowSolverBase::coupledSolverAttributePrefix(), globalResidualNorm[0], globalResidualNorm[1] ));
+    BASE::getConvergenceStats().m_residualMass = globalResidualNorm[0];
+    BASE::getConvergenceStats().m_residualVol = globalResidualNorm[1];
```
paveltomin: "Vol" is wrong here

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseFVM.cpp
```diff
@@ -228,8 +233,9 @@ real64 SinglePhaseFVM< BASE >::calculateResidualNorm( real64 const & GEOS_UNUSED
       physicsSolverBaseKernels::L2ResidualNormHelper::computeGlobalNorm( localResidualNorm[0], localResidualNormalizer[0], residualNorm );
     }
 
-    GEOS_LOG_LEVEL_RANK_0_NLR( logInfo::Convergence,
+    GEOS_LOG_LEVEL_RANK_0_NLR( logInfo::ResidualNorm,
                                GEOS_FMT( "        ( R{} ) = ( {:4.2e} )", FlowSolverBase::coupledSolverAttributePrefix(), residualNorm ));
+    BASE:: getConvergenceStats().m_residualFlow = residualNorm;
```
paveltomin: why there is no "total residual" here?

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseHybridFVM.cpp
```diff
@@ -576,6 +578,7 @@ real64 SinglePhaseHybridFVM::calculateResidualNorm( real64 const & GEOS_UNUSED_P
   }
 
   GEOS_LOG_LEVEL_RANK_0( logInfo::ResidualNorm, GEOS_FMT( "        ( R{} ) = ( {:4.2e} )", coupledSolverAttributePrefix(), residualNorm ));
+  getConvergenceStats().m_residualProppant = residualNorm;
```
paveltomin: "Proppant" is wrong here

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
paveltomin: i am not sure calling that residual `residualMass` is the right thing to do

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseFVM.cpp
paveltomin: overall, the use of "total residual" is not consistent between the solvers

src/coreComponents/physicsSolvers/fluidFlow/wells/SinglePhaseWell.cpp
```diff
@@ -954,6 +956,8 @@ SinglePhaseWell::calculateResidualNorm( real64 const & time_n,
                                                             coupledSolverAttributePrefix(), resNorm ));
   }
 
+  getConvergenceStats().m_residualMass = resNorm;
```
paveltomin: how about Renergy ?

src/coreComponents/physicsSolvers/solidMechanics/contact/SolidMechanicsEmbeddedFractures.cpp
```diff
@@ -582,12 +582,17 @@ real64 SolidMechanicsEmbeddedFractures::calculateResidualNorm( real64 const & ti
 
   // Matrix residual
   real64 const solidResidualNorm = SolidMechanicsLagrangianFEM::calculateResidualNorm( time, dt, domain, dofManager, localRhs );
+  getConvergenceStats().m_residualSolid = solidResidualNorm;
```
paveltomin: probably not needed, should be already assigned inside of `SolidMechanicsLagrangianFEM::calculateResidualNorm`

src/coreComponents/physicsSolvers/solidMechanics/contact/SolidMechanicsLagrangeContact.cpp
```diff
@@ -864,12 +869,20 @@ real64 SolidMechanicsLagrangeContact::calculateContactResidualNorm( DomainPartit
...
+                             GEOS_FMT( "        ( Rstick Rslip Ropen ) = ( {:15.6e} {:15.6e} {:15.6e} )",
+                                       stickResidual, slipResidual, openResidual ));
+
+
+  real64 totalResidualNorm = sqrt( stickResidual * stickResidual + slipResidual * slipResidual + openResidual * openResidual );
+
+  getConvergenceStats().m_residualStick = stickResidual;
+  getConvergenceStats().m_residualSlip = slipResidual;
+  getConvergenceStats().m_residualOpen = openResidual;
+  getConvergenceStats().m_totalResidual = totalResidualNorm;
```
paveltomin: this should be something like residualContact, not total

src/coreComponents/physicsSolvers/solidMechanics/contact/SolidMechanicsLagrangeContactBubbleStab.cpp
```diff
@@ -467,7 +474,13 @@ real64 SolidMechanicsLagrangeContactBubbleStab::calculateResidualNorm( real64 co
 
   real64 const contactResidual = calculateContactResidualNorm( domain, dofManager, localRhs );
 
-  return sqrt( solidResidual * solidResidual + contactResidual * contactResidual );
+  real64 const totalResidual = sqrt( solidResidual * solidResidual + contactResidual * contactResidual );
+
+  getConvergenceStats().m_residualSolid = solidResidual;
```
paveltomin: not needed

paveltomin:
@arng40 @MelReyCG @dkachuma how about this one https://github.com/GEOS-DEV/GEOS/pull/3754

src/coreComponents/physicsSolvers/SolverStatistics.cpp
```diff
@@ -38,122 +55,268 @@ SolverStatistics::SolverStatistics( string const & name, Group * const parent )
...
+}
+
+void IterationsStatistics::outputStatistics()
+{
+  if( !m_isIterativeSolver || !m_logOutput )
+    return;
+
+  {
+    TableLayout iterationLogLayout ( GEOS_FMT( "{} iterations", getParent().getName() ),
+                                     { "Components", "Iter  "} );
```
paveltomin: no need for `{ "Components", "Iter "} );`

src/coreComponents/physicsSolvers/multiphysics/CoupledReservoirAndWellsBase.hpp
```diff
@@ -209,16 +209,23 @@ class CoupledReservoirAndWellsBase : public CoupledSolver< RESERVOIR_SOLVER, WEL
   postInputInitialization() override
   {
     Base::postInputInitialization();
+    wellSolver()->getIterationStats().setIterativeSolver( false );
```
paveltomin: why?

src/coreComponents/physicsSolvers/multiphysics/CoupledSolver.hpp
```diff
@@ -96,6 +96,7 @@ class CoupledSolver : public PhysicsSolverBase
       GEOS_LOG_LEVEL_RANK_0( logInfo::Coupling,
                              GEOS_FMT( "{}: found {} solver named {}",
                                        getName(), solver->getCatalogName(), solverName ) );
+      solver->getIterationStats().setIterativeSolver( false );
```
paveltomin: why?

src/coreComponents/physicsSolvers/multiphysics/PoromechanicsSolver.hpp
```diff
@@ -113,6 +113,9 @@ class PoromechanicsSolver : public CoupledSolver< FLOW_SOLVER, MECHANICS_SOLVER
                    GEOS_FMT( "{} {}: The attribute `{}` of the flow solver must be thermal since the poromechanics solver is thermal",
                              this->getCatalogName(), this->getName(), this->flowSolver()->getName() ),
                    InputError );
+
+    flowSolver()->getIterationStats().setIterativeSolver( false );
+    solidMechanicsSolver()->getIterationStats().setIterativeSolver( false );
```
paveltomin: why?

src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
```diff
@@ -1152,6 +1183,17 @@ void PhysicsSolverBase::setupDofs( DomainPartition const & GEOS_UNUSED_PARAM( do
   GEOS_ERROR( "PhysicsSolverBase::setupDofs called!. Should be overridden." );
 }
 
+void PhysicsSolverBase::updateSolverStatistics( real64 const & time_n, real64 const & dt, integer const cycleNumber )
+{
+  getConvergenceStats().updateSolverStep( time_n, dt, cycleNumber );
```
paveltomin: this should also have the iteration number

src/coreComponents/physicsSolvers/SolverStatistics.cpp
```diff
@@ -38,122 +55,268 @@ SolverStatistics::SolverStatistics( string const & name, Group * const parent )
...
+    m_convergenceFormatter.reset( new TableCSVFormatter( *m_convergenceLayout ));
+    m_logStream << m_convergenceFormatter->headerToString( );
   }
+
+  m_logStream << m_convergenceFormatter->dataToString( m_convergenceData );
+  m_logStream.flush();
+  m_convergenceData.clear();
+}
+
+void ConvergenceStatistics::updateSolverStep( real64 const & time_n, real64 const & dt, integer const cycleNumber )
```
paveltomin: add iteration number

src/coreComponents/physicsSolvers/SolverStatistics.cpp
```diff
@@ -38,122 +55,268 @@ SolverStatistics::SolverStatistics( string const & name, Group * const parent )
...
+    { m_totalResidual, "RTotal" },
+    { m_residualNormT, "R" }
   };
 
-  // TODO: the print logic is really convoluted to accomodate the needs of the different solvers, needs simplification
+  residualsNormCells.emplace_back( TableData::CellData( {CellType::Value,
+                                                         GEOS_FMT( "{}", m_cycleNumber )} ));
+  residualsNormCells.emplace_back( TableData::CellData( {CellType::Value,
+                                                         GEOS_FMT( "{}", m_time_n )} ));
+  residualsNormCells.emplace_back( TableData::CellData( {CellType::Value,
```
paveltomin, line 213: add iteration number

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -1349,10 +1349,12 @@ void CompositionalMultiphaseBase::initializePostInitialConditionsPreSubGroups()
...
 
 void
-CompositionalMultiphaseBase::implicitStepSetup( real64 const & GEOS_UNUSED_PARAM( time_n ),
-                                                real64 const & GEOS_UNUSED_PARAM( dt ),
+CompositionalMultiphaseBase::implicitStepSetup( real64 const & time_n,
+                                                real64 const & dt,
+                                                integer const cycleNumber,
                                                 DomainPartition & domain )
 {
+  updateSolverStatistics( time_n, dt, cycleNumber );
```
paveltomin: this needs to be called at every iteration

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -2652,6 +2654,7 @@ void CompositionalMultiphaseBase::implicitStepComplete( real64 const & time,
       }
     } );
   } );
+  writeStatisticsToTable();
```
paveltomin: with current implementation, needs to be called at every iteration

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseHybridFVM.cpp
arng40: Why is this wrong `coupledSolverAttributePrefix` is Proppant, this is a totalResidual ?

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseFVM.cpp
arng40: Change to energy

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseFVM.cpp
arng40: In this case RFlow = totalResidual ? In general at the end of `calculateResidualNorm`  if we have one type of residual, is this also the totalResidual ? It's not clear for me

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseHybridFVM.cpp
arng40: same coupledSolverAttributePrefix() give "Proppant"

src/coreComponents/physicsSolvers/fluidFlow/wells/SinglePhaseWell.cpp
arng40: same as above coupledSolverAttributePrefix give "well"

src/coreComponents/physicsSolvers/solidMechanics/contact/SolidMechanicsLagrangeContact.cpp
arng40: same as above : https://github.com/GEOS-DEV/GEOS/pull/3629#discussion_r2251708977

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseHybridFVM.cpp
paveltomin: `  static string coupledSolverAttributePrefix() { return "flow"; }`

i think calling it Rflow would make the most sense

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseFVM.cpp
paveltomin: yeah, you are right, sorry for confusion
it should be Rflow here, no total

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseHybridFVM.cpp
paveltomin: no, it should give "flow"

src/coreComponents/physicsSolvers/fluidFlow/wells/SinglePhaseWell.cpp
paveltomin: there is Renergy above that does not seem to get recorded

src/coreComponents/physicsSolvers/solidMechanics/contact/SolidMechanicsLagrangeContact.cpp
paveltomin: this is not total residual, it does not make sense to assign it here, it will be overwritten outside of that function anyway

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseHybridFVM.cpp
arng40: oh, not paying enough attention,I was given a wrong redirection by the IDE

src/coreComponents/physicsSolvers/PhysicsSolverBase.hpp
arng40: change to `writestatistics`

src/coreComponents/physicsSolvers/fluidFlow/wells/SinglePhaseWell.cpp
arng40: i've updated thispart

acitrain (approved):

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseFVM.cpp
```diff
@@ -215,6 +217,9 @@ real64 SinglePhaseFVM< BASE >::calculateResidualNorm( real64 const & GEOS_UNUSED
 
     GEOS_LOG_LEVEL_RANK_0_NLR( logInfo::Convergence, GEOS_FMT( "        ( R{} ) = ( {:4.2e} )        ( Renergy ) = ( {:4.2e} )",
                                                                FlowSolverBase::coupledSolverAttributePrefix(), globalResidualNorm[0], globalResidualNorm[1] ));
+    BASE::getConvergenceStats().m_residualFlow = globalResidualNorm[0];
+    BASE::getConvergenceStats().m_residualEnergy = globalResidualNorm[1];
+    BASE::getConvergenceStats().m_totalResidual = residualNorm;
```
paveltomin: no need to assign total here

src/coreComponents/physicsSolvers/solidMechanics/contact/SolidMechanicsEmbeddedFractures.cpp
```diff
@@ -586,8 +586,12 @@ real64 SolidMechanicsEmbeddedFractures::calculateResidualNorm( real64 const & ti
   if( !m_useStaticCondensation )
   {
     real64 const fractureResidualNorm = calculateFractureResidualNorm( domain, dofManager, localRhs );
+    real64 totalResidualNorm = sqrt( solidResidualNorm * solidResidualNorm + fractureResidualNorm * fractureResidualNorm );
 
-    return sqrt( solidResidualNorm * solidResidualNorm + fractureResidualNorm * fractureResidualNorm );
+    getConvergenceStats().m_residualFracture = fractureResidualNorm;
+    getConvergenceStats().m_totalResidual = totalResidualNorm;
```
paveltomin: no need to record total here

src/coreComponents/physicsSolvers/multiphysics/PoromechanicsSolver.hpp
arng40: removed

src/coreComponents/physicsSolvers/multiphysics/CoupledSolver.hpp
arng40: removed

src/coreComponents/physicsSolvers/multiphysics/CoupledReservoirAndWellsBase.hpp
arng40: removed

src/coreComponents/physicsSolvers/solidMechanics/contact/SolidMechanicsLagrangeContactBubbleStab.cpp
```diff
@@ -510,10 +519,8 @@ real64 SolidMechanicsLagrangeContactBubbleStab::calculateContactResidualNorm( Do
   stickResidual = MpiWrapper::sum( stickResidual );
   stickResidual = sqrt( stickResidual );
 
-  if( getLogLevel() >= 1 && logger::internal::rank==0 )
-  {
-    std::cout << GEOS_FMT( "        ( Rt  ) = ( {:15.6e}  )", stickResidual );
-  }
+  GEOS_LOG_LEVEL_RANK_0( logInfo::ResidualNorm,
```
paveltomin: _NLR is needed here

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseFVM.cpp
arng40: indeed

src/coreComponents/physicsSolvers/SolverStatistics.cpp
```diff
@@ -38,122 +54,216 @@ SolverStatistics::SolverStatistics( string const & name, Group * const parent )
...
+
+  if( !m_logStream.is_open() )
   {
-    if( printOuterLoopIterations )
-    {
-      logStat( "successful outer loop iterations", m_numSuccessfulOuterLoopIterations );
-    }
-    logStat( "successful nonlinear iterations", m_numSuccessfulNonlinearIterations );
-    if( printLinearIterations ) // don't print for the outer iterations in sequential schemes
+    string_array header = {"Cycle number", "time_n (s)", "dt (s)", "Newton iteration"};
```
paveltomin: calling it "newton" is not always correct (for example for sequential solver), either nonlinear or just iteration is better

src/coreComponents/physicsSolvers/SolverStatistics.hpp
```diff
@@ -22,176 +22,321 @@
...
-  integer getNumDiscardedOuterLoopIterations() const
-  { return m_numDiscardedOuterLoopIterations; }
+  void updateSolverStep( real64 const & time_n,
+                         real64 const & dt,
+                         integer const cycleNumber,
+                         integer const newtonIter );
+
+  void updateSolverStep( real64 const & time_n,
+                         real64 const & dt,
+                         integer const cycleNumber ){}
```
paveltomin: why this empty function is needed?

arng40: i was refactoring, it will be removed

paveltomin (approved):
looks good now, thanks
two small things:
- please check the remaining minor changes in https://github.com/GEOS-DEV/GEOS/pull/3754
- please check what happened to `src/coreComponents/LvArray` and if it really needs to be changed

src/coreComponents/physicsSolvers/multiphysics/CoupledSolver.hpp
```diff
@@ -503,11 +507,20 @@ class CoupledSolver : public PhysicsSolverBase
...
+                                                  iter,
                                                   time_n,
                                                   stepDt,
                                                   domain );
 
+        forEachArgInTuple( m_solvers, [&]( auto & solver, auto )
+        {
+
+          solver->getConvergenceStats().updateSolverStep( time_n, dt, cycleNumber, iter );
+          solver->getConvergenceStats().writeConvergenceStatsToTable();
```
paveltomin: maybe move it inside of `checkSequentialConvergence`

src/coreComponents/physicsSolvers/multiphysics/PoromechanicsSolver.hpp
```diff
@@ -251,8 +252,15 @@ class PoromechanicsSolver : public CoupledSolver< FLOW_SOLVER, MECHANICS_SOLVER
...
     if( m_performStressInitialization )
       subcycling = 1;
+    std::cout << "checkSequentialConvergence" << std::endl;
+    bool isConverged = Base::checkSequentialConvergence( cycleNumber, iter, time_n, dt, domain );
+
+    flowSolver()->getConvergenceStats().updateSolverStep( time_n, dt, cycleNumber, iter );
+    flowSolver()->getConvergenceStats().writeConvergenceStatsToTable();
+
+    solidMechanicsSolver()->getConvergenceStats().updateSolverStep( time_n, dt, cycleNumber, iter );
+    solidMechanicsSolver()->getConvergenceStats().writeConvergenceStatsToTable();
```
paveltomin: don't forget to remove those

src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
```diff
@@ -905,6 +928,7 @@ real64 PhysicsSolverBase::nonlinearImplicitStep( real64 const & time_n,
     }
     else
     {
+      cleanup( time_n, dt, 0, cycleNumber, domain );
```
paveltomin: you sure this will not lead to double-calling of `cleanup`?

src/coreComponents/physicsSolvers/SolverStatistics.cpp
```diff
@@ -38,122 +54,218 @@ SolverStatistics::SolverStatistics( string const & name, Group * const parent )
...
+    iterationDataLog.addRow( "Discarded nonlinear iterations", m_numDiscardedNonlinearIterations );
+    iterationDataLog.addRow( "Discarded linear iterations", m_numDiscardedLinearIterations );
+
+    GEOS_LOG_RANK_0( statsFormatter.toString( iterationDataLog ));
+  }
+}
+
+ConvergenceStatistics::ConvergenceStatistics()
+{
+  m_convergenceLayout = std::make_unique< TableLayout >();
```
paveltomin: does this have to be a pointer?

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.hpp
```diff
@@ -103,8 +103,8 @@ class CompositionalMultiphaseFVM : public CompositionalMultiphaseBase
                            arrayView1d< real64 > const & localRhs ) override;
 
   virtual real64
-  calculateResidualNorm( real64 const & time_n,
-                         real64 const & dt,
+  calculateResidualNorm( real64 const & GEOS_UNUSED_PARAM( time_n ),
+                         real64 const & GEOS_UNUSED_PARAM( dt ),
```
paveltomin: no need for `GEOS_UNUSED_PARAM` here

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseHybridFVM.hpp
```diff
@@ -100,7 +100,7 @@ class CompositionalMultiphaseHybridFVM : public CompositionalMultiphaseBase
                            arrayView1d< real64 > const & localRhs ) override;
 
   virtual real64
-  calculateResidualNorm( real64 const & time_n,
+  calculateResidualNorm( real64 const & GEOS_UNUSED_PARAM( time_n ),
```
paveltomin: same as above

src/coreComponents/physicsSolvers/fluidFlow/ReactiveCompositionalMultiphaseOBL.hpp
```diff
@@ -145,8 +145,8 @@ class ReactiveCompositionalMultiphaseOBL : public FlowSolverBase
              DofManager & dofManager ) const override;
 
   virtual real64
-  calculateResidualNorm( real64 const & time_n,
-                         real64 const & dt,
+  calculateResidualNorm( real64 const & GEOS_UNUSED_PARAM( time ),
+                         real64 const & GEOS_UNUSED_PARAM( dt ),
```
paveltomin: same

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseHybridFVM.hpp
```diff
@@ -98,7 +98,7 @@ class SinglePhaseHybridFVM : public SinglePhaseBase
                            arrayView1d< real64 > const & localRhs ) override;
 
   virtual real64
-  calculateResidualNorm( real64 const & time_n,
+  calculateResidualNorm( real64 const & GEOS_UNUSED_PARAM( time_n ),
```
paveltomin: same

src/coreComponents/physicsSolvers/multiphysics/CoupledReservoirAndWellsBase.hpp
```diff
@@ -209,6 +209,8 @@ class CoupledReservoirAndWellsBase : public CoupledSolver< RESERVOIR_SOLVER, WEL
   postInputInitialization() override
   {
     Base::postInputInitialization();
+    Base::getIterationStats().setCSVOutput( this->m_writeStatistics >= 2 );
+    Base::getConvergenceStats().setCSVOutput( this->m_writeStatistics >= 2 );
```
paveltomin: remove

src/coreComponents/physicsSolvers/multiphysics/CoupledSolver.hpp
```diff
@@ -248,6 +247,22 @@ class CoupledSolver : public PhysicsSolverBase
...
+    {
+      solver->updateConvergenceStep( time_n, dt, cycleNumber, iteration );
+
+      if( m_writeStatistics >= 2 )
+      {
+        solver->getConvergenceStats().updateSolverStep( time_n, dt, cycleNumber, iteration );
+        solver->getConvergenceStats().writeConvergenceStatsToTable();
+      }
+    } );
+  }
```
paveltomin, line 260: do you need to update/write the parent solver stats here?

src/coreComponents/physicsSolvers/multiphysics/CoupledSolver.hpp
```diff
@@ -501,13 +515,18 @@ class CoupledSolver : public PhysicsSolverBase
...
-
         // Check convergence of the outer loop
-        isConverged = checkSequentialConvergence( iter,
+        isConverged = checkSequentialConvergence( cycleNumber,
+                                                  iter,
                                                   time_n,
                                                   stepDt,
                                                   domain );
 
+        forEachArgInTuple( m_solvers, [&]( auto & solver, auto )
```
paveltomin: maybe just call `updateConvergenceStep` here?

paveltomin: oh i see it is being called inside of `checkSequentialConvergence`, so the added `forEachArgInTuple` here should be redundant - remove

src/coreComponents/physicsSolvers/multiphysics/CoupledSolver.hpp
```diff
@@ -673,6 +696,10 @@ class CoupledSolver : public PhysicsSolverBase
   {
     setSubSolvers();
 
+    getIterationStats().setLogOutput( m_writeStatistics >= 1 );
+    getIterationStats().setCSVOutput( m_writeStatistics >= 2 );
+    getConvergenceStats().setCSVOutput( m_writeStatistics >= 2 );
```
paveltomin: remove and call base instead

src/coreComponents/physicsSolvers/simplePDE/PhaseFieldDamageFEM.hpp
```diff
@@ -93,8 +93,8 @@ class PhaseFieldDamageFEM : public PhysicsSolverBase
                                         CRSMatrixView< real64, globalIndex const > const & localMatrix,
                                         arrayView1d< real64 > const & localRhs ) override;
 
-  virtual real64 calculateResidualNorm( real64 const & time_n,
-                                        real64 const & dt,
+  virtual real64 calculateResidualNorm( real64 const & GEOS_UNUSED_PARAM( time_n ),
+                                        real64 const & GEOS_UNUSED_PARAM( dt ),
```
paveltomin: revert

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.hpp
```diff
@@ -154,8 +154,8 @@ class SolidMechanicsLagrangianFEM : public PhysicsSolverBase
                                         arrayView1d< real64 > const & localRhs ) override;
 
   virtual real64
-  calculateResidualNorm( real64 const & time_n,
-                         real64 const & dt,
+  calculateResidualNorm( real64 const & GEOS_UNUSED_PARAM( time_n ),
```
paveltomin: revert

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsStateReset.cpp
```diff
@@ -130,6 +130,7 @@ bool SolidMechanicsStateReset::execute( real64 const time_n,
     } );
   } );
 
+  m_solidSolver->getConvergenceStats().resetResidualsValue();
```
paveltomin: probably not needed here

src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
arng40: removed

src/coreComponents/physicsSolvers/multiphysics/CoupledSolver.hpp
arng40: just after updating all stats, do you want another function to write ?

paveltomin: I mean - it only calls update and write for subsolvers, but not for the parent coupled solver itself

by the way - what is solver->updateAndWriteConvergenceStep doing? it looks like things are done twice here

src/coreComponents/physicsSolvers/PhysicsSolverBase.hpp
```diff
@@ -472,6 +477,19 @@ class PhysicsSolverBase : public ExecutableGroup
...
+   * @param time_n the time at the beginning of the step
+   * @param dt the desired timestep
+   * @param cycleNumber event cycle number
+   * @param iteration current iteration
+   */
+  virtual void
+  updateAndWriteConvergenceStep( real64 const & GEOS_UNUSED_PARAM( time_n ),
+                                 real64 const & GEOS_UNUSED_PARAM( dt ),
+                                 integer const GEOS_UNUSED_PARAM( cycleNumber ),
+                                 integer const GEOS_UNUSED_PARAM( iteration ) ){}
```
paveltomin: why this is empty?

arng40: I update the func for the base class

src/coreComponents/physicsSolvers/multiphysics/CoupledSolver.hpp
arng40: duplicate -> remove

src/coreComponents/physicsSolvers/SolverStatistics.hpp
```diff
@@ -22,176 +22,377 @@
...
+                        dataRepository::Group * const parent );
+
+  /// State of log output. True when writeStatistics is set to 1
+  bool m_logOutput = false;
+
+  /// State of csv output. True when writeStatistics is set to 2
+  bool m_csvOutput = false;
+
+  /// Number of time steps
+  integer m_numTimeSteps = 0;
```
MelReyCG: Why all those attributes became public? As these are processed my methods, they should be private (could unclear to what is allowed)

src/coreComponents/physicsSolvers/SolverStatistics.hpp
```diff
@@ -22,176 +22,377 @@
...
+
+  /// Number of time step cuts
+  integer m_numTimeStepCuts = 0;
+
+  /// Number of configuration iterations in the current time step (utility variable constantly overwritten)
+  integer m_currentNumConfigIterations = 0;
```
MelReyCG: name seems better, but could it be more explained what is done during a "configuration iterations"? (maybe it is just me)

src/coreComponents/physicsSolvers/SolverStatistics.hpp
```diff
@@ -22,176 +22,377 @@
...
+
 
   /**
    * @brief Initialize the counters used for an individual time step
    */
-  void initializeTimeStepStatistics();
+  void resetCurrentTimeStepStatistics();
+
+  /**
+   * @brief Tell the solverStatistics that we are doing a configuration iteration
```
MelReyCG:
```suggestion
   * @brief Tell the solverStatistics that we have done a configuration iteration
```
so it is really specifying how we should use this class?

MelReyCG: same remark for all your explaining comments in `SolverStatistics.cpp`, they can give information to class users

src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
```diff
@@ -102,6 +102,12 @@ PhysicsSolverBase::PhysicsSolverBase( string const & name,
     setInputFlag( InputFlags::FALSE ).
     setRestartFlags( RestartFlags::WRITE_AND_READ );
 
+  registerWrapper( viewKeyStruct::writeStatisticsString(), &m_writeStatistics ).
```
MelReyCG:
```suggestion
  registerWrapper( viewKeyStruct::writeStatisticsCSVString(), &m_writeStatistics ).
```
?

src/coreComponents/physicsSolvers/PhysicsSolverBase.hpp
```diff
@@ -705,6 +722,9 @@ class PhysicsSolverBase : public ExecutableGroup
     /// @return string for the allowNonConvergedLinearSolverSolution wrapper
     static constexpr char const * allowNonConvergedLinearSolverSolutionString() { return "allowNonConvergedLinearSolverSolution"; }
 
+    /// @return string for the writeStatistics wrapper
+    static constexpr char const * writeStatisticsString() { return "writeStatistics"; }
```
MelReyCG:
```suggestion
    /// @return string for the writeStatisticsCSV wrapper
    static constexpr char const * writeStatisticsCSVString() { return "writeStatisticsCSV"; }
```
?

src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
```diff
@@ -985,8 +1007,11 @@ bool PhysicsSolverBase::solveNonlinearSystem( real64 const & time_n,
 
       // get residual norm
       residualNorm = calculateResidualNorm( time_n, stepDt, domain, m_dofManager, m_rhs.values() );
-      GEOS_LOG_LEVEL_RANK_0( logInfo::Convergence,
+
+      GEOS_LOG_LEVEL_RANK_0( logInfo::ResidualNorm,
                              GEOS_FMT( "        ( R ) = ( {:4.2e} )", residualNorm ) );
+      getConvergenceStats().m_residuals["R"] = residualNorm;
+      updateAndWriteConvergenceStep( time_n, stepDt, cycleNumber, newtonIter );
```
MelReyCG: No flag that disables these features?
If those attributes would remain private, it would be easier to disable them.

src/coreComponents/physicsSolvers/SolverStatistics.hpp
```diff
@@ -22,176 +22,377 @@
...
-   * @return Cumulative number of discarded linear iterations
+   * @brief Construct a new Convergence Statistics object
    */
-  integer getNumDiscardedLinearIterations() const
-  { return m_numDiscardedLinearIterations; }
+  ConvergenceStatistics();
+
+  /// State of csv output. True when writeSolver is set to 2
+  bool m_csvOutput = false;
+
```
MelReyCG: Maybe you want to split that in two:
```suggestion
  /// True when statistics output flags is set to 2
  bool m_csvOutputEnabled = false;

  /// Current state of csv output.
  bool m_csvOutputOpened = false;
```

And, as I requested private attributes, you can check for one flag or the other depending on if it is file or data related.

src/coreComponents/physicsSolvers/SolverStatistics.hpp
arng40: @paveltomin What do you think ? Maybe supplement the doc

src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
arng40: y

src/coreComponents/physicsSolvers/SolverStatistics.hpp
paveltomin: maybe, not sure to what extent

src/coreComponents/physicsSolvers/SolverStatistics.hpp
```diff
@@ -22,176 +22,381 @@
...
   /**
-   * @brief Output the cumulative statistics to the terminal
+   * @brief Reset the setupTime & solveTime to 0 at the end of each cycle
    */
-  void outputStatistics() const;
+  void resetSolverLinearTime();
 
   /**
-   * @return Number of time steps
+   * @brief Tell the solverStatistics that we cut the time step and we increment the cumulative counters for discarded timesteps
```
MelReyCG, line 120: Maybe, to explain the timestep cut process.
```suggestion
   * @brief Tell the solverStatistics that we cut the time step and we increment the cumulative counters for discarded timesteps
   *        Assume that we restart computing the last timestep cut we are computing (it did not converge).
```

src/coreComponents/physicsSolvers/SolverStatistics.hpp
```diff
@@ -22,176 +22,381 @@
...
+  void updateSolverStep( real64 const & time_n,
+                         real64 const & dt,
+                         integer const cycleNumber,
+                         integer const newtonIter );
+
+  /**
+   * @brief Reset the solid residuals value.
+   * Call by SolidMechanicsStateReset.
+   */
+  void resetResidualsValue();
```
MelReyCG: unused

src/coreComponents/physicsSolvers/SolverStatistics.hpp
```diff
@@ -22,176 +22,381 @@
...
+   */
+  string & getFilename()
+  { return m_convergenceFilename;  }
+
+  /**
+   * @brief Close the stream output file
+   */
+  void closeFile()
+  { m_logStream.close(); }
+
```
MelReyCG:
```suggestion
  void closeFile()
  { if(m_csvOutputOpened) { m_logStream.close(); m_csvOutputOpened = false; } }
```
(-> cpp)

MelReyCG: same for `IterationStatistics`

MelReyCG (approved):
There some few remarks present. Beyond that this seems nice.

src/coreComponents/physicsSolvers/SolverStatistics.hpp
MelReyCG: adding something like...
```
/// Represents how many times the solver tried to configure and solve the system before either converging or cutting the timestep.
/// Configuration loops can be added in order to scale the timestep delta-time, as many time as needed to achieve a convergence.
```
?

paveltomin: nope :(

MelReyCG: If I remember, the idea of renaming this stat to "config iteration" is from you. Maybe if you explain the reason why you prefered the "config" term over "outer loop", will help us to find for a fix for this proposal or to design a new one.
And/or can you explain what seems wrong / inaccurate in this proposal?

This statistic will be displayed to users which will not necessarily read the code, a documentation will be helpful.

src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
arng40: In hpp, now I only have m_solverStatistics, I can't access it directly

src/coreComponents/physicsSolvers/SolverStatistics.hpp
paveltomin: because the thing it counts is called config iteration in GEOS, not "outer loop", see https://github.com/search?q=repo%3AGEOS-DEV%2FGEOS%20configuration%20iteration&type=code
i already spent quite some time with this PR, please just merge it

herve-gross (approved):

src/coreComponents/physicsSolvers/SolverStatistics.hpp
MelReyCG: Oh ok, there was two terms in the codebase.
Personally I approved yesterday.

MelReyCG: The thing is that was not well documented, so I though that if we could start something here to clarify this term, we can have a reference for later.
There are also reference of "outer-loop" at some places, that will be the subject of another PR.

PR #3632, opened 2025-04-24
fix: error message for vtk import to avoid hard crash

paveltomin:
one change - check for collocated_nodes  presence not only in parallel but also is serial

second thing - check for global ids presence before calling
```
buildLocalToGlobal( vtkIdTypeArray::FastDownCast( faceMesh->GetCellData()->GetGlobalIds() ),
                        vtkIdTypeArray::FastDownCast( mesh->GetCellData()->GetGlobalIds() ) )

```
and for unit test - actually provide global ids
no real fixes
just more elegant crash instead of segmentation fault


src/coreComponents/mesh/generators/VTKFaceBlockUtilities.cpp
```diff
@@ -647,6 +647,14 @@ void importFractureNetwork( string const & faceBlockName,
   faceBlock.set2dElemToFaces( std::move( elem2dTo3d.elem2dToFaces ) );
   faceBlock.set2dElemToElems( std::move( elem2dTo3d.elem2dToElem3d ) );
 
+  if( num2dElements > 0 )
+  {
+    GEOS_ERROR_IF( !vtkIdTypeArray::FastDownCast( faceMesh->GetCellData()->GetGlobalIds() ),
+                   "Global IDs are required for not found for fracture mesh" );
```
CusiniM:
```suggestion
                   "Global IDs are required but not found for fracture mesh" );
```

src/coreComponents/mesh/generators/VTKFaceBlockUtilities.cpp
```diff
@@ -647,6 +647,14 @@ void importFractureNetwork( string const & faceBlockName,
...
+  }
+  GEOS_ERROR_IF( !vtkIdTypeArray::FastDownCast( mesh->GetCellData()->GetGlobalIds() ),
+                 "Global IDs are required for not found for main mesh" );
```
CusiniM:
```suggestion
                 "Global IDs are required but not found for main mesh" );
```

CusiniM (approved):

PR #3639, opened 2025-04-29
refactor: Move fluid tests under constitutive

dkachuma:
The purpose of this PR is to move the multiphase fluid tests from `unitTests/constitutiveTests` to `constitutive/unitTests`. These tests do not need to depend on solvers and do not require the solver functionality. This initial PR moves only the deadoil test with the other `testMultiFluid*` test to be moved in subsequent PRs.


paveltomin (approved):

rrsettgast (approved):
@dkachuma FYI. If you move a file use "git mv" so that doesn't think you are deleting the old file and creating a new file. This will preserve history.

If you want, you could "git rebase -i" and unto the "mv" then apply the "git mv".

dkachuma:
> @dkachuma FYI. If you move a file use "git mv" so that doesn't think you are deleting the old file and creating a new file. This will preserve history.

I have reinstated the file history. The combined move and edit was making it murky.

PR #3643, opened 2025-05-01
fix: sync perf found

paveltomin:
sync `resElemFound` to avoid different output when running serial and mpi
no real changes are expected


tjb-ltk (commented):
what was the difference in output

paveltomin:
> what was the difference in output

some ranks will keep searching for perf-to-reservoir connection even when it was already found on another rank
just confusing to see diffs when run serial and parallel

tjb-ltk (approved):

CusiniM (approved):

PR #3645, opened 2025-05-02
fix: dt for control/pres bal eqn

tjb-ltk:
Fixes for temperature derivatives of well control equation/pressure drop eqn.  phasevolrate dt just wrong, cut and paste error,   mass density off by factor of gravD.   Unit test didn't catch due to tolerance.


paveltomin (approved):
👍

dkachuma (approved):

PR #3655, opened 2025-05-08
feat: update kernel variants

rrsettgast:
This PR:

- [x] disables some kernel specifications 
- [ ] adds cmake option to specify alternative kernel files
- [ ] adds new kernel specifications for other physicsSolvers


paveltomin:
@rrsettgast i did a pass on flow kernels

rrsettgast:
@paveltomin Looks like progress! @CusiniM please have a look.

src/coreComponents/physicsSolvers/KernelLaunchSelectors.hpp
```diff
@@ -28,12 +28,7 @@ namespace internal
 template< typename S, typename T, typename LAMBDA >
 void invokePhaseDispatchLambda ( S val, T numPhases, LAMBDA && lambda )
 {
-  if( numPhases == 1 )
```
CusiniM, line 31: so we exclude the possibility of using the compositional solver for a single phase multicomponent system? I guess that's fine since we are working on a dedicated reactiveTransport solver.

CusiniM (commented):
What @paveltomin did for the flow looks good to me. Do we need the `VEM` for `SolidMechanicsFixedStressThermoPoromechanicsKernels` ?

src/coreComponents/physicsSolvers/KernelLaunchSelectors.hpp
paveltomin: we can always re-enable it later when needed, what would be nice is to have a single place to specify those constraints to keep everything consistent, but not sure it's doable

castelletto1:
> What @paveltomin did for the flow looks good to me. Do we need the `VEM` for `SolidMechanicsFixedStressThermoPoromechanicsKernels` ?

My preference is to use VEM sparingly for now, limiting its application to solid mechanics and isothermal poromechanics. We can incrementally expand its use later to support a broader range of solvers.

castelletto1:
Are `SolidMechanicsFixedStressThermoPoroElasticKernels` still needed?


https://github.com/GEOS-DEV/GEOS/blob/ce3095709aaf7614c59741e7988bbe096103a3ac/src/coreComponents/physicsSolvers/solidMechanics/kernelSpecs.json#L114-L138

I would say they are already covered by `SolidMechanicsFixedStressThermoPoromechanicsKernels`

https://github.com/GEOS-DEV/GEOS/blob/ce3095709aaf7614c59741e7988bbe096103a3ac/src/coreComponents/physicsSolvers/solidMechanics/kernelSpecs.json#L75-L112

dkachuma:
I have moved the fluid update kernel. See #3674

PR #2933, opened 2024-01-12
injection example fix

arng40:
Correction of SimpleCO2InjTutorial_smoke example due to incorrect values on :
- pvtgas.txt,
- pvtliquid.txt,
- co2flash.txt.

https://github.com/GEOS-DEV/integratedTests/pull/102

PR #3661, opened 2025-05-13
docs: Update CO2 injection example

### LINKED ISSUE #3660: Tutorial Example for CO2 injection where the plume is different between the tutorial and the files to be loaded
**Describe the bug**
I rerun tutorial example for [CO2 Injection](https://geosx-geosx.readthedocs-hosted.com/en/latest/docs/sphinx/basicExamples/co2Injection/Example.html?_sm_au_=iHVcPT2vJHFJ6nnRQ0WpHK6H8sjL6#)

First problem: The co2flash.txt file created manually with the data in the tutorial is not the correct one. This file generates a bug. The correct one is at the same level as the 2 XML files in the inputFiles/compositionalMultiphaseWell/ directory.

Image

Second problem: The plume generated is not the same between the tutorial and the files to be loaded
See the difference in the picture below: 

Image

**To Reproduce**
Go to the CO2 injection case in the documentation
Download the 2 files indicated at the top of the tutorial:
inputFiles/compositionalMultiphaseWell/simpleCo2InjTutorial_base.xml
inputFiles/compositionalMultiphaseWell/simpleCo2InjTutorial_smoke.xml
And launch the simulation with the latest version of GEOS

**Expected behavior**
We do not know the desired behavior because there are 2 possibilities. But we ask ourselves the following questions.
Where does the change in plume come from?
Which is the best rendering the one from the tutorial or the one we get now?
What do we do?

- Update of the documentation?
      * Update images to match uploaded files?
      * Or update files to find previous images?

- Why is this not detected by the regression tests? What checks are being made? Should the tests be updated?

Choice 1: Update of the XML content to find previous images
    -> Can be made by TTE US team
Choice 2: Update of the co2flash.txt content and update of the picture in the documentation
    -> Can be made by TTE Pau team

- Machine: Pangea4
- Compiler: gcc 12.2.1
- GEOS Version: e09f499

dkachuma:
This PR updates the CO2 injection example to address the issues raised in #3660

The pvt files in this case were updated at #2933. However, the update to the `co2flash.txt` file was missed. I have therefore updated the example so that the actual files are referenced directly without having to duplicate them in the rst.

The update at #2933 changed the ranges of pressure and temperature used to generate the PVT tables. These new values are more correct because they include the surface conditions required by the well. I cannot go back to the old ranges because now geos has a check to ensure that the well operates within the range of the PVT data.

According to the change at #2933, and also at #1065, I suspect the original images were generated with a temperature range that was too low. This led to the gas density being higher and therefore the gas not being as buoyant. When this was corrected, the gas became more buoyant as can be seen with the new results.

I have edited the case to make the images look a bit more like the original. I have halved the vertical permeability to curtail the gas migration upwards and reduced the porosity to 0.05 to ensure that the gas reaches the surface. The [updated example](https://geosx-geosx--3661.com.readthedocs.build/en/3661/docs/sphinx/basicExamples/co2Injection/Example.html) has the new updated images.


tjb-ltk (approved):

CusiniM (approved):

paveltomin (approved):

jhuang2601 (approved):
@dkachuma Thanks for updating this case. 
@margauxraguenel Please take a look on this fix.

margauxraguenel (approved):
Tested on our side, perfect!

rrsettgast (approved):

PR #3662, opened 2025-05-13
feat: Flash for single phase mixtures

dkachuma:
#### Summary
This PR enhances the flash calculation framework to better handle phase transitions, particularly when a previously non-existent phase appears. The goal is to maintain continuity in properties and derivatives of this previously non-existent phase during the transition from single-phase to two-phase behaviour.

#### Key Changes
- **Continuity Handling**: Captures the *incipient phase composition* from the stability test to ensure smooth property transitions when a new phase emerges.
- **Negative Flash Handling**: Negative flash results are no longer immediately truncated. Instead, they are used to compute derivatives before truncation. The incipient phase composition is retained, and only the phase fraction is set to zero.
- **Code Refactoring**:
  - `NegativeTwoPhaseFlash` logic is now split into a dedicated `_impl` file.
  - Introduced `FlashParameters` to encapsulate flash-specific settings (e.g., tolerances, iteration limits), now exposed via XML for user configurability.
- **EOS Integration**: Both equations of state (EOS) are now used to assess mixture instability.
- **Shared Utilities**: Added `Utilities.hpp` to house common functions used in both flash and stability calculations.
- **Testing**: A new unit test validates the full flash model, ensuring correctness and robustness.


paveltomin (approved):

herve-gross (approved):

.integrated_tests.yaml
```diff
@@ -1,6 +1,6 @@
 baselines:
   bucket: geosx
-  baseline: integratedTests/baseline_integratedTests-pr3746-12509-e98985d
+  baseline: integratedTests/baseline_integratedTests-pr3662-12531-02dadd8
```
dkachuma:
```suggestion
  baseline: integratedTests/baseline_integratedTests-pr3662-12537-63d0314
```

jhuang2601 (approved):

rrsettgast (approved):

PR #3666, opened 2025-05-14
refactor: split functionality of addToFractureMesh

OmarDuran:
To enable dedicated interface conditions in the physics solvers, it's important to separate the attributions of the `addToFractureMesh` function. The adopted solution is to introduce an auxiliary function, `addToSurfaceMesh`, which handles only the `faceElementsToCells` mappings.

To preserve the original functionality of addToFractureMesh, the following steps are taken:

1.  Call to addToSurfaceMesh to avoid duplicating code
2. Set the rupture time
3. Build fracture-specific edge-to-face connectivity
4. Compute geometry
5. Update set memberships

### DIFF
--- a/src/coreComponents/mesh/SurfaceElementRegion.cpp
+++ b/src/coreComponents/mesh/SurfaceElementRegion.cpp
@@ -89,11 +89,10 @@ void SurfaceElementRegion::initializePreSubGroups()
 
 localIndex SurfaceElementRegion::addToSurfaceMesh( FaceManager const * const faceManager,
                                                    ArrayOfArraysView< localIndex const >  const & originalFaceToEdgeMap,
-                                                   localIndex const faceIndices[2],
-                                                   SortedArray< localIndex > & connectedEdges )
+                                                   localIndex const faceIndices[2])
 {
   localIndex rval = -1;
-
+  SortedArray< localIndex > connectedEdges;
   arrayView2d< localIndex const > const faceToElementRegion = faceManager->elementRegionList();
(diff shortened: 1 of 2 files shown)


rrsettgast (approved):

src/coreComponents/mesh/SurfaceElementRegion.cpp
```diff
@@ -185,10 +178,27 @@ localIndex SurfaceElementRegion::addToFractureMesh( real64 const time_np1,
...
+
+  FaceElementSubRegion & subRegion = this->getUniqueSubRegion< FaceElementSubRegion >();
+  arrayView1d< real64 > const ruptureTime = subRegion.getField< fields::ruptureTime >();
+  ruptureTime( kfe ) = time_np1;
+
+  SurfaceElementSubRegion::EdgeMapType & edgeMap = subRegion.edgeList();
   // Fill the connectivity between FaceElement entries. This is essentially a copy of the
   // edgesToFaces map, but with differing offsets.
-  for( auto const & edge : connectedEdges )
+  for( localIndex a = 0; a < edgeMap.sizeOfArray( kfe ); ++a )
```
rrsettgast: This needs to be identical to the previous definition, which is a loop over `connectedEdges`, which is set using `originalFaceToEdgeMap` rather than `subRegion.edgeList`.

OmarDuran: Thank you @rrsettgast, I made it identical to the previous implementation.

PR #3667, opened 2025-05-15
feat: Add option for computing the imaging condition for RTM

sframba:
This PR adds the possibility of computing the classic imaging condition for RTM seismic imaging. This reuses the machinery for the FWI gradient, changing the integral for the imaging condition.


acitrain (approved):

MelReyCG (approved):
(approving the unit-tests part)

PR #3671, opened 2025-05-16
feat: add map/unordered_map bounds checking

arng40:
This PR adds further bounds checking in GEOS.

- Add missing stdVector
- std::map/unordered_map wrapper that has bounds checking for ::operator[]( index )
- disallow use of map/unordered_map::operator[] to perform insertion.

In the next PR we will introduce the new aliases (i.e. stdUnorderedMap, stdMap) for performing bounds_checking


src/coreComponents/common/StdContainerWrappers.hpp
```diff
@@ -66,83 +92,129 @@ class StdVectorWrapper : public std::vector< T, Allocator >
...
+{
+public:
+  using stdMap< TKEY, TVAL >::stdMap;
+};
+
+template< typename TKEY, typename TVAL >
+class mapBase< TKEY, TVAL, std::integral_constant< bool, false > > : public stdUnorderedMap< TKEY, TVAL >
+{
+  using stdUnorderedMap< TKEY, TVAL >::stdUnorderedMap;
+};
```
MelReyCG, line 216: - Do you think that would possible to rename mapBase to `mapType` (it isn't really the base class of anything)?
- Couldn't the `SORTED` parameter be a simple `bool`?
- I would use those types only when strictly needed.
```suggestion
template< typename TKEY, typename TVAL, bool SORTED >
class mapType
{};

/// @cond DO_NOT_DOCUMENT
template< typename TKEY, typename TVAL >
class mapType< TKEY, TVAL, true > : public stdMap< TKEY, TVAL >
{
public:
  using stdMap< TKEY, TVAL >::stdMap;
};

template< typename TKEY, typename TVAL >
class mapType< TKEY, TVAL, false > : public stdUnorderedMap< TKEY, TVAL >
{
  using stdUnorderedMap< TKEY, TVAL >::stdUnorderedMap;
};
```

arng40:
@rrsettgast The code compile everywhere except in constitutive/unitsTests. 
For example in `testMultiFluidDeadOil.hpp` 

for Ubuntu CUDA I have this error :
```
StdContainerWrappers.hpp:40:45: error: recursive evaluation of default argument
using std::template vector< T, Allocator> ::vector;
                                             ^
constitutive/unitTests/testMultiFluidBlackOil.cpp:80:2: note: in instantiation of member function 'geos::testing::MultiFluidBlackOilTestFixture<false>::fillPhysicalProperties' requested here
(fillPhysicalProperties)(fluid); 
````

**Comment by arng40** [comment]
I think that the compil issue might be related to inheriting constructors

**Review by MelReyCG — COMMENTED** [review]
Is it normal that there still are some direct usage of `std::vector` & `std::map`? (`unitTests/`, `event/`, `fileIO/`...)

File: src/coreComponents/dataRepository/xmlWrapper.cpp [hunk 2]
```diff
@@ -22,6 +22,7 @@
 #include "xmlWrapper.hpp"
 
 #include "common/format/StringUtilities.hpp"
+#include "common/StdContainerWrappers.hpp"
```
MelReyCG: why adding this header here?

src/coreComponents/constitutive/unitTests/testMultiFluidCO2Brine.cpp
```diff
@@ -174,10 +174,10 @@ void MultiFluidCO2BrineTestFixture< BRINE, FLASH, THERMAL >::fillPhysicalPropert
 {
   dataRepository::Group & group = fluid;
 
-  auto & phaseNames = group.getReference< string_array >( MultiFluidBase::viewKeyStruct::phaseNamesString() );
+  string_array & phaseNames = group.getReference< string_array >( MultiFluidBase::viewKeyStruct::phaseNamesString() );
   phaseNames = {"gas", "liquid"};
 
-  auto & compNames = group.getReference< string_array >( MultiFluidBase::viewKeyStruct::componentNamesString() );
+  string_array & compNames = group.getReference< string_array >( MultiFluidBase::viewKeyStruct::componentNamesString() );
```
MelReyCG: Now that we found the compilation issue, I think you can revert that.

src/coreComponents/constitutive/unitTests/CMakeLists.txt
```diff
@@ -16,8 +16,8 @@ set( gtest_geosx_tests
      testLohrenzBrayClarkViscosity.cpp
      testModifiedCamClay.cpp
      testMultiFluidBlackOil.cpp
-     testMultiFluidCO2Brine.cpp
      testMultiFluidTwoPhaseCompositionalMultiphase.cpp
+     testMultiFluidCO2Brine.cpp
      testMultiFluidDeadOil.cpp
```
MelReyCG: `C` < `T`

MelReyCG:
I think we will need one more PR for `std::array<>`.
@arng40  @rrsettgast

src/coreComponents/common/StdContainerWrappers.hpp
```diff
@@ -21,15 +37,97 @@ namespace internal
...
+    : std::vector< T, Allocator >( std::move( vec )) {}
+
+  template< typename U, typename A >
+  StdVectorWrapper( std::vector< U, A > & vec )
+    : std::vector< T, A >( vec.begin(), vec.end()) {}
+  /// @endcond
+
   /**
    * Access element at index with bounds checking if USE_STD_CONTAINER_BOUNDS_CHECKING is true.
    * Otherwise, uses operator[] for unchecked access.
```
MelReyCG, line 87:
```suggestion
   * Access element at index with bounds checking if USE_BOUNDS_CHECKING is true.
   * Otherwise, uses operator[] for unchecked access.
   * @throws std::out_of_range if index is out of bounds.
```
(and same below)

src/coreComponents/common/StdContainerWrappers.hpp
```diff
@@ -66,83 +164,129 @@ class StdVectorWrapper : public std::vector< T, Allocator >
...
+{
+public:
+  using stdMap< TKEY, TVAL >::stdMap;
+};
+
+template< typename TKEY, typename TVAL >
+class mapBase< TKEY, TVAL, std::integral_constant< bool, false > > : public stdUnorderedMap< TKEY, TVAL >
+{
+  using stdUnorderedMap< TKEY, TVAL >::stdUnorderedMap;
+};
```
MelReyCG: I think we could simplify this part:
- about the `mapBase` naming, there is no real "base" concept here, nor need for a class here,
- why not simplifying the `std::integral_constant` to a simple `bool`?
- a pattern like the `std::*_t` would be more standard.

I would suggest:
```c++
namespace internal
{

template< typename TKEY, typename TVAL, bool SORTED >
struct stdMapTypeDefinition
{};

template<typename TKEY, typename TVAL>
struct stdMapTypeDefinition< TKEY, TVAL,true >
{ using Type = stdMap< TKEY, TVAL >; };

template<typename TKEY, typename TVAL>
struct stdMapTypeDefinition< TKEY, TVAL,false >
{ using Type = stdUnorderedMap< TKEY, TVAL >; };

} // namespace internal

/**
 * @brief templated type for ordered and unordered maps.
 * @tparam TKEY key type
 * @tparam TVAL value type
 * @tparam SORTED a bool indicating whether map is ordered
 */
template< typename TKEY, typename TVAL, bool SORTED >
using stdMapType = internal::stdMapTypeDefinition<TKEY,TVAL,SORTED>::Type;

```

src/coreComponents/common/StdContainerWrappers.hpp
```diff
@@ -66,83 +164,129 @@ class StdVectorWrapper : public std::vector< T, Allocator >
...
+ * @tparam MapType The type of the underlying map (e.g., std::map).
  * @tparam Allocator Allocator type for the vector.
+ * @tparam USE_STD_CONTAINER_BOUNDS_CHECKING A boolean flag to enable or disable bounds checking.
  */
-template< typename T, typename Allocator = std::allocator< T > >
-using stdVector = std::vector< T, Allocator >;
-#endif
+template< typename MapType,
+          bool USE_BOUNDS_CHECKING = false
+          >
```
MelReyCG:
```suggestion
template< typename MapType,
          bool USE_BOUNDS_CHECKING = false >
```

src/coreComponents/common/StdContainerWrappers.hpp
```diff
@@ -66,83 +164,129 @@ class StdVectorWrapper : public std::vector< T, Allocator >
...
+  {
+    if constexpr (USE_BOUNDS_CHECKING)
+    {
+      return this->at( key );
+    }
+    else
+    {
+      return Base::operator[]( key );
+    }
+  }
```
MelReyCG: Where the behaviour of these operators is tested in fonction of all possible `USE_BOUNDS_CHECKING` values? (same remark for `stdVector`, except that it cannot be tested against `USE_BOUNDS_CHECKING=false`)
If they are not tested, we can add them in `testDataTypes.cpp`.

MelReyCG (changes requested):

src/coreComponents/dataRepository/xmlWrapper.cpp
arng40: removed

src/coreComponents/common/StdContainerWrappers.hpp
arng40: They would be a lot of changements in BufferOps.hpp imo

arng40: I renamed `mapBase` to `mapType`
I can't change `typename SORTED` to `bool SORTED`, I tried and it required a lot of work for this PR and maybe when we will jump into c++20 for concepts

src/coreComponents/common/StdContainerWrappers.hpp
arng40: add a test in `testDataTypes.cpp`

rrsettgast:
@arng40 Do you think it would be good idea to introduce the aliases (i.e. stdUnorderedMap, stdMap) in a separate PR which just aliases to std::unordered_map, and std::map so that we can review the actual changes independently from this swap...which should result in no behavioral changes with develop?

arng40:
> @arng40 Do you think it would be good idea to introduce the aliases (i.e. stdUnorderedMap, stdMap) in a separate PR which just aliases to std::unordered_map, and std::map so that we can review the actual changes independently from this swap...which should result in no behavioral changes with develop?

@rrsettgast 
I don't see any problem in splitting into 2 PR.

src/coreComponents/common/StdContainerWrappers.hpp
MelReyCG: @arng40 and same bellow: use `@throws` in doxygen, inline comments are not usually in use in this project (*comments in implementations are not a specification*)

src/coreComponents/common/StdContainerWrappers.hpp
MelReyCG: OK, so let's keep `typename SORTED` for now.

Can you just keep it in `DataTypes.hpp` rather than in `StdContainerWrappers.hpp`?

src/coreComponents/common/StdContainerWrappers.hpp
MelReyCG: nice :)

arng40:
> std::vector

They are all under PVTPackage or package that don't include DataType.hpp

src/coreComponents/codingUtilities/tests/testGeosxTraits.cpp
```diff
@@ -52,11 +52,11 @@ TEST( testGeosxTraits, HasMemberFunction_at )
   static_assert( HasMemberFunction_at< string >, "Should be true." );
   static_assert( HasMemberFunction_at< stdVector< int > >, "Should be true." );
   static_assert( HasMemberFunction_at< stdVector< double > >, "Should be true." );
-  static_assert( HasMemberFunction_at< map< int, string > >, "Should be true." );
-  static_assert( HasMemberFunction_at< std::unordered_map< int, string > >, "Should be true." );
+  static_assert( HasMemberFunction_at< std::map< int, string > >, "Should be true." );
+  static_assert( HasMemberFunction_at< std::map< int, string > >, "Should be true." );
```
MelReyCG: `static_assert( HasMemberFunction_at< stdMap< int, string > >, "Should be true." );` & same for `static_assert( HasMemberFunction_at< stdUnorderedMap< int, string > >, "Should be true." );`?

MelReyCG: and same bellow for this file?

MelReyCG: at least leave the `unordered_map` test

MelReyCG (approved):

rrsettgast (approved):
The only issue I see is that when we get a throw from usage of `at()` the stack trace isn't very nice.

paveltomin:
this still needs reviews and there is an error in "Ubuntu CUDA debug" build

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/PVTFunctionHelpers.hpp
```diff
@@ -136,8 +136,8 @@ class PTTableCoordinates
 
   array1d< array1d< real64 > > const & getCoords() const { return coords; }
 
-  static const inline stdVector< units::Unit > coordsUnits =
-  { units::Pressure, units::TemperatureInC };
+  static const inline stdVector< units::Unit > coordsUnits = stdVector< units::Unit >(
+    { units::Pressure, units::TemperatureInC } );
```
wrtobin:
```suggestion
  static const inline stdVector< units::Unit > coordsUnits{ units::Pressure, units::TemperatureInC };
```

wrtobin: this *should* still work, yeah?

arng40: still working

arng40:
All bugs have been solved and ready to merge,
@rrsettgast @paveltomin @wrtobin  @dkachuma

src/coreComponents/mesh/mpiCommunications/NeighborCommunicator.hpp
```diff
@@ -277,13 +277,13 @@ class NeighborCommunicator
     return m_sendBuffer[commID];
   }
 
-  void resizeSendBuffer( int const commID, int const newSize )
+  void resizeSendBuffer( std::vector< signed char >::size_type commID, int const newSize )
```
rrsettgast: I think we have a `size_t` defined for things like this

rrsettgast (approved):

rrsettgast:
@wrtobin Can you please review this?

src/coreComponents/common/StdContainerWrappers.hpp
```diff
@@ -106,83 +127,179 @@ class StdVectorWrapper : public std::vector< T, Allocator >
...
+  MappedType const & operator[]( KeyType const & key ) const
+  {
+    if constexpr (USE_BOUNDS_CHECKING)
+    {
+      return this->at( key );
+    }
+    else
+    {
+      return Base::operator[]( key );
+    }
```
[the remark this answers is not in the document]
wrtobin: this does seem correct, 
`operator[]` is not const in the stl since it may or may not increase the size of the map. 
have we successfully compiled with `USE_BOUNDS_CHECK == false`?

src/coreComponents/mesh/mpiCommunications/NeighborCommunicator.hpp
wrtobin: Yeah I can't see a reason why the `commID` should be dependent on the `size_type` of a completely unrelated type?

src/coreComponents/common/StdContainerWrappers.hpp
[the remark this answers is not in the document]
wrtobin: or rather do we want this to maintain the compilation failure behavior of attempting to use `operator[]` on a const map type? we might want to leave this in that case, but it would only fail when `USE_BOUNDS_CHECK == false`, which does just tend to complicate usage of the type, which we want to generally be a thin wrapper only.

wrtobin (approved):
I think this is good except for 1-2 already-flagged issues I've added feedback to.

src/coreComponents/common/StdContainerWrappers.hpp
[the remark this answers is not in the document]
wrtobin: this introduces the possibility of deferring what should be a compile-time failure to instead cause a runtime failure, which I think is probably not the way to go.
I think just dropping the `const` version of the operator might be the best solution, since that makes our wrapper interface semantically identical to the STL map.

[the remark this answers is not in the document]
arng40: > this does seem correct, `operator[]` is not const in the stl since it may or may not increase the size of the map. have we successfully compiled with `USE_BOUNDS_CHECK == false`?

In both release & debug mode, it compile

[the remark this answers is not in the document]
rrsettgast: @wrtobin I think we should drop this overload. That is the stl behavior preserving thing to do?

rrsettgast:
@paveltomin Can you review?

Testing/Temporary/LastTest.log
```diff
@@ -0,0 +1,3 @@
+Start testing: Jun 04 14:40 CEST
```
paveltomin: what is that file?

src/coreComponents/mesh/mpiCommunications/NeighborCommunicator.hpp
```diff
@@ -283,7 +283,7 @@ class NeighborCommunicator
     m_sendBuffer[commID].resize( newSize );
   }
 
-  void resizeRecvBuffer( int const commID, int const newSize )
+  void resizeRecvBuffer( std::vector< signed char >::size_type commID, int const newSize )
```
paveltomin: why do you need to change this?

src/coreComponents/mesh/unitTests/testMeshObjectPath.cpp
```diff
@@ -161,7 +161,7 @@ TEST( testMeshObjectPath, meshObjectPathConstuction )
   Group const & meshBodies = testMesh.meshBodies();
 
 
-  std::map< string, MeshObjectPath::permutationMapType > pathsAndResults =
+  map< string, MeshObjectPath::permutationMapType > pathsAndResults =
```
paveltomin, line 164: just "map" is quite ambiguous

src/coreComponents/schema/schema.xsd
```diff
@@ -1399,7 +1399,7 @@ Information output from lower logLevels is added with the desired log level
 		<xsd:attribute name="direction" type="R1Tensor" default="{0,0,0}" />
 		<!--endTime => Time at which the boundary condition will stop being applied.-->
 		<xsd:attribute name="endTime" type="real64" default="1e+99" />
-		<!--functionName => Name of a function that specifies the variation of the production rate variations of this SourceFlux.Multiplied by scale. If no function is provided, a constant value of 1 is used.The producted fluid rate unit is in kg by default, or in mole if the flow solver has a useMass of 0.-->
+		<!--functionName => Name of a function that specifies the variation of the production rate variations of this SourceFlux.Multiplied by scale. If no function is provided, a constant value of 1 is used.The produced fluid rate unit is in kg by default, or in mole if the flow solver uses moles.-->
```
paveltomin, line 1402: why in this PR?

src/coreComponents/common/StdContainerWrappers.hpp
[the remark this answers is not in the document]
MelReyCG: Wasn't the goal to remove the ability to insert with `operator[]`? (as I see in the PR description and the doxygen documentation)
Do we want different behaviours in function of `USE_BOUNDS_CHECK`?
What is the issue when `StdMapWrapper::operator[]` always rely on `at()`? (thus keeping const version & sending exception on issue)
If we drop the const version, and allow insertion on `USE_BOUNDS_CHECK==false`, it not do less, but more computation (despite this is a classic motivation of disabling bound checking).

Anyway, if we finally decide that `operator[]` is made to add elements (on `USE_BOUNDS_CHECK == false` or always), documentation must be slightly modified to reflect the code behaviour @arng40.

paveltomin (approved):

Testing/Temporary/LastTest.log
arng40: deleted

src/coreComponents/mesh/mpiCommunications/NeighborCommunicator.hpp
arng40: revert

src/coreComponents/mesh/unitTests/testMeshObjectPath.cpp
arng40: in the following PR i will change the naming

dkachuma (approved):

PR #3673, opened 2025-05-16
feat: option to cut the timestep when updateState failed

paveltomin:
activated by `chopWhenUpdateStateFailed` flag in the solver


src/coreComponents/physicsSolvers/fluidFlow/kernels/compositional/FluidUpdateKernel.hpp
```diff
@@ -34,38 +34,42 @@ namespace thermalCompositionalMultiphaseBaseKernels
...
-  static void
+  static bool
   launch( SortedArrayView< localIndex const > const & targetSet,
           FLUID_WRAPPER const & fluidWrapper,
           arrayView1d< real64 const > const & pres,
           arrayView1d< real64 const > const & temp,
           arrayView2d< real64 const, compflow::USD_COMP > const & compFrac )
   {
-    forAll< POLICY >( targetSet.size(), [=] GEOS_HOST_DEVICE ( localIndex const a )
+    RAJA::ReduceBitAnd< ReducePolicy< POLICY >, bool > status = true;
```
dkachuma, line 63: I'm not sure if this forces a synchronisation or not. Is there a way to avoid this?

jhuang2601 (approved):

tjb-ltk:
general comment.  @dkachuma  @paveltomin ..  there is a duality to the roles and responsibilities for the fluid component. seems like state variables are input to fluid model, but the fluid model owns the derived property/derivative arrays? .. so the client has to pull on fluid model for some props...   Instead of using return types for flash status on compute/update, why not put queries on the component to get flash success or max sat changes ?  The api might be more stable, the component can export new functionality , and client can pull as needed.

PR #3674, opened 2025-05-19
refactor: Fluid kernel

dkachuma:
Fluid update kernel (for #3655)
- Splits the `FluidUpdateKernel` into the declaration (`FluidUpdateKernel.hpp`) and the implementation (`FluidUpdateKernel_impl.hpp`)
- This ensures that the calls in `CompositionalMultiphaseBase.cpp` and `CompositionalMultiphaseWell.cpp` do not get to expand over all the fluid models. This should ease the burden of compiling these files.
- The implementation for specific fluids is explicitly instantiated via the kernel auto generated files
- `CompositionalMultiphaseWell` is changed to use all serial policy.


src/coreComponents/physicsSolvers/fluidFlow/kernelSpecs.json
```diff
@@ -215,5 +215,36 @@
...
+      "FLUID"
+    ],
+    "constants": [
+    ],
+    "combinations": {
+      "NAME": [
+        "FluidUpdateKernel"
+      ],
+      "POLICY": [
+        "serialPolicy",
```
CusiniM, line 233: do we need this? you d think this is always `parallelDevicePolicy<>`

dkachuma: There is this line ([CompositionalMultiphaseWell.cpp:1055](https://github.com/GEOS-DEV/GEOS/blob/develop/src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp#L1055)) in the well model that does an explicit serial call. These well launches are quite small but we can still shift them to the GPU I suppose.

tjb-ltk: The well model can change that line parallelDevicePolicy, change as if it helps....  but it would be nice to eventually test the performance of this pattern .  

Does the well models use of serialPolicy force the reservoir to use same policy , is this another instantiation that needs to be defined...  the well can be changed to use parallelDEvicePolicy

tjb-ltk: How does this work... CMAKE reads the json file and generates all the cpp files to instantiate the kernels ?  or is there some python for code generation..  ... found it generateKernels, nice... it would be good to reduce compile times of other kernels outside fluid model as well !!

src/coreComponents/physicsSolvers/fluidFlow/kernels/compositional/FluidUpdateKernel_impl.hpp
tjb-ltk, line 1: Is. fluidWrapper.numGauss() anything other than 1 ?

dkachuma: I think currently it is always 1. Potentially it could be more than 1 for discretisations other than TPFA I suppose.

paveltomin: it depends on if FE is used or not for that cell - if FE is used, constitutive has to fill all gauss points
so basically if mechanics is present - `numGauss` will be > 1 even for TPFA

src/coreComponents/physicsSolvers/fluidFlow/kernelSpecs.json
dkachuma: The requirement for a `serialPolicy` in the well does not force the reservoir to use the same. The 2 are independent. Changing the well to `parallelDevicePolicy` would simplify the code (and the compilation because we wouldn't need to include the extra generated files). However, I wouldn't want to do that at the cost of performance.

On the other hand this ([CompositionalMultiphaseWell.cpp:858-870](https://github.com/GEOS-DEV/GEOS/blob/develop/src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp#L858-L870)) uses the fluid prescribed policy which in most cases is `parallelDevicePolicy`. Not sure what the difference between the 2 launches is.

tjb-ltk: The later is used when the well is first turned on.  Connected res cell vals are used to initialize segment values, then fluid props are then computed .    the launches should be the same   at present a serial police would be fine... unless there's no overhead when problem size is on order of number of segments..

dkachuma: I've updated the call to use a serial policy.

paveltomin (approved):

PR #3675, opened 2025-05-19
Feature/bd713/clean partitioner

bd713:


src/coreComponents/mesh/mpiCommunications/PartitionerBase.hpp
```diff
@@ -0,0 +1,193 @@
...
+
+
+  int getNumPartitions() const
+  {
+    return m_numPartitions;
+  }
+  int getNumPartitions()
+  {
+    return m_numPartitions;
+  }
```
CusiniM: you don't need `non-const getters`. It would make sense if you were returning a reference but I would not do that and add a setter instead if you need one.

PR #3677, opened 2025-05-20
feat: encountered negative pressure/density cells summary output

MelReyCG:
This PR aims at adding a partial logging of the element ids that have negative pressures / densities, to allow the user to quickly see where negative pressure / densities are computed.
To enable the output, it is proposed to set the `logLevel` at least to 2 on the flow solver.

Here is an exemple of an output (from a synthetic case):
```
        compflow: 8 negative pressure values encountered. Minimum value: -126475.570 [Pa].
                  ---------------------------------------------------
                  |      Summary of negative pressure elements      |
                  |-------------------------------------------------|
                  |--------------Rank 0, 2 / 2 values---------------|
                  |     Global Id |              0 |          13200 |
                  | pressure [Pa] | -126475.570318 | -126475.570318 |
                  |               |                |                |
                  |--------------Rank 2, 2 / 2 values---------------|
                  |     Global Id |          13140 |          26340 |
                  | pressure [Pa] | -126475.570318 | -126475.570318 |
                  |               |                |                |
                  |--------------Rank 3, 2 / 2 values---------------|
                  |     Global Id |             59 |          13259 |
                  | pressure [Pa] | -126475.570318 | -126475.570318 |
                  |               |                |                |
                  |--------------Rank 5, 2 / 2 values---------------|
                  |     Global Id |          13199 |          26399 |
                  | pressure [Pa] | -126475.570318 | -126475.570318 |
                  |               |                |                |
                  ---------------------------------------------------
```
A standard component is proposed to do so this kind of "limited kernel output": the `ElementReporterOutput` & . Its goal is to:
- `ElementReporterBuffer`, allocate global memory & be able to disable the whole system (ie: by a `logLevel`),
- `ElementReporterCollector`, manage the atomic counting & collection of elements,
- `ElementReporterOutput`, output the result (i.e. in the log),
If the number of reported elements exceed the rank `ElementReporterBuffer` size, the row ends with `...`, but the elements still gets counted.

----

This table has been drawn by the new `TableTextMpiOutput` (also proposed in this PR), which allows to draw a single table on the rank 0 from multiple ranks:
```c++
        TableLayout const layout = TableLayout().
                                     setTitle( GEOS_FMT( "Summary of {} elements", valueNaming ) ).
        TableData data;
        
        // new TableMpiLayout for customization
        TableMpiLayout mpiLayout;
        mpiLayout.m_separatorBetweenRanks = true; // visually separate each ranks data

        if( signaledCount > 0 ) // <- if some ranks have an empty TableData, they will be ignored
        {
          // title of each rank table part
          mpiLayout.m_rankTitle = GEOS_FMT( "Rank {}, {} / {} values",
                                            MpiWrapper::commRank(), collectedCount, signaledCount );

          // filling rank data...
          data.addRow( stdVector< TableData::CellData >( tableColumnsCount, { CellType::MergeNext, "" } ) );
          cells[globalIdLine].front() = { CellType::Value, "Global Id" };
          // [...]
        }

        // draw the ranks table
        TableTextMpiOutput const formatter = TableTextMpiOutput( layout, mpiLayout );
        formatter.toStream( std::cout, data );
```


src/coreComponents/common/format/table/TableMpiComponents.hpp
```diff
@@ -0,0 +1,109 @@
...
+   * @param tableLayout Contain all tableColumnData names and optionnaly the table title
+   * @param mpiLayout MPI-specific layout information (default is having contiguous ranks data).
+   */
+  TableTextMpiOutput( TableLayout const & tableLayout,
+                      TableMpiLayout mpiLayout = TableMpiLayout() );
+
+  /**
+   * @brief Convert a data source to a table string.
+   * @param tableData The data source to convert.
+   * @param outputStream The target output stream for rank 0, to output the table strin grepresentation
```
arng40: fix the word

src/coreComponents/common/format/table/TableMpiComponents.cpp
```diff
@@ -0,0 +1,156 @@
...
+{
+  TableTextMpiOutputStatus status {
+    // m_isMasterRank (only the master rank does the output of the header && bottom of the table)
+    MpiWrapper::commRank() == 0,
+    // m_isContributing (some ranks does not have any output to produce)
+    !tableData.getCellsData().empty(),
+    // m_sepLine
+    ""
+  };
+
```
arng40: for clarity, could we declare  `status` as const and `m_sepLine` as a separated variable ?

src/coreComponents/common/format/table/TableMpiComponents.cpp
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  // all other ranks than rank 0 render their output in a string and comunicate its size
+  auto ranksStrsSizes = std::vector( ranksCount, 0 );
+  string const rankStr = !status.m_isMasterRank && status.m_isContributing ? localStringStream.str() : "";
+  integer const rankStrSize = rankStr.size();
+  MpiWrapper::allgather( &rankStrSize, 1, ranksStrsSizes.data(), 1 );
+
+  // we compute the memory layout of the ranks strings
+  auto ranksStrsDisps = std::vector( ranksCount, 0 );
```
arng40: ranksStrsDisplacements, not obvious on first reading😅

src/coreComponents/common/format/table/TableMpiComponents.cpp
MelReyCG: This serves to have a temporary, local, non-const structure that we can pass in the functions (to avoid passing a lot of parameters or recomputing the same logics). That will scale nicely if we want to add more local variables.
But I can set m_isMasterRank & m_isContributing to const if that is clearer.

src/coreComponents/common/format/table/TableMpiComponents.cpp
MelReyCG: OK, I renamed to `ranksStrsOffsets`, I think that is even more clear.

MelReyCG: I also used explicit types instead of `auto`

arng40 (approved):

src/coreComponents/common/format/table/TableFormatter.cpp
```diff
@@ -146,37 +146,41 @@ string TableTextFormatter::toString< TableData >( TableData const & tableData )
...
-                       tableTotalWidth );
-  outputTable( m_tableLayout, tableOutput,
-               headerCellsLayout, dataCellsLayout,
-               tableTotalWidth );
+                       tableTotalWidth, nullptr );
+
+  string const sepLine = string( tableTotalWidth, m_horizontalLine );
+  outputTableHeader( tableOutput, m_tableLayout, headerCellsLayout, sepLine );
+  outputTableData( tableOutput, m_tableLayout, dataCellsLayout );
+  outputTableBottom( tableOutput, m_tableLayout, sepLine, !dataCellsLayout.empty() );
```
paveltomin: footer?

src/coreComponents/common/format/table/TableFormatter.cpp
```diff
@@ -146,37 +146,41 @@ string TableTextFormatter::toString< TableData >( TableData const & tableData )
...
                                               CellLayoutRows & dataCellsLayout,
-                                              size_t & tableTotalWidth ) const
+                                              size_t & tableTotalWidth,
+                                              ColumnWidthModifier columnWidthModifier ) const
 {
+  RowsCellInput const & inputDataValues( tableInputData.getCellsData() );
   bool const hasColumnLayout = tableLayout.getColumnLayersCount() > 0;
-  RowsCellInput const & inputDataValues( tableInputData.getTableDataRows() );
   size_t const inputDataRowsCount = !inputDataValues.empty() ? inputDataValues.front().size() : 0;
+  size_t nbVisibleColumns = std::max( size_t( 1 ), ( hasColumnLayout ?
```
paveltomin: `size_t( 1 )` - do you really need `size_t` here?

paveltomin: `nbVisibleColumns` can be const?

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.cpp
```diff
@@ -863,76 +868,78 @@ bool CompositionalMultiphaseFVM::checkSystemSolution( DomainPartition & domain,
...
 
-        minPres  = std::min( minPres, subRegionData.localMinPres );
-        minDens = std::min( minDens, subRegionData.localMinDens );
-        minTotalDens = std::min( minTotalDens, subRegionData.localMinTotalDens );
-        numNegPres += subRegionData.localNumNegPressures;
-        numNegDens += subRegionData.localNumNegDens;
+        minPres  = std::min( minPres, subRegionData.localMinNegPres );
+        minDens = std::min( minDens, subRegionData.localMinNegDens );
+        minTotalDens = std::min( minTotalDens, subRegionData.localMinNegTotalDens );
         numNegTotalDens += subRegionData.localNumNegTotalDens;
```
paveltomin: why `numNegPres` and `numNegDens` got removed but `numNegTotalDens` was not?

herve-gross (approved):

src/coreComponents/physicsSolvers/fluidFlow/kernels/SolutionCheckKernelsHelpers.hpp
```diff
@@ -0,0 +1,136 @@
...
+
+      if( outputStart < m_elementsBuffer.size() )
+      {
+        m_elementsBuffer[outputStart].m_id = m_localToGlobalId[report.m_id];
+        m_elementsBuffer[outputStart].m_value = report.m_value;
+      }
+    }
+  }
+
+  // // currently unused version for adding multiple ids from a given kernel
```
paveltomin, line 99: remove?

src/coreComponents/physicsSolvers/fluidFlow/kernels/compositional/SolutionCheckKernel.hpp
```diff
@@ -84,44 +87,55 @@ class SolutionCheckKernel : public SolutionScalingAndCheckingKernelBase< integer
...
     m_scalingFactor( scalingFactor ),
-    m_scalingType( scalingType )
+    m_scalingType( scalingType ),
+    m_negPressureIds( negPressureIds ),
+    m_negDensityIds( negDensityIds )
   {}
 
   /**
-   * @struct StackVariables
+   * @struct KernelStats
```
paveltomin, line 99: why renamed?

src/coreComponents/physicsSolvers/fluidFlow/kernels/singlePhase/SolutionCheckKernel.hpp
```diff
@@ -53,14 +62,14 @@ struct SolutionCheckKernel
 
         if( newPres < 0.0 )
         {
-          numNegativePressures += 1;
-          minPres.min( newPres );
+          minNegPres.min( newPres );
+          negPressureIds.collectElement( atomicPolicy{}, { ei, newPres } );
```
paveltomin, line 66: will it always collect, no matter if it was requested or not?
how about performance impact?

src/coreComponents/physicsSolvers/fluidFlow/wells/LogLevelsInfo.hpp
```diff
@@ -66,6 +66,12 @@ struct WellValidity
   static constexpr std::string_view getDescription() { return "Well validity information"; }
 };
 
+struct WellValidityDetails
```
paveltomin: should be just `SoutionDetails`, nothing special about the well

paveltomin (approved):
looks ok, hopefully no performance impact

src/coreComponents/common/format/table/TableFormatter.cpp
MelReyCG: - `size_t( ... )` is needed here as c++17 does not have the "uz" postfix yet (c++23). We could define it though.
-  it can effectively be const 👍

src/coreComponents/common/format/table/TableFormatter.cpp
MelReyCG: Good suggestion

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.cpp
MelReyCG: ~You're absolutely right, missed that one since I focused on pressure issues.~

src/coreComponents/physicsSolvers/fluidFlow/kernels/SolutionCheckKernelsHelpers.hpp
MelReyCG: I was wondering if we can keep commented code for a feature that could be needed later, as to give a recommendation of implementation.

paveltomin: i am fine with that

src/coreComponents/physicsSolvers/fluidFlow/kernels/compositional/SolutionCheckKernel.hpp
MelReyCG: My goal is not to rename, but to separate the return type `KernelStats` from the other computing variables `StackVariables`, to keep them "private".
We could also name it `SolutionCheckStats`.

src/coreComponents/physicsSolvers/fluidFlow/kernels/singlePhase/SolutionCheckKernel.hpp
MelReyCG: Collection of ids is disabled if `logLevel` is not high enough. In that case, the cost is which of incrementing the counter (which already existed), and a kernel constant branch evaluation which should not be noticiable.
```c++
    ElementsReporterBuffer rankNegPressureIds{ isLogLevelActive< logInfo::Solution >( getLogLevel() ),
                        /* 0 = disabled -> */  isLogLevelActive< logInfo::SolutionDetails >( getLogLevel() ) ? 16 : 0 };
```

But you are true, I need to benchmark it. If you see an interesting case in the repo, tell me.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.cpp
MelReyCG: Ignore my previous comment, it was for a reason:
I did not added the collection system for the negative total densities, as the ids will already be collected for the negative component densities.
Do you think I should uniformize here?

src/coreComponents/physicsSolvers/fluidFlow/kernels/compositional/SolutionCheckKernel.hpp
MelReyCG: Separating the two structure can improve readability on what is actually output from the kernel.

MelReyCG: (I edited my first answer, I tried another proposal, but ended up with the same code)

paveltomin: ok renaming sounds good to me, just wanted to check why

src/coreComponents/physicsSolvers/fluidFlow/kernels/singlePhase/SolutionCheckKernel.hpp
paveltomin: i think it's fine, thanks for the explanations

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.cpp
paveltomin: yeah, maybe not in this PR and not top priority but would be nice to unify

MelReyCG:
@paveltomin I did a benchmark with a real case of 1M+ cells, with the different `logLevel` values (0, 1, 2), on CPU & GPU. This PR has no detectable performance impact.

dkachuma (approved):
Looks good to me.
Q: I would expect the global ids to be unique so what is the reason for separating the report into different ranks? Can't we just write one "table" that summarises over all ranks?

wrtobin (approved):
Looks good to me.

src/coreComponents/common/format/table/unitTests/testMpiTable.cpp
```diff
@@ -0,0 +1,149 @@
...
+      "-------------------------------------------\n"
+    },
+  };
+  for( TestCase const & testCase: testCases )
+  {
+    int const rankId = MpiWrapper::commRank();
+    int const nbRanks = MpiWrapper::commSize();
+    if( nbRanks > 1 )
+    {
+      ASSERT_EQ( nbRanks, 4 );
```
OmarDuran: This test fails unless it is run with exactly 4 MPI ranks. Consider making it more flexible or skipping it when nbRanks != 4.

OmarDuran (changes requested):
Thanks for the improvement, It looks good to me. I left a minor request.

src/coreComponents/common/format/table/unitTests/testMpiTable.cpp
MelReyCG: I picked the values in order to verify that the ordering of the rows remains consistant, I don't really want another set of values here (especially a set with less ranks).

In the `CMakeList.txt` that register this test, I set:
```
if( ENABLE_MPI )
  set( nranks 4 )
```
The part you mentionned is a guard to be sure this test is still ran with 4 ranks, and does not result in UB with unclear errors.

There are other tests in the repo that are designed to run with a fixed number of ranks, and they don't have any guard (`testGraphColoringMPI` at 8, `testVTKImport` at 2, both crash).

I can totally add a comment to explain the reason of this assertion check.

OmarDuran: This should address the comment, thanks!

OmarDuran (approved):

PR #3679, opened 2025-05-21
refactor: solvers cleanup

paveltomin:
- remove `maxStableDt` - not used anywhere
- remove `registerWrapper` for `meshTargets`
- added few more timers: `step setup`, `step complete`, `reset state` - matters more when update state is more expensive (such as compositional)
- added `setConstitutiveName` to unify
- switched integer flags to boolean in few places
- removed `FluidPropViews` in single phase solver, not really needed
- tried to reduce dependency on solvers in places where dependency is not really needed, most importantly in `SinglePhasePoromechanicsEFEM_impl.hpp` which propagates into kernels
- some other cleanup


src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -365,49 +365,17 @@ void CompositionalMultiphaseBase::registerDataOnMesh( Group & meshBodies )
     {
       if( m_hasCapPressure )
       {
-        subRegion.registerWrapper< string >( viewKeyStruct::capPressureNamesString() ).
-          setPlotLevel( PlotLevel::NOPLOT ).
-          setRestartFlags( RestartFlags::NO_WRITE ).
-          setSizedFromParent( 0 ).
-          setDescription( "Name of the capillary pressure constitutive model to use" ).
```
dkachuma, line 372: We have lost these descriptions. But I suppose they aren't user accessible.

src/coreComponents/physicsSolvers/PhysicsSolverBase.hpp
```diff
@@ -1124,6 +1123,19 @@ string PhysicsSolverBase::getConstitutiveName( ParticleSubRegionBase const & sub
...
+{
+  subRegion.registerWrapper< string >( wrapperName ).
+    setPlotLevel( dataRepository::PlotLevel::NOPLOT ).
+    setRestartFlags( dataRepository::RestartFlags::NO_WRITE ).
+    setSizedFromParent( 0 );
+
+  string & constitutiveName = subRegion.getReference< string >( wrapperName );
+  constitutiveName = getConstitutiveName< CONSTITUTIVE_TYPE >( subRegion );
+  GEOS_ERROR_IF( constitutiveName.empty(), GEOS_FMT( "{}: constitutive model not found on subregion {}",
+                                                     getDataContext(), subRegion.getName() ) );
```
dkachuma: Is there a way of getting what the constitutive type is (e.g. relperm, diffusion) ? I don't think `getDataContext()` will include this information.

dkachuma (approved):

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
paveltomin: yeah, those are internal things and not sure the description is very useful

src/coreComponents/physicsSolvers/PhysicsSolverBase.hpp
paveltomin: is something like that better:
```
***** ERROR
***** LOCATION: /home/ptls/GEOS/GEOS/src/coreComponents/physicsSolvers/PhysicsSolverBase.hpp:1136
***** Controlling expression (should be false): constitutiveName.empty()
***** Rank 0: compflow (co2_flux_3d.xml, l.5): N4geos12constitutive24RelativePermeabilityBaseE constitutive model not found on subregion cellBlock
```

paveltomin: changed it to wrapper name, and adjusted those strings to something more readable:
```
***** ERROR
***** LOCATION: /home/ptls/GEOS/GEOS/src/coreComponents/physicsSolvers/PhysicsSolverBase.hpp:1136
***** Controlling expression (should be false): constitutiveName.empty()
***** Rank 0: compflow (co2_flux_3d.xml, l.5): relative permeability constitutive model not found on subregion cellBlock
```

paveltomin:
rebaseline needed:
```
Error: /Problem/Solvers/compflow
	Group has a child 'maxStableDt' in the baseline file but not the file to compare.
Error: /Problem/Solvers/compflow
	Group has a child 'meshTargets' in the baseline file but not the file to compare.
```

src/coreComponents/physicsSolvers/PhysicsSolverBase.hpp
dkachuma: I like wrapper name here.

src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
```diff
@@ -39,7 +39,6 @@ PhysicsSolverBase::PhysicsSolverBase( string const & name,
   :
   ExecutableGroup( name, parent ),
   m_cflFactor(),
-  m_maxStableDt{ 1e99 },
```
CusiniM, line 43: are you sure wavesolvers don't use this for anything?

src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
```diff
@@ -80,11 +74,6 @@ PhysicsSolverBase::PhysicsSolverBase( string const & name,
                     "solver may be applied to these regions. The decision about what regions this solver will be"
                     "applied to rests in the EventManager." );
 
-  registerWrapper( viewKeyStruct::meshTargetsString(), &m_meshTargets ).
-    setInputFlag( InputFlags::FALSE ).
-    setRestartFlags( RestartFlags::NO_WRITE ).
-    setDescription( "MeshBody/Region combinations that the solver will be applied to." );
```
CusiniM, line 87: I believe this is used. isn't it how we do that loop... `forDiscretizationOnMeshTargets` ...

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
CusiniM: Isn't the description also used in the docs describing the c++ components though?  Maybe not for these because of where it's happening.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.hpp
```diff
@@ -314,6 +309,10 @@ class CompositionalMultiphaseBase : public FlowSolverBase
     static constexpr char const * maxSequentialCompDensChangeString() { return "maxSequentialCompDensChange"; }
     static constexpr char const * minScalingFactorString() { return "minScalingFactor"; }
 
+    static constexpr char const * relPermNamesString() { return "relative permeability"; }
+    static constexpr char const * capPressureNamesString() { return "capillary pressure"; }
```
CusiniM:
```suggestion
    static constexpr char const * relPermNamesString() { return "relativePermeability"; }
    static constexpr char const * capPressureNamesString() { return "capillaryPressure"; }
```

src/coreComponents/physicsSolvers/fluidFlow/FlowSolverBase.hpp
```diff
@@ -73,17 +73,18 @@ class FlowSolverBase : public PhysicsSolverBase
...
     static constexpr char const * allowNegativePressureString() { return "allowNegativePressure"; }
     static constexpr char const * maxAbsolutePresChangeString() { return "maxAbsolutePressureChange"; }
     static constexpr char const * maxSequentialPresChangeString() { return "maxSequentialPressureChange"; }
     static constexpr char const * maxSequentialTempChangeString() { return "maxSequentialTemperatureChange"; }
+
+    static constexpr char const * fluidNamesString() { return "fluid"; }
+    static constexpr char const * solidNamesString() { return "solid"; }
+    static constexpr char const * permeabilityNamesString() { return "permeability"; }
+    static constexpr char const * solidInternalEnergyNamesString() { return "solid internal energy"; }
+    static constexpr char const * thermalConductivityNamesString() { return "thermal conductivity"; }
```
CusiniM:
```suggestion
    static constexpr char const * solidInternalEnergyNamesString() { return "solidInternalEnergy"; }
    static constexpr char const * thermalConductivityNamesString() { return "thermalConductivity"; }
```

CusiniM (approved):

src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
paveltomin: no, didn't find any trace

src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
paveltomin: i am probably missing something - i didn't remove m_meshTargets itself but removed the `registerWrapper` call since it is not input neither it is written into restart

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.hpp
paveltomin: those can be printed out in case of input error, so i changed them into human language, is space a problem?

src/coreComponents/physicsSolvers/fluidFlow/FlowSolverBase.hpp
tjb-ltk: if a client of this string is the user, then spaces may be better, if the client lives in the developer world, then camelCase is fine.

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.hpp
tjb-ltk, line 1: What was the original intent of a structure of views...and this has been replaced with using getters for each prop ?

tjb-ltk (approved):

src/coreComponents/physicsSolvers/fluidFlow/SinglePhaseBase.hpp
paveltomin: there was a comment:
```
  /**
   * @brief Extract properties from a fluid.
   * @param fluid base reference to the fluid object
   * @return structure with property views
   *
   * This function enables derived solvers to substitute SingleFluidBase for a different,
   * unrelated fluid class, and customize property extraction. For example, it is used by
   * SinglePhaseProppantBase to allow using  constitutive::SlurryFluidBase, which does not inherit from
   * SingleFluidBase currently (but this design may need to be revised).
   */
```
but it is actually outdated since `SlurryFluidBase` does inherit from `SingleFluidBase` in the latest code, so those views are not needed anymore

PR #3683, opened 2025-05-29
refactor: include cleanup, trying to remove some dependencies

paveltomin:


src/coreComponents/fieldSpecification/SourceFluxBoundaryCondition.cpp
```diff
@@ -35,10 +34,9 @@ SourceFluxBoundaryCondition::SourceFluxBoundaryCondition( string const & name, G
   getWrapper< string >( FieldSpecificationBase::viewKeyStruct::functionNameString() ).
     setDescription( GEOS_FMT( "Name of a function that specifies the variation of the production rate variations of this {}."
                               "Multiplied by {}. If no function is provided, a constant value of 1 is used."
-                              "The producted fluid rate unit is in kg by default, or in mole if the flow solver has a {} of 0.",
+                              "The producted fluid rate unit is in kg by default, or in mole if the flow solver uses moles.",
```
castelletto1:
```suggestion
                              "The produced fluid rate unit is in kg by default, or in mole if the flow solver uses moles.",
```

src/coreComponents/physicsSolvers/multiphysics/HydrofractureSolver.cpp
```diff
@@ -25,7 +25,7 @@
 #include "physicsSolvers/multiphysics/HydrofractureSolverKernels.hpp"
 #include "physicsSolvers/solidMechanics/SolidMechanicsFields.hpp"
 #include "physicsSolvers/multiphysics/SinglePhasePoromechanics.hpp"
-#include "physicsSolvers/multiphysics/MultiphasePoromechanics.hpp"
+//#include "physicsSolvers/multiphysics/MultiphasePoromechanics.hpp"
```
castelletto1: Remove

castelletto1 (approved):

jhuang2601 (approved):

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
```diff
@@ -1498,149 +1497,100 @@ CompositionalMultiphaseWell::checkSystemSolution( DomainPartition & domain,
 
   string const wellDofKey = dofManager.getKey( wellElementDofName() );
   integer localCheck = 1;
-  if( 0 )
```
paveltomin, line 1508: @tjb-ltk please check

dkachuma (approved):

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
tjb-ltk: it should be removed...

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseUtilities.hpp
```diff
@@ -25,6 +25,26 @@
 namespace geos
 {
 
+/**
+ * @brief Options for density treatment in gravity
+ */
+enum class GravityDensityScheme : integer
+{
+  ArithmeticAverage, ///< average phase density is computed using simple arithmetic average:
+                     ///  rho_ave = 0.5 * (rho_i + rho_j)
```
rrsettgast, line 34: if either i or j doesn't have the phase, what happens?

rrsettgast (approved):

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseUtilities.hpp
paveltomin: i think densities are still computed even when phase is not present, so there will be some value for `rho` for all cells

we tried an experiment with skipping the density and viscosity compute https://github.com/GEOS-DEV/GEOS/pull/3576 but unfortunately it showed some performance degradation specifically for spe11, so we just parked it as an option, but didn't change the default behavior

PR #3708, opened 2025-06-23
chore: update LC builds

rrsettgast:


rrsettgast:
@alexbenedicto I think an update in geosPythonPackages may have broke our CI testing. Can you look into it?

PR #3718, opened 2025-07-01
feat:  Enable scalar value setting from Python via PyWrapper

bd713:
This PR introduces the `PyWrapper_setValue` method, enabling Python-to-C++ assignment of scalar values (`int, long, double, std::string`).

Previously, Python-to-C++ assignment was only possible for types like `NumPy arrays`, which support in-place modification. Scalar types lacked a setter, making it impossible to update them from Python.

In practice, this PR includes:
- A new method: `PyWrapper_setValue`
- A companion unit test: `testPyWrapper_setValue`, covering:
    - Valid types: `string, double, int, long`
    - Type mismatches (e.g., assigning a `string` to an `int` wrapper)
    - Unsupported types (e.g., `std::vector<int>`)


herve-gross:
@MelReyCG can you take a look at this one?

src/coreComponents/dataRepository/python/PyWrapper.hpp
```diff
@@ -24,6 +24,26 @@ namespace geos
...
+{
+  PyObject_HEAD
+
+  /// Documentation string for the PyWrapper Python type.
+  static constexpr char const * docString =
+    "A Python interface to geos::dataRepository::WrapperBase.";
+
+  /// Pointer to the underlying C++ WrapperBase instance.
+  dataRepository::WrapperBase * wrapper;
+};
```
corbett5: My understanding is that you moved this and the declaration for `PyWrapper_setValue` into the header so that they're available for the unit test correct?

I love that there is a unit test for the new functionality, but I do wish it didn't require modifying the source. I think ideally we'd have the unit tests themselves be in python as is done in LvArray

https://github.com/GEOS-DEV/LvArray/blob/develop/unitTests/python/testPyArray.cpp
https://github.com/GEOS-DEV/LvArray/blob/develop/unitTests/python/testPyArrayDriver.py

but I get that's pretty out of scope for this PR.

src/coreComponents/dataRepository/python/PyWrapper.cpp
```diff
@@ -100,10 +90,83 @@ static PyObject * PyWrapper_value( PyWrapper * const self, PyObject * const args
...
+  try
+  {
+    bool success = false;
+
+    if( PyUnicode_Check( pyValue ) )
+    {
+      const char * strVal = PyUnicode_AsUTF8( pyValue );
+      success = trySetValue( self->wrapper, std::string( strVal ) );
+      if( !success )
+        PyErr_SetString( PyExc_TypeError, "Wrapper is not a std::string." );
```
corbett5: Could be useful to say exactly what type `wrapper` holds.

src/coreComponents/dataRepository/python/PyWrapper.cpp
```diff
@@ -100,10 +90,83 @@ static PyObject * PyWrapper_value( PyWrapper * const self, PyObject * const args
...
+      return nullptr;
+    }
+
+    Py_RETURN_NONE;
+  }
+  catch( std::exception const & e )
+  {
+    PyErr_SetString( PyExc_RuntimeError, e.what() );
+    return nullptr;
+  }
```
corbett5: What sort of errors might we want to catch here? From what I see nothing should be throwing.

corbett5 (approved):

src/coreComponents/dataRepository/python/PyWrapper.hpp
bd713: Thanks for the reference!
Following the LvArray pattern, I reverted the header to its original state, and rewrote the test in python.

src/coreComponents/dataRepository/python/PyWrapper.cpp
bd713: Indeed... I changed it to a more informative message.

src/coreComponents/dataRepository/python/PyWrapper.cpp
bd713: Agreed, I've removed it

src/coreComponents/dataRepository/python/PyWrapper.hpp
corbett5: I didn't actually expect you to do this... so thank you! I think it's both cleaner and better testing. Hopefully it'll also make it easier for others to add similar tests in GEOS as well.

src/coreComponents/dataRepository/python/PyWrapper.cpp
corbett5: Great, thanks!

PR #3690, opened 2025-06-03
feat: Adding error YAML file in GEOS

amandinehry:
This PR aims to provide a file that stores warnings and errors regarding input data that occur during the current execution.

To do this, I created a class called `ErrorLogger` that retrieves, formats, and sends the desired information to a YAML file.
I also have (in the Logger.hpp file) two macros for each of the following cases:

- Warning
- Error
- Throw

For each of these cases, one of these macro allows to attach contextual information about the error/warning.
Each of the macros is based on the `ErrorLogger` class.

```
- type: Exception
    rank: 0
    message: >-
      compositionalMultiphaseWell (simpleCo2Inj_inc_erreurTable.xml, l.34): wrong surface pressure / temperature.
      Table input error for water phase (in table from "/path/to/randomExemple/pvtliquid_erreur.txt").
      /Functions(simpleCo2Inj_inc_erreurTable.xml,l.203)/fluid_phaseModel1_PhillipsBrineDensity_table:
      Requested pressure of 101325 [Pa] is out of the table bounds
      ( lower bound: pressure of 10000000 [Pa] -> upper bound: pressure of 75000000 [Pa] ).
    contexts:
      - dataPath: /Functions(simpleCo2Inj_inc_erreurTable.xml,l.203)/fluid_phaseModel1_PhillipsBrineDensity_table
        priority: 0
      - dataPath: /domain/MeshBodies/cartesianMesh/meshLevels/Level0/ElementRegions/elementRegionsGroup/
                  wellRegion(simpleCo2Inj_inc_erreurTable.xml,l.64)/elementSubRegions/wellRegionUniqueSubRegion/
                  ConstitutiveModels/fluid
        priority: 2
      - inputFile: /path/to/randomExemple/simpleCo2Inj_inc_erreurTable.xml
        inputLine: 34
        priority: 1
    sourceLocation:
      file: /path/to/GEOS/src/coreComponents/functions/TableFunction.cpp
      line: 191
    sourceCallStack:
      - frame0:  geos::constitutive::PVTProps::PhillipsBrineDensity::checkTablesParameters(double, double) const 
      - frame1:  geos::constitutive::CO2BrineFluid::checkTablesParameters(double, double) const 
      - frame2:  geos::CompositionalMultiphaseWell::validateWellControlsForFluid() const 
      - frame3:  geos::CompositionalMultiphaseWell::validateConstitutiveModels(geos::DomainPartition const&) const 
[...]

PR #3722, opened 2025-07-08
feat: Adding signal and external error detection & output

MelReyCG:
replaced by #3856


paveltomin:
@MelReyCG can you clean it up somehow?

dkachuma:
Is there an overlap with #3690? Would a merge of #3690 make this smaller?

MelReyCG:
I guess the only way to get this PR readable is to wait for the other one to be merged :(
image
Another way would be to open a PR in the fork just for reviewing.

PR #3725, opened 2025-07-13
feat: split FiniteElementBase into nonvirtual and virtual interfaces

rrsettgast:
This PR seeks to remove the issues caused by the virtual destructor in FIniteElementBase.

- [x] Remove operations from `FiniteElementBase`
- [x] Split derived types into virtual and non-virtual types.
- [x] use the non-virtual types in device kernels.


src/coreComponents/constitutive/fluid/twophaseimmisciblefluid/TwoPhaseImmiscibleFluid.hpp
```diff
@@ -253,7 +253,7 @@ void TwoPhaseImmiscibleFluid::KernelWrapper::
 {
   using Deriv = constitutive::multifluid::DerivativeOffset;
 
-  LvArray::forValuesInSlice( phaseDensity.derivs, []( real64 & val ) { val = 0.0; } );
+  // LvArray::forValuesInSlice( phaseDensity.derivs, []( real64 & val ) { val = 0.0; } );
```
wrtobin: these were giving the strange `basic_string` `__host__` only call from device context on lassen with clang-13, cuda-11.8

wrtobin:
@rrsettgast this *should* be fine to review then promoted to run the GPU CI and the test suite

wrtobin:
stack trace from the first failure I'm looking into:

```
***** ERROR
***** LOCATION: /tmp/geos/src/coreComponents/LvArray/src/ArraySlice.hpp:294
***** Controlling expression (should be false): index < 0 || index >= m_dims[ 0 ]
Array Bounds Check Failed: index=0 m_dims[0]=0

** StackTrace of 10 frames **
Frame 0: /tmp/geos-build/lib/libsolidMechanicsSolvers.so 
Frame 1: geos::SolidMechanicsLagrangianFEM::initializePostInitialConditionsPreSubGroups() 
Frame 2: geos::SolidMechanicsEmbeddedFractures::initializePostInitialConditionsPreSubGroups() 
Frame 3: geos::dataRepository::Group::initializePostInitialConditions() 
Frame 4: geos::dataRepository::Group::initializePostInitialConditions() 
Frame 5: geos::dataRepository::Group::initializePostInitialConditions() 
Frame 6: geos::GeosxState::applyInitialConditions() 
Frame 7: main 
Frame 8: /lib/x86_64-linux-gnu/libc.so.6 
Frame 9: __libc_start_main 
Frame 10: _start 
=====
```

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.cpp
```diff
@@ -376,11 +375,9 @@ void SolidMechanicsLagrangianFEM::initializePostInitialConditionsPreSubGroups()
...
           {
             typename FE_TYPE::StackVariables feStack;
-            finiteElement.template setup< FE_TYPE >( k, meshData, feStack );
-            localIndex const numSupportPoints =
-              finiteElement.template numSupportPoints< FE_TYPE >( feStack );
+            element.template setup< FE_TYPE >( k, meshData, feStack );
+            localIndex const numSupportPoints = element.getNumSupportPoints( feStack );
 
-//#if ! defined( CALC_FEM_SHAPE_IN_KERNEL ) // we don't calculate detJ in this case
             for( localIndex q=0; q<numQuadraturePointsPerElem; ++q )
```
wrtobin: It seems like this is the only reasonable source of the error, since its a bounds-check issue from this function, and the `initialize` call above is a no-op for basically all fetypes except the `ConformingVirtualElement`, which isn't in the `ALL_FE_TYPES` that I can tell.

src/coreComponents/constitutive/fluid/twophaseimmisciblefluid/TwoPhaseImmiscibleFluid.hpp
rrsettgast: I don't think we can just comment this out can we?

src/coreComponents/finiteElement/elementFormulations/H1_Pyramid_Lagrange1_Gauss5.hpp
```diff
@@ -640,6 +589,47 @@ H1_Pyramid_Lagrange1_Gauss5::
...
+  /**
+   * @brief Get the device-compatible implementation type.
+   * @return A pointer to the device-compatible implementation type.
+   */
+  H1_Pyramid_Lagrange1_Gauss5_impl const * getImpl() const
+  {
+    return static_cast< H1_Pyramid_Lagrange1_Gauss5_impl const * >(this);
+  }
+
+};
```
rrsettgast, line 684: @wrtobin what happened to the guards? If compiling on device, we never should need this?? I think you tried to explain but I didn't understand.

src/coreComponents/finiteElement/elementFormulations/H1_QuadrilateralFace_Lagrange1_GaussLegendre2.hpp
```diff
@@ -47,27 +47,50 @@ namespace finiteElement
...
+#endif
+  GEOS_HOST_DEVICE H1_QuadrilateralFace_Lagrange1_GaussLegendre2_impl() = default;
+  GEOS_HOST_DEVICE ~H1_QuadrilateralFace_Lagrange1_GaussLegendre2_impl() = default;
+  GEOS_HOST_DEVICE H1_QuadrilateralFace_Lagrange1_GaussLegendre2_impl( H1_QuadrilateralFace_Lagrange1_GaussLegendre2_impl const & ) = default;
+  GEOS_HOST_DEVICE H1_QuadrilateralFace_Lagrange1_GaussLegendre2_impl & operator=( H1_QuadrilateralFace_Lagrange1_GaussLegendre2_impl const & ) = default;
+  GEOS_HOST_DEVICE H1_QuadrilateralFace_Lagrange1_GaussLegendre2_impl( H1_QuadrilateralFace_Lagrange1_GaussLegendre2_impl && ) = default;
+  GEOS_HOST_DEVICE H1_QuadrilateralFace_Lagrange1_GaussLegendre2_impl & operator=( H1_QuadrilateralFace_Lagrange1_GaussLegendre2_impl && ) = default;
+#ifdef __CUDACC__
+  #pragma diag_pop
+#endif
```
rrsettgast: what is the pragma for?

src/coreComponents/physicsSolvers/solidMechanics/SolidMechanicsLagrangianFEM.cpp
rrsettgast: Not clear what line the bounds-check is referring to?

wrtobin: nope, just giving the lvarray line number and the function stack (not even memory offsets), the other comment is the stack

src/coreComponents/finiteElement/elementFormulations/H1_QuadrilateralFace_Lagrange1_GaussLegendre2.hpp
wrtobin: this is what cuda gives warnings for (decorating explicitly defaulted constructors/destructors), but was required for some older rocm compilers

per our discussion earlier this week we can probably just remove the decorators at this point, just wasn't top priority vs getting it working at the time, now that I have to debug more I'll go ahead and get this removed.

src/coreComponents/finiteElement/elementFormulations/H1_Pyramid_Lagrange1_Gauss5.hpp
wrtobin: I conflated the guards, we might be able to put them back, but symbol visibility shouldn't cause any of the corner-case size difference errors, only actually instantiating the undesired type on device. 

Basically using either `GEOS_DEVICE_COMPILE` or `__CUDACC__` / `__HIPCC__` caused symbol visibility issues in specific compilation contexts. I'll see if I can get a guard in place though.

src/coreComponents/mainInterface/ProblemManager.cpp
```diff
@@ -981,17 +980,14 @@ map< std::tuple< string, string, string, string >, localIndex > ProblemManager::
...
+                  FE_TYPE::template initialize< FE_TYPE, SUBREGION_TYPE >( nodeManager,
+                                                                           edgeManager,
+                                                                           faceManager,
+                                                                           subRegion,
+                                                                           meshData );
 
                   localIndex const numQuadraturePoints = FE_TYPE::numQuadraturePoints;
 
-//#if ! defined( CALC_FEM_SHAPE_IN_KERNEL )
-                  feDiscretization->calculateShapeFunctionGradients< SUBREGION_TYPE, FE_TYPE >( X, &subRegion, meshData, finiteElement );
```
wrtobin, line 998: @rrsettgast without this `detJ` isn't (1) resized or (2) initialized so when we go to access it in

`SolidMechanicsLagrangianFEM::initializePostInitialConditionsPreSubGroups` 

we get the oob error, presumably I should just add this back?

wrtobin: alternatively I can add a `calcDetJ` to the impl classes to calculate it on the fly if we want to go that way instead

rrsettgast: ah...we need to calculated it. We should be removing detJ completely from the data repo.

src/coreComponents/finiteElement/elementFormulations/H1_QuadrilateralFace_Lagrange1_GaussLegendre2.hpp
rrsettgast: ok. lets do that

src/coreComponents/finiteElement/elementFormulations/H1_Pyramid_Lagrange1_Gauss5.hpp
rrsettgast: The guard isn't required. I just thought it would be an indicator of when we were doing something unintended. Like having the virtual enabled class on device.

src/coreComponents/mainInterface/ProblemManager.cpp
wrtobin: okay so we wanna calculate on the fly for the mass initialization presumably, I'll take a look at getting that added and removing detJ from the CellElementSubregion.

src/coreComponents/finiteElement/elementFormulations/H1_Pyramid_Lagrange1_Gauss5.hpp
wrtobin: yeah I completely agree that a compile-time hard line to prevent instantiation of the type on device is preferable, I just couldn't get the compilation to succeed without visibility of the symbol on both passes (host and device), even if it isn't instantiated.

once the rest of the PR is working I'll come back to this and see if I can figure it out.

wrtobin:
@rrsettgast 
I removed the shape function gradient precalculation. The only place it was used was in the surface generation nodal force calculations. I rewrote those to calculate on-the-fly, but that required getting at the finite element discretization info (since prior the element type didn't matter so much, they just indexed into the appropriate arrays and assumed the array sizes defined the element support point count, quadrature point count, etc). 

That part seemed a bit messy, I feel like there might be an easier way to get the appropriate `FiniteElementBase` to downcast through the `dispatch3D` operator, but it didn't jump out at me while I was getting the calculations replaced. 

I was getting a diff on the mass calculation replacement from last week that was more significant than machine epsilon numerical rounding etc, so I'm going to try to go track down what is causing that.

wrtobin:
@rrsettgast Looks like we're good to go on this, all integrated tests passing.

paveltomin (approved):

castelletto1 (approved):

jhuang2601:
Are these failures expected or not?

```
=======================
Integrated test results
=======================
expected: 0
created: 0
batched: 0
filtered: 0
skipped: 0
running: 0
passed: 546
timedout: 0
halted: 0
lsferror: 0
failed: 2
=======================
Test failures
=======================
acous3D_pml_smoke_01
acous3D_pml_smoke_08
=======================
Overall status: FAILED
=======================
```

PR #3728, opened 2025-07-15
feat: `--validate-input` command line option

MelReyCG:
This PR aims to allow to launch GEOS to only run the loading phase, skipping any simulation and validating that GEOS can load a case without encountering any issue.

To do so, here is the proposed new command line option:
```
    -v, --validate-input, \t only do the loading phase, and not actual simulation. Useful to validate 'input'.
```
It also updates documentation accordingly.

----

This PR also fix the GEOS behavior when no command line argument are given (GEOS would throw errors and stacktrace, contrary to what is stated in the current documentation), and update the current "empty command line params" output in our docs.


dkachuma:
This is a very useful feature. Can we add it somewhere in [QuickStart.rst](https://github.com/GEOS-DEV/GEOS/blob/develop/src/docs/sphinx/QuickStart.rst?plain=1#L355-L373).

MelReyCG:
@dkachuma
Done, since I passed by I also updated the GEOS behaviour & documentation for when an user does not give any command line arguments.

paveltomin (approved):

castelletto1 (approved):

jhuang2601 (approved):
This feature is incredibly helpful.
Thanks for making it available.

herve-gross (approved):

dkachuma (approved):

rrsettgast (commented):
@MelReyCG I am missing where geos would terminate if onlyValidateInput==true. Can you point it out to me?

src/coreComponents/mainInterface/GeosxState.cpp
```diff
@@ -177,7 +177,12 @@ void GeosxState::run()
 
   GEOS_THROW_IF_NE( m_state, State::READY_TO_RUN, std::logic_error );
 
-  if( !getProblemManager().runSimulation() )
+  if( getCommandLineOptions().onlyValidateInput )
+  {
+    m_state = State::COMPLETED;
+    GEOS_LOG( "Input validation completed, terminating GEOS..." );
+  }
+  else if( !getProblemManager().runSimulation() )
```
MelReyCG, line 185: Sure, here it is ^
We set the GEOS state to completed and prevent it to run the simulation l.185. I can avoid using `else if`  here if that is not clear.

PR #3735, opened 2025-07-18
feat: Well Estimator + Constraint Selector

tjb-ltk:
1) Add well estimator solve assuming fixed reservoir conditions
    -  solve applied at beginning of Jacobian assembly of the coupled system
    -  frequency determined via input setting

2) Refactor constraint specification
    -  Introduce constraint definition in schema such as MinimumBHPConstraint or PhaseProductionConstraint to encapsulate quantities associated with each constraint type
    - Added kernels for constraint equation JacGen specialized for constraint type

3) Introduced concept of well separator internally to remove ambiguity of fluid model quantities encountered when using same fluid model for constraint flashes and well element calculations.   No input schema changes.  This should be under separate PR to include formal well separator schema

4) Introduced limiting constraint selector to work with well estimator.  At the beginning of the first n Newton Iterations , for each constraint, the well estimator applied and the most limiting constraint selected and set as the active constraint  for the remaining Newton iterations. 

5) PR currently has a fair bit of well debug code that will be removed


src/coreComponents/physicsSolvers/fluidFlow/wells/WellConstraintsBase.hpp
```diff
@@ -0,0 +1,373 @@
...
+template< typename T >
+void registerInjectionStream( array1d< real64 > & injectionStream,
+                              real64 & injectionTemperature,
+                              T & context )
+{
+  context.registerWrapper( constraintViewStruct::injectionStreamKey::injectionStreamString(), &injectionStream ).
+    setDefaultValue( -1 ).
+    setSizedFromParent( 0 ).
+    setInputFlag( dataRepository::InputFlags::OPTIONAL ).
+    setDescription( "Global component densities of the injection stream [moles/m^3 or kg/m^3]" );
```
dkachuma, line 112: I always assumed this to be fractions (mass or mole).

src/coreComponents/physicsSolvers/fluidFlow/wells/WellControls.cpp
```diff
@@ -176,6 +184,62 @@ WellControls::WellControls( string const & name, Group * const parent )
...
+Group * WellControls::createChild( string const & childKey, string const & childName )
+{
+  GEOS_LOG_RANK_0( GEOS_FMT( "{}: adding {} {}", getName(), childKey, childName ) );
+  ////const auto childTypes = { viewKeyStruct::perforationString() };
+  //GEOS_ERROR_IF( childKey != viewKeyStruct::perforationString(),
+  //               CatalogInterface::unknownTypeError( childKey, getDataContext(), childTypes ) );
+
+  Group * constraint = nullptr;
+  if( childKey == viewKeyStruct::minimumBHPConstraintString() )
+  {
```
dkachuma, line 216: I'm guessing we are expecting only one constraint of each type. Is there a way we can enforce this? Something similar to the NonLinearParameters on the solvers.

src/coreComponents/physicsSolvers/fluidFlow/wells/WellControls.hpp
```diff
@@ -340,6 +394,22 @@ class WellControls : public dataRepository::Group
...
+    /// string key for the maximum phase rate for a producer
+    static constexpr char const * phaseProductionConstraintString() { return "PhaseProductionConstraint"; }
+    /// string key for the maximum phase rate for a injection
+    static constexpr char const * phaseInjectionConstraintString() { return "PhaseInjectionConstraint"; }
+    /// string key for the maximum volume rate for a producer
+    static constexpr char const * volumeProductionConstraintString() { return "VolumeProductionConstraint"; }
+    /// string key for the maximum volume rate for a injector
+    static constexpr char const * volumeInjectionConstraintString() { return "VolumeInjectionConstraint"; }
+    /// string key for the maximum mass rate for a producer
+    static constexpr char const * massProductionConstraintString() { return "massProductionConstraint"; }
```
dkachuma, line 492:
```suggestion
    static constexpr char const * massProductionConstraintString() { return "MassProductionConstraint"; }
```

src/coreComponents/physicsSolvers/fluidFlow/wells/WellControls.hpp
```diff
@@ -340,6 +394,22 @@ class WellControls : public dataRepository::Group
...
+    /// string key for the maximum mass rate for a injector
+    static constexpr char const * massInjectionConstraintString() { return "massInjectionConstraint"; }
```
dkachuma, line 494:
```suggestion
    static constexpr char const * massInjectionConstraintString() { return "MassInjectionConstraint"; }
```

src/coreComponents/physicsSolvers/fluidFlow/wells/WellPhaseRateConstraints.hpp
```diff
@@ -0,0 +1,354 @@
...
+                 ": Invalid phase type for simulation fluid model",
+                 InputError );
+}
+
+/**
+ * @class PhaseInjectionConstraint
+ * @brief This class describes a phase rate constraint used to control a injection well.
+ */
+
+class PhaseInjectionConstraint : public PhaseConstraint
```
dkachuma, line 242: How is `PhaseInjectionConstraint` different from `VolumeInjectionConstraint`?

tjb-ltk: Probably should be called TotalVolume...

tjb-ltk: and Phase could be called PhaseVolume

src/coreComponents/physicsSolvers/fluidFlow/wells/WellControls.cpp
tjb-ltk: From the user, yes only one constraint of each type, but the code would still select the most limiting constraint. I will put code in to enforce this, and also valid combinations of constraints.  
1) There maybe more than one constraint of each type for operations like WHP or scaling, where the simulator imposes a constraint based on some allocation scheme, or for WHP, it would implement this as a BHP constraint, or other.  Constraints are usually tagged with  USER or SIM, where SIM being an internally generated constraint.  How/when these SIM constraints get registered still needs to be addressed. 
2) you can assign inj and prod constraints to a well. This is more of an oil field usecase, cycling, but for geothermal this might be of use for closed loop well, which could have 2 boundary conditions.

PR #3738, opened 2025-07-23
feat: Chevron's .cmake at host-configs/CVX

drmichaeltcvx:
These are the .cmake host configuration files for Chevron. These are all new files not affecting existing "develop" branch


drmichaeltcvx:
I am cleaning up and reducing the number of the various `host-config/CVX/*.cmake` files. As a matter of fact most are symlinks and their presence their is as a convenience to declare what options were used for building TPL/GEOS.

drmichaeltcvx:
This PR is not introducing any code changes. It simply brings in elements that are applicable to building GEOS/TPL in Chevron, namely, `host-configs/CVX` and matching scripts that preprocess and adapt the `host-configs/CVX/*.cmake` to HPC environment's peculiarities at Chevron. Please suggest proper labels, as some CI jobs fail due to that.

PR #3740, opened 2025-07-24
fix: add missing updateHydraulicApertureAndFracturePermeability for sequential

paveltomin:


Guotong-Ren:
what type of problem is solved by sequential coupling here?

paveltomin:
> what type of problem is solved by sequential coupling here?

poromechanics with conforming fractures

Guotong-Ren (approved):

jhuang2601 (approved):
Thanks for fixing this reported issue

PR #3743, opened 2025-07-28
feat: Enable mass variables in compositional models

dkachuma:
This PR enables the mass formulation for compositional models. 

All calculation in the compositional fluid model is performed with mole fractions. So when using the mass formulation the mass fractions are converted into mole fractions and then the calculation performed. This change then converts the derivatives using a chain rule from mole fraction derivatives to mass fraction derivatives.

The `PVTDriver` uses mass variables. So now with this activated, the results are based on the mass density and not molar density.


paveltomin (approved):

herve-gross:
@dkachuma @paveltomin no file changes left - should this PR be closed?

paveltomin:
all changes are already in develop

PR #3745, opened 2025-07-29
feat: oscillation scaling

paveltomin:
an option to scale the solution changes when oscillations (on Newton iteration level) are detected:
`oscillationScaling` - 0 or 1 - enables the feature
`oscillationScalingFactor` - between 0 and 1 - scaling factor to be applied when oscillations are detected
`oscillationCheckDepth` - 2 or larger - how many iterations down the solution change history will be checked
`oscillationTolerance` - check tolerance, used both to trim-out small solution changes and to check that oscillatory changes are close to each other in magnitude
`oscillationFraction` - between 0 and 1 - fraction of total cells that have oscillations


src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
```diff
@@ -1487,6 +1512,62 @@ void PhysicsSolverBase::saveSequentialIterationState( DomainPartition & GEOS_UNU
...
+  if( m_solutionHistory.size() < oscillationCheckDepth )
+    return false; // not enough history to check oscillations
+
+  RAJA::ReduceSum< parallelDeviceReduce, localIndex > oscillationCount = 0;
+
+  localIndex const numDofs = m_solutionHistory[0].size();
+  localIndex const historySize = m_solutionHistory.size();
+
+  RAJA::forall< parallelDevicePolicy<> >( RAJA::TypedRangeSegment< localIndex >( 0, numDofs ),
+                                          [&] GEOS_HOST ( localIndex const dof )
```
dkachuma:
```suggestion
                                          [=] GEOS_HOST_DEVICE ( localIndex const dof )
```

src/coreComponents/physicsSolvers/NonlinearSolverParameters.cpp
```diff
@@ -195,6 +221,20 @@ void NonlinearSolverParameters::postInputInitialization()
...
 
+  // check oscillation parameters
+  GEOS_ERROR_IF_LE_MSG( m_oscillationScalingFactor, 0.0,
+                        getWrapperDataContext( viewKeysStruct::oscillationScalingFactorString() ) << ": should be positive" );
+  GEOS_ERROR_IF_GT_MSG( m_oscillationScalingFactor, 1.0,
+                        getWrapperDataContext( viewKeysStruct::oscillationScalingFactorString() ) << ": can not be more than 1.0" );
+  GEOS_ERROR_IF_LT_MSG( m_oscillationCheckDepth, 2,
+                        getWrapperDataContext( viewKeysStruct::oscillationCheckDepthString() ) << ": can not be less than 2" );
+  GEOS_ERROR_IF_LE_MSG( m_oscillationTolerance, 0.0,
+                        getWrapperDataContext( viewKeysStruct::oscillationToleranceString() ) << ": should be positive" );
```
dkachuma: Should also be less than 1?

src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
```diff
@@ -1487,6 +1512,62 @@ void PhysicsSolverBase::saveSequentialIterationState( DomainPartition & GEOS_UNU
...
+        }
+      }
+
+      if( oscillationDetected )
+      {
+        oscillationCount += 1;
+      }
+    } );
+
+  real64 const f = static_cast< real64 >( MpiWrapper::sum( oscillationCount.get() ) ) / numDofs;
```
dkachuma: I think we should also mpi sum `numDofs` here.

src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
```diff
@@ -1371,9 +1375,26 @@ bool PhysicsSolverBase::checkSystemSolution( DomainPartition & GEOS_UNUSED_PARAM
...
+
+  // Check for oscillations
+  if( m_nonlinearSolverParameters.m_oscillationScaling && detectOscillations() )
+  {
+    scalingFactor *= m_nonlinearSolverParameters.m_oscillationScalingFactor;
+    GEOS_LOG_LEVEL_RANK_0( logInfo::NonlinearSolver,
+                           GEOS_FMT( "        {}: oscillation detected, scaling factor set to {}", getName(), scalingFactor ) );
+  }
+
+  m_solutionHistory.appendArray( localSolution.begin(), localSolution.end());
```
dkachuma: No need to populate of `m_nonlinearSolverParameters.m_oscillationScaling` is false.

dkachuma (approved):

paveltomin:
@dkachuma all your comments should be fixed now, thanks

paveltomin:
@dkachuma example how it helps a bit for _dome_ case:
```
                oscillationScalingFactor="0.66"
                oscillationCheckDepth="2"

```
image

src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
tjb-ltk, line 1: A unit test or  integrated test to exercise the detection logic would be good.  Ideally this could be exercised in isolation from a full simulation run, but this isn't always possible..

paveltomin: i enabled it for one example from integrated tests

tjb-ltk (approved):

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.cpp
```diff
@@ -615,7 +615,7 @@ real64 CompositionalMultiphaseFVM::scalingForSystemSolution( DomainPartition & d
   else
   {
     string const dofKey = dofManager.getKey( viewKeyStruct::elemDofFieldString() );
-    real64 scalingFactor = 1.0;
+    real64 scalingFactor = CompositionalMultiphaseBase::scalingForSystemSolution( domain, dofManager, localSolution );
```
jhuang2601, line 619: Can this feature be extended for singlePhase flow?

jhuang2601 (approved):

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.cpp
paveltomin: i think it should work for single phase as is now, although i didn't have a change to test

PR #3748, opened 2025-07-31
refactor: use old pore volume in volume balance equation

paveltomin:
for flow assembly kernel, for poromechanics it is already doing old pore volume

bonus: re-use `accumulationAssemblyLaunch` to avoid code duplication


castelletto1 (approved):
Looks good to me, thanks

dkachuma (approved):

PR #3753, opened 2025-08-01
feat: Add linear solve functions

dkachuma:
Extends the dense linear solver interface to include linear solves where the solution and rhs are matrices. The addition is a function to solve
$$AX=B$$
where $A$ is an $n{\times}n$ matrix and $X$ and $B$ are $n{\times}m$ matrices.

This also adds an "in-place" version of `solve`. This takes the matrix $A$ and the rhs $B$ only. In using this, the caller is allowing both the $A$ and $B$ to be destroyed with $B$ being replaced by the solution $X$. This has the advantage of not requiring extra allocations of memory within the call.


src/coreComponents/denseLinearAlgebra/denseLASolvers.hpp
```diff
@@ -132,6 +180,69 @@ bool solveThreeByThreeSystem( MATRIX_TYPE const & A, RHS_TYPE const & b, SOL_TYP
...
+ * it is overwritten with the solution vectors (i.e., `X`). The function checks for
+ * singularity using a tolerance and returns `false` if the system cannot be solved.
+ *
+ * @tparam M The number of columns in the rhs matrix.
+ * @tparam MATRIX_TYPE The type of the matrix A. Must support indexing with `A[i][j]`.
+ * @tparam SOL_TYPE The type of the rhs/solution matrix x. Must support indexing with `X[i][j]`.
+ *
+ * @param[in] A The 3x3 matrix representing the system of equations. Must have size 3x3.
+ * @param[in,out] X The 3xM matrix that will store the right-had-side and solution to the system.
+ * @return bool that sepcifies whether the solve succeeded (1) or not (0).
```
castelletto1:
```suggestion
 * @return bool that specifies whether the solve succeeded (1) or not (0).
```

src/coreComponents/denseLinearAlgebra/denseLASolvers.hpp
```diff
@@ -78,6 +78,54 @@ bool solveTwoByTwoSystem( MATRIX_TYPE const & A, RHS_TYPE const & b, SOL_TYPE &&
...
+ * it is overwritten with the solution vectors (i.e., `X`). The function checks for
+ * singularity using a tolerance and returns `false` if the system cannot be solved.
+ *
+ * @tparam M The number of columns in the rhs matrix.
+ * @tparam MATRIX_TYPE The type of the matrix A. Must support indexing with `A[i][j]`.
+ * @tparam SOL_TYPE The type of the rhs/solution matrix x. Must support indexing with `X[i][j]`.
+ *
+ * @param[in] A The 2x2 matrix representing the system of equations. Must have size 2x2.
+ * @param[in,out] X The 2xM matrix that will store the right-had-side and solution to the system.
+ * @return bool that sepcifies whether the solve succeeded (1) or not (0).
```
castelletto1:
```suggestion
 * @return bool that specifies whether the solve succeeded (1) or not (0).
```

src/coreComponents/denseLinearAlgebra/denseLASolvers.hpp
```diff
@@ -78,6 +78,54 @@ bool solveTwoByTwoSystem( MATRIX_TYPE const & A, RHS_TYPE const & b, SOL_TYPE &&
...
+ * On input, the matrix `X` contains the right-hand side values (i.e., `B`). On output,
+ * it is overwritten with the solution vectors (i.e., `X`). The function checks for
+ * singularity using a tolerance and returns `false` if the system cannot be solved.
+ *
+ * @tparam M The number of columns in the rhs matrix.
+ * @tparam MATRIX_TYPE The type of the matrix A. Must support indexing with `A[i][j]`.
+ * @tparam SOL_TYPE The type of the rhs/solution matrix x. Must support indexing with `X[i][j]`.
+ *
+ * @param[in] A The 2x2 matrix representing the system of equations. Must have size 2x2.
+ * @param[in,out] X The 2xM matrix that will store the right-had-side and solution to the system.
```
castelletto1:
```suggestion
 * @param[in,out] X The 2xM matrix that will store the right-hand-side and solution to the system.
```

castelletto1 (approved):
This looks good to me, @dkachuma -- thank you!

Side question: What was the original argument against only offering the option where the matrix and RHS are overwritten by the factorization and solution vector? Personally, I’d be in favor -- though not in this PR -- of supporting only that behavior, with the expectation that the client code is responsible for copying the matrix and RHS beforehand if they’re needed after solving the linear system.

dkachuma:
> Side question: What was the original argument against only offering the option where the matrix and RHS are overwritten by the factorization and solution vector? 

I think there just wasn't a need for that because the current application was only for 2x2 matrices where allocations are cheap and can be done on the stack.

PR #3757, opened 2025-08-05
refactor: add NLR for residual outputs

paveltomin:
most places already have it, changing the rest for consistency


dkachuma (approved):
I'm surprised that we need this. I would have expected anyone wanting a carriage return to have to explicitly put it in their message.

PR #3766, opened 2025-08-12
feat: Add InvariantImmiscibleFluid for compositional mixtures

OmarDuran:
This PR adds the InvariantImmiscibleFluid constitutive fluid model for a minimal compositional setup. The fluid is invariant, in the sense of its properties are constant and do not depend on thermodynamic states. While the existing multiphase model can handle similar two-phase immiscible behavior, this model provides a simple setup for the compositional solvers `CompositionalMultiphaseFVM` and `CompositionalMultiphaseHybridFVM`. It is useful for the following run-time checks:

- Minimal configuration for integration tests - no complex thermodynamic parameters
- Allows for fast dynamic tests in large-scale meshes using linear tracer simulations
- Allows for testing and consistency check of FV and MFD discretization on any mesh
- Allows for testing upwinding discretizations under simple gravitational segregation

Minimal setup:
```
<InvariantImmiscibleFluid name="fluid"
                          componentNames="{ H2O, C5H12, CH4 }"
                          phaseNames="{ water, oil, gas }"
                          densities="{ 1000.0, 800.0, 100.0 }"
                          viscosities="{ 0.001, 0.005, 0.00002 }"/>
```

Testing: Phase property validation for both mass/molar basis calculations


src/coreComponents/constitutive/fluid/multifluid/constant/InvariantImmiscibleFluid.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+GEOS_HOST_DEVICE
+void InvariantImmiscibleFluid::update( localIndex const k,
+                                       localIndex const q,
+                                       real64 const pressure,
+                                       real64 const temperature,
+                                       arraySlice1d< real64 const, compflow::USD_COMP - 1 > const & composition ) const
+{
+#ifdef GEOS_USE_DEVICE_CODE
+  // For device code, create slices directly and call compute
```
paveltomin: this does not look standard for geos code and should be probably avoided

src/coreComponents/constitutive/fluid/multifluid/constant/InvariantImmiscibleFluid.hpp
```diff
@@ -0,0 +1,165 @@
...
+  void checkTablesParameters( real64 pressure, real64 temperature ) const override;
+
+  /**
+   * @brief Kernel wrapper class for InvariantImmiscibleFluid
+   *        This kernel can be called on the GPU
+   */
+  class KernelWrapper final : public MultiFluidBase::KernelWrapper
+  {
+public:
+    GEOS_HOST_DEVICE
```
paveltomin: better place after the comments

paveltomin (approved):

src/coreComponents/constitutive/fluid/multifluid/constant/InvariantImmiscibleFluid.hpp
```diff
@@ -0,0 +1,165 @@
...
+  };
+
+  /**
+   * @brief Create a KernelWrapper for this fluid model
+   * @return A KernelWrapper instance for this fluid model
+   */
+  KernelWrapper createKernelWrapper() const;
+
+  GEOS_HOST_DEVICE
+  virtual void update( localIndex const k,
```
dkachuma: The current pattern is to have this method only on the `KernelWrapper`. Then we can call `InvariantImmiscibleFluid::createKernelWrapper::update()`.

src/coreComponents/constitutive/fluid/multifluid/constant/InvariantImmiscibleFluid.hpp
```diff
@@ -0,0 +1,259 @@
...
+                                                       PhaseProp::SliceType const phaseInternalEnergy,
+                                                       PhaseComp::SliceType const phaseCompFraction,
+                                                       FluidProp::SliceType const totalDensity ) const
+{
+  // Note: pressure and temperature are marked as unused as this model uses constant properties
+  GEOS_UNUSED_VAR( pressure, temperature );
+
+  using Deriv = constitutive::multifluid::DerivativeOffset;
+
+  integer nPhase = phaseDensity.value.size();
```
dkachuma:
```suggestion
  integer const nPhase = numPhases();
```
Better to rely on this object's properties than caller arguments.

src/coreComponents/constitutive/fluid/multifluid/constant/InvariantImmiscibleFluid.hpp
```diff
@@ -0,0 +1,259 @@
...
+  integer nComp = phaseCompFraction.value.size( 1 );
```
dkachuma:
```suggestion
  integer const nComp = numComponents();
```

src/coreComponents/constitutive/fluid/multifluid/constant/InvariantImmiscibleFluid.cpp
```diff
@@ -0,0 +1,91 @@
...
+  }
+  return -1;
+}
+
+void InvariantImmiscibleFluid::postInputInitialization()
+{
+  // check base inputs
+  MultiFluidBase::postInputInitialization();
+
+  integer numPhase = numFluidPhases();
```
dkachuma, line 54: I think we have an assumption that number of components equals number of phases. So we need to check `numFluidPhases()` against `numFluidComponents()`.

src/coreComponents/constitutive/fluid/multifluid/constant/InvariantImmiscibleFluid.hpp
```diff
@@ -0,0 +1,259 @@
...
+GEOS_HOST_DEVICE
+GEOS_FORCE_INLINE
+void InvariantImmiscibleFluid::KernelWrapper::update( localIndex const k,
+                                                      localIndex const q,
+                                                      real64 const pressure,
+                                                      real64 const temperature,
+                                                      arraySlice1d< real64 const, compflow::USD_COMP - 1 > const & composition ) const
+{
+  // Create slice views of the member variables at cell k, quadrature point q
+  PhaseProp::SliceType phaseFractionSlice{m_phaseFraction.value[k][q], m_phaseFraction.derivs[k][q]};
```
dkachuma: ` m_phaseFraction( k, q )` should give the slice directly. No need to index the values and derivatives separately.

dkachuma (approved):
Looks good to me.

I however have a feeling that such an incompressible fluid would be quite difficult for the fluid solvers to deal with.

src/coreComponents/constitutive/fluid/multifluid/constant/InvariantImmiscibleFluid.hpp
OmarDuran: The decorator was replaced as suggested.

src/coreComponents/constitutive/fluid/multifluid/constant/InvariantImmiscibleFluid.hpp
OmarDuran: Thank for this clear directive. I made the code compliant with your suggestion.

src/coreComponents/constitutive/fluid/multifluid/constant/InvariantImmiscibleFluid.hpp
OmarDuran: The slicing is apply directly as you have suggested.

paveltomin:
> I however have a feeling that such an incompressible fluid would be quite difficult for the fluid solvers to deal with.

maybe adding an option to provide phase compressibilities would be nice?

src/coreComponents/constitutive/fluid/multifluid/constant/InvariantImmiscibleFluid.cpp
OmarDuran: The assumption is now checked at the specified location, and the test has been updated accordingly.

OmarDuran:
> Looks good to me.
> 
> I however have a feeling that such an incompressible fluid would be quite difficult for the fluid solvers to deal with.

I can see where incompressible fluids may pose challenges in certain solver setups, especially with mechanics on top. However keep in mind that this zero-order model is intentionally designed to simplify testing and validation within compositional solvers with a minimal setup.

OmarDuran:
> > I however have a feeling that such an incompressible fluid would be quite difficult for the fluid solvers to deal with.
> 
> maybe adding an option to provide phase compressibilities would be nice?

Adding optional phase compressibilities is a valid option. However, if needed users can always rely on DeadOilFluid which already include compressibilties via tables.

src/coreComponents/constitutive/unitTests/testInvariantImmiscibleFluid.cpp
```diff
@@ -0,0 +1,475 @@
...
+
+      // Initialize the fluid by computing properties with a test composition
+      stackArray1d< real64, 3 > composition( 3 );
+      composition[0] = 0.4;
+      composition[1] = 0.3;
+      composition[2] = 0.3;
+
+      arraySlice1d< real64 const, compflow::USD_COMP - 1 > compositionS( composition );
+      StackArray< real64, 3, 3, multifluid::LAYOUT_PHASE > compositionData( 1, 1, 3 );
+      auto composition_c = compositionData[0][0];
```
dkachuma:
```suggestion
      StackArray< real64, 2, 3, compflow::LAYOUT_COMP > compositionData( 1, 3 );
      auto composition_c = compositionData[0];
```
`#include "common/DataLayouts.hpp"`

OmarDuran: Thank you for the suggestion, @dkachuma. Using slices as types in the fluid signatures is indeed non-intuitive. It may be worth a refactor in the near future to simplify the signatures, perhaps using views only.

dkachuma: Agree. The proliferation of these layouts and USDs is not great. We need to simplify/unify this somewhat.

tjb-ltk (commented):
An integrated test should be add to validate other components ability to run with this fluid model

OmarDuran:
> An integrated test should be add to validate other components ability to run with this fluid model

This PR is oriented on introducing the invariant fluid model. A set of integration tests is planned and will be added once this  PR is merged.

paveltomin:
> > An integrated test should be add to validate other components ability to run with this fluid model
> 
> This PR is oriented on introducing the invariant fluid model. A set of integration tests is planned and will be added once this PR is merged.

don't forget to move it to the merge queue if it ready

tjb-ltk:
> > An integrated test should be add to validate other components ability to run with this fluid model
> 
> This PR is oriented on introducing the invariant fluid model. A set of integration tests is planned and will be added once this PR is merged.

Fair enough , didn't realize development branch can be used this way.   What was the rationale for introducing this fluid model instead of modifying the fluid property input of other models to effectively make the prop/ders constant ?

OmarDuran:
> > > An integrated test should be add to validate other components ability to run with this fluid model
> > 
> > 
> > This PR is oriented on introducing the invariant fluid model. A set of integration tests is planned and will be added once this PR is merged.
> 
> Fair enough , didn't realize development branch can be used this way. What was the rationale for introducing this fluid model instead of modifying the fluid property input of other models to effectively make the prop/ders constant ?

The rationale is to provide a fluid model that captures zero-order compositional behavior in a minimal setup. This is more robust and transparent than adapting existing models via input hacks, and better suited for testing and validating the compositional flow functionality without having the thermodynamic complexity of the existing models.

PR #3770, opened 2025-08-12
feat: improve mimetic inner products unit testing

chauj96:
## Overview
This PR adds three sets of unit tests: **1) consistency test**, **2) linear pressure recovery tests** and **3) hydrostatic equilibrium consistency test** for multiple inner product formulations in the mimetic discretization framework. 

Work in collaboration: @OmarDuran, @castelletto1, @herve-gross, @bd713 
 
---
## 1. Consistency Test

**Functions**
`runConsistencyTest(...)`

**Purpose**
Verify that each IP (inner product) satisfies the relation:
<p align="center">
$$
NK = TC
$$
</p>
where

 - $N$ = face normal matrix

- $K$ = full permeability tensor

- $C = NK$

- $T$ = transmissibility matrix from IP formulation

We check that
<p align="center">
$$
\Vert NK - TC \Vert < \text{tol} \quad (\text{tol} = 10^{-10})
$$
</p>

**Coverage**
- 4 IP types: TPFA, QuasiTPFA, Simple, BdVLM
- 2 element types: hexahedron, tetrahedron
- 8 cases total

**Expected results**
All 8 cases pass (norm < tol)

---

## 2. Linear Pressure Recovery Test

**Reference pressure field**
<p align="center">
$$
p(x,y,z) = x + \alpha y, \quad \text{we set } \alpha = 0.3
$$
</p>

**Geometry cases**

- Two unit cubes with no distortion
- Planar distortion: two cells, shared face vertices shifted in x-direction
- Nonplanar distortion: two opposite vertices shifted in contrary directions

We can modify the level of distortion $\epsilon$, and for reference case we set $\epsilon = 0.2$.

**IP types tested**
- TPFA
- QuasiTPFA
- Simple
- BdVLM

**Expected results**
- Case 1 (unit cube): All IP types recover the linear field exactly (error < 1e-15)
- Case 2 (planar distortion): QuasiTPFA, Simple, BdVLM recover the field, while TPFA fails the recovery
- Case 3 (non-planar distortion): All IP types fail to recover linear pressure (expected due to nonplanar facets)
---

## Reference figures

planar case

nonplanar case

---
## 3. Hydrostatic Equilibrium Consistency Test

**Purpose**
Verify the four inner product schemes remain consistent when a gravity term is included. In other words, the discretization should exactly reproduce hydrostatic equilibrium and yield zero net flux.

**Test statement**
The test checks that
<p align="center">
$$
\Vert T((\pi - \mathbf{e}p_E) + C(\rho \mathbf{g})) \Vert < \text{tol} \quad (\text{tol} = 10^{-15})
$$
</p>
where

- $T$ = transmissibility matrix
- $\pi$ = vector of face pressures
- $p_E$ = cell center pressure
- $\mathbf{e}$ = $(1,\dots,1) \in \mathbb{R}^{n_f}$ ($n_f$: number of faces)
- $C$ = cell-to-face centroid displacement matrix
- $\rho$ = density
- $\mathbf{g}$ = gravity vector.

It measures the discrete fluxes induced by pressure differences plus gravity. In true hydrostatic equilibrium, pressure gradients exactly balance the weight of the fluid. If the norm above is below tolerance, it confirms that the scheme reproduces this balance without introducing spurious flow.

**Physical setting**
The domain consists of two unit cubes stacked vertically. A hydrostatic pressure field

<p align="center">
$$
p(x,y,z) = p_D - \rho g (z - z_0) \quad (p_D: \text{reference pressure at height } z = z_0,\; z_0: \text{reference elevation})
$$
</p>

is imposed with gravity acting in the negative $z$ -direction. For the test, we set $\rho = g = p_D = 1.0$. The three mesh configurations are used as well: (i) undistorted cubes, (ii) planar distortion of the shared interface and (iii) non-planar distortion of the shared interface. 

**Expected results**
All four inner product formulations (TPFA, QuasiTPFA, Simple, BdVLM) pass the test, both with and without mesh distortion.

---

## Reference figures

planar case

nonplanar case


src/coreComponents/unitTests/finiteVolumeTests/testMimeticInnerProducts.cpp
```diff
@@ -905,9 +1010,474 @@ TEST( testMimeticInnerProducts, BdVLMtetra )
...
+  real64 sum = 0;
+  // iterate over columns
+  for( localIndex j = 0; j < T.size( 1 ); ++j )
+  {
+    sum += T( row, j );
+  }
+  return sum;
+}
+
+// recompute volumn and center of distorted cell
```
bd713: typo:  `volumn` -> `volume`

src/coreComponents/unitTests/finiteVolumeTests/testMimeticInnerProducts.cpp
```diff
@@ -905,9 +1010,474 @@ TEST( testMimeticInnerProducts, BdVLMtetra )
...
+    else
+    {
+      EXPECT_GT( errTPFA, 1e-15 );       // test TPFA is inconsistent with non K-orthogonal grids
+    }
+
+  }
+}
+
+// =================== case 2: with distortion (nonplanar) ===========================
+TEST( MimeticIP_Linear, Distortion_NonPlanar_LinearPresssure )
```
bd713: typo: `Pressure` with 2 `s` and not 3

src/coreComponents/unitTests/finiteVolumeTests/testMimeticInnerProducts.cpp
```diff
@@ -905,9 +1010,474 @@ TEST( testMimeticInnerProducts, BdVLMtetra )
...
+
+  real64 A11 = aL, A22 = aR, A13 = -alphaL, A23 = -alphaR, A31 = betaL, A32 = betaR, A33 = -gamma;
+  real64 b1 = rL, b2 = rR, b3 = boundary_sum;
+
+  real64 denom = A33 - A31 * A13 / A11 - A32 * A23 / A22;
+  real64 lambdaI = ( b3 - A31 * (b1 / A11) - A32 * (b2 / A22) ) / denom;
+  real64 pL = ( b1 - A13 * lambdaI ) / A11;
+  real64 pR = ( b2 - A23 * lambdaI ) / A22;
+
+  // compute analytcal pressure values
```
bd713: analyt`i`cal

bd713 (approved):
In general, unless we're comparing against very small reference values, I prefer using relative error over absolute error. This helps ensure scale independence and reduces the need to fine-tune threshold bounds across the different tests.

src/coreComponents/unitTests/finiteVolumeTests/testMimeticInnerProducts.cpp
```diff
@@ -525,6 +525,68 @@ void makeTetra( array2d< real64, nodes::REFERENCE_POSITION_PERM > & nodePosition
...
+  EXPECT_LT( diffNorm, 1e-10 ) << testName << ": norm(NK - TC) = " << diffNorm;
+
+  if( diffNorm < 1e-10 )
+  {
+    std::cout << "[CONSISTENCY TEST PASSED] " << testName << " consistency test passed" <<  std::endl;
+  }
+  else
+  {
+    std::cout << "[CONSISTENCY TEST FAILED] " << testName << " consistency test failed: norm(NK - TC) = " << diffNorm << std::endl;
+  }
```
paveltomin: should it return anything? like status?

herve-gross (approved):

src/coreComponents/unitTests/finiteVolumeTests/testMimeticInnerProducts.cpp
chauj96: I originally added "runConsistencyTest" inside each inner product test block because I wanted to check both the transmissibility matrix comparison against the reference and the operator consistency (norm(TC - NK) < tol) at the same time. When we run the unit tests, all the inner products print out that they are consistent. 

Do you think it would be better to separate these and make the test structure clearer?

dkachuma: I'd say line 579 `EXPECT_LT( diffNorm, 1e-10 )` should be enough and it will indicate that the test has passed or failed. I don't see a need to print these extra messages.

OmarDuran (approved):

src/coreComponents/unitTests/finiteVolumeTests/testMimeticInnerProducts.cpp
chauj96: I removed the extra messages and fixed the tolerance to 1e-11.

paveltomin (approved):

PR #3773, opened 2025-08-13
feat: print out solver statistics after every cycle when log level >=2

paveltomin:


MelReyCG:
I would suggest to add a bit of documentation in `src\coreComponents\physicsSolvers\LogLevelsInfo.hpp`:
```
struct SolverExecutionDetails
{
  static constexpr int getMinLogLevel() { return 2; }
  static constexpr std::string_view getDescription() { return "More precise information on solver execution + does output after every cycle."; }
};
```

MelReyCG (approved):

paveltomin:
> I would suggest to add a bit of documentation in `src\coreComponents\physicsSolvers\LogLevelsInfo.hpp`:
> 
> ```
> struct SolverExecutionDetails
> {
>   static constexpr int getMinLogLevel() { return 2; }
>   static constexpr std::string_view getDescription() { return "More precise information on solver execution + does output after every cycle."; }
> };
> ```

done

PR #3694, opened 2025-06-04
fix: Replace stackArrayND with StackArray

victorapm:
Closes #3693

Issue #3693, opened 2025-06-04
Build with cuda 12.8 fails


victorapm:
**Describe the bug**
Compilation issues with cuda 12.8.1

**Platform (please complete the following information):**
- Machine: LLNL Matrix (H100)
- Compiler: gcc-12.1.1 + nvcc 12.8.1
- GEOS Version: c2768e6ecd648b17760afa187c44d5a7009d6b72 (develop as of 06/04/25)

**Additional context**
First error:
```
[  9%] Building CUDA object coreComponents/constitutive/CMakeFiles/constitutive.dir/fluid/multifluid/compositional/models/ImmiscibleWaterFlashModel.cpp.o
path-to-geos/src/coreComponents/constitutive/fluid/multifluid/compositional/functions/NegativeTwoPhaseFlash.hpp: In static member function ‘static void geos::constitutive::compositional::NegativeTwoPhaseFlash::computeDerivatives(geos::integer, geos::real64, geos::real64, geos::arraySlice1d<const double>&, const geos::constitutive::compositional::ComponentProperties::KernelWrapper&, const geos::constitutive::compositional::FlashData&, const geos::real64&, geos::arraySlice1d<const double, USD>&, geos::arraySlice1d<const double, USD>&, geos::arraySlice1d<double, USD2>&, geos::arraySlice2d<double, USD3>&, geos::arraySlice2d<double, USD3>&)’:
path-to-geos/src/coreComponents/constitutive/fluid/multifluid/compositional/functions/NegativeTwoPhaseFlash.hpp:383:134: error: template argument 2 is invalid
  383 |     stackArray2d< real64, maxNumComps * maxNumDofs > logLiquidFugacityDerivs( numComps, numDofs );
      |                                                                                                                                      ^
path-to-geos/src/coreComponents/constitutive/fluid/multifluid/compositional/functions/NegativeTwoPhaseFlash.hpp:383:178: error: expression list treated as compound expression in initializer [-fpermissive]
  383 |     stackArray2d< real64, maxNumComps * maxNumDofs > logLiquidFugacityDerivs( numComps, numDofs );
      |                                                                                                                                                                                  ^
path-to-geos/src/coreComponents/constitutive/fluid/multifluid/compositional/functions/NegativeTwoPhaseFlash.hpp:384:134: error: template argument 2 is invalid
  384 |     stackArray2d< real64, maxNumComps * maxNumDofs > logVapourFugacityDerivs( numComps, numDofs );
[...]

PR #3774, opened 2025-08-14
fix: add dane clang build

rrsettgast:
ruby is dead. dane is king. long live dane.

https://github.com/GEOS-DEV/thirdPartyLibs/pull/315


jhuang2601:
@rrsettgast It has already been handled by [this merged PR](https://github.com/GEOS-DEV/GEOS/pull/3708). This one can be closed.

rrsettgast:
Ruby needs to be removed

jhuang2601:
> Ruby needs to be removed

Probably, we should also remove Pangea2 and Pecan? @herve-gross

herve-gross:
Yes, we can remove Pecan and P2

wrtobin (approved):

frankfeifan (approved):

src/coreComponents/common/DataTypes.hpp
```diff
@@ -78,7 +78,7 @@ namespace geos
 using size_t      = std::size_t;
 
 /// Signed integer type.
-using integer     = std::int32_t;
+using integer     = int;
```
rrsettgast, line 81: FYI, @victorapm @wrtobin @CusiniM @dkachuma 

The issue in #3694 and #3693 came up again when I updated our cuda12 versions on the CI. This was the "problem". The compiler was trying to do a cast in a template parameter list, and apparently it is becoming stricter about such things....or it is a compiler bug.

victorapm: Interesting... But `int` and `std::int32_t` have the same size, right? Still puzzling... but thanks Randy!

PR #3775, opened 2025-08-14
fix:  Lambda capture 'mat' is not used

OmarDuran:
In file src/coreComponents/linearAlgebra/interfaces/trilinos/EpetraMatrix.cpp:592:54 there is lambda capture 'mat' 
not used.


victorapm (approved):

rrsettgast (approved):

herve-gross:
Thank you for the fix, I had the same issue.

PR #3777, opened 2025-08-15
feat: Move CO2-Brine parameters to xml

dkachuma:
This PR removes the parameter files used for CO2-Brine model input. This is replaced by direct xml entries. This is a breaking change in terms of user interaction requiring changes to input models. This is evidenced by the extensive changes required in the geos integrated tests.

The key change is the removal of the two fields `phasePVTParaFiles` and `flashModelParaFile` on the `CO2BrineFluid` models. The information that was in these files is now required within the xml itself. This includes:

- Pressure and temperature table ranges. Generally these models involve creation of tables internally and the user was expected to enter the points for these tables by specifying a min, max and step for pressure and temperature. Now the `pressureCoordinates` captures the range while the `pressureInterval` captures the step and similarly for temperature. The new formulation is different from the old in that
  - All properties have the same pressure and temperature discretisation points whereas the old setup allowed different points for the different properties (flash, density, viscosity). 
  - The new setup does not assume equally spaced points. The user can explicitly give the list of pressures in `pressureCoordinates` and ignore `pressureInterval`. This is useful in some cases where a log space might be more appropriate.
 - The salinity with a default of 0.
 - The density and viscosity Ezrokhi coefficients. These can be captured with `ezrokhiDensityCoefficients` and `ezrokhiViscosityCoefficients`. These default to 0 if not specified.
 - The solubility model can now be explicitly defined with choices between (`DuanSun` (default), `SpycherPruess` and `Tables`). The table formulation works in conjunction with the `solubilityTables` field in which the user provides (externally generated) solubility tables directly.
 
The implementation involves creation of the `BrineFluidParameters` object to contain all the parameters. This is then passed around when creating PVT models instead of lists of strings.

The [CO2-Brine documentation](https://geosx-geosx--3777.com.readthedocs.build/en/3777/coreComponents/constitutive/docs/CO2BrineFluid.html) has been updated to reflect not only the changes but also to include all previously developed features that had not been documented.

Changes in results:
- The biggest change is the use of common points for all tables. Some of the examples were using different table ranges for the different properties. This has resulted in reasonable changes in the results now that a common set of points and ranges is used. 

Example change:
```xml
    <CO2BrineEzrokhiFluid
      name="fluid"
      phaseNames="{ gas, water }"
      componentNames="{ co2, water }"
      componentMolarWeight="{ 44e-3, 18e-3 }"
      phasePVTParaFiles="{ tables/pvtgas.txt, tables/pvtliquid_ez.txt }"
      flashModelParaFile="tables/co2flash.txt" />
```
```xml
    <CO2BrineEzrokhiFluid
      name="fluid"
      phaseNames="{ gas, water }"
      componentNames="{ co2, water }"
      componentMolarWeight="{ 44e-3, 18e-3 }"
      pressureCoordinates="{1.0e5, 6e7}"
      pressureInterval="1e5"
      temperatureCoordinates="{283.15, 393.5}"
      temperatureInterval="5"
      salinity="1.901285269"
      ezrokhiDensityCoefficients="{0.1033, -2.2991e-5, -2.3658e-6}"
      ezrokhiViscosityCoefficients="{0, 0, 0}" />
```

Since this is a breaking change, detailed migration instructions are posted at #4022


MelReyCG:
This is really a good thing. This will also allow to simply unit tests.
Will the previous behaviour with external tables be kept for backward compatibility? Or is it possible to include a documentation like _"How to convert PVT parameters files to the new PVT settings."_?

dkachuma:
> Will the previous behaviour with external tables be kept for backward compatibility? 

No. This is a breaking change. The idea is to add an issue describing how to change the parameters.

paveltomin (approved):

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/BrineFluidParameters.cpp
```diff
@@ -0,0 +1,308 @@
...
+                                  viewKeyStruct::temperatureCoordinatesString(), maxTemp ),
+                        InputError );
+
+  if constexpr ( EZROKHI_DENSITY )
+  {
+    if( m_ezrokhiDensityCoefficients.empty())
+    {
+      for( integer ic = 0; ic < 3; ic++ )
+      {
+        m_ezrokhiDensityCoefficients.emplace_back( 0.0 );
```
joshua-white: I suggest we set default values here, and then remove these parameters from the integrated test XMLs unless they differ from the default.  It looks like they all use the same set of values.

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/BrineFluidParameters.cpp
```diff
@@ -0,0 +1,308 @@
...
+                          InputError );
+  }
+
+  if constexpr ( EZROKHI_VISCOSITY )
+  {
+    if( m_ezrokhiViscosityCoefficients.empty())
+    {
+      for( integer ic = 0; ic < 3; ic++ )
+      {
+        m_ezrokhiViscosityCoefficients.emplace_back( 0.0 );
```
joshua-white: Ditto.  Could set good default values.  Users may not know how to set these params properly.

inputFiles/compositionalMultiphaseWell/benchmarks/Class09Pb3/class09_pb3_drainageOnly_direct_base.xml
```diff
@@ -72,8 +72,13 @@
...
       componentMolarWeight="{ 44e-3, 18e-3 }"
-      phasePVTParaFiles="{ tables/pvtgas.txt, tables/pvtliquid_ez.txt }"
-      flashModelParaFile="tables/co2flash.txt"/>
+      pressureCoordinates="{1.0e5, 6e7}"
+      pressureInterval="1e5"
+      temperatureCoordinates="{283.15, 393.5}"
+      temperatureInterval="5"
+      salinity="1.901285269"
+      ezrokhiDensityCoefficients="{0.1033, -2.2991e-5, -2.3658e-6}"
+      ezrokhiViscosityCoefficients="{0, 0, 0}" />
```
joshua-white: Suggest we set default values and remove these coefficients wherever possible from XMLs.

src/coreComponents/constitutive/fluid/multifluid/reactive/ReactiveBrineFluid.cpp
```diff
@@ -130,17 +130,20 @@ void ReactiveBrineFluid< PHASE > ::postInputInitialization()
...
-                        "invalid number of phases",
-                        InputError, getDataContext() );
-  GEOS_THROW_IF_NE_MSG( m_phasePVTParaFiles.size(), 1,
-                        "invalid number of values in attribute ",
-                        InputError, getDataContext() );
+                        GEOS_FMT( "{}: invalid number of phases", getFullName() ),
+                        InputError );
+
+  // Salinity must not be negative
+  GEOS_THROW_IF_LT_MSG( m_salinity, 0.0,
```
joshua-white, line 137: I think this check could be handled by the BrineFluidParameters object itself.  We could centralize all checking of parameter physicality there?

src/docs/sphinx/basicExamples/co2Injection/Example.rst
```diff
@@ -165,30 +165,15 @@ Under the **Constitutive** tag, four items can be found:
...
+The step increments used to construct these tables are provided by the ``pressureInterval`` and ``temperatureInterval`` attributes. 
+Note that in this model formulation, the water component is not allowed to evaporate into the CO :sub:`2`-rich phase.
+The salinity in this case is explicitly set to 0.
 
 .. note::
-   It is the responsibility of the user to make sure that the pressure and temperature values encountered in the simulation (in the reservoir and in the well) are within the bounds specified in the PVT files. GEOS will not throw an error if a value outside these bounds is encountered, but the (nonlinear) behavior of the simulation and the quality of the results will likely be negatively impacted.  
+   It is the responsibility of the user to make sure that the pressure and temperature values encountered in the simulation (in the reservoir and in the well) are within the bounds specified by the coordinate attributes in the XML. GEOS will not throw an error if a value outside these bounds is encountered, but the (nonlinear) behavior of the simulation and the quality of the results will likely be negatively impacted.
```
joshua-white, line 176: Not specific to this PR, but this is such a common user error I wonder if we should revisit the idea that it is not our responsibility to check for out of bounds and warn the user.  Would be nice if this was a flaggable check.  We could probably even just check the initial formation pressure range and any BHP conditions flagged in the XML and say someone is likely to exceed bounds, without doing range checking at every table call mid-simulation.

joshua-white (approved):
Looks great.  Just a few comments regarding providing default values and making sure we do good parameter consistency checking when possible.

src/docs/sphinx/basicExamples/co2Injection/Example.rst
joshua-white: Or once a timestep compute the (max / min) (pressure / temp) over the whole domain and see if those values are in bounds for the table.

src/coreComponents/physicsSolvers/fluidFlow/unitTests/testFlowStatistics.cpp
```diff
@@ -847,7 +851,6 @@ TestSet getTestSet()
 TEST_F( FlowStatisticsTest, checkMultiPhaseFluxStatisticsMass )
 {
   TestSet const testSet = getTestSet();
-  writeTableFiles( testSet.inputs.tableFiles );
```
MelReyCG, line 859: You can also remove the ``TestInputs::tableFiles`` attribute, and the ``FlowStatisticsTest`` class.

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/BrineFluidParameters.cpp
```diff
@@ -0,0 +1,308 @@
...
+    setDescription( "Tolerance to used for fluid property calculation" ).
+    setDefaultValue( m_tolerance );
+
+  // Check if we have the Ezrokhi model
+  if constexpr ( EZROKHI_DENSITY )
+  {
+    fluid->registerWrapper( viewKeyStruct::ezrokhiDensityCoefficientsString(), &m_ezrokhiDensityCoefficients ).
+      setInputFlag( InputFlags::OPTIONAL ).
+      setRestartFlags( RestartFlags::NO_WRITE ).
+      setDescription( "Ezrokhi correlation coefficients for brine density." );
```
MelReyCG: Maybe add link to the doc page where these coefficients are described?

src/docs/sphinx/basicExamples/co2Injection/Example.rst
MelReyCG: IIUC, you talk about the checkTableParameter() checks. Currently the check is only done for wells to minimize performance impact. I think the best behaviour would be to warn the user by default because it is a quite invisible failure, and I join you on the fact that we should only check the min-max access after kernels.

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/BrineFluidParameters.cpp
dkachuma: For CO2 it appears that zero values are a good default. The dependence is quite minimal. So I have left the zero default and added a note in the doc.

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/BrineFluidParameters.cpp
dkachuma: Added default values `{0.1033, -2.2991e-5, -2.3658e-6}` which appear to be the "universal" values for CO2.

src/coreComponents/constitutive/fluid/multifluid/reactive/ReactiveBrineFluid.cpp
dkachuma: This fluid does not use `BrineFluidParameters` to capture the input properties - it simply creates one to use to create the internal fluid object. It has a separate parameters file that was used simply to capture the salinity. I instead removed that and added the salinity as a member of the fluid object directly - hence this separate need to validate.

There are no examples of this fluid model and in fact this file appears to not be included in the CMakelists. It looks like functionality that was not completed.

src/docs/sphinx/basicExamples/co2Injection/Example.rst
dkachuma: I think the idea of checking at least the initial state for the reservoir is good. I have added #4128 to do this.

dkachuma:
Baseline updates
----------------

### 1. Input Format Changes (No Numerical Differences)

The new implementation introduced changes to the input format. For most of the affected cases, this resulted in baseline diffs solely due to this. There are actually no numerical differences.

### 2. PVT Consistency Fixes (`testPVT` cases)

The `testPVT` cases exhibited similar diffs because they required updates to resolve inconsistencies in the original test definitions:

* **Salinity Consistency:** The original cases had inconsistent salinity values between the flash settings (salinity=0) and brine properties (salinity=1). The update ensures that only a single, consistent salinity input (salinity=1) is used across all `testPVT` cases.

* **Temperature Range Consistency:** There was a problem with inconsistent temperature ranges, where some tables had a maximum temperature of 350 while others had 460. All `testPVT` cases now uniformly use a maximum temperature of 460.

### 3. Table Evaluation Variations

The `class09_pb3_smoke_3d` cases show small numerical changes. These differences are consistent with expected variations in the table evaluation caused by kinks in the tables. The changes are minor and within acceptable tolerances.

<table>
  <tbody>
    <tr>
      <td colspan="2"><strong>1. Input Format Changes (No Numerical Differences)</strong></td>
    </tr>
    <tr>
      <td></td>
      <td>
        <code>co2_hybrid_1d_01</code>, <code>co2_hybrid_1d_02</code>, <code>co2_hybrid_1d_03</code>, <code>co2_flux_3d_01</code>, <code>co2_flux_3d_08</code>, <code>co2_flux_3d_27</code>, <code>initialization_2phase_01</code>, <code>initialization_2phase_02</code>, <code>initialization_2phase_no_cappres_01</code>, <code>initialization_2phase_no_cappres_02</code>, <code>staircase_co2_wells_3d_01</code>, <code>staircase_co2_wells_3d_08</code>, <code>staircase_co2_wells_hybrid_3d_01</code>, <code>staircase_co2_wells_hybrid_3d_08</code>, <code>isothm_mass_inj_table_01</code>, <code>isothm_mass_inj_table_02</code>, <code>isothm_vol_inj_table_01</code>, <code>isothm_vol_inj_table_02</code>, <code>staged_close_perf_1_01</code>, <code>staged_close_perf_1_04</code>, <code>staged_close_perf_2_01</code>, <code>staged_close_perf_2_04</code>, <code>staged_open_perf_1_01</code>, <code>staged_open_perf_1_04</code>, <code>staged_open_perf_2_01</code>, <code>staged_open_perf_2_04</code>, <code>PoroElastic_staircase_co2_3d_fim_01</code>, <code>PoroElastic_staircase_co2_3d_fim_04</code>, <code>PoroElastic_staircase_co2_3d_sequential_01</code>, <code>PoroElastic_staircase_co2_3d_sequential_04</code>, <code>PoroElastic_hybridHexPrism_co2_fim_3d_01</code>, <code>PoroElastic_hybridHexPrism_co2_fim_3d_04</code>, <code>PoroElastic_hybridHexPrism_co2_sequential_3d_01</code>, <code>PoroElastic_hybridHexPrism_co2_sequential_3d_04</code>, <code>co2_thermal_2d_01</code>, <code>co2_thermal_2d_04</code>, <code>ThermoPoroElastic_staircase_co2_smoke_01</code>, <code>ThermoPoroElastic_staircase_co2_smoke_04</code>, <code>thermalLeakyWell_smoke_3d_01</code>, <code>thermalLeakyWell_smoke_3d_04</code>, <code>co2_2d_plume_Z_01</code>, <code>co2_2d_plume_Z_04</code>, <code>grav_seg_drain_Z_01</code>, <code>grav_seg_drain_Z_02</code>, <code>ALM_multiphasePoromechanics_curvedFrac_smoke_01</code>, <code>ALM_multiphasePoromechanics_curvedFrac_smoke_04</code>, <code>multiphasePoromechanics_FaultModel_smoke_01</code>, <code>multiphasePoromechanics_FaultModel_smoke_04</code>, <code>multiphasePoromechanics_FaultModel_well_fim_smoke_01</code>, <code>multiphasePoromechanics_FaultModel_well_fim_smoke_04</code>, <code>multiphasePoromechanics_FaultModel_well_seq_smoke_01</code>, <code>multiphasePoromechanics_FaultModel_well_seq_smoke_04</code>[cite: 1]
      </td>
    </tr>
    <tr>
      <td colspan="2"><strong>2. PVT Consistency Fixes</strong></td>
    </tr>
    <tr>
      <td></td>
      <td>
        <code>testPVT_01</code>, <code>testPVT_CO2Brine_01</code>, <code>testPVT_CO2BrineTables_01</code>, <code>testPVT_docExample_01</code>[cite: 1]
      </td>
    </tr>
    <tr>
      <td colspan="2"><strong>3. Table Evaluation Variations</strong></td>
    </tr>
    <tr>
      <td></td>
      <td>
        <code>class09_pb3_smoke_3d_01</code>, <code>class09_pb3_smoke_3d_08</code>, <code>co2_flux_dirichlet_01</code>, <code>co2_flux_dirichlet_06</code>[cite: 1]
      </td>
    </tr>
  </tbody>
</table>

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/BrineFluidParameters.cpp
```diff
@@ -0,0 +1,308 @@
...
+    setDescription( "List of pressure values for interpolation of function values." );
+
+  fluid->registerWrapper( viewKeyStruct::pressureIntervalString(), &m_pressureInterval ).
+    setInputFlag( InputFlags::OPTIONAL ).
+    setRestartFlags( RestartFlags::NO_WRITE ).
+    setDescription( GEOS_FMT( "Step size used to evenly discretize the range pressure range given by "
+                              "{0} into a set of evenly spaced values. If a positive is givem, it defines the "
+                              "spacing between generated points across the full range from the first to the "
+                              "last input value. If zero or a negative number is given then the points in {0} "
+                              " are left as is.", viewKeyStruct::pressureIntervalString() )).
```
OmarDuran: Perhaps this should be viewKeyStruct::pressureCoordinatesString()  instead of viewKeyStruct::pressureIntervalString()?

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/BrineFluidParameters.cpp
```diff
@@ -0,0 +1,308 @@
...
+    setDescription( "List of temperature values for interpolation of function values." );
+
+  fluid->registerWrapper( viewKeyStruct::temperatureIntervalString(), &m_temperatureInterval ).
+    setInputFlag( InputFlags::OPTIONAL ).
+    setRestartFlags( RestartFlags::NO_WRITE ).
+    setDescription( GEOS_FMT( "Step size used to evenly discretize the range temperature range given by "
+                              "{0} into a set of evenly spaced values. If a positive is givem, it defines the "
+                              "spacing between generated points across the full range from the first to the "
+                              "last input value. If zero or a negative number is given then the points in {0} "
+                              " are left as is.", viewKeyStruct::temperatureIntervalString() )).
```
OmarDuran: Perhaps this should be viewKeyStruct:: temperatureIntervalString()  instead of viewKeyStruct:: temperatureCoordinatesString()?

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/BrineFluidParameters.cpp
```diff
@@ -0,0 +1,308 @@
...
+      GEOS_THROW_IF( m_solubilityTables.size() != 1 && m_solubilityTables.size() != 2,
+                     GEOS_FMT( "{}: The number of table names in {} must be 1 or 2", fullName, viewKeyStruct::solubilityTablesString() ),
+                     InputError );
+    }
+  }
+
+  // Water compressibility must be positive
+  GEOS_THROW_IF_LT_MSG( m_waterCompressibility, MultiFluidConstants::epsilon,
+                        GEOS_FMT( "{}: invalid water compressibility {}. "
+                                  "Value must be positive", fullName, viewKeyStruct::waterCompressibilityString() ),
```
OmarDuran: Perhaps it should be m_waterCompressibility

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/BrineFluidParameters.cpp
```diff
@@ -0,0 +1,308 @@
...
+  // Salinity must not be negative
+  GEOS_THROW_IF_LT_MSG( m_salinity, 0.0,
+                        GEOS_FMT( "{}: invalid salinity {}. "
+                                  "Value must not be negative", fullName, viewKeyStruct::salinityString() ),
+                        InputError );
+
+  // Flash tolerance must be positive
+  GEOS_THROW_IF_LT_MSG( m_tolerance, MultiFluidConstants::epsilon,
+                        GEOS_FMT( "{}: invalid flash tolerance {}. "
+                                  "Value must be positive", fullName, viewKeyStruct::toleranceString() ),
```
OmarDuran: Perhaps it should be m_tolerance

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/BrineFluidParameters.cpp
```diff
@@ -0,0 +1,308 @@
...
+  // Water compressibility must be positive
+  GEOS_THROW_IF_LT_MSG( m_waterCompressibility, MultiFluidConstants::epsilon,
+                        GEOS_FMT( "{}: invalid water compressibility {}. "
+                                  "Value must be positive", fullName, viewKeyStruct::waterCompressibilityString() ),
+                        InputError );
+
+  // Salinity must not be negative
+  GEOS_THROW_IF_LT_MSG( m_salinity, 0.0,
+                        GEOS_FMT( "{}: invalid salinity {}. "
+                                  "Value must not be negative", fullName, viewKeyStruct::salinityString() ),
```
OmarDuran: Perhaps it should be m_salinity

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/BrineFluidParameters.cpp
```diff
@@ -0,0 +1,308 @@
...
+  GEOS_THROW_IF_LT_MSG( minTemp, minTempInK,
+                        GEOS_FMT( "{}: Minimum temperature must be at least {}K ({} in C). "
+                                  "The lowest value provided in {} is {}K", fullName,
+                                  minTempInK, minimumTemperature,
+                                  viewKeyStruct::temperatureCoordinatesString(), minTemp ),
+                        InputError );
+  GEOS_THROW_IF_GT_MSG( maxTemp, maxTempInK,
+                        GEOS_FMT( "{}: Maximum temperature must be at most {}K ({} in C). "
+                                  "The highest value provided in {} is {}K", fullName,
+                                  maxTempInK, minimumTemperature,
```
OmarDuran: perhaps this should be maximumTemperature?

OmarDuran (approved):
Thank you, @dkachuma, this makes it much easier to set up compositional simulations.

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/BrineFluidParameters.cpp
dkachuma: You are correct.

src/coreComponents/constitutive/fluid/multifluid/CO2Brine/functions/BrineFluidParameters.cpp
dkachuma: I have actually added both the value and the field name.

jhuang2601 (approved):

PR #3715, opened 2025-06-27
feat: Gravity

bd713:
Prerequisites:
- #3791 for scatter HDF5 outputs
- #3718 for scalar value from Python

Companion PR in `geosPythonPackages`:
- [PR 110](https://github.com/GEOS-DEV/geosPythonPackages/pull/110) for Gravity solver Python interface

PR #3791, opened 2025-08-22
feat: Add scatter data history

bd713:
This PR introduces a `ScatterDataHistoryCollection` feature, enabling collection and output of time history data from scattered points.
Typical application can be found in geophysics, e.g.  data extraction at seismic receivers, or gravimeter stations (see an example in #3715 )

The implementation includes:
- `ScatterDataHistoryCollection`: responsible for collecting, organizing, and outputting scatter data from a specified provider.
- `ScatterDataProvider`:
  - Base interface for any class that supplies scatter data to the output system, defines:
     - `getNumScatterPoints()`: Returns the number of scatter points.
     - `getScatterData()`: Provides the main data array.
     - `getScatterCoordinates()`: Supplies coordinates for each point.
     - `getScatterMetadata()`: Supplies metadata (e.g., station names).
  - Enables decoupling between data sources and output logic, allowing any solver (wave equation, gravity, etc.) to act as a provider.
- HDF5 support, leveraging the output capabilities inherited from `TimeHistoryOutput`
- A comprehensive test

### DIFF
--- a/src/coreComponents/fileIO/CMakeLists.txt
+++ b/src/coreComponents/fileIO/CMakeLists.txt
@@ -37,7 +37,9 @@ set( fileIO_headers
      timeHistory/BufferedHistoryIO.hpp
      timeHistory/PackCollection.hpp
      timeHistory/HDFHistoryIO.hpp
-     timeHistory/HistoryCollection.hpp )
+     timeHistory/HistoryCollection.hpp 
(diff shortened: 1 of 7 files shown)


MelReyCG (approved):
Nice work!
Would you mind adding a simple example or test case demonstrating its usage? (in this PR or later)
This would help users understand how to apply it in practice, and allow to test it in a full case.

PR #3796, opened 2025-08-25
refactor: add solid fields

paveltomin:


src/coreComponents/constitutive/solid/CeramicDamage.cpp
```diff
@@ -46,20 +47,11 @@ CeramicDamage::CeramicDamage( string const & name, Group * const parent ):
...
-
-  registerWrapper( viewKeyStruct::lengthScaleString(), &m_lengthScale ).
-    setApplyDefaultValue( DBL_MIN ).
-    setPlotLevel( PlotLevel::NOPLOT ).
-    setDescription( "Array of quadrature point damage values" );
+  registerField< fields::solid::damage >( &m_damage );
+
+  registerField< fields::solid::jacobian >( &m_jacobian );
+
+  registerField< fields::solid::lengthScale >( &m_lengthScale );
```
MelReyCG, line 54: There is no `DECLARE_FIELD()` for those?

MelReyCG:
Apart from that it look good to me.

src/coreComponents/constitutive/solid/CeramicDamage.cpp
paveltomin: hm how does it build then... let me check

src/coreComponents/constitutive/solid/SolidFields.hpp
```diff
@@ -0,0 +1,429 @@
...
+
+DECLARE_FIELD( jacobian,
+               "jacobian",
+               array2d< real64 >,
+               1.0,
+               NOPLOT,
+               NO_WRITE,
+               "Jacobian of the deformation" );
+
+DECLARE_FIELD( lengthScale,
```
paveltomin, line 125: @MelReyCG it's here

MelReyCG: oh right, nice

MelReyCG (approved):

castelletto1 (approved):
Looks good to me, thank you!

src/coreComponents/constitutive/solid/Damage.cpp
```diff
@@ -143,16 +113,16 @@ void Damage< BASE >::postInputInitialization()
                                            " external driving force flag is turned on" );
 
   // set results as array default values
-  this->template getWrapper< array1d< real64 > >( viewKeyStruct::criticalFractureEnergyString() ).
+  this->template getWrapper< array1d< real64 > >( fields::solid::criticalFractureEnergy::key() ).
```
dkachuma: Since we have done a `registerField`, I think we should also do a `getField` (if that works).
```suggestion
  getField< fields::solid::criticalFractureEnergy >( ).
```

dkachuma (approved):

jhuang2601 (approved):

PR #3799, opened 2025-08-26
refactor: do not allow registerWrapper more than once

paveltomin:


paveltomin:
this is ready for review
rebaseline is needed because of the following three things:

```
Rank 0 is comparing /tmp/tmp.UuRaqTJ0K3/GEOS_integratedTests_working/lagrangianContactMechanics/ContactMechanics_SimpleCubes_smoke_01/0to20_restart_000000010/rank_0000000.hdf5 with /tmp/tmp.UuRaqTJ0K3/GEOS_integratedTests_baselines/lagrangianContactMechanics/ContactMechanics_SimpleCubes_smoke_01/0to20_restart_000000010/rank_0000000.hdf5 
Error: /Problem/domain/MeshBodies/mesh1/meshLevels/Level0/ElementRegions/elementRegionsGroup/Fracture/elementSubRegions/FractureSubRegion
	Group has a child 'pressure_n' in the baseline file but not the file to compare.
Error: /Problem/domain/MeshBodies/mesh1/meshLevels/Level0/ElementRegions/elementRegionsGroup/Fracture/elementSubRegions/FractureSubRegion
	Group has a child 'pressure' in the baseline file but not the file to compare.
```

```
Rank 0 is comparing /tmp/tmp.UuRaqTJ0K3/GEOS_integratedTests_working/proppant/FlowProppantTransport2d_smoke_01/0to10_restart_000000010/rank_0000000.hdf5 with /tmp/tmp.UuRaqTJ0K3/GEOS_integratedTests_baselines/proppant/FlowProppantTransport2d_smoke_01/0to10_restart_000000010/rank_0000000.hdf5 
Error: /Problem/NumericalMethods/FiniteVolume/singlePhaseTPFA/fieldName/__values__
	Datasets have different shapes and therefore can't be compared statistically: (24,), (40,).
	Character arrays have different sizes: 24, 40.
	    pressure
	    pressure  pressure
```

```
Rank 0 is comparing /tmp/tmp.UuRaqTJ0K3/GEOS_integratedTests_working/wavePropagation/acouselas3D_Q2_abc_smoke_01/0to20_restart_000000020/rank_0000000.hdf5 with /tmp/tmp.UuRaqTJ0K3/GEOS_integratedTests_baselines/wavePropagation/acouselas3D_Q2_abc_smoke_01/0to20_restart_000000020/rank_0000000.hdf5 
Error: /Problem/Solvers/elasticSolver
	Group has a child 'sourceConstantsz' in the file to compare but not the baseline file.
Error: /Problem/Solvers/elasticSolver
	Group has a child 'sourceConstantsy' in the file to compare but not the baseline file.
Error: /Problem/Solvers/elasticSolver
	Group has a child 'sourceConstantsx' in the file to compare but not the baseline file.
The files are different.
```

no real results diffs:
`INFO: No unfiltered differences were found.`

MelReyCG:
Just to understand the motivation of this PR, what advantage does that brings? Is that about fine management
of wrapper creation for safer data access?

paveltomin:
> Just to understand the motivation of this PR, what advantage does that brings? Is that about fine management of wrapper creation for safer data access?

I think it is just much safer not to register twice, imagine different solvers registering different things under the same name as an example

MelReyCG (approved):
@paveltomin Yes, I also think it complexify wrapper initialization / configuration.

rrsettgast (approved):

dkachuma (approved):

rrsettgast:
@paveltomin 
After discussing with @wrtobin , I think it is important to list out the challenges in different contexts. 

1. 2 physics solvers register a field "velocity" of the same type that means the same thing. (e.g. solid velocity on nodes)Great. Allow this!
2. 2 physics solvers register a field "velocity" of the same type that means the different things. (solid velocity and fluid velocity on nodes in FEM/FEM context) Oh no...don't allow this!

if we honor 1, we have what is implemented now. if we honor 2, then different physics packages have to prefix their field names somehow? How do we allow different physics packages to share data?


1. 2 constitutive models allocate data on the subregion with the same name and type....with a prefix...ok, but it is hard to access.
2. 2 constitutive models allocate data on the subregion with the same name and type. Oh no....they need to be independent of each other. Don't allow this!

what we have now is 1. If you don't want the prefix, then you have to have multiple field allocations from different constitutive models allocate data... but be accessible with a generic name...but be different. So then you have to add a dimension to the array. So instead of a 1d array, you make it a 2d array. Great...but now you have different types array1d vs array2d, with different interfaces. So you can't generically access them unless it is always a 2d array, and each model context knows what index it is. Same argument goes for the physics packages. If you two views named "velocity" that are same type but different, then what do you do? And what if you have two views with the same name and different types?

rrsettgast:
@paveltomin I guess we could have the field names have a default, but also specifiable in the XML. Thus the user can really have full control over what names are given and what names are expected from different contexts...but that would be kind of heavy.

Issue #3778, opened 2025-08-15
Producer well injecting when BHP limit is higher than reservoir


dkachuma:
[TESTCASE.zip](https://github.com/user-attachments/files/21803221/TESTCASE.zip)

When there is a producer well under BHP control and the BHP is set to a value that's too high for the well to produce, it starts injecting.

**To Reproduce**
Please see the [attached case](https://github.com/user-attachments/files/21803221/TESTCASE.zip). This has a simple `InternalMesh` and a single producer (1 segment, 1 perforation) in the middle of the reservoir. Initially the controlling BHP is 280 bar which is OK to produce. After 10 days this is switched to 320 bar which is too high for the reservoir. In this case the well starts to inject and the reservoir pressure starts to go up. We get this
```
PRODUCER: BHP (at the specified reference elevation): 32000000.000000004 Pa
PRODUCER: Total rate: 163.58947798970624 mol/s; total surface volumetric rate: 3.8497711670206285 sm3/s
PRODUCER: Phase 0 surface volumetric rate: -4.051372363392528e-11 sm3/s
PRODUCER: Phase 1 surface volumetric rate: 3.8497711670206285 sm3/s
```
Positive rates mean injection.

**Expected behaviour**
The well should just shut when the BHP is too high.

**Platform (please complete the following information):**
- GEOS Version 1.1.0 `1561f01eb2`

paveltomin:
https://github.com/GEOS-DEV/GEOS/pull/3800

paveltomin:
@dkachuma ok to close?

PR #3800, opened 2025-08-29
feat: enable crossflow logic for producers

paveltomin:
should address https://github.com/GEOS-DEV/GEOS/issues/3778


dkachuma (approved):
Would this still work if the user disables cross flow?

PR #3801, opened 2025-08-29
fix: Managing geometric objects margins

### LINKED ISSUE #1159: Geometric object margins
Automatically extend geometric object bounds by an epsilon.

_I'm always frustrated when_ I have to manually specify inclusive margins for a `Box` (or read someone else's input file with such margins) e.g.
```xml
<Box xMin="-0.01  8.99 -0.01" 
     xMax=" 1.01 10.01  0.01"/>
```
This is error prone, and if you refine the mesh, you (may) have to manually change the margins as well.

We could automatically extend the bounds of each geometric object by an `eps` in each direction, where `eps = ulp * numeric_limits<real64>::epsilon() * meshBody->m_globalLengthScale` and `ulp` could be between 2 and 10. The above declaration would then look like
```xml
<Box xMin=" 0.0  9.0  0.0" 
     xMax=" 1.0 10.0  0.0"/>
```
which to me is pretty unambiguous, although this opinion may be controversial.

* We could also have `eps` as another (optional) input attribute of each geometric shape, so the user would have full control on the margin size.  Whether it should default to zero or to the length-scale-based value above is a good question. 
* `eps` could also be set on the entire `Geometry` block rather than individual shapes, although I think this is a worse alternative.

paveltomin:
fix https://github.com/GEOS-DEV/GEOS/issues/1159


klevzoff (approved):
Finally! :smile: 

Lower priority, but for consistency it would make sense to also apply this to other objects (`ThickPlane`, `Cylinder`, ...). If you decide to do this, `m_eps` and corresponding code to compute it could be moved to the base class.

rrsettgast:
@paveltomin I would move `m_eps` to the base class. Also I would call it `m_epsilon` since it doesn't really save anything to abbreviate. Also I would make it settable in the xml.

paveltomin:
> @paveltomin I would move `m_eps` to the base class. Also I would call it `m_epsilon` since it doesn't really save anything to abbreviate. Also I would make it settable in the xml.

done

src/coreComponents/mesh/simpleGeometricObjects/SimpleGeometricObjectBase.cpp
```diff
@@ -27,17 +29,27 @@ SimpleGeometricObjectBase::SimpleGeometricObjectBase( string const & name,
...
 }
 
+void SimpleGeometricObjectBase::postInputInitialization()
+{
+  // determine m_epsilon
+  m_epsilon = std::numeric_limits< real64 >::max();
+  DomainPartition & domain = this->getGroupByPath< DomainPartition >( "/Problem/domain" );
+  domain.forMeshBodies( [&]( MeshBody const & meshBody )
+  {
+    m_epsilon = std::min( m_epsilon, 5 * std::numeric_limits< real64 >::epsilon() * meshBody.getGlobalLengthScale() );
```
MelReyCG: The geometric object's epsilon should not be pre-computed here, as the compared value is not yet known here.

This will not work for some values (even with 5x factor, some values can have their next representable value have almost a `10 * std::numeric_limits<T>::epsilon` factor, I'm not sure if it exists with higher factors).

MelReyCG (commented):
(see comment below first)
Another solution if fearing the next / previous value to be compared because of computation imprecision, is rather than using the epsilon constant, to use the [function that really get the exact number before & after](https://en.cppreference.com/w/cpp/numeric/math/nextafter.html) for the comparisons, but that technique can fail after additions/substraction operations.

This is why I suggest to:

- either set `m_epsilon` as a relative "margin" factor, with a default of `3x10^-16`.
Which would give `if( coord0[i] < m_min[i] - m_epsilon*m_min[i] || coord0[i] > m_max[i] + m_epsilon*m_max[i] )`

or 

- generalize `m_thickness` for all geometric objects, with a min default of `3*10^-11`

src/coreComponents/mesh/simpleGeometricObjects/SimpleGeometricObjectBase.cpp
klevzoff: > This will not work for some values (even with 5x factor, some values can have their next representable value have almost a 10 * std::numeric_limits<T>::epsilon factor, I'm not sure if it exists with higher factors).

That's why we're multiplying by `meshBody.getGlobalLengthScale()` - this serves as an upper bound on the range of coordinate values that will be encountered during mesh generation. We end up setting a margin that is equal to 5 ULP of _the largest possible value_, which will be higher than any roundoff error one might encounter during mesh coordinate generation. We can also increase the 5 ULP to 10 ULP or more if this feels unsafe. The point is to have a predictable safety margin that is constant everywhere in the domain, and is also overridable by the user.

klevzoff:
> either set m_epsilon as a relative "margin" factor, with a default of 3x10^-16.
Which would give if( coord0[i] < m_min[i] - m_epsilon*m_min[i] || coord0[i] > m_max[i] + m_epsilon*m_max[i] )

This breaks down when `m_min[i] = 0` (most meshes will be anchored at the origin and will likely have at least one box with 0 coordinates), so you need an extra "absolute" margin for those cases... which brings us back to defining one. Though I do understand the rationale of using relative tolerances.

src/coreComponents/mesh/simpleGeometricObjects/SimpleGeometricObjectBase.cpp
MelReyCG: IIUC, `meshBody.getGlobalLengthScale()` get the diagonal length of the mesh bounding box in the current partition, isn't it?
And the mesh can be offcentered, is that true?

MelReyCG:
IIUC, `meshBody.getGlobalLengthScale()` get the diagonal length of the mesh bounding box in the current partition, isn't it?
And the mesh can be offcentered, is that true?

rrsettgast (approved):

rasimHZ (approved):

src/coreComponents/mesh/simpleGeometricObjects/SimpleGeometricObjectBase.cpp
klevzoff: > IIUC, `meshBody.getGlobalLengthScale()` get the diagonal length of the mesh bounding box in the current partition, isn't it? And the mesh can be offcentered, is that true?

No, it's a _global_ length scale, computed globally across the entire mesh, e.g. [here](https://github.com/GEOS-DEV/GEOS/blob/develop/src/coreComponents/mesh/generators/VTKUtilities.cpp#L2098-L2101). 

For meshes with coordinates very far away from the origin (e.g. `Xmin ~ 1e15, Xmax-Xmin ~ 1e3`) this might not work well, sure. There will likely be other problems with such meshes too (e.g. inaccurate cell volume calculations) but if someone has a good reason for it, they can manually override the default margin.

src/coreComponents/mesh/simpleGeometricObjects/SimpleGeometricObjectBase.cpp
```diff
@@ -27,17 +29,28 @@ SimpleGeometricObjectBase::SimpleGeometricObjectBase( string const & name,
...
-}
-
-
-SimpleGeometricObjectBase::~SimpleGeometricObjectBase()
-{}
 
+  registerWrapper( viewKeyStruct::epsilonString(), &m_epsilon ).
+    setApplyDefaultValue( 1e-6 ).
+    setInputFlag( dataRepository::InputFlags::OPTIONAL ).
+    setDescription( "Tolerance for coordinate checks" );
```
rrsettgast:
```suggestion
    setDescription( "Absolute tolerance for coordinate checks. This is an upper bound value. The minimum of the specified epsilon and 1e-12 * GlobalLengthScale will be used, where GlobalLengthScale is the length of the domain." );
```

rrsettgast (approved):

src/coreComponents/mesh/simpleGeometricObjects/SimpleGeometricObjectBase.cpp
paveltomin: thanks, I clarified the logic a bit here

MelReyCG (approved):
LGTM

Issue #1876, opened 2022-04-14
Nested namespace structure in solvers


francoishamon:
We use very long namespace names, such as `ThermalCompositionalMultiphaseBaseKernels` or `IsothermalCompositionalMultiphaseFVMKernels`. 

It seems that having a more nested namespace structure might be one way to deal with it, e.g. `::geosx::flow::multiphase::kernels::isothermal` or `::geosx::kernels::flow::multiphase::isothermal` (we'd have to think about the order). The benefit is being able to reference any level of this hierarchy by an explicit local alias or a `using` directive, and thus reduce typing.

_Originally posted by @klevzoff in https://github.com/GEOSX/GEOSX/pull/1225#discussion_r841341981_

PR #3803, opened 2025-09-01
refactor: nested namespace structure

paveltomin:
implement proposal from https://github.com/GEOS-DEV/GEOS/issues/1876, for flow solvers only for now

also move zformulation into a separate namespace and simplify flux compute kernel call logic a bit


paveltomin:
This builds and tests are ok, ready for a look

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -75,8 +75,8 @@ CompositionalMultiphaseBase::CompositionalMultiphaseBase( const string & name,
   m_allowCompDensChopping( 1 ),
   m_useTotalMassEquation( 1 ),
   m_useSimpleAccumulation( 1 ),
-  m_minCompDens( isothermalCompositionalMultiphaseBaseKernels::minDensForDivision ),
-  m_minCompFrac( isothermalCompositionalMultiphaseBaseKernels::minCompFracForDivision )
+  m_minCompDens( kernels::fluidFlow::compositional::minDensForDivision ),
```
MelReyCG: Why not having a kernel namespace per namespace? (when needed)
```suggestion
  m_minCompDens( fluidFlow::compositional::kernels::minDensForDivision ),
```
As it is proposed here, it seems like a duplicated tree of solver namespaces.

paveltomin: @rrsettgast @MelReyCG current version is like that:
image
i think what @MelReyCG is suggesting makes a lot of sense and i can swap it around, just confirm that this is the choice
i can also try to wrap solvers into namespaces in a follow-up

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -717,7 +716,7 @@ void CompositionalMultiphaseBase::updateFluidModel( ObjectManagerBase & dataGrou
     using ExecPolicy = typename FluidType::exec_policy;
     typename FluidType::KernelWrapper fluidWrapper = castedFluid.createKernelWrapper();
 
-    thermalCompositionalMultiphaseBaseKernels::
+    kernels::fluidFlow::compositional::
```
dkachuma: Should there be a `thermal` version of this?

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
```diff
@@ -1427,22 +1426,11 @@ void CompositionalMultiphaseBase::assembleSystem( real64 const GEOS_UNUSED_PARAM
...
-  }
-  else
-  {
-    assembleFluxTerms( dt,
-                       domain,
-                       dofManager,
-                       localMatrix,
-                       localRhs );
-  }
+  assembleFluxTerms( dt,
```
dkachuma, line 1429: ~~Do we no longer need the jump stabilised version of this?~~

I see now where this is moved to.

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
rrsettgast: @paveltomin I would agree with swapping the kernels to the end

tjb-ltk: @rrsettgast @MelReyCG  @paveltomin 
Thinking about future consolidation or not...  In Jacobian assemble singlePhase and compositional specialization implement the same interfaces...  separation of single phase and compositional introduces a layer that may not be needed.    Ideally there would be one fluid flow and inside its assemble/update/check methods, switches/factories to call the specialized versions based on model definition. Why duplicate the algorithmic flow code when it doesn't provide any performance optimizations.  having single algorithmic flow at a higher level with specializations inside the algorithmic component methods might be easier to maintain.. going down that path maybe something like fluildFlow:kernels::    would be more appropriate...

paveltomin: i switched to
`fluidFlow::kernels::compositional::...`
that seems logical and it matches the file structure, and it allows to share some kernels and definitions defined at top level between single-phase and compositional in a more clean way
image

dkachuma (approved):

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
paveltomin: this one is general, both for thermal and isothermal

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseBase.cpp
paveltomin: @MelReyCG @rrsettgast @tjb-ltk is it good now?

OmarDuran: I’m not fully involved in this discussion, but I agree that matching the namespace to the file structure is more logical and cleaner.

PR #3808, opened 2025-09-05
feat: Input error when a box does not select anything

arng40:
When a box (usually a user-defined volume) is not selected and the deck does not involve a `SurfaceGenerator`, the simulation is stopped and output an error. 
 
 ```
***** Rank 0: bcPressure (thermalLeakyWell_base.xml, l.168): this FieldSpecification targets (an) empty set(s).
Set 'east':
  - Does not capture: ElementRegions/aquiferBottom
  - Instead, captures: faceManager, edgeManager, nodeManager
Set 'north':
  - Does not capture: ElementRegions/aquiferBottom
  - Instead, captures: faceManager, edgeManager, nodeManager
Set 'south':
  - Does not capture: ElementRegions/aquiferBottom
  - Instead, captures: faceManager, edgeManager, nodeManager
Set 'west':
  - Does not capture: ElementRegions/aquiferBottom
  - Instead, captures: faceManager, edgeManager, nodeManager
```
The user can precise this error state by adding a `errorSetMode` in the targeted `FieldSpecification` wrapper with "silent|warning|error".

Also update the example on 5 units tests where the box defined didn't include the targeted mesh.

Edited the _Thermal Leaky Well_ example. The *smoke* version of this was not implementing the boundary conditions correctly. These have been changed from element based to face based boundary conditions so that they are applicable to both the smoke and benchmark cases. This case has also changed similar to issue #3660. The images have therefore been updated to the current values.

### CODE REVIEW

**Review by paveltomin — DISMISSED** [review]
i am ok with the change but you will need to update all the hydrofrac inputs...

File: src/coreComponents/fieldSpecification/FieldSpecificationManager.cpp [hunk 1]
```diff
@@ -237,14 +237,38 @@ void FieldSpecificationManager::validateBoundaryConditions( MeshLevel & mesh ) c
...
+                                 "The box {} does not select any region.\n"
+                                 "If the simulation does not involve the SurfaceGenerator, check the content of the set `{}` in `{}`.\n",
+                                 fs.getDataContext(), object.getDataContext(),
+                                 mapEntry.first, fs.getObjectPath());
+            if( fs.getErrorAsWarning() )
+            {
+              std::stringstream line;
+              line << fs.getDataContext().getLine();
+              message << GEOS_FMT( "You can set `errorAsWarning` (l. {}) to `0` to disable the error.",
+                                   line.str() );
```
MelReyCG: - This is introspecting too much into DataContext, this class help to prevent coupling with potencial input data. Rewrite without modifying this class.
- The line likely does not exists, as the user did not know it.

MelReyCG (changes requested):

herve-gross (approved):

src/coreComponents/fieldSpecification/FieldSpecificationManager.cpp
```diff
@@ -237,14 +237,36 @@ void FieldSpecificationManager::validateBoundaryConditions( MeshLevel & mesh ) c
...
+        {
+          if( mapEntry.first == object.getName())
+          {
+            std::ostringstream message;
+            message << GEOS_FMT( "{}: this FieldSpecification targets (an) empty set(s)\n"
+                                 "The box {} does not select any region.\n"
+                                 "If the simulation does not involve the SurfaceGenerator, check the content of the set `{}` in `{}`.\n",
+                                 fs.getDataContext(), object.getDataContext(),
+                                 mapEntry.first, fs.getObjectPath());
+            if( fs.getErrorAsWarning() )
```
MelReyCG: - Isn't this condition inverted?
- Can you be a bit more specific with this flag name? (something like, `emptySetErrorAsWarning`)

MelReyCG:
Please explain motivation of exemples modifications in the PR description.

src/coreComponents/fieldSpecification/FieldSpecificationBase.cpp
```diff
@@ -81,6 +81,15 @@ FieldSpecificationBase::FieldSpecificationBase( string const & name, Group * par
     setApplyDefaultValue( 1.0e99 ).
     setInputFlag( InputFlags::OPTIONAL ).
     setDescription( "Time at which the boundary condition will stop being applied." );
+
+  registerWrapper( viewKeyStruct::emptySetErrorModeString(), &m_emptySetErrorMode ).
+    setApplyDefaultValue( setErrorMode::Error ).
+    setInputFlag( InputFlags::OPTIONAL ).
+    setDescription( "Set the log state when we a “set” does not target any region\n"
+                    "When set to 0, no output\n"
```
dkachuma: I'd use `EnumStrings` here to allow users to enter the strings instead of 0,1,2. 
```xml
  errorAsWarning="Silent|Warning|Error"
```
or something like that.

src/coreComponents/fieldSpecification/FieldSpecificationBase.hpp
```diff
@@ -108,6 +108,8 @@ class FieldSpecificationBase : public dataRepository::Group
   /// deleted move assignement
   FieldSpecificationBase & operator=( FieldSpecificationBase && ) = delete;
 
+  enum setErrorMode : integer { Silent = 0, Error = 1, SurfaceGeneratorWarning = 2 };
```
dkachuma:
```suggestion
  enum SetErrorMode : integer { Silent = 0, Error = 1, SurfaceGeneratorWarning = 2 };
```

src/coreComponents/fieldSpecification/FieldSpecificationBase.hpp
```diff
@@ -591,6 +603,8 @@ class FieldSpecificationBase : public dataRepository::Group
   /// The name of a function used to turn on and off the boundary condition.
   string m_bcApplicationFunctionName;
 
+  /// Value indicating whether we converts an error into a warning
+  integer m_emptySetErrorMode;
```
dkachuma:
```suggestion
  SetErrorMode m_emptySetErrorMode;
```

src/coreComponents/fieldSpecification/FieldSpecificationBase.cpp
```diff
@@ -81,6 +81,15 @@ FieldSpecificationBase::FieldSpecificationBase( string const & name, Group * par
     setApplyDefaultValue( 1.0e99 ).
     setInputFlag( InputFlags::OPTIONAL ).
     setDescription( "Time at which the boundary condition will stop being applied." );
+
+  registerWrapper( viewKeyStruct::emptySetErrorModeString(), &m_emptySetErrorMode ).
+    setApplyDefaultValue( setErrorMode::Error ).
```
dkachuma: Default should be current behaviour, no? With `SurfaceGeneratorWarning`.

dkachuma (approved):

src/coreComponents/fieldSpecification/FieldSpecificationBase.hpp
```diff
@@ -108,6 +108,13 @@ class FieldSpecificationBase : public dataRepository::Group
   /// deleted move assignement
   FieldSpecificationBase & operator=( FieldSpecificationBase && ) = delete;
 
+  enum class SetErrorMode : integer
+  {
+    silent,
+    error,
+    surfaceGeneratorWarning
+  };
+
```
MelReyCG:
```suggestion
  enum class SetErrorMode : integer
  {
    silent,
    warning,
    error,
  };

```

src/coreComponents/fieldSpecification/FieldSpecificationBase.hpp
```diff
@@ -591,6 +599,8 @@ class FieldSpecificationBase : public dataRepository::Group
   /// The name of a function used to turn on and off the boundary condition.
   string m_bcApplicationFunctionName;
 
+  /// Value indicating whether we converts an error into a warning
+  string m_emptySetErrorMode;
```
MelReyCG: This is not how enum-strings are meant to be used.
```suggestion
  SetErrorMode m_emptySetErrorMode;
```

src/coreComponents/fieldSpecification/FieldSpecificationBase.cpp
arng40: Wouldn't it better to change the default behavior to error ? Either ways, if we have a surfaceGenerator the behavior will be changed to warning

src/coreComponents/fieldSpecification/FieldSpecificationManager.cpp
```diff
@@ -195,26 +250,65 @@ void FieldSpecificationManager::validateBoundaryConditions( MeshLevel & mesh ) c
...
+          break;
+        case  FieldSpecificationBase::SetErrorMode::error:
+          GEOS_THROW( message.str(), InputError );
+          break;
+        case  FieldSpecificationBase::SetErrorMode::warning:
+          message << GEOS_FMT( "As the simulation includes a SurfaceGenerator, the set may be modified later",
+                               fs.getDataContext() );
+          GEOS_WARNING( message.str() );
+          break;
+      }
```
MelReyCG: - Remove empty case?
- Why `GEOS_FMT()` if no `{}`?
- Possible to take into account SurfaceGenerator presence? (raise a flag from somewhere, or check the data-repo tree?)

```suggestion
      switch( EnumStrings< FieldSpecificationBase::SetErrorMode >::fromString( wrapper.getDefaultValue() ))
      {
        case FieldSpecificationBase::SetErrorMode::error:
          if( isSurfaceGenCase )
            GEOS_THROW( message.str(), InputError );
          break;
        case  FieldSpecificationBase::SetErrorMode::surfaceGeneratorWarning:
          if( isSurfaceGenCase )
            message << "As the simulation includes a SurfaceGenerator, the set may be modified later";
          GEOS_WARNING( message.str() );
          break;
      }
```

src/coreComponents/fieldSpecification/FieldSpecificationManager.cpp
```diff
@@ -195,26 +250,65 @@ void FieldSpecificationManager::validateBoundaryConditions( MeshLevel & mesh ) c
...
     }
 
+    MpiWrapper::barrier( MPI_COMM_GEOS );
```
MelReyCG: - Why this Barrier?

MelReyCG (changes requested):

src/coreComponents/fieldSpecification/FieldSpecificationManager.cpp
arng40: indeed not necessary

src/coreComponents/fieldSpecification/FieldSpecificationManager.cpp
arng40: cannot remove empty case, compil error

MelReyCG: still lack the if for the throw, isn't it?

src/coreComponents/fieldSpecification/FieldSpecificationManager.cpp
```diff
@@ -105,41 +164,36 @@ void FieldSpecificationManager::validateBoundaryConditions( MeshLevel & mesh ) c
...
-      if( targetGroup.hasWrapper( fieldName ) ||
-          flag == InputFlags::FALSE || // no need to check input if the input flag is false (Aquifer, Traction, Equilibrium do not target a
-                                       // field)
-          targetGroup.getName() == MeshLevel::groupStructKeys::faceManagerString() ) // the field names of the face BCs are not always
-                                                                                     // registered on
-                                                                                     // the faceManager...
+      if( targetGroup.hasWrapper( fieldName ) ||flag == InputFlags::FALSE ||
+          targetGroup.getName() == MeshLevel::groupStructKeys::faceManagerString() )     // the field names of the face BCs are not always
+                                                                                         // registered on
+                                                                                         // the faceManager...
```
MelReyCG: Why did you delete those comments? Not useful?
```suggestion
      InputFlags const flag = fs.getWrapper< string >( FieldSpecificationBase::viewKeyStruct::fieldNameString() ).getInputFlag();
      isTargetSetCreated.at( setName ) = 1;
      if( targetGroup.hasWrapper( fieldName ) ||flag == InputFlags::FALSE ||
          targetGroup.getName() == MeshLevel::groupStructKeys::faceManagerString() )     // the field names of the face BCs are not always
                                                                                         // registered on the faceManager...
```

src/coreComponents/fieldSpecification/FieldSpecificationManager.cpp
```diff
@@ -67,24 +68,82 @@ void FieldSpecificationManager::expandObjectCatalogs()
...
+
 
     // Step 1: collect all the set names in a map (this is made necessary by the "apply" loop pattern
 
     string_array const & setNames = fs.getSetNames();
     for( size_t i = 0; i < setNames.size(); ++i )
     {
-      isTargetSetEmpty[setNames[i]] = 1;
       isTargetSetCreated[setNames[i]] = 0;
+      setTypesMap[setNames[i]].push_back( {expectedSetType, 1 } );
```
MelReyCG: Is it possible to have doubles? Should you use a `std::set` rather than a `std::vector`

src/coreComponents/fieldSpecification/FieldSpecificationManager.cpp
```diff
@@ -67,24 +68,82 @@ void FieldSpecificationManager::expandObjectCatalogs()
   }
 }
 
+/// @brief alias for the map allowing to know the existance of given element types (node, edge, cell...)
+/// Represented as : { { "setName", { ObjectPathType (= An element container) , [0 | 1] (not empty) }, ... }, ... }
```
MelReyCG: !!!
```suggestion
/// Represented as : { { "setName", { ObjectPathType (= An element container) , [0 | 1] (is empty) }, ... }, ... }
```

src/coreComponents/fieldSpecification/FieldSpecificationManager.cpp
```diff
@@ -67,24 +68,82 @@ void FieldSpecificationManager::expandObjectCatalogs()
...
+ * @tparam NEXT_TYPES The remaining Types
+ * @param mesh Holds all the managers.
+ * @param targetType The target fieldSpecification object path type
+ * @param setName The current setName to be evaluated in all the managers
+ * @param setTypesMap The map containing the relation between the setNames and their managers
+ */
+template< typename TYPE, typename ... NEXT_TYPES >
+void forManagerInMeshLevel( MeshLevel const & mesh,
+                            MeshObjectPath::ObjectTypes const targetType,
+                            string const & setName, SetNameToTypesMap & setTypesMap )
```
MelReyCG: + -> MeshObjectPath
+ do a functor to set `setTypesMap`

src/coreComponents/fieldSpecification/FieldSpecificationManager.cpp
```diff
@@ -67,24 +68,82 @@ void FieldSpecificationManager::expandObjectCatalogs()
   }
 }
 
+/// @brief alias for the map allowing to know the existance of given element types (node, edge, cell...)
+/// Represented as : { { "setName", { ObjectPathType (= An element container) , [0 | 1] (not empty) }, ... }, ... }
+using SetNameToTypesMap = std::map< std::string, std::vector< std::pair< MeshObjectPath::ObjectTypes, localIndex > > >;
```
MelReyCG: You want a `std::map< std::string, std::map< MeshObjectPath::ObjectTypes, localIndex > >` here

src/coreComponents/fieldSpecification/FieldSpecificationManager.cpp
```diff
@@ -105,41 +164,36 @@ void FieldSpecificationManager::validateBoundaryConditions( MeshLevel & mesh ) c
...
     } );
 
     // Step 3: MPI synchronization
     isFieldNameFound = MpiWrapper::max( isFieldNameFound );
 
-    for( std::pair< string const, localIndex > & mapEntry : isTargetSetEmpty )
+    string_array setNamesToErase;
+    for( auto & [setName, typesValue] : setTypesMap )
+    {
+      auto & typeValue =  std::get< 1 >( typesValue.front() );
```
MelReyCG: please use meaningful names (like `isTypeExisting`)

MelReyCG: I think `setTypesMap` can be improved in that way too.

MelReyCG (changes requested):
Data type issue

herve-gross:
@castelletto1 @corbett5 @bd713  : this one needs your help because the CI does not pass, and @arng40 does not understand why. Thank you!

bd713:
> @castelletto1 @corbett5 @bd713 : this one needs your help because the CI does not pass, and @arng40 does not understand why. Thank you!

When comparing with the reference result, the CI detects that we have introduced a new `errorSetMode`. 
As a result, it sees the outputs as different and considers the test failed... even though the numerical results themselves are likely unchanged.

rrsettgast:
@arng40 
There seem to be some non-trivial changes to the xml files, which may be leading to some of your integratedTest fails. Please revert those changes back to get a valid test result. Then you can change them and rebaseline once you verify that you haven't introduced a diff through code changes.

castelletto1:
As @bd713 pointed out, all integrated tests have restart hdf5 files that show a structural change due to addition of `setErrorMode`, precisely:

```
"Group has a child 'errorSetMode' in the file to compare but not the baseline file."
```

The only two tests that show numerical diffs are:

-  `/compositionalMultiphaseFlow/benchmarks/thermalLeakyWell/thermalLeakyWell_smoke_3d_04/0162.python3_thermalLeakyWell_smoke_3d_04_2_restartcheck_.log`
- `/compositionalMultiphaseFlow/benchmarks/thermalLeakyWell/thermalLeakyWell_smoke_3d_01/0158.python3_thermalLeakyWell_smoke_3d_01_2_restartcheck_.log`

Given the PR description:

> Edited the Thermal Leaky Well example. The smoke version of this was not implementing the boundary conditions correctly. These have been changed from element based to face based boundary conditions so that they are applicable to both the smoke and benchmark cases. This case has also changed similar to issue https://github.com/GEOS-DEV/GEOS/issues/3660. The images have therefore been updated to the current values.

... it’s not surprising to see numerical diffs in these two tests.

Unless I am missing something, I would say we are good to go.

arng40:
With the revert of thermalLeakyWell and the script set to "errorSetMode" I've got 
```
INFO: Total number of log files processed: 1041

INFO: No unfiltered differences were found.

INFO: The following file(s) had at least one error block that was filtered:
```

paveltomin:
does https://geosx-geosx.readthedocs-hosted.com/en/latest/docs/sphinx/advancedExamples/validationStudies/carbonStorage/thermalLeakyWell/Example.html need to be updated?

src/coreComponents/dataRepository/Group.cpp
```diff
@@ -289,7 +289,7 @@ void Group::printDataHierarchy( integer const indent ) const
   for( auto & view : wrappers() )
   {
     GEOS_LOG( string( indent, '\t' ) << "-> " << view.second->getName() << " : "
-                                     << LvArray::system::demangleType( *view.second ) );
+                                     << LvArray::system::demangleType( *view.second ) << " size : "<<view.second->size());
```
paveltomin: add spaces to keep formatting consistent

src/coreComponents/fieldSpecification/FieldSpecificationBase.cpp
```diff
@@ -81,6 +81,15 @@ FieldSpecificationBase::FieldSpecificationBase( string const & name, Group * par
     setApplyDefaultValue( 1.0e99 ).
     setInputFlag( InputFlags::OPTIONAL ).
     setDescription( "Time at which the boundary condition will stop being applied." );
+
+  registerWrapper( viewKeyStruct::errorSetModeString(), &m_emptySetErrorMode ).
+    setInputFlag( InputFlags::OPTIONAL ).
+    setApplyDefaultValue( SetErrorMode::error ).
+    setDescription( "Set the log state when we a “set” does not target any region\n"
```
paveltomin: remove "we"

src/coreComponents/fieldSpecification/FieldSpecificationBase.cpp
```diff
@@ -81,6 +81,15 @@ FieldSpecificationBase::FieldSpecificationBase( string const & name, Group * par
...
+                    "When set to \"silent\", no output\n"
+                    "When set to \"warning\", output an error\n"
+                    "When set to \"error\", output a throw\n" );
```
paveltomin: ????
and "error" above

src/coreComponents/fieldSpecification/FieldSpecificationBase.hpp
```diff
@@ -591,6 +603,8 @@ class FieldSpecificationBase : public dataRepository::Group
   /// The name of a function used to turn on and off the boundary condition.
   string m_bcApplicationFunctionName;
 
+  /// Value indicating whether we converts an error into a warning
```
paveltomin: grammar

src/coreComponents/fieldSpecification/FieldSpecificationManager.cpp
```diff
@@ -69,22 +70,26 @@ void FieldSpecificationManager::expandObjectCatalogs()
 
 void FieldSpecificationManager::validateBoundaryConditions( MeshLevel & mesh ) const
 {
+  DomainPartition const & domain = this->getGroupByPath< DomainPartition >( "/Problem/domain" );
+  Group const & meshBodies = domain.getMeshBodies();
```
paveltomin, line 74: is this needed?

src/coreComponents/fieldSpecification/FieldSpecificationManager.cpp
```diff
@@ -115,31 +120,35 @@ void FieldSpecificationManager::validateBoundaryConditions( MeshLevel & mesh ) c
       //       - FieldSpecification that do not target a field, like Aquifer, Traction, Equilibrium, etc. For these, the check is not
       // necessary (the user cannot mess up)
       //       - Face boundary conditions that target cell-based quantities, like the face BC of the flow solvers
-      if( targetGroup.hasWrapper( fieldName ) ||
-          flag == InputFlags::FALSE || // no need to check input if the input flag is false (Aquifer, Traction, Equilibrium do not target a
-                                       // field)
-          targetGroup.getName() == MeshLevel::groupStructKeys::faceManagerString() ) // the field names of the face BCs are not always
-                                                                                     // registered on
-                                                                                     // the faceManager...
+      if( targetGroup.hasWrapper( fieldName ) ||flag == InputFlags::FALSE ||
```
paveltomin: formatting

paveltomin (approved):

paveltomin:
> does https://geosx-geosx.readthedocs-hosted.com/en/latest/docs/sphinx/advancedExamples/validationStudies/carbonStorage/thermalLeakyWell/Example.html need to be updated?

oh i see it was indeed updated...

paveltomin (approved):

src/coreComponents/fieldSpecification/FieldSpecificationManager.cpp
arng40: removed the duplication and let this one

MelReyCG (approved):

jhuang2601 (approved):

acitrain (approved):

PR #3811, opened 2025-09-09
fix: issue 3532

### LINKED ISSUE #3532: A bug in the FieldSpecification of the hydraulic fracturing tutorial example
**Describe the bug**
When assigning `bulk_modulus`, `shear_modulus`, and all initial stress conditions in `FieldSpecifications` in [heterogeneousInSitu_base.xml](https://github.com/GEOS-DEV/GEOS/blob/develop/inputFiles/hydraulicFracturing/heterogeneousInSitu_base.xml), we currently have `objectPath="ElementRegions"`, which is incomplete. We should change it to `objectPath="ElementRegions/Domain"`, otherwise, it will check if the `Fracture` region also has these variables and apparently it does not. 

In [heterogeneousInSitu_smoke.xml](https://github.com/GEOS-DEV/GEOS/blob/develop/inputFiles/hydraulicFracturing/heterogeneousInSitu_smoke.xml), we redefine the `FieldSpecifications` with the correct `objectPath`, so that's why there is no issue running `integratedTests`. However, in [heterogeneousInSitu_benchmark.xml](https://github.com/GEOS-DEV/GEOS/blob/develop/inputFiles/hydraulicFracturing/heterogeneousInSitu_benchmark.xml), it directly takes `FieldSpecifications` from the base .xml, so it will give error when running it. 

**To Reproduce**
Simply run `./geosx_preprocessed -i heterogeneousInSitu_benchmark.xml` should throw the error like something below 

`Rank 0: Group /domain/MeshBodies/mesh1/meshLevels/Level0/ElementRegions/elementRegionsGroup/Fracture(heterogeneousInSitu_benchmark.xml.preprocessed,l.142)/elementSubRegions/FractureSubRegion has no wrapper named rock_bulkModulus`

**Additional context**
We should also check the convergence of the benchmark problem.

paveltomin:
should fix https://github.com/GEOS-DEV/GEOS/issues/3532


frankfeifan (approved):
Thanks for taking care of this issue!

jhuang2601 (approved):

PR #3812, opened 2025-09-10
fix: Implement state initialisation for compositional fluid

dkachuma:
This PR adds a k-value reset for compositional fluid. This allows the well model to force a re-initialisation of the k-values when well conditions switch to guarantee a better fluid state.

### DIFF
--- a/src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
+++ b/src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
@@ -680,6 +680,7 @@ void CompositionalMultiphaseWell::updateVolRatesForConstraint( WellElementSubReg
 
   string const & fluidName = subRegion.getReference< string >( viewKeyStruct::fluidNamesString() );
   MultiFluidBase & fluid = subRegion.getConstitutiveModel< MultiFluidBase >( fluidName );
+  fluid.initializeState();
   integer const isThermal = fluid.isThermal();
   arrayView3d< real64 const, constitutive::multifluid::USD_PHASE > const & phaseFrac = fluid.phaseFraction();
   arrayView4d< real64 const, constitutive::multifluid::USD_PHASE_DC > const & dPhaseFrac = fluid.dPhaseFraction();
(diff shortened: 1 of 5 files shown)


src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
```diff
@@ -902,6 +903,8 @@ void CompositionalMultiphaseWell::updateFluidModel( WellElementSubRegion & subRe
   string const & fluidName = subRegion.getReference< string >( viewKeyStruct::fluidNamesString() );
   MultiFluidBase & fluid = subRegion.getConstitutiveModel< MultiFluidBase >( fluidName );
 
+  fluid.initializeState();
```
paveltomin: feels not right to call it here all the time?

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
```diff
@@ -680,6 +680,7 @@ void CompositionalMultiphaseWell::updateVolRatesForConstraint( WellElementSubReg
 
   string const & fluidName = subRegion.getReference< string >( viewKeyStruct::fluidNamesString() );
   MultiFluidBase & fluid = subRegion.getConstitutiveModel< MultiFluidBase >( fluidName );
+  fluid.initializeState();
```
paveltomin: also calling it here feels a bit "random"

dkachuma: Yes it's not very elegant. The well is using the same object to do the reservoir and surface condition flashes. This is requiring the k-values from the previous flash to be cleared out every time.

tjb-ltk: In the estimator PR a separate fluid model is created for the constraint flash.   could be added to this PR, its relatively simple.... otherwise yes not elegant but needed for now...

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
dkachuma: I have removed the calls from the well model. I will leave that to @tjb-ltk with the well estimator code to put the calls in the right place. But we still need this means to reset the k-values.

paveltomin (approved):

tjb-ltk (approved):

PR #3813, opened 2025-09-10
feat: Well Separator

tjb-ltk:
Create a "separator" fluid model to compute quantities associated with constraint equation.  P & T conditions are user defined as either surface,  top well element, or reservoir region conditions.  Composition is taken from well element associated with constraint equation.  Previously the fluid model that computed well element quantities was also used for constraint flash, and shared property arrays.    This limited surface separator definitions , code design, and other fluid model optimizations.


src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
```diff
@@ -565,6 +565,15 @@ void CompositionalMultiphaseWell::initializePostInitialConditionsPreSubGroups()
       string const & fluidName = subRegion.getReference< string >( viewKeyStruct::fluidNamesString() );
       MultiFluidBase & fluid = subRegion.getConstitutiveModel< MultiFluidBase >( fluidName );
       fluid.setMassFlag( m_useMass );
+
+      // setup fluid separator
+      WellControls & wellControls = getWellControls( subRegion );
```
paveltomin, line 593: this will probably be more clean sitting inside of well control?

paveltomin (approved):

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
tjb-ltk: Yes but I couldn't come up with clean way of creating inside well control,  it needs the subregion.... when we change well control to become the compositionalMultiphaseWell,  and compositionalMultiphaseWell becomes the WellManager , then its already in the right place, but     In the well estimator dev there several other objects being push/pulled from well control because the object model isn't quite right...

rrsettgast (approved):

dkachuma (approved):

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
```diff
@@ -565,10 +589,16 @@ void CompositionalMultiphaseWell::initializePostInitialConditionsPreSubGroups()
       string const & fluidName = subRegion.getReference< string >( viewKeyStruct::fluidNamesString() );
       MultiFluidBase & fluid = subRegion.getConstitutiveModel< MultiFluidBase >( fluidName );
       fluid.setMassFlag( m_useMass );
+      // setup fluid separator
+      WellControls & wellControls = getWellControls( subRegion );
+      string const fluidSeparatorName = wellControls.getName() + "Separator";
+      std::unique_ptr< constitutive::ConstitutiveBase >  fluidSeparatorPtr  = fluid.deliverClone( fluidSeparatorName, &fluid );
```
dkachuma:
```suggestion
      std::unique_ptr< constitutive::ConstitutiveBase >  fluidSeparatorPtr  = fluid.deliverClone( fluidSeparatorName, &wellControls );
```

PR #3815, opened 2025-09-11
fix: stats output logic

paveltomin:


src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
```diff
@@ -127,11 +127,11 @@ PhysicsSolverBase::PhysicsSolverBase( string const & name,
 void PhysicsSolverBase::postInputInitialization()
 {
   m_solverStatistics.setOutputFilesName( getName() );
-  m_solverStatistics.makeDir( m_writeStatisticsCSV >= 2 );
+  m_solverStatistics.makeDir( m_writeStatisticsCSV >= 1 );
 
   getIterationStats().setTableName( getName() );
-  getIterationStats().setLogOutputState( m_writeStatisticsCSV >= 1 );
-  getIterationStats().setCSVOutputState( m_writeStatisticsCSV >= 2 );
+  getIterationStats().setLogOutputState( true );
```
MelReyCG, line 133: No log level for that?

Just suggesting, feel free to ignore / adapt
```suggestion
  getIterationStats().setLogOutputState( isLogLevelActive< Statistics >( this->getLogLevel() ) );
```

MelReyCG (approved):

src/coreComponents/physicsSolvers/multiphysics/CoupledSolver.hpp
```diff
@@ -482,7 +482,7 @@ class CoupledSolver : public PhysicsSolverBase
       for( iter = 0; iter < solverParams.m_maxIterNewton; iter++ )
       {
         // Increment the solver statistics for reporting purposes
-        getIterationStats().incrementConfigIteration();
+        getIterationStats().updateNonlinearIteration( 0 );
```
MelReyCG, line 485: Wow, didn't catch this one in previous PR

src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
paveltomin: By default it always printed iteration stats at the end of the simulation and I think we better preserve that. I recently made it to print iteration stats after every cycle for log level >= 2 which is useful for convergence investigations. That logInfo:: Statistics has a different flavor, I was also confused yesterday, maybe we should rename it to make more specific.

arng40 (approved):

jhuang2601 (approved):

PR #3816, opened 2025-09-12
fix: workaround for vtkRedistributeDataSetFilter

bd713:
This PR introduces a workaround for a regression in `vtkRedistributeDataSetFilter` observed in VTK versions 9.4.2 and newer.

Since VTK v9.4.2, `vtkRedistributeDataSetFilter` fails to correctly redistribute hybrid meshes containing **tetrahedra** or **pyramids**, resulting in **_silent_ cell loss**. This issue does not seem to happen in VTK v9.3.1, but has been reproduced in both v9.4.2 and v9.5.1 (on large hybrid meshes >1M cells and at high rank counts >48), even in minimal standalones **independent of GEOS**. 

For context, our parallel workflow relies on `vtkRedistributeDataSetFilter`'s kd-tree capability to perform an initial, balanced distribution of any VTK mesh that is first read only by rank 0. This step is a prerequisite for parallel partitioners like ParMetis, which assume all ranks already hold a portion of the mesh.

In practice, this PR implements a "detect-and-fallback" mechanism:
* After applying `vtkRedistributeDataSetFilter`, the code checks whether any cells were lost.
* If cell loss is detected, it falls back to a basic `scatterByBlock` method, which redistributes the mesh from rank 0 using contiguous cell blocks.

As opposed to a kd-tree, this fallback approach does not rely on any geometric information, so using  a partitioning tool such as `ParMetis` or `PTScotch` in post-processing is highly recommended to optimize the resulting partitions.


MelReyCG:
We will test this one on our side.
Is there an issue on KitWare side for the kdTree issue?

src/coreComponents/mesh/generators/VTKUtilities.cpp
```diff
@@ -705,12 +804,30 @@ redistributeByKdTree( vtkDataSet & mesh )
...
+  vtkSmartPointer< vtkDataSet > result = vtkDataSet::SafeDownCast( rdsf->GetOutputDataObject( 0 ) );
+  
+  // Verify we didn't lose any cells
+  vtkIdType localOutputCells = result->GetNumberOfCells();
+  vtkIdType globalOutputCells = MpiWrapper::allReduce( localOutputCells, MpiWrapper::Reduction::Sum, MPI_COMM_GEOS );
+  
+  if( globalOutputCells != globalInputCells )
+  {
+    GEOS_LOG_RANK_0( "WARNING: VTK KdTree redistribution lost " << (globalInputCells - globalOutputCells) 
+                      << " elements. Falling back to block redistribution." );
```
MelReyCG:
```suggestion
    GEOS_WARNING( "VTK KdTree redistribution lost " << (globalInputCells - globalOutputCells) 
                      << " elements! Falling back to block redistribution." );
```

bd713:
> Is there an issue on KitWare side for the kdTree issue?

Not yet, I need to coordinate with @margauxraguenel for that.

paveltomin (approved):

jhuang2601 (approved):
Thanks for fixing this hidden issue. 
With it, missing tet elements can be detected and recovered.

PR #3821, opened 2025-09-17
test: enable parallel for contact tests

### LINKED ISSUE #2014: Reactivate parallel smoke test for contact mechanics
In PR #1992 the only parallel integrated test using a mesh that is not internally generated, i.e. `ContactMechanics_SingleFracCompression_03`, has been disabled because of ghosting problem (see issue #841). The test should reactivated after #841 has been resolved.

paveltomin:
~if works, will resolve https://github.com/GEOS-DEV/GEOS/issues/2014~
didn't work but it is still worth to activate some parallel tests


paveltomin:
does not work unfortunately
image

rrsettgast (approved):

inputFiles/lagrangianContactMechanics/contactMechanics.ats
```diff
@@ -50,15 +50,15 @@ decks = [
         name="ContactMechanics_SingleFracCompression_smoke",
         description=
         "Single tilted fracture subjected to remote compression (unstructured grid)",
-        partitions=((1, 1, 1), ),
+        partitions=((1, 1, 1),),
```
jhuang2601, line 53: So the issue persists when using the current contact solver and VTK mesh in parallel runs. 
And ALM solver is stable.

jhuang2601 (approved):
Thanks for this update.

inputFiles/lagrangianContactMechanics/contactMechanics.ats
paveltomin: right

Issue #3823, opened 2025-09-18
Crash in Debug Mode Multiphase Contact Mechanics


dkachuma:
**Describe the bug**
The simulation case crashes at the first time step in debug mode due to bad offsets, despite completing initialization successfully. The setup involves multiphase contact mechanics with no sources or sinks, and uses `HydrostaticEquilibrium` and `MultiphasePoromechanicsConformingFracturesInitialization`. If the setup is incorrect, the code should fail gracefully with a clear error message.

**To Reproduce**
1. Load the attached case [zip file](https://github.com/user-attachments/files/22412501/multiphaseFault.zip).
2. Run in debug mode.
3. Observe the crash at the first time step.
4. Inspect the stack trace for offset-related failure.

**Expected behavior**
The simulation should either proceed correctly or fail gracefully during initialization with a clear diagnostic message if the setup is invalid.

**Screenshots**
Stack trace is attached in the bug report.

**Platform (please complete the following information):**
- Machine: pine
- Compiler: gcc 12.2.0
- GEOS Version: 1.1.0 (develop, sha1: 96356b00ba)

**Additional context**
- Case includes multiphase contact mechanics.
- No sources or sinks are defined.
- Initialization completes without error.
- Crash occurs at first time step in debug mode.
- Stack trace and case zip are attached.

[multiphaseFault.zip](https://github.com/user-attachments/files/22412501/multiphaseFault.zip)
[stackTrace.txt](https://github.com/user-attachments/files/22412504/stackTrace.txt)

jhuang2601:
@dkachuma I can reproduce the same issue in `debug` mode:
```
  Number of element for each fracture state: stick =            0 | new slip =            0 | slip =             0 | open =           800
  Attempt:  0, ConfigurationIter:  0, NewtonIter:  0
***** ERROR
***** LOCATION: /shared/data1/Users/j0551570/Compilation/Develop_250918/GEOS/src/coreComponents/LvArray/src/CRSMatrixView.hpp:545
***** Controlling expression (should be false): !(columns[ pos ] == cols[ i ])

  columns[ pos ] = 57723
  cols[ i ] = 57699


** StackTrace of 24 frames **
Frame 0: void LvArray::CRSMatrixView<double, long long const, int const, LvArray::ChaiBuffer>::addToRowBinarySearchUnsorted<RAJA::policy::sequential::seq_atomic>(int, long long const*, double const*, long long) const 
[...]

PR #3834, opened 2025-09-25
fix: clean up set sparsity pattern logic + matrix setup for poromechanics with contact and wells

paveltomin:
~this is preparatory cleanup for fixing https://github.com/GEOS-DEV/GEOS/issues/3823~
in `setupSystem`, matrix pattern build is moved into a separate function to allow code re-use and calling pattern build separately from full setup
merged changes from https://github.com/GEOS-DEV/GEOS/pull/3839


paveltomin:
this is ready for review

src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
```diff
@@ -1209,6 +1209,13 @@ void PhysicsSolverBase::setupSystem( DomainPartition & domain,
   solution.create( dofManager.numLocalDofs(), MPI_COMM_GEOS );
 }
 
+void PhysicsSolverBase::setSparsityPattern( DomainPartition & GEOS_UNUSED_PARAM( domain ),
+                                            DofManager & dofManager,
+                                            SparsityPattern< globalIndex > & pattern )
+{
+  dofManager.setSparsityPattern( pattern );
```
castelletto1, line 1217: At some point we may want to adopt a consistent style and chose between `GEOS_UNUSED_PARAM` and `GEOS_UNUSED_VAR`.

castelletto1 (approved):

PR #3839, opened 2025-09-26
fix: matrix setup for poromechanics with contact and wells

### LINKED ISSUE #3823: Crash in Debug Mode Multiphase Contact Mechanics

paveltomin:
depends on https://github.com/GEOS-DEV/GEOS/pull/3834
trying to fix https://github.com/GEOS-DEV/GEOS/issues/3823
- one fix is for case with wells, to avoid overwriting matrix pattern when wells are added
- another fix is for multiphase where component loop was missing in two places

### DIFF
--- a/src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
+++ b/src/coreComponents/physicsSolvers/PhysicsSolverBase.cpp
@@ -1197,7 +1197,7 @@ void PhysicsSolverBase::setupSystem( DomainPartition & domain,
   if( setSparsity )
   {
     SparsityPattern< globalIndex > pattern;
-    setSparsityPattern( domain, dofManager, pattern );
+    setSparsityPattern( domain, dofManager, localMatrix, pattern );
     localMatrix.assimilate< parallelDevicePolicy<> >( std::move( pattern ) );
   }
   localMatrix.setName( this->getName() + "/matrix" );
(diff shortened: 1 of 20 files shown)


paveltomin:
Ready for review

castelletto1 (approved):

PR #3828, opened 2025-09-23
feat: Adding error YAML file in GEOS

MelReyCG:
This PR aims to provide a file that stores warnings and errors regarding input data that occur during the current execution.

To do this, I created a class called `ErrorLogger` that retrieves, formats, and sends the desired information to a YAML file.
I also have (in the Logger.hpp file) two macros for each of the following cases:

- Warning
- Error
- Throw

For each of these cases, one of these macro allows to attach contextual information about the error/warning.
Each of the macros is based on the `ErrorLogger` class.

```
- type: Exception
    rank: 0
    message: >-
      compositionalMultiphaseWell (simpleCo2Inj_inc_erreurTable.xml, l.34): wrong surface pressure / temperature.
      Table input error for water phase (in table from "/path/to/randomExemple/pvtliquid_erreur.txt").
      /Functions(simpleCo2Inj_inc_erreurTable.xml,l.203)/fluid_phaseModel1_PhillipsBrineDensity_table:
      Requested pressure of 101325 [Pa] is out of the table bounds
      ( lower bound: pressure of 10000000 [Pa] -> upper bound: pressure of 75000000 [Pa] ).
    contexts:
      - dataPath: /Functions(simpleCo2Inj_inc_erreurTable.xml,l.203)/fluid_phaseModel1_PhillipsBrineDensity_table
        priority: 0
      - dataPath: /domain/MeshBodies/cartesianMesh/meshLevels/Level0/ElementRegions/elementRegionsGroup/
                  wellRegion(simpleCo2Inj_inc_erreurTable.xml,l.64)/elementSubRegions/wellRegionUniqueSubRegion/
                  ConstitutiveModels/fluid
        priority: 2
      - inputFile: /path/to/randomExemple/simpleCo2Inj_inc_erreurTable.xml
        inputLine: 34
        priority: 1
    sourceLocation:
      file: /path/to/GEOS/src/coreComponents/functions/TableFunction.cpp
      line: 191
    sourceCallStack:
      - frame0:  geos::constitutive::PVTProps::PhillipsBrineDensity::checkTablesParameters(double, double) const 
      - frame1:  geos::constitutive::CO2BrineFluid::checkTablesParameters(double, double) const 
      - frame2:  geos::CompositionalMultiphaseWell::validateWellControlsForFluid() const 
      - frame3:  geos::CompositionalMultiphaseWell::validateConstitutiveModels(geos::DomainPartition const&) const 
[...]

PR #3856, opened 2025-10-08
feat: Adding signal and external error detection & output

MelReyCG:
This PR is based on Amandine work on adding error YAML file in GEOS (PR #3828), and aims at adding a detection & management inside GEOS of 1. Error signals, 2. External errors from dependencies, in order to be able to manage & output them in the log & error YAML file.

Managing those external errors gives us the opportunity to:
- output error message & context **reliably** in the log, even if the stderr get lost or used for another reason,
- be sure to detect any kernel / system allocator errors and add the stack-trace of these errors,
- factorize them with external tools / scripts, thus highlighting which are the source rank(s) of the issue.
- also has the effect to prevent the stacktrace to be cut by other ranks message, which could previously happen on a signal.

Without the pipe redirection proposed here, we encounter a lot of errors without any stacktrace and too minimalist (and sometimes even no message at all, just a cut log) on many HPC platforms and on the CI. The branch has been tested to provide the lacking info on many issues on P4.

This work also gives us the opportunity to tag later each dependency message (system, LvArray, Hypre, ...) to quickly identify / filter issues source.

----

(Replaces #3722)


src/coreComponents/common/logger/ExternalErrorHandler.cpp
```diff
@@ -0,0 +1,204 @@
...
+  // backup original descriptor
+  m_originalStreamTarget = ::dup( m_redirectedStream );
+  if( m_originalStreamTarget == m_errorResult )
+  {
+    ::close( m_deviationPipe.readEnd() );
+    ::close( m_deviationPipe.writeEnd() );
+    GEOS_WARNING( "Failed to duplicate original descriptor: " + std::string( strerror( errno ) ) );
+  }
+
+  // Redirect stderr to pipe
```
arng40, line 100:
```suggestion
  // make the deviation pipe point to what the original pipe point
```

src/coreComponents/common/logger/ExternalErrorHandler.cpp
```diff
@@ -0,0 +1,204 @@
...
+  if( ::dup2( m_deviationPipe.writeEnd(), m_redirectedStream ) == m_errorResult )
+  {
+    ::close( m_originalStreamTarget );
+    ::close( m_deviationPipe.readEnd() );
+    ::close( m_deviationPipe.writeEnd() );
+    m_originalStreamTarget = m_disabledPipe;
+    GEOS_WARNING( "Failed to redirect stream: " + std::string( strerror( errno ) ) );
+  }
+
+  // Close the write end of the parent pipe (we write through stderr now)
```
arng40:
```suggestion
  // Close the writeEnd of the deviation pipe since the original pipe point to it
we can now read through the pipe
```

src/coreComponents/common/logger/ExternalErrorHandler.cpp
```diff
@@ -0,0 +1,204 @@
...
+  ::close( m_deviationPipe.writeEnd() );
+  m_deviationPipe.writeEnd() = m_disabledPipe;
+
+  // various optimizations
+  m_unprocessedData.reserve( 16384 );
+
+  #ifdef __linux__
+  ::fcntl( m_deviationPipe.readEnd(), F_SETPIPE_SZ, 1048576 );
```
arng40:
```suggestion
  maybe add a comment for this line
```

src/coreComponents/common/logger/ExternalErrorHandler.cpp
```diff
@@ -0,0 +1,204 @@
...
+  enableStderrPipe( false );
+}
+
+ExternalErrorHandler & ExternalErrorHandler::instance()
+{
+  static ExternalErrorHandler instance;
+  return instance;
+}
+
+void ExternalErrorHandler::enableStderrPipe( bool enable )
```
arng40: enableStderrPipeDeviation ?

arng40:
Can you give some output example in the description of your PR please

MelReyCG:
@rrsettgast @wrtobin @paveltomin I need a review, this work will be useful for further debugging tasks

paveltomin:
> @rrsettgast @wrtobin @paveltomin I need a review, this work will be useful for further debugging tasks

i more or less understand what is done here but don't really understand how it works, sorry
seems very low-level, can you share some examples - situation before and after ?

MelReyCG:
@paveltomin thanks for the feedback, I'll work on the code clarity!

paveltomin (approved):

wrtobin (approved):
My only note is we might prefer `ExternalSignalHandler` just for specificity/clarity vs our existing error handling mechanisms, but not a hard requirement.

MelReyCG:
I would like to keep the name general as this component does not process only signals, but also any message from external code (as an exemple, VTK error / warning messages).
Some messages may appear as (non-crashing) error for now, but we will be able categorize them with this component (or a sub-component of it).
Thanks for the review

PR #3871, opened 2025-10-17
fix: Fix the calculate ghost logic

jafranc:
I think there is a missed condition in the ghosting regarding `calcultateNormResidual` that:
 - Run into a SegFault in Debug (probably in bound check too)
 - Prevent correct convergence in Release

Indeed, as we are browsing the faces of the `SurfaceElementRegion` is is necessary but not sufficient to check that the `face element` is not a ghost, as we write in its neighbors `faces` residual position, it should also be checked that those neighbors do not belong to ghost either.

- [x] Avoid those cases by rejecting writing out-of -bound 
- [x] Check that it restore convergence and is consistent with the single rank solution
- [x] Use `faceManager` 's ghosting


src/coreComponents/constitutive/ConstitutivePassThru.hpp
```diff
@@ -307,6 +307,27 @@ struct ConstitutivePassThru< DamageBase >
 };
 
 
+template<>
+struct ConstitutivePassThru< PorousSolid< ElasticIsotropic, ConstantPermeability > >
```
paveltomin: not sure if you want to keep this here in this PR

paveltomin (approved):

PR #3875, opened 2025-10-21
fix: Fix scatterByBlock initialization

bd713:
Correctly initialize the VTK structure, even for ranks with no data.

### DIFF
--- a/src/coreComponents/mesh/generators/VTKUtilities.cpp
+++ b/src/coreComponents/mesh/generators/VTKUtilities.cpp
@@ -793,8 +793,14 @@ scatterByBlock( vtkDataSet & mesh )
   }
   else
   {
-    // Other ranks have empty mesh
-    localParts->SetNumberOfPartitions( 0 );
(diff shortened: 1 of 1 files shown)


rrsettgast (approved):
Why would we have an empty mesh of a partition?

bd713:
> Why would we have an empty mesh of a partition?

When `scatterByBlock` is called, only rank 0 has the full mesh; all other ranks have no data.
Rank 0 prepares the partitions to be sent to all ranks via `vtk::redistribute`.

Even if a rank has no data **before** `vtk::redistribute`, it still must set the partition count in `localParts` to match the total number of partitions (comm size). There is also an assert in the `vtk::redistribute` function that enforces this requirement.
To be explicit, the corresponding partitions are marked as empty.

**After** redistribute, all ranks should have a non-empty mesh partition, except in some very pathological cases (e.g. more ranks than elements in the mesh).

paveltomin:
@bd713 merge?

bd713:
> @bd713 merge?

yes, thanks!

PR #3880, opened 2025-10-23
fix: sequential poromechanics for single-phase

paveltomin:
fix a bug introduced in https://github.com/GEOS-DEV/GEOS/pull/3485
basically mass that is used in accumulation term was not updated with porosity change after mechanics leading to always converged sequential outer loop


castelletto1 (approved):
Looks good to me, thank you!

jhuang2601 (approved):
Thanks for the fix and it works for the Mandel_Seq case.

rrsettgast (approved):

PR #3884, opened 2025-10-27
fix: correct the effective stress calculation for thermoporomechanics

frankfeifan:
This PR aims to correct the current GEOS effective stress calculation such that it is consistent with Terzaghi definition. Detailed explanations on the issue and this change can be seen in the file attached. 
[THMEffStress_GEOS.pdf](https://github.com/user-attachments/files/26957377/THMEffStress_GEOS.pdf)

- [x] Correct the effective and total stress calculation as well as `dTotalStress_dTemperature` in `PorousSolid` for the thermal case
- [x] Add an integrated test to verify the implementation
- [x] Remove the thermal strain part from the `avgPlasticStrain` calculation. (assumed `avgPlasticStrain` to be purely mechanical deformation). 
- [x] Revise the THM tutorial benchmark problem accordingly ([PR#4110](https://github.com/GEOS-DEV/GEOS/pull/4110))

Another PR(s) needed to fix following items

- Fix the porosity update for both full implicit case and fixed-stress case (see formulation in the last slide of the pdf above). It can be complex in terms of derivative calculations. 
- Prepare tutorial examples with these newly added tests


frankfeifan:
An integrated test is added to verify the implementation: https://github.com/GEOS-DEV/GEOS/blob/deaa1bda781b98215e21a7671b55756f434a5604/inputFiles/thermoPoromechanics/ThermoDruckerPrager_1DCooling_smoke.xml

The previous GEOS total stress calculation cannot match the analytical solution:
[displacementPlots.pdf](https://github.com/user-attachments/files/23267666/displacementPlots.pdf)
[stressPlots.pdf](https://github.com/user-attachments/files/23267667/stressPlots.pdf)


With this fix, the analytical solution can be matched:
[displacementPlots.pdf](https://github.com/user-attachments/files/23267671/displacementPlots.pdf)
[stressPlots.pdf](https://github.com/user-attachments/files/23267673/stressPlots.pdf)

jhuang2601 (approved):
Great update, thanks!

herve-gross (approved):

PR #3634, opened 2025-04-24
feat: Single-phase reactive transport combined with HPCReact

frankfeifan:
This branch aims to implement the single-phase reactive transport model into GEOS. The following tasks are listed for completion:

- [x] A `SinglePhaseReactiveTransport` solver (derived from `SinglePhaseBase`) that contains all needed functions to solve the problem and `SinglePhaseReactiveTransportFields.hpp` to store all necessary fields 
- [x] `AccumulationKernel` and `FluxComputeKernel` to assemble the local rhs and matrix
- [x] Make everything work for both isothermal and thermal
- [x] Plug in the chemistry module from this [repo](https://github.com/GEOS-DEV/HPCReact) through a `ReactiveSinglePhaseFluid` constitutive model 
- [x] A unit test `testSinglePhaseReactiveTransport` to verify the implementation of Jacobian matrix
- [x] Examples for representative reaction systems
- [x] A porosity model that updates the solid porosity with chemical reaction 
- [x] Update `surfaceArea` (added but need test)
- [x] MoMaS verification (transport + all equilibrium) and generic chain reaction verification (transport + all kinetic)

PR #3887, opened 2025-10-29
feat: Sequential ChemoMechanics Coupling

frankfeifan:
This PR will couple `SinglePhaseReactiveTransport` solver (https://github.com/GEOS-DEV/GEOS/pull/3634) with `SolidMechanicsFEM` in a sequential manner first. The preliminary goal is to account for the volume expansion or contraction and its resulting mechanical response (e.g., stress) due to chemical precipitation or dissolution. 

- [x] Incorporate an inelastic strain term based on mineral reaction in the stress calculation in a `CoupledSolid`
- [ ] Some integrated tests or unit tests if necessary
- [x] An appropriate porosity and permeability model 
- [x] Couple with the phase-field for damage/fracture modeling (maybe another PR)


src/coreComponents/constitutive/permeability/CarmanKozenyPermeability.hpp
```diff
@@ -151,9 +151,11 @@ void CarmanKozenyPermeabilityUpdate::compute( real64 const & porosity,
 {
   real64 const constant = pow( m_sphericity*m_particleDiameter, 2 ) / 150;
 
-  real64 const permValue = constant * pow( porosity, 3 )/ pow( (1 - porosity), 2 );
+  real64 const eps = 1e-20;
```
frankfeifan, line 154: @azibitsker lets make it an input option later instead of hard code

PR #3891, opened 2025-10-30
feat: add a warning when direct linear solver is used for > 100k dofs system

paveltomin:
add a warning when direct linear solver is used for > 100k dofs system


victorapm (approved):
Great! Just an idea - add at the end: "Consider using an iterative solver for better performance"

castelletto1:
It also depends whether the serial (`directParallel=0`) vs the parallel (`directParallel=1`, default) is used. In some cases, having a direct solver for problem sizes larger than 100k may still be useful, in particular if the system matrix is constant and the factorization can be reduced.

castelletto1 (approved):
Regardless of my earlier comment, I agree with having the WARNING report the information.

OmarDuran (approved):
Thanks, Pavel. This is an useful addition

PR #3892, opened 2025-10-30
fix: Add Clang warning flag to suppress 64-to-32 bit conversion warnings

OmarDuran:
Adds `-Wno-shorten-64-to-32` to Clang warning flags to suppress noisy 64-to-32 bit conversion warnings while keeping other strict checks enabled.

This PR is connected with issue #3893


rrsettgast:
We shouldn't do this. We should fix the problem instead. Can you output an example of the warnings? I don't see them on my Mac. If the operations is OK, i.e. our integer_conversion routines, then we can isolate the warning suppression with pragmas

OmarDuran:
I am closing this PR due to current resource constraints.

PR #3900, opened 2025-11-04
refactor: Enforce arrayRef regex & Handling and XML Spacing Rules

arng40:
To reference an array of groups aka `groupNameRefArray`, we ensure we have to write `attribute="{*}"` .
We cannot longer write for example :  `attribute="*"` 

Also allow to have spaces spaces before and after a value in an attribute :
```
someAttributes="{ 0.0, 1.0, 10.0, 100.0 }"
coordinates="   { 0.0, 1.0, 10.0, 100.0 }"
```


herve-gross (approved):

src/coreComponents/dataRepository/unitTests/testXmlWrapper.cpp
```diff
@@ -375,11 +372,12 @@ TEST( testXmlWrapper, testGroupNamesFormats )
...
       GroupNameTest( groupNameRegex, "test-name" ),
       GroupNameTest( groupNameRegex, "test.name" ),
+      GroupNameTest( groupNameRegex, "  test.name" ),
+      GroupNameTest( groupNameRegex, "  test.name   " ),
     };
     for( GroupNameTest const & input : workingInputs )
     {
       EXPECT_NO_THROW( xmlWrapper::stringToInputVariable( groupName, input.m_valueToTest, input.m_regex ) );
-      EXPECT_STREQ( input.m_valueToTest.c_str(), groupName.c_str() );
     }
```
MelReyCG: Why removing that assertion?

src/coreComponents/dataRepository/unitTests/testXmlWrapper.cpp
```diff
@@ -301,7 +300,6 @@ INSTANTIATE_TEST_SUITE_P(
                      std::make_tuple( "1.234gamma", 0, true ),
                      std::make_tuple( "1.2.3", 0, true ),
                      std::make_tuple( "1e2.3 ", 0, true ),
-                     std::make_tuple( "1 ", 0, true ),
                      std::make_tuple( "1e", 0, true ),
```
MelReyCG, line 309: Why removing that case instead of adding one to test your new feature? (space before)

src/coreComponents/dataRepository/xmlWrapper.hpp
```diff
@@ -354,7 +354,7 @@ stringToInputVariable( T & target, string const & value, Regex const & regex )
 
   std::istringstream ss( value );
   ss >> target;
-  GEOS_THROW_IF( ss.fail() || !ss.eof(),
+  GEOS_THROW_IF( ss.fail(),
```
MelReyCG: if eof reaching is not controled, I'm afraid that text context can be ignored.
Is there a reason to remove that?

src/coreComponents/codingUtilities/RTTypes.cpp
```diff
@@ -220,7 +223,7 @@ rtTypes::RegexMapType rtTypes::createBasicTypesRegexMap()
     { string( CustomTypes::plotLevel ), Regex( intRegex, intDesc ) },
     { string( CustomTypes::groupName ), Regex( groupNameRegex, groupNameDesc ) },
     { string( CustomTypes::groupNameRef ), Regex( groupNameRefRegex, groupNameRefDesc ) },
-    { string( CustomTypes::groupNameRefArray ), constructArrayRegex( groupNameRefRegex, groupNameRefDesc, 1 ) }
+    { string( CustomTypes::groupNameRefArray ), Regex( groupNameRefArrayRegex, groupNameRefArrayDesc ) }
```
MelReyCG, line 226: As you no longer use ``constructArrayRegex()``, this array format is under-tested.
Are you sure things like ``"{a,b,,d}"`` is prohibed? I think a lot of code consider that array entries are always non-null (null options are for ``groupNameRef``).

MelReyCG (changes requested):
need a few more work on unit tests

src/coreComponents/dataRepository/unitTests/testXmlWrapper.cpp
arng40: revert

src/coreComponents/dataRepository/xmlWrapper.hpp
arng40: revert and add << std::ws

src/coreComponents/dataRepository/unitTests/testXmlWrapper.cpp
arng40: Since the regex has been updated this case no longer works

MelReyCG: Since your last commit, it is not prohibed nor accepted.
Take clear decision on "1  " and "  1".

src/coreComponents/dataRepository/xmlWrapper.hpp
```diff
@@ -350,9 +350,10 @@ template< typename T >
 std::enable_if_t< traits::CanStreamInto< std::istringstream, T > >
 stringToInputVariable( T & target, string const & value, Regex const & regex )
 {
-  validateString( value, regex );
+  string_view stringTrimed = stringutilities::trimSpaces( value );
+  validateString( string( stringTrimed ), regex );
```
MelReyCG: I think you're getting issues in your unit tests because of the ordering of your validations:
```suggestion
  validateString( value, regex );
  string_view stringTrimed = stringutilities::trimSpaces( value );
```

src/coreComponents/dataRepository/unitTests/testXmlWrapper.cpp
```diff
@@ -407,6 +407,92 @@ TEST( testXmlWrapper, testGroupNamesFormats )
...
+      GroupNameTest( groupNameRefArrayRegex, "{}" ),
+      GroupNameTest( groupNameRefArrayRegex, " {} " ),
+      GroupNameTest( groupNameRefArrayRegex, " \t {} \n " ),
+      GroupNameTest( groupNameRefArrayRegex, "{groupName}" ),
+      GroupNameTest( groupNameRefArrayRegex, "{123name}" ),
+      GroupNameTest( groupNameRefArrayRegex, "{name123}" ),
+      GroupNameTest( groupNameRefArrayRegex, "{a-Z0-9./*[]-_,}" ),
+      GroupNameTest( groupNameRefArrayRegex, "{name.with-special_chars}" ),
+      GroupNameTest( groupNameRefArrayRegex, "{path/to/resource*}" ),
+      GroupNameTest( groupNameRefArrayRegex, "{[arrayElement]}" ),
```
MelReyCG, line 444: Can you add these ones in non-array ``groupName`` test please?

src/coreComponents/dataRepository/unitTests/testXmlWrapper.cpp
```diff
@@ -407,6 +407,92 @@ TEST( testXmlWrapper, testGroupNamesFormats )
...
+      GroupNameTest( groupNameRefArrayRegex, "{groupName}} " ),
+      GroupNameTest( groupNameRefArrayRegex, "{}groupName" ),
+      GroupNameTest( groupNameRefArrayRegex, "groupName{}" ),
+      GroupNameTest( groupNameRefArrayRegex, "test{groupName}" ),
+      GroupNameTest( groupNameRefArrayRegex, "{groupName}test" ),
+      GroupNameTest( groupNameRefArrayRegex, "{groupName} test" ),
+      GroupNameTest( groupNameRefArrayRegex, "{element with space, another}" ),
+      GroupNameTest( groupNameRefArrayRegex, "{element, element with space, 123.456, a-b}" ),
+      GroupNameTest( groupNameRefArrayRegex, "{ space at ends } " ),
+      GroupNameTest( groupNameRefArrayRegex, "{valuewith,,commas }" ),
```
MelReyCG, line 491: Add one with all alowed whitespaces kind in-between the commas (``{hello, \t\n\r ,world}``)

src/coreComponents/dataRepository/unitTests/testXmlWrapper.cpp
```diff
@@ -375,11 +372,16 @@ TEST( testXmlWrapper, testGroupNamesFormats )
...
+      GroupNameTest( groupNameRegex, "  test.name   " ),
     };
     for( GroupNameTest const & input : workingInputs )
     {
       EXPECT_NO_THROW( xmlWrapper::stringToInputVariable( groupName, input.m_valueToTest, input.m_regex ) );
-      EXPECT_STREQ( input.m_valueToTest.c_str(), groupName.c_str() );
+      std::string groupNameComp;
+      std::istringstream ss( input.m_valueToTest );
+      ss >> groupNameComp >> std::ws;
+      EXPECT_STREQ( groupNameComp.c_str(), groupName.c_str() );
```
MelReyCG: please revert

MelReyCG (changes requested):

src/coreComponents/dataRepository/unitTests/testXmlWrapper.cpp
MelReyCG: Your last fix...
```
      EXPECT_STREQ( groupName.c_str(), groupName.c_str() );
```
... is like a `a == a` check, it does not check anything.

MelReyCG (changes requested):

src/coreComponents/dataRepository/unitTests/testXmlWrapper.cpp
```diff
@@ -374,12 +380,13 @@ TEST( testXmlWrapper, testGroupNamesFormats )
...
       GroupNameTest( groupNameRegex, "test_name" ),
       GroupNameTest( groupNameRegex, "test-name" ),
-      GroupNameTest( groupNameRegex, "test.name" ),
+      GroupNameTest( groupNameRegex, " testname " ),
+      GroupNameTest( groupNameRegex, " \t testname \n " ),
     };
     for( GroupNameTest const & input : workingInputs )
     {
       EXPECT_NO_THROW( xmlWrapper::stringToInputVariable( groupName, input.m_valueToTest, input.m_regex ) );
-      EXPECT_STREQ( input.m_valueToTest.c_str(), groupName.c_str() );
```
MelReyCG, line 382: Here is the place that needs to be reworked.

MelReyCG (changes requested):

MelReyCG (changes requested):
please address last comment

src/coreComponents/codingUtilities/RTTypes.cpp
```diff
@@ -160,28 +160,31 @@ rtTypes::RegexMapType rtTypes::createBasicTypesRegexMap()
...
+  string_view const realRegex = "\\s*[+-]?[\\d]*([\\d]\\.?|\\.[\\d])[\\d]*([eE][-+]?[\\d]+|\\s*)";
 
   string_view const R1Desc = "Input value must be a R1Tensor, an array of 3 real numbers surrounded by braces and separated by commas (eg.  \"{ 1, .25, +2.3}\", \"{ -.4, 5.6e7, -8E-9\", etc.) .";
   string const R1Regex = "\\s*\\{\\s*(" + string( realRegex ) + "\\s*,\\s*){2}" + string( realRegex ) + "\\s*\\}\\s*";
   string_view const R2Desc = "Input value must be a R2SymTensor, an array of 6 real numbers surrounded by braces and separated by commas (eg.  \"{ 1, .25, +2.3, -.4, 5.6e7, -8E-9\", etc.) .";
   string const R2Regex = "\\s*\\{\\s*(" + string( realRegex ) + "\\s*,\\s*){5}" + string( realRegex ) + "\\s*\\}\\s*";
 
   string_view const strDesc = "Input value must be a string that cannot be empty, contain any whitespaces nor the characters  , { }";
-  string_view const strRegex = "[^,\\{\\}\\s]+\\s*";
+  string_view const strRegex = "\\s*[^,\\{\\}\\s]+\\s*";
```
dkachuma, line 171: Is this now out of sync with the message above? Do we now allow leading spaces?

dkachuma (approved):

src/coreComponents/codingUtilities/RTTypes.cpp
arng40: The test now authorize to have spaces to write smth like :
```
someAttributes="{ 0.0, 1.0, 10.0, 100.0 }"
coordinates="   { 0.0, 1.0, 10.0, 100.0 }"
```

MelReyCG (approved):

src/coreComponents/codingUtilities/RTTypes.cpp
MelReyCG: @dkachuma It is a proposal indeed. Do not hesitate to tell if you see any issue.

jhuang2601 (approved):

src/coreComponents/codingUtilities/RTTypes.cpp
dkachuma: I see. It makes sense now. Thanks

PR #3902, opened 2025-11-05
refactor: Centralize error output

arng40:
Remove code duplication found in  `GEOS_THROW`, `GEOS_ERROR`, `GEOS_WARNING` and put into a static function in `ErrorLogger`.
Called  while `flushErrorMsg()`.
Move all Exceptions under GeosExceptions.hpp


src/coreComponents/common/logger/ErrorHandling.hpp
```diff
@@ -119,6 +121,9 @@ class ErrorLogger
     std::vector< ErrorContext > m_contextsInfo;
     /// the stack trace
     std::vector< std::string > m_sourceCallStack;
+    /// the string stack trace
+    std::string m_stringCallStack;
```
MelReyCG: Why do you need it to be stored twice?

MelReyCG: You can just loop over the frames texts, it gives the opportunity to later format them independantly.

src/coreComponents/common/logger/ErrorHandling.hpp
```diff
@@ -230,6 +235,24 @@ class ErrorLogger
     bool m_isValidStackTrace = false;
   };
 
+  /**
+   * @brief Retrieve all informations from the ErrorLogger and format into an ascii message
+   * @param errMsg Class containing all the error/warning information
+   * @param rank The rank where the error/warning happened
+   * @return The error message displayed to the standard output
+   */
+  static string formatMsgToAscii( ErrorLogger::ErrorMsg const & errMsg, string const & rank )
```
MelReyCG: Follow the function goal: "formatX", and not "logX" (and stick to "formatX" since that is more flexible).
```suggestion
  static void formatMsgToAscii( std::ostream & outStream, ErrorLogger::ErrorMsg const & errMsg, string const & rank )
```

src/coreComponents/common/logger/Logger.hpp
```diff
@@ -156,36 +156,23 @@
...
-                                         __FILE__, \
-                                         __LINE__ ); \
-        msgStruct.setCause( __cause ); \
-        msgStruct.addCallStackInfo( stackHistory ); \
-        msgStruct.addContextInfo( GEOS_DETAIL_REST_ARGS( __VA_ARGS__ ) ); \
-        GEOS_ERROR_LOGGER_INSTANCE.flushErrorMsg( msgStruct ); \
-      } \
+      std::ostringstream msgoss; \
+      msgoss << GEOS_DETAIL_FIRST_ARG( __VA_ARGS__ ); \
+      std::ostringstream causemsgsoss; \
```
MelReyCG: Keep the original "__" prefix (and names?)
They helped to not have name shadowing.

src/coreComponents/common/unitTests/testUnits.cpp
```diff
@@ -155,3 +155,11 @@ TEST( Units, SystemDurationFormatTest )
...
   }
 }
+
+
+int main( int ac, char * av[] )
+{
+  ::testing::InitGoogleTest( &ac, av );
+  int const result = RUN_ALL_TESTS();
+  return result;
+}
```
MelReyCG, line 165: thanks !

MelReyCG (changes requested):

src/coreComponents/common/logger/Logger.hpp
```diff
@@ -230,44 +218,35 @@
...
+          .addRank( ::geos::logger::internal::g_rank ) \
+          .addCallStackInfo( LvArray::system::stackTrace( true ) ); \
       } \
-      throw GEOS_DETAIL_FIRST_ARG( __VA_ARGS__ )( __oss.str() ); \
+      GEOS_GLOBAL_LOGGER.currentErrorMsg() \
+        .addToMsg( __msgoss.str() ) \
+        .addContextInfo( GEOS_DETAIL_REST_ARGS( __VA_ARGS__ ) ); \
+      std::ostringstream __asciiErrormSG; \
+      ErrorLogger::formatMsgToAscii( GEOS_GLOBAL_LOGGER.currentErrorMsg(), \
+                                     __asciiErrormSG ); \
```
MelReyCG: Is it really useful to output the message here to the log?
The ``ErrorMsg`` get more and more info along the ``catch`` potencial cascade, then at the ``main()`` catch, it get fully flushed to the YAML. Why not doing the same for the log?

src/coreComponents/common/logger/Logger.hpp
```diff
@@ -158,34 +158,22 @@
...
+      __causemsgsoss << CAUSE_MESSAGE; \
+      ErrorLogger::ErrorMsg msgStruct( ErrorLogger::MsgType::Error, \
+                                       __msgoss.str(), \
+                                       ::geos::logger::internal::g_rank, \
+                                       __FILE__, \
+                                       __LINE__ ); \
+      msgStruct.setCause( __causemsgsoss.str() ); \
+      msgStruct.addCallStackInfo( LvArray::system::stackTrace( true ) ); \
+      msgStruct.addContextInfo( GEOS_DETAIL_REST_ARGS( __VA_ARGS__ ) ); \
+      ErrorLogger::formatMsgToAscii( msgStruct, GEOS_GLOBAL_LOGGER.getErrorStream() ); \
```
MelReyCG: I feel like ``formatMsgToAscii()`` should be done internally by ``flushErrorMsg()``. 
Is it passing the unit test on your side?
Do you have a error/throw/warning test series? (I think you should have like at least 5 tests to validate, but unit tests should do that)

src/coreComponents/common/logger/ErrorHandling.hpp
```diff
@@ -230,6 +232,30 @@ class ErrorLogger
     bool m_isValidStackTrace = false;
   };
 
+  /**
+   * @brief Retrieve all informations from the ErrorLogger and format into a stream
```
MelReyCG:
```suggestion
   * @brief Retrieve all informations from the ErrorMsg and format into a stream.
```

And **IF** it is needed to be public, you can add:
```
   *        Useful for formating a message log text externally.
```

src/coreComponents/common/logger/ErrorHandling.cpp
```diff
@@ -230,73 +230,76 @@ void ErrorLogger::streamMultilineYamlAttribute( std::string_view msg, std::ofstr
...
-                             m_filename, isOutputFileEnabled() ) );
+      yamlFile << "\n";
+      yamlFile.flush();
+      errorMsg = ErrorMsg();
+      GEOS_LOG_RANK( GEOS_FMT( "The error file {} has been appended.", m_filename ) );
+    }
+    else
+    {
+      GEOS_LOG_RANK( GEOS_FMT( "Unable to open error file for writing.\n- Error file: {}\n- Error file enabled = {}.\n",
+                               m_filename, g_errorLogger.isOutputFileEnabled() ) );
```
MelReyCG: Why?

src/coreComponents/dataRepository/unitTests/testErrorHandling.cpp
```diff
@@ -210,11 +210,12 @@ TEST( ErrorHandling, testYamlFileExceptionOutput )
...
-    "- frame0: ",
-    "- frame1: ",
-    "- frame2: "
-  } );
+        __FILE__, line1 ),
+      "- frame0: ",
+      "- frame1: ",
+      "- frame2: "
+    } );
+  }
```
MelReyCG: Syntax error.

src/main/main.cpp
```diff
@@ -76,11 +75,8 @@ int main( int argc, char *argv[] )
...
   catch( std::exception const & e )
   {
-    GEOS_LOG( e.what() );
-    if( ErrorLogger::global().isOutputFileEnabled() )
-    {
-      ErrorLogger::global().flushErrorMsg( ErrorLogger::global().currentErrorMsg() );
-    }
+    ErrorLogger::global().getErrorStream() << e.what();
+    ErrorLogger::global().flushErrorMsg( ErrorLogger::global().currentErrorMsg() );
     LvArray::system::callErrorHandler();
```
MelReyCG: I'm not sure this is the right direction (this is related with my other comments).

MelReyCG (changes requested):

src/coreComponents/common/logger/ErrorHandling.hpp
```diff
@@ -105,6 +112,8 @@ class ErrorLogger
   {
     /// the error type (Warning, Error or Exception)
     MsgType m_type = ErrorLogger::MsgType::Undefined;
+    /// The signal received and formatted
+    std::string m_signal;
```
MelReyCG: That little guy seems to never be initialized (`addSignalToMsg()`?)
Maybe you can just remove it and read the `Attribute::Signal` entry

src/coreComponents/common/logger/ErrorHandling.cpp
```diff
@@ -41,6 +41,61 @@ static constexpr std::string_view g_level3Next =  "        ";
 
 ErrorLogger g_errorLogger{};
 
+/**
+ * @brief Retrieve all informations from the ErrorMsg and format and write into a stream.
+ * @param errMsg Class containing all the error/warning information
+ * @param oss The output stream to write the content to.
+ */
+void ErrorLogger::writeToAscii( ErrorLogger::ErrorMsg const & errMsg, std::ostream & oss )
```
MelReyCG: maybe move that just over ``writeToYaml``?

src/coreComponents/common/logger/ErrorHandling.hpp
```diff
@@ -75,6 +79,9 @@ class ErrorLogger
     /// The key is a field of the Attribute enumeration and is converted to a string for writing in the YAML
     map< Attribute, std::string > m_attributes;
 
+    /// String containing the target object name followed by the the file and line declaring it.
+    string m_dataDisplayString;
```
MelReyCG: As the auto-constructor is extensively used in your PR, I would suggest moving ``m_dataDisplayString`` at top of ``m_attributes`` for clarity and to force to give this parameters.

src/coreComponents/common/logger/ErrorHandling.cpp
```diff
@@ -41,6 +41,61 @@ static constexpr std::string_view g_level3Next =  "        ";
...
+{
+  /// HEADER
+  oss << "***** " << ErrorLogger::toString( errMsg.m_type ) << "\n";
+  if( !errMsg.m_signal.empty())
+  {
+    oss<< "***** SIGNAL: "<< errMsg.m_signal <<"\n";
+  }
+  oss << "***** LOCATION: " << errMsg.m_file<< " l." << errMsg.m_line << "\n";
+  oss << "***** " << errMsg.m_cause  << "\n";
```
MelReyCG: Be careful, it seems like ErrorMsg can be constructed without those. You can output them conditionnally or remove the empty ``ErrorMsg`` constructor (there are also solutions between the two).

src/coreComponents/common/logger/ErrorHandling.hpp
```diff
@@ -230,6 +239,13 @@ class ErrorLogger
     bool m_isValidStackTrace = false;
   };
 
+  /**
+   * @brief Format all information in ErrorMsg and write it to the specified output stream
+   * @param errMsg The struct containing the error/warning object
+   * @param oss The output stream
+   */
+  static void writeToAscii( ErrorLogger::ErrorMsg const & errMsg, std::ostream & oss );
+
```
MelReyCG, line 288: Same remark, can you move it with the associated functions?
Maybe I see why you ordered it that way, yes it is static, but it is a function, so you can place it wherever you want.
If you prefer it as it is, don't hesitate to tell.

src/main/main.cpp
```diff
@@ -74,16 +73,20 @@ int main( int argc, char *argv[] )
...
+  catch( geos::Exception & e )
+  { // GEOS generated exceptions management
+    ErrorLogger::global().flushErrorMsg(  ErrorLogger::global().currentErrorMsg() );
+    basicCleanup();
+    // lvarray error handler is just program termination
     LvArray::system::callErrorHandler();
+  }
+  catch( std::exception const & e )
+  { // native exceptions management
+    ErrorLogger::ErrorMsg & errMsg = ErrorLogger::global().currentErrorMsg();
```
MelReyCG: I think it lacks a lot of information here: at this point, IIRC, ``GEOS_THOW()`` has never been called, so the same information must be provided to the error message.

Appart from that, this part is a lot better!

MelReyCG (changes requested):

src/coreComponents/common/logger/ErrorHandling.cpp
```diff
@@ -228,31 +259,121 @@ void ErrorLogger::streamMultilineYamlAttribute( std::string_view msg, std::ofstr
...
+{
+  static constexpr string_view PREFIX = "***** ";
+  // --- HEADER ---
+  oss << PREFIX << ErrorLogger::toString( errMsg.m_type ) << "\n";
+  if( !errMsg.m_file.empty())
+  {
+    oss << PREFIX<< "LOCATION: " << errMsg.m_file;
+    if( errMsg.m_line > 0 )
+    {
+      oss << " l." << errMsg.m_line;
```
MelReyCG: You changed the format here in a way that prevent us, developpers, to ctrl+click the file location for quick access.

The ``LOCATION`` macro is:
```
#define LOCATION __FILE__ ":" STRINGIZE( __LINE__ )
```

MelReyCG: Also, you put an if there but not in the YAML format.

src/coreComponents/common/logger/ErrorHandling.hpp
```diff
@@ -139,7 +164,13 @@ class ErrorLogger
...
               std::string_view msgFile,
               integer msgLine )
-      : m_type( msgType ), m_msg( msgContent ), m_ranksInfo( {rank} ), m_file( msgFile ), m_line( msgLine ) {}
+      :
+      m_type( msgType ),
+      m_msg( msgContent ),
+      m_ranksInfo( {rank} ),
+      m_file( msgFile ),
+      m_line( msgLine ),
+      m_commit( false ) {}
```
MelReyCG: -> cpp

src/coreComponents/common/logger/ErrorHandling.hpp
```diff
@@ -148,7 +179,7 @@ class ErrorLogger
      *              default is false
      * @return Reference to the current instance for method chaining.
      */
-    ErrorMsg & addToMsg( std::exception const & e, bool toEnd = false );
+    ErrorMsg & addToMsg( std::exception const & e, bool const toEnd = false );
```
MelReyCG: Those ``const`` added to boolean values are implementation details, not interface matter, it should remain in the cpp.

MelReyCG: same for ``MsgType`` and ``integer`` below

src/coreComponents/common/logger/ErrorHandling.hpp
```diff
@@ -220,15 +251,102 @@ class ErrorLogger
     bool isValidStackTrace() const
     { return m_isValidStackTrace; }
 
+    /**
+     * @return Whether the error message has been finalized and committed.
+     */
+    bool isCommited() const
+    { return m_commit; }
```
MelReyCG: About this "commit" concept:
1. Why is it needed? What problem does it aim to solve? Is the "commited" state really needed?
2. ``ErrorMsg`` is a data structure with some helper methods for building it, I don't think it should have any state in it.
3. I see your new ``ErrorMsgBuilder`` class, with a lot of copied method, with some of them having silenciously no effect. Could the methods be better distributed?

I propose ``ErrorMsg`` being a POD, with no state, and ``ErrorMsgBuilder`` having all the building methods.
Also, I would not bind ``ErrorMsgBuilder`` to an ``ErrorLogger`` instance but to the ``ErrorMsg`` instance. This way you don't need this two ways dependancies anymore, which does not exist for good reasons (duplicating ``flush()`` & violating encapsulation).

src/coreComponents/common/logger/ErrorHandling.hpp
```diff
@@ -268,11 +386,10 @@ class ErrorLogger
 
   /**
    * @brief Gives acces to the error message that is currently being constructed,
-   *        potencially at various application layers
-   *        Use flushErrorMsg() when the message is fully constructed and you want it to be output
-   * @return Reference to the current instance for method chaining.
+   *        potencially at various application layers (Typically for exceptions)
```
MelReyCG: "Typically for exceptions"
Are there other use-cases?

src/coreComponents/common/logger/ErrorHandling.hpp
```diff
@@ -220,15 +251,102 @@ class ErrorLogger
...
+    ErrorLogger & m_errorContext;
   };
+  /**
+   * @brief Creates and returns a new ErrorMsgBuilder to begin constructing an error message.
+   * @return A new ErrorMsgBuilder instance
+   */
+  ErrorMsgBuilder beginLogger()
+  {
+    return ErrorMsgBuilder( *this );
+  }
```
MelReyCG: I see that you tried to remove the "current error message" read-write accessor for encapsulation & clarity purpose, but I'm not sure the solution is better than the original: ``beginLogger()`` can be called multiple time in an exception stack scenario. I think ``buildCurrentErrorMsg()`` would be better named.

src/coreComponents/dataRepository/unitTests/testErrorHandling.cpp
```diff
@@ -193,7 +204,7 @@ TEST( ErrorHandling, testYamlFileExceptionOutput )
     rank: 0
     message: >-
       Table input error.
-      Empty Group: Base Test Class (file.xml, l.23)
+      Empty Group
     contexts:
       - priority: 2
         inputFile: /path/to/file.xml
```
MelReyCG, line 218: The main context information are lost (`Base Test Class (file.xml, l.23)`).

src/coreComponents/common/logger/ErrorHandling.hpp
```diff
@@ -220,15 +251,102 @@ class ErrorLogger
...
+    { m_errorContext.m_currentErrorMsg.commitErrorMsg(); }
+    /**
+     * @copydoc ErrorLogger::flushErrorMsg()
+     */
+    void flush()
+    { m_errorContext.flushErrorMsg(); }
+
+private:
+    ///@copydoc ErrorLogger::m_errorContext
+    ErrorLogger & m_errorContext;
```
MelReyCG: If you keep it (I don't advise to):
```suggestion
    ErrorLogger & m_errorLogger;
```

If you change it:
```suggestion
    ErrorMsg & m_errorMsg;
```

MelReyCG (changes requested):

src/coreComponents/common/logger/ErrorHandling.hpp
```diff
@@ -66,21 +69,24 @@ class ErrorLogger
...
-     *
+     * @brief Construct to initialize ErrorContext given a string containing the context and his priority
+     * @param attributes Map containing contextual information about the error
+     * @param priority Priority level assigned to an error context.
      */
-    integer m_priority = 0;
+    ErrorContext( map< Attribute, std::string >  attributes, integer priority ):
+      m_dataDisplayString( "" ),
+      m_attributes( attributes ),
+      m_priority( priority ) {};
```
MelReyCG: ``m_dataDisplayString`` is allowed as empty? Isn't that the source of your issue in the unit test?

src/coreComponents/common/initializeEnvironment.cpp
```diff
@@ -134,26 +128,16 @@ void setupLogger()
...
-      }
+      ErrorLogger::global().beginLogger()
+        .addSignalToMsg( signal )
+        .setType( ErrorLogger::MsgType::Error )
+        .addRank( ::geos::logger::internal::g_rank )
+        .addCallStackInfo( stackHistory )
+        .addContextInfo(
+        ErrorContext{ { { ErrorContext::Attribute::Signal, std::to_string( signal ) } }, 1 },
+        ErrorContext{ { { ErrorContext::Attribute::DetectionLoc, string( "signal handler" ) } }, 0 } )
+        .flush();
```
MelReyCG: Maybe a unit test for signals would be a good thing.

MelReyCG (changes requested):

src/coreComponents/common/logger/ErrorHandling.hpp
MelReyCG: live discussions:
1. the problem is the `if`, written in the throw macro, which break this scenario:
```
  try {
    GEOS_THROW_IF( parseFloat(x), "float invalid", geos::RuntimeError );
  } catch ( geos::exception const & e )
  {
    GEOS_LOG("aie");
  }

  GEOS_THROW_IF( MpiWrapper::commRank() > 0, "RelpermDriver should only be run in serial", geos::RuntimeError );
```
We agree that we have to remove this `if`, but the `isCommited()` solution is not optimal, we could just remove the if.

src/coreComponents/common/logger/ErrorHandling.hpp
```diff
@@ -288,12 +405,32 @@ class ErrorLogger
...
    */
-  void flushErrorMsg( ErrorMsg & errorMsg );
+  void writeToYaml();
+
+  /**
+   * @brief Write all the information retrieved about the error/warning message into the output stream specified and
+   * optionnaly into a yaml file
+   * @param errorMsg a constant reference to the ErrorMsg
+   */
+  void flushErrorMsg();
```
MelReyCG:
```suggestion
  /**
   * @brief Write all the information retrieved about the error/warning message into the YAML file
   * @param errorMsg a constant reference to the error
   */
  void writeToYaml( ErrorLogger::ErrorMsg const & errMsg );

  /**
   * @brief Write all the information retrieved about the error/warning message into the instance
   * outputs (stream specified + optional yaml file)
   * @param errorMsg a reference to the ErrorMsg to output, and will be re-initialized
   */
  void flushErrorMsg( ErrorLogger::ErrorMsg & errMsg );
```

src/coreComponents/common/logger/ErrorHandling.hpp
MelReyCG: We can also rename the `currentErrorMsg` concept to `currentExceptionMsg`
-> `ErrorMsgBuilder buildCurrentExceptionMsg()` / `ErrorMsg const & getCurrentExceptionMsg()`

src/coreComponents/dataRepository/unitTests/testErrorHandling.cpp
```diff
@@ -264,6 +280,85 @@ TEST( ErrorHandling, testYamlFileErrorOutput )
...
+      "***** {}\n"
+      "***** Rank 0\n"
+      "***** Message from {}:\n"
+      "{}\n"
+      "***** Additional contexts:\n"
+      "***** - {}\n"
+      "***** - {}\n",
+      testErrorLogger.getCurrentExceptionMsg().m_file, line1,
+      testErrorLogger.getCurrentExceptionMsg().m_cause,
+      context.toString(),
```
MelReyCG: Shouldn't you switch with the most important context here?

src/coreComponents/common/logger/GeosExceptions.hpp
```diff
@@ -0,0 +1,250 @@
...
+      m_cachedWhat = "GEOS Exception (formatting failed)";
+    }
+
+  }
+
+private:
+  /// Formatted exception message for what() method
+  string m_cachedWhat;
+  /// Error message logger for structured error reporting
+  DiagnosticMsg m_errorMsg;
```
MelReyCG: Why?

MelReyCG: isn't that in ``currentExceptionMsg()``?

src/coreComponents/common/logger/GeosExceptions.hpp
```diff
@@ -0,0 +1,250 @@
...
+      oss << "***** " << cause << "\n";
+      oss << "***** Rank  " << rank << ": "<< msg <<"\n\n";
+      oss << stackTrace;
+      m_cachedWhat = oss.str();
+    } catch( ... )
+    {
+      m_cachedWhat = "GEOS Exception (formatting failed)";
+    }
+
+  }
```
MelReyCG:
```suggestion
private:
  static thread_local std::ostringstream formattingOSS;

public:
  void prepareWhat( ErrorMsg const & msg ) noexcept
  {
    formattingOSS.str("");
    formattingOSS.clear();

    ErrorLogger::writeToAscii( msg, formattingOSS );
    m_cachedWhat = formattingOSS.bad ? "Exception formatting error!" : formattingOSS.str();
    m_errorMsg = msg;
  }
```

src/coreComponents/dataRepository/unitTests/testErrorHandling.cpp
```diff
@@ -264,6 +280,85 @@ TEST( ErrorHandling, testYamlFileErrorOutput )
...
+  }
+  catch( std::exception & e )
+  {
+
+    testErrorLogger.initCurrentExceptionMessage()
+      .setType( MsgType::Exception )
+      .addToMsg( e.what() )
+      .addRank( ::geos::logger::internal::g_rank )
+      .addCallStackInfo( LvArray::system::stackTrace( true ) )
+      .get();
```
MelReyCG: who get what?

src/coreComponents/common/logger/ErrorHandling.hpp
```diff
@@ -21,214 +21,228 @@
...
-    void addContextInfoImpl( ErrorContext && ctxInfo );
+  /**
+   * @brief Add contextual information about the error/warning
+   * @param ctxInfo rvalue of the ErrorContext class
+   */
+  DiagnosticMsgBuilder & addContextInfoImpl( ErrorContext && ctxInfo );
+  /// The diagnosticMsg being constructed
+  DiagnosticMsg m_errorMsg;
+  /// The target diagnosticMsg
+  DiagnosticMsg * m_targetErrorMsg = nullptr;
```
MelReyCG:
```suggestion
  /// The target diagnosticMsg
  DiagnosticMsg & m_targetErrorMsg;
```

src/coreComponents/common/logger/ErrorHandling.hpp
```diff
@@ -240,10 +254,10 @@ class ErrorLogger
   GEOS_HOST static ErrorLogger & global();
```
MelReyCG, line 296: I see your point, but just use thread_local here

src/coreComponents/common/logger/ErrorHandling.hpp
MelReyCG: Here is what is needed : `ErrorContext( string formattedContext, map< Attribute, std::string >  attributes, integer priority ):`

src/coreComponents/common/logger/ErrorHandling.hpp
```diff
@@ -267,41 +287,75 @@ class ErrorLogger
...
   /// Indicate whether the write to YAML command line option is enabled
   bool m_writeYaml = false;
   /// YAML file name
   std::string_view m_filename = "errors.yaml";
+  /// The stream used for the log output. By default used std::cout
+  std::ostream & m_stream = std::cout;
+
+  /**
+   * @brief Write all the information retrieved about the error/warning message into the YAML file
+   * @param errorMsg a constant reference to the error
```
MelReyCG: not const

src/coreComponents/common/logger/ErrorHandling.hpp
```diff
@@ -267,41 +287,75 @@ class ErrorLogger
...
+   */
+  void flushCurrentExceptionMessage();
+
+  /**
+   * @brief Write all the information retrieved about the diagnostic message into the instance
+   * outputs (stream specified + optional yaml file)
+   * @param errorMsg a reference to the ErrorMsg to output, and will be re-initialized
+   * @note Used for warnings and non-exception errors
+   */
+  void flushErrorMsg( DiagnosticMsg & errMsg );
```
MelReyCG, line 381: provide a oss parameter (with a default to ``std::cout``)?

src/coreComponents/common/logger/ErrorHandling.hpp
MelReyCG: (removed in favor of a ErrorContext instance)

src/coreComponents/common/logger/ErrorHandling.hpp
```diff
@@ -21,214 +21,228 @@
...
+};
+
+/**
+ * @brief Builder class for constructing DiagnosticMsg  objects
+ */
+class DiagnosticMsgBuilder
+{
+public:
+  static DiagnosticMsgBuilder init()
+  { return DiagnosticMsgBuilder();}
```
MelReyCG:
```
  static ErrorMsgBuilder init( ErrorMsg & msg,
                               MsgType msgType,
                               std::string_view msgContent,
                               integer rank,
                               std::string_view msgFile,
                               integer msgLine );
```
What do you think of this suggestion? ^
Is that too constraining?

(internally, that would call the builder methods)

MelReyCG (changes requested):

src/coreComponents/common/logger/GeosExceptions.hpp
arng40: removed

src/coreComponents/dataRepository/unitTests/testErrorHandling.cpp
arng40: renamed

src/coreComponents/common/logger/ErrorHandling.hpp
arng40: ok for init with parameters but I removed file & line. For the signal we don't have file & line

src/coreComponents/common/logger/ErrorHandling.hpp
arng40: We have now 2 flush methods

src/coreComponents/common/logger/GeosExceptions.hpp
```diff
@@ -60,41 +54,21 @@ struct Exception : public std::exception
...
+    ErrorLogger::writeToAscii( msg, m_formattingOSS );  
+    m_cachedWhat = m_formattingOSS.bad() ? "Exception formatting error!" : m_formattingOSS.str();  
   }
 
 private:
   /// Formatted exception message for what() method
   string m_cachedWhat;
-  /// Error message logger for structured error reporting
-  DiagnosticMsg m_errorMsg;
+  inline static thread_local std::ostringstream m_formattingOSS;
```
MelReyCG: add a cpp for it?
(and also for `insertExMsg()` and `InputError` & `SimulationError` constructors)

src/coreComponents/common/logger/ErrorHandling.hpp
MelReyCG: Sorry, my advise seems wrong as `std::ostream & m_stream = std::cout;` exists.

src/coreComponents/common/logger/Logger.hpp
```diff
@@ -224,18 +224,16 @@
...
       __causemsgsoss << CAUSE_MESSAGE; \
-      GEOS_GLOBAL_LOGGER.initCurrentExceptionMessage() \
-        .setType( MsgType::Exception ) \
-        .setCodeLocation( __FILE__, __LINE__ ) \
-        .setCause( __causemsgsoss.str() ) \
-        .addRank( ::geos::logger::internal::g_rank ) \
-        .addCallStackInfo( LvArray::system::stackTrace( true ) ) \
-        .addToMsg( __msgoss.str() ) \
-        .addContextInfo( GEOS_DETAIL_REST_ARGS( __VA_ARGS__ )).get(); \
+      DiagnosticMsg exceptionMsg =  GEOS_GLOBAL_LOGGER.initCurrentExceptionMessage() \
```
MelReyCG: wouldn't `initCurrentExceptionMessage()` need 
```
                               std::string_view msgContent,
                               integer rank,
                               std::string_view msgFile,
                               integer msgLine
```
?

src/main/main.cpp
```diff
@@ -82,10 +83,10 @@ int main( int argc, char *argv[] )
   }
   catch( std::exception const & e )
   { // native exceptions management
-    ErrorLogger::global().flushErrorMsg( DiagnosticMsgBuilder::init()
-                                           .setType( MsgType::Exception )
-                                           .addToMsg( e.what() )
-                                           .addRank( ::geos::logger::internal::g_rank )
+    DiagnosticMsg diagnosticMsg;
+    ErrorLogger::global().flushErrorMsg( DiagnosticMsgBuilder::init( diagnosticMsg,
```
MelReyCG: not `initCurrentExceptionMessage()`?

src/coreComponents/common/logger/Logger.hpp
```diff
@@ -164,7 +164,8 @@
         .addRank( ::geos::logger::internal::g_rank ) \
         .addCallStackInfo( LvArray::system::stackTrace( true ) ) \
         .addToMsg( __msgoss.str() ) \
-        .addContextInfo( GEOS_DETAIL_REST_ARGS( __VA_ARGS__ )).get(); \
+        .addContextInfo( GEOS_DETAIL_REST_ARGS( __VA_ARGS__ )) \
+        .getDiagnosticMsg(); \
```
MelReyCG: There are a lot of unnecessary `getDiagnosticMsg()` in all usages.

src/coreComponents/common/initializeEnvironment.cpp
```diff
@@ -134,26 +127,22 @@ void setupLogger()
...
+      ErrorLogger::global().flushErrorMsg( DiagnosticMsgBuilder::init( diagnosticMsg,
+                                                                       MsgType::Error, "Signal: ",
+                                                                       ::geos::logger::internal::g_rank )
+                                             .addSignalToMsg( signal )
+                                             .addCallStackInfo( stackHistory )
+                                             .addContextInfo(
+                                             ErrorContext{"",
+                                                          { { ErrorContext::Attribute::Signal,
+                                                            std::to_string( signal ) } },
+                                                          1 },
```
MelReyCG: we could do that automatically in `addSignalToMsg()`, what do you think? (it would then be named `addSignal()`)

src/coreComponents/common/logger/ErrorHandling.hpp
MelReyCG: Sorry, my advise seems wrong as `std::ostream & m_stream = std::cout;` exists.

src/coreComponents/common/logger/ErrorHandling.hpp
MelReyCG: Also I was wondering, what about the `thread_local` instances initialization?

src/coreComponents/dataRepository/unitTests/testErrorHandling.cpp
MelReyCG: I still have the same remaks

src/coreComponents/dataRepository/unitTests/testErrorHandling.cpp
```diff
@@ -264,6 +282,87 @@ TEST( ErrorHandling, testYamlFileErrorOutput )
...
 }
 
+
+TEST( ErrorHandling, testLogFileExceptionOutput )
+{
+  ErrorLogger testErrorLogger;
+
+  size_t line1;
+
+  try
```
MelReyCG, line 295: Maybe add a comment for the test intentions:
```suggestion
  // test that the first context with default priority comes after the most important one, but before
  // a second context with default priotity
  try
```

src/coreComponents/common/logger/ErrorHandling.hpp
MelReyCG: An approach would be to remove this `thread_local` and add mutexes in output functions ([see](https://en.cppreference.com/w/cpp/thread/mutex.html#:~:text=std%3A%3Alock_guard%3Cstd%3A%3Amutex%3E))

src/coreComponents/common/logger/Logger.hpp
MelReyCG: still 2 left

src/coreComponents/dataRepository/unitTests/testErrorHandling.cpp
```diff
@@ -305,6 +406,57 @@ TEST( ErrorHandling, testYamlFileAssertOutput )
...
+  {
+  #if defined( GEOS_USE_MPI )
+    int mpi = 0;
+    MPI_Initialized( &mpi );
+    if( mpi )
+    {
+      MPI_Abort( MPI_COMM_WORLD, EXIT_FAILURE );
+    }
+  #endif
+    std::abort();
```
MelReyCG: maybe just re-call ``setupLogger()`` to reset GEOS

src/coreComponents/common/logger/GeosExceptions.hpp
MelReyCG: Don't forget to relocate the functions

src/coreComponents/common/logger/ErrorHandling.hpp
```diff
@@ -267,41 +328,82 @@ class ErrorLogger
...
-  ErrorMsg m_currentErrorMsg;
+  DiagnosticMsg m_getCurrentExceptionMsg;
   /// Indicate whether the write to YAML command line option is enabled
   bool m_writeYaml = false;
   /// YAML file name
   std::string_view m_filename = "errors.yaml";
+  /// The stream used for the log output. By default used std::cout
+  std::ostream & m_stream = std::cout;
+  /// Avoid concurrent access between threads for outputs
+  static std::mutex m_errorHandlerMutex;
```
MelReyCG: I would like to advise a less prohibiting mutex:
- non static to not block between logger instances,
- 1 mutex for each concurrent stream scenario (1 for yaml, 1 for log).

MelReyCG (approved):

src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp
tjb-ltk, line 1: As a developer, when is the modifyCurrentExceptionMessage needed ? what's the use case..

arng40: In geos there are some rethrow cases where we modify the current exception to add more context i.e
**CO2BrineFluid l.190**

herve-gross (approved):

dkachuma (approved):

tjb-ltk (approved):

acitrain (approved):

jhuang2601 (approved):

OmarDuran (approved):

victorapm (approved):

PR #3919, opened 2025-12-06
Add DeprecatedGroup and ObsoleteGroup implementations with unit tests

wrtobin:
Just putting this up for review for the moment, not married to any design decisions and I assume we'll want to change output message structure.

This basically just introduces mechanisms to lifecycle internal types that have been replaced/supplanted while informing the user in obvious ways and maintaining the schema generation/documentation and parse>init process for that type until it drops completely from the project.

To deprecate any `Group` subclass just, for e.g. `class Foo : public ExecutableGroup` (inherits from any group subclass):

```cpp
class Foo : public DeprecatedGroup< ExecutableGroup >
```

It basically just injects itself into the inheritance hierarchy and adds the deprecation/obsolescence behavior.

### DIFF
--- a/src/coreComponents/dataRepository/DeprecatedGroup.hpp
+++ b/src/coreComponents/dataRepository/DeprecatedGroup.hpp
@@ -1,0 +1,173 @@
+/*
+ * ------------------------------------------------------------------------------------------------------------
+ * SPDX-License-Identifier: LGPL-2.1-only
+ *
+ * Copyright (c) 2016-2024 Lawrence Livermore National Security LLC
+ * Copyright (c) 2018-2024 TotalEnergies
+ * Copyright (c) 2018-2024 The Board of Trustees of the Leland Stanford Junior University
+ * Copyright (c) 2023-2024 Chevron
+ * Copyright (c) 2019-     GEOS/GEOSX Contributors
+ * All rights reserved
+ *
+ * See top level LICENSE, COPYRIGHT, CONTRIBUTORS, NOTICE, and ACKNOWLEDGEMENTS files for details.
+ * ------------------------------------------------------------------------------------------------------------
+ */
+
+/**
+ * @file DeprecatedGroup.hpp
+ *
+ * This file provides utilities for marking Group classes as deprecated while maintaining
+ * full functionality.
+ *
+ * @section deprecated_lifecycle Deprecated Stage in Group Lifecycle
+ *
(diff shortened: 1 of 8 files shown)


src/coreComponents/dataRepository/DeprecatedGroup.hpp
```diff
@@ -0,0 +1,173 @@
...
+    string const message = getDeprecationMessage();
+
+    GEOS_LOG_RANK_0( GEOS_FMT(
+                       "\n"
+                       "********************************************************************************\n"
+                       "* DEPRECATION WARNING\n"
+                       "* Group: {}\n"
+                       "* {}\n"
+                       "********************************************************************************",
+                       className, message ) );
```
MelReyCG, line 153: Why not using a true ``GEOS_WARNING``?

src/coreComponents/dataRepository/ObsoleteGroup.hpp
```diff
@@ -0,0 +1,495 @@
...
+    GEOS_LOG_RANK_0( GEOS_FMT(
+                       "\n"
+                       "********************************************************************************\n"
+                       "* OBSOLETE GROUP WARNING\n"
+                       "* Group: {}\n"
+                       "* {}\n"
+                       "* This group has been removed and will cause an error if used.\n"
+                       "* (Group object created for input parsing compatibility only)\n"
+                       "********************************************************************************",
+                       className, message ) );
```
MelReyCG, line 451: Same remark, why not ``GEOS_WARNING``?

src/coreComponents/dataRepository/ObsoleteGroup.hpp
```diff
@@ -0,0 +1,495 @@
...
+ *
+ * Any code path that attempts to treat an obsolete group's wrappers as real
+ * storage (e.g. calling Wrapper<T>::reference(), resize to hold state, etc.)
+ * is a programming error in the caller. Tests intentionally restrict themselves
+ * to WrapperBase-level queries when interacting with NoopWrapper.
+ *
+ * @tparam T The type that would normally be wrapped
+ */
+template< typename T >
+class NoopWrapper final : public WrapperBase
```
MelReyCG, line 78: As metadata function as another wrapper, could it be used to document some aspects of a given Group?
If yes, it would be cool to have this class in its own file :)
If not, maybe it should be named after its final role (i.e. ``DeprecatedWrapper``)

Issue #3826, opened 2025-09-20
[AUTO] Test Python Package Integration - PR #121


rrsettgast:
Automated integration test request from geosPythonPackages CI  **Python Package Details:** - Repository: https://github.com/GEOS-DEV/geosPythonPackages.git - Branch: benedicto/ci/testGeosIntegration - PR Number: 121 - Requested by: alexbenedicto  **Test Instructions:** This issue was automatically created to trigger GEOS integration testing. The workflow should process this request and close the issue when complete.

PR #3928, opened 2025-12-22
fix: single phase statistics CSV generation

arng40:
Missing "," in csv headers and average/min pressure have been reversed


paveltomin (approved):

herve-gross (approved):

MelReyCG (changes requested):
Uniformize with the `compositionalMultiphaseStatistics.cpp`:
- use a TableLayout in the rows output
- Also, if it is not too much out of scope, you can keep the formatter as a private member (preventing any different parameter count in the CSV header & data)

MelReyCG (approved):
You can ignore my previous comment as I am doing it in #3826

tjb-ltk (approved):

dkachuma (approved):

PR #3938, opened 2026-01-20
feat: Configurable Quadrature Rules for triangles/quads in ALM solver

bd713:
This PR:

- Removes hard-coded quadrature order for contact interface elements (triangles and quadrilaterals) in `SolidMechanicsAugmentedLagrangianContact`
- Relies on `useHighOrderQuadratureRule` setting from `FiniteElementDiscretization`, allowing:
  - Consistent quadrature accuracy between 3D and 2D elements
  - Users to control quadrature accuracy via XML input


paveltomin (approved):

jafranc (approved):

jhuang2601 (approved):
Many thanks for fixing it!

castelletto1:
Do we want the quadrature rule to be exposed to the user?

https://github.com/GEOS-DEV/GEOS/blob/58178e42765c158f7a2f941eea098078b6f4c14a/src/coreComponents/finiteElement/FiniteElementDiscretization.cpp#L53C1-L56C91

As far as I can tell having `m_useHighOrderQuadratureRule == 1` is needed to correctly integrate bubble contributions on tetrahedra and triangular faces when the ALM solver (`SolidMechanicsAugmentedLagrangianContact`) is used. Same thing applies to `SolidMechanicsLagrangeContactBubbleStab`, which was implemented by @CusiniM for the induced seismicity project and which would need some refactoring to include `slip` and `open` fracture states. 

In my opinion, having an optional attribute, namely `useHighOrderQuadrature`, needed in the two solvers above, which must be set to `1` (for tetrahedral element subregions) is not a good design. I would not expose the option to the user and set it to true in the solver constructor.

src/coreComponents/physicsSolvers/solidMechanics/contact/SolidMechanicsLagrangeContactBubbleStab.cpp
```diff
@@ -99,6 +99,42 @@ void SolidMechanicsLagrangeContactBubbleStab::registerDataOnMesh( Group & meshBo
...
+          hasTetrahedra = true;
+        }
+      } );
+    } );
+  } );
+
+  GEOS_ERROR_IF( hasTetrahedra && useHighOrderQuadrature != 1,
+                 GEOS_FMT( "{}: Tetrahedral meshes require useHighOrderQuadratureRule=\"1\" for correct integration of bubble contributions. "
+                           "Please add this attribute to your FiniteElements/{} XML block.",
+                           getName(), getDiscretizationName() ) );
```
castelletto1: 👍

castelletto1 (approved):

PR #3940, opened 2026-01-22
fix: Fix transmissibility calculation

dkachuma:
This PR fixes the transimissibility calculated between a cell and a surface element. This follows the same treatment as `CellElementStencilTPFA`.

As a side, the logging on `vtkExtractEdges` results in some info messages in the log. I am disabling the logging temporarily while this filter is running.


castelletto1 (approved):

paveltomin (approved):

OmarDuran (approved):
Thank you for the fix @dkachuma

jhuang2601 (approved):
Great, it is working for SEAM case. 
The negative transmissibility is related to the old issue with vtk (inconsistent surface normal with MPI)

src/coreComponents/finiteVolume/FaceElementToCellStencil.hpp
jafranc: Shouldn't the `m_transMultiplier` be there too ?

dkachuma: Yes this was missing in the derivatives. I have added it. I have also included the fix in the method overload

jhuang2601:
Expected failures for MPI runs and hence rebaseline
```
=======================
Test failures
=======================
ALM_multiphasePoromechanics_curvedFrac_smoke_02
ALM_singlephasePoromechanics_curvedFrac_smoke_02
multiphasePoromechanics_FaultModel_smoke_04
multiphasePoromechanics_FaultModel_well_fim_smoke_04
multiphasePoromechanics_FaultModel_well_seq_smoke_04
singlePhasePoromechanics_FaultModel_smoke_04
```

PR #3941, opened 2026-01-22
fix: Fix thermal conductivity

dkachuma:
Somewhere along the line the proper resizing of the rock thermal conductivity for the volume weighted model was lost. This was resulting in a seg fault in models using this.


paveltomin (approved):

jhuang2601 (approved):
Nice catch

jafranc (approved):

PR #3954, opened 2026-01-30
fix: Fix bug in L2 norm calculation for isothermal

dkachuma:
For isothermal models, the normaliser for the L2 normal calculation was not being populated leading to a divide by zero down the line. This PR simply sets this to the minimum value to bypass the floating point exception.


paveltomin (approved):

src/coreComponents/physicsSolvers/fluidFlow/CompositionalMultiphaseFVM.cpp
tjb-ltk, line 1: Should defaults in well also be set ?

dkachuma: Well looks to be handled differently. There's no volume balance and it seems it's all sized correctly (1 for isothermal and 2 for thermal). [CompositionalMultiphaseWell.cpp:1379-1388](https://github.com/GEOS-DEV/GEOS/blob/develop/src/coreComponents/physicsSolvers/fluidFlow/wells/CompositionalMultiphaseWell.cpp#L1379-L1388)

PR #3957, opened 2026-02-02
test: mixed-dimensional consistency

### LINKED ISSUE #2959: Adding check and warning message for fracture pressure
Sanity check is missing to verify the initialization of fracture pressure.

For this [tutorial example](https://github.com/GEOS-DEV/GEOS/pull/2958/files#diff-40c035969f966e550a519f9cf65c96ee4665afa22e0a37233d01afc7dd1ad3bf), if following block is missing in the xml file, 
```
    <FieldSpecification
      name="prescribedFaultPressure"
      setNames="{ all }"
      objectPath="ElementRegions/Fault"
      fieldName="pressure"
      scale="35.0e6"/>
```
simulation result deviates away from the corresponding analytical solution.
image

Sanity check and warning message are expected after creating fracture surface.

### LINKED ISSUE #3909: Initial Pressure in the Fracture
This issue arises when we model Hydraulic Fracture, as in the basic example in the tutorials. When running the hydraulic fracture simulation with an initial pressure >0, the simulated pressure in the fracture always starts at zero or at much lower pressures than the assigned pi, even if the properties have been defined to model leak-off and poroelasticity. I have defined initial pore pressure in various ways, as shown below (for example). However, the fracture pressure always starts at much lower pressures (it seems that setting the initial pressure does not have an effect on the fracture pressure). 

    <FieldSpecification
      name="initialPressureFracture"
      initialCondition="1"
      setNames="{ all }"
      objectPath="ElementRegions/Fracture/FractureSubRegion"
      fieldName="pressure"
      scale="4.999e6"/>

It would be beneficial to illustrate a scenario where the initial pressure is not zero. For example, the basic HF example has a pore pressure CSV, but it is never used.  There are numerous conflicts with solvers, etc., when this is activated, and when conflict is resolved, the pore pressure does not follow the set value(s). It is essential to observe how pressure increases during injection and how it changes during fall-off as a function of closure pressure.

OmarDuran:
This PR introduces a new suite of integration tests designed to verify the consistency and correctness of several core capabilities: the Finite Element Method (FEM) mechanical solver `SolidMechanicsAugmentedLagrangianContact`, single-phase flow, compositional hydrostatic equilibrium, and the workflow for defining fractures using the `SurfaceGenerator` class.

These tests specifically target the solution of mixed-dimensional problems generated with the internal Surface Generator, where fractures (2D entities) are conformal within a 3D matrix mesh. The test suite also includes verification of the topological modifications induced in the mesh by fracture generation.

Among its key benefits, this PR enables a more robust initialization procedure for mechanical problems with fractures. It allows for the following workflow: initialize the intact rock problem, perform a fracturing step (generation of fractures), initialize fracture tractions based on the already computed surrounding stress, and finally equilibrate the entire system with fractures before the actual dynamic simulation.
```
<Events minTime="0.0" maxTime="1.0">
    <SoloEvent name="IntactRockInitialization" targetTime="0.0" beginTime="0.0" target="/Tasks/PRE.SPLIT.STEP"/>
    <SoloEvent name="FracturingStep" target="/Solvers/SurfaceGen" targetTime="0.0" beginTime="0.0"/> 
    <SoloEvent name="FracturedRockInitialization" targetTime="0.0" beginTime="0.0" target="/Tasks/POST.SPLIT.STEP"/>
    <PeriodicEvent name="solverApplications" target="/Solvers/mechSolver" forceDt="1.0"/>
</Events>
```

The following four C++ integration test files are introduced; **all other code modifications are strictly test-driven** and were required to support these new integration tests.

All simulations shown in the image sequence were executed using 12 MPI ranks.
C1: Mesh configuration containing five fractures with no topological connection to the domain boundary.
(Expected Euler characteristic of the resulting 3D mesh: 2)

C2: Mesh configuration containing three intersecting fractures that are topologically connected to the domain boundary.
(Expected Euler characteristic of the resulting 3D mesh: 8)

C3: Mesh configuration containing three perfectly aligned fractures.
(Expected Euler characteristic of the resulting 3D mesh: 2)

### testFEMConsistency.cpp (Serial and MPI test)
The test ensures that the mechanical solvers preserve stress equilibrium and correctly handle contact/interface conditions across different mesh types and fracture configurations.

The physical setup corresponds to a pure compression case (-10 [MPa]), with no shear induced by the boundary conditions. As a result, the stress field is constant, which allows for direct verification. For volume elements, the averaged computed stress tensor is checked against the applied boundary tractions. For fracture surfaces, the traction vector 
𝑡=𝜎⋅𝑛 is applied neighboring volume cells onto the fracture faces, computed, and verified on both sides of the fracture against the expected traction. The value of the normal traction (-10 [MPa]) can be easily verified because the loading condition is hydrostatic and the displacement solution is trilinear and exact.
The test ensures that the mechanical solvers preserve stress equilibrium and correctly handle contact/interface conditions across different mesh types and fracture configurations.

Regardless of the DFN geometry, the normal traction is -10[MPa] and the displacement jump is zero; that is, continuity of stresses and displacements is obtained:
C1
stress_state

C2
stress_state_full_span

### testMixedDimHydrostaticEquilibrium.cpp  (Serial and MPI test)

This test verifies that a hydrostatic equilibrium state defined in both the matrix (3D elements) and fractures (2D elements) with compatible boundary conditions, it is indeed a valid solution of the underlying mixed-dimensional equilibrium problem.

Because the system is initialized exactly at equilibrium, executing a single simulation step should not modify the solution regardless of the time step size. The test therefore checks that the numerical solver preserves the hydrostatic state without introducing spurious pressure or saturation changes.

Regardless of the DFN geometry, saturation and pressure on fractures and matrix should be aligned:
C1
sat_pressure_example

C2
sat_pressure_example_full_span

### testMixedDimSinglePhaseFlow.cpp (Serial and MPI test)

This test verifies that the single-phase flow solver reproduces exact linear pressure solutions in mixed-dimensional settings. It ensures that the solver correctly handles contact and interface conditions between matrix and fracture elements across different mesh configurations, while preserving the exactness of the analytical pressure solution.

C3 satisfies K-orthogonality; therefore, exactness in linear pressure solutions is expected:
pressure_state

### testSurfaceGenerator.cpp (Serial and MPI test)
This test validates the correctness of the Surface Generator by ensuring that the mesh undergoes the expected topological modifications during fracture generation.

The tests cover a variety of fracture configurations and fracture networks (62 configurations), which are compiled in the following directory:`src/coreComponents/integrationTests/meshTests/dfn_market`. It verifies:
- The expected number of fractures created in the mesh
- The correct duplication of nodes required for fracture separation
- The Euler characteristic of the mesh before and after the split, ensuring that the topological transformation is consistent

Special acknowledgments to @rrsettgast for his guidance and suggestions during the implementation of these changes, and to @castelletto1 for providing the five-fracture mesh configurations.

https://github.com/GEOS-DEV/LvArray/pull/354


src/coreComponents/mesh/FaceElementSubRegion.cpp
```diff
@@ -1238,4 +1258,90 @@ void FaceElementSubRegion::fixNeighboringFacesNormals( FaceManager & faceManager
 
 }
 
+void FaceElementSubRegion::orderKf1NodesConsistentlyWithKf0( FaceManager & faceManager,
+                                                             NodeManager const & nodeManager )
+{
```
rrsettgast, line 1263: This breaks the right hand rule assumption of the node ordering for an external face. This must be true for all external faces in the code. The contact kernels should understand this and act accordingly rather than reorder the nodes. There should be unwinding when dealing with the nodes in contact algorithms.

OmarDuran: Based on our internal discussion, this change does not break the right-hand rule assumption. Instead, it corrects an existing issue where the node ordering was already violating that assumption.

src/coreComponents/physicsSolvers/FieldApplicator.cpp
```diff
@@ -0,0 +1,429 @@
...
+
+  registerWrapper( viewKeyStruct::fieldSpecificationNamesString(), &m_fieldSpecificationNames ).
+    setRTTypeName( rtTypes::CustomTypes::groupNameRefArray ).
+    setInputFlag( dataRepository::InputFlags::REQUIRED ).
+    setDescription( "Array containing the field specifications to apply to newly created elements" );
+
+  registerWrapper( viewKeyStruct::solverNameString(), &m_solverName ).
+    setRTTypeName( rtTypes::CustomTypes::groupNameRef ).
+    setInputFlag( dataRepository::InputFlags::OPTIONAL ).
+    setDescription( "Name of the flow solver to use for fluid state initialization (required for compositional flow)" );
```
rrsettgast, line 58: This shouldn't be here. The applicator should be solver agnostic.

herve-gross (approved):
Great work to improve robustness. Thank you.

src/coreComponents/mesh/graphs/GraphColoringBase.cpp
```diff
@@ -67,32 +67,28 @@ bool GraphColoringBase::isColoringValid( const std::vector< camp::idx_t > & adjn
...
-    MpiWrapper::iRecv( &neighborColors[i], 1, neighborRank, 0, comm, &requests[i * 2 + 1] );
-  }
-
-  // Wait for all asynchronous communications to complete
-  MpiWrapper::waitAll( requests.size(), requests.data(), MPI_STATUSES_IGNORE );
-
+  // Use a collective allgather to retrieve every rank's color.
+  // The original point-to-point iSend/iRecv approach with a fixed tag is
+  // fragile in Release builds: messages from successive calls to getColor()
+  // (one per SurfaceGenerator time-step) can linger in the MPI queue and be
```
bd713, line 73: Since the partitioning remains unchanged, there is no real need to recompute and verify the coloring at every `surfaceGenerator` time-step. 
In the future, we should compute and check the coloring once (similar to PR3847), and `getColor()` would simply return the stored value.

src/coreComponents/mesh/mpiCommunications/SpatialPartition.cpp
```diff
@@ -117,10 +119,66 @@ int SpatialPartition::getColor()
...
   else
   {
-    // External partitioner such as ParMetis or PTScotch (for VTK external mesh)
-    std::vector< camp::idx_t > adjncy;
-    adjncy.reserve( m_metisNeighborList.size());
-    std::copy( m_metisNeighborList.begin(), m_metisNeighborList.end(), std::back_inserter( adjncy ));
+    // External partitioner such as ParMetis or PTScotch (for VTK external mesh).
+    //
+    // findNeighborRanks() runs independently on each rank using bounding-box intersection,
+    // which can produce ASYMMETRIC neighbor lists in Release builds: rank A may include rank B
```
bd713, line 125: Should we make `findNeighborRanks` symmetric in the first place?

src/coreComponents/mesh/graphs/RLFGraphColoringMPI.cpp
```diff
@@ -54,6 +55,39 @@ std::vector< int > RLFGraphColoringMPI::colorGraph( const std::vector< camp::idx
   std::vector< int > colors;
   if( rank == 0 )
   {
+    // Symmetrize the gathered graph before coloring.
+    // findNeighborRanks() can produce asymmetric neighbor lists in Release builds
```
bd713, line 59: Fully agreed that we need more robust QC on the input.
Now, how redundant is this with the symmetrization already performed on the split graph in the calling function?
Could we simply check that the symmetry assumption is not violated and raise an error, instead of fixing the graph in a different place?

src/coreComponents/linearAlgebra/interfaces/trilinos/EpetraMatrix.cpp
```diff
@@ -581,8 +581,18 @@ void EpetraMatrix::computeScalingVector( EpetraVector & scaling ) const
...
+    // so we set scaling to 1.0 (no scaling)
+    if( weights[c] > std::numeric_limits< real64 >::epsilon() )
+    {
+      // TODO: consider digit truncation like hypre does
+      weights[c] = LvArray::math::sqrt( LvArray::math::invSqrt( weights[c] ) );
+    }
+    else
+    {
+      weights[c] = 1.0;
+    }
```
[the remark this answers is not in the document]
OmarDuran: limits are included

src/coreComponents/linearAlgebra/interfaces/petsc/PetscMatrix.cpp
```diff
@@ -706,8 +706,18 @@ void PetscMatrix::computeScalingVector( PetscVector & scaling ) const
...
+    // so we set scaling to 1.0 (no scaling)
+    if( weights[c] > std::numeric_limits< real64 >::epsilon() )
+    {
+      // TODO: consider digit truncation like hypre does
+      weights[c] = LvArray::math::sqrt( LvArray::math::invSqrt( weights[c] ) );
+    }
+    else
+    {
+      weights[c] = 1.0;
+    }
```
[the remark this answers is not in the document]
OmarDuran: limits are included.

src/coreComponents/mesh/FaceElementSubRegion.cpp
```diff
@@ -1238,4 +1258,90 @@ void FaceElementSubRegion::fixNeighboringFacesNormals( FaceManager & faceManager
...
+    if( numNodes0 < 4 )
+      return;
+
+    // For each node in kf0, find the closest node in kf1 and build a reordering.
+    // After mesh splitting, collocated nodes are at (nearly) identical positions.
+    localIndex const numNodes = numNodes0;
+    constexpr localIndex MAX_FACE_NODES = 8;
+    localIndex reorderedKf1[MAX_FACE_NODES];
+    bool matched[MAX_FACE_NODES];
+
```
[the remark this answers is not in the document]
OmarDuran: This PR is intended to support geometries where fractures are defined on tetrahedral or hexahedral regions of the mesh. General polyhedral faces are out of scope (only tetrahedra and hexahedra are supported).

An error message is included to reflect this constraint.

src/coreComponents/mesh/generators/ParMETISInterface.cpp
```diff
@@ -58,23 +58,34 @@ meshToDual( ArrayOfArraysView< idx_t const, idx_t > const & elemToNodes,
...
+      graph.appendToArray( k, adjncy + xadj[k], adjncy + xadj[k+1] );
+    } );
 
-  METIS_Free( xadj );
-  METIS_Free( adjncy );
+    METIS_Free( xadj );
+    METIS_Free( adjncy );
+  }
 
   return graph;
```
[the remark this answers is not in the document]
OmarDuran: The frees are guarded by null-check to avoid leaking.

src/coreComponents/physicsSolvers/surfaceGeneration/ParallelTopologyChange.cpp
```diff
@@ -35,6 +41,40 @@ using namespace dataRepository;
...
+
+    if( launchErr != cudaSuccess )
+    {
+      printf( "  launch: %s\n", cudaGetErrorString( launchErr ) );
+    }
+
+    if( syncErr != cudaSuccess )
+    {
+      printf( "  sync: %s\n", cudaGetErrorString( syncErr ) );
+    }
```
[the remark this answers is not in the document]
OmarDuran: The checks are guarded by a debug flag to avoid unconditional device-wide synchronization in production.

OmarDuran:
At the current time, Copilot’s reviews and suggestions are far from correct. It seems the quality of such AI reviews degenerates with the amount of modifications and the complexity of the codebase.

castelletto1 (approved):

PR #3964, opened 2026-02-05
fix: Add fix to fracture state in case of high cohesion

npillardou:
This PR tries to correct a condition found in the ALM method.

The implementation is based from the LM method (  physicsSolvers/solidMechanics/contact/SolidMechanicsLagrangeContact.cpp:2274-2335) : 

 - when normal traction is smaller than normalTractionTolerance, check if cohesion allows stick state before deciding to transition to slip (fix for high cohesion materials)

Some tests have been performed showing a better consistency in Aramis case with very high cohesion (1e20): 
Previous implementation:
image

With the fix: 
image

### DIFF
--- a/src/coreComponents/physicsSolvers/solidMechanics/contact/kernels/SolidMechanicsALMKernelsBase.hpp
+++ b/src/coreComponents/physicsSolvers/solidMechanics/contact/kernels/SolidMechanicsALMKernelsBase.hpp
@@ -151,7 +151,7 @@ struct UpdateStateKernel
         // When normal traction is very small, check if cohesion allows stick state
         // before deciding to transition to slip (fix for high cohesion materials)
         real64 dLimitTau_dTraction = 0.0;
-        real64 const limitTau = contactWrapper.computeLimitTangentialTractionNorm( 0.0, dLimitTau_dTraction );
+        real64 const limitTau = contactWrapper.computeLimitTangentialTractionNorm( localTractionNew[ 0 ], dLimitTau_dTraction );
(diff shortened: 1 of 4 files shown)


src/coreComponents/physicsSolvers/solidMechanics/contact/kernels/SolidMechanicsALMKernelsBase.hpp
```diff
@@ -144,13 +144,24 @@ struct UpdateStateKernel
...
+        // before deciding to transition to slip (fix for high cohesion materials)
+        real64 dLimitTau_dTraction = 0.0;
+        real64 const limitTau = contactWrapper.computeLimitTangentialTractionNorm( localTractionNew[ 0 ], dLimitTau_dTraction );
+        real64 const currentTau = LvArray::math::sqrt( localTractionNew[1] * localTractionNew[1] +
+                                                       localTractionNew[2] * localTractionNew[2] );
+
+        if( currentTau > limitTau )
+        {
+          // Tangential traction exceeds cohesion limit: transition to slip
+          LvArray::tensorOps::fill< 3 >( localTractionNew, 0.0 );
```
jhuang2601: @castelletto1 @jafranc 
If `fields::contact::FractureState::Slip`, why zero out `localTractionNew`?
For open element, it is correct; but for the slipped ones, I guess the normal traction (`localTractionNew[0]`) should be preserved?

npillardou: I think you're right, I kept it the same as before, but logically the traction should remain the same, if not zero.

jhuang2601 (approved):

PR #3836, opened 2025-09-26
feat: Split statistics components computation & output

MelReyCG:
Issue: #3822 
On top of PR : #3829 

Planned refactor: The goal is to have a way to query statistics for solver (reservoir reference pressure for wells), with a clear separation with output statistics component.

The solution proposed here aims at giving a single data storage for each component needing statistics computation. In practices, the statistics are no longer recomputed for each convergence step, but only once, at start of each iteration (it gives the same result since the operation is read only).

To achieve these goals, the PR propose to **separate each roles, instead of having everything in each `*PhaseStatistics` class**:
- `RegionStatistics` classes for resulting data storages at different levels:
  * sub-regions mpi-local kernel compute read-backs,
  * (cell element) regions,
  * solver discretisation `MeshLevel`,
- `StatsAggregator` classes to compute & reduce the statistics to rank 0, can be instanciated as solver helpers (**as done for wells in the PR**),
- `StatsTask` `Group`, the user component to schedule flow statistics computation & output to log & CSV.

Old stats data architecture (stats are sticked to regions, and shared by all components):
```
Problem : ProblemManager
|-> domain : DomainPartition
    |-> MeshBodies : Group
        |-> cartesianMesh : MeshBody
            |-> meshLevels : Group
                |-> Level0 : MeshLevel
                |   |-> nodeManager : NodeManager
                |   |   |  * RegionStatistics (aggregate, mpi reduced)  <- stats data
                |   |   |-> sets : Group
                |   |       | * all : Wrapper< index array >
                |   |       | * xneg : Wrapper< index array >
                |   |       [...] (other element sets)
                |   |
                |   |-> ElementRegions : ElementRegionManager
                |   |   |  * RegionStatistics (aggregate, mpi reduced)  <- stats data
                |   |   |-> Channel : CellElementRegion
                |   |   |   |-> cb-0_0_0 : CellElementSubRegion
                |   |   |   |   | * pressure : Wrapper< real64 array >
                |   |   |   |   | * temperature : Wrapper< real64 array >
                |   |   |   |   | * RegionStatistics (compute read-back)  <- stats data
                |   |   |   |   [...] (other fields)
                |   |   |   |
                |   |   |   |-> cb-0_0_1 : CellElementSubRegion
                |   |   |   |   | * pressure : Wrapper< real64 array >
                |   |   |   |   | * temperature : Wrapper< real64 array >
                |   |   |   |   | * RegionStatistics (compute read-back)  <- stats data
[...]

PR #3966, opened 2026-02-06
feat: Additional information on the table log perforation 

arng40:
Add well-subregion & rank informations.
Use of MPI Table.

Before the modification :
```
--------------------------------------------------------------
|               Well 'EL001_2' Perforation Table             |
|------------------------------------------------------------|
| Perforation  |        Coordinates         |  Well element  |
|--------------|----------------------------|----------------|
|           0  |  [000000, 1000000, -0000]  |             0  |
|           1  |  [000000, 1000000, -0001]  |             1  |
|           2  |  [000000, 1000000, -0002]  |             2  |
|           3  |  [000000, 1000000, -0003]  |             3  |
|           4  |  [000000, 1000000, -0004]  |             4  |
|           5  |  [000000, 1000000, -0005]  |             5  |
|           6  |  [000000, 1000000, -0006]  |             6  |
--------------------------------------------------------------
```
After
```
------------------------------------------------------------------------------------------------------------------------
|                                           Well 'EL001_2' Perforation Table                                           |
|----------------------------------------------------------------------------------------------------------------------|
|  Rank  |  Perforation  |  Well element  |        Coordinates         |  Cell region  |  Cell sub-region  |  Cell ID  |
|--------|---------------|----------------|----------------------------|---------------|-------------------|-----------|
|    42  |            0  |             0  |  [000000, 1000000, -0000]  |         Cook  |      7_hexahedra  |  1011111  |
|    42  |            1  |             1  |  [000000, 1000000, -0001]  |         Cook  |      7_hexahedra  |  1011112  |
|    42  |            2  |             2  |  [000000, 1000000, -0002]  |    Johanssen  |      8_hexahedra  |  1011113  |
|    42  |            3  |             3  |  [000000, 1000000, -0003]  |    Johanssen  |      8_hexahedra  |  1011114  |
|    42  |            4  |             4  |  [000000, 1000000, -0004]  |    Johanssen  |      8_hexahedra  |  1011115  |
|    42  |            5  |             5  |  [000000, 1000000, -0005]  |    Johanssen  |      8_hexahedra  |  1011116  |
|    42  |            6  |             6  |  [000000, 1000000, -0006]  |    Johanssen  |      8_hexahedra  |  1011117  |
------------------------------------------------------------------------------------------------------------------------
```


src/coreComponents/mesh/WellElementSubRegion.hpp
```diff
@@ -522,6 +561,15 @@ class WellElementSubRegion : public ElementSubRegionBase
 
   /// Number of  segment per rank
   array1d< localIndex > m_elementPerRank;
+
+  /// The reservoir element that contains the perforation
+  globalIndex m_reservoirElementID;
+  /// the target region name for the reservoir element
+  string m_regionName;
+  /// the target sub region name for the reservoir element
+  string m_subRegionName;
```
Melcx: A ``WellElementSubRegion`` can only have one communicating region cell element?
Isn't this data already stored somewhere?

src/coreComponents/mesh/WellElementSubRegion.hpp
```diff
@@ -461,6 +480,26 @@ class WellElementSubRegion : public ElementSubRegionBase
...
+   * @param[inout] eiMatched the element index of the reservoir element that contains "location", if any
+   * @param[inout] giMatched the element global index of the reservoir element that contains "location", if any
+   */
+  bool searchLocalElements( MeshLevel const & mesh,
+                            real64 const (&location)[3],
+                            localIndex const & searchDepth,
+                            localIndex const & targetRegionIndex,
+                            localIndex & esrMatched,
+                            localIndex & eiMatched,
+                            globalIndex & giMatched,
```
Melcx: As you may have a lot of parameters here, maybe a functor with a structured parameter would be simpler to call than a lot of in-out parameters.
Also, the return type is not documented, it is a EOF-style-bool, right? With a functor you could have the loop inside of the search method.

src/coreComponents/mesh/WellElementSubRegion.hpp
arng40: yep I revert wellElementSubRegion cpp & hpp

src/coreComponents/common/format/table/unitTests/testMpiTable.cpp
```diff
@@ -99,13 +100,82 @@ TEST( testMpiTables, testDifferentRankData )
       "-------------------------------------------\n"
     },
   };
+  for( TestCase const & testCase: testCases )
+  {
+    int const rankId = MpiWrapper::commRank();
+    int const nbRanks = MpiWrapper::commSize();
+    if( nbRanks > 1 )
+    {
+      ASSERT_EQ( nbRanks, 4 );
```
MelReyCG: Why moving that & removing the message?
I think you can leave this test as it was.

src/coreComponents/common/format/table/unitTests/testMpiTable.cpp
```diff
@@ -99,13 +100,82 @@ TEST( testMpiTables, testDifferentRankData )
...
+  };
 
   int const rankId = MpiWrapper::commRank();
   int const nbRanks = MpiWrapper::commSize();
-  ASSERT_EQ( nbRanks, 4 ) << "This unit test cases are designed for exactly 4 ranks to check row ordering consistency.";
-
-  for( TestCase const & testCase: testCases )
+  if( nbRanks > 1 )
   {
+    ASSERT_EQ( nbRanks, 4 );
```
MelReyCG: You can use the same as in the other test.

src/coreComponents/common/format/table/unitTests/testMpiTable.cpp
```diff
@@ -99,13 +100,82 @@ TEST( testMpiTables, testDifferentRankData )
...
+    string m_expectedResult;
+  };
+
+  TestCase const testCase =
+  {
+    {   // m_ranksValues: in this test, rank 2 has no value
+      { {1, 0.502} },
+      { {2, 0.624}, {3, 0.791} },
+      {},
+      { {4, 0.243}, {5, 0.804}, {6, 0.302} },
```
MelReyCG: I don't understand what is tested if the data is already sorted. Wouldn't it get the same table if we removed the sorting function?

src/coreComponents/common/format/table/TableData.hpp
```diff
@@ -302,5 +302,17 @@ void TableData2D::addCell( real64 const rowValue, real64 const columnValue, T co
   m_data.get_inserted( rowValue ).get_inserted( columnValue ) =  GEOS_FMT( "{}", value );
 }
 
+// Custom Comp function;
+namespace tabledatasorting
```
MelReyCG: - code style: ``tableDataSorting``
- As it is declared at this level (``TableData.hpp`` and not ``TableMpiComponents.hpp``), I would have expected a sorting mecanism available in ``TableData``. Could the sorting function be easily moved? (``TableData::sort(functor) -> std::sort()``)

src/coreComponents/mesh/ElementRegionManager.cpp
```diff
@@ -229,6 +230,66 @@ void ElementRegionManager::generateWells( CellBlockManagerABC const & cellBlockM
 
   } );
 
+  forElementRegions< WellElementRegion >( [&]( WellElementRegion const & wellRegion ){
+    WellElementSubRegion const &
+    wellSubRegion = wellRegion.getGroup( ElementRegionBase::viewKeyStruct::elementSubRegions() )
+                      .getGroup< WellElementSubRegion >( wellRegion.getSubRegionName() );
```
MelReyCG: Prefer ``getSubRegion()`` over reimplementing ``dataRepository`` access.

src/coreComponents/mesh/ElementRegionManager.cpp
```diff
@@ -229,6 +230,66 @@ void ElementRegionManager::generateWells( CellBlockManagerABC const & cellBlockM
...
+    TableLayout const layoutPerforation ( GEOS_FMT( "Well '{}' Perforation Table",
+                                                    wellRegion.getWellGeneratorName()),
```
MelReyCG: No control for this table logging?

src/coreComponents/mesh/ElementRegionManager.cpp
```diff
@@ -229,6 +230,66 @@ void ElementRegionManager::generateWells( CellBlockManagerABC const & cellBlockM
...
+        auto const & meshElems = wellSubRegionPerforationData->getMeshElements();
+
+        localIndex const targetRegionIndex = meshElems.m_toElementRegion[iperfLocal];
+        localIndex const targetSubRegionIndex = meshElems.m_toElementSubRegion[iperfLocal];
+
+        ElementRegionBase const & region =
+          meshLevel.getElemManager().getRegion< ElementRegionBase >( targetRegionIndex );
+
+        ElementSubRegionBase const & subRegion = region.getSubRegion< ElementSubRegionBase >( targetSubRegionIndex );
+        integer const localWellElemIndices = wellSubRegion.getGlobalWellElementIndex()[iperfLocal];
```
MelReyCG: Could ``localWellElemIndices`` be renamed?

src/coreComponents/mesh/ElementRegionManager.cpp
```diff
@@ -229,6 +230,66 @@ void ElementRegionManager::generateWells( CellBlockManagerABC const & cellBlockM
...
+        localCoords.emplace_back( wsrPerfLocation[iperfLocal][0] );
+        localCoords.emplace_back( wsrPerfLocation[iperfLocal][1] );
+        localCoords.emplace_back( wsrPerfLocation[iperfLocal][2] );
+        localPerfoData.addRow( globalIperf[iperfLocal], localWellElemIndices, localCoords,
+                               region.getName(), subRegion.getName(), cellId, rankId );
+      }
+    }
+
+    TableMpiLayout mpiLayout;
+    TableTextMpiOutput formatter = TableTextMpiOutput( layoutPerforation, mpiLayout );
```
MelReyCG: If you want to, you can rename it ``TableTextMpiFormatter``

src/coreComponents/mesh/ElementRegionManager.cpp
```diff
@@ -229,6 +230,66 @@ void ElementRegionManager::generateWells( CellBlockManagerABC const & cellBlockM
...
+
+    formatter.setSortingFunc(
+      []( std::vector< TableData::CellData > const & row1,
+          std::vector< TableData::CellData > const & row2 ) {
+      return tabledatasorting::positiveNumberStringComp( row1[1].value, row2[1].value );
```
MelReyCG: Did you forgot to set that to ``[0]``?

src/coreComponents/common/MpiWrapper.hpp
```diff
@@ -331,6 +331,85 @@ struct MpiWrapper
...
+  /**
+   * @brief Structure holding the result from all the gather operation
+   * @tparam CONTAINER The container type holding the data.
+   */
+  template< typename CONTAINER >
+  struct GatherResult
+  {
+    CONTAINER data;                   // Collected data
+    stdVector< integer > counts;      // Number of elements per row
+    stdVector< integer > offsets;     // Starting index for each row in 'data'
```
MelReyCG: Code style
```suggestion
    /// Collected data
    CONTAINER data;
    /// Number of elements per row
    stdVector< integer > counts;
    /// Starting index for each row in 'data'
    stdVector< integer > offsets;
```

MelReyCG: + I would explain a bit more what ``data`` & ``CONTAINER`` are. I think CONTAINER must be contiguous, right? can you give some examples of its types?

src/coreComponents/common/MpiWrapper.hpp
```diff
@@ -331,6 +331,85 @@ struct MpiWrapper
...
+      std::is_trivially_copyable_v< VALUE_T > &&
+      std::is_same_v< decltype(std::declval< CONTAINER >().data()), VALUE_T * > &&
+      std::is_same_v< decltype(std::declval< CONTAINER >().size()), std::size_t >
+      >
+    >
+  static GatherResult< CONTAINER >
+  gatherBufferRank0( CONTAINER const & localBuffer )
+  {
+    integer const numRanks = MpiWrapper::commSize();
+    integer const numValues = static_cast< integer >(localBuffer.size());
```
MelReyCG:
```suggestion
    integer const numLocalValues = static_cast< integer >(localBuffer.size());
```

src/coreComponents/common/MpiWrapper.cpp
```diff
@@ -518,6 +518,28 @@ template<> MPI_Datatype getMpiPairType< double, double >()
 
 } /* namespace internal */
 
+template<>
+void MpiWrapper::gatherStringOnRank0< std::function< void(string_view) > >
+  ( string_view rankStr, std::function< void(string_view) > && func )
+{
```
MelReyCG: By declaring this in the CPP instead of in the HPP, you're basically constraining the templated type ``FUNC`` to be ``std::function< void(string_view) >``. What about removing the template and be explicit on the expected type in the HPP?

src/coreComponents/common/format/table/TableMpiComponents.cpp
```diff
@@ -118,45 +147,76 @@ void TableTextMpiOutput::outputTableDataToRank0( std::ostream & tableOutput,
...
   }
-
-  // all other ranks than rank 0 render their output in a string and comunicate its size
-  stdVector< integer > ranksStrsSizes = stdVector< integer >( ranksCount, 0 );
   string const rankStr = !status.m_isMasterRank && status.m_isContributing ? localStringStream.str() : "";
-  integer const rankStrSize = rankStr.size();
-  MpiWrapper::gather( &rankStrSize, 1, ranksStrsSizes.data(), 1, 0 );
+  stdVector< string > strsAccrossRanks;
+
+  MpiWrapper::gatherStringOnRank0( rankStr, std::function< void(string_view) >( [&]( string_view str ){
```
MelReyCG: Is it possible to
```suggestion
  MpiWrapper::gatherStringOnRank0( rankStr, [&] ( string_view str ) {
```

src/coreComponents/common/format/table/TableMpiComponents.cpp
```diff
@@ -118,45 +147,76 @@ void TableTextMpiOutput::outputTableDataToRank0( std::ostream & tableOutput,
...
+    initalizeTableGrids( m_tableLayout, tableData,
+                         headerCellsLayout, dataRows, errorRows,
+                         tableTotalWidth, columnWidthModifier );
+    status.m_sepLine = string( tableTotalWidth, m_horizontalLine );
+  }
 
-    for( integer rankId = 1; rankId < ranksCount; ++rankId )
+  if( m_sortingFunctor )
+  {
+    gatherSortAndOutput( tableOutput, dataRows, status );
```
MelReyCG: I don't get why headerCellsLayout & errorCellsLayout get ignored in this branch, are they discarded?
Are the header & footer (with errors) correctly output?

src/coreComponents/common/format/table/TableMpiComponents.cpp
```diff
@@ -118,45 +147,76 @@ void TableTextMpiOutput::outputTableDataToRank0( std::ostream & tableOutput,
...
+    !tableData.getCellsData().empty(),
+    // m_hasContent
+    false,
+    // m_sepLine
+    ""
+  };
+
+  CellLayoutRows headerCellsLayout;
+  CellLayoutRows dataRows;
+  CellLayoutRows errorRows;
```
MelReyCG: You renamed a part of the variables, but not all, is there a reason?

src/coreComponents/common/format/table/unitTests/testMpiTable.cpp
```diff
@@ -125,15 +195,62 @@ TEST( testMpiTables, testDifferentRankData )
...
+    formatter.setSortingFunc( []( std::vector< TableData::CellData > const & row1,
+                                  std::vector< TableData::CellData > const & row2 ) {
+      return tabledatasorting::positiveNumberStringComp( row1[0].value, row2[0].value );
+    } );
+
     std::ostringstream oss;
     formatter.toStream( oss, data );
     if( rankId == 0 )
     {
+      std::cout << "ma boula " ""<< oss.str()<<std::endl;
```
MelReyCG: ????

src/coreComponents/common/format/table/TableMpiComponents.hpp
```diff
@@ -98,11 +110,54 @@ class TableTextMpiOutput : public TableTextFormatter
...
-                               CellLayoutRows const & dataCellsLayout,
-                               Status & status ) const;
 
+  /**
+   * @brief Parse each row of a CellLayoutRows into its string representation.
+   * @param cellLayoutRows The target CellLayoutRows to parse
+   * @param rowsAsString The result vector of strings
+   */
+  void parseCellLayoutRows( CellLayoutRows const & cellLayoutRows,
+                            stdVector< string > & rowsAsString ) const;
```
MelReyCG:
```suggestion
  void cellRowsToStrings( CellLayoutRows const & cellLayoutRows,
                          stdVector< string > & rowsAsString ) const;
```
(parse would be the opposite)

src/coreComponents/common/format/table/TableMpiComponents.cpp
MelReyCG: I suspect that you sort formatted data, with spacing being able to influence the sorting.
Wouldn't it be easier to, in the case of the sorting functor, 1. start first by collecting the other ranks ``TableData`` lines (before any CellLayoutRows matter), 2. do a classic ``TableTextFormatter::toStream()``?

MelReyCG (changes requested):

src/coreComponents/common/format/table/TableMpiComponents.cpp
arng40: compilation issue [-fpermissive]

src/coreComponents/mesh/ElementRegionManager.cpp
arng40: I sort by local well indices

src/coreComponents/common/format/table/TableMpiComponents.cpp
arng40: An oversight , I renamed them.

src/coreComponents/common/format/table/TableMpiComponents.cpp
arng40: This part is indeed misleading, `gatherSortAndOutput` call toString which do all the formatting with `headerRows` and `errorRows`

arng40: I'll create a new branch and submit my new proposal

src/coreComponents/common/format/table/unitTests/testMpiTable.cpp
arng40: fixed

src/coreComponents/common/format/table/unitTests/testMpiTable.cpp
arng40: fixed

src/coreComponents/mesh/ElementRegionManager.cpp
arng40: added

src/coreComponents/mesh/PerforationFields.hpp
```diff
@@ -61,7 +61,7 @@ DECLARE_FIELD( reservoirElementIndex,
 DECLARE_FIELD( reservoirElementGlobalIndex,
                "reservoirElementGlobalIndex",
                array1d< globalIndex >,
-               0,
+               -1,
```
MelReyCG: Why do you need it? Is it safe? It scares me :D

src/coreComponents/common/format/table/unitTests/testMpiTable.cpp
MelReyCG: you did not revert.. as you did, it seems like ``MpiWrapper::commRank()`` and ``MpiWrapper::commSize()`` can vary.

src/coreComponents/common/format/table/TableMpiComponents.cpp
MelReyCG: So why can I do that in #3836 ? 
```c++
  m_aggregator->forRegionStatistics( mesh, meshRegionsStatistics,
                                     [&] ( CellElementRegion & region, RegionStatistics & stats )
  {
    outputRegionStats( GEOS_FMT( "Region '{}'", region.getName() ), stats );
  } );
```

src/coreComponents/common/MpiWrapper.hpp
```diff
@@ -331,6 +331,53 @@ struct MpiWrapper
...
 
+  /**
+   * @brief Structure holding the result from all the gather operation
+   * @tparam CONTAINER The container type holding the data.
+   * The underlying storage in CONTAINER must be contiguous
+   */
+  template< typename CONTAINER >
+  struct GatherResult
+  {
+    // Collected data who must be trivially copyable
```
MelReyCG: I don't think there will be any phylosophical debate for that kind of data: who -> which

src/coreComponents/common/MpiWrapper.hpp
```diff
@@ -331,6 +331,53 @@ struct MpiWrapper
...
+    CONTAINER data;
+    // Number of elements per row
```
MelReyCG: "row"? what do you think of "rank string"?

src/coreComponents/common/format/table/TableData.hpp
```diff
@@ -302,5 +323,91 @@ void TableData2D::addCell( real64 const rowValue, real64 const columnValue, T co
   m_data.get_inserted( rowValue ).get_inserted( columnValue ) =  GEOS_FMT( "{}", value );
 }
 
+// Serialisation/ Deserialisation utils for common
+namespace serialBuffer
```
MelReyCG:
```suggestion
namespace basicSerialization
```

src/coreComponents/common/format/table/TableMpiComponents.hpp
```diff
@@ -98,11 +110,31 @@ class TableTextMpiOutput : public TableTextFormatter
...
-  void outputTableDataToRank0( std::ostream & tableOutput,
-                               PreparedTableLayout const & tableLayout,
-                               CellLayoutRows const & dataCellsLayout,
-                               Status & status ) const;
+  /**
+   * @brief Parse a string row to a TablaData cells.
+   * @param rowString The string row string to parse.
+   * @return The parsed row as a vector of CellData.
+   */
+  stdVector< TableData::CellData > parseStringRow( string_view rowString ) const;
```
MelReyCG: it is now dead code isn't it?

src/coreComponents/common/format/table/unitTests/testMpiTable.cpp
MelReyCG: randomize more please, they are already ordered on each rank + leave the empty one between the others

src/coreComponents/common/format/table/unitTests/testMpiTable.cpp
MelReyCG: You can use the same as in the other test.

src/coreComponents/mesh/ElementRegionManager.cpp
MelReyCG: Shouldn't you sort by perforation id?

MelReyCG (changes requested):

src/coreComponents/common/MpiWrapper.hpp
arng40: rank seems more appropriate

src/coreComponents/mesh/PerforationFields.hpp
arng40: If it matches a GI of 0, it will not be added to the table with this cond 
l 247 `if( cellId != -0 )`

src/coreComponents/mesh/WellElementSubRegion.cpp
```diff
@@ -965,12 +970,21 @@ void WellElementSubRegion::connectPerforationsToMeshElements( MeshLevel & mesh,
...
+      { // all ranks exit the search
         break;
+      }
+    }
+
+    if( !globalResElemFound )
+    {
+      GEOS_WARNING( GEOS_FMT( "Perforation {} not maching any regions", iperfGlobal ));
     }
+    //set de boolean
```
MelReyCG: 👀

src/coreComponents/mesh/WellElementSubRegion.cpp
```diff
@@ -965,12 +970,21 @@ void WellElementSubRegion::connectPerforationsToMeshElements( MeshLevel & mesh,
...
+      globalResElemFound |= (bool) MpiWrapper::allReduce( (integer) localResElemFound, MpiWrapper::Reduction::LogicalOr );
+      if( globalResElemFound )
+      { // all ranks exit the search
         break;
+      }
+    }
+
+    if( !globalResElemFound )
+    {
+      GEOS_WARNING( GEOS_FMT( "Perforation {} not maching any regions", iperfGlobal ));
```
MelReyCG: instead of id, output data-ctx of perforation please

src/coreComponents/mesh/PerforationData.hpp
```diff
@@ -152,6 +152,23 @@ class PerforationData : public ObjectManagerBase
    */
   arrayView1d< localIndex > getWellElements() { return m_wellElementIndex; }
 
+  /**
+   * @return an array containing the values of the local perforations connected to a reservoir element
```
MelReyCG:
```suggestion
   * @return an array of booleans each value of hasLocalPerforationInReservoir() per perforation.
```

MelReyCG (approved):

MelReyCG:
The integrated tests now runs successfully on this PR.

src/coreComponents/common/format/table/TableData.hpp
jafranc, line 54: typo ?

src/coreComponents/common/format/table/TableData.cpp
```diff
@@ -59,7 +59,81 @@ TableData & TableData::operator=( TableData const & other )
 
 bool TableData::operator<( TableData const & other ) const
 {
-  return m_rows < other.m_rows;
+  if( other.getCellsData().size()!= getCellsData().size())
+    return false;
```
jafranc, line 63: Does this need a comment in the function doc ?

src/coreComponents/common/format/table/TableData.cpp
```diff
@@ -59,7 +59,81 @@ TableData & TableData::operator=( TableData const & other )
...
+{
+  if( comparingTable.getCellsData().size()!= getCellsData().size())
+    return false;
+  for( size_t i = 0; i < getCellsData().size(); i++ )
+  {
+    if( getCellsData()[i].data()->value  != comparingTable.getCellsData()[i].data()->value )
+      return false;
+  }
+  return true;
+}
```
jafranc, line 83:
```suggestion
bool TableData::operator==( TableData const & other ) const
{
  if( other.getCellsData().size()!= getCellsData().size())
    return false;
  for( size_t i = 0; i < getCellsData().size(); i++ )
  {
    if( getCellsData()[i].data()->value  != other.getCellsData()[i].data()->value )
      return false;
  }
  return true;
}
```

src/coreComponents/common/format/table/TableData.hpp
```diff
@@ -55,7 +55,14 @@ class TableData
   bool operator<( TableData const & other ) const;
 
   /**
-   * @brief Representing a data in TableData
+   * @brief Comparison operator for data rows
+   * @param comparingTable The tableData values to compare
+   * @return The comparison result
+   */
+  bool operator==( TableData const & comparingTable ) const;
```
jafranc, line 62:
```suggestion
  bool operator==( TableData const & other ) const;
```

src/coreComponents/common/format/table/TableData.hpp
```diff
@@ -302,5 +323,91 @@ void TableData2D::addCell( real64 const rowValue, real64 const columnValue, T co
...
+ * @param b The string to compare
+ * @return True if a is greater than b
+ */
+bool positiveNumberStringComp( string_view a, string_view b );
+}
+
+template< typename T >
+void basicSerialization::serializePrimitive ( T const data, stdVector< buffer_unit_type > & out )
+{
+  static_assert( std::is_trivially_copyable_v< T > );
```
jafranc, line 396: shouldn't it be `std::enable_if` - ed ?

src/coreComponents/common/format/table/TableMpiComponents.hpp
```diff
@@ -89,6 +98,9 @@ class TableTextMpiOutput : public TableTextFormatter
 
   TableMpiLayout m_mpiLayout;
 
+  /// The custom comparison function object for std::sort
+  std::unique_ptr< SortingFunc > m_sortingFunctor;
```
jafranc, line 102: `std::optional` ?

src/coreComponents/common/format/table/TableMpiComponents.hpp
```diff
@@ -74,6 +75,14 @@ class TableTextMpiOutput : public TableTextFormatter
   template< typename DATASOURCE >
   void toStream( std::ostream & outputStream, DATASOURCE const & tableData ) const;
 
+  /**
+   * @brief Set the Sorting Func object
+   * @param func The callable comparison function object
+   */
+  void setSortingFunc( SortingFunc && func )
+  { m_sortingFunctor = std::make_unique< SortingFunc >( std::move( func )); }
```
jafranc, line 83: Feel that `std::optional` would be clearer on intentions

jafranc (approved):
Great work !! Thanks

src/coreComponents/common/format/table/TableData.cpp
tjb-ltk, line 1: Could you show an example of how to use addRow when the number of arguments is runtime defined ? 
For example reporting a table of phase  / component / phase component properties...  
where number of phases and components is runtime defined  
.and row would have labels  water density,  gas density , comp1_massfrac, comp2_massfrac, comp3_massfrac ,  phasecompfrac_water_comp1,, phasecompfrac_water_comp2, phasecompfrac_comp3, ....

tjb-ltk (changes requested):

src/coreComponents/common/format/table/TableData.cpp
MelReyCG: I've got one [here ->](https://github.com/GEOS-DEV/GEOS/pull/3836/changes#diff-d08762b25775af40df58eeae407acb8447ae290e09dc7d7cdadebcfb2c78da30R190)

src/coreComponents/common/format/table/TableData.cpp
MelReyCG: Clarified the documentation, and reimplemented those two comparison operators to match the described logic.

src/coreComponents/common/format/table/TableData.cpp
tjb-ltk: This looks good.  What about reporting quantities at the newton iteration level ?

tjb-ltk (approved):

PR #3967, opened 2026-02-09
fix: 2D/3D cell co-location in parallel mesh redistribution

bd713:
### Problem
The current `redistributeMeshes()` partitions the entire mesh (both 2D and 3D cells) all at once using graph partitioning.
However, there is no guarantee that a 2D surface element will be assigned to a rank containing its adjacent 3D face.
This can lead to configurations where 2D cells are orphaned on ranks with no access to their 3D neighbor faces. 
See screenshot below for a real life illustration.
A side effect is that the nodes of these orphaned 2D surfaces are seen as owned by that rank and will prevent them from being properly ghosted in the 3D mesh.

### Solution
In this PR, the redistribution workflow is refactored in 3 steps to **ensure co-location of 2D and 3D cells**:
1. Separate 2D and 3D cells 
2. Redistribute 3D cells independently using existing graph partitioning routines
3. Assign each 2D cell to the rank owning one of its 3D neighbors (deterministic tie-breaking uses minimum global ID)

NB: When using MeshDoctor splitting, the **fracture elements** suffer the exact same problem. These will be addressed in a separate PR.

image

### Rebaseline
As the 2D elements are no longer directly included in the graph to be partitioned, the resulting partitions may differ from the baseline.
This notably affect the following tests:
- impermeableFault_smoke_04
- permeableFault_smoke_04

In addition, we now reject 2D orphan cells that resulted from a bug in MeshDoctor ([geosPythonPackages fix (PR #219)](https://github.com/GEOS-DEV/geosPythonPackages/pull/219)). This required changing the split mesh for:
- ALM_singlephasePoromechanics_curvedFrac_smoke_01
- ALM_singlephasePoromechanics_curvedFrac_smoke_04
- ALM_multiphasePoromechanics_curvedFrac_smoke_01
- ALM_multiphasePoromechanics_curvedFrac_smoke_04


rrsettgast:
@bd713 why not just merge the 2d and 3d meshes into a single mesh object?

src/coreComponents/mesh/generators/VTKUtilities.cpp
```diff
@@ -830,7 +831,385 @@ scatterByBlock( vtkDataSet & mesh )
   return result;
 }
 
+/**
```
castelletto1, line 863:
```suggestion
/**
 * @brief Extract cells by index and create index mapping to original mesh
 *
 * @param[in] mesh Source mesh
 * @param[in] indices Cell indices to extract
 * @param[out] output Deep copy of extracted cells
 * @param[out] mapping Mapping from output cell indices to original mesh indices
 */
 static void extractCellsByIndices( vtkDataSet & mesh,
                                   vtkIdList * indices,
                                   vtkSmartPointer< vtkUnstructuredGrid > & output,
                                   array1d< vtkIdType > & mapping )
{
  vtkNew< vtkExtractCells > extractor;
  extractor->SetInputDataObject( &mesh );
  extractor->SetExtractAllCells( false );
  extractor->SetCellList( indices );
  extractor->Update();

  output = vtkSmartPointer< vtkUnstructuredGrid >::New();
  output->DeepCopy( extractor->GetOutput() );

  mapping.resize( indices->GetNumberOfIds() );
  for( vtkIdType i = 0; i < indices->GetNumberOfIds(); ++i )
  {
    mapping[i] = indices->GetId( i );
  }
}

/**
```


Adding an helper function to avoid doing the same thing for 3D and 2D cells.

Is there a way to avoid using `DeepCopy`?

src/coreComponents/mesh/generators/VTKUtilities.cpp
```diff
@@ -830,7 +831,385 @@ scatterByBlock( vtkDataSet & mesh )
...
+
+  cells2D = vtkSmartPointer< vtkUnstructuredGrid >::New();
+  cells2D->DeepCopy( extractor2D->GetOutput() );
+
+  // Store index mapping for 2D cells
+  cells2DToOriginal.resize( indices2D->GetNumberOfIds() );
+  for( vtkIdType i = 0; i < indices2D->GetNumberOfIds(); ++i )
+  {
+    cells2DToOriginal[i] = indices2D->GetId( i );
+  }
```
castelletto1:
```suggestion
  // Extract and deep-copy 3D cells
  extractCellsByIndices( mesh, indices3D, cells3D, cells3DToOriginal );

  // Extract and deep-copy 2D cells
 extractCellsByIndices( mesh, indices2D, cells2D, cells2DToOriginal );
```

src/coreComponents/mesh/generators/VTKUtilities.cpp
```diff
@@ -830,7 +831,385 @@ scatterByBlock( vtkDataSet & mesh )
...
+{
+  GEOS_MARK_FUNCTION;
+
+  // Retrieve global cell ID array
+  vtkDataArray * globalCellIds = mesh.GetCellData()->GetGlobalIds();
+  GEOS_ERROR_IF( globalCellIds == nullptr,
+                 "Global cell IDs must be present in mesh for 2D-3D neighbor mapping" );
+
+
+// Build reverse lookup: original mesh index -> global cell ID (3D cells only)
```
castelletto1:
```suggestion
  // Build reverse lookup: original mesh index -> global cell ID (3D cells only)
```

src/coreComponents/mesh/generators/VTKUtilities.cpp
```diff
@@ -830,7 +831,385 @@ scatterByBlock( vtkDataSet & mesh )
...
+    {
+      vtkIdType const neighborIdx = neighborCells->GetId( n );
+
+      // Only consider 3D cells
+      if( mesh.GetCell( neighborIdx )->GetCellDimension() == 3 )
+      {
+        auto it = original3DToGlobalId.find( neighborIdx );
+        GEOS_ERROR_IF( it == original3DToGlobalId.end(),
+                       GEOS_FMT( "3D neighbor at index {} not found in 3D cell mapping", neighborIdx ) );
+
```
castelletto1:
```suggestion
      // Check if neighbor is a 3D cell and get its global ID
      auto it = original3DToGlobalId.find( neighborIdx );
      if( it != original3DToGlobalId.end() )  // Is 3D cell
      {
        neighbor3DGlobalIds.emplace_back( it->second );
      }
      // Non-3D neighbors (2D/1D/0D) are silently skipped
```

src/coreComponents/mesh/generators/VTKUtilities.cpp
```diff
@@ -830,7 +831,385 @@ scatterByBlock( vtkDataSet & mesh )
...
+    offsets[r] = totalCells;
+    totalCells += cellCounts[r];
+  }
+
+  // Gather local global IDs
+  array1d< int64_t > localGlobalIds( local3DCells );
+  for( vtkIdType i = 0; i < local3DCells; ++i )
+  {
+    localGlobalIds[i] = static_cast< int64_t >( redistributed3DGlobalIds->GetTuple1( i ) );
+  }
```
castelletto1: Since global IDs are `int64_t`, I would expect `offsets` to be as well `int64_t` and not `int`.

src/coreComponents/mesh/generators/VTKUtilities.cpp
```diff
@@ -830,7 +831,385 @@ scatterByBlock( vtkDataSet & mesh )
...
+
+  // All-gather global IDs to all ranks
+  array1d< int64_t > allGlobalIds( totalCells );
+  MpiWrapper::allgatherv( localGlobalIds.data(), static_cast< int >( local3DCells ),
+                          allGlobalIds.data(), cellCounts.data(), offsets.data(),
+                          comm );
+
+// Build complete partition map: global cell ID -> owning rank
+  stdUnorderedMap< int64_t, int64_t > complete3DPartitionMap;
```
castelletto1:
```suggestion
  stdUnorderedMap< int64_t, int > complete3DPartitionMap;
```

Rank is `int`

src/coreComponents/mesh/generators/VTKUtilities.cpp
```diff
@@ -830,7 +831,385 @@ scatterByBlock( vtkDataSet & mesh )
...
+  complete3DPartitionMap.reserve( totalCells );
+
+  for( int r = 0; r < numRanks; ++r )
+  {
+    for( int i = 0; i < cellCounts[r]; ++i )
+    {
+      int64_t const globalId = allGlobalIds[offsets[r] + i];
+      complete3DPartitionMap.emplace( globalId, r ); // ← FIXED
```
castelletto1:
```suggestion
      complete3DPartitionMap.emplace( globalId, r );
```

src/coreComponents/mesh/generators/VTKUtilities.cpp
```diff
@@ -830,7 +831,385 @@ scatterByBlock( vtkDataSet & mesh )
...
+  for( vtkIdType i = 0; i < local3DCells; ++i )
+  {
+    localGlobalIds[i] = static_cast< int64_t >( redistributed3DGlobalIds->GetTuple1( i ) );
+  }
+
+  // All-gather global IDs to all ranks
+  array1d< int64_t > allGlobalIds( totalCells );
+  MpiWrapper::allgatherv( localGlobalIds.data(), static_cast< int >( local3DCells ),
+                          allGlobalIds.data(), cellCounts.data(), offsets.data(),
+                          comm );
```
castelletto1: This is going to be very expensive for large meshes.

src/coreComponents/mesh/generators/VTKUtilities.cpp
```diff
@@ -830,7 +831,385 @@ scatterByBlock( vtkDataSet & mesh )
...
+    appendFilter->AddInputData( local2DCells );
+    appendFilter->MergePointsOn();  // Ensures shared nodes are not duplicated
+    appendFilter->Update();
+
+    mergedMesh->DeepCopy( appendFilter->GetOutput() );
+
+    //GEOS_LOG_RANK( GEOS_FMT( "Rank merged {} 3D cells + {} 2D cells = {} total",
+    //                         redistributed3D->GetNumberOfCells(),
+    //                         local2DCells->GetNumberOfCells(),
+    //                         mergedMesh->GetNumberOfCells() ) );
```
castelletto1:
```suggestion
```

Remove

src/coreComponents/mesh/generators/VTKUtilities.cpp
```diff
@@ -830,7 +831,385 @@ scatterByBlock( vtkDataSet & mesh )
...
+ * have multiple 3D neighbors (e.g., internal fractures).
+ *
+ * @param[in] neighbors2Dto3D Neighbor map from build2DTo3DNeighbors()
+ * @param[in] globalIdTo3DPartition Complete map of 3D cell global ID -> rank assignment
+ * @param[in] comm MPI communicator (unused, kept for API consistency)
+ * @return Partition (rank) assignment for each 2D cell
+ */
+static array1d< int64_t >
+assign2DCellsTo3DPartitions( ArrayOfArrays< vtkIdType, int64_t > const & neighbors2Dto3D,
+                             stdUnorderedMap< int64_t, int64_t > const & globalIdTo3DPartition )
```
castelletto1:
```suggestion
assign2DCellsTo3DPartitions( ArrayOfArrays< vtkIdType, int64_t > const & neighbors2Dto3D,
                             stdUnorderedMap< int64_t, int > const & globalIdTo3DPartition )
```

src/coreComponents/mesh/generators/VTKUtilities.cpp
```diff
@@ -830,7 +831,385 @@ scatterByBlock( vtkDataSet & mesh )
...
+
+    // Deterministic tie-breaking: choose neighbor with minimum global ID
+    int64_t minGlobalId = neighbors[0];
+    for( localIndex n = 1; n < neighbors.size(); ++n )
+    {
+      if( neighbors[n] < minGlobalId )
+      {
+        minGlobalId = neighbors[n];
+      }
+    }
```
castelletto1:
```suggestion
    int64_t minGlobalId = *std::min_element( neighbors.begin(), neighbors.end() );
```

src/coreComponents/mesh/generators/VTKUtilities.cpp
```diff
@@ -830,7 +831,385 @@ scatterByBlock( vtkDataSet & mesh )
...
+
+/**
+ * @brief Assign partition ownership for 2D cells based on their 3D neighbor locations
+ *
+ * Uses a deterministic tie-breaking rule (minimum global ID) to assign 2D cells that
+ * have multiple 3D neighbors (e.g., internal fractures).
+ *
+ * @param[in] neighbors2Dto3D Neighbor map from build2DTo3DNeighbors()
+ * @param[in] globalIdTo3DPartition Complete map of 3D cell global ID -> rank assignment
+ * @param[in] comm MPI communicator (unused, kept for API consistency)
```
castelletto1:
```suggestion
```

src/coreComponents/mesh/generators/VTKUtilities.cpp
```diff
@@ -830,7 +831,385 @@ scatterByBlock( vtkDataSet & mesh )
...
+  GEOS_MARK_FUNCTION;
+
+  array1d< int64_t > partitions2D( neighbors2Dto3D.size() );
+
+  for( localIndex i = 0; i < neighbors2Dto3D.size(); ++i )
+  {
+    auto neighbors = neighbors2Dto3D[i];
+
+    GEOS_ASSERT_MSG( neighbors.size() > 0,
+                     "2D cell should have at least one neighbor (enforced by build2DTo3DNeighbors)" );
```
castelletto1: ASSERT or ERROR? Is the check needed here? `build2DTo3DNeighbors` should catch this earlier.

src/coreComponents/mesh/generators/VTKUtilities.cpp
bd713: Added helper function.
Switched to shallowCopy and also removed the deep copies that separated 2D and 3D cells. Now rely on indices.

src/coreComponents/mesh/generators/VTKUtilities.cpp
bd713: Done

src/coreComponents/mesh/generators/VTKUtilities.cpp
bd713: Done

src/coreComponents/mesh/generators/VTKUtilities.cpp
bd713: Removed

src/coreComponents/mesh/generators/VTKUtilities.cpp
bd713: Limited gather operation to only the relevant indices

src/coreComponents/mesh/generators/VTKUtilities.cpp
bd713: Removed

castelletto1 (approved):

PR #3969, opened 2026-02-09
fix: SEQ Coupling for Multiphase Poromechanics

jhuang2601:
This PR aimed to 

- [ ] fix the SEQ Coupling for Multiphase Poromechanics

- [ ] enable full run with `subcycling="1"` for GNL+ALM: https://github.com/Total-RD/geos-dataset/pull/37

### DIFF
--- a/src/coreComponents/physicsSolvers/multiphysics/MultiphasePoromechanics.hpp
+++ b/src/coreComponents/physicsSolvers/multiphysics/MultiphasePoromechanics.hpp
@@ -107,6 +107,31 @@ protected:
       GEOS_ERROR( GEOS_FMT( "{}: MGR strategy is not implemented for {}", this->getName(), this->getCatalogName()));
   }
 
+  virtual void mapSolutionBetweenSolvers( DomainPartition & domain, integer const solverType ) override
+  {
+    GEOS_MARK_FUNCTION;
+
+    Base::mapSolutionBetweenSolvers( domain, solverType );
+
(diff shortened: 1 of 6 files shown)


src/coreComponents/physicsSolvers/multiphysics/MultiphasePoromechanics.hpp
```diff
@@ -107,6 +107,31 @@ class MultiphasePoromechanics : public PoromechanicsSolver< FLOW_SOLVER, MECHANI
...
+      this->template forDiscretizationOnMeshTargets<>( domain.getMeshBodies(), [&]( string const &,
+                                                                                    MeshLevel & mesh,
+                                                                                    string_array const & regionNames )
+      {
+
+        mesh.getElemManager().forElementSubRegions< CellElementSubRegion >( regionNames, [&]( localIndex const,
+                                                                                              auto & subRegion )
+        {
+          // update mass after porosity change due to mechanics solve
+          this->flowSolver()->updateMass( subRegion );
```
paveltomin: it does not work for multiphase the same way as for single phase, this change should not matter
what issue did you observe?

npillardou: For me this line is not compiling because updateMass() is not defined in CompositionalMultiphaseBase.
We should switch to updateCompAmount()

PR #3973, opened 2026-02-16
Fix thermal well perforation energy flux

npillardou:
This PR tries to fix a thermal well issue at top perforations.

According to the image below, for a single phase water injection, the temperature drops significantly lower than injection temperature (293.15K ) which is not logical for water injection case (left figure). 

image


tjb-ltk:
Hi Nicolas

Could you upload the associated model.   There is a restriction, probably not documented, where the top segment cannot contain a perforation.  
Thanks

tjb-ltk:
Could  you move/change the perf/seg definition so that the top seg doesn't have a perf and then make the comparison.  Currently it looks like the top seg has 2 perforations...

npillardou:
> Could you move/change the perf/seg definition so that the top seg doesn't have a perf and then make the comparison. Currently it looks like the top seg has 2 perforations...

Hello Thomas, 
I updated the xml file with perforation like you suggested. However, the same issue appear in the top layers. 

I also noticed a MGR issue when the injection starts from 0, I don't know if it is also your case : 
```
    <TableFunction
      name="injectorTotalRateTable"
      inputVarNames="{ time }"
      coordinates="{  0, 100.0e6 }"
      values="{ 1e-2, 1e-2 }" 
      interpolation="lower"/> 
```

npillardou:
Based on discussion with @tjb-ltk, I added an assertion when a perforation is placed on the top well element (wellhead).

dkachuma (commented):
This is quite a restriction. Almost all the well cases that I have created have a perforation in the first segment. Can we simply add a "dummy" segment for the boundary condition?

tjb-ltk:
@npillardou ...  Nicolas... could you add this fix to compositional ?  and then I'll put it behind a switch... we can make it default doesn't matter to me, as long as models can be tested with/without option...  @dkachuma we could add a segment internally,  if we were going to do that dev, I would propose we extend it to also include modifying segmentation to support one perf / seg.

PR #3977, opened 2026-02-20
feat: Change to local Newell's normal

jafranc:
This PR draft a proposal for switching from Newell's normal method based on global coordinates to local, _i.e._ from 
`\mathbf{n} = \sum_i P_i x P_(i+1)%n` to `\mathbf{n} = \sum_i (P_i - P_0) x (P_(i+1)%n - P_0)`.

This could avoid some precision loss when dealing with large coordinates. 

Tests are also more complete.


dkachuma (approved):
Looks good to me and indeed should have been included in the original implementation. [2280](https://github.com/GEOS-DEV/GEOS/pull/2280#issuecomment-1421016683)

castelletto1 (approved):

castelletto1:
@jafranc Please confirm that the failing integrated tests are behaving as expected, and verify the new baselines so we can proceed with merging.

jafranc:
> @jafranc Please confirm that the failing integrated tests are behaving as expected, and verify the new baselines so we can proceed with merging.

I am looking into that

jafranc:
> > @jafranc Please confirm that the failing integrated tests are behaving as expected, and verify the new baselines so we can proceed with merging.
> 
> I am looking into that

Most of the failed test are exhibiting 1e-9 diff, which I would expect but the `deadOilEgg`:
```
Error: /Problem/Solvers/compositionalMultiphaseWell/wellControls7/dCurrentBHP
	Arrays of types float64 and float64 have 4 values of which 1 fail both the relative and absolute tests.
		Max absolute difference is at index (np.int64(2),): value = -98.10000000000002, base_value = -0.0
		Max relative difference is at index (np.int64(2),): value = -98.10000000000002, base_value = -0.0
	Statistics of the q values greater than 1.0 defined by absolute tolerance: N = 1
		max = 98100000.00000003, mean = 24525000.000000007, std = 42478546.05562673
		max is at index (np.int64(2),), value = -98.10000000000002, base_value = -0.0
	Statistics of the q values greater than 1.0 defined by relative tolerance: N = 0
Error: /Problem/Solvers/compositionalMultiphaseWell/wellControls7/dCurrentTotalVolumetricRate
	Arrays of types float64 and float64 have 5 values of which 1 fail both the relative and absolute tests.
		Max absolute difference is at index (np.int64(1),): value = 1.6560123574322675e-06, base_value = -7.86199250195236e-06
		Max relative difference is at index (np.int64(1),): value = 1.6560123574322675e-06, base_value = -7.86199250195236e-06
	Statistics of the q values greater than 1.0 defined by absolute tolerance: N = 1
		max = 9.518004859384629, mean = 1.9036009718769258, std = 3.8072019437538516
		max is at index (np.int64(1),), value = 1.6560123574322675e-06, base_value = -7.86199250195236e-06
	Statistics of the q values greater than 1.0 defined by relative tolerance: N = 0
```

strangly enough when compared  proposed and `develop` normals:
```
  normal ( 0 1 0)                                                                   | normal ( 0 1 -5.78069e-16)                                                        
 normal ( -1 0 0)                                                                  | normal ( -1 0 0)                                                                  
 normal ( 0 0 -1)                                                                  | normal ( 0 0 -1)                                                                  
 normal ( 0 1 0)                                                                   | normal ( 0 1 -1.90883e-15)                                                        
 normal ( 1 0 0)                                                                   | normal ( 1 0 -2.40862e-15)                                                        
 normal ( 0 0 -1)                                                                  | normal ( 0 0 -1)                                                                  
 normal ( 0 1 0)                                                                   | normal ( 0 1 -2.75787e-15)                                                        
 normal ( 1 0 0)                                                                   | normal ( 1 0 0)                                                                   
 normal ( 0 0 -1)                                                                  | normal ( 0 0 -1)                                                                  
 normal ( 0 1 0)                                                                   | normal ( 0 1 2.6555e-15)                                                          
 normal ( 1 0 0)                                                                   | normal ( 1 0 -4.81724e-15)                                                        
 normal ( 0 0 -1)                                                                  | normal ( 0 0 -1)                                                                  
 normal ( 0 1 0)                                                                   | normal ( 0 1 6.6237e-15)                                                          
 normal ( 1 0 0)                                                                   | normal ( 1 0 9.63448e-15)                                                         
 normal ( 0 0 -1)                                                                  | normal ( 0 0 -1)                                                                  
 normal ( 0 1 0)                                                                   | normal ( 0 1 9.38759e-15)                                                         
 normal ( 1 0 0)                                                                   | normal ( 1 0 9.63448e-15)   
 ```
 
 so I am looking into upwinding or a normal conditions that could flip

**Comment by jafranc** [comment]
After discussion with @tjb-ltk, the discrepancy in baseline seems normally introduced by change of formula. I'll update baselines

**Comment by jafranc** [comment]
@OmarDuran @rrsettgast @jhuang2601 @bd713 can you have a look at it ?

File: src/coreComponents/mesh/unitTests/testComputationalGeometry.cpp [hunk 1]
```diff
@@ -70,4 +71,57 @@ TEST( testComputationalGeometry, checkCentroid3DPolygon )
...
+  const real64 EXPECTED_CENTER[3] = { 1.e6 + (xa+xb)/3, 1.e6 + (ya+yb)/3., 1.e6 + (za+zb)/3};
+
+  const real64 EXPECTED_NORMAL[3] = { (ya*zb-za*yb)/2/EXPECTED_AREA, (za*xb-xa*zb)/2/EXPECTED_AREA, (xa*yb-ya*xb)/2/EXPECTED_AREA};
+
+  EXPECT_LT( abs( faceArea - EXPECTED_AREA ), 1e-6 );
+  EXPECT_LT( abs( faceNormal[0]-EXPECTED_NORMAL[0] ), 1e-6 );
+  EXPECT_LT( abs( faceNormal[1]-EXPECTED_NORMAL[1] ), 1e-6 );
+  EXPECT_LT( abs( faceNormal[2]-EXPECTED_NORMAL[2] ), 1e-6 );
+  EXPECT_LT( abs( faceCenter[0]-EXPECTED_CENTER[0] ), 1e-6 );
+  EXPECT_LT( abs( faceCenter[0]-EXPECTED_CENTER[0] ), 1e-6 );
```
jhuang2601:
```suggestion

```
This is the duplicate of the above line

jhuang2601 (approved):

src/coreComponents/mesh/utilities/ComputationalGeometry.hpp
```diff
@@ -282,6 +286,12 @@ real64 centroid_3DPolygon( arraySlice1d< localIndex const > const pointsIndices,
   }
   else
   {
+    for( localIndex a=0; a<numberOfPoints; ++a )
+    {
+      GEOS_LOG_RANK( "Points: " << points[ pointsIndices[ a ] ] << " " << pointsIndices[ a ] );
+    }
+    GEOS_WARNING( "Null area found : " << area );
```
bd713: if the intent is to continue on null area, I would guard the warning with `#if !defined(GEOS_DEVICE_COMPILE)`, otherwise it will probably trap on gpu, like an error.

src/coreComponents/physicsSolvers/surfaceGeneration/SurfaceGenerator.cpp
```diff
@@ -40,6 +40,7 @@
 
 #include <algorithm>
 #include <deque>
+#include <limits>
```
bd713: is it needed in the end? I see 1e-108 is used eventually

bd713 (approved):

rrsettgast (approved):
It certainly should have been done like this in the first place.

one note is that for a quadrilateral (4 vertices) the answer is trivial:

A = 1/2 (x_3 - x_1) x ( x_4 - x_2 )

"A" is the face area vector, which is not equal to the actual area, but for purposes of fluxing across the surface, it is appropriate. 

Also, if you look at the original implementation before (https://github.com/GEOS-DEV/GEOS/pull/2280) which is more of a "fan Area" method, the normal calculation is equivalent to the Newell's method with the offset, but the area was more close to the actual surface area rather than the projected area that is more appropriate for FV fluxing. I wonder if you will get the same issues that prompted the change in the first place?

src/coreComponents/mesh/utilities/ComputationalGeometry.hpp
rrsettgast: null area should be an error.

src/coreComponents/physicsSolvers/surfaceGeneration/SurfaceGenerator.cpp
jafranc: Good catch !

src/coreComponents/mesh/utilities/ComputationalGeometry.hpp
jafranc: Gotcha ! I just added a warning where there was a silent pass. But fully on board for a proper ERROR :)
I amend it in this sense

PR #3978, opened 2026-02-20
fix: Separator flash fix

tjb-ltk:
Well separator used for constraint properties did not supply temperature when calling flash .

New baseline required for single phase fluid cases with well setting useSurfaceConditions="1"


tjb-ltk:
@dkachuma

Should the fluid model interface should throw if being called when model is thermal ???

src/coreComponents/physicsSolvers/fluidFlow/wells/SinglePhaseWell.cpp
```diff
@@ -282,11 +284,12 @@ void SinglePhaseWell::updateVolRateForConstraint( ElementRegionManager const & e
   string const wellControlsName = wellControls.getName();
   bool const logSurfaceCondition = isLogLevelActive< logInfo::WellControl >( wellControls.getLogLevel());
   integer const useSurfaceConditions = wellControls.useSurfaceConditions();
-  real64 flashPressure;
+  real64 flashPressure, flashTemperature;
   if( useSurfaceConditions )
   {
     // use surface conditions
     flashPressure = wellControls.getSurfacePressure();
+    flashTemperature = wellControls.getSurfaceTemperature();
```
dkachuma, line 293: Do we ensure that users supply this for single phase?

src/coreComponents/physicsSolvers/fluidFlow/wells/SinglePhaseWell.cpp
```diff
@@ -350,12 +357,19 @@ void SinglePhaseWell::updateVolRateForConstraint( ElementRegionManager const & e
         if( useSurfaceConditions )
         {
           // we need to compute the surface density
-          fluidWrapper.update( iwelemRef, 0, flashPressure );
+          if constexpr ( IS_THERMAL )
```
dkachuma: Since both fluid interfaces have the temperature dependent interface (the isothermal fluid simply ignores it), I would remove this branch and have a single call.

src/coreComponents/physicsSolvers/fluidFlow/wells/SinglePhaseWell.cpp
```diff
@@ -366,7 +380,15 @@ void SinglePhaseWell::updateVolRateForConstraint( ElementRegionManager const & e
         else
         {
           real64 const refPres = pres[iwelemRef];
-          fluidWrapper.update( iwelemRef, 0, refPres );
+          real64 const refTemp = temp[iwelemRef];
+          if constexpr ( IS_THERMAL )
```
dkachuma: Similarly here. No need for the branch.

dkachuma (approved):
Looks good. The fact that the thermal fluid has the interface without the temperature is strange. I think we can adopt the multiphase approach that we will always call with the temperature. Remove the interface without the temperature from the fluid models and see what fails to compile.

src/coreComponents/physicsSolvers/fluidFlow/wells/SinglePhaseWell.cpp
tjb-ltk: we do now

PR #3979, opened 2026-02-22
feat: Solution scaling for thermal single phase

frankfeifan:
This PR aims to enable scaling factor for temperature in the thermal single-phase flow and well models. Following changes are made:

- [x] Get and print out individual max. changes and scaling factors for the pressure and temperature
- [x] The `scalingFactor` output by `scalingForSystemSolution` is a global min of pressure and temperature scaling factors
- [x] Apply individual scaling factors to the pressure and temperature accordingly in both reservoir flow and well models

### DIFF
--- a/src/coreComponents/physicsSolvers/fluidFlow/wells/SinglePhaseWell.cpp
+++ b/src/coreComponents/physicsSolvers/fluidFlow/wells/SinglePhaseWell.cpp
@@ -1177,7 +1177,7 @@ SinglePhaseWell::applySystemSolution( DofManager const & dofManager,
   dofManager.addVectorToField( localSolution,
                                wellElementDofName(),
                                well::pressure::key(),
-                               scalingFactor,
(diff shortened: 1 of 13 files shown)


src/coreComponents/physicsSolvers/fluidFlow/wells/SinglePhaseWell.cpp
tjb-ltk, line 1: give the 1 and 2 a name... varOffset or something descriptor

tjb-ltk (approved):
Good addition and cleanup of single/compositional well scaling paramerters

src/coreComponents/physicsSolvers/fluidFlow/wells/SinglePhaseWell.cpp
frankfeifan, line 1187: Not sure which scaling factor should be used here for the `connectionRate`

PR #3980, opened 2026-02-23
feat: Add command-line option to control developper information output in diagnostic messages

MelReyCG:
This PR propose to introduce a new command-line option to control diagnostic log verbosity in GEOS. The goal is to reduce the verbosity of logs by default while still providing detailed developer information when needed.

New option:
```
-d, --diagnostic-info-level,  Diagnostic message information level:
                                0 = (default) basic information,
                                1 = developper information (source file, stack-trace) for errors only,
                                2 = developper information for warnings & errors
```

For an error, `-d 0` would give:
```
***** ExternalError
***** Rank 4
***** Message from Signal (detected from Signal Handler):
Floating point error encountered:
Unknown reason.
```

... and `-d 1` would give:
```
***** ExternalError
***** Rank 4
***** Message from Signal (detected from Signal Handler):
Floating point error encountered:
Unknown reason.

***** StackTrace of 15 frames
  - Frame  0:  /lib64/libc.so.6 
  - Frame  1:  void geos::isothermalCompositionalMultiphaseBaseKernels::internal::kernelLaunchSelectorCompSwitch<int, geos::isothermalCompositionalMultiphaseBaseKernels::GlobalComponentFractionKernelFactory::createAndLaunch<RAJA::policy::sequential::seq_exec>(int, geos::ObjectManagerBase&)::{lambda(auto:1)#1}>(int, geos::isothermalCompositionalMultiphaseBaseKernels::GlobalComponentFractionKernelFactory::createAndLaunch<RAJA::policy::sequential::seq_exec>(int, geos::ObjectManagerBase&)::{lambda(auto:1)#1}&&) 
  - Frame  2:  geos::CompositionalMultiphaseWell::updateGlobalComponentFraction(geos::WellElementSubRegion&) const 
[...]
  - Frame 11:  geos::GeosxState::run() 
  - Frame 12:  main 
  - Frame 13:  __libc_start_main 
  - Frame 14:  _start 
*****
```

The last option, `-d 2`, also enables the output of stack trace and source file for warnings.

Stack traces and source files are still included in the YAML output through the `-e` option, but now users can control the presence of developper information *in the log* through this new `-d` option.


src/coreComponents/common/logger/ErrorHandling.hpp
```diff
@@ -111,13 +112,28 @@ struct ErrorContext
...
   Error,
   ExternalError,
   Warning,
   Exception,
-  Undefined
+  Count // internal use, keep at last
+};
+
+/**
+ * @enum MsgType
```
arng40: You mean "DiagnosticInfoLevel"

src/coreComponents/common/logger/ErrorHandling.hpp
```diff
@@ -23,6 +23,7 @@
 #include "common/DataTypes.hpp"
 #include "common/format/Format.hpp"
 #include "common/format/StringUtilities.hpp"
+#include <array>
```
arng40: remove

src/coreComponents/common/logger/ErrorHandling.hpp
```diff
@@ -111,13 +112,28 @@ struct ErrorContext
...
+enum class MsgType : integer
 {
+  Undefined,
   Error,
   ExternalError,
   Warning,
   Exception,
-  Undefined
+  Count // internal use, keep at last
+};
```
arng40, line 123: A suggestion, `typeToString` may be redundant with `EnumString`.
You could use `EnumString ` instead of `typeToString` by moving this enum to a separate class.

src/coreComponents/common/logger/GeosExceptions.cpp
```diff
@@ -19,6 +19,7 @@
 
 
 #include "common/logger/GeosExceptions.hpp"
+#include "common/logger/ErrorHandling.hpp"
```
arng40: remove

src/coreComponents/common/logger/ErrorHandling.hpp
MelReyCG: I just renamed ``ErrorLogger::toString()``, but yes I can refactor if this function is not useful.
It is not in a class right?

MelReyCG: Ok, no I'm not able to do this with enum-strings. It is to avoid circular header dependancy.

arng40: Maybe put MsgType  in another file, it will avoid the circular dependency

MelReyCG: `MsgType` would depends on `EnumString`, but it depends on `Logger.hpp`, then on `ErrorHandler`
We should break that dependancy by removing logging of `EnumString`

arng40 (approved):

herve-gross (approved):

src/docs/sphinx/QuickStart.rst
```diff
@@ -350,7 +350,7 @@ We can do a quick check that the geosx executable is working properly by calling
 
   ./bin/geosx --help
 
-This should print out a brief summary of the available command line arguments:
+This should print out the available command line arguments with their description:
```
dkachuma, line 353:
```suggestion
This should print out the available command line arguments with their descriptions:
```

src/coreComponents/mainInterface/initialization.cpp
```diff
@@ -222,6 +228,12 @@ std::unique_ptr< CommandLineOptions > parseCommandLineOptions( int argc, char *
         commandLineOptions->onlyValidateInput = true;
       }
       break;
+      case DIAGNOSTIC_INFO_LEVEL:
+      {
+        DiagnosticInfoLevel infoLevel = DiagnosticInfoLevel( std::stoi( opt.arg ) );
```
dkachuma, line 233: Not sure how these work but do we need to check in case the user gives something like `-d -3`?

dkachuma: Just seen how it's implemented. Ignore this comment.

dkachuma (approved):

src/coreComponents/common/logger/ErrorHandling.hpp
```diff
@@ -322,6 +340,14 @@ class ErrorLogger
   void setOutputFilename( std::string_view filename )
   { m_filename = filename; }
 
+  /**
+   * @brief Set the diagnostic messages information level. The higher, the more verbose and
+   *        developper-oriented the messages will be.
```
kdrienCG:
```suggestion
   *        developer-oriented the messages will be.
```

src/coreComponents/common/logger/ErrorHandling.hpp
```diff
@@ -407,6 +442,8 @@ class ErrorLogger
   std::mutex m_errorHandlerAsciiMutex;
   /// Avoid concurrent access between threads for yaml outputs
   std::mutex m_errorHandlerYamlMutex;
+  /// Indicated if the source file information output is enabled for a given message type
+  stdArray< bool, size_t( MsgType::Count ) > m_msgTypeSourceInfoEnabled;
```
kdrienCG: (Might not be a suggestion/review but more like a question)
Here `m_msgTypeSourceInfoEnabled` is correctly initialized in the `setDiagnosticInfoLevel( DiagnosticInfoLevel::Basic )` call in the `ErrorLogger()` constructor. 
But if someone modifies the code and for some reason `setDiagnosticInfoLevel( ... )` is called too late, or not called at all (through another constructor? or only in an `if` condition? or something else?) `m_msgTypeSourceInfoEnabled` will be left uninitialized and produce UB (maybe using garbage booleans producing unexpected results).

Value initialization with empty braces (where `bool`s are set to `false` by default when value initializing) could prevent UB, and always produce the same result if this theoretical unhappy mentioned earlier path is run.
```suggestion
  stdArray< bool, size_t( MsgType::Count ) > m_msgTypeSourceInfoEnabled {};
```

I've still not come across value initialization in GEOS so maybe there is a reason for it?

MelReyCG: Indeed a good catch, ``std::array`` are not automatically initialized, and there could be other code paths later.

kdrienCG (approved):

MelReyCG:
@herve-gross @bd713 What do you think of enabling the error.yaml by default, and keeping the log not verbose (`-d 0` default)

victorapm:
Needs code owner review from @OmarDuran , @bd713 and/or @rrsettgast

MelReyCG:
Need to be sure that:
- the stack-trace is still output by default somewhere (yaml enabled by default OR stack-trace in log by default),
- we are able to run on a restart and re-enable that option (are commandline-option mutable between restart).

Thanks for remarks @ dev-meeting :)

PR #3987, opened 2026-03-02
feat: WHP constraint  

tjb-ltk:
Wellhead pressure constraint option


herve-gross (approved):
Hefty commit! I am approving just to not hold it up, but a thorough review and test would be needed.

dkachuma (commented):
I think we should merge the other PRs first. Just to isolate the schema changes from the actual functional WHP changes.

tjb-ltk:
We still need to validate with compositional VFP run,  so a far bit away from merging..

PR #3994, opened 2026-03-12
feat: inner products for mixed method

chauj96:
This PR adds the implementation of `computeM`, which computes the inner product matrix $M$ used in the MFD mixed method formulation. 

1) The function has been implemented for the following inner product types:
- `TPFAInnerProduct`
- `BdVLMInnerProduct`
- `QuasiTPFAInnerProduct`
- `SimpleInnerProduct`
Each class now includes a `computeM` routine that construct the inner product matrix $M$ required for the mixed formulation.

2) Testing
Unit tests were added in `testMimeticInnerProducts` to verify the correctness of the newly implemented `computeM` functions. All tests currently pass.

3) The centroid computation has been updated to evaluate the exact volume of a polyhedron with planar polygonal facets. This property is verified by a unit test. This induces a rebaseline.

+) Minor variable naming updates were applied in `SimpleInnerProduct` to keep the geometry nomenclature consistent with the other inner product implemenations.


src/coreComponents/finiteVolume/mimeticInnerProducts/TPFAInnerProduct.hpp
```diff
@@ -129,6 +140,68 @@ TPFAInnerProduct::compute( arrayView2d< real64 const, nodes::REFERENCE_POSITION_
...
+  real64 const weightTolerance = 1e-30 * lengthTolerance;
+
+  // initialize M to zero
+  for( localIndex i = 0; i < NF; ++i )
+  {
+    for( localIndex j = 0; j < NF; ++j )
+    {
+      M[i][j] = 0.0;
+    }
+  }
```
castelletto1:
```suggestion
LvArray::tensorOps::fill< NF, NF >( M, 0.0 );
```

src/coreComponents/finiteVolume/mimeticInnerProducts/SimpleInnerProduct.hpp
```diff
@@ -195,6 +209,121 @@ SimpleInnerProduct::compute( arrayView2d< real64 const, nodes::REFERENCE_POSITIO
...
+  real64 tParam = 2.0 * ( elemPerm[0] + elemPerm[1] + elemPerm[2] );
+  real64 scale = elemVolume / tParam;
+
+  for( localIndex i = 0; i < NF; ++i )
+  {
+    for( localIndex j = 0; j < NF; ++j )
+    {
+      M[i][j] = CKCt[i][j] + scale * Usc[i][j];
+    }
+  }
```
castelletto1:
```suggestion
  // 4) M = CKCt + (v / t) * A^{-1} * ( I - Q Q^T ) * A^{-1}
  real64 invA[NF];
  for (localIndex i = 0; i < NF; ++i)
  {
    invA[i] = real64(1) / A[i];
  }
  
  // scale = elemVolume / tParam, where tParam = 2 * trace(K)
  real64 const tParam = real64(2) * (elemPerm[0] + elemPerm[1] + elemPerm[2]);
  real64 const scale  = elemVolume / tParam;
  
  for (localIndex i = 0; i < NF; ++i)
  {
    real64 const invAiScaled = scale * invA[i];

    real64 const qi0 = Qmat[i][0];
    real64 const qi1 = Qmat[i][1];
    real64 const qi2 = Qmat[i][2];
  
    for (localIndex j = i; j < NF; ++j)
    {
      // qdot = (Q Q^T)_{ij} = sum_k Qik * Qjk, with k in [0,2]
      real64 const qdot = qi0*Qmat[j][0] + qi1*Qmat[j][1] + qi2*Qmat[j][2];
  
      real64 const u = (i == j ? real64(1) : real64(0)) - qdot;
  
      real64 const mij = CKCt[i][j] + (invAiScaled * invA[j]) * u;
  
      M[i][j] = mij;
      M[j][i] = mij; // symmetry fill
    }
  }
```

src/coreComponents/finiteVolume/mimeticInnerProducts/SimpleInnerProduct.hpp
```diff
@@ -195,6 +209,121 @@ SimpleInnerProduct::compute( arrayView2d< real64 const, nodes::REFERENCE_POSITIO
...
+
+  // 1) Compute C, N, A
+  real64 C[ NF ][ 3 ] = {{ 0 }};
+  real64 N[ NF ][ 3 ] = {{ 0 }};
+  real64 A[ NF ] = { 0.0 };
+
+  for( localIndex i = 0; i < NF; ++i )
+  {
+    real64 fCenter[3], fNormal[3];
```
castelletto1: Let's use a consistent nomenclature across inner products, e.g.:

- `fCenter` vs `faceCenter`
- `fNormal` vs `faceNormal`
- `vec_cf` vs `cellToFaceVector`
- ...

Let's keep in mind when refactoring nomenclature, that kernels will eventually be used also for flow within surfaces (faults, fractures). Cell seems appropriate. Instead of faces (or edges), we should probably use the term facets.

src/coreComponents/finiteVolume/mimeticInnerProducts/TPFAInnerProduct.hpp
chauj96: Thank you, I’ve checked the changes!

src/coreComponents/finiteVolume/mimeticInnerProducts/SimpleInnerProduct.hpp
chauj96: I have updated the naming in 'SimpleInnerProduct' to keep it consistent with TPFA/QuasiTPFA. Thank you for the feedback!

OmarDuran (approved):
Thank you @chauj96, nice work.

bd713 (approved):
Let's update the PR description to clearly explain that the change in centroid computation is the reason a rebaseline is needed.

src/coreComponents/mesh/utilities/ComputationalGeometry.hpp
```diff
@@ -271,6 +271,24 @@ real64 centroid_3DPolygon( arraySlice1d< localIndex const > const pointsIndices,
   {
     LvArray::tensorOps::normalize< 3 >( normal );
     area *= 0.5;
+
```
jafranc, line 289: Can you refactor that above, that would avoid duplication as you used almost the same steps.

src/coreComponents/finiteVolume/mimeticInnerProducts/BdVLMInnerProduct.hpp
```diff
@@ -112,18 +138,13 @@ BdVLMInnerProduct::compute( arrayView2d< real64 const, nodes::REFERENCE_POSITION
                                                  faceNormal,
                                                  areaTolerance );
 
-    LvArray::tensorOps::copy< 3 >( cellToFaceVec, faceCenter );
-    LvArray::tensorOps::subtract< 3 >( cellToFaceVec, elemCenter );
+    MimeticInnerProductHelpers::computeCellToFacetVector( cellToFaceVec, faceCenter, elemCenter );
```
jafranc, line 141: Maybe those two should be moved in `ComputationalGeometry` header so we can factorize use later

src/coreComponents/finiteVolume/mimeticInnerProducts/QuasiTPFAInnerProduct.hpp
```diff
@@ -90,6 +115,125 @@ QuasiTPFAInnerProduct::compute( arrayView2d< real64 const, nodes::REFERENCE_POSI
                                                                 transMatrix );
 }
 
+template< localIndex NF >
+GEOS_HOST_DEVICE
+void
+QuasiTPFAInnerProduct::computeM( arrayView2d< real64 const, nodes::REFERENCE_POSITION_USD > const & nodePosition,
```
jafranc, line 121: Looks like a great refactor, I am unclear on why it is not called in `compute` and mostly duplicated from it.

jafranc (changes requested):
Great work on rewriting as inner products rather than transmissivities.

Although I am unclear on why it is not called in `compute` replacing the old way and just duplicated from it up to the transposition.

src/coreComponents/finiteVolume/mimeticInnerProducts/BdVLMInnerProduct.hpp
OmarDuran: These are helpers for mimetic inner product constructions. While they are pure geometry operations, I’m not sure computeCellToFacetVector has other uses in GEOs beyond mimetic operators. Le's keep it here.

src/coreComponents/finiteVolume/mimeticInnerProducts/QuasiTPFAInnerProduct.hpp
OmarDuran: In general M and T exhibit some duality in the sense of M = T^{-1}. However there is no cheap map from T to M for the non-diagonal families, including QuasiTPFA construction.

src/coreComponents/mesh/utilities/ComputationalGeometry.hpp
OmarDuran: Agreed, this is a good point. The suggested refactor is now included.

OmarDuran:
> Great work on rewriting as inner products rather than transmissivities.
> 
> Although I am unclear on why it is not called in `compute` replacing the old way and just duplicated from it up to the transposition.

Thanks! The reason `computeM` doesn't replace the body of `compute` is that they build different matrices: `compute` assembles the inverse inner product W = (N K Nᵀ + t·P_C diag(N K Nᵀ) P_C)/V for the hybrid (flux-elimination) scheme, while `computeM` assembles the primal M = C K⁻¹ Cᵀ/V + (V/t)·P_N diag(N K Nᵀ)⁻¹ P_N for the mixed saddle-point formulation, where fluxes stay unknowns. The dual structure of M↔W (C↔N, K↔K⁻¹, t↔1/t) exists by consistency construction, but the stabilizations differ, so in general M ≠ W⁻¹, if we want to call one from the other would need a dense per-element inversion in-kernel. The actual shared part is now factored into `computeCellToFaceGeometry` (C, N, face areas) and `computeConsistencyTerm` (C K⁻¹ Cᵀ/V).

jafranc (approved):

jafranc:
Great refactor 👍

PR #3995, opened 2026-03-13
fix: correct three phases hydro. equilibrium with no capillarity.

OmarDuran:
This PR improves the logic  for the hydrostatic equilibrium with no capillarity. For the condition without a capillarity model, there are no $P_c$ curves to invert, and the compositions from the user's elevation tables are preserved. 

An integrated test is included to verify that hydrostatic equilibrium is a solution of the PDE under compatible boundary conditions without sources or wells. The test asserts that the computed equilibrium values remain unchanged throughout the simulation.

### DIFF
--- a/src/coreComponents/integrationTests/fluidFlowTests/testFixedDimensionalHydrostaticEquilibriumFixture.hpp
+++ b/src/coreComponents/integrationTests/fluidFlowTests/testFixedDimensionalHydrostaticEquilibriumFixture.hpp
@@ -1,0 +1,322 @@
+/*
+ * ------------------------------------------------------------------------------------------------------------
+ * SPDX-License-Identifier: LGPL-2.1-only
+ *
+ * Copyright (c) 2016-2024 Lawrence Livermore National Security LLC
+ * Copyright (c) 2018-2024 TotalEnergies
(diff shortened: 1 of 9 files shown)


src/coreComponents/integrationTests/fluidFlowTests/testFixedDimensionalHydrostaticEquilibriumFixture.hpp
```diff
@@ -0,0 +1,322 @@
...
+          subRegion.getField< fields::flow::compAmount_n >();
+
+        integer const numComp = compAmount.size( 1 );
+
+        for( localIndex k = 0; k < subRegion.size(); ++k )
+        {
+          for( integer ic = 0; ic < numComp; ++ic )
+          {
+            real64 const accumulation = std::fabs( compAmount[k][ic] - compAmount_n[k][ic] );
+            EXPECT_NEAR( accumulation, 0.0, accumulation_tolerance )
```
jafranc:
```suggestion
            real64 const accumulation = compAmount[k][ic] - compAmount_n[k][ic];
            EXPECT_NEAR( accumulation, 0.0, accumulation_tolerance )
```
I think `EXPECT_NEAR` is handling the absolute valure logic

rrsettgast (approved):

PR #3996, opened 2026-03-14
feat: Change molarity concentration to molality concentration

frankfeifan:
This PR changes convention of the concentration fed into `HPCReact` from molarity **[mol/m3 of solution]** or **[mol/L of solution]** to molality **[mol/kg of solvent]** using a factor called `solventMassPerSolutionVolume` (not `solventDensity` anymore) to be consistent with convention. 

This is a preparatory step for introducing the [activity model](https://github.com/GEOS-DEV/HPCReact/pull/17). So now, `HPCReact` is treated as a strictly molality-based library, and every conversion to a volumetric basis happens on the GEOS side through a single factor.

Specific changes include:

- [x] Switched the concentration passed to HPCReact from molarity to molality, converting back at every caller: accumulation, advective and diffusive flux, Dirichlet and source terms, and their thermal counterparts.

- [x] Introduced `solventMassFraction` as an optional input on the reactive fluid model (default 1, validated to lie in (0, 1]). HPCReact is molality-based, so species amounts are converted with `molarity = molality * solventMassFraction * fluidDensity`, where the density comes from the single fluid model and so already carries the pressure and temperature dependence. For now the fraction is prescribed and kept constant. Ideally it follows the chemical concentration, which thus needs species molar masses. Where that update would be computed and launched (GEOS or HPCReact) needs more discussion and deserves another PR.

- [x] Sourced the `solventMassFraction ` from the solver's single reactive fluid model rather than a per-subregion scan, and turned the previously silent single-fluid-model assumption into an explicit input error, ensuring a single fluid model defined across all regions for a uniform dof layout.

- [x] Restored the finite-difference Jacobian test to `dt = 1.0`, since the reduction to 0.01 was a workaround for the inflated reaction term.

- [x] Rebaseline and added 1D serial chain reaction into the integrated test set

Note: No `HPCReact` change is required, as the conversion factor lives in GEOS now tentatively.


dkachuma (approved):

jafranc (approved):
Looks great 👍

tjb-ltk (approved):
A future discussion could consider posing this as multicomponent single phase option where specializations could be this could be this option or something like tracers ..  not sure about the implications when starting with a num component generalization...

frankfeifan:
> A future discussion could consider posing this as multicomponent single phase option where specializations could be this could be this option or something like tracers .. not sure about the implications when starting with a num component generalization...

Yeah. That's a good point. We may need a higher-level generic solver base for such multi-component single phase option (can inherit from SinglePhaseBase as I did for this reactive transport solver).

src/coreComponents/constitutive/fluid/reactivefluid/ReactiveSinglePhaseFluid.cpp
```diff
@@ -31,6 +31,9 @@ namespace reactivefluid
 {
 
 using namespace hpcReact::bulkGeneric;
+using namespace hpcReact::geochemistry;
+using namespace hpcReact::ChainGeneric;
+using namespace hpcReact::MoMasBenchmark;
```
OmarDuran: Double check if we actually need this.

```suggestion

```

OmarDuran (commented):
@frankfeifan thank you for taking care of computing actual mass balance. 

I might be mistaken, but only with constant ρs, the conversions are fully correct. I’m unsure whether they remain consistent for a compressible fluid.

I also refactored a few files to accommodate these changes. See the patch below; if it makes sense, feel free to apply it.

[molality_solventMassFraction.patch](https://github.com/user-attachments/files/32321203/molality_solventMassFraction.patch)

frankfeifan:
> @frankfeifan thank you for taking care of computing actual mass balance.
> 
> I might be mistaken, but only with constant ρs, the conversions are fully correct. I’m unsure whether they remain consistent for a compressible fluid.
> 
> I also refactored a few files to accommodate these changes. See the patch below; if it makes sense, feel free to apply it.
> 
> [molality_solventMassFraction.patch](https://github.com/user-attachments/files/32321203/molality_solventMassFraction.patch)

Thanks, @OmarDuran! Yes, you're right. My changes for now considers a constant ρs for simplicity, as the main goal is to ensure we convert the concentration to the right unit fed into HPCReact. That said, I took a look of the patch, and I think it's a genius idea and exactly aligned with my TODO in the code. Eventually, we may also want to allow this `solventMassFraction` to vary based on the dissolved species mass, but that can be done in a future PR. This is a good intermediate step and not that complicated at this moment, so I've applied this. Thank you very much!

OmarDuran (approved):

PR #3997, opened 2026-03-16
fix: LogPart length

arng40:
Fix a length bug where when we have a description `>100 characters`  the logPart maxLength is not taken into account


src/coreComponents/common/format/LogPart.cpp
```diff
@@ -61,7 +61,7 @@ void LogPart::setMaxWidth( size_t const & maxWidth )
 
 double clamp( double v, double min, double max )
 {
-  return std::min( max, std::max( min, v ));
+  return LvArray::math::min( max, LvArray::math::max( min, v ));
```
MelReyCG, line 64: Thanks for this change, could be added to the automatic code rules and the docs. There's some more in log-parts.

src/coreComponents/common/format/LogPart.cpp
```diff
@@ -70,60 +70,59 @@ void LogPart::formatDescriptions( LogPart::Description & description,
   stdVector< string > & formattedLines = formattedDescription.m_lines;
   size_t const borderSpaceWidth = m_nbBorderChar * 2 + m_borderMargin * 2;
 
-  size_t const formattingCharSize = borderSpaceWidth;
+  size_t const decorationWidth = borderSpaceWidth;
```
MelReyCG, line 73: Names are better in your PR, indeed.

MelReyCG (approved):

PR #4000, opened 2026-03-17
feat: Prevent ghost elements from modifying their own data in contact solvers

bd713:
WIP
**Objective**
Prevent ghost elements from modifying their own data in contact solvers
Instead, only receive updates via proper synchronization.


rrsettgast:
@bd713 were these getting update with incorrect values? I ask because ghost elements are allowed to update their stress state since it derives from the displacement field which is synchronized?

bd713:
@rrsettgast No. It's more about having consistent ghost handling across the solver, than fixing a known bug.

PR #3332, opened 2024-09-05
docs: edits from joss review

rrsettgast:
- added links to input file for Tutorial 1
- added reference/link XML schema in Tutorial 1
- modified documentation for `NumericalMethods` section in Tutorial 1
- added note on future availability of permeability as a full tensor.
- added calls to `basicCleanup()` in `main()` for cases where it would be skipped.
- added note on the `Mandel` example for using the python plotting script.

PR #4001, opened 2026-03-18
fix: Infinite hang when MPI process crashes

arng40:
The bug occurs when you launch a simulation with multiple ranks and one of the ranks calls a GEOS_THROW. This results in an infinite hang


MelReyCG (approved):

rrsettgast:
The second call to `basicCleanup()` was done intentionally in:
https://github.com/GEOS-DEV/GEOS/pull/3332

to avoid cleanup issues. Are you sure this doesn't reintroduce the issues that it fixed?

MelReyCG (changes requested):
Ok, so let's change the strategy:
- separate anything that is not about cleaning up from `basicCleanup()` to a new function (`finalizeRun()`? something else?),
- ensure to call `basicCleanup()` in any scenario, to ensure it is called, but any logging of a potencial error must be done before.

src/coreComponents/common/MpiWrapper.cpp
```diff
@@ -152,11 +153,19 @@ void internal::ManagedResources::finalize()
...
+  {
+    MpiWrapper::commFree( MPI_COMM_GEOS );
+    internal::getManagedResources().finalize();
+    MPI_CHECK_ERROR( MPI_Finalize() );
+  }
+  else
+  {
+    LvArray::system::callErrorHandler();
+  }
 #endif
```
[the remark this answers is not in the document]
MelReyCG: @arng40

src/coreComponents/common/MpiWrapper.cpp
```diff
@@ -152,11 +153,19 @@ void internal::ManagedResources::finalize()
...
-  MPI_CHECK_ERROR( MPI_Finalize() );
+  if( !inError )
+  {
+    MpiWrapper::commFree( MPI_COMM_GEOS );
+    internal::getManagedResources().finalize();
+    MPI_CHECK_ERROR( MPI_Finalize() );
+  }
+  else
+  {
+    LvArray::system::callErrorHandler();
```
MelReyCG: It seems very strange that the MpiWrapper is responsible of the termination of the program.
Consider relocating.

src/main/main.cpp
```diff
@@ -71,15 +74,14 @@ int main( int argc, char *argv[] )
...
     return 0;
   }
   catch( geos::Exception & e )
   { // GEOS generated exceptions management
     ErrorLogger::global().flushCurrentExceptionMessage();
-    basicCleanup();
-    // lvarray error handler is just program termination
-    LvArray::system::callErrorHandler();
+    basicCleanup( true );
+    std::abort();
```
MelReyCG: Why is it not the LvArray error handler that terminates the program here?

MelReyCG (changes requested):

MelReyCG (approved):

MelReyCG (approved):

jafranc (approved):

PR #4006, opened 2026-03-27
feat: Add MGR strategy for thermal single-phase flow with wells and poromechanics

victorapm:
cc @castelletto1 @tjb-ltk @frankfeifan 

Adds MGR strategy and fixes a formulation bug.

The FIM Jacobian was missing well–reservoir perforation couplings. That is now fixed in this PR. FIM `assembleSystem` already added perforation residuals (mass and energy), but `setupSystem` built the matrix pattern only from `DofManager`. Perforation off-diagonals are not in the `DofManager` graph; sequential flow+wells adds them in `CoupledReservoirAndWellsBase::setSparsityPattern`. In RelWithDebInfo, `addToRowBinarySearchUnsorted` can drop those Jacobian entries.

The well energy residual was assembled without a matching Jacobian, so the first Newton update blew it up and FIM looked unstable.

`PoromechanicsSolver` now takes the flow solver’s sparsity pattern, the same approach `PoromechanicsConformingFractures` already used. For `ReservoirAndWells` that includes perforation couplings; for standalone flow it is the same DofManager pattern as before.


frankfeifan:
> cc @castelletto1 @tjb-ltk @frankfeifan
> 
> Would you be to help by adding a regression test? Thanks!

Thanks, Victor! I can add one based on the sequential case I made for thermoporomechanics reservoir solver.

victorapm:
Thanks, Frank! Feel free to push to this PR

frankfeifan:
> Thanks, Frank! Feel free to push to this PR

Hey Victor @victorapm, I've added a fim case for `ReservoirThermoPoroElastic` and moved both fim and sequential cases to `thermoPoromechanics` folder. I don't see critical issue in the linear solver for the fim case, and it runs well. The log file is attached. 
[output_fim.log](https://github.com/user-attachments/files/26325864/output_fim.log)

However, I found the solutions from the fim and sequential are quite different, e.g., see `deltaPressure` below (left is fim and right is sequential). And I see the fim case shows stability issue at early time. Maybe it can be mitigated by tuning some parameters in xml or mgr recipe? Please feel free to take a look.  

Screenshot 2026-03-28 at 1 09 41 PM

victorapm:
Thanks Frank! Maybe it's an issue with the FIM formulation? @tjb-ltk do you know if all the Jacobian components are being computed correctly?

tjb-ltk:
@frankfeifan @victorapm I'll run the cases and check the well behavior.

PR #4008, opened 2026-03-30
fix: Fracture/3D cell co-location in parallel mesh redistribution

bd713:
Replaces #3985

### Description
Similarly to 2D elements (see #3967), when VTK meshes with fractures are redistributed across MPI ranks, fracture elements must stay co-located with their neighboring 3D cells.

In practice, this PR introduces a super-cell partitioning strategy: cells connected through a fracture are grouped into super-cells that are treated as atomic units during partitioning. 
ParMETIS then operates on the super-cell graph instead of the individual cell graph, guaranteeing co-location.

### Rebaseline needed
The changed fracture-to-cell colocation logic alters internal index arrays and adds a new partitionFractureWeight field.


dkachuma (approved):

src/coreComponents/integrationTests/meshTests/testVTKImport.cpp
```diff
@@ -229,7 +250,7 @@ class TestFractureImport : public ::testing::Test
       fracture->Allocate( numQuads );
       for( vtkIdType const * q: quad )
       {
-        fracture->InsertNextCell( VTK_QUAD, numPoints, q );
+        fracture->InsertNextCell( VTK_QUAD, 4, q );
```
jafranc: isn't it weird to hardcode there and still use `numPoints` above ?

src/coreComponents/mesh/generators/ParMETISInterface.cpp
```diff
@@ -147,5 +147,61 @@ partition( ArrayOfArraysView< idx_t const, idx_t > const & graph,
...
+                   int const numRefinements )
+{
+  array1d< idx_t > part( graph.size() );
+  if( numParts == 1 )
+  {
+    return part;
+  }
+
+  array1d< real_t > tpwgts( numParts );
+  tpwgts.setValues< serialPolicy >( 1.0f / static_cast< real_t >( numParts ) );
```
jafranc, line 169: isn't parmetis dividing in equal weights if `NULL` is pass as tpwgts ? If so, it might avoid a check for zero-div here

src/coreComponents/mesh/generators/CollocatedNodes.cpp
```diff
@@ -24,29 +24,51 @@ namespace geos::vtk
...
+      GEOS_LOG_RANK_0( GEOS_FMT( "Available point data fields in '{}':", faceBlockName ) );
+      for( int i = 0; i < faceMesh->GetPointData()->GetNumberOfArrays(); ++i )
+      {
+        GEOS_LOG_RANK_0( GEOS_FMT( " - {} of type '{}'",
+                                   faceMesh->GetPointData()->GetArrayName( i ),
+                                   faceMesh->GetPointData()->GetArray( i )->GetDataTypeAsString() ) );
+      }
+      GEOS_ERROR( GEOS_FMT( "Could not find valid field \"{}\" for fracture \"{}\" on this rank.",
+                            COLLOCATED_NODES,
+                            faceBlockName ) );
```
jafranc:
```suggestion
CollocatedNodes::CollocatedNodes( string const & faceBlockName,
                                  vtkSmartPointer< vtkDataSet > faceMesh,
                                  bool performGlobalCheck )
{
  string const COLLOCATED_NODES = "collocated_nodes";

  vtkIdTypeArray const * collocatedNodes = vtkIdTypeArray::FastDownCast(
    faceMesh->GetPointData()->GetArray( COLLOCATED_NODES.c_str() ) );

  // Determine whether the field is missing:
  //   -  if globalCheck: reduce across ranks so an empty-mesh detection is any ranks
  bool const isMissing = performGlobalCheck
    ? ( MpiWrapper::max( reinterpret_cast< std::uintptr_t >( collocatedNodes ) ) == 0 )
    : ( collocatedNodes == nullptr );

  if( isMissing )
  {
    GEOS_LOG_RANK_0( GEOS_FMT( "Available point data fields in '{}':", faceBlockName ) );
    for( int i = 0; i < faceMesh->GetPointData()->GetNumberOfArrays(); ++i )
    {
      GEOS_LOG_RANK_0( GEOS_FMT( " - {} of type '{}'",
                                 faceMesh->GetPointData()->GetArrayName( i ),
                                 faceMesh->GetPointData()->GetArray( i )->GetDataTypeAsString() ) );
    }
    GEOS_ERROR( GEOS_FMT( "Could not find valid field \"{}\" for fracture \"{}\"{}.",
                          COLLOCATED_NODES,
                          faceBlockName,
                          performGlobalCheck ? "" : " on this rank" ) );
  }
  else
    init( collocatedNodes );
  }
  
```

src/coreComponents/mesh/generators/ParMETISInterface.cpp
```diff
@@ -147,5 +147,61 @@ partition( ArrayOfArraysView< idx_t const, idx_t > const & graph,
...
+
+  idx_t wgtflag = 2;  // vertex weights only
+  idx_t numflag = 0;
+  idx_t ncon = 1;
+  idx_t npart = numParts;
+
+  // Options: [use_defaults, log_level, seed, coupling]
+  // PARMETIS_PSR_UNCOUPLED = 0 (default - uses PartitionSmallGraph)
+  // PARMETIS_PSR_COUPLED = 1 (forces distributed algorithm)
+  idx_t options[4] = { 1, 0, 2022, 0 };
```
jafranc:
```suggestion
  idx_t options[4] = { 1, 0, 2022, PARMETIS_PSR_UNCOUPLED };
```

src/coreComponents/mesh/generators/VTKUtilities.cpp
```diff
@@ -237,59 +238,38 @@ vtkSmartPointer< vtkCellArray > getCellArray( vtkSmartPointer< vtkDataSet > mesh
...
+  // Create strictly allocated ArrayOfArrays using resizeFromCapacities
   ArrayOfArrays< INDEX_TYPE, INDEX_TYPE > elemToNodes;
   elemToNodes.template resizeFromCapacities< parallelHostPolicy >( numCells, nodeCounts.data() );
 
-  vtkIdTypeArray const & globalPointId = *vtkIdTypeArray::FastDownCast( meshes.getMainMesh()->GetPointData()->GetGlobalIds() );
+  vtkIdTypeArray const & globalPointId = *vtkIdTypeArray::FastDownCast( mesh->GetPointData()->GetGlobalIds() );
 
   // GetCellAtId() is conditionally thread-safe, use POLICY argument
-  forAll< POLICY >( num3dCells, [&cells, &globalPointId, elemToNodes = elemToNodes.toView()] ( localIndex const cellIdx )
+  forAll< POLICY >( numCells, [&cells, &globalPointId, elemToNodes = elemToNodes.toView()] ( localIndex const cellIdx )
```
jafranc: why aren't we using `GetCellPoints()` here ?
https://vtk.org/doc/nightly/html/classvtkDataSet.html#ab4d677c257a58e4eb4a80757f9b371ea

jafranc: This might the need for passing `cells` through the interface and the `IsShareable` logic

src/coreComponents/mesh/generators/VTKUtilities.cpp
```diff
@@ -1225,29 +1396,294 @@ redistribute2DAndMergeWith3D( vtkSmartPointer< vtkDataSet > redistributed3D,
...
+    vtkIdType const origCellIdx = cells3DIndices[i];
+    vtkCell * cell = originalMesh.GetCell( origCellIdx );
+    vtkIdList * pointIds = cell->GetPointIds();
+
+    for( vtkIdType p = 0; p < pointIds->GetNumberOfIds(); ++p )
+    {
+      vtkIdType const nodeLocalId = pointIds->GetId( p );
+      vtkIdType const nodeGlobalId = static_cast< vtkIdType >( meshGlobalNodeIds->GetTuple1( nodeLocalId ) );
+
+      nodeGlobalIdToCells3D.get_inserted( nodeGlobalId ).insert( origCellIdx );
```
jafranc, line 1582: isn't it `CellPoints()` again ?

src/coreComponents/mesh/generators/VTKSuperCellPartitioning.cpp
```diff
@@ -0,0 +1,1052 @@
...
+    }
+  }
+
+  GEOS_LOG_RANK_0( GEOS_FMT( "Built fracture graph: {} 3D cells connected via {} fracture elements",
+                             fractureGraph.size(), totalFractureElements ) );
+
+  // -----------------------------------------------------------------------
+  // Step 2: Find the maximum global ID from ALL cells (not just fracture cells)
+  // -----------------------------------------------------------------------
+  vtkIdType maxGlobalId = 0;
```
jafranc: `GetMaxId()` ?

src/coreComponents/mesh/generators/VTKSuperCellPartitioning.cpp
```diff
@@ -0,0 +1,1052 @@
...
+        }
+      }
+    }
+  }
+
+  // -----------------------------------------------------------------------
+  // Step 6: Symmetrize
+  // -----------------------------------------------------------------------
+  // Collect ALL edges to send (to any rank)
+  std::vector< pmet_idx_t > myEdgesSrc;
```
jafranc: any particular reason to have all these `myVar` convention ?

src/coreComponents/mesh/generators/VTKSuperCellPartitioning.cpp
```diff
@@ -0,0 +1,1052 @@
...
+  array1d< int > allEdgeCounts( numRanks );
+  MpiWrapper::allgather( &myEdgeCount, 1, allEdgeCounts.data(), 1, comm );
+
+  // Compute displacements for symmetrization
+  array1d< int > edgeDispls( numRanks + 1 );
+  edgeDispls[0] = 0;
+  for( int r = 0; r < numRanks; ++r )
+  {
+    edgeDispls[r + 1] = edgeDispls[r] + allEdgeCounts[r];
+  }
```
jafranc:
```suggestion
  int myEdgeCount = static_cast< int >( myEdgesSrc.size() );

  // Gather counts from all ranks
  array1d< int > allEdgeCounts( numRanks );
  MpiWrapper::allgather( &myEdgeCount, 1, allEdgeCounts.data(), 1, comm );

  // Compute displacements for symmetrization
  array1d< int > edgeDispls( numRanks + 1 );
  edgeDispls[0] = 0;
  std::partial_sum(allEdgeCounts.begin(), allEdgeCounts.end(),
                 edgeDispls.begin() + 1);

```

src/coreComponents/mesh/generators/VTKSuperCellPartitioning.cpp
```diff
@@ -0,0 +1,1052 @@
...
+                                 rank, i, vertexWeights[i] ) );
+      }
+    }
+  }
+
+  // -----------------------------------------------------------------------
+  // Check 6: Graph symmetry for local edges
+  // -----------------------------------------------------------------------
+  {
+    std::unordered_set< uint64_t > localOutgoingNeighbors;
```
jafranc: can we refactor this duplication into a single locally declared function ?

src/coreComponents/mesh/generators/VTKSuperCellPartitioning.cpp
```diff
@@ -0,0 +1,1052 @@
...
+  }
+
+}
+
+
+// =============================================================================================
+// SECTION 5: PARTITION UNPACKING (after ParMETIS)
+// =============================================================================================
+array1d< int64_t >
+unpackSuperCellPartitioning(
```
jafranc: as `packing/unpacking` is rather specific in GEOS context, unless related, I would suggest renaming

src/coreComponents/mesh/generators/VTKSuperCellPartitioning.cpp
```diff
@@ -0,0 +1,1052 @@
...
+
+  vtkIdTypeArray * globalIds =
+    vtkIdTypeArray::SafeDownCast( cells3D->GetCellData()->GetGlobalIds() );
+
+  GEOS_ERROR_IF( !globalIds, "3D mesh missing global IDs" );
+
+  // -----------------------------------------------------------------------
+  // Step 1: Build 3D cell connectivity graph via fractures
+  // -----------------------------------------------------------------------
+  std::map< vtkIdType, std::set< vtkIdType > > fractureGraph;
```
jafranc: isn't the use of `stdMap` rather than `std::map` a directive ?

src/coreComponents/mesh/generators/VTKSuperCellPartitioning.cpp
```diff
@@ -0,0 +1,1052 @@
...
+    }
+  }
+
+  vtkIdType totalSplitSuperCells = MpiWrapper::sum( numSplitSuperCells, comm );
+
+  GEOS_ERROR_IF( totalSplitSuperCells > 0,
+                 GEOS_FMT( "Partitioning failed: {:L} super-cells were split across ranks!",
+                           totalSplitSuperCells ) );
+
+  return cellPartitioning;
```
jafranc, line 1031: Can't we group these do-check architecture into a single loop leveraging set's lookup insertion idiom (like above) and disjunction on the second value of the tuple ?

jafranc (approved):
Great ! Thanks, some shortening suggestion here and there

src/coreComponents/integrationTests/meshTests/testVTKImport.cpp
bd713: Made more explicit

src/coreComponents/mesh/generators/CollocatedNodes.cpp
bd713: Done

src/coreComponents/mesh/generators/ParMETISInterface.cpp
bd713: yes

src/coreComponents/mesh/generators/VTKUtilities.cpp
bd713: yes, that part was refactored. 
@jafranc, could you review the changes and confirm that they correctly addressed your comments.

src/coreComponents/mesh/generators/VTKSuperCellPartitioning.cpp
bd713: Duplication removed

src/coreComponents/mesh/generators/VTKSuperCellPartitioning.cpp
bd713: renamed `expandSuperCellPartitioningToCells`

src/coreComponents/mesh/generators/VTKSuperCellPartitioning.cpp
bd713: yes, changed all of them in favor of StdContainerWrappers

src/coreComponents/mesh/generators/VTKSuperCellPartitioning.cpp
bd713: done

src/coreComponents/mesh/generators/VTKSuperCellPartitioning.cpp
bd713: done

src/coreComponents/mesh/generators/VTKUtilities.cpp
bd713: done

src/coreComponents/mesh/generators/VTKSuperCellPartitioning.cpp
bd713: used `GetValueRange`

src/coreComponents/mesh/generators/ParMETISInterface.cpp
bd713: image

Tried this, but it makes the unit tests crash. 
But will protect the division.

src/coreComponents/mesh/generators/VTKSuperCellPartitioning.cpp
bd713: all changed to `localVar`

jafranc (approved):

src/coreComponents/mesh/generators/VTKUtilities.cpp
jafranc: Exactly what I meant ✔️

src/coreComponents/mesh/generators/ParMETISInterface.cpp
jafranc: Looks like it shouldn't by implementation
https://github.com/KarypisLab/ParMETIS/blob/8ee6a372ca703836f593e3c450ca903f04be14df/libparmetis/weird.c#L17-L60

src/coreComponents/mesh/generators/VTKSuperCellPartitioning.hpp
```diff
@@ -0,0 +1,167 @@
...
+
+/**
+ * @brief Super-cell metadata for constrained partitioning
+ *
+ * Groups cells that must remain co-located (e.g., fracture-connected cells).
+ */
+struct SuperCellInfo
+{
+  /// SuperCellId to global cell IDs
+  std::map< vtkIdType, std::vector< vtkIdType > > superCellToOriginalCells;
```
castelletto1: Should `stdMap` (or `map`) be used instal of `std::map` (see [here](https://geosx-geosx.readthedocs-hosted.com/en/latest/docs/sphinx/developerGuide/Contributing/CodeGeosFeatures.html#std-standard-library-usage))? Same for `std ::vector` (`stdVector`)?

castelletto1:
@OmarDuran, @MelReyCG : can you please take a look at this PR?

PR #4018, opened 2026-04-04
refactor: Change nodeset import from Int64 to unsigned integer

castelletto1:
**Changes**
- Switched nodeset arrays from `Int64` to unsigned integer (`UInt8/16/32/64`, `UInt8` recommended)
- Replaced virtual function calls with direct memory access
- Improved nodeset import performance

**Requirements**
- Nodesets now expect unsigned integer arrays (binary membership flags: node belongs or doesn't belong)
- VTK files must provide nodesets as unsigned integer arrays, any other format arrays will be rejected

**Notes**
- Current implementation still reads full `PointData` arrays (one per nodeset)
- Further optimization: read only node indices directly to avoid processing all mesh points


src/coreComponents/mesh/generators/VTKUtilities.cpp
```diff
@@ -2846,24 +2845,52 @@ void importNodesets( integer const logLevel,
...
-                   GEOS_FMT( "Target nodeset '{}' not found in mesh", nodesetNames[i] ),
+                   GEOS_FMT( "Target nodeset '{}' not found in mesh",
+                             nodesetNames[i] ),
                    InputError );
-    vtkTypeInt64Array const & nodesetMask = *vtkTypeInt64Array::FastDownCast( curArray );
 
-    SortedArray< localIndex > & targetNodeset = nodeSets.get_inserted( nodesetNames[i] );
-    for( localIndex j=0; j < numPoints; ++j )
+    // Explicit type check: must be UInt8 (VTK_UNSIGNED_CHAR)
+    GEOS_THROW_IF( curArray->GetDataType() != VTK_TYPE_UINT8,
```
dkachuma: Seems a bit stringent to throw. I would say we can have any (unsigned) integer type and internally cast to 1 byte perhaps with a warning. I think UINT8 is actually difficult to visualise in some versions of paraview and some may prefer to use UINT32.

dkachuma (approved):

src/coreComponents/mesh/generators/VTKUtilities.cpp
castelletto1: OK, now it is now recommended to use `UInt8`, but any unsigned integer type will work.

src/coreComponents/mesh/generators/VTKUtilities.cpp
```diff
@@ -2829,14 +2829,87 @@ string buildCellBlockName( ElementType const type, int const regionId )
...
+
+  // Bitmask for membership (Least Significant Bit)
+  constexpr ValueType mask = 1;
+
+  // Step 1: Extract binary flags from mask (parallel)
+  forAll< parallelHostPolicy >( numPoints, [rawData, &positions, mask]( localIndex const j )
+  {
+    positions[j] = rawData[j] & mask;
+  } );
+
```
jafranc:
```suggestion

  // Step 1: Extract binary flags from mask (parallel)
  forAll< parallelHostPolicy >( numPoints, [rawData, &positions, mask]( localIndex const j )
  {
    positions[j] = ( rawData[j] == 1 );
  } );

```

To me the bitwise op. is confusing and can give the false impression that numbering points set incrementally would results in any kind of sequential treatment while it is not

src/coreComponents/mesh/generators/VTKUtilities.cpp
```diff
@@ -2829,14 +2829,87 @@ string buildCellBlockName( ElementType const type, int const regionId )
...
+  // Step 4: Allocate exact-size output array
+  array1d< localIndex > nodeIndices( count );
+
+  // Step 5: Scatter flagged node indices to compacted array (parallel)
+  forAll< parallelHostPolicy >( numPoints, [rawData, &nodeIndices, &positions, mask]( localIndex const j )
+  {
+    if( rawData[j] & mask )
+    {
+      nodeIndices[ positions[j] ] = j;
+    }
```
jafranc:
```suggestion
    if( rawData[j] )
    {
      nodeIndices[ positions[j] ] = j;
    }
```

jafranc (approved):
Great first step towards direct import

src/coreComponents/mesh/generators/VTKUtilities.cpp
castelletto1: Done. Any unsigned integer type is supported now.

src/coreComponents/mesh/generators/VTKUtilities.cpp
castelletto1: Done

PR #4012, opened 2026-04-01
feat: add PermeabilitySpecification as an high-level FieldSpecification

kdrienCG:
This PR proposes to add an easy way to add higher-level components that create FieldSpecifications without the boilerplate.

___

As an example, to define the permeability you have to specify 3 differents FieldSpecifications.
One for the x-axis, a second for the y-axis, and a last one for the z-axis.
There is also some implementation details that could be hidden, like the `component` XML attribute for example.

```xml
<FieldSpecifications>
  <FieldSpecification
      name="permx"
      component="0"
      initialCondition="1"
      setNames="{ all }"
      objectPath="ElementRegions/reservoir"
      fieldName="rockPerm_permeability"
      functionName="permxFunc"
      scale="9.869233e-16"/>

  <FieldSpecification
      name="permy"
      component="1"
      initialCondition="1"
      setNames="{ all }"
      objectPath="ElementRegions/reservoir"
      fieldName="rockPerm_permeability"
      functionName="permyFunc"
      scale="9.869233e-16"/>

  <FieldSpecification
      name="permz"
      component="2"
      initialCondition="1"
      setNames="{ all }"
      objectPath="ElementRegions/reservoir"
      fieldName="rockPerm_permeability"
      functionName="permzFunc"
      scale="9.869233e-16"/>
</FieldSpecifications>
```

With this PR, their would be a `PermeabilitySpecification` component that would create _behind the scene_ those three FieldSpecifications to keep full compatibility with other parts that rely on `FieldSpecificationBase`. One could think of PermeabilitySpecification as a blueprint or recipe to create FieldSpecificationBase objects.

Here is how it looks like for `PermeabilitySpecification`:

```xml
<FieldSpecifications>
    <PermeabilitySpecification
      name="perm"
      setNames="{ all }"
      regionNames="{ reservoir }"
      fieldName="rockPerm_permeability"
      functionName="permFunc"
      scales="{9.869233e-16, 9.869233e-16, 9.869233e-16}"/>
</FieldSpecifications>
```

___

The following architecture is proposed:

image

This way, similar components could be added by providing a data class (like `PermeabilitySpecification`) and a factory with a `generate` method to describe how to transform the high-level components into one or multiple "low-level" `FieldSpecificationBase`.

PR #4019, opened 2026-04-07
feat: add PermeabilitySpecification as an high-level FieldSpecification

kdrienCG:
(Previously #4012)
(Discussion: #4092)

This PR proposes to add an easy way to add higher-level components that create FieldSpecifications without the boilerplate.

The goal of this component, along with future specialized field specifications (temperature, porosity, etc.), is to  improve the experience for both the users and developers.
 
Compared to the standard `FieldSpecification`, these components would:
- Expose only relevant XML attributes, with documentation tailored for its specific `fieldName`. It would also benefit from auto-generated documentation (see #4063 that can even show value limits and enable automatic validation).
- Establish a standard way to document user-facing fields. As of now, when a developer adds a field he must not forget to document if it is usable by the user. We want to provide a standard way to do so.
- Provide specialized validation directly in the data, limiting the user to entering only coherent values (via `postInputInitialization()` for example).
- Hide GEOS implementation details from users. For example, each specialized specification already knows whether it targets cells, nodes, edges, or faces. When multiple targets are valid, the choice can be exposed as an enum-string instead of requiring users to learn some implementation details to use it.

The `Group` system supports all of this: it gives developers a single place to document, validate, and optionally process user input. That is why this architecture is built on `Group`.

Hiding application data behind an interface is a common pattern that:
- improves stability of input files when the implementation changes,
- reduces the risk of users silently breaking a simulation, and
- avoids teaching internal details such as "_do not use the `initialPressure` fieldname, it is auto-filled by the pressure field; use `pressure` with `initialCondition=1`_" or "_do not touch `deltaPressure` it is internal._"
 
The flexibility offered by `FieldSpecification` is *not* affected. The goal is not to remove it, but to provide a safer default alternative for common use cases, while still keeping `FieldSpecification` available for advanced user, scenarios that require explicit data access and testing cases. Specialized components are built on top of it to ensure testing and support.
 
More specialized field specifications will follow. For example, component fractions with constraints such as sum = 1.0, implemented by creating multiple `FieldSpecification` entries under the hood.

___

As an example, to define the permeability you have to specify 3 differents FieldSpecifications.
One for the x-axis, a second for the y-axis, and a last one for the z-axis.
There is also some implementation details that could be hidden, like the `component` XML attribute for example.

```xml
<FieldSpecifications>
  <FieldSpecification
      name="permx"
      component="0"
      initialCondition="1"
      setNames="{ all }"
      objectPath="ElementRegions/reservoir"
      functionName="permxFunc"
      scale="9.869233e-16"/>

  <FieldSpecification
      name="permy"
      component="1"
      initialCondition="1"
      setNames="{ all }"
      objectPath="ElementRegions/reservoir"
      functionName="permyFunc"
      scale="9.869233e-16"/>

  <FieldSpecification
      name="permz"
      component="2"
      initialCondition="1"
      setNames="{ all }"
      objectPath="ElementRegions/reservoir"
      functionName="permzFunc"
      scale="9.869233e-16"/>
</FieldSpecifications>
```

With this PR, their would be a `PermeabilitySpecification` component that would create _behind the scene_ those three FieldSpecifications to keep full compatibility with other parts that rely on `FieldSpecificationBase`. One could think of PermeabilitySpecification as a blueprint or recipe to create FieldSpecificationBase objects.

Here is how it looks like for `PermeabilitySpecification`:

```xml
<FieldSpecifications>
    <PermeabilitySpecification
      name="perm"
      setNames="{ all }"
      regionNames="{ reservoir }"
      permeabilityModelName="rockPerm"
      functionNames="permFunc"
      scales="{9.869233e-16, 9.869233e-16, 9.869233e-16}"/>
</FieldSpecifications>
```

___

The following architecture is proposed:

ps_processors_architecture_1


rrsettgast:
Hi @kdrienCG 
This is something we have been meaning to get to for a while. But not just for perm. Can you implement a generalization to allow for processing and application of arbitrary tensor inputs? There are many things that this would be useful for besides permeability. I think we could read in a "scale" as any array type, and apply appropriately. Then this could be used for stresses, displacements, or any other non-scalar field without multiple FieldSpecificaiton blocks.

kdrienCG:
> Hi @kdrienCG This is something we have been meaning to get to for a while. But not just for perm. Can you implement a generalization to allow for processing and application of arbitrary tensor inputs? There are many things that this would be useful for besides permeability. I think we could read in a "scale" as any array type, and apply appropriately. Then this could be used for stresses, displacements, or any other non-scalar field without multiple FieldSpecificaiton blocks.

Hi @rrsettgast , and thanks for the suggestion
I agree there should be a generalized way to do this, beyond permeability.

I propose doing that in a follow-up PR so we can keep this one focused on specialized components and easier to review.

I've also put together a prototype to test the generalized approach, I can open a separate PR so we can agree on what its interface should be.

src/coreComponents/fieldSpecification/docs/FieldSpecification.rst
```diff
@@ -9,4 +9,6 @@ Initial and Boundary Conditions
    EquilibriumInitialCondition
 
    AquiferBoundaryCondition
+   
+   PermeabilitySpecification
```
MelReyCG, line 13: Hmm.. I don't see that in the generated readthedocs build.

src/coreComponents/fieldSpecification/docs/PermeabilitySpecification.rst
```diff
@@ -0,0 +1,49 @@
...
+The following illustrates a **PermeabilitySpecification** for a single region.
+
+.. code-block:: xml
+
+   <FieldSpecifications>
+     ...
+     <PermeabilitySpecification
+       name="perm"
+       setNames="{ all }"
+       regionNames="{ reservoir/block1 }"
```
MelReyCG, line 28: Can we do that?
```suggestion
       regionNames="{ reservoir }"
```

And what about that to target all region? (maybe you'd have to implement something on ``*`` in the factory)
```suggestion
       regionNames="{ * }"
```

Maybe we should these syntaxes in the examples.

src/coreComponents/fieldSpecification/docs/PermeabilitySpecification.rst
```diff
@@ -0,0 +1,49 @@
...
+
+####################################################
+Permeability specification
+####################################################
+
+Overview
+======================
+
+**PermeabilitySpecification** is an optional, higher-level XML tag that you can place in the **FieldSpecifications** block. 
+It describes a 3-axis permeability on one or more element regions.
```
MelReyCG, line 11:
```suggestion
**PermeabilitySpecification** is a XML tag that you can place in the **FieldSpecifications** block to describe a 3-axis permeability on one or more element regions.
It is an higher-level component than **FieldSpecification**, specialized on permeability.
```

rrsettgast:
@kdrienCG I would really insist that don't create a dedicated FieldSpecification for permeability. This should be handled in the generic way, with an improvement on the processing of "scale" to allow for tensor input, and the subsequent processing.

src/coreComponents/fieldSpecification/FieldSpecificationBase.hpp
```diff
@@ -524,6 +504,24 @@ class FieldSpecificationBase : public dataRepository::Group
...
   }
 
+  /**
+   * Mutator
+   * @param[in] component The component index
+   */
+  void setComponent( int component )
+  {
+    m_component = component;
+  }
```
MelReyCG, line 514: 1 liners (only the methods you added, as you did the rest in another PR)

src/coreComponents/fieldSpecification/FieldSpecificationABC.hpp
```diff
@@ -0,0 +1,107 @@
...
+ * @class FieldSpecificationABC
+ *
+ * Abstract Base Class grouping multiple types of field specifications.
+ */
+class FieldSpecificationABC : public dataRepository::Group
+{
+public:
+
+  /**
+   * @defgroup alias and functions to defined statically initialized catalog
```
MelReyCG, line 44:
```suggestion
   * @defgroup alias and functions to define statically initialized catalog
```

src/coreComponents/fieldSpecification/docs/FieldSpecification.rst
MelReyCG: maybe its an issue due to the aborting of the jobs on the last CI compilation

src/coreComponents/fieldSpecification/PermeabilitySpecificationFactory.cpp
```diff
@@ -0,0 +1,73 @@
...
+void PermeabilitySpecificationFactory::generate( FieldSpecificationABC const & specification,
+                                                 dataRepository::Group & manager ) const
+{
+  auto ps = dynamic_cast< PermeabilitySpecification const * >( &specification );
+
+  stdArray< string, 3 > suffixes = {{ "_x", "_y", "_z" }};
+
+  R1Tensor scales = ps->getScales();
+
+  for( string const & regionName : ps->getRegionNames() )
```
MelReyCG, line 43: Is it possible to add a validation of the region names?

MelReyCG:
Hi @rrsettgast, the goal here was to offer a user-oriented *interface*, as `FieldSpecification` is more developper oriented with no real interface between user configuration and internal GEOS data.
- `objectPath` gives access to all the mesh data repository, what one can use for a given field is not documented, and per-field validation would be tricky to integrate directly in it (moreover, the data-hierarchy is coupled with all user inputs),
- `fieldName` exposes data that could be considered as internal for most users, and if a user does not dive into all GEOS example, no clear list of them really exists anywhere (if you don't open all the fields headers, which are developper oriented).

I understand that this very proposal of the idea of specialized user-oriented components for each common field can seem quite empty for now, as there is no validation for the field for now, but it offer an opportunity to add specialized validations, dedicated documentation and a stable interface.
Maybe it is not the best approach, or maybe this component is not the best example, but we would like to give it a try. At term more components would be added, for high-level user configuration usecases.

Adding a vectorized version of the `FieldSpecification` is something we planned to do anyway, I'll be away for one month, in the meantime, @kdrienCG can prioritize the development of the vectorized `FieldSpecification`, which would then base this PR on.

src/coreComponents/fieldSpecification/PermeabilitySpecificationFactory.cpp
```diff
@@ -0,0 +1,73 @@
...
+ */
+
+void PermeabilitySpecificationFactory::generate( FieldSpecificationABC const & specification,
+                                                 dataRepository::Group & manager ) const
+{
+  auto ps = dynamic_cast< PermeabilitySpecification const * >( &specification );
+
+  stdArray< string, 3 > suffixes = {{ "_x", "_y", "_z" }};
+
+  R1Tensor scales = ps->getScales();
```
MelReyCG, line 41: We would need to ``GEOS_ERROR()`` on negative values.

MelReyCG:
Also, we don't have region name array on `FieldSpecification`, it is also an improvement that this PR brings. We also could add that on the `FieldSpecification`.

rrsettgast:
> Hi @rrsettgast, the goal here was to offer a user-oriented _interface_, as `FieldSpecification` is more developper oriented with no real interface between user configuration and internal GEOS data.
> 

I don't understand what exactly is meant when you say that FieldSpecification has "no real interface between user configuration and internal GEOS data. It is the specification of "field data". I read it as an implication that a "user" will not understand the concept that permeability is stored as a field. This is pretty basic.

> * `objectPath` gives access to all the mesh data repository, what one can use for a given field is not documented, and per-field validation would be tricky to integrate directly in it (moreover, the data-hierarchy is coupled with all user inputs),

This seems like it is a documentation issue. 

I do not understand what you mean by "per-field validation".  Do you mean checking that the specification makes sense? The criteria for this check is not determined at the specification action level.

The coupling of data with user inputs is intentional.

> * `fieldName` exposes data that could be considered as internal for most users, and if a user does not dive into all GEOS example, no clear list of them really exists anywhere (if you don't open all the fields headers, which are developper oriented).
> 
This is a documentation issue. How will the user know they can use a "PermeabilitySpecification" block, and what the available attributes/inputs are for that block? I guess they would have to find the specification somewhere in the documentation or examples? Is this any harder than finding the examples with "FieldSpecification" setting the permeability field?

Also, lucky for us that claude or codex/ChatGPT can easily figure this stuff out now. Just give the agent access to the examples and documentation and it will do it for you.

> I understand that this very proposal of the idea of specialized user-oriented components for each common field can seem quite empty for now, as there is no validation for the field for now, but it offer an opportunity to add specialized validations, dedicated documentation and a stable interface. Maybe it is not the best approach, or maybe this component is not the best example, but we would like to give it a try. At term more components would be added, for high-level user configuration usecases.
> 
Can you give examples of the type of validation that must occur specifying a field? Why must it be a stand alone class in field specification rather than validation based on field name and specified solvers? I think it is correct to say that the solvers should be the ones decided what input specifications are valid not the actual FieldSpecification, and a mechanism exist for this already. Do the "Initialization" functions from Group not do this?

> Adding a vectorized version of the `FieldSpecification` is something we planned to do anyway, I'll be away for one month, in the meantime, @kdrienCG can prioritize the development of the vectorized `FieldSpecification`, which would then base this PR on.

This really would solve a lot of "user interface" issues. However again, claude and codex handle the replication for you without any error.

src/coreComponents/fieldSpecification/PermeabilitySpecificationFactory.cpp
kdrienCG: Added in [#588cc90](https://github.com/GEOS-DEV/GEOS/pull/4019/commits/588cc9006652df838f20bdb079667cbbaf2b8d6e)

src/coreComponents/fieldSpecification/FieldSpecificationBase.hpp
kdrienCG: Resolved with merge commit [cd4a134](https://github.com/GEOS-DEV/GEOS/pull/4019/commits/cd4a13465e56a053651fc8ce4a0978c23111419c)

src/coreComponents/fieldSpecification/FieldSpecificationABC.hpp
```diff
@@ -0,0 +1,107 @@
...
+
+namespace geos
+{
+class Function;
+
+
+/**
+ * @class FieldSpecificationABC
+ *
+ * Abstract Base Class grouping multiple types of field specifications.
```
MelReyCG:
```suggestion
 * Abstract Base Class for concrete field modifiers (`FieldSpecification`) and high-level user-defined field specification.
```

src/coreComponents/fieldSpecification/PermeabilitySpecification.hpp
```diff
@@ -0,0 +1,217 @@
...
+  PermeabilitySpecification( PermeabilitySpecification const & ) = delete;
+
+  /// Defaulted move constructor
+  PermeabilitySpecification( PermeabilitySpecification && ) = default;
+
+  /// deleted copy assignment
+  PermeabilitySpecification & operator=( PermeabilitySpecification const & ) = delete;
+
+  /// deleted move assignement
+  PermeabilitySpecification & operator=( PermeabilitySpecification && ) = delete;
```
MelReyCG: is it possible to remove that as it is already defined in `FieldSpecificationABC`?

src/coreComponents/fieldSpecification/PermeabilitySpecification.cpp
```diff
@@ -0,0 +1,168 @@
...
+    setInputFlag( InputFlags::REQUIRED ).
+    setSizedFromParent( 0 ).
+    setDescription( "Names of sets that the boundary condition is applied to.\n"
+                    "A set can contain heterogeneous elements in the mesh (volumes, nodes, faces, edges).\n"
+                    "A set can be be defined by a 'Geometry' component, or correspond to imported sets in case of an external mesh" );
+
+  registerWrapper( viewKeyStruct::regionNamesString(), &m_regionNames ).
+    setRTTypeName( rtTypes::CustomTypes::groupNameRefArray ).
+    setInputFlag( InputFlags::REQUIRED ).
+    setDescription( "Names of the regions that boundary condition is applied to." );
```
MelReyCG: 📝
```suggestion
  registerWrapper( viewKeyStruct::regionNamesString(), &m_regionNames ).
    setRTTypeName( rtTypes::CustomTypes::groupNameRefArray ).
    setInputFlag( InputFlags::REQUIRED ).
    setDescription( "Names of the cell regions that boundary condition is applied to." );
```

src/coreComponents/fieldSpecification/PermeabilitySpecification.cpp
```diff
@@ -0,0 +1,168 @@
...
+                 getDataContext() );
+}
+
+template<>
+void expandFieldSpecification< PermeabilitySpecification >( PermeabilitySpecification const & permSpec,
+                                                            dataRepository::Group & manager )
+{
+  for( string const & regionName : permSpec.getRegionNames() )
+  {
+    string const objectPath = "ElementRegions/" + regionName;
```
MelReyCG: 📝 We need a validation here:
- path with region is valid
- type of region is `CellElementRegion`

src/coreComponents/fieldSpecification/PermeabilitySpecification.cpp
```diff
@@ -0,0 +1,168 @@
...
+  registerWrapper( viewKeyStruct::regionNamesString(), &m_regionNames ).
+    setRTTypeName( rtTypes::CustomTypes::groupNameRefArray ).
+    setInputFlag( InputFlags::REQUIRED ).
+    setDescription( "Names of the regions that boundary condition is applied to." );
+
+  registerWrapper( viewKeyStruct::fieldNameString(), &m_fieldName ).
+    setRTTypeName( rtTypes::CustomTypes::groupNameRef ).
+    setInputFlag( InputFlags::OPTIONAL ).
+    setDescription( "Name of field that boundary condition is applied to.\n"
+                    "A field can represent a physical variable. (pressure, temperature, global composition fraction of the fluid, ...)" );
```
MelReyCG: 📝 Should not be customizable as it is always `permeability`.
`getFieldName()` could be pure virtual and read from the `DECLARE_FIELD` struct.

src/coreComponents/fieldSpecification/PermeabilitySpecification.cpp
```diff
@@ -0,0 +1,168 @@
...
+  registerWrapper( viewKeyStruct::errorSetModeString(), &m_emptySetErrorMode ).
+    setInputFlag( InputFlags::OPTIONAL ).
+    setApplyDefaultValue( FieldSpecification::SetErrorMode::error ).
+    setDescription( GEOS_FMT( "Set the log state when a “set” does not target any region\n"
+                              "When set to \"{}\", no output.\n"
+                              "When set to \"{}\", output a warning.\n"
+                              "When set to \"{}\", output a throw.\n",
+                              EnumStrings< FieldSpecification::SetErrorMode >::toString( FieldSpecification::SetErrorMode::silent ),
+                              EnumStrings< FieldSpecification::SetErrorMode >::toString( FieldSpecification::SetErrorMode::warning ),
+                              EnumStrings< FieldSpecification::SetErrorMode >::toString( FieldSpecification::SetErrorMode::error )  ));
```
MelReyCG: 📝 As they will always be present, why not inheriting them from the ABC?

src/coreComponents/fieldSpecification/PermeabilitySpecification.cpp
```diff
@@ -0,0 +1,168 @@
...
+    setRTTypeName( rtTypes::CustomTypes::groupNameRef ).
+    setInputFlag( InputFlags::OPTIONAL ).
+    setDescription( "Name of field that boundary condition is applied to.\n"
+                    "A field can represent a physical variable. (pressure, temperature, global composition fraction of the fluid, ...)" );
+
+  registerWrapper( viewKeyStruct::componentString(), &m_component ).
+    setApplyDefaultValue( -1 ).
+    setInputFlag( InputFlags::OPTIONAL ).
+    setDescription( "Component of field (if tensor) to apply boundary condition to.\n"
+                    "The component must use the order in which the phaseNames have been defined in the Constitutive Element." );
```
MelReyCG: 📝 I don't think this one is useful for the `PermeabilitySpecification` class.

MelReyCG (changes requested):

src/coreComponents/fieldSpecification/PermeabilitySpecification.cpp
```diff
@@ -136,6 +137,27 @@ void expectValidCellRegion( DomainPartition const & domain,
...
+/**
+ * @brief Validate that a certain permeability model name exists in the domain
+ * @param domain The domain
+ * @param permeabilityModelName The name of the permeability model to validated
+ * @param permSpec Reference to the current object to print its DataContext
+ * @note Throws if the permeability model doesn't exists
+ */
+void expectValidPermeabilityModel( DomainPartition const & domain,
+                                   string const & modelName,
+                                   PermeabilitySpecification const & permSpec )
```
MelReyCG, line 128: ✌️

src/coreComponents/fieldSpecification/FieldSpecification.hpp
```diff
@@ -50,16 +49,9 @@ class FieldSpecification : public dataRepository::Group
...
-   */
-  enum class SetErrorMode : integer
-  {
-    silent,
-    error,
-    warning
-  };
+
+  // @copydoc FieldSpecificationABC::SetErrorMode
+  using SetErrorMode = FieldSpecificationABC::SetErrorMode;
```
MelReyCG: dead code

src/coreComponents/fieldSpecification/PermeabilitySpecification.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+  registerWrapper( viewKeyStruct::regionNamesString(), &m_regionNames ).
+    setRTTypeName( rtTypes::CustomTypes::groupNameRefArray ).
+    setInputFlag( InputFlags::REQUIRED ).
+    setDescription( "Names of the cell regions that boundary condition is applied to." );
+
+  registerWrapper( viewKeyStruct::permeabilityModelNameString(), &m_permeabilityModelName ).
+    setRTTypeName( rtTypes::CustomTypes::groupNameRef ).
+    setInputFlag( InputFlags::REQUIRED ).
+    setDescription( "Name of the constitutive permeability model." );
```
MelReyCG, line 41: 📝Optional, `ConstitutiveBase::makeFieldName` won't be called if leaved empty.

src/coreComponents/fieldSpecification/PermeabilitySpecification.hpp
```diff
@@ -0,0 +1,144 @@
...
+
+  /**
+   * Accessor
+   * @return const reference to m_fieldName
+   */
+  virtual string const & getFieldName() const override
+  {
+    static string const fieldName = fields::permeability::permeability::key();
+    return fieldName;
+  }
```
MelReyCG, line 109:
```suggestion
  virtual string_view getFieldName() const override
  { return fields::permeability::permeability::key(); }
```

src/coreComponents/fieldSpecification/PermeabilitySpecification.cpp
```diff
@@ -0,0 +1,190 @@
...
+ * @brief Validate that a region with the name @p regionName exists and is a valid cell region
+ * @param domain The domain
+ * @param regionName The name of the region to validated
+ * @param permSpec Reference to the current object to print its DataContext
+ * @note Throws if the region doesn't exists or isn't a cell region
+ */
+void expectValidCellRegion( DomainPartition const & domain,
+                            string const & fullRegionName,
+                            PermeabilitySpecification const & permSpec )
+{
```
MelReyCG, line 99: I think we can propose that as a standard method in the ABC + it has some sense to be a member function for the context retrieval.
```suggestion
template< typename ELEMENT_REGION_T >
void FieldSpecificationABC::expectValidCellRegion( DomainPartition const & domain,
                                                   string_view fullRegionName ) const
{
```

MelReyCG (changes requested):

src/coreComponents/fieldSpecification/PermeabilitySpecification.hpp
kdrienCG: A lot of call sites expect a string const &, it will be a subject for another PR if you agree.

PR #4021, opened 2026-04-08
feat: Add Young Modulus & Poisson import from VTK mesh

npillardou:
This feature adds register data fields for Poisson's ratio and Young Modulus directly from VTK mesh, avoiding any external conversion operations. 

It also offers the possibility to obtain them as output by adding (according to user needs) : 

`    <Outputs>
          <VTK 
               name="vtkOutput" 
               fieldNames="{ rock_youngModulus, rock_poissonRatio }" />  
    </Outputs>
`


src/coreComponents/constitutive/solid/ElasticIsotropic.cpp
```diff
@@ -147,6 +151,62 @@ void ElasticIsotropic::postInputInitialization()
...
+
+  localIndex numConverted = 0;
+  localIndex numInvalidE  = 0;
+
+  for( localIndex k = 0; k < m_youngModulus.size(); ++k )
+  {
+    // youngModulus default is 0: negative values are invalid, zero means not imported
+    if( youngMod[k] < 0.0 )
+    {
+      GEOS_WARNING( GEOS_FMT( "ElasticIsotropic '{}': element {} has negative Young's modulus ({:.6e}). "
```
dkachuma: This warning might flood the log if all the cells have negative values. I wouldn't issue the warning within the loop. Within the loop, simply count how many values are "bad" (perhaps pick out the worst case as well) and then outside the loop reduce over all ranks.

src/coreComponents/constitutive/solid/ElasticIsotropic.cpp
```diff
@@ -147,6 +151,62 @@ void ElasticIsotropic::postInputInitialization()
...
+                             "but Poisson's ratio ({:.6f}) is outside the valid range (-0.5, 0.5). "
+                             "Both youngModulus and poissonRatio must be provided together.",
+                             getName(), k, youngMod[k], nu[k] ) );
+
+    bulkMod[k]  = conversions::youngModAndPoissonRatio::toBulkMod( youngMod[k], nu[k] );
+    shearMod[k] = conversions::youngModAndPoissonRatio::toShearMod( youngMod[k], nu[k] );
+    ++numConverted;
+  }
+
+  GEOS_WARNING_IF( numInvalidE > 0,
```
dkachuma: Instead of issuing the warning on each rank, maybe sum `numInvalidE` over all ranks and issue a single warning on rank 0.

src/coreComponents/constitutive/solid/ElasticIsotropic.cpp
```diff
@@ -147,6 +151,62 @@ void ElasticIsotropic::postInputInitialization()
...
+      ++numInvalidE;
+      continue;
+    }
+
+    // youngModulus default is 0: not imported — skip silently
+    if( !( youngMod[k] > 0.0 ) )
+      continue;
+
+    // E was imported and is positive: nu must also be valid
+    GEOS_ERROR_IF( nu[k] <= -0.5 || nu[k] >= 0.5,
```
jhuang2601: For most common materials, Poisson’s ratio falls within the range of 0 to 0.5.

jhuang2601 (approved):

src/coreComponents/constitutive/solid/ElasticIsotropic.cpp
```diff
@@ -54,6 +54,10 @@ ElasticIsotropic::ElasticIsotropic( string const & name, Group * const parent ):
   registerField< fields::solid::bulkModulus >( &m_bulkModulus );
 
   registerField< fields::solid::shearModulus >( &m_shearModulus );
+
+  registerField< fields::solid::youngModulus >( &m_youngModulus );
+
+  registerField< fields::solid::poissonRatio >( &m_poissonRatio );
```
dkachuma, line 61: I think this causes a change to the `schema.xsd.other`. Can you upload this file as well please.

dkachuma (approved):
My understanding of (nu, E, K and G) is that we need only 2 of these. What combinations are currently supported. Currently you can provide K and G. You're proposing nu and E I suppose, but can we do any other combinations. Maybe just add a note in the documentation. For example what happens if the user tries to load K and E?

src/coreComponents/constitutive/solid/ElasticIsotropic.cpp
npillardou: I agree with you, but there was some checks already done around line 80-84  that were with this range : 
```
  if( nu > -0.5 && nu < 0.5 )
  {
    ++numConstantsSpecified;
    errorCheck += "nu, ";
  }
```

So I kept them as it  is.

herve-gross (approved):

PR #4027, opened 2026-04-10
fix: Fix inverse capillary pressure

dkachuma:
This PR fixes the inverse capillary pressure calculation for 3-phase cases where the oil phase disappears. A correction factor is applied to the Newton step to ensure that we do not evaluate the capillary pressure curves outside their bounds.


bd713 (approved):
Thanks for fixing it. I confirm that the test now passes

src/coreComponents/constitutive/capillaryPressure/InverseCapillaryPressure.hpp
```diff
@@ -290,7 +290,11 @@ class InverseCapillaryPressureUpdate
         real64 stepSize = 1.0;
         if( epsilon < LvArray::math::abs( dS ))
         {
-          real64 const phaseStepSize = LvArray::math::max((minSaturation[phase0] - saturation[phase0])/dS, (maxSaturation[phase0] - saturation[phase0])/dS );
+          real64 phaseStepSize = LvArray::math::max((minSaturation[phase0] - saturation[phase0])/dS, (maxSaturation[phase0] - saturation[phase0])/dS );
+          if( phaseStepSize < 1.0 )
+          {
+            phaseStepSize *= (1.0 - STEP_TOLERANCE);
+          }
```
rrsettgast, line 297: So if the phaseStepSize is less than 1.0, you are reducing by a small factor? Can you expand on how this helps?

dkachuma: The phaseStepSize is less than 1 if we don't want to go outside the bounds of the capillary pressure curve. Reducing it further ensures that we are truly away from the boundary. In some cases we were ending up on the boundary itself. If evaluated at the boundary, some of the capillary pressure curves give a zero derivative.

rrsettgast: sorry. but what is the effect of the guard `if( phaseStepSize < 1.0 )`. Is a phaseStepSize=1.0 ok?

rrsettgast (approved):

PR #4030, opened 2026-04-16
feat: Copy/archive input XML files into the output directory

kdrienCG:
(Previously #4003)

This PR proposes to archive the XML input deck to the output directory, so that every set of results is accompanied by the exact input that produced it.

When GEOS is run with the new archive command line option (`-a LEVEL`), all XML input files (and every files that are included via the `<Included>` XML tag) are flattened into a single file in a timestamped directory.

Following suggestions from @rrsettgast and @bd713, the XML input files are "flattened" into a single XML file. 

The following command line argument is added to specifiy the level of archiving wanted:
```
-a, --archive,  Archiving strategy's level:
                  0 = no archiving (default),
                  1 = XML inputs only,
                  2 = XML inputs and the XSD schema
```

---

Given a run like:
```
geosx -i foo.xml -o OUTPUT/ -a 2
```

The following is created:
```
.
└─ OUTPUT/   
   └─ archive_inputFiles/
      └─ 20260320_103034/
         ├─ input.xml
         └─ schema.xsd
```

---

<details>

<summary>Previous proposition</summary>

This PR proposes to automatically archive the XML input deck to the output directory, so that every set of results is accompanied by the exact input that produced it.

When GEOS is run with an output directory specified (`-o <dir>`), all XML input files (and every files that are included via the `<Included>` XML tag) are copied into the output directory with a timestamp.

---

Given a run like:
```
geosx -i foo.xml -o OUTPUT/
```

The following is created:
```
.
└─ OUTPUT/   
   └─ archive_inputFiles/
      └─ 20260320_103034/
         ├─ foo.xml
         │
         ├─ include1.xml
         │
         └─ some_subdir/
            └─ include2.xml
```

When included files are "behind" the input tree of the main XML, the following structure is proposed:
```
.
└─ OUTPUT/   
   └─ archive_inputFiles/
      └─ 20260320_103034/
         ├─ foo.xml
         │
         ├─ __one_level_behind_include.xml
         │
         ├─ __one_level_behind_folder/           
         |  └─ baz.xml
         │
         └─ ____two_level_behind_folder/      
            └─ buzz.xml
```
Files outside the input tree (reached via `../`) are prefixed with `__` for every `../` from `foo.xml`.

</details>


src/main/main.cpp
```diff
@@ -38,6 +39,11 @@ int main( int argc, char *argv[] )
 
     outputVersionInfo();
 
+    if( commandLineOptions->archiveInputDeck )
```
MelReyCG: to be more explicit, I would

```suggestion
    if( commandLineOptions->archiveInputDeck > 0 )
```

herve-gross:
@bd713 moving this to the review queue

src/coreComponents/fileIO/Outputs/ArchiveInputDeck.cpp
```diff
@@ -0,0 +1,252 @@
...
+    }
+  }
+
+  return flatDoc;
+}
+
+void copySchemaToArchive( string const & archiveDir )
+{
+  std::filesystem::path const candidates[] = {
+    GEOS_SCHEMA_SOURCE_PATH
```
bd713: Comma needed to actually construct the 2 candidates.
Otherwise schema archiving always fails in my tests.

```suggestion
    GEOS_SCHEMA_SOURCE_PATH,
```

src/coreComponents/dataRepository/xmlWrapper.cpp
```diff
@@ -268,6 +269,91 @@ string buildMultipleInputXML( string_array const & inputFileList,
   return inputFileName;
 }
 
+void collectIncluded( string const & filePath,
```
bd713: Are these 2 functions still needed for the current flattening strategy? 
If not, I would suggest removing them for now.

src/coreComponents/common/initializeEnvironment.hpp
```diff
@@ -93,6 +93,9 @@ struct CommandLineOptions
 
   /// Print memory usage in data repository
   real64 printMemoryUsage = -1.0;
+
+  /// Set the archiving level
+  integer archiveInputDeck = 1;
```
bd713: This feature will be incredibly useful in our production environment. That said, I would personally keep the existing GEOS default behavior (no archiving) unchanged for regular usage, CI, etc. 
Same for example as CSV output, which isn't on by default.

bd713 (commented):
Thanks, @kdrienCG. 

Could you also update the GEOS documentation accordingly? 
Typically [QuickStart.rst](https://geosx-geosx.readthedocs-hosted.com/en/latest/docs/sphinx/QuickStart.html#:~:text=USAGE%3A%20geosx,schema%20specified.%20Exiting...) for the new CLI flag, and maybe worth a brief description under [Advanced XML Features](https://geosx-geosx.readthedocs-hosted.com/en/latest/coreComponents/fileIO/doc/InputXMLFiles.html#advanced-xml-features).

src/coreComponents/dataRepository/xmlWrapper.cpp
kdrienCG: removed in [81084b8](https://github.com/GEOS-DEV/GEOS/pull/4030/changes/81084b8170f9eb935ce11bb0cd96abff4c1d7acc)

dkachuma (approved):

src/coreComponents/mainInterface/initialization.cpp
```diff
@@ -266,6 +268,12 @@ std::unique_ptr< CommandLineOptions > parseCommandLineOptions( int argc, char *
         ErrorLogger::global().createFile();
       }
       break;
+      case ARCHIVE:
+      {
+        integer const level = std::stoi( opt.arg );
```
jhuang2601, line 273: Please add a range check (only valid for 0/1/2) with error handling

src/coreComponents/common/initializeEnvironment.hpp
jhuang2601: I echo @bd713. Please keep `0 = no archiving` as default

jhuang2601 (changes requested):

src/coreComponents/fileIO/Outputs/ArchiveInputDeck.cpp
```diff
@@ -0,0 +1,252 @@
...
+    std::filesystem::path const destination = std::filesystem::path( archiveDir ) / "schema.xsd";
+    std::filesystem::copy_file( source,
+                                destination,
+                                ec );
+
+    if( ec )
+    {
+      GEOS_WARNING( GEOS_FMT( "Failed to copy XSD schema to archive '{}': {}",
+                              destination.string(), ec.message() ) );
+      return;
```
MelReyCG: Don't you want to `continue;` here?

src/coreComponents/fileIO/Outputs/ArchiveInputDeck.cpp
```diff
@@ -0,0 +1,252 @@
...
+string makeTimestamp()
+{
+  auto const now = std::chrono::system_clock::now();
+  auto const time_t_now = std::chrono::system_clock::to_time_t( now );
+  std::ostringstream timestampStream;
+  timestampStream << std::put_time( std::localtime( &time_t_now ), "%Y%m%d_%H%M%S" );
+  return timestampStream.str();
+}
+
+void stripMetadataAttributes( xmlWrapper::xmlNode node )
```
MelReyCG, line 53: Are you sure that you don't break something here? `xmlWrapper::xmlNode` is a wrapper around internal pugi data, I'm afraid you remove `xmlWrapper::filePathString` & `xmlWrapper::charOffsetString` definitively. Does it affect the line logging feature in `GEOS_ERROR`/`GEOS_WARNING`?

src/coreComponents/fileIO/Outputs/ArchiveInputDeck.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+}
+
+
+void archiveInputDeck( string_array const & inputFileNames,
+                       string const & outputDirectory,
+                       string_array const & xmlTagOrder,
+                       integer const level )
+{
+  if( level == 0 || inputFileNames.empty() || outputDirectory.empty() )
```
MelReyCG:
```suggestion
  if( level <= 0 || inputFileNames.empty() || outputDirectory.empty() )
```
?

src/coreComponents/fileIO/Outputs/ArchiveInputDeck.hpp
```diff
@@ -0,0 +1,56 @@
...
+#include "common/DataTypes.hpp"
+
+namespace geos
+{
+
+namespace archiveInputDeck
+{
+
+/**
+ * @brief Archive the XML input deck (and optionally the XSD schema) into the
```
MelReyCG, line 32: I would add a note that it should not be called on multiple ranks

MelReyCG (commented):
As there is no unitTest, how did you test this feature?

src/main/main.cpp
kdrienCG: I resolve it as it is not present anymore

src/coreComponents/fileIO/Outputs/ArchiveInputDeck.cpp
kdrienCG: `stripMetadataAttributes()` runs on a copy/separate XML doc (`flatDoc`). It should not affect the XML document used for input parsing

victorapm:
This is a nice feature, thanks! What else needs to be done so we can merge it into develop?

PR #4033, opened 2026-04-21
feat: Add new scatter strategies for initial mesh redistribution

bd713:
## Problem

The current strategy to scatter the mesh initially loaded by rank 0 uses `vtkRedistributeDataSetFilter` (kd-tree partitioning). This approach **can take several minutes for large meshes** (10s of millions of elements) when redistributed among 1000+ ranks.

**Recent example:** A 136M cell mesh redistributed to 1024 ranks takes about **17 minutes** with the current kd-tree approach.

## Solution

This PR adds partitioning strategies that bypass the VTK library's bottleneck, available via the new `scatterMethod` XML attribute.

### Available strategies

| Method | Description | Comment |
|--------|-------------|----------|
| `kdtree` | VTK built-in kd-tree (default) | Spatial locality, but can be slow and current version can lose tetra or pyramid cells |
| `contiguous` | Contiguous cell-index blocks | Fast, but does not guarantee spatial locality |
| `cartesian` | Regular Cartesian grid using `-x/-y/-z` partition counts | Does not guarantee load balancing if mesh is irregular |
| `rcb` | Recursive Coordinate Bisection along longest bounding-box axis | Spatial locality + load balance |

## Remark
The resulting partitions can be further refined and improved via calls to `ParMetis` or `PTScotch` with `partitionRefinement>0`

## Benchmark
```
==============================================================================================================
SUMMARY (1024 ranks, 135901500 cells)
==============================================================================================================
Method        | Time (s)  | Cells       | Status  | Min cells | Max cells | Imbalance   | BBox Overlap
--------------------------------------------------------------------------------------------------------------
KdTree        | 1020.194  | 135901494   | LOSS    | 126714    | 139764    | 1.053103    | 1.086       
Contiguous    |   15.822  | 135901500   | OK      | 132716    | 132717    | 1.000005    | 2.861       
Cartesian     |   22.005  | 135901500   | OK      | 116157    | 152266    | 1.147304    | 1.095       
RCB           |   34.289  | 135901500   | OK      | 132716    | 132717    | 1.000005    | 1.116       
--------------------------------------------------------------------------------------------------------------

Speedup relative to each method:
  KdTree        64.48x slower than fastest
  Contiguous    1.00x slower than fastest <-- fastest
  Cartesian     1.39x slower than fastest
  RCB           2.17x slower than fastest

Best load balance: Contiguous (imbalance ratio: 1.000005x)
Best compactness:  KdTree (bbox overlap: 1.086)
==============================================================================================================
```

## Rebaseline needed
57/67 failures are XML metadata diff: `Group has a child 'scatterMethod' in the file to compare but not the baseline file.`
10/67 are real partitioning changes in contact cases as we switched from Morton's algorithm to RCB


victorapm:
That looks great, Bertrand. Even though `contiguous` is the fastest, we might want to have `RCB` as default since `bbox` values look much better and the additional cost is not too bad. cc @castelletto1

dkachuma (approved):
Looks good to me.

joshua-white (approved):
Nice.  Only suggestion might be to add a log note if a "large" mesh and kdtree are detected, suggesting user consider something else.

